PR #236, opened 2017-06-28
Add SecondsToDurationConverter

arturs105:
This adds an `MvxValueConverter` to convert seconds from `TimeEntry.Duration` to a human readable duration in format `h:mm:ss`. Will be used in time entries log.


amulware:
I'm fine with hardcoding the time format for now, but later we'll want to take into account the time format chosen by the user most likely. Might require comment from @pe0ny

arturs105:
The format could be passed as the parameter for the converter. At the beginning I made it that way (see first commit), but there were some issues with durations longer than 24 hours.

PR #229, opened 2017-06-21
Create build configurations for unit and api tests

amulware:
Upsides:
- Unit, and especially api integration tests build significantly faster locally
- Unit tests run faster on Bitrise (2-3 minutes, instead of 5-6)

Downsides:
- With the current setup, Bitrise now does not try to compile Daneel, so any build errors there would not automatically get noticed
- Slightly more maintenance, and more complex build file

I think this might still be worth it, and we could always have a different flow run on Bitrise, to make sure everything compiles, if we like.

Thoughts?

PR #237, opened 2017-06-28
Create build configurations for unit and api tests V2

amulware:
Supersedes #229

As far as I can tell, it works great with @willsb's changes, both locally and on bitrise.

The switch to debug makes the biggest difference, but locally it will now default to only build what is needed for the tests you run, making it even faster.


amulware:
I skimmed the build log, which looks fine, and checked that it runs the same number of tests, which is does. And I indeed didn't have to change anything.

PR #238, opened 2017-06-28
Add time entry grouping by day for TimeEntriesViewModel

arturs105:


Toggl.Foundation.Tests/MvvmCross/ViewModels/TimeEntriesLogViewModelTests.cs
```diff
@@ -0,0 +1,125 @@
...
+            public async Task GroupsTimeEntriesFromTheSameDay()
+            {
+                DateTime[] dateTimes =
+                {
+                    new DateTime(2017, 6, 7, 12, 48, 0),
+                    new DateTime(2017, 6, 7, 10, 10, 10),
+                    new DateTime(2017, 6, 8, 15, 16, 17),
+                    new DateTime(2017, 6, 8, 16, 17, 18)
+                };
+				TimeEntry[] timeEntries =
```
willsb: Sneaky tabs 🕵️

Toggl.Foundation.Tests/MvvmCross/ViewModels/TimeEntriesLogViewModelTests.cs
```diff
@@ -0,0 +1,125 @@
...
+                DateTime[] dateTimes =
+                {
+                    new DateTime(2017, 5, 1, 1, 1, 1),
+                    new DateTime(2017, 5, 2, 1, 1, 1),
+                    new DateTime(2017, 5, 3, 1, 1, 1),
+                    new DateTime(2017, 5, 3, 1, 1, 2),
+                    new DateTime(2017, 5, 3, 1, 1, 3)
+                };
+                //Time entries are created out of order on purpose
+				TimeEntry[] timeEntries =
```
willsb: Tabs

Toggl.Foundation.MvvmCross/ViewModels/TimeEntriesLogViewModel.cs
```diff
@@ -0,0 +1,36 @@
...
+            this.dataSource = dataSource;
+
+            TimeEntries = new ObservableCollection<IGrouping<DateTime, ITimeEntry>>();
+        }
+
+        public async override Task Initialize()
+        {
+            await base.Initialize();
+            var timeEntries = await dataSource.TimeEntries.GetAll();
+            var groups = timeEntries.OrderBy(te => te.Start)
```
willsb: I believe this should be `OrderByDescending`

Toggl.Foundation.MvvmCross/ViewModels/TimeEntriesLogViewModel.cs
```diff
@@ -0,0 +1,36 @@
...
+                                    .GroupBy(te => te.Start.Date);
```
willsb: Does `GroupBy` guarantee that the elements inside the `IGrouping` will have the same order they had before being grouped?

Toggl.Foundation.MvvmCross/ViewModels/TimeEntriesLogViewModel.cs
```diff
@@ -0,0 +1,36 @@
...
+
+        public ObservableCollection<IGrouping<DateTime, ITimeEntry>> TimeEntries { get; }
+
+        public TimeEntriesLogViewModel(ITogglDataSource dataSource)
+        {
+            Ensure.Argument.IsNotNull(dataSource, nameof(dataSource));
+            
+            this.dataSource = dataSource;
+
+            TimeEntries = new ObservableCollection<IGrouping<DateTime, ITimeEntry>>();
```
willsb: Why not make this construction inline?

Toggl.Foundation.MvvmCross/ViewModels/TimeEntriesLogViewModel.cs
```diff
@@ -0,0 +1,36 @@
...
+        }
+
+        public async override Task Initialize()
+        {
+            await base.Initialize();
+            var timeEntries = await dataSource.TimeEntries.GetAll();
```
willsb: While I agree that `await`ing is the best approach here for now, we will have refreshing mechanisms after the sync algo is implemented, so keep in mind that this will eventually change.

Toggl.Foundation.Tests/MvvmCross/ViewModels/TimeEntriesLogViewModelTests.cs
willsb: Also: 
`var timeEntries = dateTimes.Select(date => new TimeEntry { Start = date });`

Toggl.Foundation.Tests/MvvmCross/ViewModels/TimeEntriesLogViewModelTests.cs
```diff
@@ -0,0 +1,125 @@
...
+                    new DateTime(2017, 6, 8, 16, 17, 18)
+                };
+				TimeEntry[] timeEntries =
+				{
+                    new TimeEntry { Start = dateTimes[0] },
+                    new TimeEntry { Start = dateTimes[1] },
+                    new TimeEntry { Start = dateTimes[2] },
+                    new TimeEntry { Start = dateTimes[3] }
+				};
+				var observable = Observable.Return(timeEntries);
```
willsb: More tabs here

Toggl.Foundation.Tests/MvvmCross/ViewModels/TimeEntriesLogViewModelTests.cs
```diff
@@ -0,0 +1,125 @@
...
+				TimeEntry[] timeEntries =
+                {
+                    new TimeEntry { Start = dateTimes[3] },
+                    new TimeEntry { Start = dateTimes[0] },
+					new TimeEntry { Start = dateTimes[2] },
+                    new TimeEntry { Start = dateTimes[1] },
+                    new TimeEntry { Start = dateTimes[4] },
+					new TimeEntry { Start = dateTimes[3] }
+				};
+				var observable = Observable.Return(timeEntries);
```
willsb: Tabs

Toggl.Foundation.MvvmCross/ViewModels/TimeEntriesLogViewModel.cs
amulware: It does.

amulware: However, I'm cringing a bit that it's not using a more efficient algorithm, if the list is already sorted. Probably not worth it making the code more complicated for, but if there is a solution of similar complexity, that would be cool.

willsb: Add a `OrderByAndGroup` extension 🐎

amulware: That method would take two different key selectors, not sure that's a good idea. I'd rather have a `SectionBy` extension method. But `GroupBy` works fine too, and the overhead will probably not be noticable. We can optimise later, make it work first.

Toggl.Foundation.Tests/MvvmCross/ViewModels/TimeEntriesLogViewModelTests.cs
```diff
@@ -0,0 +1,94 @@
...
+
+        public class TheTimeEntriesProperty : TimeEntriesLogViewModelTest
+        {
+            [Property]
+            public void OrdersGroupsProperly(DateTime[] dateTimes)
+            {
+                var timeEntries = dateTimes.Select(date => new TimeEntry { Start = date });
+                var observable = Observable.Return(timeEntries);
+                DataSource.TimeEntries.GetAll().Returns(observable);
+                ViewModel.TimeEntries.Clear();
```
arturs105: This `Clear` method works here, but if we are going to continue using FsCheck, then we need to rework the tests a bit, because this test method is called multiple times and the ViewModel object is the same between those calls. I think it would be better if each time the test is run ViewModel had a fresh state.

willsb:
After an internal discussion, I'm taking over this PR

willsb:
@amulware Please review this when you have some time

Toggl.Foundation.MvvmCross/ViewModels/TimeEntriesLogViewModel.cs
```diff
@@ -0,0 +1,39 @@
...
+            this.dataSource = dataSource;
+        }
+
+        public async override Task Initialize()
+        {
+            await base.Initialize();
+
+            (await dataSource
+                    .TimeEntries
+                    .GetAll())
```
amulware: I think it's better readable to have this await into a variable, and go on with that.

Toggl.Foundation.MvvmCross/ViewModels/TimeEntriesLogViewModel.cs
```diff
@@ -0,0 +1,39 @@
...
+using Toggl.Multivac.Models;
+
+namespace Toggl.Foundation.MvvmCross.ViewModels
+{
+    public class TimeEntriesLogViewModel : BaseViewModel
+    {
+        private readonly ITogglDataSource dataSource;
+
+        public ObservableCollection<IGrouping<DateTime, ITimeEntry>> TimeEntries { get; }
+            = new ObservableCollection<IGrouping<DateTime, ITimeEntry>>();
```
amulware: If one time entry changes, we have to change a whole grouping each time. Is this desired? In general, is there a plan for will changes will be propagated?

Toggl.Foundation.Tests/Generators/ViewModelGenerators.cs
```diff
@@ -0,0 +1,33 @@
...
+using Toggl.Foundation.MvvmCross.ViewModels;
+using Toggl.Multivac.Extensions;
+using TimeEntry = Toggl.Ultrawave.Models.TimeEntry;
+using FuncDate = System.Func<System.DateTimeOffset, System.DateTimeOffset>;
+
+namespace Toggl.Foundation.Tests.Generators
+{
+    public static class ViewModelGenerators
+    {
+        public static Gen<TimeEntriesLogViewModel> ForTimeEntriesLogViewModel(FuncDate applyToDates)
```
amulware: The type is called `Gen`? Brrrr

Toggl.Foundation.Tests/Generators/ViewModelGenerators.cs
```diff
@@ -0,0 +1,33 @@
...
+        {
+            return Arb.Default.Array<DateTimeOffset>().Generator.Select(dateTimes =>
```
amulware: Can you explain to me what this line does exactly?

Toggl.Foundation.Tests/MvvmCross/ViewModels/TimeEntriesLogViewModelTests.cs
```diff
@@ -0,0 +1,74 @@
...
+            }
+        }
+
+        public class TheTimeEntriesProperty
+        {
+            [Property]
+            public Property ShouldBeOrderedAfterInitialization()
+            {
+                var generator = ViewModelGenerators.ForTimeEntriesLogViewModel(d => d).ToArbitrary();
+                return Prop.ForAll(generator, viewModel =>
```
amulware: Can you link me to some docs or explain exactly how property testing works with this?

I was under the impression we could just take the viewModel, or maybe a list of time entries as argument for the test.

Toggl.Foundation.Tests/MvvmCross/ViewModels/TimeEntriesLogViewModelTests.cs
```diff
@@ -0,0 +1,74 @@
...
+
+            [Property]
+            public Property ShouldHaveOrderedGroupsAfterInitialization()
+            {
+                var generator =
+                    ViewModelGenerators
+                        .ForTimeEntriesLogViewModel(d =>
+                            new DateTime(
+                                DateTime.UtcNow.Year, DateTime.UtcNow.Month,
+                                d.Day, d.Hour, d.Minute, d.Second
```
amulware: Will this not break if sticking dates from a 31 day month into a shorter one? In general, using `Now` here, and doing so twice in a row (will return different numbers) is maybe not the best idea. In the end this should work for any month in any year after all.

Toggl.Foundation.Tests/MvvmCross/ViewModels/TimeEntriesLogViewModelTests.cs
```diff
@@ -0,0 +1,74 @@
...
+
+        public class TheTimeEntriesProperty
+        {
+            [Property]
+            public Property ShouldBeOrderedAfterInitialization()
+            {
+                var generator = ViewModelGenerators.ForTimeEntriesLogViewModel(d => d).ToArbitrary();
+                return Prop.ForAll(generator, viewModel =>
+                {
+                    viewModel.Initialize().GetAwaiter().GetResult();
```
amulware: Can't this (and below) just be `.Wait()` or something?

Toggl.Foundation.MvvmCross/ViewModels/TimeEntriesLogViewModel.cs
willsb: This is the current idea: Changing the IGrouping and then notifying the UI.

Toggl.Foundation.MvvmCross/ViewModels/TimeEntriesLogViewModel.cs
willsb: I disagree, but don't care enough 😄

Toggl.Foundation.Tests/Generators/ViewModelGenerators.cs
willsb: You must be new here

Toggl.Foundation.Tests/Generators/ViewModelGenerators.cs
willsb: Creates a generator that creates arrays of DateTimeOffset. The rest is LINQ

Toggl.Foundation.Tests/MvvmCross/ViewModels/TimeEntriesLogViewModelTests.cs
willsb: We could, yeah:

```
[Property( Arbitrary = new [] { typeof(ViewModelGenerators.ForTimeEntriesLogViewModelArbitrary) })]
public void ShouldBeOrderedAfterInitialization(TimeEntriesLogViewModel viewModel)
{
    //Test
}
```

But this lacks type safety we get here (we can basically pass any type to the attribute while here we have to ensure the generator is correct) and @arturs105 said the error reporting is not as nice (this gives some very nice error reporting, with shrinkers and everything)

Toggl.Foundation.Tests/MvvmCross/ViewModels/TimeEntriesLogViewModelTests.cs
willsb: It could, yes

Toggl.Foundation.Tests/MvvmCross/ViewModels/TimeEntriesLogViewModelTests.cs
willsb: This will break horribly, yes, I'll change it

Toggl.Foundation.MvvmCross/ViewModels/TimeEntriesLogViewModel.cs
amulware: If by changing you mean re-creating and replacing, ok, sure.

Toggl.Foundation.Tests/Generators/ViewModelGenerators.cs
amulware: And I assume, this being a default generator, it will basically generate random dates? Ok.

Toggl.Foundation.Tests/MvvmCross/ViewModels/TimeEntriesLogViewModelTests.cs
```diff
@@ -48,18 +48,18 @@ public Property ShouldHaveOrderedGroupsAfterInitialization()
...
                         ).ToArbitrary();
 
                 return Prop.ForAll(generator, viewModel =>
                 {
-                    viewModel.Initialize().GetAwaiter().GetResult();
+                    viewModel.Initialize().Wait();
                     foreach (var grouping in viewModel.TimeEntries)
-                    {
+                    {
+                        grouping.Count().Should().BeGreaterOrEqualTo(1);
```
amulware: Why this here?

Toggl.Foundation.Tests/MvvmCross/ViewModels/TimeEntriesLogViewModelTests.cs
```diff
@@ -48,18 +48,18 @@ public Property ShouldHaveOrderedGroupsAfterInitialization()
...
                 var generator =
                     ViewModelGenerators
-                        .ForTimeEntriesLogViewModel(d =>
-                            new DateTime(
-                                DateTime.UtcNow.Year, DateTime.UtcNow.Month,
-                                d.Day, d.Hour, d.Minute, d.Second
-                           )
+                        .ForTimeEntriesLogViewModel(
+                            d => d.Year == DateTime.UtcNow.Year 
+                              && d.Month == DateTime.UtcNow.Month
```
amulware: With a strict filter like this, is it smart enough to still generate significant numbers of dates?

Toggl.Foundation.Tests/MvvmCross/ViewModels/TimeEntriesLogViewModelTests.cs
willsb: To ensure we're not creating empty Groups

Toggl.Foundation.Tests/MvvmCross/ViewModels/TimeEntriesLogViewModelTests.cs
willsb: This is ensured by the `ArrayOf(100)` part of the generator

amulware: I'll be curious to see how long it takes to generate 100 random samples in the last month. :D

Toggl.Foundation.Tests/MvvmCross/ViewModels/TimeEntriesLogViewModelTests.cs
amulware: Shouldn't that be a different test?

willsb: Yeah, it should

Toggl.Foundation.Tests/MvvmCross/ViewModels/TimeEntriesLogViewModelTests.cs
```diff
@@ -0,0 +1,74 @@
...
+                var generator = ViewModelGenerators.ForTimeEntriesLogViewModel(d => d).ToArbitrary();
+                return Prop.ForAll(generator, viewModel =>
+                {
+                    viewModel.Initialize().Wait();
+
+                    for (int i = 1; i < viewModel.TimeEntries.Count(); i++)
+                    {
+                        var dateTime1 = viewModel.TimeEntries.ElementAt(i - 1).Key;
+                        var dateTime2 = viewModel.TimeEntries.ElementAt(i).Key;
+                        dateTime1.Should().BeOnOrAfter(dateTime2);
```
arturs105: I think we should use `BeAfter()`, because we are comparing time entry groups (by day), and if two groups have the same datetime, then something must be wrong, i.e. each group should have unique datetime.

arturs105:
Just came to my mind: doesn't the view model need a list of projects as well? Or any other way to get the project colors?

Toggl.Foundation.Tests/Generators/ViewModelGenerators.cs
```diff
@@ -0,0 +1,51 @@
...
+using TimeEntry = Toggl.Ultrawave.Models.TimeEntry;
+
+namespace Toggl.Foundation.Tests.Generators
+{
+    public static class ViewModelGenerators
+    {
+        public static Gen<TimeEntriesLogViewModel> ForTimeEntriesLogViewModel(MonthPredicate filter)
+        {
+            var monthsGenerator = Gen.Choose(1, 12).Where(filter);
+            var yearGenerator = Gen.Choose(2007, DateTime.UtcNow.Year);
```
amulware: Arbitrary year?

Toggl.Foundation.Tests/Generators/ViewModelGenerators.cs
```diff
@@ -0,0 +1,51 @@
...
+                       var year = yearGenerator.Sample(0, 1).First();
+
+                       var observable = dateTimes
+                            .Select(newDateWithGenerator(monthsGenerator, year))
+                            .Select(d => new TimeEntry { Start = d })
+                            .Apply(Observable.Return);
+                       
+                        source.TimeEntries.GetAll().Returns(observable);
+
+                       return viewModel;
```
amulware: Formatting

willsb: What about it?

amulware: Either this or the line above is indented incorrectly.

Toggl.Foundation.Tests/Generators/ViewModelGenerators.cs
willsb: Yup

PR #239, opened 2017-06-28
Add OnboardingNavigationController

willsb:
This adds a Navigation bar to the onboarding flow (and also to the LoginView, when it comes). This also makes the text on the title bar white.

Note that I disabled user interaction in the OnboardingView because I want us to be able to click through it. Why not hide it altogether? Because it it’s invisible then the status bar text won’t be white. Go figure.


amulware:
Can't you just set the status bar text colour separately?

willsb:
I'd rather not repeat that code, since this is a mechanism that the presenter gives to us for free (we'll need to create a `NavigationController` anyways)

PR #248, opened 2017-07-01
Select first view based on credentials

willsb:
Now we are only registering the LoginManager if absolutely needed.

I removed the `DatabaseFactory` because that closure was simply delaying failures and served us no purpose. If needed be, we can always add it back for supporting multiple databases; for the time being, KISS applies.


Toggl.Foundation.Tests/Login/LoginManagerTests.cs
```diff
@@ -41,26 +60,8 @@ public void ThrowsIfAnyOfTheArgumentsIsNull(bool useApiFactory, bool useDatabase
             }
         }
 
-        public class TheLoginMethod
+        public class TheLoginMethod : LoginManagerTest
```
amulware: We should also test that the correct errors are returned, if database or api misbehave (or login fails for that matter).

Toggl.Foundation/Login/LoginManager.cs
```diff
@@ -51,5 +51,22 @@ public IObservable<ITogglDataSource> Login(Email email, string password)
                         return new TogglDataSource(database, api);
                     });
         }
+
+        public ITogglDataSource IsLoggedIn()
```
amulware: From the name I expect this to be a bool property. Please rename.

Toggl.Foundation/Login/LoginManager.cs
```diff
@@ -51,5 +51,22 @@ public IObservable<ITogglDataSource> Login(Email email, string password)
...
+        {
+            return database.User
+                       .Single()
+                       .Select(dataSourceFromUser)
+                       .Catch(Observable.Return<ITogglDataSource>(null))
+                       .Wait();
```
amulware: Are we sure we want this to run synchronously? We could load/initialise more of the app, while this is running, no?

Toggl.Foundation/Login/LoginManager.cs
willsb: I'll rename once you point a better name

Toggl.Foundation.Tests/Login/LoginManagerTests.cs
```diff
@@ -12,26 +12,45 @@
...
         {
-            private static readonly IApiFactory ApiFactory = Substitute.For<IApiFactory>();
-            private static readonly DatabaseFactory DatabaseFactory = () => Substitute.For<ITogglDatabase>();
-
             [Theory]
             [InlineData(true, false)]
             [InlineData(false, true)]
             [InlineData(false, false)]
-            public void ThrowsIfAnyOfTheArgumentsIsNull(bool useApiFactory, bool useDatabaseFactory)
+            public void ThrowsIfAnyOfTheArgumentsIsNull(bool useApiFactory, bool useDatabase)
```
willsb: It's being tested everywhere, we have to construct this guy before using it 😄

Toggl.Foundation.Tests/Login/LoginManagerTests.cs
```diff
@@ -91,42 +92,67 @@ public void ThrowsIfYouPassInvalidParameters(string email, string password)
             [Fact]
             public async Task EmptiesTheDatabaseBeforeTryingToLogin()
```
willsb: True that.

Toggl.Foundation.Tests/Login/LoginManagerTests.cs
```diff
@@ -91,42 +92,67 @@ public void ThrowsIfYouPassInvalidParameters(string email, string password)
...
             {
-                await loginManager.Login(Email, Password);
+                await LoginManager.Login(Email, Password);
 
-                await database.User.Received().Create(Arg.Is<IDatabaseUser>(receivedUser => receivedUser.Id == user.Id));
+                await Database.User.Received().Create(Arg.Is<IDatabaseUser>(receivedUser => receivedUser.Id == User.Id));
             }
 
             [Fact]
             public async Task TheUserToBePersistedShouldNotHaveIsDirtySetToFalse()
```
willsb: 👍

Toggl.Foundation/Login/LoginManager.cs
willsb: This happens in a pretty sensitive part of the app. If this takes a bit too long and we don't register the app start, this might make the app crash. We should keep it sync and only change it if we feel it's slowing the app down. A single database access shouldn't be a problem, though (and if it is, we can start storing the credentials elsewhere).

Toggl.Foundation.MvvmCross/App.cs
```diff
@@ -1,11 +1,31 @@
...
 using Toggl.Foundation.MvvmCross.ViewModels;
 
 namespace Toggl.Foundation.MvvmCross
 {
     public class App : MvxApplication
     {
+        private readonly ILoginManager loginManager;
+
+        public App(ILoginManager loginManager)
+            => this.loginManager = loginManager;
```
willsb: Yet, there are a some of them throughout the application. Did it start bothering you only now? 🤔

amulware: I mostly started noticing now. But yeah, I've also been thinking about whether I like expression bodied void-return methods.

Toggl.Foundation.Tests/Login/LoginManagerTests.cs
amulware: I withdraw the comment.

Toggl.Foundation/Login/LoginManager.cs
amulware: We'll measure it (later).

willsb:
This will become a mess. I'll split in a PR for the LoginManager and another one for the AppStart part.

PR #250, opened 2017-07-03
Allow LoginManager to check if the user is logged in

willsb:


Toggl.Foundation/Login/LoginManager.cs
```diff
@@ -51,5 +51,22 @@ public IObservable<ITogglDataSource> Login(Email email, string password)
                         return new TogglDataSource(database, api);
                     });
         }
+
+        public ITogglDataSource IsLoggedIn()
```
amulware: Still don't like the name. Should be something that communicates what the method is actually doing. `TryLoggingInLocally`, but better. `TryConnectToDaraSource`, `LogInToDataSourceIfLoggedIn` dunno.

willsb: `InitializeDataSourceIfLoggedIn`?

PR #253, opened 2017-07-03
Add suggestion provider

### LINKED ISSUE #185: Basic suggestions
Implement a flexible suggestions framework allowing for multiple providers that can be extended in the future. (discuss details/requirements as needed)

The first and simplest provider should show the most used time entry in the last two weeks, matching by description, project and task.

arturs105:
Closes #185


Toggl.Foundation/Suggestions/MostUsedTimeEntryProvider.cs
```diff
@@ -0,0 +1,65 @@
...
+using Toggl.Multivac.Models;
+using Toggl.Ultrawave.Models;
+
+namespace Toggl.Foundation.Suggestions
+{
+    public class MostUsedTimeEntryProvider : ISuggestionProvider
+    {
+        private readonly ITogglDataSource dataSource;
+
+        public MostUsedTimeEntryProvider(ITogglDataSource dataSource)
```
willsb: Maybe this should only depend on DB

Toggl.Foundation/Suggestions/MostUsedTimeEntryProvider.cs
```diff
@@ -0,0 +1,65 @@
...
+        {
+            this.dataSource = dataSource;
+        }
+
+        public IObservable<List<ITimeEntry>> GetSuggestions()
+        {
+            return Observable.Create<List<ITimeEntry>>(async observer =>
+            {
+                var timeEntries = await dataSource.TimeEntries.GetAll();
```
willsb: If this were to depend on DB, you could simply filter directly

Toggl.Foundation/Suggestions/MostUsedTimeEntryProvider.cs
```diff
@@ -0,0 +1,65 @@
...
+                timeEntries = timeEntries.Where(te => isInLastTwoWeeks(te.Start));
+                var mostUsedTimeEntry = getMostUsedTimeEntry(timeEntries);
```
willsb: This could be simplified to:

```csharp
var suggestion = dataSource.TimeEntries
    .GetAll(inLastTwoWeeks)
    .GroupBy(te => new { te.Description, te.ProjectId, te.TaskId })
    .OrderByDescending(g => g.Count())
    .FirstOrDefault();

if (suggestion == null)
    return Observable.Empty();

var timeEntry = new TimeEntry {  Description = x.Description, ProjectId = x.ProjectId, TaskId = x.TaskId };
return Observable.Return(new List<ITimeEntry> { timeEntry });

```

arturs105: I thought, that ordering the list would be too much useless work, since we only need one largest value, but after reading about `OrderByDescending` being deferred, i agree with you.

Toggl.Foundation/Suggestions/MostUsedTimeEntryProvider.cs
```diff
@@ -0,0 +1,57 @@
...
+
+            this.database = database;
+        }
+
+        public IObservable<List<ITimeEntry>> GetSuggestions()
+        {
+            return Observable.Create<List<ITimeEntry>>(async observer =>
+            {
+                var list = new List<ITimeEntry>();
+                var mostUsedGroup = (await database.TimeEntries.GetAll(isInLastTwoWeeks))
```
willsb: I don't like that we are `awaiting` the database observable here. Why would we? The will make exception handling harder. We should instead combine observables to compose our own.

Toggl.Foundation/Suggestions/MostUsedTimeEntryProvider.cs
```diff
@@ -0,0 +1,57 @@
...
+                {
+                    list.Add(new TimeEntry
+                    {
+                        Description = mostUsedGroup.Key.Description,
+                        ProjectId = mostUsedGroup.Key.ProjectId,
+                        TaskId = mostUsedGroup.Key.TaskId
+                    });
+                }
+                    
+                observer.OnNext(list);
```
willsb: I don't like this idea of emitting an empty list. An observable can complete without emitting anything (and if there is no meaningful data, it should).

Toggl.Foundation/Suggestions/MostUsedTimeEntryProvider.cs
```diff
@@ -0,0 +1,57 @@
...
+                return Disposable.Empty;
```
willsb: This observable needs to complete

Toggl.Foundation/Suggestions/MostUsedTimeEntryProvider.cs
```diff
@@ -0,0 +1,57 @@
...
+            });
+        }
+
+        private bool isInLastTwoWeeks(ITimeEntry timeEntry)
+        {
+            var twoWeeks = TimeSpan.FromDays(14);
+            var delta = DateTime.Now - timeEntry.Start;
```
willsb: `UtcNow`?

Toggl.Foundation/Suggestions/MostUsedTimeEntryProvider.cs
arturs105: You mean use methods like `Select` on the database observable right?

willsb: Precisely

arturs105:
I got rid of awaiting the database observable, but the problem with empty list still remains. I could solve it with a class-level `bool` variable, which would be set to true in `selectSuggestions()` and then an empty observable could be returned from `GetSuggestions()`, but this is not the cleanest solution, is there a better way?

willsb:
Since in the end we'll combine multiple suggestion providers, I suggest (kek) returning an `IObservable<ITimeEntry>` instead. If any suggestion provider has the need to return multiple time entries at once, they can emit multiple items. This will simplify the logic for creating, combining and consuming the providers :)

arturs105:
Oh, yeah, i didn't know, that we'll combine providers - i thought that in future we might have a provider which returns multiple suggestions, but yeah, this makes sense.

Toggl.Foundation.Tests/Suggestions/MostUsedTimeEntryProviderTests.cs
```diff
@@ -0,0 +1,27 @@
...
+    {
+        public class TheGetSuggestionMethod
+        {
+            [Fact]
+            public void ThrowsIfThereAreNoTimeEntries()
+            {
+                var database = Substitute.For<ITogglDatabase>();
+                var provider = new MostUsedTimeEntryProvider(database);
+
+                Action getSuggestionAction = () => provider.GetSuggestion().Wait();
```
willsb: `provider.GetSuggestion().IsEmpty().Wait().Should().BeTrue()`

Toggl.Foundation.Tests/Suggestions/MostUsedTimeEntryProviderTests.cs
```diff
@@ -0,0 +1,27 @@
...
+
+                getSuggestionAction.ShouldThrow<InvalidOperationException>();
+            }
+        }
+    }
+}
```
willsb: \n

Toggl.Foundation/Suggestions/MostUsedTimeEntryProvider.cs
```diff
@@ -0,0 +1,57 @@
...
+                return null;
+
+            return new TimeEntry
+            {
+                Description = mostUsedGroup.Key.Description,
+                ProjectId = mostUsedGroup.Key.ProjectId,
+                TaskId = mostUsedGroup.Key.TaskId
+            };
+        }
+
```
willsb: Extra space

Toggl.Foundation/Suggestions/MostUsedTimeEntryProvider.cs
```diff
@@ -0,0 +1,57 @@
...
+
+        private bool isNotNull(ITimeEntry timeEntry)
+            => timeEntry != null;
+    }
+}
```
willsb: \n

Toggl.Foundation.Tests/Suggestions/MostUsedTimeEntryProviderTests.cs
arturs105: 👍

arturs105:
More tests are already on the way

Toggl.Foundation/Suggestions/MostUsedTimeEntryProvider.cs
arturs105: I don't see any problem with newlines here

Toggl.Foundation/Suggestions/MostUsedTimeEntryProvider.cs
arturs105: Don't see any

Toggl.Foundation.Tests/Suggestions/MostUsedTimeEntryProviderTests.cs
```diff
@@ -0,0 +1,79 @@
...
+                var suggestion = await provider.GetSuggestion();
+
+                suggestion.Description.Should().Be("te0");
+                suggestion.TaskId.Should().Be(0);
+                suggestion.ProjectId.Should().Be(0);
+            }
+
+            private IObservable<IEnumerable<IDatabaseTimeEntry>> createObservable(ITimeEntry[] timeEntries)
+            {
+                var databaseTimeEntries = timeEntries.Select(te => Foundation.Models.TimeEntry.Clean(te));
```
willsb: No need for the lamda, `Select(Foundation.Models.TimeEntry.Clean)` works fine

PR #254, opened 2017-07-03
Add UI Tests for Login

willsb:
Squash only after watching all the tests running :robot:


Toggl.Daneel.Tests.UI/LoginTests.cs
```diff
@@ -0,0 +1,55 @@
+﻿using NUnit.Framework;
+using Xamarin.UITest;
+using Xamarin.UITest.iOS;
+using static Toggl.Daneel.Tests.UI.Extensions.LoginExtensions;
```
arturs105: Why not simply `using Toggl.Daneel.Tests.UI.Extensions`?

willsb: Because this only imports the portion of the extensions we need; These have a tendency of growing quite a lot

PR #256, opened 2017-07-04
Login handling

willsb:
The user now no longer has to login whenever he opens the app, YAY

- I'm using a custom `AppStart` class because it makes everything simpler for both handling dependencies and testing.
- I'm intercepting the result of `InitializeNavigationService` because this way I don't have to use the ServiceLocator.

Since this is pretty straight forward, squash using the 🔑  emoji as soon as approved


Toggl.Foundation.MvvmCross/App.cs
```diff
@@ -1,11 +1,55 @@
...
 namespace Toggl.Foundation.MvvmCross
 {
     public class App : MvxApplication
     {
+        private readonly AppStart appStart;
+
+        public App(ILoginManager loginManager, IMvxNavigationService navigationService)
+        {
+            Ensure.Argument.IsNotNull(loginManager, nameof(loginManager));
+            Ensure.Argument.IsNotNull(navigationService, nameof(navigationService));
```
amulware: Should these be tested?

Toggl.Foundation.MvvmCross/App.cs
```diff
@@ -1,11 +1,55 @@
...
+
+            appStart = new AppStart(loginManager, navigationService);
+        }
+
         public override void Initialize()
-            => RegisterAppStart<OnboardingViewModel>();
+            => RegisterAppStart(appStart);
```
amulware: I assume this can't easily be tested?

willsb: They can't, no. They will be covered by UI tests, though :)

Toggl.Foundation.MvvmCross/App.cs
willsb: Well, this class is meant to be dumb and the check done in AppStart is tested. What do you say? I honestly have no opinion on this.

PR #260, opened 2017-07-05
Update NuGet packages

willsb:
This updates:

- Realm
- PropertyChanged.Fody
- MvvmCross
- Minor dependencies

The newest version of `PropertyChanged.Fody` does not require any attributes (in fact, it does not compile if you use the now deprecated `ImplementPropertyChangedAttribute`, so removing it is needed for the project to build)

Squash whenever


amulware:
If it does not have any attributes.. does that mean that all properties on all types now cause the PropertyChanged code to run?

willsb:
All auto implemented props in types that implement INotifyPropertyChanged aka all types that need this behavior

PR #261, opened 2017-07-05
Reduce asset size

willsb:
I ran all the image files through [ImageOptim](https://imageoptim.com/mac). The outcome is ~43% size reduction 💪 

image

I also removed the *Pixel images, that were not being used anywhere.

Squash whenever


amulware:
Are none of the removed file headers needed for iOS to display the images correctly?

Also, did you apply any form of lossy compression, or only entirely lossless one?

willsb:
Lossless. It removed only junky EXIF.

Issue #78, opened 2017-04-26
Begin implementation of timer page


amulware:
Let's get up a scaffolding to support the three parts of the interface: timeline, suggestions, and log.

Each of the three will get its own issue.

<img width="374" alt="screen shot 2017-04-26 at 14 44 13" src="https://cloud.githubusercontent.com/assets/3987892/25435005/e1328a92-2a8e-11e7-948b-a56d0d6013e1.png">

amulware:
There is some code here that could possible be reused: #121 (up to whoever works on this)

amulware:
Such out of date. Much old. Wow.

PR #263, opened 2017-07-06
Add timer suggestion cell

arturs105:
Used to display suggestions. Can be seen in #78.


arturs105:
Just noticed that the last cell in the image doesn't have a project label. Should that be a different cell or should this cell be rearranged if the project is not set?

amulware:
I think it should be the same cell. I much rather have the branching logic right there in the UI, than having to create different views based on the model.

willsb:
I strongly disagree, Paul. No reason for increasing the complexity of a View when we can simply add multiple files.

amulware:
You rather duplicate virtually the entire view for a slight difference in layout? And if the design changes change both?

arturs105:
Bottom row (`ProjectLabel` and `TaskLabel`) can be removed by setting `BottomRowView.Hidden = true`, `DescriptionLabel` will automatically recenter.

PR #268, opened 2017-07-10
Add suggestions view model

arturs105:


Toggl.Foundation.MvvmCross/ViewModels/SuggestionsViewModel.cs
```diff
@@ -0,0 +1,32 @@
...
+namespace Toggl.Foundation.MvvmCross.ViewModels
+{
+    public sealed class SuggestionsViewModel : BaseViewModel
+    {
+        private readonly ISuggestionProvider[] suggestionProviders;
+
+        public ObservableCollection<ITimeEntry> Suggestions { get; }
+            = new ObservableCollection<ITimeEntry>();
+
+        public SuggestionsViewModel(params ISuggestionProvider[] suggestionProviders)
```
willsb: Include an MvvmCross specific interface that's easy to inject; I don't know how the IoC system is going to deal with params + array parameters. In theory it shouldn't be a problem, but I don't know if registering and array is a good idea.

Toggl.Foundation.MvvmCross/ViewModels/SuggestionsViewModel.cs
```diff
@@ -0,0 +1,32 @@
...
+
+            this.suggestionProviders = suggestionProviders;
+        }
+
+        public async override Task Initialize()
+        {
+            await base.Initialize();
+
+            foreach (var provider in suggestionProviders)
+                Suggestions.Add(await provider.GetSuggestion());
```
willsb: This sort of defeats the purposes of observables. Awaiting only returns the last one. Some providers will want to return multiple suggestions (and some won't return anything). Also, some might even error. This does not cover any of those cases.

Toggl.Foundation.MvvmCross/ViewModels/SuggestionsViewModel.cs
arturs105: So something like this?

```
interface ISuggestionProviderContainer
{
    List<ISuggestionProvider> Providers { get; }
}
```

willsb: Precisely

Toggl.Foundation.Tests/MvvmCross/ViewModels/SuggestionsViewModelTests.cs
```diff
@@ -0,0 +1,78 @@
...
+
+            [Fact]
+            public async Task WorksIfProviderHasMultipleSuggestions()
+            {
+                var provider = Substitute.For<ISuggestionProvider>();
+                var timeEntries = new[]
+                {
+                    new TimeEntry{ Description = "te1" },
+                    new TimeEntry{ Description = "te2" },
+                    new TimeEntry{ Description = "te3" }
```
amulware: I'd like spaces before the `{`.

Toggl.Foundation.Tests/MvvmCross/ViewModels/SuggestionsViewModelTests.cs
```diff
@@ -0,0 +1,78 @@
...
+                };
+                var observable = Observable.Create((IObserver<ITimeEntry> arg) =>
```
amulware: `arg` could be named better.

Toggl.Foundation.MvvmCross/ViewModels/SuggestionsViewModel.cs
```diff
@@ -0,0 +1,32 @@
...
+using Toggl.Multivac.Models;
+
+namespace Toggl.Foundation.MvvmCross.ViewModels
+{
+    public sealed class SuggestionsViewModel : BaseViewModel
+    {
+        private readonly ISuggestionProviderContainer suggestionProviders;
+
+        public ObservableCollection<ITimeEntry> Suggestions { get; }
+            = new ObservableCollection<ITimeEntry>();
```
amulware: What is the reasoning behind using this interface, if we don't actually need all the properties in it? Would it not make sense to have an interface/model with exactly the properties needed?

Toggl.Foundation.Tests/MvvmCross/ViewModels/SuggestionsViewModelTests.cs
```diff
@@ -0,0 +1,78 @@
...
+            {
+                Action tryingToConstructWithEmptyParameters =
+                    () => new SuggestionsViewModel(null);
+
+                tryingToConstructWithEmptyParameters
+                    .ShouldThrow<ArgumentNullException>();
+            }
+        }
+
+        public class TheSuggestionsProperty
```
amulware: I think this should also test that it throws if there are no/null providers, and the case where none of the given providers return any suggestions.

Toggl.Foundation.MvvmCross/ViewModels/SuggestionsViewModel.cs
arturs105: Do you mean `ITimeEntry`? I didn't think it was necessary to create a new interface for this. We could also use tuples.

Toggl.Foundation.Tests/MvvmCross/ViewModels/SuggestionsViewModelTests.cs
arturs105: The view model can't even be created without providers. `SuggestionProviderContainer` throws an exception if it gets created with no/null providers. That behaviour is already tested in `SugestionProviderContainerTests`.

Your second point is valid though.

Toggl.Foundation.MvvmCross/ViewModels/SuggestionsViewModel.cs
amulware: This would be a big tuple, so better to have a separate model, if we would change it. What do you think @willsb?

amulware:
Apart from the question about whether to use a separate model or not, I'm fine with this.

If we added a separate model, the same could be used for continuation of other time entries. Just a thought.

Toggl.Foundation.MvvmCross/ViewModels/SuggestionsViewModel.cs
willsb: We will have a PR to change the ITimeEntries in the TImeEntriesLogViewModel to a new model. I think the same model could be used here, no? But that's a different PR

PR #271, opened 2017-07-12
Add header title converter

willsb:
Converts a date into the title of the header (the one that appears in the TimeEntriesLogViewController)

Squash ASAP


arturs105:
What about tests?

PR #273, opened 2017-07-13
Change tappable areas of onboarding

willsb:
They are much easier to hit now. Thank me later.

Squash whenever


Toggl.Daneel/ViewControllers/OnboardingViewController.xib
```diff
@@ -248,23 +261,5 @@ START TRACKING</string>
...
-        <image name="icDownS" width="8" height="5"/>
-        <image name="icProjects" width="16" height="12"/>
-        <image name="icRefresh" width="20" height="14"/>
-        <image name="icReports" width="16" height="18"/>
-        <image name="icSettings" width="16" height="16"/>
-        <image name="icSyncerror" width="16" height="16"/>
-        <image name="icTags" width="14" height="13"/>
-        <image name="icUpS" width="8" height="5"/>
-        <image name="playIcon" width="50" height="50"/>
-        <image name="stopIcon" width="34" height="34"/>
```
amulware: Where did these go?

arturs105:
I see how this increases the tappable areas and it (sorta) works. It works perfectly if you tap below the buttons, but it does not quite work above the buttons. Seems like the `UIScrollView` handles the touches and the buttons don't react to them. Is there a way to pass the touches through the scroll view?

Toggl.Daneel/ViewControllers/OnboardingViewController.xib
willsb: Interface builder changes files when you open them. Just ignore this

willsb:
On it, @arturs105

arturs105:
The issue still persists. Decreasing the width of the `UIScrollView` and setting `ClipsToBounds = false` might work, but then swiping wouldn't work at the edges.

willsb:
I don't think the issue is that big. I'm using it on the smallest possible screen and it works just fine (and much better than previously). If you're clicking on the scroll view then you're obviously not clicking the next button. But I might be biased, so I summon @amulware

arturs105:
I guess you're right - the buttons aren't that hard to press.

PR #275, opened 2017-07-13
Add SuggestionsEmptyViewCell

willsb:
Self-explanatory name. The cell is initialized with random values when it gets created, as discussed on slack.

Squash when you see a randomly selected blue project.


Toggl.Daneel/Views/SuggestionsEmptyViewCell.cs
```diff
@@ -0,0 +1,52 @@
+﻿using System;
+using Foundation;
+using UIKit;
+
+namespace Toggl.Daneel.Views
+{
+    public partial class SuggestionsEmptyViewCell : UITableViewCell
+    {
+        private static readonly Random Random = new Random();
```
amulware: Random is not thread safe. Will this always be used on only one thread?

willsb: AwakeOnNib is sure to run on the MainThread. This is safe :+1:

willsb:
Is this approved? :smile:

PR #281, opened 2017-07-17
Fix incorrect behavior when validating email

willsb:
This is very specific. Since we were depending on the `Email` property and actually using the `email` field, the property was being raised at the wrong time, making it so that, by deleting characters after inputing a valid email, an invalid email could be used during the Login step. Not really a bug, since this is the behavior we specified.

Squash whenever.


amulware:
It looks fine, but I don't understand why it didn't work before? Deleting characters still changed the email field and should raise and reevaluate the visibility property, no?

willsb:
It was reevaluating it _before_ updating the `email` field; hence the reevaluation was happening before the prop was ready and the check used the version of `Email` that existed before the user deleted one character.

amulware:
Oh, I see. That is indeed very subtle and not at all obvious.

PR #282, opened 2017-07-18
Update projects dependencies

kornelijepetak:
I have consolidated versions of various assemblies, which were clashing during runtime (at least on Windows):

* `System.Net.Http` to `4.3.2`
* `System.Value.Tuple` to `4.3.1`
* `System.Diagnostics.DiagnosticSource` to `4.4.0`
* `System.Linq.Queryable` to `4.3.0`

#### Potential issue:
The tests weren't passing on MSBuild, because different projects have had different versions of same assemblies. They still do and as far as I know, the only way to solve it is to add `app.config` file for each of those projects (that have conflicting package versions) and to do an assembly binding redirect to the newer version. 

```xml
<assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
    <dependentAssembly>
        <assemblyIdentity name="System.Net.Http" publicKeyToken="b03f5f7f11d50a3a" culture="neutral" />
        <bindingRedirect oldVersion="0.0.0.0-4.1.1.1" newVersion="4.1.1.1" />
      </dependentAssembly>
</assemblyBinding>
```

Visual Studio does that by default when such packages are installed. However, I kept removing those files because they weren't in the repository before. Why does XBuild handle things differently and how it does assembly binding, that's something to maybe look into. If I leave those `app.config`, they get copied to the output folder as `<dll-name>.config` and then all the tests pass.

It would be best if somebody tried this on Mac before merging.

Also, should we keep app.config files with the project (for Windows?)


Toggl.Foundation.MvvmCross/packages.config
```diff
@@ -8,8 +8,11 @@
   <package id="MvvmCross.Plugin.Color" version="5.0.5" targetFramework="portable45-net45+win8+wpa81" />
   <package id="MvvmCross.Plugin.Visibility" version="5.0.5" targetFramework="portable45-net45+win8+wpa81" />
   <package id="PropertyChanged.Fody" version="2.1.3" targetFramework="portable45-net45+win8+wpa81" developmentDependency="true" />
+  <package id="System.Collections" version="4.3.0" targetFramework="portable45-net45+win8+wpa81" />
```
amulware: There are a few more here than before. How come?

kornelijepetak: Because MvvmCross depends on NETStandard (1.6.1).
`System.Value.Tuple` dependencies on NETStandard 1.* are:
* `Sytem.Collections >= 4.3.0`
* `Sytem.Resources.ResourceManager >= 4.3.0`
* `Sytem.Runtime >= 4.3.0`

amulware:
Please wait for the review from @willsb as well (since you requested it).

kornelijepetak:
Can you comment on the app.config issue? They are required by MSBuild (.NET) and probably ignored by XBuild (mono). Are we leaving things as are?

Toggl.Foundation.MvvmCross/packages.config
willsb: [I'll just leave this here](https://fsprojects.github.io/Paket/paket-why.html)

kornelijepetak:
But I haven't commited `app.config` files. Why did it change anything for you?

PR #283, opened 2017-07-18
Fix invalid test condition

kornelijepetak:
I believe logically the test was incorrect. 

It was possible for the test to pass even though there is no a single client with all three tested properties.
The returned list should contain a client in its entirety.


amulware:
Any idea why the diff shows almost the entire file as changed? Line endings?

kornelijepetak:
Probably. But I don't get it. Whichever version (of 4) of the file I open, I see LF as a line ending.

amulware:
When squashing, please make sure there is an emoji in that commit message ;)

kornelijepetak:
If you look at where the changes begin, it seems that it's more probable that the issue lies with whitespace than with line endings (first few lines are ok, diffs start where indentation starts). But I've checked even that and all versions have 4 spaces indentation. Strange.

Issue #105, opened 2017-05-08
Create Project API


amulware:
For endpoints see #29.

For now it should be sufficient to add GETting of all the user's projects, and POSTing of new ones. This should be exposed in a nicer way as `Create()` or similar. (Editing them using PUT is not needed.)

Please `curl` the endpoints to see the exact format of the data it returns. Otherwise feel free to orient yourself on the already existing model, endpoints, etc.

Models and clients should have unit tests.

Please submit multiple small PRs for easier reviewing.

- [ ] endpoints
- [ ] models
- [ ] api client
- [ ] integration tests

amulware:
This was started here: #63

Feel free to ignore or reuse as appropriate.

kornelijepetak:
Implemented in f4d471ae74ec88d237999d4841d1b150a1859bed.

PR #284, opened 2017-07-18
Make premium Project features nullable and ignore in JSON.

kornelijepetak:
Make premium features nullable and ignore them when serializing in JSON.

Completes model part of #105


willsb:
I don't think we can simply ignore then, can we? We should ignore based on whether the feature is available or not, right?

ping @arturs105 and @amulware

amulware:
@willsb That's correct, and there is an issue for that. It requires workspace features, and some thinking, so no way we can do this right now. For the time being, I'm quite fine with ignoring the properties. I see no harm in that.

amulware:
@willsb Does this work for you?

willsb:
Since the issue is there to remind us, yeah, I guess it's fine. But I'm still a bit 🤔  over this

amulware:
I understand. We'll have to think about this more, but for now, this makes things not break horribly.

PR #286, opened 2017-07-18
Api projects integration tests

kornelijepetak:
Create integration tests for Projects API.


kornelijepetak:
I don't get it why ProjectsApi.cs is changed again completely. Both versions show as spaces/LF to me. The previous and the new one. Anyone else?

kornelijepetak:
Nope, unfortunately it seems that the previous one was CRLF. Now it should be OK (with LF).

Toggl.Ultrawave.Tests.Integration/ProjectsApiTests.cs
```diff
@@ -0,0 +1,138 @@
...
+                    persistedProject.Name.Should().Be(project.Name);
+                    persistedProject.ClientId.Should().Be(project.ClientId);
+                    persistedProject.IsPrivate.Should().Be(project.IsPrivate);
+                    persistedProject.Color.Should().Be(project.Color);
+                }
+            }
+
+            private static async Task<Project> createNewProject(ITogglApi togglClient, int workspaceID, bool isActive = true, bool createClient = false)
+            {
+                Client client = null;
```
willsb: This could be inline with `?:`

Toggl.Ultrawave.Tests.Integration/ProjectsApiTests.cs
```diff
@@ -0,0 +1,138 @@
...
+                project = await createNewProject(togglClient, user.DefaultWorkspaceId, isActive: false);
+                await togglClient.Projects.Create(project);
+
+                var activeProjects = await CallEndpointWith(togglClient);
+
+                noProjects.Should().BeNull();
+                activeProjects.Should().BeNull();
+            }
+
+            public class TheCreateMethod : AuthenticatedPostEndpointBaseTests<Project>
```
willsb: Why is this nested inside the `GetAllMethod`?

Toggl.Ultrawave.Tests.Integration/ProjectsApiTests.cs
```diff
@@ -0,0 +1,138 @@
...
+                var projects = await CallEndpointWith(togglClient);
+
+                projects.Should().HaveCount(2);
+
+                projects.Should().Contain(project => isTheSameAs(projectAPosted, project));
+                projects.Should().Contain(project => isTheSameAs(projectBPosted, project));
+            }
+
+            [Fact, LogTestInfo]
+            public async System.Threading.Tasks.Task ReturnsOnlyActiveProjects()
```
willsb: Either alias this or our own Task. Full path is gross

Toggl.Ultrawave.Tests.Integration/ProjectsApiTests.cs
```diff
@@ -0,0 +1,138 @@
...
+                var projectAPosted = await togglClient.Projects.Create(projectA);
+
+                var projectB = await createNewProject(togglClient, user.DefaultWorkspaceId);
+                var projectBPosted = await togglClient.Projects.Create(projectB);
+
+                var projects = await CallEndpointWith(togglClient);
+
+                projects.Should().HaveCount(2);
+
+                projects.Should().Contain(project => isTheSameAs(projectAPosted, project));
```
willsb: This could be chained with `.And`

willsb: Shouldn't this check that all returned projects are the ones we have locally? iow, that there are no backend created projects among this?

Toggl.Ultrawave.Tests.Integration/ProjectsApiTests.cs
```diff
@@ -0,0 +1,138 @@
...
+
+                [Fact, LogTestInfo]
+                public async System.Threading.Tasks.Task CreatesNewProject()
+                {
+                    var (togglClient, user) = await SetupTestUser();
+
+                    var project = await createNewProject(togglClient, user.DefaultWorkspaceId);
+                    var persistedProject = await CallEndpointWith(togglClient, project);
+
+                    persistedProject.Name.Should().Be(project.Name);
```
willsb: Why not `ShouldBeEquivalentTo`?

Toggl.Ultrawave.Tests.Integration/ProjectsApiTests.cs
amulware: Is it not testing that?

kornelijepetak: It is testing it, logically. If it has 2 and it contains one and the other. Then that's that.

willsb: Oh yeah, with count. Makes sense.

Toggl.Ultrawave.Tests.Integration/ProjectsApiTests.cs
kornelijepetak: By error. Gonna fix it.

Toggl.Ultrawave.Tests.Integration/ProjectsApiTests.cs
kornelijepetak: The problem is not in the Task, but in the Project class (because of which I had to introduce its namespace, which also contains our own Task). What would be the appropriate alias for Project, ProjectModel?

Toggl.Ultrawave.Tests.Integration/ProjectsApiTests.cs
kornelijepetak: Because `persistedProject` and `project` will differ in the `Id`, making them _not equivalent_, even though they are (`Id` is not set on in-memory object before POSTing, so it should be ignored).

Toggl.Ultrawave.Tests.Integration/ProjectsApiTests.cs
kornelijepetak: Since I am awaiting the results, this could become less readable than it is now. If you still want to replace it with `?:`, that's ok.

Toggl.Ultrawave.Tests.Integration/ProjectsApiTests.cs
willsb: Yep, that's a good alias

Toggl.Ultrawave.Tests.Integration/ProjectsApiTests.cs
willsb: I see. Leave it as is

Issue #215, opened 2017-06-16
Create workspace feature API


amulware:
Related to #214

For the endpoints see #29

amulware:
I'd like an integration test that checks that the API returns **exactly** the list of features known to the app, and a second test that checks that all the `name` fields in the feature json are exactly the same as the feature enumeration value names (except in :snake: case).

The `name` field is not actually important as such (the id is what we will use for all logic, since this is guaranteed to be fixed) but it would still be good to get a failing test when the names do change for some reason (which they should only in very special circumstances).

amulware:
(The reason to check that the list is exactly the same is so that we are automatically notified by our tests if a feature is removed/added.)

Issue #214, opened 2017-06-16
Create workspace feature models


amulware:
This might not be as straight forward as other models to convert between API and database models, because of the strange (nested) json format. A good common representation has to be found.

For the endpoints, see: #29

amulware:
Please feel free to post a proposal here for discussion before implementing it, or if you are sure about it, just go ahead.

kornelijepetak:
I can see some possible solutions for this.

1) Restructure the model to enable direct serialization and use that in API throughout the rest of the app.

2) Use different, custom `IJsonSerializer` instance for `WorkspaceFeaturesApi`, and manually adapt the structure from the response to the model.  

3) Add additional temporary (DTO) classes in the serialization part of the process that would mimic the HTTP response and then adapt them in-memory to what we actually want to use in the app, thus hiding those DTO classes from the rest of the app.

Out of those three options, 1st seems bad because the response format is counterintuitive for almost all other operations. Between 2nd and the 3rd, I guess it's better to use a few new classes that are used only in a single part of the process and not visible from then on, than to implement custom parsing logic with JSON.NET. 

3rd option would look something like this (don't mind the class names):

```c#
internal class WorkspaceFeatureCollectionDefinitionDTO
{
    public int WorkspaceId { get; set; }
    public List<WorkspaceFeatureDefinitionDTO> Features { get; set; }
}

internal class WorkspaceFeatureDefinitionDTO
{
    public WorkspaceFeatureId FeatureID { get; set; }
    public string Name { get; set; }
    public bool Enabled { get; set; }
}

internal static class WorkspaceFeatureCollectionExtensions
{
    internal static List<WorkspaceFeature> ToWorkspaceFeatures(
        this List<WorkspaceFeatureCollectionDefinitionDTO> workspaceFeatureCollection)
        => workspaceFeatureCollection
        .SelectMany(wf => wf.Features.Select(f => f.ToWorkspaceFeature(wf.WorkspaceId)))
        .ToList();

    internal static WorkspaceFeature ToWorkspaceFeature(this WorkspaceFeatureDefinitionDTO feature, int workspaceId)
        => new WorkspaceFeature
        {
            WorkspaceId = workspaceId,
            Name = feature.Name,
            Enabled = feature.Enabled,
            FeatureId = feature.FeatureID
        };
}
```

Json response would be naturally deserialized in API into DTO classes and then converted into real entities (i.e. `WorkspaceFeature`) with those two ext. methods.

Any comments?

[...]

PR #295, opened 2017-07-20
Workspace features

kornelijepetak:
Changes related to #214, #215 as a work-in-progress.

Certain things are still to be noted:

* It's not determined how to persist `WorkspaceFeature` in the database (it has an undefined primary key `Id`).
* Sometimes totally unrelated test fail randomly if run in parallel (rate limiting?)
* This implementation goes along with a 3rd version discussed on #214, which means that if decision is to be made the other way, this PR will need to change accordingly.

TL;DR
I put this here not to lose the work I've done (during discussions) if it can be helpful later on.


simonrozsival:
This PR, which closes #214 and also closes #215, is now ready for reviews.

Deserialization of the workspace features is implemented, as well as the API client for getting the list of workspaces features. Both integration and unit tests are included.

One notable thing is the interface `Toggl.Multivac.Models.IWorkspaceFeature`, which does not implement `IBaseModel` as the features do not have `Id`s. I also moved it from the `Models` directory into `SpecificModels` directory (better name suggestions welcome), so it is not used by generators (which look for the contents of the directory `Toggl.Multivac/Models`).

simonrozsival:
I moved the `IWorkspaceFeature` to the common models' directory as @willsb requested.

Toggl.Ultrawave/ApiClients/WorkspaceFeaturesApi.cs
```diff
@@ -0,0 +1,91 @@
...
+
+        internal class WorkspaceFeatureDTO
+        {
+            public WorkspaceFeatureId FeatureId { get; set; }
+            public string Name { get; set; }
+            public bool Enabled { get; set; }
+        }
+    }
+
+    internal static class WorkspaceFeatureCollectionExtensions
```
amulware: Why extensions instead of methods on the objects above?

Toggl.Ultrawave/ApiClients/WorkspaceFeaturesApi.cs
```diff
@@ -0,0 +1,91 @@
...
+        public IObservable<List<IWorkspaceFeature>> GetEnabledFeatures()
+        {
+            return CreateObservable<List<WorkspaceFeatureCollectionDTO>>(endPoints.Get, AuthHeader)
+                .Select(list =>
+                    list.ToWorkspaceFeatures()
+                        .Where(wf => wf.Enabled)
+                        .ToList());
+        }
+
+        public IObservable<List<IWorkspaceFeature>> GetEnabledFeaturesForWorkspace(long workspaceId)
```
amulware: Why does this get all and then filter? There is an endpoint for getting the features of a specific workspace.

Also, what is this needed for? I don't see any use cases for our app.

Toggl.Ultrawave/ApiClients/WorkspaceFeaturesApi.cs
```diff
@@ -0,0 +1,91 @@
...
+
+        public IObservable<List<IWorkspaceFeature>> GetAll()
+        {
+            return CreateObservable<List<WorkspaceFeatureCollectionDTO>>(endPoints.Get, AuthHeader)
+                .Select(list => list
+                    .ToWorkspaceFeatures()
+                    .ToList());
+        }
+
+        public IObservable<List<IWorkspaceFeature>> GetEnabledFeatures()
```
amulware: I don't understand why this exists. If the consumer wants to filter, they can do that themselves.

However, I see no use case for the filtered list either.

Toggl.Ultrawave/ApiClients/WorkspaceFeaturesApi.cs
```diff
@@ -0,0 +1,91 @@
...
+        {
+            public WorkspaceFeatureId FeatureId { get; set; }
+            public string Name { get; set; }
+            public bool Enabled { get; set; }
+        }
+    }
+
+    internal static class WorkspaceFeatureCollectionExtensions
+    {
+        internal static IEnumerable<IWorkspaceFeature> ToWorkspaceFeatures(this IEnumerable<WorkspaceFeatureCollectionDTO> collection)
```
amulware: This returns a list, why not make the return type specific and avoid the wasteful `ToList()` above?

Toggl.Ultrawave.Tests.Integration/WorkspaceFeaturesEnumTests.cs
```diff
@@ -0,0 +1,64 @@
...
+{
+    public class WorkspaceFeaturesIdEnumTests
+    {
+        public class BackendEnumValuesCheckTest : AuthenticatedEndpointBaseTests<List<(WorkspaceFeatureId FeatureId, string Name)>>
+        {
+            protected override IObservable<List<(WorkspaceFeatureId FeatureId, string Name)>> CallEndpointWith(ITogglApi togglApi)
+                => (togglApi.WorkspaceFeatures as WorkspaceFeaturesApi).GetAllRaw();
+
+            [Fact, LogTestInfo]
+            public async Task TestsWhetherEnumValuesMatchBackendResponse()
```
amulware: This test's name has different semantics than all our other tests.

Toggl.Ultrawave/ApiClients/WorkspaceFeaturesApi.cs
simonrozsival: This is original @kornelijepetak's implementation. I wasn't sure if you discussed it or not and I just couldn't find it. I will 🔥 it.

Toggl.Ultrawave/ApiClients/WorkspaceFeaturesApi.cs
simonrozsival: Same as above -> 🔥 .

Toggl.Ultrawave/ApiClients/WorkspaceFeaturesApi.cs
simonrozsival: The point is to have DTO objects just store the data without any business logic inside. The extension method is a bit of a hack, but it makes clear what the purpose of those objects is. I won't mind placing them inside the classes.

amulware: I believe that would allow the classes to be private, right? That seems like a nicer solution to me, yeah.

simonrozsival: It must be internal, because I need to access the DTO in tests: https://github.com/toggl/mobileapp/pull/295/files#diff-667bbee88f9bebceb2e74338711eeb93R26

I could expose the deserialization to the tests in some different way, but that means extra code just because of the tests, which I don't think is a better solution. Or I could remove the serialization test and say that deserialization is tested in fact in the integration tests... but that also doesn't feel right because this is the case of more complicated deserialization and the test in fact shows exactly why the DTO is needed and how is it used.

amulware: Hmm. I see.

simonrozsival: Does this mean you approve the extension method or do still request changes here?

amulware: Oh, please move the methods. Otherwise, ok for now. I'm not super happy with it, but also am not sure how I would prefer it.

simonrozsival:
I have rewritten the PR but as it changed significantly, I decided to close this one and open #386 instead.

PR #296, opened 2017-07-20
Add custom activity indicator

arturs105:
Adds a rotating activity indicator that uses custom images.


Toggl.Daneel/Views/ActivityIndicator.cs
```diff
@@ -0,0 +1,73 @@
+﻿using System;
+using CoreAnimation;
+using CoreGraphics;
+using Foundation;
+using UIKit;
+
+namespace Toggl.Daneel.Views
+{
+    [Register(nameof(ActivityIndicator))]
+    public class ActivityIndicator : UIView
```
willsb: Let's make naming consistent and call this `ActivityIndicatorView`

Toggl.Daneel/Views/ActivityIndicator.cs
```diff
@@ -0,0 +1,73 @@
...
+    {
+        private UIImageView imageView;
+        private CABasicAnimation animation;
+        private string animationKey = "rotationAnimation";
```
willsb: Make this a constant

Toggl.Daneel/Views/ActivityIndicator.cs
```diff
@@ -0,0 +1,73 @@
...
+
+        private string imageResource;
+        public string ImageResource
```
willsb: This prop is really confusing. Why are you setting the backing field on get?

Toggl.Daneel/Views/ActivityIndicator.cs
```diff
@@ -0,0 +1,73 @@
...
+        }
+
+        public ActivityIndicator(IntPtr handle)
+            : base(handle)
+        {
+        }
+
+        private void init()
+        {
+            imageView = new UIImageView(Frame)
```
willsb: Couldn't you instantiate this on construction and add it here, making the `UIImageView` readonly?

Toggl.Daneel/Views/ActivityIndicator.cs
```diff
@@ -0,0 +1,73 @@
...
+            {
+                Image = UIImage.FromBundle(ImageResource),
+                ContentMode = UIViewContentMode.Center
+            };
+            AddSubview(imageView);
+            
+            animation = CABasicAnimation.FromKeyPath("transform.rotation.z");
```
willsb: This could also be created on construction to make the animation readonly

Toggl.Daneel/Views/ActivityIndicator.cs
```diff
@@ -0,0 +1,73 @@
...
+            animation.Duration = 0.8;
+            animation.Cumulative = false;
+            animation.RepeatCount = float.PositiveInfinity;
+            animation.RemovedOnCompletion = false;
+            animation.By = NSObject.FromObject(Math.PI * 2);
+        }
+
+        public void Start()
+        {
+            imageView.Layer.AddAnimation(animation, animationKey);
```
willsb: Why would this ever stop? It should be spinning constantly and, if it shouldn't be spinning, it should be simply static

Toggl.Daneel/Views/ActivityIndicator.cs
```diff
@@ -0,0 +1,73 @@
...
+        public void Stop()
+        {
+            var transform = imageView.Layer.PresentationLayer.Transform;
+            imageView.Layer.RemoveAnimation(animationKey);
+            imageView.Layer.Transform = transform;
+        }
+
+        public override void AwakeFromNib()
+        {
+            init();
```
willsb: There's no NIB associated to it, why are you using this method?

Toggl.Daneel/Views/ActivityIndicator.cs
willsb: Also, wouldn't it be simpler to subview `UIImageView` instead?

arturs105: It probably would, but a while ago @amulware told that he prefers composition over inheritance. Either is fine with me.

Toggl.Daneel/Views/ActivityIndicator.cs
arturs105: Oh, that's a mistake. I meant to set the value on the backing field itself, not in the getter.

Toggl.Daneel/Views/ActivityIndicator.cs
arturs105: Well this is called, when this view is added to another nib, but I guess I could refactor all the initialisation to constructors.

Toggl.Daneel/Views/ActivityIndicator.cs
arturs105: I was so preoccupied with whether or not I could, I didn’t stop to think if I should.

arturs105: Bit since this view will be hidden at some times, what happens with the animation? Does it have to be removed, or does iOS take care of this?

arturs105:
Just tried out different stuff and came to a few conclusions:

The `AwakeFromNib` override is unnecessary and will be removed.

Constructor `ActivityIndicatorView(IntPtr handle)` is needed to use this control from interface designer. And it must call `base(handle)`, so it can't call other constructors. That leaves us 3 options:
1. Leave the `init()` method as is and call it from all constructors (can't make `animation` and `imageView` `readonly`).
2. Make `animation` and `imageView` read only and duplicate initialisation code in the constructors.
3. Something I am missing.

Toggl.Daneel/Views/ActivityIndicator.cs
willsb: It will probably be there, but it won't render so /shrug

Toggl.Daneel/Views/ActivityIndicator.cs
willsb: Composition is fine in many a place, but if we're nesting views for no reason we're simply increasing complexity

willsb:
You can simply remove one constructor. IAGNI applies to the one since you're not using it anyways and it'll simplify the whole thing. Regarding the `duplicate init code` part, you could use a `Lazy<T>`, but now you'll only have a single constructor so it's fine :)

Toggl.Daneel/Views/ActivityIndicator.cs
arturs105: I see what you mean. So this is a case where inheritance wins.

Toggl.Daneel/Views/ActivityIndicatorView.cs
```diff
@@ -0,0 +1,46 @@
...
+            set => Image = UIImage.FromBundle(imageResource = value);
+        }
+
+        public ActivityIndicatorView(CGRect frame)
+            : base (frame)
+        {
+            init();
+        }
+
+        public ActivityIndicatorView(IntPtr handle)
```
willsb: This [should not have any initialization logic](https://developer.xamarin.com/api/constructor/UIKit.UIView.UIView/p/System.IntPtr/). 

Keep the constructor, but remove the call to `init`.

Bu doing so, you also completely remove the need of the init method, so you can simply do everything in the constructor instead :)

Toggl.Foundation.Tests/Toggl.Foundation.Tests.csproj
```diff
@@ -36,9 +36,9 @@
     <Reference Include="System" />
     <Reference Include="System.Core" />
     <Reference Include="System.Data" />
-    <Reference Include="System.ValueTuple, Version=4.0.1.1, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51, processorArchitecture=MSIL">
-      <HintPath>..\packages\System.ValueTuple.4.3.1\lib\netstandard1.0\System.ValueTuple.dll</HintPath>
-    </Reference>
+    <Reference Include="System.ValueTuple, Version=4.0.1.1, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51, processorArchitecture=MSIL">
```
willsb: I don't think any of those changes in .csprojs other than `Toggl.Daneel.csproj` is an actual change. Please `git checkout develop -- path/to/file/asd.csproj` before merging this.

Toggl.Daneel/Views/ActivityIndicatorView.cs
```diff
@@ -0,0 +1,46 @@
...
+        {
+            init();
+        }
+
+        private void init()
+        {
+            Image = UIImage.FromBundle(ImageResource);
+            ContentMode = UIViewContentMode.Center;
+
+            animation.Duration = 0.8;
```
willsb: Magic constant is magic

Toggl.Daneel/Views/ActivityIndicatorView.cs
arturs105: Thanks, I had missed this. But we can't remove `init`, if we want to use this control from .xib files - the other constructor does not get called and no animation is applied. Can be solved by calling init from `AwakeFromNib`.

Issue #182, opened 2017-06-07
Set up syncing state machine


amulware:
We have to decide how to implement the state machine responsible for syncing.

There may be different technical approaches, and we should figure out which will work best for us in the long term in terms of stability and maintainability.

willsb:
Consider [this](https://github.com/dotnet-state-machine/stateless)

PR #297, opened 2017-07-20
Add state machine for syncing algorithm

amulware:
Ref #182

This is the first and lowest level part of the syncing algorithm.

In itself, the state machine as implemented has absolutely no dependency on any of Toggl's models or business logic. It merely is the foundation on which I am building an easily expandable algorithm.

The types and concepts used here are:
- `IStateResult` - the outcome of a state
- `ITransition` - a transition from a state (knows the associated state result, and potential data that can be carried to the next state)
- `ITransitionHandlerProvider` - think of this as the flowchart that knows how all states are connected. Given a state result, it will return a `TransitionHandler` (it's basically a dictionary)
- `TransitionHandler` - turns a transition (obtained from the previous state) into the next state (this makes sure parameters are passed on correctly, etc.)

Each state is an `IObservable<ITransition>`, which is expected to return a single `ITransition` when it is done. That `ITransition` has an `IStateResult` associated with the outcome of the state. The `IStateResult` is handed to the `ITransitionHandlerProvider` which returns a `TransitionHandler` that can take the `ITransition` and return the next `IObservable<ITransition>` state.

That cycle continues until an error occurs, or we reach a state result that is not hooked up to another state, i.e. when syncing is done.

While not necessary to pass the tests, I already included the implementations of `ITransition` and `IStateResult`, since these are likely to be final (not to mention trivial).

As an example, here is how I imagine we will implement the various states:
```
class CheckIfAnythingToSyncState
{
    public StateResult<SyncItem> ItemToSync { get; } = new StateResult<SyncItem>();
    public StateResult NoItemToSync { get; } = new StateResult();

    public CheckIfAnythingToSyncState(ITogglDatabase database)
    {
        this.database = database;
    }

    public IObservable<Transition> Start()
    {
        return database
            .GetAnItemToSync()
            .SingleOrDefaultAsync()
            .Select(item => item == null
                ? NoItemToSync.Transition()
                : ItemToSync.Transition(item)
                );
    }
}
```
(Note that strictly speaking there is no need for separate classes for the states, but I think it will be nicer to test and organise that way. In either case, this state machine does not care.)

Lastly, my apologies for the messed up project and packages file. I'm not sure what went wrong there. If you can find the problem with them, please let me know and I'll fix it.


Toggl.Foundation.Tests/Toggl.Foundation.Tests.csproj
```diff
@@ -1,237 +1,230 @@
-<?xml version="1.0" encoding="utf-8"?>
```
willsb: `Diff too big to load` 🤔

Toggl.Foundation.Tests/packages.config
```diff
@@ -1,69 +1,69 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
```
willsb: You seem to fancy Windows 😄

Toggl.Foundation/Sync/StateMachine/StateMachine.cs
```diff
@@ -0,0 +1,81 @@
...
+
+        private readonly Subject<StateMachineEvent> stateTransitions = new Subject<StateMachineEvent>();
+        public IObservable<StateMachineEvent> StateTransitions { get; }
+
+        private readonly ITransitionHandlerProvider transitionHandlerProvider;
+        private readonly IScheduler scheduler;
+
+        private IDisposable currentState;
+
+        public StateMachine(ITransitionHandlerProvider transitionHandlerProvider, IScheduler scheduler)
```
willsb: Any reason for not defaulting this guy to `Scheduler.Main`?

Toggl.Foundation/Sync/StateMachine/StateMachine.cs
```diff
@@ -0,0 +1,81 @@
...
+        {
+            var transitionHandler = transitionHandlerProvider.GetTransitionHandler(transition.Result);
+
+            if (transitionHandler == null)
+            {
+                currentState = null;
+                reachDeadEnd(transition);
+            }
+            else
+                performTransition(transition, transitionHandler);
```
willsb: Please early return or add braces here. Braces on if and no braces on else is weird

Toggl.Foundation/Sync/StateMachine/StateMachineEvent.cs
```diff
@@ -0,0 +1,45 @@
+using System;
+
+namespace Toggl.Foundation.Sync
+{
+    public abstract class StateMachineEvent
+    {
+        public abstract override string ToString();
```
willsb: `abstract override` makes me happy ❤️

Toggl.Foundation/Sync/StateMachine/StateMachine.cs
```diff
@@ -0,0 +1,81 @@
...
+        }
+
+        private void reachDeadEnd(ITransition transition)
+        {
+            stateTransitions.OnNext(new StateMachineDeadEnd(transition));
+        }
+
+        private void reportError(Exception exception)
+        {
+            stateTransitions.OnNext(new StateMachineError(exception));
```
willsb: Why onNext instead of actually erroring? Can the StateMachine continue running after errors?

Toggl.Foundation.Tests/Sync/StateMachineTests.cs
```diff
@@ -0,0 +1,285 @@
...
+
+            [Fact]
+            public void ReportsErrorIfTheStateCompletesWithoutTransition()
+            {
+                var transition = GetTransitionSubstitute(_ => Observable.Empty<ITransition>());
+
+                StateMachine.PerformTransition(transition);
+
+                Events.ShouldBeSameEventsAs(
+                    Transition(transition),
```
willsb: Why is transition reported if it errored without transition? (similar case in test above)

Toggl.Foundation.Tests/Sync/StateMachineTests.cs
```diff
@@ -0,0 +1,285 @@
...
+            {
+                var stateSubject = new Subject<ITransition>();
+                var transition = GetTransitionSubstitute(_ => stateSubject.AsObservable());
+                var transition2 = GetTransitionSubstitute(_ => Observable.Never<ITransition>());
+
+                StateMachine.PerformTransition(transition);
+                Scheduler.AdvanceBy(TimeSpan.FromSeconds(55).Ticks);
+                stateSubject.OnNext(transition2);
+                stateSubject.OnCompleted();
+                Scheduler.AdvanceBy(TimeSpan.FromSeconds(55).Ticks);
```
willsb: Is this second .AdvanceBy needed?

Toggl.Foundation.Tests/Sync/StateMachineTests.cs
```diff
@@ -0,0 +1,285 @@
...
+            }
+
+            [Fact]
+            public void GetsNewStateObservableEveryTimeStateIsEntered()
+            {
+                var transition = GetTransitionSubstitute();
+                var deadEnd = GetTransitionSubstitute();
+                var observablesGotten = 0;
+                SetTransitionHandler(transition.Result, _ =>
+                    {
```
willsb: Isn't indenting here one level above correct?

Toggl.Foundation.Tests/Sync/StateMachineTests.cs
```diff
@@ -0,0 +1,285 @@
...
+                    Transition(transition),
+                    Error
+                );
+            }
+
+            [Fact]
+            public void GetsNewStateObservableEveryTimeStateIsEntered()
+            {
+                var transition = GetTransitionSubstitute();
+                var deadEnd = GetTransitionSubstitute();
```
willsb: These two lines made me realize this method is named in a little confusing way. I think `GetNewTransitionSubstitute` is a better name

Toggl.Foundation.Tests/Sync/StateMachineTests.cs
```diff
@@ -0,0 +1,285 @@
...
+                stateSubject.OnCompleted();
+
+                Events.ShouldBeSameEventsAs(
+                    Transition(transition),
+                    Error
+                );
+            }
+
+            [Fact]
+            public void GetsNewStateObservableEveryTimeStateIsEntered()
```
willsb: I believe this is misnamed. It should be `Notifies a new event` or `Gets new state` rather than `Gets New State Observable`

Toggl.Foundation.Tests/packages.config
amulware: As I said in the OP, it's a mess, and I have no clue why. Some whitespace changed, but I can't in what way locally. Fix suggestions/PRs welcome.

Toggl.Foundation/Sync/StateMachine/StateMachine.cs
amulware: No reason. Why should it though? In general, I like for things to be explicit.

Toggl.Foundation/Sync/StateMachine/StateMachine.cs
amulware: Will see what I can do!

Toggl.Foundation/Sync/StateMachine/StateMachine.cs
amulware: Yes. The state machine can be restarted afterwards (for the next sync). I could have done it differently, where you need to create a new state machine every time you wanna start it up, but I saw no reason to. I think this way it is easier to ensure that we don't do multiple things in parallel.

Toggl.Foundation/Sync/StateMachine/StateMachineEvent.cs
amulware: 💙

arturs105: 💚

Toggl.Foundation.Tests/Sync/StateMachineTests.cs
amulware: Because the transition is performed just fine (i.e. the follow-up state (Observable.Empty) is run). The error occurs because that state does not itself return a new transition.

Toggl.Foundation.Tests/Sync/StateMachineTests.cs
amulware: I'll split the test into two.

Toggl.Foundation.Tests/Sync/StateMachineTests.cs
amulware: The reason I named it this is that I wanted to make sure it calls the transition handler properly, creating a completely new state, instead of somehow using the same by resubscribing or something. I can rename it to `GetsNewState` if you like, but that seems essentially synonymous to me. The observable is the state is the observable.

Toggl.Foundation.Tests/Sync/StateMachineTests.cs
amulware: Ok.

Toggl.Foundation.Tests/Sync/StateMachineTests.cs
amulware: ✔️

Toggl.Foundation/Sync/StateMachine/StateMachine.cs
amulware: Not to mention that we only have to keep a state machine around, instead of a thing that can make state machines (some factory type that we don't need this way).

Toggl.Foundation.Tests/packages.config
willsb: Since there aren't many files to add back you could simply reset this and add them via VS 4 mac?

`git checkout develop -- path/to/Toggl.Foundation.Tests/packages.config path/to/Toggl.Foundation.Tests/packages.config`

Toggl.Foundation/Sync/StateMachine/StateMachine.cs
willsb: Omitting the Scheduler is pretty common in Rx methods, but sure

Toggl.Foundation.Tests/Sync/StateMachineTests.cs
willsb: Got it

Toggl.Foundation.Tests/Sync/StateMachineTests.cs
willsb: I see. Then it's not misnamed, but misleading. Could use a better name, though I can't think of one now. Ping @arturs105

Toggl.Foundation/Sync/StateMachine/StateMachine.cs
willsb: Makes sense

Toggl.Foundation/Sync/StateMachine/StateMachine.cs
```diff
@@ -0,0 +1,82 @@
...
+
+            StateTransitions = stateTransitions.AsObservable();
+        }
+
+        public void PerformTransition(ITransition transition)
+        {
+            Ensure.Argument.IsNotNull(transition, nameof(transition));
+            
+            if (currentState != null)
+                throw new InvalidOperationException("Cannot start state machine if it is already running.");
```
arturs105: If I understand correctly, then "Cannot perform a transition if another transition is in progress" would be better. Or is this method meant for starting the state machine?

amulware:
FIxed most erroneous whitespace changes. *shakes hands at windows/git/vs/whatever*

Toggl.Foundation/Sync/StateMachine/StateResult.cs
```diff
@@ -0,0 +1,21 @@
+namespace Toggl.Foundation.Sync
+{
+    public interface IStateResult { }
```
amulware: I could use an empty base class too, but I don't see a way of doing what I'm doing (the state result objects are used essentially as key to retrieve transition handlers) without a shared type. The reason there is more than one is that this makes passing parameters very convenient. I think this will be clearer once I submit the transition handler provider implementation. If you can come up with a better solution though, I won't complain.

Toggl.Foundation/Sync/StateMachine/StateMachineEvent.cs
```diff
@@ -0,0 +1,45 @@
...
+    internal class StateMachineTransition : StateMachineEvent
+    {
+        public ITransition Transition { get; }
+
+        public StateMachineTransition(ITransition transition)
+        {
+            Transition = transition;
+        }
+
+        public override string ToString() => "State machine is transitioning to new state.";
```
amulware: True, maybe we'll do that in the future. I did not want to complicate the interfaces too much for now, so there is very little information coming out of the whole syncing system right now.

amulware: Again, concrete proposals are welcome. :)

Toggl.Foundation/Sync/StateMachine/StateMachine.cs
amulware: Yes, this method is the one public way of controlling the machine, and can only be used to start it (with a given entry point transition), if it is not running. I could see us renaming the method to `Start` or similar if you prefer.

Toggl.Foundation.Tests/Sync/StateMachineTests.cs
arturs105: 🤷‍♂️

Toggl.Foundation/Sync/StateMachine/StateMachine.cs
arturs105: In that case I would prefer `Start`.

amulware:
@willsb @arturs105 Please review again. All points have been addressed to the best of my ability, except for the test naming (which I think is clear the way it is).

PR #298, opened 2017-07-20
Make git use only LFs in repository

kornelijepetak:
We should really use this.
It will prevent stray CRs in the codebase.


amulware:
We should probably have a PR that normalises all line endings of already existing files. Otherwise, we'll have strange diffs for weeks/months.

PR #299, opened 2017-07-24
Add test method templates

amulware:
This adds the following four templates: `fact` `factasync` `theory` and `property` for the respective test methods.

I have no clue if these work anywhere but in VS.

Squash if you are afraid of RSI.


arturs105:
Should this work on VS for Mac?

amulware:
I have no clue.

amulware:
After investigation, it seems that only VS on windows uses these template settings. Please approve it, or give reasons why you would not.

PR #305, opened 2017-07-25
Add the option to delete time entries.

arturs105:


Toggl.Foundation/DataSources/Interfaces/ITimeEntriesSource.cs
```diff
@@ -7,5 +7,6 @@ namespace Toggl.Foundation.DataSources
     public interface ITimeEntriesSource
     {
         IObservable<IEnumerable<ITimeEntry>> GetAll();
+        void Delete(ITimeEntry timeEntry);
```
amulware: Do we want the interface to be this? If I'm not mistaken, in the current implementation this will override the existing time entry with the one given, but also set the deleted flag, so it could be used to do more than just deleting it. Do we care? Are there better options?

arturs105: The only other option I see is deleting by id. And "deleting" by setting the `IsDeleted` will probably make syncing easier.

amulware: Setting the deleted flag is not the issue I have. I'm worried that there could be unintended side effects of changing other fields as well. Maybe it doesn't matter at all because they will be deleted, but maybe there is some case where it does matter, I do not know.

amulware: I confirmed that we do not care about making changes to a time entry if it is also being deleted. So the interface could be changed to take only the id. The question is if we like having a method like this taking an integer parameter? - This makes me think we really want type-safe ids...

arturs105: Well this method will probably be called from a view model and some view models don't hold any references to time entries, only some properties of them, so I think passing an id would be better.

The worst thing that can happen is an unused id is passed and we can't find anything with that in the database, so we either throw or ignore.

But then again, if an id is passed, we have to look up the time entry before deleting it.

Why not have both methods?

amulware: Why not have both: I consider the one where we delete a time entry by passing the whole thing to be sort of undefined behaviour. That's my main issue with the whole thing. Also: We always look up the time entry already anyways (within the transaction, be sure to keep it that way).

Speaking of transactions: Shouldn't this method return an observable? I.e. doesn't the repository.Update() return one? If nobody subscribes to that, the update will never actually take place.

arturs105: So deleting by id it is.

>  If nobody subscribes to that, the update will never actually take place.

Did not know about that. So it should return `IOBservable<ITimeEntry>`?

amulware: I think IObservable<Unit> to indicate success or failure should be enough.

Toggl.Foundation/DataSources/TimeEntriesDataSource.cs
```diff
@@ -19,6 +20,11 @@ public TimeEntriesDataSource(IRepository<IDatabaseTimeEntry> repository)
         }
 
         public IObservable<IEnumerable<ITimeEntry>> GetAll()
-            => repository.GetAll();
+            => repository.GetAll(te => !te.IsDeleted);
+
+        public IObservable<ITimeEntry> Delete(int id)
```
amulware: I still feel like this should be an `IObservable<Unit>`. Why would we ever want to get the time entry returned?

amulware:
Also, this needs tests. :)
Please include a test that simulates what happens when the time entry is deleted between the two repository calls.

Toggl.Foundation/DataSources/TimeEntriesDataSource.cs
```diff
@@ -19,6 +20,11 @@ public TimeEntriesDataSource(IRepository<IDatabaseTimeEntry> repository)
...
+            => repository.GetById(id)
+                         .Select(te => Models.TimeEntry.Dirty(te, true))
```
willsb: Please follow my suggestion from Slack and create a `Deleted` method instead of this. I have to read the method signature to understand what's going on with this boolean flag and if you're changing it so it gets deleted, it should be dirty already.

```csharp
repository
    .GetById(id)
    .Select(TimeEntry.Deleted)
    .Do(repositoty.Update)
    .Select(_ => Unit.Default);
```

Toggl.Foundation.Tests/Suggestions/MostUsedTimeEntryProviderTests.cs
```diff
@@ -68,7 +68,7 @@ public void ReturnsEmptyObservableIfThereAreNoTimeEntries()
 
             private IObservable<IEnumerable<IDatabaseTimeEntry>> createObservable(ITimeEntry[] timeEntries)
             {
-                var databaseTimeEntries = timeEntries.Select(Foundation.Models.TimeEntry.Clean);
+                var databaseTimeEntries = timeEntries.Select(te => Models.TimeEntry.Clean(te));
```
willsb: Needless closure if you follow my method :shipit:

Toggl.Foundation/DataSources/TimeEntriesDataSource.cs
amulware: We discussed this and I pointed out that Clean-Deleted is a valid state.

If you want more explicit methods, there have to be two of them. `CleanDeleted` and `DirtyDeleted`.

arturs105: But do we really need `CleanDeleted`? Will there ever be a use case, when we need to construct a deleted time entry that has been deleted?

And also: how about having 2 dirty methods - `Dirty(ITimeEntry te)` and `Dirty(ITImeEntry te, bool isDeleted)`? That would sole the needles closure issue.

Toggl.Foundation/DataSources/TimeEntriesDataSource.cs
```diff
@@ -19,6 +22,12 @@ public TimeEntriesDataSource(IRepository<IDatabaseTimeEntry> repository)
...
 
         public IObservable<IEnumerable<ITimeEntry>> GetAll()
-            => repository.GetAll();
+            => repository.GetAll(te => !te.IsDeleted);
+
+        public IObservable<Unit> Delete(int id)
+            => repository.GetById(id)
+                         .Select(TimeEntry.DirtyDeleted)
+                         .SelectMany(repository.Update)
+                         .Select(_ => Unit.Default);
```
amulware: Why do we return the Unit instead of ignoring elements?

Toggl.Foundation.Tests/DataSources/TimeEntriesDataSourceTests.cs
```diff
@@ -0,0 +1,76 @@
...
+                var dataSource = new TimeEntriesDataSource(repository);
+
+                Action tryingToDeleteRemovedTimeEntry = ()
+                    => dataSource.Delete(timeEntry.Id).Wait();
+
+                tryingToDeleteRemovedTimeEntry.ShouldThrow<EntityNotFoundException>();
+            }
+
+            [Fact]
+            public void TryingToDeleteNonExistantTimeEntryThrowsEntityNotFoundException()
```
amulware: This tests really just tests if the repository error is passed on correctly, nothing more. Needs better name.

Toggl.Foundation/DataSources/TimeEntriesDataSource.cs
arturs105: > I think `IObservable<Unit>` to indicate success or failure should be enough.

You said that last week.

Toggl.Foundation.Tests/DataSources/TimeEntriesDataSourceTests.cs
```diff
@@ -0,0 +1,76 @@
...
+using UltrawaveTimeEntry = Toggl.Ultrawave.Models.TimeEntry;
+
+namespace Toggl.Foundation.Tests.DataSources
+{
+    public class TimeEntriesDataSourceTests
+    {
+        public class TheDeleteMethod
+        {
+            [Fact]
+            public async Task UpdatesTheRepository()
```
amulware: Ideally, we should test that the model passed with `update` is identical to the one returned with `getbyid`. We should also test that the dirty flag is set as well.

Consider whether these should be the same or different tests.

Toggl.Foundation.Tests/DataSources/TimeEntriesDataSourceTests.cs
```diff
@@ -0,0 +1,76 @@
...
+                repository.GetAll().Returns(observable);
+                repository.GetById(Arg.Is(toBeDeleted.Id)).Returns(Observable.Return(toBeDeleted));
+
+                await dataSource.Delete(toBeDeleted.Id);
+
+                await repository.Received().Update(Arg.Is<IDatabaseTimeEntry>(te => te.IsDeleted && te.Id == toBeDeleted.Id));
+            }
+
+            [Fact]
+            public void TimeEntryGetsDeletedBeforeDeleteFinishes()
```
amulware: This name does not specify what should happen in that case.

Toggl.Foundation/DataSources/TimeEntriesDataSource.cs
amulware: Sure, but that is handled by the observable completed/error behaviour. The code right now will also return the `Unit.Default` element in the success case, which we will ignore. I wonder why we should even return it.

arturs105: Yeah, I see what you mean. We are only interested in `onError` and `onCompleted`. Will see what can be done.

Toggl.Foundation.Tests/DataSources/TimeEntriesDataSourceTests.cs
arturs105: I think those should be different tests.

Toggl.Foundation.Tests/DataSources/TimeEntriesDataSourceTests.cs
arturs105: 👍

Toggl.Foundation.Tests/DataSources/TimeEntriesDataSourceTests.cs
arturs105: 👍 Naming is hard

PR #307, opened 2017-07-25
Add View for StartTimeEntry

### LINKED ISSUE #311: Animations: Card
- This applies for both: time entry creation and edit cards.

ios-card

```javascript
cardInCurve = "cubic-bezier(0.23, 1, 0.32, 1)"
cardOutCurve = "cubic-bezier(0.86, 0, 0.07, 1)"
EnterTiming = .225
LeaveTiming = .195
```

willsb:
As the name implies. It shows up as a modal when you click the start button.

Closes #311

@pe0ny, please take a look at the gif :point_down: 

Squash when: ever


Toggl.Daneel/ViewControllers/MainViewController.cs
```diff
@@ -20,7 +20,7 @@ public override void ViewDidLoad()
             base.ViewDidLoad();
 
             var bindingSet = this.CreateBindingSet<MainViewController, MainViewModel>();
-
+            bindingSet.Bind(StartTimeEntryButton).To(vm => vm.StartTimeEntryCommand);
```
arturs105: The name `StartTimeEntryCommand` implies that this actually starts counting the time, maybe change to `ShowTimeEntryDialog` or `ShowStartTimeEntryDialog`?

Toggl.Daneel/ViewControllers/StartTimeEntryViewController.cs
```diff
@@ -0,0 +1,58 @@
...
+            => BottomDistanceConstraint.Constant = e.FrameBegin.Height + 0;
+
+        private void keyboardWillHide(object sender, UIKeyboardEventArgs e)
+            => BottomDistanceConstraint.Constant = 0;
+
+        private void prepareTextFields()
+        {
+            var stringAttributes = new CTStringAttributes(
+                new UIStringAttributes { ForegroundColor = Color.StartTimeEntry.Placeholder.ToNativeColor() }.Dictionary
+            );
```
arturs105: Should we use Egyptian style braces? Ping @amulware about this.

Toggl.Foundation.MvvmCross/Helper/Color.cs
```diff
@@ -27,5 +27,12 @@ public static class TimeEntriesLog
         {
             public static readonly MvxColor ButtonBorder = new MvxColor(243, 243, 243);
         }
+
+        public static class StartTimeEntry
+        {
+            public static readonly MvxColor Caret = new MvxColor(76, 217, 100);
```
arturs105: Is this for the checkmark at the bottom of the start time entry view? If so I suggest naming this `Checkmark`, since it isn't really a caret.

Toggl.Daneel/ViewControllers/StartTimeEntryViewController.cs
amulware: You mean the Egyptian parentheses that are already here? If so: I think that is fine.
If you mean splitting the UIStringAttributes initialisation into multiple lines using Egyptian braces, that is fine with me too.

arturs105:
I did a test on an iPhone and I think the animation looks kinda chaotic - it bounces quite quickly and out of screen, but I'll let @pe0ny be the judge of that.

Also the bottom sheet with the buttons doesn't always appear (see the video). Locking and unlocking the device makes them appear.

The buttons were also hidden on the simulator, when the autocomplete/suggestions were displayed above the keyboard.

[Video](https://www.youtube.com/watch?v=r-_CY3Vmqro)

Toggl.Foundation.MvvmCross/Helper/Color.cs
willsb: It's for the, well, the caret 😄 
The caret has a different color, check the designs

Toggl.Daneel/ViewControllers/MainViewController.cs
willsb: It does actually start counting the time, but it's transactional and can be cancelled by the user

Toggl.Foundation.MvvmCross/Helper/Color.cs
arturs105: Is that the green thingy on the bottom right? That doesn't look like a caret to me.

willsb: image
image
image
image
image

arturs105: Isn't that thing called a cursor? But it has exactly the same colour as the checkmark.

willsb: I don't know if it's the same color, but that thing is an image, so its color does not need to be anywhere

arturs105: Ok, I agree with that, but I still think that the name should be changed to `Cursor`, because caret (as described by [Wikipedia](https://en.wikipedia.org/wiki/Caret) and [Google images](https://www.google.lv/search?q=caret&client=safari&rls=en&source=lnms&tbm=isch&sa=X&ved=0ahUKEwiFqbu746bVAhXhCpoKHeu1D9oQ_AUICigB&biw=1440&bih=762)) is a completely different thing.

PR #310, opened 2017-07-26
Add Logout to DataSource

willsb:
Right now, logging out means deleting the user from the storage. The meaning might change in the future, but the method should be called logout instead of `Delete user`.

I thought about putting this in the login manager and rename it to something like `SessionManager`, but I didn’t like the idea. I’m mentioning just in case someone suggests; be aware that my vote is no.

Squash whenever, but be aware that this will conflict will #309, so whichever gets approved first will require some conflict solving.


amulware:
Should a log out not clear the entire database?

Toggl.Foundation/Login/LoginManager.cs
```diff
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Reactive;
```
amulware: Who dis?

Toggl.Foundation/DataSources/Interfaces/IUserSource.cs
```diff
@@ -1,10 +1,6 @@
...
-using Toggl.Multivac.Models;
-
 namespace Toggl.Foundation.DataSources
 {
     public interface IUserSource
     {
     }
-}
\ No newline at end of file
+}
```
amulware: Unrelated cleanup.

PR #316, opened 2017-07-27
Add state machine orchestrator

amulware:
Clean PR is clean. 💅 

Some background:

The state machine from #297 was very low-level and entirely Toggl-agnostic.

This new class, the StateMachineOrchestrator knows a little bit more about Toggl. It knows that there is a PullSync, a PushSync, and Sleep state for our syncing system (as signified by the included `enum`).

It also knows how to start the state machine into the correct state using the properties of the injected StateMachineEntrypoints.

It's main responsibility then is to start the state machine as indicated, watch over it, report back when a Pull or Push has completed (Sleep never completes, since it does not actually cause the state machine to start). It also enforces (threw exceptions, but no worry those will never happen in production) that the state machine is not started while it is already busy.

This class contains observables that will be used to further control it from the next higher level of abstraction, but these observables will also be made available by that level to be consumed by the UI to see what the syncing system is doing (Pull/Push/Sleep) on a high level.

Note that this class is not thread-safe. That (very necessary) functionality will come on the next level.


Toggl.Foundation/Sync/StateMachineEntryPoints.cs
```diff
@@ -0,0 +1,8 @@
+namespace Toggl.Foundation.Sync
+{
+    public sealed class StateMachineEntryPoints
+    {
+        public StateResult StartPullSync { get; } = new StateResult();
```
amulware: What is the same object? StateResults are used keys in a dictionary (ITransitionHandlerProvider) so they have to be unique, even though they don't do much.

Toggl.Foundation/Sync/StateMachine/Transition.cs
```diff
@@ -15,7 +15,7 @@ public Transition(StateResult result)
         }
     }
 
-    internal sealed class Transition<T> : ITransition
+    public sealed class Transition<T> : ITransition
```
amulware: Basically, I had a lot of trouble keeping these internal, because that means I have to make tests internal too, which doesn't really work, or not easily at least. I'm sure it can be done, but I'd then rather take a look at this again once all parts of the syncing machine are merged.

Toggl.Foundation/Sync/SyncState.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace Toggl.Foundation.Sync
+{
+    public enum SyncState
+    {
+        Sleep = 0,
```
amulware: It shouldn't make a difference. I have the habit of adding it. If you prefer I can remove it. I sorta like to specify it to indicate what the default state is. (And that I mean one of the states to be the default)

Toggl.Foundation.Tests/Sync/StateMachineOrchestratorTests.cs
```diff
@@ -0,0 +1,288 @@
...
+                new StateMachineOrchestrator(stateMachine, new StateMachineEntryPoints());
+                
+                transitions.Received().Subscribe(Arg.Any<IObserver<StateMachineEvent>>());
+            }
+        }
+
+        public class TheStateProperty : StateMachineOrchestratorBaseTests
+        {
+            [Fact]
+            public void ReturnsSleepAtFirst()
```
amulware: Sure.

Toggl.Foundation.Tests/Sync/StateMachineOrchestratorTests.cs
```diff
@@ -0,0 +1,288 @@
...
+            {
+                Orchestrator.State.Should().Be(Sleep);
+            }
+        }
+        
+        public abstract class StateChangeMethod : StateMachineOrchestratorBaseTests
```
amulware: Ok.

Toggl.Foundation.Tests/Sync/StateMachineOrchestratorTests.cs
```diff
@@ -0,0 +1,288 @@
...
+        {
+            protected abstract SyncState ExpectedState { get; }
+            protected abstract void CallMethod();
+
+            private Action callingMethod => CallMethod;
```
amulware: The tests read nicer.

Toggl.Foundation.Tests/Sync/StateMachineOrchestratorTests.cs
```diff
@@ -0,0 +1,288 @@
...
+            [Fact]
+            public void ShouldNotCauseCompletedEvent()
+            {
+                CallMethod();
+
+                CompletedEvents.Should().BeEmpty();
+            }
+        }
+
+        public abstract class PullPushSyncMethod : StateChangeMethod
```
amulware: Ok.

Toggl.Foundation/Sync/IStateMachineOrchestrator.cs
```diff
@@ -0,0 +1,15 @@
...
+{
+    public interface IStateMachineOrchestrator
+    {
+        SyncState State { get; }
+        IObservable<SyncState> StateObservable { get; }
+        IObservable<SyncState> SyncCompleteObservable { get; }
+        
+        void StartPushSync();
+        void StartPullSync();
+        void GoToSleep();
```
willsb: Agreed

Toggl.Foundation.Tests/Sync/StateMachineOrchestratorTests.cs
willsb: I see, but why not name the abstract method like that then? That's not nitpicking, I'm just curious :smile:

amulware: Because I still want to `CallMethod();` 

The real question maybe is why is it an action-property instead of a method? Is there even a difference? Who knows!

Toggl.Foundation/Sync/StateMachineOrchestrator.cs
```diff
@@ -0,0 +1,79 @@
...
+using Toggl.Multivac;
+
+namespace Toggl.Foundation.Sync
+{
+    internal sealed class StateMachineOrchestrator : IStateMachineOrchestrator
+    {
+        private readonly IStateMachine stateMachine;
+        private readonly StateMachineEntryPoints entryPoints;
+
+        private readonly BehaviorSubject<SyncState> state = new BehaviorSubject<SyncState>(SyncState.Sleep);
```
arturs105: I find this name a bit confusing (especially on line 18). Maybe `stateBehaviourSubject`?

Toggl.Foundation/Sync/SyncState.cs
```diff
@@ -0,0 +1,9 @@
...
+        Pull,
+        Push
+    }
+}
```
arturs105: \n

Toggl.Foundation/Sync/StateMachineOrchestrator.cs
amulware: That seems very Hungarian to me, and would not be very consistent (with for example the `syncComplete` subject just below. But I can see the confusion and will see what I can do.

amulware: In general, we should maybe have a discussion on how we think observables should be named, both privately and publicly.

amulware:
@willsb @arturs105 Please review again! :)

PR #317, opened 2017-07-27
Time service

willsb:
:clock1: Service is useful for:
 
:clock2: Abstracting time 
:clock3: Help testing 
:clock4: Synchronize UI that depends on the current time

etc  :clock5:, etc :clock6:, etc :clock7: 

Squash if there’s no time to wait :clock8: (there actually is no time at all :clock9:)


willsb:
Good call! This whole interface basically screams for us to expose `IConnectableObservable` instead of a simple `IObservable` to make its intentions clear. Here, have some: @paul 🌮  🌮

Toggl.Foundation/TimeService.cs
```diff
@@ -0,0 +1,27 @@
...
+
+        public TimeService(IScheduler scheduler)
+        {
+            CurrentDateTimeObservable =
+                Observable
+                    .Interval(TimeSpan.FromSeconds(1), scheduler)
+                    .Select(_ => DateTimeOffset.UtcNow)
+                    .Publish();
+
+            disposable = CurrentDateTimeObservable.Connect();
```
amulware: Do we have to hold on to this?

willsb: I don't see harm in doing so

Toggl.Daneel/Startup/Setup.cs
```diff
@@ -1,68 +1,72 @@
...
-            return navigationService;
+        {
+            navigationService = base.InitializeNavigationService();
+            return navigationService;
         }
-
+
         protected override IMvxApplication CreateApp()
         {
             base.InitializeFirstChance();
```
arturs105: I know this isn't a change in this PR, but why is this called from `CreateApp()`?

Toggl.Foundation.MvvmCross/ViewModels/MainViewModel.cs
```diff
@@ -9,16 +9,19 @@ namespace Toggl.Foundation.MvvmCross.ViewModels
 {
```
arturs105: GH doesn't allow adding comments to the first 8 lines, but the first `using` statement has become unnecessary.

Toggl.Daneel/Startup/Setup.cs
willsb: Because it's responsible for creating the `IMvxApplication`. Part of the MvvmCross lifecycle

PR #318, opened 2017-07-29
Fix flaky suggestions test

willsb:
Problem 1: This test depends on a ViewModel, so it depends on MvvmCross
Problem 2: Tests are async and the order is not ensured

Result: If this test is ever the first to run (as it was [here](https://www.bitrise.io/build/0515d881909c5875)) the whole MvvmCross singleton thingy is not initialized and will explode horribly in our faces.

Fix: Make the test depend on MvvmCross

Squash whenever.


arturs105:
Is the dependency on MvvmCross in `MvxViewModel`, because I didn't see anything that would cause problems in `SuggestionsViewModel` and `BaseViewModel`? If that is the case we should probably do this for all view model tests.

willsb:
The problem is [here](https://github.com/MvvmCross/MvvmCross/blob/develop/MvvmCross/Core/Core/ViewModels/MvxNotifyPropertyChanged.cs#L39). This is already being done, as `BaseViewModelTests` inherits from `BaseMvvmCrossTests`. The problem only happens in this one class :)

PR #324, opened 2017-07-31
Add the option to specify modal view height

arturs105:


willsb:
I'm not a big fan of this approach. It looks really fragile, cumbersome and it `CGRect`s. I'd rather try and use constraints instead, since those are more resilient to changes in the window size (something that happens when the user receives a call, for example)

arturs105:
Good point, constraints probably would be better. Where do you suggest these constraints be set? In the view controller itself or somewhere else?

amulware:
This has not changed in almost a week. What's gonna happen? @arturs105 @willsb

arturs105:
I am going to look at this today, but probably have to wait for @willsb since he might have a better idea how to do this.

arturs105:
So the height for modals can now be set from the view controller by setting the `PreferredContentSize`. The size can also be changed/animated from the view controller by manipulating the `Frame` and `PreferredContentSize` properties.

Toggl.Daneel/Presentation/TogglPresenter.cs
```diff
@@ -55,9 +55,9 @@ protected virtual void ShowNestedViewController(UIViewController viewController,
         protected virtual void ShowModalCardViewController(UIViewController viewController, MvxBasePresentationAttribute attribute, MvxViewModelRequest request)
         {
             viewController.ModalPresentationStyle = UIModalPresentationStyle.Custom;
-            viewController.TransitioningDelegate = FromBottomTransitionDelegate;
```
willsb: Is the order change still needed?

Toggl.Daneel/Presentation/Transition/FromBottomTransitionDelegate.cs
```diff
@@ -25,6 +25,7 @@ public IUIViewControllerAnimatedTransitioning GetAnimationControllerForDismissed
         public IUIViewControllerInteractiveTransitioning GetInteractionControllerForDismissal(IUIViewControllerAnimatedTransitioning animator)
             => swipeInteractionController.InteractionInProgress ? swipeInteractionController : null;
 
-        public void WireToViewController(UIViewController vc) => swipeInteractionController.WireToViewController(vc);
```
willsb: `git checkout develop -- Toggl.Daneel/Presentation/Transition/FromBottomTransitionDelegate.cs`

arturs105:
Yesterday I tested only on an emulator and it worked, just tried out on an iPhone and it still works (with and without `PreferredContentSize`).

PR #327, opened 2017-08-01
Add view model for edit view

arturs105:
The `confirm` method hasn't been implemented on the data source yet.

~~The `delete` method is being added in #305.~~


Toggl.Foundation.MvvmCross/ViewModels/EditViewModel.cs
```diff
@@ -0,0 +1,76 @@
...
+            DeleteCommand = new MvxCommand(delete);
+            ConfirmCommand = new MvxCommand(confirm);
+            CloseCommand = new MvxAsyncCommand(close);
+        }
+
+        private void delete()
+        {
+            deleteDisposable = dataSource
+                .Delete(Id)
+                .Subscribe(null, onDeleteError, onDeleteCompleted);
```
willsb: Why not simply pass `close` instead?

willsb: Also, consider creating an overload that only takes error and completed as parameters

Toggl.Foundation.MvvmCross/ViewModels/EditViewModel.cs
```diff
@@ -0,0 +1,76 @@
...
+using MvvmCross.Core.Navigation;
+using MvvmCross.Core.ViewModels;
+using PropertyChanged;
+using Toggl.Foundation.DataSources;
+using Toggl.Multivac;
+using Toggl.Multivac.Models;
+
+namespace Toggl.Foundation.MvvmCross.ViewModels
+{
+    public class EditViewModel : BaseViewModel
```
willsb: `BaseViewModel<ITimeEntry>`, since the TimeEntry is a crucial part not present here

willsb: I'd like this to be called `EditTimeEntryViewModel`, for consistency with `StartTimeEntryViewModel` (and also because we might have other edit views)

Toggl.Foundation.MvvmCross/ViewModels/EditViewModel.cs
arturs105: `close` has the wrong return type. And what is the general case for navigation service methods - should they be awaited, or can they be used as fire and forget?

arturs105:
You mean the time service is needed for updating the duration field when editing running time entries?

willsb:
Exactly

Toggl.Foundation.MvvmCross/ViewModels/EditTimeEntryViewModel.cs
```diff
@@ -0,0 +1,102 @@
...
+            CloseCommand = new MvxAsyncCommand(close);
+        }
+
+        public override async Task Initialize(ITimeEntry parameter)
+        {
+            await base.Initialize();
+
+            Id = parameter.Id;
+            Description = parameter.Description;
+            Duration = TimeSpan.FromSeconds(parameter.Duration);
```
amulware: We should not use time entry duration. In fact, we might remove it from the model entirely. Better to use the nullable end date which has much more sensible semantics.

amulware: And of course, take into account that we can edit running time entries as well.

Toggl.Foundation.Tests/MvvmCross/ViewModels/EditTimeEntryViewModelTests.cs
```diff
@@ -0,0 +1,33 @@
...
+using Toggl.Foundation.DataSources;
+using Toggl.Foundation.MvvmCross.ViewModels;
+using Toggl.Foundation.Tests.Generators;
+using Xunit;
+
+namespace Toggl.Foundation.Tests.MvvmCross.ViewModels
+{
+    public class EditTimeEntryViewModelTests
+    {
+        public class TheConstructor : BaseViewModelTests<EditTimeEntryViewModel>
```
willsb: Test `CloseCommand` and `DeleteCommand `

PR #328, opened 2017-08-01
Add all named colors from zeplin

arturs105:


amulware:
I observe that pinkish gray is really just gray.

PR #329, opened 2017-08-01
Generate class data for constructor

willsb:
Right now this is too error prone. Too much copy pasta. It’s better now. Most of your “whys” are most likely solved reading [this](http://hamidmosalla.com/2017/02/25/xunit-theory-working-with-inlinedata-memberdata-classdata/) (the ClassData section, to be more specific). Feel free to ask any other questions.


amulware:
Arturs is right, we should have that test too, just to make sure.

Toggl.Foundation.Tests/Generators/ConstructorTestData.cs
```diff
@@ -0,0 +1,83 @@
...
+            [Theory]
+            [InlineData(typeof(TwoParameterConstructorTestData))]
+            [InlineData(typeof(ThreeParameterConstructorTestData))]
+            public void NeverReturnsASequenceWhereAllElementsAreTrue(Type contructorTestDataType)
+            {
+                var testData = Activator.CreateInstance(contructorTestDataType) as ConstructorTestData;
+
+                testData.Any(x => x.All(y => (bool)y))
+                        .Should()
+                        .BeFalse();
```
amulware: Can this not use the `Should().NotContain()` (or something like that) methods?

Toggl.Foundation.Tests/Generators/ConstructorTestData.cs
```diff
@@ -0,0 +1,83 @@
...
+            }
+
+            [Theory]
+            [InlineData(typeof(TwoParameterConstructorTestData), 2)]
+            [InlineData(typeof(ThreeParameterConstructorTestData), 3)]
+            public void ReturnsAllPossiblePermutationsForTheSpecifiedNumberOfParameters(Type contructorTestDataType, int parameterCount)
```
amulware: This test is very complicated and hard to read. What is wrong with two simple property tests taking 2/3 bool parameters, making an array out of them and running a 'contains' check? (returning early for all true)

I don't see what we have to gain by making this as difficult to read as the class being tested.

Toggl.Foundation.Tests/Generators/ConstructorTestData.cs
willsb: No because of the nested lists

Toggl.Foundation.Tests/Generators/ConstructorTestData.cs
```diff
@@ -0,0 +1,79 @@
...
+                var testData = Activator.CreateInstance(contructorTestDataType) as ConstructorTestData;
+                testData.Any(x => x.All(y => (bool)y)).Should().BeFalse();
+            }
+
+            [Property]
+            public void ReturnsAllPossiblePermutationsForTwoParameters(bool first, bool second)
+            {
+                if (first && second) return;
+
+                var array = new[] { first, second }.Select(x => (object)x);
```
amulware: `new object[] { first, second }` doesn't work (and is the cast necessary anyways?)

PR #330, opened 2017-08-02
Add TimeEntry builder

willsb:
As discussed yesterday. I had to make properties private so they can be set via other constructors. The builder itself only requires the properties needed for a very basic time entry. It will become more complete as we implement a more complex time entry adding mechanism.

Squash whenever


Toggl.Foundation/Models/FoundationModels.tt
```diff
@@ -26,17 +27,17 @@ namespace Toggl.Foundation.Models
 
         WriteLine($"    internal partial class {className} : IDatabase{className}");
         WriteLine( "    {");
-        WriteLine( "        public int Id { get; }\n");
+        WriteLine( "        public int Id { get; private set; }\n");
```
arturs105: Why are the setters added? They aren't used in the builder.

Toggl.Foundation/Models/TimeEntry.cs
```diff
@@ -0,0 +1,65 @@
...
+        {
+            public static Builder SetStart(DateTimeOffset start) => new Builder(start);
+
+            public DateTimeOffset Start { get; }
+
+            public bool IsDirty { get; private set; }
+
+            public bool Billable { get; private set; }
+
+            public string Description { get; private set; }
```
arturs105: Are these the only properties that will be allowed to set? Or is this just for now and more properties will get added as needed?

Toggl.Foundation/Models/TimeEntry.cs
```diff
@@ -0,0 +1,65 @@
...
+            public Builder SetBillable(bool billable)
+            {
+                Billable = billable;
+                return this;
+            }
+
+            public void EnsureValidity()
+            {
+                if (string.IsNullOrEmpty(Description))
+                    throw new InvalidOperationException("You need to set the description before building");
```
arturs105: I suggest changing this to "...before building a time entry", since we'll have builders for several models.

Toggl.Foundation/Models/TimeEntry.cs
```diff
@@ -0,0 +1,65 @@
+﻿using System;
+using Toggl.Multivac.Models;
+
+namespace Toggl.Foundation.Models
+{
+    internal partial class TimeEntry
+    {
+        internal sealed class Builder
+        {
+            public static Builder SetStart(DateTimeOffset start) => new Builder(start);
```
simonrozsival: It seems strange to me that you create the builder by calling SetStart, shouldn't there be some standard name of the method (something like "CreateBuilder") which would be used in the other builders? Of course with different parameters  required for each entity.

Toggl.Foundation/Models/FoundationModels.tt
amulware: I'm not sure I like the models now being mutable either. Even if only privately.

Toggl.Foundation/Models/TimeEntry.cs
```diff
@@ -0,0 +1,65 @@
...
+            public string Description { get; private set; }
+
+            private Builder(DateTimeOffset start)
+            {
+                Start = start;
+            }
+
+            public TimeEntry Build() => new TimeEntry(this);
+
+            public Builder SetIsDirty(bool isDirty)
```
amulware: Wouldn't `WithIsDirty()` (etc below) be better naming, if we are chaining already?

Toggl.Foundation/Models/TimeEntry.cs
```diff
@@ -0,0 +1,65 @@
...
+                return this;
+            }
+
+            public Builder SetBillable(bool billable)
+            {
+                Billable = billable;
+                return this;
+            }
+
+            public void EnsureValidity()
```
amulware: Each time entry should also have a valid start date, and be part of a workspace. We also have to see where we get the time entry id from.

Toggl.Foundation/Models/TimeEntry.cs
willsb: They will be added as needed, yes

Toggl.Foundation/Models/TimeEntry.cs
willsb: 👍

Toggl.Foundation/Models/TimeEntry.cs
willsb: We can vote. I prefer `Set`, but I don't really mind if the majority prefers `With`

Toggl.Foundation/Models/TimeEntry.cs
willsb: Not sure it's worth validating the workspace and the id part now. Let's make this incremental

Toggl.Foundation/Models/TimeEntry.cs
willsb: I did that to ensure that start gets set when constructing. Consumers have no other option. But since we are validating everything anyways, might as well have a `CreateBuilder()` method instead.

Toggl.Foundation/Models/TimeEntry.cs
```diff
@@ -0,0 +1,81 @@
...
+                Billable = billable;
+                return this;
+            }
+
+            private void ensureValidity()
+            {
+                if (Start == default(DateTimeOffset))
+                    throw new InvalidOperationException(string.Format(ErrorMessage, "start date"));
+
+                if (string.IsNullOrEmpty(Description))
```
amulware: Why though? Empty descriptions are supported.

Toggl.Foundation/Models/TimeEntry.cs
```diff
@@ -0,0 +1,81 @@
...
+                    throw new InvalidOperationException(string.Format(ErrorMessage, "description"));
+            }
+        }
+
+        internal static TimeEntry Stopped(ITimeEntry runningTimeEntry, DateTimeOffset stop)
+            => new TimeEntry(runningTimeEntry, stop);
```
amulware: Where will it validate that stop>=start?

amulware: And why is this not done using a `With` method and `New<>` like we discussed in the past?

Toggl.Foundation/Models/TimeEntry.cs
willsb: Are they? Didn't know that. Should I still check for null, though?

amulware: I see no real issue with 'null', and I expect that our API would turn it into "" anyways... But yeah, sure, let's enforce "" instead anyways.

Toggl.Foundation/Models/TimeEntry.cs
willsb: Good points!

I feel we can start with `WithStopDate` and, if things get out of hand, we add the other constructor.

Regarding the validation: It should happen both here and in the ViewModel

Toggl.Foundation/Models/TimeEntry.cs
```diff
@@ -61,8 +61,13 @@ private void ensureValidity()
             }
         }
 
-        internal static TimeEntry Stopped(ITimeEntry runningTimeEntry, DateTimeOffset stop)
-            => new TimeEntry(runningTimeEntry, stop);
+        internal TimeEntry WithStopDate(DateTimeOffset stop)
+        {
+            if (Start < stop)
```
amulware: This test is inverted.

PR #332, opened 2017-08-02
Add state tree

amulware:
💅 This is a simple container class implementing the `ITransitionHandlerProvider` interface.
This tree is essentially the flow chart that we will build once on startup and then inject into the state machine.

The way this will work is as follows. If we have this following example state:
```
class CheckIfAnythingToSyncState
{
    public StateResult<SyncItem> ItemToSync { get; } = new StateResult<SyncItem>();
    public StateResult NoItemToSync { get; } = new StateResult();

    public CheckIfAnythingToSyncState(ITogglDatabase database)
    {
        this.database = database;
    }

    public IObservable<Transition> Start()
    {
        return database
            .GetAnItemToSync()
            .SingleOrDefaultAsync()
            .Select(item => item == null
                ? NoItemToSync.Transition()
                : ItemToSync.Transition(item)
                );
    }
}
```

We could hook up the `NoItemToSync` result to the same start method (for infinite retries):
```
state = new CheckIfAnythingToSyncState(db);
tree.ConfigureTransition(state.NoItemToSync, state.Start);
```
Of course that is a contrived example, and most results would lead to different states, not the same. Still the syntax holds.


Toggl.Foundation.Tests/Sync/StateTreeTests.cs
```diff
@@ -0,0 +1,149 @@
...
+            where TStateResult : class, new()
+            where TStateFactory : class
+        {
+            protected StateTree Tree { get; } = new StateTree();
+
+            protected abstract void CallMethod(TStateResult result, TStateFactory factory);
+            protected abstract TStateFactory ToStateFactory(Action action);
+
+            [Theory]
+            [InlineData(true, false)]
```
willsb: These should use the recently merged `ClassData` instead

Toggl.Foundation.Tests/Sync/StateTreeTests.cs
```diff
@@ -0,0 +1,149 @@
...
+
+                callingMethodWithNull.ShouldThrow<ArgumentNullException>();
+            }
+
+            [Fact]
+            public void ReturnsNullIfThereAreNoHandlers()
+            {
+                var handler = tree.GetTransitionHandler(Substitute.For<IStateResult>());
+
+                handler.Should().BeNull();
```
willsb: Just to make sure I understood it all: This null is treated as "This is a deadend", correct?

Toggl.Foundation.Tests/Sync/StateTreeTests.cs
```diff
@@ -0,0 +1,149 @@
...
+                tree.ConfigureTransition(new StateResult(), () => null);
+                tree.ConfigureTransition(new StateResult<object>(), _ => null);
+
+                var handler = tree.GetTransitionHandler(Substitute.For<IStateResult>());
+
+                handler.Should().BeNull();
+            }
+
+            [Fact]
+            public void ReturnsHandlerThatCallsProvidedStateFactory()
```
willsb: This lack the indefinite article after `Returns` and a definite one after `Calls`

Toggl.Foundation.Tests/Sync/StateTreeTests.cs
```diff
@@ -0,0 +1,149 @@
...
+                tree.ConfigureTransition(stateResult, () => expectedResult);
+
+                var handler = tree.GetTransitionHandler(stateResult);
+                var actualResult = handler(null);
+
+                actualResult.Should().Be(expectedResult);
+            }
+
+            [Fact]
+            public void ReturnsHandlerThatCallsProvidedGenericStateFactory()
```
willsb: This lack the indefinite article after `Returns` and a definite one after `Calls`

Toggl.Foundation/Sync/StateTree.cs
```diff
@@ -0,0 +1,39 @@
+using System;
+using System.Collections.Generic;
+using Toggl.Multivac;
+
+namespace Toggl.Foundation.Sync
+{
+    public sealed class StateTree : ITransitionHandlerProvider
```
willsb: I think it would be a lot clearer to name this `TransitionHandlerProvider`. When I read tree I implied there would be nodes/leaves, but this is simply a Dictionary on steroids

Toggl.Foundation.Tests/Sync/StateTreeTests.cs
amulware: Yes.

Toggl.Foundation/Sync/StateTree.cs
amulware: It's a bit like the queue that isn't a real queue... :P

amulware:
Actually, I won't rename the test methods to add articles. It makes them longer and not at all more readable. We also are not consistently using articles otherwise. What is/was your argument/reasoning here?

PR #333, opened 2017-08-02
Add sync manager

### LINKED ISSUE #182: Set up syncing state machine

amulware:
Closes #182.

This one is big and important. It contains two classes that used to be one, but it made more sense to me to split them for easier testing, and clearer code.

The `SyncManager` does little but make sure that all state-changes (Sleep, Push, Pull) are performed in a threadsafe manner, no matter who tries to tell it what to do when.

The thread safety test works well (removing locks breaks it horribly), but is unfortunately extremely ugly.

Secondly, there is the `SyncStateQueue` which hides the key point that each Pull sync is followed by a Push, and in general makes sure we don't forget to sync in the right way, even if the user/system prompts us to sync while we are already doing so.

The interface of `SyncManger` sort of hides the pull/push behaviour, and in essence only exposes it as either `Push`, or `FullSync`. This means it sort of implies the implementation of the queue, which of course is not guaranteed, but I'm not sure how to improve this - except for maybe changing the `QueuePullSync` method to `QueueFullSync` on the queue?
Suggestions welcome.

In general, I'm sure this can be improved, so please point at things!


Toggl.Foundation/Sync/SyncStateQueue.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        public bool StartNextQueuedState(IStateMachineOrchestrator orchestrator)
+        {
+            if (pulledLast)
+            {
+                startPush(orchestrator);
+            }
+            else if (pullSyncQueued)
+            {
+                startPull(orchestrator);
```
simonrozsival: This is technically not a queue - if you start with all bools set to false and then you call `QueuePushSync()` and `QueuePullSync()` in either order and afterwards `StartNextQueuedState()`, it would always start with Pull and then Push (which is what should happen if I understand the specs correctly).

The implementation is quite straightforward and I like it, I just expected something else as it is called *queue*. It is rather a very short todo list...

Toggl.Foundation/Sync/ISyncStateQueue.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace Toggl.Foundation.Sync
+{
+    public interface ISyncStateQueue
+    {
+        void QueuePushSync();
+        void QueuePullSync();
+
+        bool StartNextQueuedState(IStateMachineOrchestrator orchestrator);
```
amulware: I did not really want to expose that to the sync manager, who does not need to know. But I also agree that booleans are bad. Do you think returning the new state is the lesser evil?

Toggl.Foundation.Tests/Sync/SyncManagerTests.cs
```diff
@@ -0,0 +1,343 @@
...
+                Orchestrator.State.Returns((SyncState)stateValue);
+
+                SyncManager.State.Should().Be((SyncState)stateValue);
+            }
+        }
+
+        public class TheStateObservable : SyncManagerTestBase
+        {
+            [Fact]
+            public void ShouldReturnObservableFromOrchestrator()
```
amulware: See the discussion on the state-tree pr.

amulware:
About the queue: Suggestions for a better name are welcome, but as you say: It's _sorta_ like a queue, and I was not able to find anything better.

Toggl.Foundation/Sync/ISyncStateQueue.cs
willsb: I think it's better, yes

amulware:
Also added a test for that return value which was missing. All yours @willsb.

PR #334, opened 2017-08-02
API exceptions

### LINKED ISSUE #150: Improve API error return value
Right now we are returning pretty useless exceptions with the http content of the response as message.

Soon we will also need _at least_ the http error code, for both the status endpoint, and all other endpoints as well.

simonrozsival:
Closes #150 

This PR includes specific exception classes for different errors of API requests. I also extracted the large switch statement from the `BaseApi.cs` file into a separate helper.

All of these exceptions should be tested, but since this PR already contains too many files, I would suggest putting those tests into a separate PR.

Squash whenever.


Toggl.Ultrawave/Exceptions/ApiErrorResponseException.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System;
+using System.Net;
+
+namespace Toggl.Ultrawave.Exceptions
+{
+    public class ApiErrorResponseException : ApiException
+    {
+        public readonly HttpStatusCode HttpCode;
```
willsb: Public fields are a big no 🙅‍♂️ 
Use properties instead.

Though in this case, I'm against exposing/storing the HttpCode altogether.

Toggl.Ultrawave/Helpers/ApiExceptions.cs
```diff
@@ -0,0 +1,52 @@
+﻿using Toggl.Ultrawave.Exceptions;
+using Toggl.Ultrawave.Network;
+
+namespace Toggl.Ultrawave.Helpers
+{
+    class ApiExceptions
```
willsb: Make access modifiers explicit

willsb: Also, this helper method could be in the `ApiErrorResponseException` class, in my opinion

simonrozsival:
I created those exceptions based on a list of API errors of our backend Paul gave me.

Toggl.Ultrawave/Helpers/ApiExceptions.cs
simonrozsival: It could be there, I don't like the idea of having a static method inside of an exception class. But if others agree with your suggestion, I can move the code there.

Toggl.Ultrawave/Exceptions/ApiErrorResponseException.cs
simonrozsival: OK, I'll turn that into a property.

There might be a case when the API returns a different status code than we expect. Then this is a catch-it-all case. I could add one more exception, which would be something like `UnknownErrorResponseException` which would hold the HTTP status code and others thus won't have to.

amulware: I don't want an inherited catch-all class. The base class can and should be used for that in my opinion.

amulware: And I agree with exposing the status code so that we can handle it separately if needed.

Toggl.Ultrawave/Exceptions/TooManyRequestsException.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System.Net;
+
+namespace Toggl.Ultrawave.Exceptions
+{
+    public sealed class TooManyRequestsException : ClientErrorException
+    {
+        /// <summary>
+        /// HTTP status code 429 "Too Many Redirects" is not included in the System.Net.HttpStatusCode enum.
+        /// </summary>
```
amulware: Remove XML.

Toggl.Ultrawave/Helpers/ApiExceptions.cs
amulware: I'm indifferent about where this is and am fine with leaving it here.

Toggl.Ultrawave/Exceptions/ApiErrorResponseException.cs
amulware: Actually, scratch that. (see review below)

Toggl.Ultrawave/Exceptions/ForbiddenException.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System.Net;
+
+namespace Toggl.Ultrawave.Exceptions
+{
+    public sealed class ForbiddenException : ApiErrorResponseException
```
simonrozsival: Yes, I forgot to change it.

Toggl.Ultrawave/Exceptions/ClientErrorException.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System.Net;
+
+namespace Toggl.Ultrawave.Exceptions
+{
+    public abstract class ClientErrorException : ApiErrorResponseException
+    {
+        protected ClientErrorException(string errorMessage, HttpStatusCode httpCode)
```
simonrozsival: I don't think this is necessary at runtime as it is an abstract class - you can't instantiate it with some random HTTP status code, It is then up to the classes inheriting from this one.

simonrozsival: But this is irrelevant - I will create the *unknown error* class and all the other won't store the status code.

simonrozsival:
I think I addressed all your comments.

willsb:
@simonrozsival Please dismiss our reviews when you address all comments, as specified [here](https://github.com/toggl/mobile-docs/blob/develop/pull-request-etiquette.md#making-changes-to-pull-requests)

Toggl.Ultrawave/Exceptions/ClientDeprecatedException.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System.Net;
+
+namespace Toggl.Ultrawave.Exceptions
+{
+    public sealed class ClientDeprecatedException : ClientErrorException
+    {
+        /// <summary>
+        /// HTTP status code 418 "I Am A Teapot" is not included in the System.Net.HttpStatusCode enum.
+        /// </summary>
```
simonrozsival: I removed them.

Toggl.Ultrawave/Exceptions/ServerErrorException.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System.Net;
+
+namespace Toggl.Ultrawave.Exceptions
+{
+    public abstract class ServerErrorException : ApiErrorResponseException
+    {
+        protected ServerErrorException(string errorMessage, HttpStatusCode httpCode)
```
simonrozsival: Same as above.

Toggl.Ultrawave/Exceptions/TooManyRequestsException.cs
simonrozsival: Removed.

simonrozsival:
Thanks for pointing that out to me, @willsb! I'll read the documentation thoroughly once more.

amulware:
Please also include a note what issue this closes in the PR description as outlined in the docs.

Toggl.Ultrawave/Exceptions/ClientErrorException.cs
amulware: Let's still test (in a future PR is fine) that the correct exception is created for each code. :)

simonrozsival: I am already working on the tests. I will also create an issue so we don't forget about it.

amulware:
Now, please squash with appropriate emoji, according to the guidelines. :D

PR #335, opened 2017-08-02
Add IObservable.Subscribe(onError, onCompleted) extension

arturs105:


amulware:
Would it not be nicer (more intuitive) to switch the order of arguments? i.e. `(onCompleted, onError)` to have the desired behaviour first, and the exception after?

arturs105:
@willsb It wouldn’t work because multivac doesn’t have the assembly for that extension method.

@amulware I did this because of the method Subscribe(onNext, onError, onCompletes) - it has error handler before the complete handler.

amulware:
I see. Strange choice. Go ahead then.

PR #336, opened 2017-08-02
Fix namespace for ObservableCollectionExtensions

arturs105:


amulware:
Apart from now failing, why did this need to be changed?

willsb:
It's inside the Extensions folder, but its namespace was the root folder's. It fails now because he most likely forgot to change the namespace in places where this was being used.

amulware:
We have some folders that do not provide namespace on purpose though. Should we have some sort of policy for this?

PR #337, opened 2017-08-02
Start time entry

willsb:
This is starting to look like an actual app :shipit: 

Squash whenever


Toggl.Daneel/LinkerPleaseInclude.cs
```diff
@@ -1,117 +1,117 @@
-using System.Collections.Specialized;
```
amulware: This file did not change, except for whitespace, please exclude.

Toggl.Foundation/DataSources/TimeEntriesDataSource.cs
```diff
@@ -30,5 +42,21 @@ public IObservable<Unit> Delete(int id)
...
+
+        public IObservable<ITimeEntry> Start(DateTimeOffset startTime, string description, bool billable)
+            => TimeEntry.Builder.Create()
+                .SetStart(startTime)
+                .SetDescription(description)
+                .SetBillable(billable)
+                .SetIsDirty(true)
+                .Build()
+                .Apply(repository.Create)
+                .Do(currentTimeEntrySubject.OnNext);
```
amulware: Does this wait for a response from the repo before setting the currently running entry? Wouldn't that potentially delay the UI? I'd rather have it set the current entry, and then update it after the repo call if needed (though not sure why it would be, unless there is some sort of failure)

willsb: What do you mean "Delay the UI"? You mean a hiccup after the dismissal of the StartTimeEntryViewController? From my tests, this didn't happen (Realm is blazing fast) and I'd rather only set the currently running TE after I know it's safely stored

amulware: Yeah, I mean that you don't get to see the time entry, but it actually looks like nothing happened for 100-200ms, which might confuse people. Regarding the speed of realm, did you test under otherwise heavy load (on call, or doing something else in the background), with tens of thousands of entities in the database?

Toggl.Foundation.Tests/DataSources/UserDataSourceTests.cs
```diff
@@ -0,0 +1,95 @@
...
+
+            protected DateTimeOffset ValidTime { get; } = DateTimeOffset.Now;
+
+            protected IRepository<IDatabaseTimeEntry> Repository { get; } = Substitute.For<IRepository<IDatabaseTimeEntry>>();
+
+            public UserDataSourceTest()
+            {
+                TimeEntriesSource = new TimeEntriesDataSource(Repository);
+                Repository.Create(Arg.Any<IDatabaseTimeEntry>())
+                          .Returns(info => info.Arg<IDatabaseTimeEntry>().Apply(Observable.Return));
```
amulware: This really doesn't need the `Apply`.

willsb: Let me use my pet method :shipit:

PR #339, opened 2017-08-02
Fix CRLF issues

willsb:
Some more :P

Squash whenever


amulware:
This PR does something else than changing only whitespace. Please at least change title and try again.

willsb:
🙄

PR #340, opened 2017-08-02
Preserve ViewModels

willsb:
This is the first part of fixing the release®

This preserves the constructors for our ViewModels and for the Bindings, allowing the project to run in Debug mode.

There still are a few issues (converters, for example, don’t work), but those will be fixed in different PRs

Squash whenever.


amulware:
I'm really not a big fan of this. We are bound to forget to add future stuff to these classes and run into troubles. Is there no other solution?

willsb:
You can make the Linker skip entire assemblies and carry loads of useless code, but I'm 100% against using weaker linker settings. I'm afraid this approach doesn't require us to like it :|

amulware:
Except, how much code will we realistically have in our projects that will not be used?

willsb:
The problem is not in our projects. This is mostly covered by those generators (since the linker will only skip code that gets reached by reflection, and only by reflection). 

The problem is carrying MvvmCross in its entirety to each one of our apps

amulware:
I understand, but you said we can skip an assembly. Can we not skip the assemblies of our projects only? Or are they not separate from the packages, etc.?

amulware:
Where did the `TimeEntriesViewModel.cs` go?

willsb:
It was a file lying there and getting in the way of linkage. It was renamed to `TimeEntriesLogViewModel`, but probably sneaked in during some merge. I 🔥  it.

Issue #100, opened 2017-05-05
Convert model interface to concrete implementation in API


amulware:
Otherwise the serializer won't work with attributes.

amulware:
It might be enough to add constructors to all models that take their own interface and set all their properties and then calling it like this before serialisation:

```
var apiTag = tag as Tag ?? new Tag(tag);
```

tag tag.

Also it would be cool if we could somehow check that we aren't missing this conversion anywhere (by checking in the serialiser that we are serialising a concrete class, not an interface? Alternatively, we can cover this in unit tests. Opinions welcome).

willsb:
👍 for covering it all in unit tests.

amulware:
If possible, we should generate the constructors for the models, instead of writing them manually.

willsb:
We are already doing it! Jus approve #104 to see :D

PR #341, opened 2017-08-02
Preserve Ultrawave models

willsb:
The models here are needed for every client, but they get stripped away by the linker because there are no direct references to their constructors. The `PreserveAttribute` makes the linker ignore the class and its members when stripping code.

Squash whenever


amulware:
How come these models aren't generated?

willsb:
We might need to do a lot of annotation on those models because of serialization

amulware:
Right.

amulware:
Thing is, this will not be needed, once we fix #100. Is this still worth merging?

willsb:
I don't think the model generation should apply Ultrawave. The models are very tailored for our needs (specially because of annotations due to, say, backend weirdly naming fields or ignore-on-serialize rules).

>Is this still worth merging?
The app literally can't (de)serialize anything without merging this, so yes, there's value

amulware:
I'm not sure what you are talking about. With #100, the model classes will be explicitly mentioned in code. Should that not prevent them from being removed?

Also, I'm not arguing for generation in any way, your comment about annotation is valid.

willsb:
Oh, I see your point. I thought you were still defending generation.

It boils down to: Can we wait until #100 is fixed before releasing test builds? If yes, close this. Otherwise, merge this.

amulware:
We can wait for #100, but let's get around to that soon then.

PR #344, opened 2017-08-04
Stop time entries

willsb:
Pretty straight forward method to stop the currently running TimeEntry (and notify that there no longer is a time entry running).

Squash whenever


Toggl.Foundation/DataSources/TimeEntriesDataSource.cs
```diff
@@ -53,10 +53,12 @@ public IObservable<ITimeEntry> Start(DateTimeOffset startTime, string descriptio
                 .Apply(repository.Create)
                 .Do(currentTimeEntrySubject.OnNext);
 
+        public IObservable<ITimeEntry> Stop(DateTimeOffset stopTime)             => repository                     .GetAll(runningTimeEntries)                     .SelectMany(timeEntries =>
```
amulware: Please fix formatting.

Toggl.Foundation.Tests/DataSources/TimeEntriesDataSourceTests.cs
```diff
@@ -98,7 +99,34 @@ public class TheStartMethod : TimeEntryDataSourceTest
                 await Repository.Received().Create(Arg.Is<IDatabaseTimeEntry>(te => te .Start == currentlyRunningTimeEntry.Start));
             }
         }
+         public class TheStopMethod : TimeEntryDataSourceTest         {
```
willsb: Okay, Xamarin Studio simply gave up trying on this specific branch. This is shit.

Toggl.Foundation/Models/TimeEntry.cs
```diff
@@ -75,3 +86,8 @@ private TimeEntry(Builder builder)
         }
     }
 }
+
+internal static class TimeEntryExtensions
+{
+    public static TimeEntry With(this IDatabaseTimeEntry self, DateTimeOffset stop) => new TimeEntry(self, stop);
```
arturs105: Why is this an extension method and not just a regular method on the `TimeEntry` class? And also why is it outside the namespace?

Toggl.Foundation/DataSources/TimeEntriesDataSource.cs
```diff
@@ -55,7 +55,16 @@ public IObservable<ITimeEntry> Start(DateTimeOffset startTime, string descriptio
                 .Build()
                 .Apply(repository.Create)
                 .Do(safeSetCurrentlyRunningTimeEntry);
-    
+                
+        public IObservable<ITimeEntry> Stop(DateTimeOffset stopTime)
+            => repository
+                    .GetAll(runningTimeEntries)
```
amulware: Will this load all time entries into memory, or will it run the query in realm directly somehow?

Toggl.Foundation.Tests/DataSources/TimeEntriesDataSourceTests.cs
```diff
@@ -111,6 +110,41 @@ public class TheStartMethod : TimeEntryDataSourceTest
             }
         }
 
+        public class TheStopMethod : TimeEntryDataSourceTest
```
amulware: We need a test for failure cases as well. What if the repository does not return any entries? Also, as far as I can tell we are not actually testing the argument to GetAll - so we cannot be sure it will actually take the running time entry.

Toggl.Foundation/DataSources/TimeEntriesDataSource.cs
```diff
@@ -55,7 +55,16 @@ public IObservable<ITimeEntry> Start(DateTimeOffset startTime, string descriptio
...
+                    .SelectMany(timeEntries =>
+                        timeEntries.Single()
+                            .With(stopTime)
+                            .Apply(repository.Update)
+                            .Do(_ => safeSetCurrentlyRunningTimeEntry(null)));
```
amulware: This is two transactions again. What if sync modifies the same time entry in some way, or what if sync makes some other time entry the running one?

Toggl.Foundation/Models/TimeEntry.cs
willsb: Because this is meant to work even with stuff that come from the DB. It should be inside a namespace, though

Toggl.Foundation/DataSources/TimeEntriesDataSource.cs
willsb: 👍

PR #345, opened 2017-08-04
Ensure thread safety for current Time Entry

willsb:
Apparently Realm has many a thread problem with their entities being used directly in the UI. 

This PR ensures that  `RealmTimeEntries` are never exposed outside the DataSource.

Squash whenever


Toggl.Foundation/DataSources/TimeEntriesDataSource.cs
```diff
@@ -51,12 +51,22 @@ public IObservable<ITimeEntry> Start(DateTimeOffset startTime, string descriptio
...
-            => currentTimeEntrySubject.OnNext(timeEntries.SingleOrDefault());
+            => safeSetCurrentlyRunningTimeEntry(timeEntries.SingleOrDefault());
+            
+        private void safeSetCurrentlyRunningTimeEntry(ITimeEntry timeEntry)
+        {
+            var next = timeEntry;
+            if (timeEntry != null)
+            {
+                next = TimeEntry.Clean(timeEntry);
+            }
```
amulware: I think using the ternary operator would be clearer here.

PR #346, opened 2017-08-04
Fix TimeEntry realm table

willsb:
No setter = no property weaving = table without columns = no persistency :sob: 

Squash whenever.


Toggl.PrimeRadiant.Realm/Models/RealmTimeEntry.cs
```diff
@@ -2,39 +2,40 @@
 using Realms;
 using Toggl.PrimeRadiant.Models;
 using System.Collections.Generic;
+using Toggl.Multivac.Models;
```
amulware: Why this guy?

PR #347, opened 2017-08-04
Use LocalId for persisting entities locally

willsb:
Realm does not have auto increment Ids. A GUID is a lot cheaper than querying the entire DB for checking the last valid id, so this is why I’m using it.

The additional constructors are there to ensure that objects that exist in the database will have their `LocalId` preserved.

Squash whenever.


amulware:
We discussed and decided to use the same property for local ids and to use negative numbers, since that will make things significantly easier and less error prone. I am not at all a fan of using a separate id like this, since now every place that has to check/get/whatever by id has to consider two different fields, in a variety of combinations.

amulware:
Instead we could simply have a single row in the database to store the last used negative id, and fetch that, and even have the fetching operator update the value in the database to ensure this won't break.

Unless you have a good solution for how to handle having two different ids, I cannot approve this.

amulware:
Please also refer to internal discussion (and consider continuing there).

amulware:
Won't happen.

PR #350, opened 2017-08-04
Model interfaces to implementation api

### LINKED ISSUE #100: Convert model interface to concrete implementation in API

simonrozsival:
Every `Toggl.Ultrawave` model now has a constructor which copies data from its interface. I included unit tests for the constructors (one test per model).


amulware:
At a quick glance, this looks good, but it does not close the issue. For that, these constructors also have to be used in all places where models are serialised to ensure we never send non-ultrawave model implementations to the serialiser (since only these models have the appropriate attributes for json property names and other oddities). I'd like this usage to be as safe as possible. If at all possible, I'd like it if we could not even serialise the interfaces themselves, but only the concrete model to ensure the correct types are used. But I'm not sure how feasible that is.

simonrozsival:
OK, I'll include that in the PR.

amulware:
Or let's maybe have a second one? easier to review. If you just remove the 'Closes' part above, we can probably merge this quickly.

amulware:
@willsb Please make sure to check the generator, since you know more about that stuff.

Toggl.Ultrawave.Tests/Models/BaseModelTests.cs
```diff
@@ -9,6 +11,15 @@ public abstract class BaseModelTests<T>
         protected abstract T ValidObject { get; }
 
         [Fact]
+        public void HasConstructorWithOneParameter()
```
willsb: Adding `where T : new()` as a type constraint achieves the same result in a more elegant way

simonrozsival: I agree that using `Activator` isn't elegant at all. But how does that constraint achieve the same result? That constraint just enforces that the type `T` has *a public parameterless constructor* (https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/new-constraint), whereas I wanted to test that there is a constructor which copies all of the properties from the argument into the new instance of the type `T`.

willsb: Oh, I see. My bad, I misread the test. Either way, please keep the constraint, since the parameterless constructor is needed for serialization 😄 

Also, you might wanna rename this guy to make it more specific (`Has one parametersless constructor that copies blablabla`)

simonrozsival: OK

PR #352, opened 2017-08-04
Set negative Ids to entities

willsb:
- Changed all ids to longs to avoid clashes (this could potentially be done in a separate PR, but the changes are straight forward and they all happen in a single commit, so it’s simple to spot errors, it any)
- Introduces the `IIdProvider` interface
- Implement it in Realm
- Use it when starting a TimeEntry

Now we can have more than one time entry at a time in the database. Sweet 

Squash ASAP


willsb:
Yeah, lots of CRLF x.x

PR #355, opened 2017-08-07
Add Stylecop to all projects

### LINKED ISSUE #38: Use linting
Consider EditorConfig, and checking this on Bitrise.

### LINKED ISSUE #321: Add StyleCop to all projects
I've been experimenting with it, it seems to work very well, and we can configure it to cause build errors when set rules are broken. We only have to see if it works on Mac as well.

Here is a PR that shows how easily this can be done: https://github.com/beardgame/utilities/pull/109

Ref #38

amulware:
What is the correct capitalisation? I have no idea.

In either case, this works in VS on windows and mac alike, as well as in Bitrise and any other Roslyn compiler. Note how some of the commits failed to build on Bitrise (and in <2m) after I enabled some of the rules, and then build again once I fix the issues.

This closes #321, closes #38.

There are many more rules and more detailed settings we can use for Stylecop, but we can enable those over time to not get too crazy PRs.


amulware:
Note that the Toggl.ruleset is shared between all projects. I thought that, while less flexible, this would be easier and less prone to errors.

amulware:
I don't. But I'd say we enable them over time, as many as possible, if they support our code style. The ones I enabled here for example clearly do, since the two things fixed were isolated cases.

PR #356, opened 2017-08-08
Fix typo causing builds to fail

amulware:
Squash if you want to look at empty onboarding screens on your phone.


willsb:
mfw linksip
image

PR #357, opened 2017-08-09
Use Realm relationships in database

willsb:
As discussed in slack. Check #streeling for more information.

Squash ASAP.


Toggl.Foundation/Models/FoundationModels.cs
```diff
@@ -17,6 +17,8 @@ internal partial class Client : IDatabaseClient
 
         public DateTimeOffset? ServerDeletedAt { get; }
 
+        public IDatabaseWorkspace Workspace { get; }
```
willsb: It's generated by T4 templates and then compiled rather than using `.g.cs` shenanigans, so changes should be committed

willsb:
@amulware There's no need to add the `Ignore` attribute on get-only props

PR #361, opened 2017-08-09
Add EditTimeEntryViewController

arturs105:
This adds just the view controller, bindings will come in another PR once this and #327 get approved.


amulware:
Please remove the comment. If something is so important it needs a `// TODO:` comment to remind us, use an issue instead. Kthxbai.

PR #365, opened 2017-08-10
Update MvvmCross

willsb:
Note on the third commit: 

From version 5.1 onwards, the NavigationService now depends on App. Since our App depends on NavigationService, nothing would compile.

I changed it so that the needed dependencies get injected when initializing rather than when constructing and it’s all fine and ServiceLocator free :shipit: 

Squash whenever

### DIFF
--- a/Toggl.Daneel/Startup/Setup.cs
+++ b/Toggl.Daneel/Startup/Setup.cs
@@ -6,6 +6,7 @@ using MvvmCross.iOS.Platform;
 using MvvmCross.iOS.Views.Presenters;
 using MvvmCross.Platform;
 using MvvmCross.Platform.Platform;
+using MvvmCross.Platform.Plugins;
 using Toggl.Daneel.Presentation;
 using Toggl.Daneel.Services;
 using Toggl.Foundation;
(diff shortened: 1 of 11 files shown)


Toggl.Daneel/Startup/Setup.cs
```diff
@@ -39,15 +40,23 @@ public Setup(MvxApplicationDelegate applicationDelegate, IMvxIosViewPresenter pr
...
         }
 
-        protected override IMvxNavigationService InitializeNavigationService()
+        protected override void InitializeLastChance()
         {
-            navigationService = base.InitializeNavigationService();
-            return navigationService;
+            base.InitializeLastChance();
+
+            Mvx.RegisterSingleton<IPasswordManagerService>(new OnePasswordService());
```
arturs105: Is there a reason for this to be in `InitializeLastChance` instead of `InitializeApp`?

Toggl.Foundation.MvvmCross/App.cs
```diff
@@ -10,21 +10,19 @@ namespace Toggl.Foundation.MvvmCross
 {
     public class App : MvxApplication
     {
-        private readonly AppStart appStart;
+        public override void Initialize()
```
arturs105: Why is the empty override necessary?

Toggl.Daneel/Startup/Setup.cs
willsb: Because this is supposed to override the stup implementation that's provided for platforms that do not support Password managers

Toggl.Foundation.MvvmCross/App.cs
willsb: To make it explicitly do nothing, even if the base method does something

PR #366, opened 2017-08-10
Fix nested model query in realm

willsb:
Straight forward: 

Nullability is fine and expressions can’t have the null-coalescing operator, hence the need for variables.

Squash whenever


Toggl.PrimeRadiant.Realm/Models/RealmConstructors.tt
```diff
@@ -56,8 +56,10 @@ namespace Toggl.PrimeRadiant.Realm
             if (isRealmProperty)
             {
                 var entityName = propName.Replace("Id", "");
+                WriteLine($"            var actual{entityName}Id = entity?.{entityName}Id ?? 0;");
+
                 propName = $"Realm{entityName}";
-                assignment = $"realm.All<{propName}>().Single(x => x.Id == {entityName}Id)";
+                assignment = $"realm.All<{propName}>().SingleOrDefault(x => x.Id == actual{entityName}Id)";
```
simonrozsival: Wouldn't it make more sense to skip the query altogether if we know in advance, that it will return the default value? The ternary operator could be used here instead `Realm{entityName} = entity == null ? null : realm.All<...>().Single(x => ...);` or is there some reason why this is a better way to do it?

Toggl.PrimeRadiant.Realm/Models/RealmConstructors.tt
```diff
@@ -56,8 +56,10 @@ namespace Toggl.PrimeRadiant.Realm
             if (isRealmProperty)
             {
                 var entityName = propName.Replace("Id", "");
+                WriteLine($"            var actual{entityName}Id = entity?.{entityName}Id ?? 0;");
```
simonrozsival: I understand that there is probably no entity with the ID equal to zero, but I still don't like using magical constants.

willsb: It's a bit magical, yes. But comparing stuff to null throws horribly, so there's that

Toggl.PrimeRadiant.Realm/Models/RealmConstructors.tt
willsb: Actually there is none, you are correct and your solution is beautiful 🌈

Toggl.PrimeRadiant.Realm/Models/RealmConstructors.cs
```diff
@@ -23,8 +23,8 @@ public RealmClient(IDatabaseClient entity, Realms.Realm realm)
         public RealmClient(IClient entity, Realms.Realm realm)
         {
             Id = entity.Id;
-            var actualWorkspaceId = entity?.WorkspaceId ?? 0;
-            RealmWorkspace = realm.All<RealmWorkspace>().SingleOrDefault(x => x.Id == actualWorkspaceId);
+            var skipWorkspaceFetch = entity?.WorkspaceId == null || entity.WorkspaceId == 0;
```
amulware: A client always belongs to a workspace, so this is not really a legal check.

amulware: Similar for all below. Everything always has a workspace, and the value cannot be null or 0.

Toggl.PrimeRadiant.Realm/Models/RealmConstructors.cs
```diff
@@ -116,12 +116,12 @@ public RealmTask(ITask entity, Realms.Realm realm)
         {
             Id = entity.Id;
             Name = entity.Name;
-            var actualProjectId = entity?.ProjectId ?? 0;
-            RealmProject = realm.All<RealmProject>().SingleOrDefault(x => x.Id == actualProjectId);
-            var actualWorkspaceId = entity?.WorkspaceId ?? 0;
-            RealmWorkspace = realm.All<RealmWorkspace>().SingleOrDefault(x => x.Id == actualWorkspaceId);
-            var actualUserId = entity?.UserId ?? 0;
-            RealmUser = realm.All<RealmUser>().SingleOrDefault(x => x.Id == actualUserId);
+            var skipProjectFetch = entity?.ProjectId == null || entity.ProjectId == 0;
```
amulware: Tasks always have projects as well.

Toggl.PrimeRadiant.Realm/Models/RealmConstructors.cs
```diff
@@ -163,8 +163,8 @@ public RealmTimeEntry(ITimeEntry entity, Realms.Realm realm)
             TagIds = entity.TagIds;
             At = entity.At;
             ServerDeletedAt = entity.ServerDeletedAt;
-            var actualUserId = entity?.UserId ?? 0;
-            RealmUser = realm.All<RealmUser>().SingleOrDefault(x => x.Id == actualUserId);
+            var skipUserFetch = entity?.UserId == null || entity.UserId == 0;
```
amulware: I'm not 100% on this, but I believe that this can also not be unset.

amulware:
Did I put my comments on the wrong commit? 😕

PR #368, opened 2017-08-10
Add TimeSpan to duration string converter

arturs105:
This converter is meant for the edit view and is different from the other one as this adds the unit and trims the leading zeros:

12:12:12 => 12:12:12
12:12 => 12:12 min
12 => 12 sec


Toggl.Foundation.MvvmCross/Converters/TimeSpanToDurationWithUnitValueConverter.cs
```diff
@@ -0,0 +1,38 @@
...
+{
+    public class TimeSpanToDurationWithUnitValueConverter : MvxValueConverter<TimeSpan, string>
+    {
+        protected override string Convert(TimeSpan value, Type targetType, object parameter, CultureInfo culture)
+        {
+            var unit = getUnitFor(value);
+            var durationBuilder = new StringBuilder();
+
+            if (value.TotalHours > 1)
+                durationBuilder.Append(((int)value.TotalHours).ToString("00"))
```
arturs105: This cast will throw an exception if `TotalHours > int.Max`, can we safely assume no one will track 2 billion hours?

simonrozsival: I guess this depends on if you can put in arbitrary start/end date or if there are some limits in those pickers.

simonrozsival: We could do something like `return value.TotalHours > OneHunderedYearsInHours ? "way too long..." : convertToString(value);`

PR #369, opened 2017-08-10
Add since parameter storage to database

### LINKED ISSUE #359: Store `since` parameters in database
To not have to download too much data on every pull, some of the API's endpoints support a `?since=` parameter, which will not download data that has not be changed since the given date.

We need to store the last such date for each entity separately in the database.

I propose we store this in a single object with one property for each. No need to complicate.

amulware:
Closes #359

Pretty straight forward I think. I decided to store these in a single object, since at least right now we will always need them all at the same time. There is an argument to be made to set them one by one, as requests return, but I'm not sure that makes things significantly easier (the fetching algo will be a bit involved anyways, we'll see when we get there).


amulware:
@willsb Yes, only the sync algo needs access to this. I'm all up for limiting access somehow, but then we need a second database interface with 'only sync stuff' and then we have to pass around both types of databases and *breathes in* erm.. yeah. Suggestions welcome :D

PR #370, opened 2017-08-10
Change Rate type to Float

willsb:
I had deserialization problems when fetching projects from backend because the rate was 1.4, so a float is what we need ¯\_(ツ)_/¯

Squash whenever


amulware:
Any particular reason why float, not double?

willsb:
Yeah, just checked on the backend project, they use a Float64. I'll change it :)

PR #372, opened 2017-08-11
Implement Get Since endpoints

### LINKED ISSUE #360: Add GET `?since=` endpoints to api
Ref #359.

Workspaces, tags, clients, projects, tasks and time entries should get additional Get ?since= end points.

amulware:
Closes #360

These will be needed for faster fetching and syncing of data.

So far only the client endpoints are implemented, but feel free to review, since the others will be analogous.

Note that there is one thing I'm not testing: Whether the API actually uses the last-changed date as expected (and I know it does) or the date the entity was created. I cannot test this easily for anything but time entries since only those have update-API implemented, but I'll add a test very similar to the existing ones for time entries when I add the endpoint.

Also note that add the `long ToUnixTimeSeconds(this DateTimeOffset dateTime)` extension method, since we cannot use .net 4.6 yet. It has the same signature and does the same thing (minus boundary checks, because it's not terribly necessary for us) as the one from the framework, so switching this out in the future will be painless.

### DIFF
--- a/Toggl.Ultrawave/ApiClients/Interfaces/IClientsApi.cs
+++ b/Toggl.Ultrawave/ApiClients/Interfaces/IClientsApi.cs
@@ -7,6 +7,7 @@ namespace Toggl.Ultrawave.ApiClients
     public interface IClientsApi
     {
         IObservable<List<Client>> GetAll();
+        IObservable<List<Client>> GetAllSince(DateTimeOffset threshold);
         IObservable<Client> Create(Client client);
     }
 }
--- a/Toggl.Multivac/Extensions/DateTimeOffsetExtensions.cs
+++ b/Toggl.Multivac/Extensions/DateTimeOffsetExtensions.cs
@@ -1,0 +1,17 @@
+﻿using System;
+
+namespace Toggl.Multivac.Extensions
+{
+    public static class DateTimeOffsetExtensions
+    {
+        public static long ToUnixTimeSeconds(this DateTimeOffset dateTime)
+        {
+            // constant and implementation taken from .NET reference source:
+            // https://referencesource.microsoft.com/#mscorlib/system/datetimeoffset.cs
+            const long unixEpochSeconds = 62_135_596_800L;
+
+            var seconds = dateTime.UtcDateTime.Ticks / TimeSpan.TicksPerSecond;
+            return seconds - unixEpochSeconds;
+        }
+    }
+}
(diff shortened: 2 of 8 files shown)


Toggl.Ultrawave/ApiClients/Interfaces/IClientsApi.cs
```diff
@@ -7,6 +7,7 @@ namespace Toggl.Ultrawave.ApiClients
     public interface IClientsApi
     {
         IObservable<List<Client>> GetAll();
+        IObservable<List<Client>> GetAllSince(DateTimeOffset threshold);
```
simonrozsival: Wouldn't it be cleaner to just have one `GetAll(DateTimeOffset? threshold)` which would add the since parameter if the value is not `null`?

Also will we add any more query parameters in the future? We could use something like a dictionary to store multiple parameters and then convert it into query string to make it more general. It is probably not a big deal for `Clients`, but there might be more parameters for different API clients and I suppose it would be nice to have a similar methods in the interfaces.

simonrozsival: (I can see the PR is still wip but I wanted you to consider changing this as you are still working on it atm)

amulware: I prefer not using default parameters in public APIs. If this were ever to become a package, default parameters can cause problems with binary level compatibility. (a weak argument for sure)
In general, I think having two methods makes it more expressive, and clear.

There also will not be any more parameters in the future. This is the only exception.

simonrozsival: OK, it makes sense to do it like this.

amulware:
Still getting some failing integration tests. Not sure why. Will have to investigate further.

amulware:
I had a lot of struggle getting the tests to run, mostly because the time entries used in our tests had weird semi-valid data. I changed that to more a bit more sense now, and everything works like a charm.

Review away.

Toggl.Ultrawave.Tests.Integration/ProjectsApiTests.cs
```diff
@@ -1,7 +1,8 @@
 ﻿﻿﻿using System;
 using System.Linq;
 using System.Collections.Generic;
-using System.Reactive.Linq;
+  using System.Linq.Expressions;
+  using System.Reactive.Linq;
```
simonrozsival: extra spaces

amulware: Huh.

amulware:
Btw: I probably should have said somewhere below that I ran integration tests, and link to it passing on Bitrise. I hope it suffices to say that yes indeed, they pass.

PR #375, opened 2017-08-14
Add edit view bindings

arturs105:
This adds almost all bindings and other necessary stuff for edit view, however edit view is still not complete - the rest will come in other PRs.


Toggl.Daneel/ViewControllers/EditTimeEntryViewController.cs
```diff
@@ -17,6 +23,86 @@ public override void ViewDidLoad()
         {
             base.ViewDidLoad();
 
+            DurationLabel.Font = DurationLabel.Font.GetMonospacedDigitFont();
+
+            var size = CGSize.Empty;
+            size.Height = View.Frame.Height;
+            PreferredContentSize = size;
```
willsb: If width is ignored, why use a `CGSize`? 🤔

Toggl.Daneel/ViewControllers/EditTimeEntryViewController.cs
```diff
@@ -17,6 +23,86 @@ public override void ViewDidLoad()
...
+            var durationConverter = new TimeSpanToDurationWithUnitValueConverter();
+            var dateConverter = new DateToTitleStringValueConverter();
+            var timeConverter = new DateTimeToTimeConverter();
+            var visibilityConverter = new MvxVisibilityValueConverter();
+            var inverterVisibilityConverter = new MvxInvertedVisibilityValueConverter();
+
+            var bindingSet = this.CreateBindingSet<EditTimeEntryViewController, EditTimeEntryViewModel>();
+
+            //Text
+            bindingSet.Bind(DescriptionLabel)
```
willsb: I'd make those with only `Bind` and `To` inline, since they aren't that big

Toggl.Foundation.MvvmCross/ViewModels/EditTimeEntryViewModel.cs
```diff
@@ -20,32 +20,36 @@ public class EditTimeEntryViewModel : MvxViewModel<ITimeEntry>
...
 
-        public DateTimeOffset? EndTime { get; set; }
+        private DateTimeOffset? endTime;
+        public DateTimeOffset? EndTime
+        {
+            get => endTime;
+            set
+            {
+                endTime = value;
+                if (endTime != null)
```
willsb: What happens if the user changes it to null back again?

Toggl.Daneel/ViewControllers/EditTimeEntryViewController.cs
arturs105: Because `CGSize` is a `struct` so `PrefferedContentSize.Height` can't be set. But I could change this to `PreferredContentSize = new CGSize { Height = height }`

Toggl.Foundation.MvvmCross/ViewModels/EditTimeEntryViewModel.cs
arturs105: Can it even be done? We could again subscribe to time service ticks.

Toggl.Daneel/ViewControllers/EditTimeEntryViewController.cs
willsb: What's the problem with setting the width too and simply setting it to `View.Frame`?

Toggl.Foundation.MvvmCross/ViewModels/EditTimeEntryViewModel.cs
```diff
@@ -8,44 +8,52 @@
...
+        public DateTimeOffset? EndTime
+        {
+            get => endTime;
+            set
+            {
+                endTime = value;
+                if (endTime == null)
+                    subscribeToTimeServiceTicks();
+                else
+                    tickingDisposable.Dispose();
```
willsb: What happens if you call dispose multiple times?

This would be better if written like:

```
if (endTime != null)
{
    tickingDisposable?.Dispose();
    tickingDisposable = null;
    return;
}

subscribeToTimeServiceTicks();
```

willsb: Also, add a value comparison on top, similar to what Fody weaves on all of our properties, to avoid unnecessary subscriptions, something like `if (endTime == value) return;`

Toggl.Ultrawave/packages.config
```diff
@@ -3,7 +3,7 @@
   <package id="modernhttpclient" version="2.4.2" targetFramework="portable45-net45+win8+wpa81" />
   <package id="Newtonsoft.Json" version="10.0.3" targetFramework="portable45-net45+win8+wpa81" />
   <package id="StyleCop.Analyzers" version="1.0.2" targetFramework="portable45-net45+win8+wpa81" developmentDependency="true" />
-  <package id="System.Reactive.Core" version="3.0.0" targetFramework="net45" />
-  <package id="System.Reactive.Interfaces" version="3.0.0" targetFramework="net45" />
-  <package id="System.Reactive.Linq" version="3.0.0" targetFramework="net45" />
+  <package id="System.Reactive.Core" version="3.0.0" targetFramework="net45" requireReinstallation="true" />
```
willsb: what are those?

Toggl.Foundation.MvvmCross/ViewModels/EditTimeEntryViewModel.cs
```diff
@@ -8,44 +8,58 @@
...
 using Toggl.Foundation.DataSources;
 using Toggl.Multivac;
-using Toggl.Multivac.Models;
+using Toggl.PrimeRadiant.Models;
 using static Toggl.Multivac.Extensions.ObservableExtensions;
 
 namespace Toggl.Foundation.MvvmCross.ViewModels
 {
-    public class EditTimeEntryViewModel : MvxViewModel<ITimeEntry>
+    public class EditTimeEntryViewModel : MvxViewModel<IDatabaseTimeEntry>
```
willsb: One thing I just noticed. It might be better to wrap the id around a parameter object (similar to what we're doing on the LoginView and the StartTimeEntryView) and re-fetch the `IDatabaseTimeEntry` here. Much faster and safer (no serialization involved and we're sure to get the entire object)

Toggl.Daneel/ViewControllers/EditTimeEntryViewController.cs
```diff
@@ -17,6 +24,85 @@ public override void ViewDidLoad()
...
+            bindingSet.Bind(StartTimeLabel)
+                      .To(vm => vm.StartTime)
+                      .WithConversion(timeConverter);
+            
+            //Commands
+            bindingSet.Bind(CloseButton).To(vm => vm.CloseCommand);
+            bindingSet.Bind(DeleteButton).To(vm => vm.DeleteCommand);
+
+            //Description visibility
+            bindingSet.Bind(AddDescriptionView)
```
willsb: Consider add one empty line between each one of the multiline binding descriptions. It makes scanning faster

Toggl.Daneel/ViewControllers/EditTimeEntryViewController.cs
```diff
@@ -17,6 +24,85 @@ public override void ViewDidLoad()
...
+            PreferredContentSize = View.Frame.Size;
+
+            resizeSwitch();
+
+            var durationConverter = new TimeSpanToDurationWithUnitValueConverter();
+            var dateConverter = new DateToTitleStringValueConverter();
+            var timeConverter = new DateTimeToTimeConverter();
+            var visibilityConverter = new MvxVisibilityValueConverter();
+            var inverterVisibilityConverter = new MvxInvertedVisibilityValueConverter();
+            var colorConverter = new MvxRGBValueConverter();
```
willsb: Does this work with strings too?

PR #376, opened 2017-08-14
Add color binding

arturs105:


willsb:
Why is this needed? You can simply use the included `MvxColorValueConverter` that's included with the plugin

arturs105:
I did not know about that. So how would it be used to bind a view's background colour?

willsb:
```
bindingSet.Bind(SomeView)
          .For(v => v.BackgroundColor)
          .To(vm => vm.SomeColor)
          .WithConvertion(colorConverter);
```

arturs105:
🤦‍♂️ I forgot that bindings can be added for regular properties.

PR #385, opened 2017-08-15
Linker safety

willsb:
This ensures that the all our assemblies are safe for linkage, once and for all 💪 

Instead of relying on the error prone `--linkskip` args, we'll instead annotate our libraries as [`LinkerSafe`](https://developer.xamarin.com/guides/ios/advanced_topics/linker/). This has the added benefit of making the libraries safe for everyone consuming them, not only Daneel. Since the amount of apps will grow, this is a nice plus.

_"Why can I still see one `--linkskip`?"_, you ask me. Realistically, it makes no sense not to skip linkage on the actual app projects; after all they are not meant to be consumed but anyone and anything being removed is just something that depends on reflection.

Since we are now sure that everything works, new code _has_ to ensure that things continue working. If anything breaks, it's solely the committer's fault, and no one else's, for not testing the code on the Simulator/Device.

This is best reviewed by looking at each commit individually. Please also do a full functional test on both device and Simulator before approving. I want a review from every dev in the team, just in case.


arturs105:
So how do we decide which new classes should have the `PreserveAttribute` and which shouldn't?

amulware:
That is a good question. There should be a guide/explanation of some sort.

willsb:
There's no "deciding". New stuff that's not statically referenced anywhere (aka things that rely on reflection) will explode during runtime given our linker settings. So we should use `Preserve` or else the new changes won't even work.

Obvious targets are models that need serializing and ViewModels, both of which rely heavily on reflection because of serialization/IoC

amulware:
What about the guide/explanation I mentioned? Will you add that here, separately, or do you object?

willsb:
I don't see the need for us to have documentation here. This is something Xamarin Devs in general should know and Xamarin maintains very up to date docs on this regard ([here](https://developer.xamarin.com/guides/ios/advanced_topics/linker/), [here](https://developer.xamarin.com/guides/ios/advanced_topics/native_interop/) and [here](https://developer.xamarin.com/guides/android/advanced_topics/linking/)) so there's nothing I can write any further.

I can, however, provide a document that explains our linking strategies for newcomers if you find it necessary. I'd prefer to do so in a different PR, though.

Toggl.Ultrawave/Models/Workspace.cs
```diff
@@ -1,9 +1,11 @@
 ﻿using System;
-﻿using Newtonsoft.Json;
+using Newtonsoft.Json;
+using Toggl.Multivac;
 using Toggl.Multivac.Models;
 
 namespace Toggl.Ultrawave.Models
 {
+    [Preserve(AllMembers = true)]
```
amulware: Why not generate these with the constructors?

amulware:
:shipit:

PR #387, opened 2017-08-15
UI Test login

willsb:
This adds UI tests for login, testing both success and failure.

Note that there now is a `Credentials.cs` file. Replace the constants there with valid credentials before running the tests so they work. 

In a future PR, UI tests will run from cake, making the whole process a lot simpler.

Squash whenever


willsb:
There are no additional conditions. It only has to be valid credentials. Any other thing is irrelevant for the time being. If it ever becomes relevant, it will be noted. 

Also, there is no validation on the validity of provided credentials; if you don't provide valid credentials, the tests simply fail.

Toggl.Daneel.Tests.UI/LoginTests.cs
```diff
@@ -51,5 +51,29 @@ public void TheBackButtonShowsTheEmailFieldIfThePasswordFieldIsVisible()
...
             app.Screenshot("Login email page.");
         }
+
+        [Test]
+        public void TheNextButtonAfterInputtingAnInvalidPasswordShowsTheErrorLabel()
+        {
+            app.EnterText(ValidEmail);
+            app.GoToPasswordScreen();
+
+            app.EnterText("123456");
```
amulware: I rather have this use the credentials password and append a character to ensure it is not a valid password.

Toggl.Daneel.Tests.UI/Credentials.cs
```diff
@@ -0,0 +1,9 @@
+using System;
```
arturs105: Unnecessary

Toggl.Daneel.Tests.UI/LoginTests.cs
```diff
@@ -51,5 +51,29 @@ public void TheBackButtonShowsTheEmailFieldIfThePasswordFieldIsVisible()
...
+            app.GoToPasswordScreen();
+
+            app.EnterText($"{Credentials.Password}123456");
+            app.TryLoginAndFail();
+
+            app.Screenshot("Login email page.");
+        }
+
+        [Test]
+        public void TheNextButtonAfterInputtingAnValidPasswordShowsTheMainScreen()
```
arturs105: It should be "TheNextButtonAfterInputting**A**ValidPasswordShowsTheMainScreen" instead of "TheNextButtonAfterInputting**An**ValidPasswordShowsTheMainScreen"

arturs105: Also this test fails, but I assume it's because login is broken right now.

simonrozsival:
Wouldn't it be better to create a new user account before the tests are run?

Toggl.Daneel.Tests.UI/LoginTests.cs
willsb: Did you replace your variables? If you did, this should work

arturs105: Yeah, I did replace them. Maybe it's because I haven't set up the cert for staging api.

PR #388, opened 2017-08-15
Run UI tests on cake

willsb:
To test the UI tests:

`bash build.sh —target Tests.UI`

I had to change the target build architectures to x64 only; the NUnit runner fails otherwise. That won’t be a problem anyways.

One now needs to expose `TOGGL_UI_TEST_USERNAME ` and `TOGGL_UI_TEST_PASSWORD ` as environment variables. Take this into account when configuring the Bitrise flow.

This should remain `on-hold` until #387 is merged


Toggl.Daneel.Tests.UI/Configuration.cs
```diff
@@ -0,0 +1,15 @@
...
+    public static class Configuration
+    {
+        public static iOSApp GetApp()
+            => ConfigureApp
+                .iOS
+                .AppBundle("../iPhoneSimulator/Debug/Toggl.Daneel.app")
+                .EnableLocalScreenshots()
+                .StartApp();
+    }
+}
```
arturs105: \n

willsb:
It's possible, yes. It requires [passing the simulator's device id](https://developer.xamarin.com/guides/testcloud/uitest/working-with/selecting-ios-simulator/offline.pdf) when creating the `iOSApp`

PR #391, opened 2017-08-16
Add example for private const

arturs105:


amulware:
I would require an additional review, if @willsb had not already implicitly approved these changes here: #392

willsb:
🚀

PR #394, opened 2017-08-16
Use a specific suggestion model

willsb:
We should use this, a specific model for suggestions, instead of returning raw time entries. This will give us more flexibility if we need to change how suggestions work.

Squash whenever


Toggl.Foundation/Suggestions/Suggestion.cs
```diff
@@ -0,0 +1,29 @@
...
+
+namespace Toggl.Foundation.Suggestions
+{
+    public sealed class Suggestion
+    {
+        public string Description { get; } = "";
+
+        public long? ProjectId { get; } = null;
+
+        public long? TaskId { get; } = null;
```
amulware: If we have a task id, shouldn't we have a task name as well?

willsb: Agreed

PR #399, opened 2017-08-17
Use time entry events in the log

willsb:
Now our lists react to new time entries being added without the need of restarting the application. YAY

Please do a functional test before approving this.


Toggl.Foundation.MvvmCross/ViewModels/TimeEntriesLogViewModel.cs
```diff
@@ -52,10 +55,55 @@ public TimeEntriesLogViewModel(ITogglDataSource dataSource, ITimeService timeSer
...
+            var timeEntriesInDay = new List<TimeEntryViewModel> { new TimeEntryViewModel(timeEntry) };
+
+            var collection = TimeEntries.FirstOrDefault(x => x.Date == indexDate);
+            if (collection != null)
+            {
+                timeEntriesInDay.AddRange(collection);
+                TimeEntries.Remove(collection);
+            }
+
+            var newCollection = new TimeEntryViewModelCollection(timeEntriesInDay.GroupBy(x => x.Start.Date).Single());
```
simonrozsival: This seems quite strange to me. I would prefer adding another constructor to the `TimeEntryViewModelCollection` which would take a `DateTime` and `IEnumerable`? It would make it much more readable.

Toggl.Foundation.MvvmCross/ViewModels/TimeEntriesLogViewModel.cs
```diff
@@ -52,10 +55,55 @@ public TimeEntriesLogViewModel(ITogglDataSource dataSource, ITimeService timeSer
...
+
+        private void onTimeEntryCreated(IDatabaseTimeEntry timeEntry)
+        {
+            var indexDate = timeEntry.Start.Date;
+            var timeEntriesInDay = new List<TimeEntryViewModel> { new TimeEntryViewModel(timeEntry) };
+
+            var collection = TimeEntries.FirstOrDefault(x => x.Date == indexDate);
+            if (collection != null)
+            {
+                timeEntriesInDay.AddRange(collection);
```
simonrozsival: The order of the groups by days is taken care of, but the order of time entries in a day is not? Or is the order of the time entries in a day handled somewhere else, so it makes sense to put the new time entry at the beginning of the list?

Toggl.Foundation.MvvmCross/ViewModels/TimeEntriesLogViewModel.cs
```diff
@@ -1,19 +1,24 @@
 ﻿using System.Collections.ObjectModel;
+using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Reactive.Linq;
+using System.Threading;
```
arturs105: Not needed

Toggl.Foundation.MvvmCross/ViewModels/TimeEntriesLogViewModel.cs
willsb: Right now, the only possible add event is starting, so this works. Good catch, though, this will eventually break

simonrozsival: You can have TEs which have start and stop in the future - so once we start pulling data from the API server, this could break.

PR #400, opened 2017-08-17
Add binding to billable button

willsb:
This adds the structure needed for changing the colors of the buttons once they get highlighted. Since right now only the billable flag can actually change colors, I’m only adding the bindings to it.

Squash whenever


Toggl.Daneel/Converters/BoolToUIColorConverter.cs
```diff
@@ -0,0 +1,23 @@
...
+
+namespace Toggl.Foundation.MvvmCross.Converters
+{
+    public class BoolToUIColorConverter : MvxValueConverter<bool, UIColor>
+    {
+        private readonly UIColor onColor;
+        private readonly UIColor offColor;
+
+        public BoolToUIColorConverter(MvxColor onColor, MvxColor offColor)
+        {
```
arturs105: What about null checks and tests?

willsb: This is in Daneel, so this will be covered by the UI tests. Null checks can be added, though

PR #401, opened 2017-08-18
Add more methods to the tags API client

simonrozsival:
The constructor in `Tag.cs` is temporary and should be removed after https://github.com/toggl/backend/issues/5182 is fixed.


Toggl.Ultrawave/Models/Tag.cs
```diff
@@ -12,5 +13,14 @@ public sealed partial class Tag : ITag
...
 
         public DateTimeOffset At { get; set; }
+
+        [JsonConstructor]
+        public Tag(long id, long workspaceId, string name, DateTimeOffset? at)
+        {
+            Id = id;
+            WorkspaceId = workspaceId;
+            Name = name;
+            At = at ?? new DateTimeOffset(DateTime.UtcNow);
```
amulware: This does not make me happy. Not sure what else to do about it right now.

simonrozsival: Me neither, the user's time settings could be wrong and there could be a problems with synchronization of tags later. But I don't think there is a better workaround for the null value at the moment.

amulware: If the user's time is wrong, all sorts of things can go wrong. We might want to use a time service in the future that syncs with our backend and ignores the local time if it can. But for now, let's assume users have the correct time.

PR #404, opened 2017-08-18
Add FetchAllSinceState

amulware:
This is the first sync state machine state implemented. It is also the first state that is run when starting a PULL sync.

It takes the since parameters from the database and kicks off a bunch of parallel API calls with those parameters.

The followup states will each take one of these observables (in order from workspaces to time entries) and persist the result to the database.

I'm not entirely happy with the verbosity of both the state's code and the tests, though the problem here mostly is that it is doing a lot of similar work, which is hard to separate. Any suggestions are welcome.


Toggl.Foundation/Sync/States/FetchAllSinceState.cs
```diff
@@ -0,0 +1,49 @@
...
+                connectedReplayed(getSinceOrDefault(sinceDates.TimeEntries, api.TimeEntries.GetAll, api.TimeEntries.GetAllSince))
+            );
+
+            observer.OnNext(FetchStarted.Transition(observables));
+            observer.OnCompleted();
+
+            return () => { };
+        });
+
+        private static IObservable<T> connectedReplayed<T>(IObservable<T> observable)
```
simonrozsival: Shouldn't this method be rather in the `FetchObservables` class? I would say that this class has responsibility of obtaining the since dates and deciding whether to call GetAll or GetAllSince, the other class holds the observables and thus should have the responsibility of processing them in the constructor.

amulware: I would prefer the other object to remain a simple data transfer object, and keep actual behaviour within the states.

Toggl.Foundation/Sync/States/FetchAllSinceState.cs
```diff
@@ -0,0 +1,49 @@
...
+        {
+            var replayed = observable.Replay();
+            replayed.Connect();
+            return replayed;
+        }
+
+        private static IObservable<T> getSinceOrDefault<T>(
+            DateTimeOffset? threshold, Func<IObservable<T>> @default, Func<DateTimeOffset, IObservable<T>> since)
+            => threshold.HasValue ? since(threshold.Value) : @default();
```
amulware: The main reason is that using the ternary operator above makes the code a lot longer and more difficult to check for correctness there:

```
connectedReplayed(sinceDates.Clients.HasValue ? api.Clients.GetAllSince(sinceDates.Clients.Value) : api.Clients.GetAll()),
```

If the general consensus is to is to inline this, we can do so, but I think using the helper function makes this clearer, and avoids some repetition that could cause mistakes.

Toggl.Foundation/Sync/States/FetchAllSinceState.cs
amulware: We could consider making this an extension method btw since it is quite a useful bit of logic (I use the same in the StateMachineManager I believe), but that should maybe be done in a separate PR unless you insist.

simonrozsival: If other states would use it, then I would go with reusing it somehow. Extension method seems like a better option than creating some base class.

amulware: I'm not sure if this will be used in any other states. It is only necessary here because we are running several requests at the same time, which is something I want to avoid as much as possible. But we'll have to see.

amulware: I can still make it an extension method though and will do so in a separate PR.

Toggl.Foundation/Sync/States/FetchAllSinceState.cs
simonrozsival: I don't think that the length of the code bothers us. But I take your point with repetitive code. Should you keep the helper function, then I would suggest at changing the order of the parameters of the function to correspond to the order in the ternary operator. I would also prefer renaming `@default` to something different, like `all`, which doesn't need the ugly `@` symbol in its name.

PR #407, opened 2017-08-21
Storing tags in Realm

### LINKED ISSUE #167: Fody can't IList Realm
When building I get:
```
 Toggl.PrimeRadiant -> bin/Release/Toggl.PrimeRadiant.dll
    Fody: Fody (version 2.0.6.0) Executing
Toggl.PrimeRadiant.Realm/Models/RealmTimeEntry.cs(26,37,26,41): warning : Fody/RealmWeaver: SKIPPING RealmTimeEntry.Tags because it is an IList but its generic type is not a RealmObject subclass, so will not persist. [Toggl.PrimeRadiant.Realm/Toggl.PrimeRadiant.Realm.csproj]
Toggl.PrimeRadiant.Realm/Models/RealmTimeEntry.cs(28,36,28,40): warning : Fody/RealmWeaver: SKIPPING RealmTimeEntry.TagIds because it is an IList but its generic type is not a RealmObject subclass, so will not persist. [Toggl.PrimeRadiant.Realm/Toggl.PrimeRadiant.Realm.csproj]
```

No good.

### LINKED ISSUE #380: Figure out how to store tags
Ref #167

As far as I know (test this) tags can be created by pushing their names when pushing/putting time entries.
The tag ids do not need to be pushed, and backend should return them (new ids for new tags, existing ids for existing tags (test this!)).

There is also an endpoint to fetch a list of all tags. As far as I can tell, these can be used for auto completion, but serve no other purpose.

Regarding storage of tags in PR, I think it would be most efficient to store them in a string, separated by some magic character (which? check with backend if there is a character that cannot occur in tag names, or find some fool proof way to escape). I don't think there is ever any reason to store tag ids on a time entry, since backend does matching on names anyways when we push/put (test/check this as well).

When working on this, consider all the scenarios that can happen with respect to tags.

Ideally, submit findings and proposal before implementation.

### LINKED ISSUE #477: Try removing tag names from time entry models
The API should soon support handling tags entirely by Id just as any other entity.

I see no reason to keep the tag names (a list of strings) in each time entry model.

If there really is none, this should be scrapped.

simonrozsival:
Closes #380, closes #477, and also closes #167.

We cannot just store tag names concatenated into one string, because tag names can be edited in the webapp and so we need to store them in the database separately and reference them from the time entries.

Tag names list was removed, we don't have to use it anymore because of the changed API endpoint for creating time entries.


Toggl.PrimeRadiant.Realm/Models/RealmConstructors.tt
```diff
@@ -53,8 +55,10 @@ namespace Toggl.PrimeRadiant.Realm
             var isProperty = line.Contains("get;");
             if (!isProperty) continue;
 
+            var isEnumerable = trimmedLine.StartsWith("IEnumerable<");
```
willsb: Couldn't you check if the property has a setter on line 55 instead?

simonrozsival: That would be great, but currently none of the Multivac model interfaces has any setters defined.

willsb: Then 🤷‍♂️

simonrozsival: We should definitely do something about it later when we work on #409

amulware:
Putting on hold until we do tags.

Toggl.PrimeRadiant.Realm/Models/RealmConstructors.cs
```diff
@@ -208,8 +208,10 @@ public void SetPropertiesFrom(IDatabaseTimeEntry entity, Realms.Realm realm)
             Start = entity.Start;
             Stop = entity.Stop;
             Description = entity.Description;
-            TagNames = entity.TagNames;
-            TagIds = entity.TagIds;
+            RealmTags.Clear();
+            var allRealmTags = entity.TagIds.Select(id => realm.All<RealmTag>().Single(x => x.Id == id));
```
simonrozsival: So far I didn't find a better way how to do this. I count on Realm being super fast. Suggestions are welcome.

willsb: I think you're looking for [backlink/inverse relationships](https://realm.io/docs/xamarin/latest/#inverse-relationships`). Let me know if these suffice your needs, otherwise we can take a look at this together :)

willsb:
Can you investigate whether this tackles #167 as well and add it to the list of closed issues?

simonrozsival:
These warnings go away. So this also closes #167.

Toggl.PrimeRadiant.Realm/Models/RealmConstructors.cs
simonrozsival: I don't think I could use backlinks - this is a m:n relationship and the `Tag` model does not have any reference to the time entries it belongs to. Even if there was some direct connection Realm would know about, I want to load all the tags, which have `Id` in `entity.TagIds`. That is something very easy to do in SQL, but I'm not sure how to do it in Realm. The expression in `Where` can be only very limited (I tried e.g. `realm.All<RealmTag>().Where(x => entity.TagIds.Contains(x.Id))` but that doesn't work.

On the other hand - what is the maximum number of tags somebody used for a single time entry - 20? And how many tags will be in the database - 1000? That should be quite fast, definitely faster than querying suggestions, and that seems to be really fast.

willsb: Just pointing out that Realm's `IQueryable`s can't handle accessing stuff that's not in the DB (this included calculated properties, since those are not stored), so in order to have accessing `TagIds` working, you'd have to first enumerate everything and do it in memory.

No, as you said, this should be fast enough

willsb: I just realized that this should be `Where` instead of `Single`

simonrozsival: No, it should be Single. Why did you think it should be Where? I suppose we won't have multiple tags with the same id.

arturs105: Why not just `realm.All<RealmTag>().Where(t => TagIds.Contains(t.Id));`?

PR #408, opened 2017-08-21
Make sure the edit view can handle long projects/tasks/clients

arturs105:
This adds correct wrapping behaviour for projects/tasks/clients in the edit view:

- Projects/tasks/clients wrap by words, so all 3 are displayed in a single `UILabel`
- The coloured project dot must be aligned with the first line of the text, so it's added as a `NSTextAttachemt`


Toggl.Daneel/Converters/ProjectTaskClientToAttributedStringValueConverter.cs
```diff
@@ -0,0 +1,89 @@
...
+            if (!string.IsNullOrEmpty(value.project))
+                builder.Append($" { value.project }");
+            
+            if (!string.IsNullOrEmpty(value.task))
+                builder.Append($": { value.task }");
+            
+            if (!string.IsNullOrEmpty(value.client))
+                builder.Append($" { value.client }");
+
+
```
arturs105: This one is on purpose to separate the code visually. Lines 32 - 41 do one thing and lines 44 - 52 do another and I think two empty lines look better.

simonrozsival: Then you make a separate private function which joins the names and returns a string.

simonrozsival: Then you make a separate private function which joins the names and returns a string.

arturs105: Okay, a few more refactors coming in then.

arturs105:
Yeah, I had a feeling that this wouldn't do. And I didn't know about the value combiners. Will look into them.

willsb:
Is this `wip` or `on-hold`?

Toggl.Daneel/ProjectTaskClientValueCombiner.cs
```diff
@@ -0,0 +1,93 @@
...
+
+            var project = steps.ElementAt(0).GetValue().ToString();
+            var task = steps.ElementAt(1).GetValue()?.ToString();
+            var client = steps.ElementAt(2).GetValue()?.ToString();
+            var fontHeight = (double)steps.ElementAt(3).GetValue();
+            var projectColor = steps.ElementAt(4).GetValue().ToString();
+
+            var text = buildString(project, task, client);
+
+            var image = UIImage.FromBundle("icProjectDot").ImageWithRenderingMode(UIImageRenderingMode.AlwaysTemplate);
```
willsb: No me gusta this literal string

Toggl.Daneel/ProjectTaskClientValueCombiner.cs
```diff
@@ -0,0 +1,93 @@
...
+        {
+            var imageSize = dot.Image.Size;
+            var y = (fontHeight - imageSize.Height) / 2;
+            dot.Bounds = new CGRect(0, y, imageSize.Width, imageSize.Height);
+        }
+
+        private void setProjectDotColor(NSMutableAttributedString text, string hexColor)
+        {
+            var range = new NSRange(0, 1);
+            var color = (UIColor)colorConverter.Convert(hexColor, typeof(MvxColor), null, CultureInfo.CurrentCulture);
```
willsb: That looks like a bit of an overkill, but I can't think of a better way

Toggl.Daneel/ViewControllers/EditTimeEntryViewController.cs
```diff
@@ -34,19 +35,19 @@ public override void ViewDidLoad()
...
             bindingSet.Bind(DescriptionTextField).To(vm => vm.Description);
-            bindingSet.Bind(ProjectLabel).To(vm => vm.Project);
-            bindingSet.Bind(ClientLabel).To(vm => vm.Client);
             bindingSet.Bind(BillableSwitch).To(vm => vm.Billable);
             bindingSet.Bind(DurationLabel)
                       .To(vm => vm.Duration)
                       .WithConversion(durationConverter);
             
+            bindingSet.Bind(ProjectTaskClientLabel, $"AttributedText ProjectTaskClient(Project, Task, Client, {ProjectTaskClientLabel.Font.CapHeight}, ProjectColor)")
+                      .For(v => v.AttributedText);
```
willsb: Is the `.For` even needed?

Toggl.Daneel/ProjectTaskClientValueCombiner.cs
```diff
@@ -0,0 +1,93 @@
...
+            var task = steps.ElementAt(1).GetValue()?.ToString();
+            var client = steps.ElementAt(2).GetValue()?.ToString();
+            var fontHeight = (double)steps.ElementAt(3).GetValue();
+            var projectColor = steps.ElementAt(4).GetValue().ToString();
+
+            var text = buildString(project, task, client);
+
+            var image = UIImage.FromBundle("icProjectDot").ImageWithRenderingMode(UIImageRenderingMode.AlwaysTemplate);
+            var dot = new NSTextAttachment { Image = image };
+            //There neeeds to be a space before the dot, otherwise the colors don't work
```
willsb: Hackity hackity hack, is this a known bug in UILabel?

Toggl.Daneel/ProjectTaskClientValueCombiner.cs
arturs105: Any suggestions?

Toggl.Daneel/ProjectTaskClientValueCombiner.cs
arturs105: Yes, found this solution in StackOverflow

Toggl.Daneel/ViewControllers/EditTimeEntryViewController.cs
arturs105: Weirdly, yes. If i remove it, the value is set to `Text` instead of `AttributedText`. If i remove the target property from the binding string, it also doesn't work. So in the end, the target property needs to be set in two ways for this to work.

Toggl.Daneel/ProjectTaskClientValueCombiner.cs
arturs105: https://stackoverflow.com/questions/29041458/how-to-set-color-of-templated-image-in-nstextattachment

Toggl.Daneel/ProjectTaskClientValueCombiner.cs
willsb: Make it a const inside the combiner, at least, given we don't have any means of creating strongly typed images

PR #410, opened 2017-08-21
Enable more Stylecop spacing rules

amulware:
I enabled pretty much all of them, except for those that caused a huge amount of errors (SA1004, SA1005, SA1028) and those that conflict with how we do things (SA1008, SA1009).

Most of the others either had no or very few violations.

The only things I can see as being controversial are:
- not allowing spaces around the expression inside interpolated strings: `$"{this} instead of { this }"`
- not allowing a space after `new` for array initialisers: `new[] { this }` instead of `new [] { this}`

But I think there is great value in consistency.

Squash if you like.


willsb:
I'm fine with giving up the string interpolation spacing in exchange for consistency 👍

PR #412, opened 2017-08-21
Add bindings to SuggestionViewCell

willsb:
No more placeholders.

Squash whenever


Toggl.Daneel/Converters/BoolToConstantValueConverter.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System;
+using System.Globalization;
+using MvvmCross.Platform.Converters;
+
+namespace Toggl.Daneel.Converters
+{
+    public sealed class BoolToConstantValueConverter : MvxValueConverter<bool, nfloat>
```
arturs105: What about making this generic `BoolToConstantValueConverter<T> : MvxValueConverter<bool, T>`? That way we wouldn't need any more bool to something converters and `BoolToUIColorConverter` could also be removed.

Toggl.Daneel/ViewSources/SuggestionsTableViewSource.cs
```diff
@@ -54,5 +69,7 @@ protected override UITableViewCell GetOrCreateCellFor(UITableView tableView, NSI
             => tableView.DequeueReusableCell(item == null ? emptyCellIdentifier : cellIdentifier, indexPath);
 
         public override nfloat GetHeightForRow(UITableView tableView, NSIndexPath indexPath) => 64;
+
+        public override nfloat GetHeightForHeader(UITableView tableView, nint section) => 54;
```
arturs105: I think it's 45 in Zeplin.

Toggl.Daneel/Converters/BoolToConstantValueConverter.cs
willsb: Nice idea 👌

Toggl.Daneel/ViewSources/SuggestionsTableViewSource.cs
willsb: Correct

PR #398, opened 2017-08-17
Workspace features Realm models

simonrozsival:
Closes #397 

This PR includes only models, there is no repository in `Database`, as none is needed.

I also believe the models should not implement `IDatabaseSyncable`. The features collection is read-only and it should change only when new data is fetched from the API server.

Issue #409, opened 2017-08-21
Fix RealmConstructors.tt and other generators


amulware:
The generator generates code that does not compile at the moment.

- [x] _RealmConstructors.tt_ assumes that everything ending with `Id` is a reference to a different entity and it will try to load that entity. This fails for `IWorkspaceFeature` and the enum `WorkspaceFeatureId`. The easiest solution would be to remove the `Id` suffix from this specific property.
- [x] _RealmConstructors.tt_ assumes that every model should have an `Id` property
- [x] _RealmConstructors.tt_ assumes that every Multivac model has corresponding Prime Radiant `IDatabase...` interface
- [x] _RealmConstructors.tt_ assumes that every property of the Multivac model interface will have also a setter defined in the concrete class. This fails in some cases - especially for lists.
- [x] _FoundationModels.tt_ does not have using for `Toggl.Multivace.Models`. This is a problem for entities which contain list of other entities (`IWorkspaceFeature` - `IWorkspaceFeatureCollection`)

amulware:
Please see if any of this still applies.

simonrozsival:
All of these sub-issues were already resolved and so I close this issue.

PR #421, opened 2017-08-22
Prime Radiant Workspace models

simonrozsival:
This PR contains only definitions of the models to reduce the number of changed files in PR #398.


Toggl.PrimeRadiant/Models/IDatabaseWorkspaceFeature.cs
```diff
@@ -0,0 +1,6 @@
+﻿using Toggl.Multivac.Models;
+
+namespace Toggl.PrimeRadiant.Models
+{
+    public interface IDatabaseWorkspaceFeature : IWorkspaceFeature, IDatabaseSyncable { }
```
simonrozsival: You are right, it is not needed, I will remove it.

simonrozsival: The only problem with that is that the generator expects every IDatabase.. to implement it, so the RealmConstructors must be edited to generate code which would compile.

amulware: Or we have to not generate the code for these types. I'm not sure what the best solution is right now.

simonrozsival:
This PR contains just the models. The repository is in PR #398 - I split it in two to make it easier for reviewing.

The repository, which is included in `ITogglDatabase` exposes `IObservable<IDatabaseWorkspaceFeatureCollection> GetByWorkspaceId(long id)`. The `RealmWorkspaceFeatureCollection` exposes `bool IsEnabled(WorkspaceFeatureId feature)`.

amulware:
Let's put this on hold until needed.

simonrozsival:
This must be closed after #409 so the generator works correctly (now one line of the `FoundationConstructors.cs` file is manually edited).

simonrozsival:
This is now ready for a review.

PR #422, opened 2017-08-22
Use env variable instead of commit hash

### LINKED ISSUE #396: Use environmental variables for user agent in API integration tests
This would allow us to simplify the cake script, and allow us to easily run individual tests from within the IDE.

simonrozsival:
Closes #396 

Configuration of integration tests in Bitrise should be updated accordingly when this is merged.

Merge whenever.


willsb:
I'm afraid this is not what was proposed by that issue. The idea was that `Toggl.Ultrawave.Tests.Integration` would fetch the variable during runtime in order to simplify the cake script.

simonrozsival:
You are right, I didn't read the whole thread thoroughly. I will try again :D

PR #424, opened 2017-08-22
Fix several issues of RealmConstructors.tt

simonrozsival:
This addresses the first three issues described in #409 

Squash whenever


Toggl.PrimeRadiant.Realm/Models/RealmConstructors.tt
```diff
@@ -27,33 +27,54 @@ namespace Toggl.PrimeRadiant.Realm
...
+                    WriteLine( "        }\n");
+                }
+
+                WriteLine($"        public {realmClassName}({interfaceName} entity, Realms.Realm realm)");
+                WriteLine( "        {");
+
+                if (extendsIBaseModel)
+                {
+                    WriteLine("            Id = entity.Id;");
+                    continue;
```
willsb: Is the continue really meant to be inside this if and not on the outer if block?

PR #425, opened 2017-08-22
Add starter credentials to integration tests

arturs105:
This adds the option to use starter features in integration tests, if these environment variables are set:
- `TOGGL_INTEGRATION_TEST_STARTER_USERNAME`
- `TOGGL_INTEGRATION_TEST_STARTER_PASSWORD`

Free and premium credentials could also be added, but they are not needed for now and can be added later.


Toggl.Ultrawave.Tests.Integration/Helper/Credentials.cs
```diff
@@ -0,0 +1,11 @@
+namespace Toggl.Ultrawave.Tests.Integration.Helper
```
simonrozsival: I am not sure if this should be in integration tests. I believe starter features will be needed more in UI tests rather than in integration tests.

Toggl.Ultrawave.Tests.Integration/Helper/Credentials.cs
```diff
@@ -0,0 +1,11 @@
...
+{
+    internal static class Credentials
```
simonrozsival: I would prefer a more descriptive name of the class - something like `TestUserCredentials`. We could then avoid the confusion between this class and `Toggl.Ultrawave.Network.Credentials` and get rid of the aliases.

arturs105: This was specifically added to test `TasksApi`, which needs access to starter features.

Toggl.Ultrawave.Tests.Integration/Helper/Credentials.cs
arturs105: This was specifically added to test `TasksApi`, which needs access to starter features.

simonrozsival: OK then. Somebody will dig into it when it is actually needed in other projects.

amulware:
Every single test currently creates a new user to ensure a predictable environment. How can we still do that using these fixed accounts? In general, how will these accounts be used?

arturs105:
We need to somehow test premium/starter features and for that we need accounts with premium/starter workspaces. And since the tests are run in staging, the environment will be reset every now and then.

But I see the problem with predictable environment. Maybe each test could set up (and maybe tear down) the environment for the test - create whatever is needed (project, workspace, time entry etc.)

amulware:
So we would have to write tests differently for paid features. Not a big fan of that. But maybe we can in the future find a way of unifying things to work for everything. Alternatively, we need a way to generate test accounts with paid features enabled in staging, which I would personally prefer over this hard coded stuff anyway. I'll look into the latter at some point.

Toggl.Ultrawave.Tests.Integration/Helper/TestUserCredentials.cs
```diff
@@ -0,0 +1,11 @@
+namespace Toggl.Ultrawave.Tests.Integration.Helper
+{
+    internal static class TestUserCredentials
+    {
+        public static class Starter
```
amulware: I do not want our code to refer to packages such as 'starter'. What matters for us are workspace features, so we should refer to things in a way that corresponds with that.

In this case maybe something like `TestUserCredentials.WithProFeature` might work.

PR #426, opened 2017-08-23
Change keyboard type for starting a time entry

arturs105:
This keyboard type is meant for easier access to "@" and "#" characters.


willsb:
Can't this be changed via the interface builder?

Issue #314, opened 2017-07-27
Improve test coverage


amulware:
I ran dotCover, and here are some things that we could improve:

Not tested:
- Either.Match (both overloads)
- Test.Adapter - no test for working case?
- HttpExtensions.AddRange
- ApiConfiguration

Also, models and exceptions are not tested anywhere, but this might be alright.

amulware:
Update: this list needs an update first.

heytherewill:
🔥

PR #433, opened 2017-08-23
Add query methods to the DataSources

willsb:
Needed for the suggestions to work.

Squash whenever


Toggl.Foundation/DataSources/TimeEntriesDataSource.cs
```diff
@@ -45,6 +45,9 @@ public TimeEntriesDataSource(IIdProvider idProvider, IRepository<IDatabaseTimeEn
         public IObservable<IEnumerable<IDatabaseTimeEntry>> GetAll()
             => repository.GetAll(te => !te.IsDeleted);
 
+        public IObservable<IEnumerable<IDatabaseTimeEntry>> GetAll(Func<IDatabaseTimeEntry, bool> predicate)
```
amulware: Should this not be an expression to make use of IQueryable?

Toggl.Foundation/DataSources/Interfaces/IProjectsSource.cs
```diff
@@ -6,7 +6,8 @@ namespace Toggl.Foundation.DataSources
 {
     public interface IProjectsSource
     {
-        IObservable<IEnumerable<IDatabaseProject>> GetAll();
         IObservable<IDatabaseProject> GetById(long id);
+        IObservable<IEnumerable<IDatabaseProject>> GetAll();
+        IObservable<IEnumerable<IDatabaseProject>> GetAll(Func<IDatabaseProject, bool> predicate);
```
amulware: Should there not be more tests for both of these methods to ensure they actually return all items that they should?

Toggl.Foundation/DataSources/TimeEntriesDataSource.cs
willsb: As discussed internally and because of [this doc](https://github.com/realm/realm-dotnet/blob/0fa97f6128b45168fea03d83691f6e3be3515d79/Docs/linqsupport.md#predicate-operations), we should stick with funcs for a while

Toggl.Foundation/DataSources/Interfaces/IProjectsSource.cs
willsb: I think we didn't add those because the are just returning from repository, which is tested. But I agree. I'll add that to #314

Issue #436, opened 2017-08-23
Add batch update/insert/delete with conflict resolution to database repositories


amulware:
Similar to #434, however, the conflict resolution behaviour should be injected from Foundation.

I see this working something like this:

```
parameters:
IEnumerable<IDatabaseModel> items
Func<IDatabaseModel, Expression<Func<IDatabaseModel, bool>> getRealmItemExpressionFor
Func<IDatabaseModel, IDatabaseModel, ResolveMode> howToResolve

using(new transaction)
    foreach(item in items)
    {
        var realmItem = repo.Get(getRealmItemExpressionFor(item)) // maybe this can just be done by id instead if that's all we ever want to do here?
        var resolveMode = howToResolve(item, realmItem)
        if (resolveMode == TakeNew)
            // somehow write all values from item to realmItem, without creating a new item?! (could generate a method to set all fields)
        else if (resolveMode == TakeOld)
            // do nothing
    }
```

Very pseudo. Much incomplete.

Consider the real use cases, and that we have to handle at least: overriding data, creating new entity, deleting existing entity

We also have to make sure that any references inside the updated model are persisted correctly, which might mean we have to generate more code to update a RealmModel from an IDatabaseModel while resolving references.

simonrozsival:
I'm closing this manually, because it wasn't referenced well in the description of the pull request #467.

PR #443, opened 2017-08-24
Batch update of entities

simonrozsival:
This PR is related issue #436 but it does not close it. The other part of the issue - conflict resolution - will be part of a separate follow-up PR.

The core of the implementation is in the `RealmAdapter` class and it is intended to be used through the `Repository` class. Conflict resolution as well as matching Realm objects to the entities returned from the API server are injected through the function's arguments.

All of the entities are processed in one transaction and each of them is handled in an appropriate way. I consider 4 conflict resolutions so far:

- create a new record in the database
- update the local entity with the received data
- delete the record stored in the local database.
- ignore the received data

> We also have to make sure that any references inside the updated model are persisted correctly, which might mean we have to generate more code to update a RealmModel from an IDatabaseModel while resolving references.

The consistency of the references is ensured in the cases of creating a new record and updating a new one through the `SetPropertiesFrom` function which is called directly when updating an old record and also called internally by the constructors of all Realm models when a new record is stored in the database.


Toggl.PrimeRadiant.Realm/RealmAdapter.cs
```diff
@@ -15,7 +16,9 @@ internal interface IRealmAdapter<TModel>
 
         TModel Create(TModel entity);
 
-        TModel Update(TModel entity);   
+        TModel Update(TModel entity);
+
+        IEnumerable<TModel> BatchUpdate(IEnumerable<TModel> entities, Func<TModel, TModel, ConflictResolutionMode> conflictResolution);
```
amulware: The interface looks like it should support all cases we have right now. :+1:

Toggl.PrimeRadiant.Realm/RealmAdapter.cs
```diff
@@ -68,5 +103,40 @@ private void doTransaction(TModel entity, Action<Realms.Realm, TRealmEntity> tra
...
+            if (!resolvers.TryGetValue(resolveMode, out (Func<Realms.Realm, TRealmEntity, TRealmEntity> process, bool useNewEntity) resolver)) {
+                throw new IndexOutOfRangeException($"Unknown conflict resolution mode {resolveMode}");
+            }
+
+            var chosenEntity = resolver.useNewEntity ? convertToRealm(entity, realm) : old;
+            return resolver.process(realm, chosenEntity);
+        }
+
+        private TRealmEntity getSingleEntityMatching(Realms.Realm realm, TModel entity)
+            => realm.All<TRealmEntity>().Single(x => matchEntity(x, entity));
```
amulware: Should this not be SingleOrDefault()?

Toggl.PrimeRadiant.Realm/RealmAdapter.cs
```diff
@@ -24,24 +27,56 @@ internal sealed class RealmAdapter<TRealmEntity, TModel> : IRealmAdapter<TModel>
...
+            {
+                foreach (var entity in entities)
+                {
+                    var oldEntity = getSingleEntityMatching(realm, entity);
+                    var resolveMode = conflictResolution(oldEntity, entity);
+                    var resolvedEntity = resolveEntity(realm, oldEntity, entity, resolveMode);
+
+                    if (resolvedEntity != null)
+                    {
+                        resolvedEntities.Add(resolvedEntity);
```
amulware: Why have this if `resolveEntity` can do this already on its own? Especially since deletion will have to be in `resolveEntity` like this, so having the add here is strange.

Toggl.PrimeRadiant.Realm/RealmAdapter.cs
```diff
@@ -68,5 +103,40 @@ private void doTransaction(TModel entity, Action<Realms.Realm, TRealmEntity> tra
...
+                { ConflictResolutionMode.Update, (update, true) },
+                { ConflictResolutionMode.Delete, (delete, false) },
+                { ConflictResolutionMode.Ignore, (ignore, false) }
+            };
+
+        private static TRealmEntity create(Realms.Realm realm, TRealmEntity model)
+            => realm.Add(model);
+
+        private static TRealmEntity update(Realms.Realm realm, TRealmEntity model)
+            => realm.Add(model, true);
```
amulware: This seems broken btw. Without a primary key, this should fail, right?

Toggl.PrimeRadiant.Realm/RealmAdapter.cs
```diff
@@ -68,5 +103,40 @@ private void doTransaction(TModel entity, Action<Realms.Realm, TRealmEntity> tra
...
+        private static TRealmEntity ignore(Realms.Realm realm, TRealmEntity model) => model;
+
+        private TRealmEntity resolveEntity(Realms.Realm realm, TRealmEntity old, TModel entity, ConflictResolutionMode resolveMode)
+        {
+            if (!resolvers.TryGetValue(resolveMode, out (Func<Realms.Realm, TRealmEntity, TRealmEntity> process, bool useNewEntity) resolver)) {
+                throw new IndexOutOfRangeException($"Unknown conflict resolution mode {resolveMode}");
+            }
+
+            var chosenEntity = resolver.useNewEntity ? convertToRealm(entity, realm) : old;
+            return resolver.process(realm, chosenEntity);
```
amulware: I think these few lines are not going to work like this. Or if they do, it is not clear to me at all why. Would it maybe be simpler to use a switch-case here instead of the dictionary above?

Toggl.PrimeRadiant.Realm/RealmAdapter.cs
simonrozsival: You are right, if the entity is not in the database, than this would crash.

Toggl.PrimeRadiant.Realm/RealmAdapter.cs
simonrozsival: I had a switch at first which did exactly the same and rewrote it to make it more compact but I think I will turn this back into the switch to make it more readable.

Toggl.PrimeRadiant.Realm/RealmAdapter.cs
simonrozsival: I considered that and decided it would be better to keep the list only in the `BatchUpdate` function and use return value to create the collection of the resolved entities. The deletion might seem strange, I understand that.

Toggl.PrimeRadiant.Realm/RealmAdapter.cs
```diff
@@ -48,6 +51,35 @@ public TModel Update(TModel entity)
...
+            Ensure.Argument.IsNotNull(conflictResolution, nameof(conflictResolution));
+
+            var resolvedEntities = new List<TRealmEntity>();
+
+            var realm = Realms.Realm.GetInstance();
+            using (var transaction = realm.BeginWrite())
+            {
+                foreach (var entity in entities)
+                {
+                    var oldEntity = realm.All<TRealmEntity>().SingleOrDefault(x => matchEntity(x, entity));
```
willsb: Why not cache the call to `realm.All<TRealmEntity>()`?

Toggl.PrimeRadiant.Realm/RealmAdapter.cs
```diff
@@ -48,6 +51,35 @@ public TModel Update(TModel entity)
             return doModyfingTransaction(entity, (realm, realmEntity) => realmEntity.SetPropertiesFrom(entity, realm));
         }
 
+        public IEnumerable<TModel> BatchUpdate(IEnumerable<TModel> entities, Func<TModel, TModel, bool> matchEntity, Func<TModel, TModel, ConflictResolutionMode> conflictResolution)
```
willsb: This will be mostly comparison by id, right? I think this will greatly benefit from being an `Expression<Func>` rather than a `Func`

Toggl.PrimeRadiant.Realm/RealmAdapter.cs
```diff
@@ -73,5 +105,28 @@ private static TModel doTransaction(Func<Realms.Realm, TModel> transact)
...
+
+                case ConflictResolutionMode.Update:
+                    old.SetPropertiesFrom(entity, realm);
+                    return old;
+
+                case ConflictResolutionMode.Ignore:
+                    return old;
+
+                default:
+                    throw new IndexOutOfRangeException($"Unknown conflict resolution mode {resolveMode}");
```
willsb: This should be a `DatabaseException`, shouldn't it?

Toggl.PrimeRadiant.Realm/RealmAdapter.cs
amulware: I think I misunderstood what this was doing before. I wonder if it will ever be useful to return the whole list... I mean, if someone wants to observe what happens, they can inject that in the conflict resolver, right? Or am I overthinking this? Or is there a good reason to have this? Actually... might be good to pass a list of changed entities or their ids to the UI or something.. maybe it's fine.

Toggl.PrimeRadiant.Realm/RealmAdapter.cs
amulware: @willsb Will that work though since it's on TModel, not on TRealmEntity?

Toggl.PrimeRadiant.Realm/RealmAdapter.cs
amulware: I doubt it would be a significant performance impact, but I'm fine with changing it also.

Toggl.PrimeRadiant.Realm/RealmAdapter.cs
amulware: Should it though? There is nothing wrong with the database here, but with the injected conflict resolver.

Toggl.PrimeRadiant.Realm/RealmAdapter.cs
willsb: In principle it will as long as it follows what's in [this doc](https://github.com/realm/realm-dotnet/blob/0fa97f6128b45168fea03d83691f6e3be3515d79/Docs/linqsupport.md#predicate-operations). AKA: As long as it does not access computed properties or the complex properties of `IDatabaseModel`.

amulware: Even if it is through an interface? Well.. maybe. :D

Toggl.PrimeRadiant.Realm/RealmAdapter.cs
simonrozsival: I tried to mimic the behavior of `Update` (returns the updated entity), `Create` (returns the new entity), and `Delete` (does not return anything). Maybe we can remove it and return only if we needed it for some specific reason.

Toggl.PrimeRadiant.Realm/RealmAdapter.cs
simonrozsival: I admit the `IndexOutOfRangeException` should not be used here. But `DatabaseException` also seems weird, as @amulware pointed out. I could use `ArgumentException`, since the parameter of the function is invalid.

amulware: I'd be ok with that.

PR #457, opened 2017-08-25
Make realm not rely on primary keys

amulware:
The reason being that entity ids can change, so that's weird and finicky, so we rather not rely on them in any way. We can always index them for speed if we need to.

The way this was done is by giving all realm models a method that overrides all properties from a given model. That is then used for updating the model gotten from Realm which was previously only used to check if the model existed.

We could use the same method for creation as well, and remove the `convertToRealm` delegate from `RealmAdapter`, but that's more than we need right now.

Apart from removing the need for Ids being primary keys, this should also make it easier to create properly working batch updates for fetching, and the updates after POSTing (which changes id).


amulware:
Oh yeah, since I had to change it I slightly refactored and fixed the generator as well to not generate broken and unused models.

Toggl.PrimeRadiant.Realm/Models/RealmConstructors.tt
```diff
@@ -23,6 +21,15 @@ namespace Toggl.PrimeRadiant.Realm
...
         var realmClassName = $"Realm{className}";
+		
+        var databaseInterfaceName = $"IDatabase{className}";
+        var databaseInterfaceFileName = Path.Combine(filePath, "../../Toggl.PrimeRadiant/Models/", databaseInterfaceName + ".cs");
+        var databaseInterfaceExists = File.Exists(databaseInterfaceFileName);
+
+		if (!databaseInterfaceExists)
+			break;
+		
+		var updatesFromInterface = databaseInterfaceExists ? databaseInterfaceName : interfaceName;
```
simonrozsival: Tabs

Toggl.PrimeRadiant.Realm/Models/RealmConstructors.tt
```diff
@@ -34,38 +41,34 @@ namespace Toggl.PrimeRadiant.Realm
...
-                var databaseInterfaceFileName = Path.Combine(filePath, "../../Toggl.PrimeRadiant/Models/", databaseInterfaceName + ".cs");
-                var databaseInterfaceExists = File.Exists(databaseInterfaceFileName);
 
+                WriteLine($"        public {realmClassName}({updatesFromInterface} entity, Realms.Realm realm)");
+                WriteLine( "        {");
+				WriteLine( "            SetPropertiesFrom(entity, realm);");
+				WriteLine( "        }\n");
+					
+				WriteLine($"        public void SetPropertiesFrom({updatesFromInterface} entity, Realms.Realm realm)");
+				WriteLine( "        {");
```
simonrozsival: Tabs

Toggl.PrimeRadiant.Realm/Models/RealmConstructors.tt
```diff
@@ -34,38 +41,34 @@ namespace Toggl.PrimeRadiant.Realm
...
+				}
 
-                if (extendsIBaseModel)
-                {
-                    WriteLine("            Id = entity.Id;");
-                }
+				if (extendsIBaseModel)
+				{
+					WriteLine("            Id = entity.Id;");
+				}
```
simonrozsival: Tabs

Toggl.PrimeRadiant.Realm/Models/RealmConstructors.tt
```diff
@@ -34,38 +41,34 @@ namespace Toggl.PrimeRadiant.Realm
...
-                    WriteLine($"        public {realmClassName}({databaseInterfaceName} entity, Realms.Realm realm)");
-                    WriteLine($"            : this(entity as {interfaceName}, realm)");
-                    WriteLine( "        {");
                     WriteLine( "            IsDirty = entity.IsDirty;");
-                    WriteLine( "        }\n");
-                }
-
-                WriteLine($"        public {realmClassName}({interfaceName} entity, Realms.Realm realm)");
-                WriteLine( "        {");
+				}
```
simonrozsival: Tabs

Toggl.PrimeRadiant.Realm/Models/RealmConstructors.tt
```diff
@@ -23,6 +21,15 @@ namespace Toggl.PrimeRadiant.Realm
 
         var className = interfaceName.Substring(1, interfaceName.Length - 1);
         var realmClassName = $"Realm{className}";
+
```
simonrozsival: Tabs

Issue #441, opened 2017-08-24
Get first editting up and running


amulware:
Let's start small and only allow editing of descriptions of time entries. This should work with both running and no longer running entries.

- [x] Edit view opens from the log
- [x] Edit view opens from the running timer
- [x] User can edit the description
- [ ] Changes are persisted

PR #458, opened 2017-08-25
Make description editable

arturs105:
Related to #441 (doesn't completely close it yet)


Toggl.Foundation.MvvmCross/Helper/Color.cs
```diff
@@ -52,5 +52,13 @@ public static class StartTimeEntry
 
             public static readonly MvxColor SeparatorColor = new MvxColor(181, 188, 192);
         }
+
+        public static class EditTimeEntry
+        {
+            //This should be the same as pinkishGrey, but iOS makes the color a bit darker,
+            //when applied to UITextField.AttributedPlaceholder, so this is made a bit
+            //lighter than the actual color.
+            public static readonly MvxColor Placeholder = new MvxColor(215, 215, 215);
```
willsb: If you are including iOS specific remarks, why is this in the shared portion of the code?

Toggl.Daneel/ViewControllers/EditTimeEntryViewController.cs
```diff
@@ -30,6 +35,18 @@ public override void ViewDidLoad()
 
             resizeSwitch();
 
+            var placeholderAttributes = new UIStringAttributes
```
willsb: I'd move all this view manipulation related code away from ViewDidLoad and into its own method, like we are doing in every other VC (similar to what you did on the VC for `resizeSwitch`)

Toggl.Foundation.MvvmCross/Helper/Color.cs
arturs105: Forgot that this is shared. Will change

Toggl.Daneel/ViewControllers/EditTimeEntryViewController.cs
arturs105: 👍

amulware:
Please reference appropriate issues in the description.

Issue #444, opened 2017-08-24
Project auto completion in time entry creation dialog


amulware:
Bit of a complicated guy. Details here: #315

Issue #442, opened 2017-08-24
Description auto completion in time entry creation dialog


amulware:
- [x] Show previous time entries based on description search string (search by word in description, client, project, task)
- [x] Let user select one of these, filling in description and project from the selected one
- [x] User can still edit description/project after completion
- [x] Have all that persist correctly when starting the time entry

PR #460, opened 2017-08-25
Add start time entry suggestions

willsb:
Ref #442 and #444

Now you can get suggestions as you type! You still can’t select any of the suggestions, but this adds the barebones needed to make the following PR smaller and easier to review.

This queries the Database and does all transformations in a background thread, marshaling to the UI thread using the `MvxObservableCollection`.

Check #mobile-dev if you need a scripting to fetch real world data before testing this.

Tests are coming, but if you can review that in the meantime to speed things up, feel free to


Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntrySuggestions/SuggestionComparer.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace Toggl.Foundation.MvvmCross.ViewModels.StartTimeEntrySuggestions
+{
+    public sealed class SuggestionComparer : IEqualityComparer<ITimeEntrySuggestionViewModel>
+    {
+        public static SuggestionComparer Comparer { get; } = new SuggestionComparer();
```
arturs105: I think `Instance` would be a better name.

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntrySuggestions/TimeEntrySuggestionViewModel.cs
```diff
@@ -1,10 +1,16 @@
...
+﻿using System.Collections.Generic;
+using System.Linq;
+using MvvmCross.Core.ViewModels;
 using Toggl.PrimeRadiant.Models;
 
 namespace Toggl.Foundation.MvvmCross.ViewModels.StartTimeEntrySuggestions
 {
     public sealed class TimeEntrySuggestionViewModel : MvxNotifyPropertyChanged, ITimeEntrySuggestionViewModel
     {
+        internal static IEnumerable<TimeEntrySuggestionViewModel> FromTimeEntry(
```
arturs105: Should be in plural: `FromTimeEntries`.

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
```diff
@@ -69,6 +78,33 @@ public StartTimeEntryViewModel(ITogglDataSource dataSource, ITimeService timeSer
...
+            if (string.IsNullOrEmpty(TextFieldInfo.Text)) return;
+
+            var queryText = TextFieldInfo.Text;
+            var suggestionType = SuggestionType.TimeEntries;
+
+            var stringToSearch = TextFieldInfo.Text.Substring(0, TextFieldInfo.CursorPosition);
+            var indexOfQuerySymbol = stringToSearch.LastIndexOfAny(querySymbols);
+            if (indexOfQuerySymbol >= 0)
+            {
+                suggestionType = SuggestionType.Projects;
```
arturs105: I assume suggestion type will later be determined by the query symbol that was found right? And for now it works only for projects?

willsb: Yeah. I'm not doing any comparisons right now because tags are on hold

Toggl.Foundation.Tests/MvvmCross/ViewModels/StartTimeEntryViewModelTests.cs
```diff
@@ -103,5 +107,200 @@ public sealed class TheDoneCommand : StartTimeEntryViewModelTest
...
+                [InlineData("Nothing")]
+                [InlineData("Testing Toggl mobile apps")]
+                public async Task WhenTheUserBeginsTypingADescription(string description)
+                {
+                    var textFieldInfo = new TextFieldInfo(description, 0);
+
+                    await ViewModel.Initialize(DateParameter.WithDate(DateTimeOffset.UtcNow));
+                    ViewModel.TextFieldInfo = textFieldInfo;
+
+                    var expectedAmountOfCalls = description.Split(' ').Count();
```
arturs105: `expectedAmountOfCalls` isn't used anywhere. I assume it was meant to be passed to the `Received()` call. Also similar stuff in the test below.

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
```diff
@@ -69,6 +78,33 @@ public StartTimeEntryViewModel(ITogglDataSource dataSource, ITimeService timeSer
...
+
+                var startingIndex = indexOfQuerySymbol + 1;
+                var stringLength = TextFieldInfo.Text.Length - indexOfQuerySymbol - 1;
+                queryText = TextFieldInfo.Text.Substring(startingIndex, stringLength);
```
amulware: Not a big fan of the mutable variable. It confused me at first. Consider defining it outside the if, and assign it here and in an else block. Alternatively, extract the whole thing as a function that returns the correct thing.

amulware: If you do the method, it would probably return a tuple of string and suggestion type.

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
```diff
@@ -81,5 +117,40 @@ public StartTimeEntryViewModel(ITogglDataSource dataSource, ITimeService timeSer
...
             await navigationService.Close(this);
         }
+        
+        private IObservable<IEnumerable<ITimeEntrySuggestionViewModel>> querySuggestions(
+            (IEnumerable<string> WordsToQuery, SuggestionType SuggestionType) tuple)
+        {
+            if (!tuple.WordsToQuery.Any())
+                return Observable.Return(Enumerable.Empty<ITimeEntrySuggestionViewModel>());
+
+            if (tuple.SuggestionType == SuggestionType.Projects)
```
amulware: Switch case and a named method for each query?

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
```diff
@@ -81,5 +117,40 @@ public StartTimeEntryViewModel(ITogglDataSource dataSource, ITimeService timeSer
...
+                return tuple.WordsToQuery
+                    .Aggregate(empty<IDatabaseProject>(), (obs, word) => obs.Merge(queryProjects(word)))
```
amulware: This merges several observables, one for each word, right? Shouldn't there be an `.Intersect()` or something? This way we will even get doubles right? (solved by Distinct, but that is not needed if doing an intersection)

Or is this intersection performed somewhere else?

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
```diff
@@ -81,5 +117,40 @@ public StartTimeEntryViewModel(ITogglDataSource dataSource, ITimeService timeSer
...
+
+        private void onSuggestions(IEnumerable<ITimeEntrySuggestionViewModel> suggestions)
+        {
+            Suggestions.Clear();
+            Suggestions.AddRange(suggestions.Distinct(SuggestionComparer.Instance));
+        }
+
+        private IObservable<IEnumerable<IDatabaseTimeEntry>> queryTimeEntries(string word)
+            => dataSource.TimeEntries
+                .GetAll(te => te.Description.Contains(word) || (te.Project != null && te.Project.Name.Contains(word)));
```
amulware: Can realm handle this as an expression, or will the lambda run in memory?

amulware: I think this should filter by client name as well (and tasks in the future).

willsb: As we've discussed earlier, `te.Project` would have to be `te.RealmProject` in order for this to run in Realm :/

amulware: So close.. yet so far away...

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
willsb: Fixed 👌

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
willsb: Not needed for the time being. When we introduce Tags, maybe. But as of now, it just makes things more cumbersome

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
willsb: 👌

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
```diff
@@ -126,12 +127,12 @@ private void OnTextFieldInfoChanged()
 
             if (tuple.SuggestionType == SuggestionType.Projects)
                 return tuple.WordsToQuery
-                    .Aggregate(empty<IDatabaseProject>(), (obs, word) => obs.Merge(queryProjects(word)))
+                    .Aggregate(dataSource.Projects.GetAll(), (obs, word) => obs.Select(filterProjectsByWord(word)))
```
amulware: A shiver runs down my spine. But at least it's honest and straight forward now about doing things in memory.

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
```diff
@@ -1,6 +1,9 @@
 ﻿using System;
-using System.Collections.ObjectModel;
+using System.Collections.Generic;
+using System.Linq;
+using System.Linq.Expressions;
```
arturs105: ❌

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
```diff
@@ -142,15 +143,17 @@ private void onSuggestions(IEnumerable<ITimeEntrySuggestionViewModel> suggestion
...
+                timeEntries.Where(
+                    te => te.Description.ContainsIgnoringCase(word)
+                       || (te.Project != null && te.Project.Name.ContainsIgnoringCase(word))
+                       || (te.Project?.Client != null && te.Project.Client.Name.ContainsIgnoringCase(word)));
+
+        private Func<IEnumerable<IDatabaseProject>, IEnumerable<IDatabaseProject>> filterProjectsByWord(string word)
+            => projects =>
+                projects.Where(
+                    p => p.Name.ContainsIgnoringCase(word)
+                      || (p.Client != null && p.Client.Name.ContainsIgnoringCase(word)));
```
amulware: Can't use ?. (and ?? if needed) here instead of the null check? or does that get more verbose?

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntrySuggestions/ProjectSuggestionViewModel.cs
```diff
@@ -23,13 +24,16 @@ IEnumerable<IDatabaseProject> projects
...
             ProjectName = project.Name;
             ProjectColor = project.Color;
 
             if (project.Client == null) return;
 
             ClientName = project.Client.Name;
         }
+
+        public override int GetHashCode()
+            => (ProjectName + ClientName + ProjectColor).GetHashCode();
```
amulware: Is a GetHashCode needed? If so, why does it do string concatenation on seemingly arbitrary properties?

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntrySuggestions/SuggestionComparer.cs
```diff
@@ -3,16 +3,16 @@
...
+    public sealed class SuggestionComparer : IEqualityComparer<BaseTimeEntrySuggestionViewModel>
     {
         public static SuggestionComparer Instance { get; } = new SuggestionComparer();
 
         private SuggestionComparer() { }
 
-        public bool Equals(ITimeEntrySuggestionViewModel x, ITimeEntrySuggestionViewModel y)
-            => x.Id == y.Id;
+        public bool Equals(BaseTimeEntrySuggestionViewModel x, BaseTimeEntrySuggestionViewModel y)
+            => x.GetHashCode() == y.GetHashCode();
```
amulware: This does not look like a proper implementation of Equals to me at all. If it is, it relies on an implementation of GetHashCode that is very counter to what that method is supposed to mean.

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
willsb: Still fast af 😄

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
willsb: I wrote this query considering Realm expressions, which can't have the null coalescence operator. Honestly, I prefer to leave it this way, since the readability won't be any better

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntrySuggestions/SuggestionComparer.cs
willsb: This is just used to `.Distinct()` the objects so that we don't show similar time entries as different suggestions. If I could distinct with a lambda, that's be fine for me, but I need this interface instead.

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntrySuggestions/ProjectSuggestionViewModel.cs
willsb: This could be any method, really. I just need the means to implement an `IEqualityComparer` capable of selecting Distinct time entries based on the info the user sees on the screen.

amulware: Then this method was broken, as is the one that replaces it. There is no guarantee that it will return unique values.

amulware:
My main issue was not addressed:

GetHashCode should never be used to imply equality. The logical relationship is the other way around (equals -> same hashcode).

Instead, do a member wise comparison, instead of using concatenation which has nothing to do with the problem (performance was not my main concern, though it is a waste of memory to create that many strings), or using hash codes of any kind which are not guaranteed to be unique.

Issue #434, opened 2017-08-23
Add batch update/insert/delete to database repositories


amulware:
The method should take a list of the respective entities and persist them all.

If an entity is already known (check by id), it is updated. Otherwise, it is created.
However, if the entity is set to deleted, and it exists locally, it should be deleted instead.

The trickiest bit about this might be to ensure all realm references are gotten correctly in both cases. We may have to generate more code to update an existing realm model from another one while preserving/updating references appropriately.

---

If this turns out to be a special case of #436, maybe having only that method would be enough. Certainly, if that means we don't have to make PrimeRadiant understand when do create/update/delete (thinking especially about delete here, since that requires some knowledge about the models).

arturs105:
And by "it should be deleted instead" do you mean set the `IsDeleted` or really delete?

Issue #435, opened 2017-08-23
Persist workspaces, clients, projects after fetch


amulware:
This might be possible to do in a generic way, or maybe we need separate implementations.

The idea is simple:

- wait for the respective API observable and take the list of entities fetched from the API
- stuff those entities into the database, matching by ID to update already known entities and otherwise creating new ones
- if the persisting succeeds, change the appropriate since-date in the passed-along contained, and push the whole object into the database (there is no way of changing a single since-date right now. If you can find a non-ugly way of updating single since parameters, feel free to propose so below)
- pass down the same object, with updated since parameter

Ideally, we can do this in a batch update since Realm is a lot more efficient when writing more data in fewer transactions. See: #434

PR #462, opened 2017-08-25
Simple conflict resolution

simonrozsival:
Ref #436 and #434.

This PR contains a static class `PreferLatter` which handles conflict resolution for Workspaces, Clients, Projects, Tags and Tasks. The logic for choosing the correct resolution is based on comparing the time of last change (`At`) and whether or not is the entity deleted (`ServerDeletedAt`).

Unfortunately there is no common interface which would define `At` and `ServerDeletedAt` and it is probably not possible to introduce one at the moment. Workspaces and Tags still have nullable `At` and tags and tasks do not have the `ServerDeletedAt` field (it is not included in the responses from the API server). Therefore the `Resolve` function has to have several overloaded variants (one for every `IDatabase...` model). Ideas for a cleaner implementation are welcome.

Property-based tests ✅  are included for all of the different types of entities. Please check that the logic corresponds to the desired behavior.

This PR does not close the issues - the resolution of time entries must be done before we can close them.

Squash whenever.


willsb:
First: I don't like the fact that this is a static class. I'd prefer to have an `IConflictResolver<T> where T : IBaseModel, IDatabaseSyncable` to make injecting simpler. Then you'd have something like a `PreferLatterConflictResolver<T> : IConflictResolver<T>` with a constructor like `PreferLatterConflictResolver(Func<T, DateTimeOffset> getAt, Func<T, DateTimeOffset> getServerDeletedAt)`. These constructor functions can be replaced by abstract members and a concrete implementation.

Second: Why is introducing new interfaces an issue? We are at the best time possible for introducing new stuff: Since the app hasn't been shipped yet, we don't care about database migration if you need to change models! I actually like having a common interface, that'd make part of what I proposed in the first point unnecessary.

Last but not least: F# would allow us to create type constraints that are based on members, so this is why I can never shut up about it 😄

simonrozsival:
I will consider what you suggest and get back to it on Monday.

The issue with the interfaces is that the entities don't have that much in common: `DateTimeOffset? At` vs. `DateTimeOffset At` plus `DateTimeOffset? ServerDeletedAt` vs. no possibility of having such information.

The issues with `At` will be resolved in the future - tags are already fixed, I am not sure if it has been already rolled out, and I created an issue in backend today for workspaces. The `ServerDeletedAt` is missing in tags and tasks and we would have to create issues for that as well. And we probably will have to, because I am not sure how are we supposed to identify that a tag or task has been deleted.

Yeah, type constraints based on members would be great here...

simonrozsival:
I removed the static class and did something similar to what you proposed in the first paragraph.

Toggl.Foundation/Sync/ConflictResolution/PreferLatter.cs
```diff
@@ -0,0 +1,55 @@
+﻿using Toggl.Foundation.Sync.ConflictResolution.Selectors;
+using Toggl.PrimeRadiant;
+using Toggl.Multivac;
+using Toggl.PrimeRadiant.Models;
+
+namespace Toggl.Foundation.Sync.ConflictResolution
+{
+    internal sealed class PreferLatter<T> : IConflictResolver<T>
```
amulware: 'Latter' indicates the second given, not necessarily the newer one. I think `PreferNewer` would be a better name.

Toggl.Foundation/Sync/ConflictResolution/PreferLatter.cs
```diff
@@ -0,0 +1,55 @@
...
+
+            if (selector.IsDeleted(next))
+                return existing != null
+                    ? ConflictResolutionMode.Delete
+                    : ConflictResolutionMode.Ignore;
+
+            if (existing == null)
+                return ConflictResolutionMode.Create;
+
+            if (selector.LastModified(existing) > selector.LastModified(next))
```
amulware: It might help readability to give these conditions a name by introducing a variable.

Toggl.Foundation/Sync/ConflictResolution/PreferLatter.cs
```diff
@@ -0,0 +1,55 @@
...
+                return ConflictResolutionMode.Ignore;
+
+            if (selector.LastModified(existing) < selector.LastModified(next))
```
amulware: What's the use of this condition if we return the same below regardless?

simonrozsival:
Thanks for the comments. There are many things I must improve, removing the WIP label was a bit too early.

> Is taking modified dates into account enough? What if the entity has not been modified locally, should we not always take the foreign one?

Then the `At` of the incoming entity must be >= to the one of the local one and it will be updated. But maybe just looking at `.IsDirty` could be more readable. And maybe the case `LastModified(existing) > LastModified(next)` cannot even ever happen. I have to think about it for a while => lines 31-36 of the `PreferLatter` class might be useless then. But maybe it it is not. Maybe we start (for some reason - manual initiation? should that be even possible?) two syncs at once and two sets of data arrive, something changed between those two requests and now the more up-to-date one is processed first and the data is processed and persisted. Then the older data is processed and as the `IsDirty == false`, it will be overriden without considering the `At` of the server.

This seems like an edge case, which I am not sure if it might be possible even over some very bad network connection with long delays. But we were talking about possible tens of MBs of data, which can take a lot of time over e.g. EDGE.

> And is this at all taking into account the 5 second rule, or any other such details? Also, what do we do if the local entity has been deleted in the meantime, but that deletion has not been synced yet?
>
> Or, rephrased: What exactly is the use case for this conflict resolver?

I wanted to implement conflict resolution for the "simple" cases (Projects, Clients, Workspaces, Tags, Tasks) first and then either extend it or make a completely different resolver for the "complicated" time entries. Maybe I should have referenced issue #435 to avoid the confusion.

Why these cases are simpler: We consider editing and deleting of time entries in themobile app only and that is where the 5-sec rule comes into consideration.

Toggl.Foundation/Sync/ConflictResolution/PreferLatter.cs
simonrozsival: I know about that. I wanted to make it more explicit. I can remove it to make it less confusing.

Toggl.Foundation/Sync/ConflictResolution/PreferLatter.cs
simonrozsival: OK, I will rename it.

Toggl.Foundation/Sync/ConflictResolution/PreferLatter.cs
simonrozsival: Why not, might be helpful.

Toggl.Foundation/Sync/ConflictResolution/PreferNewer.cs
```diff
@@ -0,0 +1,58 @@
...
+        {
+            Ensure.Argument.IsNotNull(serverEntity, nameof(serverEntity));
+
+            var wasDeletedOnServer = selector.IsDeleted(serverEntity) && localEntity != null;
+            if (wasDeletedOnServer)
+                return ConflictResolutionMode.Delete;
+
+            var wasDeletedOnServerButWasntStoredLocallyBefore =
+                selector.IsDeleted(serverEntity)
+                && localEntity == null;
```
amulware: I propose:
```
if (selector.IsDeleted(serverEntity))
    return localEntity == null ? Ignore : Delete;
```
To replace all the above (and would say you should use `using static` for the enum).

amulware: Maybe this is overcomplicating it, but you _could_ add helper methods to be able to write:
```
if (isDeleted(serverEntity))
```

amulware:
This is currently not building because the project file was not saved after the renaming btw. Otherwise, I think it's almost done?

simonrozsival:
Yes. I already have a fix locally, I will push the 🐛 fix + 💄 conditions in a minute.

Issue #438, opened 2017-08-23
Persist data after fetch


amulware:
We want to persist workspaces, tags, clients, projects, tasks and time entries in order while passing on the parallel API request observables. The order is important because later entries might depend on earlier ones to be stored correctly (except that tags can happen anytime between workspaces and time entries).

For now, we will not bother ourselves with dealing with the user being removed from workspaces. We will also not worry about tags and tasks just yet.

That leaves workspaces, clients, projects and time entries.

Only the latter can be changed within our app, so that is the only one we have to run conflict resolution for.

Hence, let's split this into two parts:

- [x] persist workspaces, clients, projects: #435
- [ ] persist time entries: #437

PR #467, opened 2017-08-28
Persist workspaces, clients, and projects after fetch

### LINKED ISSUE #434: Add batch update/insert/delete to database repositories

simonrozsival:
Closes #434, #435, and #436.

This PR includes data persisting state for the sync state machine which takes a `FetchObservables` object as its input and awaits all the API requests and persists the data into the local database using batch updates. The order of persisting the different requests defined in #438 is implemented and there is a test to ensure it. There is no _parameter_ of the result of this state at the moment, as the input of a following state is not specified yet.

❕ Dealing with time entries isn't included in this pull request.


simonrozsival:
I refactored the code after an internal discussion so now:

- there is no one big state, but rather 3 smaller states
- observables are used instead of tasks
- since-parameter is not mutable any more

This PR should be ready for reviews now.

Toggl.Foundation/Sync/States/BasePersistState.cs
```diff
@@ -0,0 +1,54 @@
...
+            {
+                var fetch = transition.Parameter;
+                var since = fetch.SinceParameters;
+
+                var sinceParameters = await FetchObservable(fetch)
+                    .SingleAsync()
+                    .Select(ConvertToDatabaseEntities)
+                    .SelectMany(BatchUpdate(database))
+                    .Select(LastUpdated(since))
+                    .Select(UpdateSinceParameters(since));
```
amulware: The since parameters have to be updated with the newest At date among all the fetched entities. It seems to me that BatchUpdate will not return the same entities passed in, will it? So the since date may not be updated quite right?

What exactly is BatchUpdate supposed to return btw?

Toggl.Foundation/Sync/States/FetchObservables.cs
```diff
@@ -16,13 +16,20 @@ internal sealed class FetchObservables
         public FetchObservables(ISinceParameters sinceParameters,
             IObservable<List<IWorkspace>> workspaces, IObservable<List<IClient>> clients,
             IObservable<List<IProject>> projects, IObservable<List<ITimeEntry>> timeEntries
+        ) : this(null, sinceParameters, workspaces, clients, projects, timeEntries)
+        {
+        }
+
+        public FetchObservables(FetchObservables old, ISinceParameters sinceParameters = null,
+            IObservable<List<IWorkspace>> workspaces = null, IObservable<List<IClient>> clients = null,
+            IObservable<List<IProject>> projects = null, IObservable<List<ITimeEntry>> timeEntries = null
```
amulware: Are all these observable arguments used somewhere? And does the sinceParameters one have to be optional?

Toggl.Foundation.Tests/Sync/States/PersistStateTests.cs
```diff
@@ -0,0 +1,141 @@
...
+                    }));
+                var fetchTransition = new Transition<FetchObservables>(new StateResult<FetchObservables>(), fetchObservables);
+
+                Action fetchTwice = () => state.Start(fetchTransition).Wait();
+
+                fetchTwice.ShouldThrow<InvalidOperationException>();
+            }
+
+            [Fact]
+            public void TrigersBatchUpdates()
```
amulware: Typo. Also the method is `BatchUpdate`, singular, so I think this should be `TriggersBatchUpdate`.

Toggl.Foundation/Sync/States/FetchObservables.cs
simonrozsival: I added this method to create a new instance of FetchObservables based on the older one and change just a small part of it. I wanted to use that in tests, but I didn't create a test for all of the different entity types yet.

You are right that there is no reason for sinceParameters to be optional.

Toggl.Foundation/Sync/States/BasePersistState.cs
simonrozsival: The BatchUpdate returns the list of the entities, which were persisted to the database - the items which were deleted and the outdated ones are filtered out. Some of the deleted entities might have `at` greater than the ones left, so you are right about that. I will fix that.

Toggl.Foundation/Sync/ConflictResolution/PreferNewerFactory.cs
```diff
@@ -0,0 +1,23 @@
+﻿using Toggl.Foundation.Sync.ConflictResolution.Selectors;
+using Toggl.PrimeRadiant.Models;
+
+namespace Toggl.Foundation.Sync.ConflictResolution
+{
+    internal static class PreferNewerFactory
```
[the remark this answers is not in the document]
simonrozsival: I will try to come up with something else.

Toggl.Foundation/Sync/States/BasePersistState.cs
```diff
@@ -0,0 +1,57 @@
...
+
+                var entities = FetchObservable(fetch)
+                    .SingleAsync()
+                    .Select(ConvertToDatabaseEntities);
+
+                var sinceParameters = await entities
+                    .Select(LastUpdated(since))
+                    .Select(UpdateSinceParameters(since));
+
+                await entities.SelectMany(BatchUpdate(database));
```
amulware: You run the `entities` 'query' twice, so I think `ConvertToDatabaseEntities` might be applied twice as well, no?

Might it not be better to make the list of entities concrete and work on that? I don't quite see why the since parameter query has to be async in either case.

Alternatively put these into a single query again, and extract the since parameters with a `Do()` call before the batch update call.

Toggl.Foundation/Sync/States/BasePersistState.cs
```diff
@@ -0,0 +1,57 @@
...
+using Toggl.PrimeRadiant.Models;
+
+namespace Toggl.Foundation.Sync.States
+{
+    internal abstract class BasePersistState<TInterface, TDatabaseInterface>
+        where TDatabaseInterface : TInterface
+    {
+        private readonly ITogglDatabase database;
+
+        public StateResult<FetchObservables> PersistingFinished { get; } = new StateResult<FetchObservables>();
```
amulware: Regarding setting up the state machine, would it maybe read nicer if this was `FinishedPersisting`?
```
handlerProvider.ConfigureTransition(persistWorkspaces.FinishedPersisting, persistClients.Start);
```

Toggl.Foundation/Sync/States/FetchObservables.cs
```diff
@@ -16,13 +16,20 @@ internal sealed class FetchObservables
...
-            SinceParameters = sinceParameters;
-            Workspaces = workspaces;
-            Clients = clients;
-            Projects = projects;
-            TimeEntries = timeEntries;
+            SinceParameters = old?.SinceParameters ?? sinceParameters;
+            Workspaces = old?.Workspaces ?? workspaces;
+            Clients = old?.Clients ?? clients;
+            Projects = old?.Projects ?? projects;
+            TimeEntries = old?.TimeEntries ?? timeEntries;
```
amulware: Do I see correctly that all the observable parameters are only used in the tests? In that case, Please remove them here and write a method for this in the test project, if needed.

We should avoid having any code pandering to tests.

amulware:
Oh..
one more thing

Please also test:
- that it throws when batch updating throws
- that it does not update since when batch updating throws
- that the emitted parameters have the correct updated since parameters in them (otherwise future states will override things badly - maybe we really should have a way of setting just single parameters?)

Toggl.Foundation/Sync/States/BasePersistState.cs
simonrozsival: Good point.

Toggl.Foundation/Sync/States/BasePersistState.cs
```diff
@@ -24,15 +25,13 @@ public IObservable<ITransition> Start(Transition<FetchObservables> transition)
                 var fetch = transition.Parameter;
                 var since = fetch.SinceParameters;
 
-                var entities = FetchObservable(fetch)
-                    .SingleAsync()
-                    .Select(ConvertToDatabaseEntities);
+                var entities = await FetchObservable(fetch).SingleAsync();
+                var list = entities.Select(ConvertToDatabaseEntity).ToList();
```
amulware: `databaseEntities`?

Toggl.Foundation/Sync/States/BasePersistState.cs
```diff
@@ -24,15 +25,13 @@ public IObservable<ITransition> Start(Transition<FetchObservables> transition)
...
 
-                var sinceParameters = await entities
-                    .Select(LastUpdated(since))
-                    .Select(UpdateSinceParameters(since));
+                var lastUpdated = LastUpdated(since, list);
+                var sinceParameters = UpdateSinceParameters(since, lastUpdated);
 
-                await entities.SelectMany(BatchUpdate(database));
+                await BatchUpdate(database, list);
```
amulware: Why not do this before calculating the last updated and new since parameters, given they are only used below?

amulware:
Otherwise I approve.

Toggl.Foundation/Sync/States/BasePersistState.cs
simonrozsival: I didn't really think about this that much since these operations are independent. BatchUpdate might fail though and then calculating SinceParameters was pointless... => I swaped it.

Toggl.Foundation/Sync/States/BasePersistState.cs
```diff
@@ -0,0 +1,56 @@
...
+
+        public StateResult<FetchObservables> FinishedPersisting { get; } = new StateResult<FetchObservables>();
+
+        protected BasePersistState(ITogglDatabase database)
+        {
+            this.database = database;
+        }
+
+        public IObservable<ITransition> Start(Transition<FetchObservables> transition)
+            => Observable.Create<ITransition>(async observer =>
```
willsb: I strongly prefer combining observables over wrapping TPL code in `Observable.Create`. The reason we do it in PrimeRadiant/Ultrawave is because there's no other way. Here, I'd go for something among the lines of: 

```csharp
return FetchObservable(fetch)
    .SingleAsync()
    .Select(entities => entities.Select(ConvertToDatabaseEntity))
    .SelectMany(databaseEntities => BatchUpdate(database, databaseEntities.ToList()))
    .Select(databaseEntities => LastUpdated(since, databaseEntities))
    .Select(lastUpdated => UpdateSinceParameters(since, lastUpdated))
    .Do(database.SinceParameters.Set)
    .Select(sinceParameters => new FetchObservables(fetch, sinceParameters))
    .Select(FinishedPersisting.Transition);
```

Here, every step in the transformation is clear (even the side effects), unlike things like `await BatchUpdate `'s ignored result and `return () => { }`

Toggl.Foundation.Tests/Sync/States/PersistStateTests.cs
```diff
@@ -0,0 +1,236 @@
...
+                var oldSinceParameters = new SinceParameters(null, at);
+                var entities = CreateComplexList(at);
+                var fetchTransition = CreateTransition(oldSinceParameters, entities);
+                SetupDatabaseBatchUpdateToThrow(database, () => new TestException());
+
+                try
+                {
+                    state.Start(fetchTransition).SingleAsync().Wait();
+                }
+                catch (TestException e) {
```
willsb: MSBuild complains this `e` is not used anywhere; consider deleting it

Toggl.Foundation.Tests/Sync/States/PersistStateTests.cs
```diff
@@ -0,0 +1,236 @@
...
+                var fetchTransition = CreateTransition(oldSinceParameters, entities);
+                SetupDatabaseBatchUpdateMocks(database, entities);
+
+                state.Start(fetchTransition).SingleAsync().Wait();
+
+                database.SinceParameters.Received().Set(Arg.Is<ISinceParameters>(
+                    (newSinceParameters) => OtherSinceDatesDidntChange(oldSinceParameters, newSinceParameters, at)));
+            }
+
+            public void SinceDatesAreNotUpdatedWhenBatchUpdateThrows()
```
willsb: Where is this method being used? There are no references to it and it doesn't have a `Fact` attribute on it

Toggl.Foundation.Tests/Sync/States/PersistStateTests.cs
```diff
@@ -0,0 +1,236 @@
...
+                    state.Start(fetchTransition).SingleAsync().Wait();
+                }
+                catch (TestException e) {
+                    Console.WriteLine("Caught TestException");
+                }
+
+                database.SinceParameters.DidNotReceiveWithAnyArgs().Set(null);
+            }
+
+            public void PassesTheNewSinceParametersThroughTheTransition()
```
willsb: Similar to above

Toggl.Foundation.Tests/Sync/States/PersistStateTests.cs
simonrozsival: Thanks. I forgot to add the `[Fact]`. I will add it.

Toggl.Foundation.Tests/Sync/States/PersistStateTests.cs
simonrozsival: 🔥

Toggl.Foundation/Sync/States/BasePersistState.cs
simonrozsival: The result of `BatchUpdate` should be ignored in this case, we need the original sequence to determine the new since parameter. But otherwise this looks good, I will use it!

willsb: I see. Well, a `.Do` will make it very explicit that you're ignoring the result 😄

PR #468, opened 2017-08-28
Persist time entry description edits

### LINKED ISSUE #441: Get first editting up and running

arturs105:
Closes #441 

Time entry description editing is now working properly.

The DTO for now is quite bare, but other stuff will be added to it later.

Squash whenever.


willsb:
I think that we should stick with the Builder pattern instead. You can use the DTO for, well, data transfer, but you should still use the Builder internally in Foundation to make things consistent.

Toggl.Foundation/DataSources/TimeEntriesDataSource.cs
```diff
@@ -94,5 +101,24 @@ private void safeSetCurrentlyRunningTimeEntry(IDatabaseTimeEntry timeEntry)
             var next = timeEntry == null ? null : TimeEntry.Clean(timeEntry);
             currentTimeEntrySubject.OnNext(next);
         }
+
+        private TimeEntry createUpdatedTimeEntry(IDatabaseTimeEntry timeEntry, EditTimeEntryDto dto)
+            => TimeEntry.Builder.Create(dto.Id)
+                        .SetDescription(dto.Description)
+                        .SetAt(timeEntry.At)
```
simonrozsival: We also have to remember the time when the entity was edited for conflict resolution when we pull data from the server.

Toggl.Foundation/Models/TimeEntry.cs
```diff
@@ -56,6 +81,74 @@ public Builder SetBillable(bool billable)
...
+            {
+                if (tagNames == null) return this;
+                TagNames.AddRange(tagNames);
+                return this;
+            }
+
+            public Builder SetTagIds(IEnumerable<long> tagIds)
+            {
+                if (tagIds == null) return this;
+                TagIds.AddRange(tagIds);
```
willsb: Why name this `Set` if you're effectively `Add`ing? (similar above, on `SetTagNames`)

Toggl.Foundation/Models/TimeEntry.cs
```diff
@@ -56,6 +81,74 @@ public Builder SetBillable(bool billable)
...
+                return this;
+            }
+
+            public Builder SetIsDeleted(bool isDeleted)
+            {
+                IsDeleted = isDeleted;
+                return this;
+            }
+
             private void ensureValidity()
```
willsb: Hasn't this changed? Considering Simon's comment, maybe we should ensure the `At` date is set (and then fix it elsewhere in the App).

arturs105: Yeah, but this builder isn't used just to update a time entry. There may be cases where the `At` field doesn't need to be set.

Toggl.Foundation/Models/TimeEntry.cs
arturs105: We could either rename to `AddTagNames` or clear the existing tags before adding the new ones (then the name would be correct).

Toggl.Foundation/DataSources/TimeEntriesDataSource.cs
```diff
@@ -94,5 +101,24 @@ private void safeSetCurrentlyRunningTimeEntry(IDatabaseTimeEntry timeEntry)
...
+                        .SetTaskId(timeEntry.TaskId)
+                        .SetUserId(timeEntry.UserId)
+                        .SetBillable(timeEntry.Billable)
+                        .SetTagNames(timeEntry.TagNames)
+                        .SetIsDeleted(timeEntry.IsDeleted)
+                        .SetProjectId(timeEntry.ProjectId)
+                        .SetCreatedWith(timeEntry.CreatedWith)
+                        .SetWorkspaceId(timeEntry.WorkspaceId)
+                        .SetServerDeletedAt(timeEntry.ServerDeletedAt)
+                        .SetAt(DateTimeOffset.Now)
```
amulware: This needs to use the time service.

arturs105:
I also added a test to check that other properties are not changed. Note that such properties do not include the ones that will be editable in the future, but are not implemented currently, such as tags, project, task etc.

Toggl.Foundation/Models/TimeEntry.cs
simonrozsival: Well I don't think there is ever a case when `At` is not set. `DateTimeOffset` is not nullable and the default value is "2001-01-01 12:00:00". But I would prefer if the Builder had a nullable `At` and we validated that it was properly checked. Unless you can think of a case where not having `At` makes sense.

amulware:
There should always be a valid At date, not having it makes no sense.

Toggl.Foundation.MvvmCross/ViewModels/EditTimeEntryViewModel.cs
```diff
@@ -133,7 +135,10 @@ private void onDeleteCompleted()
 
         private void confirm()
         {
-            throw new NotImplementedException();
+            var dto = new EditTimeEntryDto { Id = Id, Description = Description };
+            confirmDisposable = dataSource.TimeEntries
+                                          .Update(dto)
+                                          .Subscribe((Exception ex) => close(), () => close());
```
willsb: Should we create issues to decide what happens when this errors?

Toggl.Foundation.Tests/DataSources/TimeEntriesDataSourceTests.cs
```diff
@@ -327,5 +351,77 @@ public void PropagatesErrorIfTimeEntryIsNotInRepository()
...
+        public sealed class TheUpdateMethod : TimeEntryDataSourceTest
+        {
+            private EditTimeEntryDto prepareTest()
+            {
+                var observable = Observable.Return(DatabaseTimeEntry);
+                Repository.GetById(Arg.Is(DatabaseTimeEntry.Id)).Returns(observable);
+                return new EditTimeEntryDto { Id = DatabaseTimeEntry.Id, Description = "New description" };
+            }
+
+            private bool ensurePropertiesNotChanged(IDatabaseTimeEntry timeEntry)
```
willsb: `ensurePropertiesDidNotChange`

Toggl.Foundation.Tests/DataSources/TimeEntriesDataSourceTests.cs
```diff
@@ -327,5 +351,77 @@ public void PropagatesErrorIfTimeEntryIsNotInRepository()
...
+            }
+
+            [Fact]
+            public async ThreadingTask UpdatesTheIsDirtyProperty()
+            {
+                var dto = prepareTest();
+
+                await TimeEntriesSource.Update(dto);
+
+                await Repository.Received().Update(Arg.Is<IDatabaseTimeEntry>(te => te.IsDirty == true));
```
willsb: No  need to compare it with true

Toggl.Foundation.Tests/DataSources/TimeEntriesDataSourceTests.cs
```diff
@@ -327,5 +351,77 @@ public void PropagatesErrorIfTimeEntryIsNotInRepository()
...
+            }
+
+            [Fact]
+            public async ThreadingTask UpdatesTheAtProperty()
+            {
+                var dto = prepareTest();
+
+                await TimeEntriesSource.Update(dto);
+
+                await Repository.Received().Update(Arg.Is<IDatabaseTimeEntry>(te => te.At != DatabaseTimeEntry.At));
```
willsb: This should check that te.At > DatabaseTimeEntry.At

Toggl.Foundation.MvvmCross/ViewModels/EditTimeEntryViewModel.cs
amulware: Apart from that it really shouldn't? Later all these sorts of things need to be logged/reported, but for now, I'm not sure what we can really do here.

Issue #437, opened 2017-08-23
Persist time entries after fetch


amulware:
Time entries need conflict resolution because they can be changed inside the app so this is a bit trickier than persisting other entities.

What there is to do:
- wait for API observable to return the list of time entries
- persist all time entries into the database using conflict resolution for those already existing, and adding all others
- if the persisting succeeds, change the appropriate since-date in the passed-along contained, and push the whole object into the database (there is no way of changing a single since-date right now. If you can find a non-ugly way of updating single since parameters, feel free to propose so below)
- pass down the same object, with updated since parameter

To improve performance with Realm, and avoid race conditions, we cannot fetch from Realm, and then write an updated time entry, but have to do both of that inside a single transaction. While we are at it, we might as well update the whole batch at once. 🐎 

The way the conflict resolution should work as follows:
- if the time entry does not exist locally
  - if the entry is deleted, do nothing
  - otherwise add it
- if the time entry exists locally
  - use the conflict resolution strategy outlined in our webapp's docs (see internal documentation)

This will require new functionality on our repositories as well: #436

amulware:
One more edge case that is very important:

We need to ensure 100%, that there will never be two running time entries (entries without stop time).
This case somehow needs to be detected and dealt with. The webapp has the logic to deal with it, but as for the detection and how to implement it, I'm not sure right now. But it has to happen in the same transaction that also persists the entries.

amulware:
For now, we won't ensure the single-running-entry condition, since it makes things rather complicated.

PR #474, opened 2017-08-29
Resolve conflicts for time entries and other mutable entities

simonrozsival:
Ref #437. This PR contains extended implementation of conflict resolution for fetching data for entities which might be edited in the mobile app (at the moment that means just time entries).

If we already have the entity stored locally and it is clean, than we always override it with the incoming data.

If the loclal entity is dirty, then we protect the local changes from being overridden when we are pulling data from the server by checking the time of last change.

_Edge case_: After we already send the data to the server (PUT), but couldn't persist the response (the user edited the entity locally in the meantime), the entity is already update on server and its `at` attribute is possibly greater than the one we have locally. If a there is a _pull_ operation before another push, then we would override the local data. We try to prevent this by subtracting a "margin of error" from the timestamp of the incoming data - this is an inspiration from the webapp ("the 5 second solution"). The margin of error for the time entities is set to the 5 seconds for now.

The implementation of the conflict resolver which was created previously for simple ("immutable") entities is just a special case of this resolver, where there is no margin of error. In fact, these entities will stay _clean_ and we will never choose the ignore conflict resolution mode.

What this PR does NOT include:
- resolving two running time entries conflict (there is a separate issue #476)
- persisting the time entries into the local database


Toggl.Foundation.Tests/Sync/ConflictResolution/PreferNewerTests.cs
```diff
@@ -77,20 +76,77 @@ public void IgnoreWhenTheIncomingDataHasSomeServerDeletedAtButThereIsNoExistingE
...
+
+            var mode = resolver.Resolve(existingEntity, incomingEntity);
+
+            mode.Should().Be(ConflictResolutionMode.Update);
+        }
+
+        [Property]
+        public void UpdateIncommingIfUserMadeChangesLocallyRecentlyButThereIsNoMarginOfError(DateTimeOffset existing, DateTimeOffset incoming)
+        {
+            if (existing > incoming || existing < incoming.AddSeconds(-1)) return;
```
amulware: I'm worried that a condition like this might cause all runs of the test to simple return early instead of actually running the test. Is there something we can do to make it more likely the code below will actually run?

Toggl.Foundation.Tests/Sync/ConflictResolution/PreferNewerTests.cs
```diff
@@ -77,20 +76,77 @@ public void IgnoreWhenTheIncomingDataHasSomeServerDeletedAtButThereIsNoExistingE
...
+            var existingEntity = new TestModel(existing, dirty: true);
+            var incomingEntity = new TestModel(incoming);
+
+            var mode = resolver.Resolve(existingEntity, incomingEntity);
+
+            mode.Should().Be(ConflictResolutionMode.Update);
+        }
+
+        [Property]
+        public void UpdateIncommingIfUserMadeChangesLocallyRecentlyButThereIsNoMarginOfError(DateTimeOffset existing, DateTimeOffset incoming)
```
amulware: "UpdateIncoming"? I'm not sure the name makes sense.

Toggl.Foundation/Sync/ConflictResolution/PreferNewer.cs
```diff
@@ -26,7 +37,10 @@ public ConflictResolutionMode Resolve(T localEntity, T serverEntity)
             if (localEntity == null)
                 return Create;
 
-            var receivedDataIsOutdated = selector.LastModified(localEntity) > selector.LastModified(serverEntity);
+            if (selector.IsDirty(localEntity) == false)
```
amulware: We have no guideline on this, but I prefer `!selector.IsDirty(localEntity)`.
Not sure if this is worth enforcing consistency on though (either way)?

simonrozsival: I personally prefer the more verbose comparison. :D

Toggl.Foundation.Tests/Sync/ConflictResolution/PreferNewerTests.cs
simonrozsival: Oh..

Toggl.Foundation.Tests/Sync/ConflictResolution/PreferNewerTests.cs
```diff
@@ -25,7 +32,12 @@ public void ThrowsWhenIncomingEntityIsNull()
         [Property]
         public void IgnoreOutdatedIncomingDataWhenLocalEntityIsDirty(DateTimeOffset existing, DateTimeOffset incoming)
         {
-            if (existing <= incoming) return;
+            if (existing <= incoming)
+            {
+                var tmp = existing;
+                existing = incoming;
+                incoming = tmp;
```
amulware: Please use tuple syntax: `(existing, incoming) = (incoming, existing);`

amulware:
:shipit:

Issue #448, opened 2017-08-24
Start/stop time edit dialog


amulware:
This will be used for editing the start time, stop time, and choosing the start time when creating a time entry.

See zeplin, poke @pe0ny for designs/functionality.

- [x] Modify `ModalPresentationController` or create a new one for smaller dialogs
- [x] Create the dialog
- [x] Wait for MvvmCross 5.2
- [x] Return selected value from the edit dialog

pe0ny:
Just a note: we don't have option to edit stop time - only start time and duration (which also changes end time)

arturs105:
This was actually closed in #566

Issue #446, opened 2017-08-24
Duration edit screen


amulware:
This will be used for both editing and creating time entries, except that it has to work slightly differently in those cases. The similarities in UI are enough to share the code though I believe.

@pe0ny is the expert on how exactly the UI is gonna work, consider poking her (or maybe she can add details here).

PR #475, opened 2017-08-30
Add presentation controller for a modal dialog

arturs105:
Ref #448 and #446 

This adds necessary classes to present view controllers as modal dialogs like this:
aug-30-2017 09-49-52


willsb:
I believe this guy can't be dismissed by swiping down, am I right?

arturs105:
Nope, it can't be swiped down.

PR #483, opened 2017-08-31
Add start date picker dialog to edit view

arturs105:
Ref #448 

The selected value isn't set in the edit view for now, because we need to wait for the new version of MvvmCross or do it in a hacky way.

The changes in `TogglPresenter` are because presenting the dialog wasn't possible if the `MasterNavigationController` is already presenting the edit view.

Squash whenever


arturs105:
After looking at the binding sources, it looks like it is supposed to work with `DateTime`s out of the box. But it doesn't. Should be working now

Toggl.Foundation.MvvmCross/ViewModels/SelectDateTimeDialogViewModel.cs
```diff
@@ -0,0 +1,47 @@
...
+    [Preserve(AllMembers = true)]
+    public class SelectDateTimeDialogViewModel : MvxViewModel<DateParameter, DateParameter>
+    {
+        private IMvxNavigationService navigationService;
+
+        private DateTime dateTime;
+        public DateTime DateTime
+        {
+            get => dateTime;
+            set => dateTime = DateTime.SpecifyKind(value, DateTimeKind.Utc);
```
willsb: This is a bit ugly, but still better than a converter, I think

Toggl.Daneel/ViewControllers/SelectDateTimeDialogViewController.cs
```diff
@@ -0,0 +1,48 @@
...
+
+            bindingSet.Apply();
+        }
+
+        private void prepareDatePicker()
+        {
+            var screenWidth = UIScreen.MainScreen.Bounds.Width;
+            PreferredContentSize = new CGSize
+            {
+                //ScreenWidth - 32 for 16pt marings on both sides
```
willsb: Typo when writing `margins`

Toggl.Foundation.MvvmCross/ViewModels/SelectDateTimeDialogViewModel.cs
```diff
@@ -0,0 +1,47 @@
...
+            this.navigationService = navigationService;
+
+            CloseCommand = new MvxAsyncCommand(close);
+            SaveCommand = new MvxCommand(save);
+        }
+        
+        public override Task Initialize(DateParameter parameter)
+        {
+            DateTime = parameter.GetDate().LocalDateTime;
+            return Task.FromResult(0);
```
willsb: We agreed we should `await base. Initialize()` instead

Issue #315, opened 2017-07-27
Time entry creation view: Autocomplete


pe0ny:
- When user is typing without using @ or #, we suggest time entries
- When user hits # we suggest tags aka 'tags suggestion mode':
![toggl - ios - tags 1x](https://user-images.githubusercontent.com/5064445/28671765-7efe0d9c-72e6-11e7-99a8-edc5d1a6fd4b.png)
- When user hits # and then backspace, tags suggestion mode will be closed.
- To get out of tag suggestion mode, user needs to 
a) Pick a tag
b) Hit backspace enough times to remove # 
c) Tap tags button (the green one)
- When user hits @ we suggest projects (with clients and tasks) aka 'projects suggestion mode':
![toggl - ios - add new entry project search and autocomplete 1x](https://user-images.githubusercontent.com/5064445/28671174-65d56c9a-72e4-11e7-9517-2259bb9c590a.png)
- [x] When user hits @ and then backspace, projects suggestion mode will be closed.
- [x] When user hits @, starts typing and hits space, we still suggest projects. This allows entering project names with spaces.
- To get out of project suggestion mode, user needs to:
  - [x] Pick a project (including 'No project')
  - [x] Hit backspace enough times to remove @ 
  - [x] Press the project button (green folder button right above keyboard)
- [x] Show 'No project' option always when project suggestion mode is open
- [x] When user has typed in description AND picked a project AND starts typing in normal mode (not in projects suggestion mode), we do not suggest time entries anymore. *(For future reference, we can consider suggesting time entries with picked project)*
- [x] When user is in project suggestion mode and taps 'No project', it closes the project suggestion mode and adds 'no project' tag in the input field.
- [x] When user is in project suggestion mode and has typed in *something*, eg '@abc-project' and taps 'No project', it closes the project suggestion mode and replaces whatever user had typed in with 'no project' tag.
- [x] When user is in project suggestion mode and taps green project-icon below, project suggestion mode closes.
- [x] When user is in project suggestion mode and has typed in *something*, eg '@abc-project' and taps green project-icon below, project suggestion mode closes (we don't store the string user had typed in before).
- [x] When user hits second @ after already picked one project, projects suggestion mode won't open again, instead user can keep typing in normal/description mode using @ in description

- [x] When user types in description, hits @ (enters projects suggestion mode) and then moves cursor into description (text typed in before), then user is still in projects suggestion mode until he/she starts typing
[...]

PR #488, opened 2017-08-31
Suggest "No Project" when starting time entries

willsb:
Ref #444 

This adds the suggestion of “No Project” when opening the Project Suggestion mode with no queries, as pointed by #315 

Squash whenever.


Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntrySuggestions/ProjectSuggestionViewModel.cs
```diff
@@ -20,7 +31,14 @@ IEnumerable<IDatabaseProject> projects
 
         public string ProjectColor { get; }
 
-        public ProjectSuggestionViewModel(IDatabaseProject project)
+        private ProjectSuggestionViewModel()
+        {
+            ProjectId = 0;
+            ProjectName = Resources.NoProject;
+            ProjectColor = "#A3A3A3";
```
amulware: Not sure if this should be in the shared project, but I suppose it's sorta business logic and not just UI.

willsb:
Issue #315 clearly mentions:
>Show 'No project' option only when user hasn't typed in anything after '@'

Unless @pe0ny meant that this is meant to be the only thing that appears, in which case this can be pretty easily solved :)

amulware:
Let's get clarification to make sure.

PR #489, opened 2017-08-31
Allow selection of TimeEntry suggestions

### LINKED ISSUE #442: Description auto completion in time entry creation dialog

willsb:
Closes #442 

This one allows users to select a time entry suggestion. It sets the suggestion as the description and also sets the project (though this is still not visible, since this is a rather complex change and will be in another PR).

Persisting the Project Id is covered by #487


amulware:
Just wondering: Will this also add the @project-name label thing to the text field?

willsb:
@amulware It will not, this is part of a different PR

PR #490, opened 2017-08-31
Login UI fixes

### LINKED ISSUE #482: Log in view UI fixes
- [x] Check back, 'Log in', and next font sizes and weights from zeplin - these don't seem to be correct
- [x] Back/next are not aligned correctly in login view (check padding 16pt from left and right)
- [x] Remove 1password icon from login view since it won't be needed in beta (low prio)

willsb:
Closes #482 

Only the second item is still missing


amulware:
If the second item is missing, then this does not close the issue.

willsb:
Forgot the WIP label :P

Issue #440, opened 2017-08-24
Set and hook up first state machine for fetching on app start


amulware:
What the title says. Requires #438

We should see about where to expose the state machine, but it's probably gonna be in the main data source somewhere. I'm not entirely sure who should be in charge of creating the sync manager for the data source, but maybe it does not matter that much right now.

We should then trigger a full sync anytime the user logins or opens the app when logged in. This trigger should happen from the UI I think, so that the app has control over it, and can consume the created observable.

- [x] Set up state machine
- [x] Expose state machine in data source
- [x] Trigger full sync on login/app start when logged in

willsb:
For reference:  The best point in the lifecycle to start the sync is [here](https://github.com/toggl/mobileapp/blob/develop/Toggl.Foundation.MvvmCross/App.cs#L43), but it should be made it so that starting the sync machine on a `BackgroundTask` or on an Android `Service` is simple, allowing for silent background syncs :shipit:

amulware:
@willsb how do you mean "it should be made it so that starting the sync machine on a BackgroundTask or on an Android Service is simple"? The state machine is all observy and stuff so it will already run almost everything in the background.

willsb:
That much I get it, I'm sure you take your observy stuff serious!

What I meant is that the state machine should be thought to be started in situations such as iOS' [Backgorund Fetch](https://developer.apple.com/library/content/documentation/iPhone/Conceptual/iPhoneOSProgrammingGuide/BackgroundExecution/BackgroundExecution.html) and Android's [Background Service](https://developer.android.com/training/run-background-service/create-service.html), where you have some limitations (namely: MvvmCross is not fully available, etc).

simonrozsival:
I am closing this issue manually because the PR just referenced the issue, but didn't say it was closing it.

PR #492, opened 2017-09-01
Setup state machine

simonrozsival:
Ref #440 

Some issues I would like to consult:

- [x] The transitions are created in the `TogglSyncService`, which is only a wrapper for the `SyncManager` otherwise, and it is exposed in the `ITogglDataSource`. I am not sure if it is a good idea, maybe there should be rather some factory which would just create a `SyncManager`.
- [x] I am not sure which `IScheduler` should be used - now I went with `Scheduler.CurrentThread`.
- [x] for now I put only the `FullSync()` in `App.cs`, ~~but it should be running in some background thread I suppose, so that is one thing to consider.~~ as it as all _observy_, it is already async.
- [x] There is a problem with `SinceParameters` - the object is retrieved from Realm, but is passed to a different thread and as soon as it is accessed, an exception is thrown. `Foundation` should convert the Realm models to Foundation models as soon as it receives them (@willsb suggested  #having a `.Wrap(realmModel)` function).


Toggl.Foundation/TogglSyncService.cs
```diff
@@ -0,0 +1,56 @@
...
+            var persistWorkspaces = new PersistWorkspacesState(database);
+            var persistClients = new PersistClientsState(database);
+            var persistProjects = new PersistProjectsState(database);
+            var persistTimeEntries = new PersistTimeEntriesState(database);
+
+            transitions.ConfigureTransition(entryPoint, fetchAllSince.Start);
+            transitions.ConfigureTransition(fetchAllSince.FetchStarted, persistWorkspaces.Start);
+            transitions.ConfigureTransition(persistWorkspaces.FinishedPersisting, persistClients.Start);
+            transitions.ConfigureTransition(persistClients.FinishedPersisting, persistProjects.Start);
+            transitions.ConfigureTransition(persistProjects.FinishedPersisting, persistTimeEntries.Start);
```
amulware: Maybe we want extension methods to allow chaining? `.configure(result, start).configure(result, start)...`

Doesn't have to be part of this PR but could be.

Toggl.PrimeRadiant.Realm/Repository.cs
```diff
@@ -30,7 +30,7 @@ public IObservable<IEnumerable<TModel>> BatchUpdate(IEnumerable<TModel> entities
             Ensure.Argument.IsNotNull(entities, nameof(entities));
             Ensure.Argument.IsNotNull(conflictResolution, nameof(conflictResolution));
 
-            return CreateObservable(() => Adapter.BatchUpdate(entities, matchById, conflictResolution));
+            return Observable.Return(Adapter.BatchUpdate(entities, matchById, conflictResolution));
```
amulware: What's this for?

Toggl.Foundation/DataSources/TogglDataSource.cs
```diff
@@ -20,6 +21,7 @@ public TogglDataSource(ITogglDatabase database, ITogglApi api)
             User = new UserDataSource(database.User, api.User);
             Projects = new ProjectsDataSource(database.Projects);
             TimeEntries = new TimeEntriesDataSource(database.IdProvider, database.TimeEntries);
+            SyncService = new TogglSyncService(database, api, Scheduler.CurrentThread);
```
amulware: Why this scheduler? Won't that force the state machine to run on whatever thread runs this code? That sounds bad. I figure it should just use the generic default scheduler instead.

Toggl.Foundation/TogglSyncService.cs
simonrozsival: We could refactor it later, this is quite verbose. On the other hand this is not something you would write very often and is not that important. But that is up to you.

Toggl.PrimeRadiant.Realm/Repository.cs
simonrozsival: The first thing didn't work. I didn't investigate this very much yesterday. I can check later if it is unnecessary, when the whole thing works.

Toggl.Foundation/DataSources/TogglDataSource.cs
simonrozsival: As I wrote in the description of the PR, some help here would be much appreciated.

Toggl.PrimeRadiant.Realm/Repository.cs
simonrozsival: It really doesn't work without this change.

simonrozsival: The problem is that we don’t ever subscribe to the observable returned by the Repository.BatchUpdate, so the func is never executed.

amulware: Do we want the BatchUpdate observable to be hot though? that means that we may or may not get the returned values when we do want them. What is the desired behaviour here?

simonrozsival: I would argue that you want to do an update ASAP. But if we decide to leave this as a cold observable, we can connect to it in the BasePersistState (it is in fact a side effect there, the return value is now thrown away).

simonrozsival:
This is now ready for a review. The app has to be restarted after synchronization is complete to show the persisted data in the UI.

Toggl.Foundation/Sync/States/BasePersistState.cs
```diff
@@ -25,7 +25,7 @@ public IObservable<ITransition> Start(FetchObservables fetch)
             return FetchObservable(fetch)
                 .SingleAsync()
                 .Select(entities => entities.Select(ConvertToDatabaseEntity).ToList())
-                .Do(databaseEntities => BatchUpdate(database, databaseEntities))
+                .Do(databaseEntities => BatchUpdate(database, databaseEntities).Wait())
```
amulware: Why wait sync with `.Wait()` and not async with `await`?

Toggl.Foundation/Sync/States/FetchAllSinceState.cs
```diff
@@ -20,7 +21,8 @@ public FetchAllSinceState(ITogglDatabase database, ITogglApi api)
 
         public IObservable<ITransition> Start() => Observable.Create<ITransition>(observer =>
         {
-            var sinceDates = database.SinceParameters.Get();
+            var realmSinceDates = database.SinceParameters.Get();
```
amulware: Don't include reference to realm. Foundation does not know what database is being used.

PR #493, opened 2017-09-01
Project suggestions button

willsb:
Ref #444 

This allows the usage of the Projects Suggestion button. Fiddle around with it on the simulator, just to make sure it’s all behaving. Tests have it covered, but just in case®

Squash whenever


willsb:
> Should i be able to turn off project suggesting by pressing the project button? 

Yes

> [...] should the suggestion list be emptied then? 

Depends. It will rerun the query, so if there's something in the text box, it will search for time entries.

> [...] should the project button turn gray after i turn it off?

Yes

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
```diff
@@ -138,6 +156,19 @@ private void OnTextFieldInfoChanged()
 
             await navigationService.Close(this);
         }
+
+        private void removeProjectQueryFromDescriptionIfNeeded()
+        {
+            var indexOfProjectQuerySymbol = TextFieldInfo.Text.IndexOf(projectQuerySymbol);
```
simonrozsival: I think the project query symbol should be the very _last_ "@" in the query, not the first one. When I type in "Send email to some@one.com @emails" and then I call this method, I would expect the result to be "Send email to some@one.com ", but with this implementation I get "Send email to some". Also shouldn't it check, that there is a _space_ before the `projectQuerySymbol`?

willsb: We ditched this space thing on slack. Also, you approach doesn't allow for the @ symbol on project queries, so something is also lost. This PR didn't even change the implementation of the query parser, so this shouldn't be subject of reviews

simonrozsival: OK

PR #494, opened 2017-09-02
Api deserialization error

willsb:
This will help us debug. A lot.

Squash whenever


Toggl.Ultrawave/ApiClients/BaseApi.cs
```diff
@@ -64,9 +65,16 @@ protected IObservable<T> CreateObservable<T>(Endpoint endpoint, IEnumerable<Http
...
-                    observer.OnCompleted();
+                    try
+                    {
+                        var data = await Task.Run(() => serializer.Deserialize<T>(response.RawData)).ConfigureAwait(false);
+                        observer.OnNext(data);
+                        observer.OnCompleted();
+                    }
+                    catch
+                    {
+                        observer.OnError(new DeserializationException<T>(response.RawData));
```
simonrozsival: We loose some information here - I would prefer passing the caught exception to the thrown exception.

Toggl.Ultrawave/Exceptions/DeserializationException.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System;
+namespace Toggl.Ultrawave.Exceptions
+{
+    public sealed class DeserializationException<T> : ApiException
+    {
+        private const string errorMessage = "An error occured while trying to deserializing type {0}. Check the Json property for more info.";
```
amulware: Grammar: *trying to deserializ**e**~~ing~~

Toggl.Ultrawave/ApiClients/BaseApi.cs
amulware: Agreed. Also, we should not include the full json string here, or at least not in release builds. This may be a lot of data for fetch endpoints, and will contain sensitive user data which should not show up in logs.

willsb: Passing the Exception does not help at all, since Newtonsoft.Json throws some pretty cryptic exceptions. Regarding logging, you do know that OnError does not log, right? Who logs what depends only on what we decide to log and we should simply decide not to log this exception for release builds.

Toggl.Ultrawave/Exceptions/DeserializationException.cs
willsb: Derp

Toggl.Ultrawave/ApiClients/BaseApi.cs
amulware: Yet don't we have to log _something_ if serialization fails for release?

I can very much see one of us adding an rx extension that logs all errors and copy paste it everywhere. Dangerous. And I better not catch that person logging user details :P

willsb: There's no Copy pasta in the world that would screw this up. I am using a specifically named property here. The person would have to log the `Json` property and that wouldn't pass review. If Simon really wants to I can wrap the original (useless) Newtonsoft exception, but I also don't see the need to do that

amulware: But you are including RawData in the exception, if that exception is logged, would it not log that data, i.e. the json of the user's data?

willsb: Raw data is not included in the base Exception, it's only used by the Json property. The base Exception gets only the type that failed to deserialize.

amulware: I see. Maybe I'll remember to read the code first next time. The use looked a lot like a message parameter to me.

PR #499, opened 2017-09-04
Reposiory replace operation

simonrozsival:
Adds a new operation to the `IRepository` which is useful for replacing one entity with another - i.e., an update when the ID of the entity changes as well. I want to use this for persisting data after `POST` requests.


Toggl.PrimeRadiant.Tests/RepositoryTests.cs
```diff
@@ -64,5 +67,42 @@ await Enumerable.Range(0, numberOfItems)
...
             entities.Count().Should().Be(numberOfItems);
         }
+
+        [Fact]
+        public void TheReplaceMethodThrowsIfThereIsNoEntityWithThatIdInTheRepository()
+        {
+            Func<Task> callingReplaceInAnEmptyRepository =
+                async () => await Replace(new TTestModel(), new TTestModel());
+
+            callingReplaceInAnEmptyRepository
```
amulware: As this variable name shows, this test tests empty, not 'no entity with that id'. Why not have two tests?

Toggl.PrimeRadiant.Tests/RepositoryTests.cs
```diff
@@ -64,5 +67,42 @@ await Enumerable.Range(0, numberOfItems)
...
+        {
+            var oldTestEntity = GetModelWith(123);
+            var nextTestEntity = GetModelWith(456);
+            await Create(oldTestEntity);
+
+            await Replace(oldTestEntity, nextTestEntity).SingleAsync();
+            Func<Task> callingReplaceForTheSecondTime =
+                async () => await Replace(oldTestEntity, nextTestEntity);
+
+            callingReplaceForTheSecondTime
```
amulware: Testing the method with a call to the method is a bit cringe-worthy, unless you mean to test calling-twice behaviour.

simonrozsival:
After internal discussion, I will close this and send a different PR instead.

PR #500, opened 2017-09-04
Change repository update and delete to use Ids explicitely

simonrozsival:
After internal discussion - avoid using `Replace` operation.


Toggl.PrimeRadiant.Realm/BaseStorage.cs
```diff
@@ -21,22 +21,21 @@ protected BaseStorage(IRealmAdapter<TModel> adapter)
         }
 
         public IObservable<TModel> Update(TModel entity)
+            => Update(entity.Id, entity);
```
amulware: Consider making this method an extension on IRepository inside foundation to keep the database interface lighter.

Toggl.PrimeRadiant.Tests/BaseStorageTests.cs
```diff
@@ -30,18 +30,18 @@ public void TheUpdateMethodThrowsIfThereIsNoEntityWithThatIdInTheRepository()
         [Fact]
         public async Task TheUpdateMethodAlwaysReturnsASingleElement()
         {
-            var testEntity = new TTestModel();
+            var testEntity = GetModelWith(12345);
             await Create(testEntity);
 
-            var element = await Update(testEntity).SingleAsync();
+            var element = await Update(12345, testEntity).SingleAsync();
```
amulware: Use property of the model of constant instead of repeating literal.

Toggl.PrimeRadiant.Tests/Realm/TestAdapter.cs
```diff
@@ -21,8 +21,11 @@ public TestModel Create(TestModel entity)
         }
 
         public TestModel Update(TestModel entity)
+            => Update(entity.Id, entity);
```
amulware: I don't think this method is used anywhere, is it?

Toggl.PrimeRadiant.Tests/RepositoryTests.cs
```diff
@@ -64,5 +64,20 @@ await Enumerable.Range(0, numberOfItems)
...
+            var oldTestEntity = GetModelWith(123);
+            var nextTestEntity = GetModelWith(456);
+            await Create(oldTestEntity);
+
+            await Update(oldTestEntity.Id, nextTestEntity).SingleAsync();
+            Func<Task> callingReplaceForTheSecondTime =
+                async () => await Repository.GetById(123);
+
+            callingReplaceForTheSecondTime
+                .ShouldThrow<EntityNotFoundException>();
```
amulware: Please test that the id was actually changed properly, instead of inferring it with this second call. Instead, consider having a separate test about what happens when calling twice but changing id with the first call.

PR #501, opened 2017-09-04
Unsyncable models

simonrozsival:


amulware:
I like the enum, and having a property that makes consuming it easier is fine for me (though only in foundation then), though we may wanna rename it to something like "HasUnsyncedChanges" to not make it confusing? (Though I can see this being confusing in either case...)

More comments welcome @willsb @arturs105

amulware:
Actually, maybe using more expressive state names, compared to `Clean` and `Dirty` might help too?

`Synced`
`NeedsSyncing`

?

Toggl.Foundation/Models/FoundationModels.cs
```diff
@@ -19,7 +20,9 @@ internal partial class Client : IDatabaseClient
 
         public IDatabaseWorkspace Workspace { get; }
 
-        public bool IsDirty { get; }
+        public SyncStatus SyncStatus { get; }
+
+        public bool IsDirty => SyncStatus != SyncStatus.InSync;
```
willsb: Any reason why this can't be an extension method instead?

willsb: Looking at the PR, it doesn't seem like we are even using it anywhere. For UI, we can have value converters. Otherwise, using an extension method seems like a smarter approach than copy pasta-ing the definition of IsDirty in every implementation

amulware: Sounds good to me. The extension should then also be on the IDatabaseSyncable interface. Much nicer. Or if not needed, we can just not have it. Even better.

simonrozsival: I will try to remove it tomorrow morning, otherwise I like the idea of extension method on `IDatabaseSyncable`

PR #503, opened 2017-09-05
Syncable repository

simonrozsival:
This repository extends the basic one with the functionality of returning the entities, which should be synced with the server.


amulware:
Why do this in a new repository, instead of just having a query on the existing one?

simonrozsival:
I wanted to stick to the open/closed principle (SOLID). It seems like a good idea here - I also didn't want to make the `IRepository` interface leaky again (@willsb).

amulware:
I'm not sure I understand.

Why cant:
```
.Where(needsSync).OrderBy(lastChangeSelector).FirstOrDefault()
```

Be run on the existing repository?

amulware:
Shouldn't that... just work?

amulware:
(Assuming we use expressions instead of funcs, which it should do anyways.)

simonrozsival:
After internal discussion we decided we don't need this.

PR #504, opened 2017-09-05
Push time entries state

### LINKED ISSUE #497: Sync state: Find oldest time entry that needs syncing
Should be as simple as running a query that returns the oldest entry that is dirty, and does not have the sync-failed flag set.

This state will have multiple results, at least that it found or didn't find such a time entry. Potentially we also want to distinguish whether to push or put the time entry, though this is maybe better done in the follow-up state to keep things simple.

simonrozsival:
Closes #497

This is meant as the first state of the push pipeline and will be used for looping until there is no entity to be synced. The base state is generic and has generic ✅  too. For now there is only concrete implementation for time entries.


Toggl.Foundation/Sync/States/Push/PushTimeEntriesState.cs
```diff
@@ -0,0 +1,23 @@
...
+            : base(database)
+        {
+        }
+
+        protected override IObservable<IDatabaseTimeEntry> GetOldestUnsynced(ITogglDatabase database)
+            => database.TimeEntries.GetAll(entity => entity.SyncStatus == SyncStatus.SyncNeeded)
+                .SingleAsync()
+                .Select(entities => entities
+                    .OrderBy(entity => entity.At)
+                    .FirstOrDefault());
```
amulware: Can't most of this query be in the base class?

Toggl.Foundation.Tests/Sync/States/Push/BasePushStateTests.cs
```diff
@@ -0,0 +1,137 @@
...
+                Action callingStart = () => state.Start().SingleAsync().Wait();
+
+                callingStart.ShouldThrow<Exception>();
+            }
+
+            protected abstract BasePushState<TModel> CreateState(ITogglDatabase database);
+
+            protected abstract void SetupRepositoryToReturn(ITogglDatabase database, TModel[] entities);
+
+            protected abstract void SetupRepositoryToReturnMoreThanOneEnumerableOfEnties(ITogglDatabase database);
```
amulware: Typo.

Toggl.Foundation.Tests/Sync/States/Push/BasePushStateTests.cs
```diff
@@ -0,0 +1,137 @@
...
+                SetupRepositoryToReturn(database, new[] { entity, entity2, entity3 });
+
+                var transition = state.Start().SingleAsync().Wait();
+                var parameter = ((Transition<TModel>)transition).Parameter;
+
+                transition.Result.Should().Be(state.PushEntity);
+                parameter.Should().Be(entity2);
+            }
+
+            public void ThrowsWhenRepositoryReturnsMultipleEntitiesToSync()
```
amulware: But.. the repo will often return multiple. Is this just bad naming?

simonrozsival: Bad naming.

PR #505, opened 2017-09-05
Push one time entry

simonrozsival:
Ref #497 

This state is meant to be the second in the pipeline for pushing time entries. It decides whether how to process a single time entry (POST, PUT, DELETE, just remove from the local storage without any HTTP request).

It is implemented generally with generics and generic tests, so it will be easily applicable to the other types of entities.


Toggl.Foundation/Sync/States/Push/PushTimeEntryState.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System;
+using Toggl.PrimeRadiant.Models;
+
+namespace Toggl.Foundation.Sync.States
+{
+    internal sealed class PushTimeEntryState : BasePushEntityState<IDatabaseTimeEntry>
```
amulware: What is the point of this class? The base class should have access to both of the properties, right?

Toggl.Foundation.Tests/Sync/States/Push/BasePushEntityStateTests.cs
```diff
@@ -0,0 +1,105 @@
...
+        [Fact]
+        public void ReturnsCreateTransition()
+            => helper.ReturnsCreateTransition();
+
+        [Fact]
+        public void ReturnsUpdateTransition()
+            => helper.ReturnsUpdateTransition();
+
+        [Fact]
+        public void ReturnsDeleteTransition()
```
amulware: Under what condition? (applies to all tests)

Also, this is missing a test for the fourth state result, and what happens if null is passed, no?

Toggl.Foundation/Sync/States/Push/BasePushEntityState.cs
```diff
@@ -0,0 +1,47 @@
...
+                .Select(entity =>
+                    IsDeleted(entity)
+                        ? WasNotPublished(entity)
+                            ? deleteLocally(entity)
+                            : delete(entity)
+                        : WasNotPublished(entity)
+                            ? create(entity)
+                            : update(entity));
+
+        private IObservable<TModel> createObservable(TModel entity)
```
amulware: Not sure if this method is really necessary but alright.

Toggl.Foundation/Sync/States/Push/PushedEntity.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System;
+using Toggl.Multivac.Models;
+using Toggl.PrimeRadiant;
+
+namespace Toggl.Foundation.Sync.States
+{
+    internal sealed class PushedEntity<TModel>
```
amulware: Not sure what this class is for. The second constructor isn't even used, and one of the properties is always null.

Toggl.Foundation/Sync/States/Push/PushTimeEntryState.cs
simonrozsival: It does not have access to `.IsDeleted`. It is true that I can move `WasNotPublished` to the base class.

Toggl.Foundation/Sync/States/Push/PushedEntity.cs
simonrozsival: The second constructor will be used in the following state. The class holds the original entity and then the deserialized one to detect if it has changed meanwhile the HTTP request is processed.

Toggl.Foundation/Sync/States/Push/BasePushEntityState.cs
simonrozsival: I wanted to put in the not null check, just for sure. If we decided that we don't need the check, then this method would be indeed useless.

Toggl.Foundation.Tests/Sync/States/Push/BasePushEntityStateTests.cs
simonrozsival: You are right, I forgot to add one `[Fact]`. I will also add the test and think of better naming.

simonrozsival:
I think this is actually a nice way to branch the states. We can then have three other simple tests for the individual operations (assuming that delete locally will be somehow similar to the normal Delete state - just skipping the HTTP request.)

Toggl.Foundation/Sync/States/Push/PushTimeEntryState.cs
amulware: So only time entries have that property? It seems to me it should be part of the IDatabaseSyncable. Would that not make sense? (Even if we can't delete other entities anytime soon, for consistency and future proofing.)

Toggl.Foundation/Sync/States/Push/BasePushEntityState.cs
amulware: I'm fine with the check, though in that case, it should probably return an observable that errors, not throw...?

Toggl.Foundation/Sync/States/Push/PushedEntity.cs
amulware: But it's not needed here then, is it? It would only need to be used with the next state. And even if we have this decision state be its own state (still not 100% convinved, but lets see), the http request and following database update really should be a single state anyways I think, since they represent a single operation that should not be split, or cancelled part way through.

PR #506, opened 2017-09-06
Project Selection

### LINKED ISSUE #444: Project auto completion in time entry creation dialog

willsb:
Closes #444 

This PR does a lot, but everything is so nested together that splitting it would make reviewing too confusing.

TextFieldInfo now has properties to deal with selected project (in the future, it can easily have a list of tags)

I’m using UITextView instead of UITextField because iOS has two classes for doing the exact same thing, only UITextView can have multiple lines (and for some reason doesn’t have a placeholder, we’ll have to see about that in different PR).

Project suggestion is pretty straightforward.

This PR only adds colored text as the Project Token, with no border. This is because there’s a bug in Xamarin that makes it impossible to fetch the attributed string attributes in the View’s Draw method. We also need to see how we’ll deal with projects that spam more than 1 line. I’ll open a different issue to deal with that so we can at least have the core functionality rolling.

A functional text is not optional.

Here's what it looks like right now:


Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntrySuggestions/TextFieldInfo.cs
```diff
@@ -24,9 +44,20 @@ public override bool Equals(object obj)
             => obj is TextFieldInfo && Equals((TextFieldInfo)obj);
 
         public override int GetHashCode()
-            => (Text + CursorPosition).GetHashCode();
+        {
+            unchecked
```
arturs105: Without the `unchecked` integer overflow would result in run time error?

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntrySuggestions/TextFieldInfo.cs
```diff
@@ -24,9 +44,20 @@ public override bool Equals(object obj)
...
+            {
+                return (Text.GetHashCode() * 397) ^
```
arturs105: Why 397?

Toggl.Daneel/Extensions/TextExtensions.cs
```diff
@@ -0,0 +1,71 @@
+﻿using System;
+using System.Globalization;
+using Foundation;
+using MvvmCross.Platform.UI;
+using MvvmCross.Plugins.Color;
+using MvvmCross.Plugins.Color.iOS;
+using Toggl.Daneel.Views;
```
arturs105: 1 and 7 🚫

Toggl.Daneel/Extensions/TextExtensions.cs
```diff
@@ -0,0 +1,71 @@
...
+using Toggl.Foundation.MvvmCross.ViewModels.StartTimeEntrySuggestions;
+using UIKit;
+
+namespace Toggl.Daneel.Extensions
+{
+    public static class TextExtensions
+    {
+        private static readonly NSParagraphStyle paragraphStyle;
+        private static readonly MvxRGBValueConverter converter = new MvxRGBValueConverter();
+        private static readonly UIColor strokeColor = Color.StartTimeEntry.SeparatorColor.ToNativeColor();
```
arturs105: I think this should be a different color (not `Color.StartTimeEntry.SeparatorColor`).

Toggl.Daneel/Extensions/TextExtensions.cs
```diff
@@ -0,0 +1,71 @@
...
+using MvvmCross.Plugins.Color;
+using MvvmCross.Plugins.Color.iOS;
+using Toggl.Daneel.Views;
+using Toggl.Foundation.MvvmCross.Helper;
+using Toggl.Foundation.MvvmCross.ViewModels.StartTimeEntrySuggestions;
+using UIKit;
+
+namespace Toggl.Daneel.Extensions
+{
+    public static class TextExtensions
```
arturs105: This class could also have a method to add the project dot (from the project task client combiner), but that's for another PR.

Toggl.Daneel/Extensions/TextExtensions.cs
```diff
@@ -0,0 +1,71 @@
...
+        public static string GetDescription(this UITextView self, TextFieldInfo info)
+        {
+            if (string.IsNullOrEmpty(info.ProjectName))
+                return self.Text;
+
+            return self.Text.Replace(info.PaddedProjectName(), "");
+        }
+
+        public static string PaddedProjectName(this TextFieldInfo self)
+            => string.IsNullOrEmpty(self.ProjectName) ? self.ProjectName : $" {self.ProjectName} ";
```
arturs105: `string.IsNullOrEmpty(self.ProjectName) ? "" : $" {self.ProjectName} ";` Shorter this way. Also - why not have this as a property on `TextFieldInfo`?

Toggl.Daneel/Extensions/TextExtensions.cs
```diff
@@ -0,0 +1,71 @@
...
+        public static NSAttributedString GetAttributedText(this TextFieldInfo self)
+        {
+            var projectName = self.PaddedProjectName();
+            var fullText = $"{self.Text}{projectName}";
+            var result = new NSMutableAttributedString(fullText);
+
+            result.AddAttributes(new UIStringAttributes
+            {
+                BaselineOffset = 5,
+                StrokeColor = UIColor.Clear,
```
arturs105: Isn't the stroke color clear by default?

Toggl.Daneel/Extensions/TextExtensions.cs
```diff
@@ -0,0 +1,71 @@
...
+                ParagraphStyle = paragraphStyle,
+                Font = UIFont.SystemFontOfSize(16, UIFontWeight.Regular),
+            }, new NSRange(0, self.Text.Length));
+
+            if (string.IsNullOrEmpty(self.ProjectColor))
+                return result;
+
+            var hexColor = self.ProjectColor;
+            var color = (UIColor)converter.Convert(hexColor, typeof(MvxColor), null, CultureInfo.CurrentCulture);
```
arturs105: Isn't target type `UIColor`?

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
```diff
@@ -89,7 +89,18 @@ private void selectSuggestion(BaseTimeEntrySuggestionViewModel suggestion)
                     var description = timeEntrySuggestion.Description;
 
                     ProjectId = timeEntrySuggestion.ProjectId;
-                    TextFieldInfo = new TextFieldInfo(description, description.Length);
+                    TextFieldInfo = new TextFieldInfo(description, description.Length,
```
arturs105: I think this would be more readable:

```
TextFieldInfo = new TextFieldInfo(
    description,
    description.Length, 
    timeEntrySuggestion.ProjectName,
    timeEntrySuggestion.ProjectColor
);
```

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntrySuggestions/TextFieldInfo.cs
willsb: Yes

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntrySuggestions/TextFieldInfo.cs
willsb: https://stackoverflow.com/a/102878/3465182

Toggl.Daneel/Extensions/TextExtensions.cs
willsb: Agreed. I think it should even contain the `NSAttributedString` used by the card that displays the currently running Time Entry

Toggl.Daneel/Extensions/TextExtensions.cs
willsb: Because this is something iOS specific, might not be needed in every platform

Toggl.Daneel/Extensions/TextExtensions.cs
willsb: This doesn't matter, really. It'll be refactored when MvvmCross 5.2 is out. Also, it was ©️ 🍝  from other place in the app

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
willsb: Agreed

Issue #498, opened 2017-09-04
Sync state: push time entry


amulware:
This can mean either POST or PUT, which might behave differently, so we should see if one or two states for this make more sense, and how to exactly wire it up.

This state should do both the push and the persistence of the result since that is sort of an 'atomic operation'.

To keep things simple, we decided that for now, we will always override local data from a POST.

From a PUT however, we will only override local data, if that data has not been changed since the push was started (as simple as comparing the two local AT dates I imagine).

PR #508, opened 2017-09-06
Sync: Create new time entry state (POST)

simonrozsival:
Ref #498 

This PR contains a state for creating a new time entry (HTTP POST) and then persists the data received from the server to the local database.


Toggl.Foundation/Sync/States/Push/BaseCreateEntityState.cs
```diff
@@ -0,0 +1,43 @@
...
+
+namespace Toggl.Foundation.Sync.States
+{
+    internal abstract class BaseCreateEntityState<TModel>
+        where TModel : class, IBaseModel, IDatabaseSyncable
+    {
+        private readonly ITogglApi api;
+        private readonly ITogglDatabase database;
+
+        public StateResult<TModel> CreatingFailed { get; } = new StateResult<TModel>();
```
amulware: I'm fine doing this later, but this has to pass on the exception as well, especially for API exceptions.

Toggl.Foundation.Tests/Sync/States/Push/BaseCreateEntityStateTests.cs
```diff
@@ -0,0 +1,129 @@
...
+        public void ReturnsFailTransitionWhenHttpFails()
+            => helper.ReturnsFailTransitionWhenHttpFails();
+
+        [Fact]
+        public void ReturnsFailTransitionWhenDatabaseOperationFails()
+            => helper.ReturnsFailTransitionWhenDatabaseOperationFails();
+
+        [Fact]
+        public void ReturnsSuccessfulTransitionWhenEveyrthingWorks()
+            => helper.ReturnsSuccessfulTransitionWhenEveyrthingWorks();
```
amulware: Typos

Toggl.Foundation/Sync/States/Push/BaseCreateEntityState.cs
```diff
@@ -12,7 +12,7 @@ internal abstract class BaseCreateEntityState<TModel>
         private readonly ITogglApi api;
         private readonly ITogglDatabase database;
 
-        public StateResult<TModel> CreatingFailed { get; } = new StateResult<TModel>();
+        public StateResult<PushFailure<TModel>> CreatingFailed { get; } = new StateResult<PushFailure<TModel>>();
```
amulware: Why not use a tuple?

amulware:
Comment shows as outdated because I made it on a commit, even though it is still relevant. Github, you silly.

Toggl.Foundation/Sync/States/Push/BaseCreateEntityState.cs
simonrozsival: I first did this with tuples and then decided to make it a separate class to make it more convenient for sharing between the states. I like this second approach more than using tuples when the states are so loosely coupled.

Issue #37, opened 2017-04-14
Write Api integration tests


amulware:
We should test all our assumptions about how the Api works, which means testing all endpoints, for all successful scenarios, as well as all imaginable failure scenarios.

As much as possible, we should test models separately:

- [x] User
- [x] Workspace
- [x] Client
- [x] Project
- [ ] Task
- [x] Tag
- [x] Time entry

We also should test the various workspace settings, time entry constraints, relation between the models, and more.

Issue #34, opened 2017-04-14
Create Api clients


amulware:
We need at least:

- [x] User
- [x] Workspace
- [x] Client
- [x] Project
- [ ] Task
- [x] Tag
- [x] Time Entry

These should be implemented separately and tested appropriately.

arturs105:
Project and Task API clients have been implemented [here](https://github.com/toggl/mobileapp/tree/projects-api) and [here](https://github.com/toggl/mobileapp/tree/tasks-api), however there are some issues with testing them related to premium features. When #207 is resolved, we should be able to use them properly.

Issue #29, opened 2017-04-14
Create Api Endpoints


amulware:
There is a number of endpoints we have to implement.
The list below is mostly complete (for our purposes here), but may change over time, and also has some inconsistencies, as well as gaps, and a bunch of question marks as to which exactly are needed.

I will improve this list in the future as our specs solidify.

These should be implemented in separate PRs. This issue serves merely to have an overview and status report.

**On the HttpMethod**

Note that the different HTTP methods mostly follow the following pattern:
- GET: retrieve a list of objects, or a single one if an id is given
- POST: create a new object
- PUT: update an existing object with the given id
- DELETE: delete an existing object with the given id

Workspace
- [x] GET /me/workspaces
- [ ] POST /workspaces (needed soon(tm))
- [x] GET /workspaces/{WID} (is this needed?)
- [ ] ~~PUT /workspaces/{WID} (is this needed?)~~
- [ ] GET /workspaces/{WID}/preferences
- [ ] ~~POST /workspaces/{WID}/preferences (probably not needed)~~
- [ ] GET /workspaces/{WID}/preferences/time_entry_constraints
- [ ] ~~POST/PUT /workspaces/{WID}/preferences/time_entry_constraints (probably not needed)~~

Client
- [x] GET /me/clients
- [x] POST /workspaces/{WID}/clients
- [ ] ~~PUT /workspaces/{WID}/clients/{CID} (is this needed?)~~

Projects
- [x] GET /me/projects
- [x] POST /workspaces/{WID}/projects
- [ ] ~~GET /workspaces/{WID}/projects/{PID} (is this needed?)~~
- [ ] ~~PUT /workspaces/{WID}/projects/{PID} (is this needed?)~~
- [ ] ~~DELETE /workspaces/{WID}/projects/{PID} (is this needed?)~~

Tasks
- [x] GET /me/tasks
- [x] POST /workspaces/{workspace_id}/projects/{project_id}/tasks

Time Entries
- [x] GET /me/time_entries
- [x] POST /workspaces/{WID}/time_entries
- [x] PUT /workspaces/{WID}/time_entries/{TID}
- [x] DELETE /workspaces/{WID}/time_entries/{TID}

Tags
- [x] GET /me/tags
- [x] POST /workspaces/{WID}/tags

Workspace Features
- [x] GET /me/features
- [ ] ~~GET /workspaces/{workspace_id}/subscription/features (probably not needed)~~

amulware:
I updated the post above:

[...]

PR #535, opened 2017-09-07
Update time entries

simonrozsival:
Ref #29, #34, #37 

Allow PUT HTTP requests for time requests. Run integration tests when doing review.

~~WIP: An empty list of tag ids is considered invalid by the API. This must be solved ~~either~~ by fixing the API ~~or by preprocessing the data before serialization~~.~~

Edit: There is a test which is skipped due to a bug in the API. It should be added back when the API bug is fixed in a separate PR.


amulware:
This looks fine to me, and I'd like it merged. There is a general question we have to answer first though:

If the API has a bug, do we write our code and tests assuming it worked correctly? (Once the app is in user's hands, this may not be practical.)

Or do we write tests testing the broken behaviour, until it is fixed, and correct the tests then?

I don't like us having to wait for API bugs to be fixed before we can resume work and one way or another I would like our tests to represent the bug (either by failing or by testing for the bug's existence) so nothing falls through the cracks.

willsb:
We write tests for both behaviors and then we `Fact(Skip="Api has ze bugz, running that other test instead")` in the test that has the appropriate behavior. This makes it very easy to track the current state of things, IMO.

simonrozsival:
This should be resolved quite soon, I reported it in the morning and it is already being processed: https://github.com/toggl/backend/issues/5306

I can add a test, which adds a tag and adds it to the time entry, which will work right now.

And I don't think we should skip this test - it is an integration test and it doesn't block merging.

amulware:
I prefer doing what @willsb suggests. But if this will indeed be fixed soon, we don't have to do it here. 

For anything that stays around for longer though, having it pass until backend changes, and then being able to make a very simple PR removing the skip and bug-test sounds pretty good to me.

willsb:
What Paul said, but:

> [...] I don't think we should skip this test - it is an integration test and it doesn't block merging.

While it doesn't block merging, our currently CI-flow runs those tests a few times a day, making false red lights quite annoying

amulware:
I don't even care about annoying, but if red lights are gonna be a common thing we will start ignoring them (like we are currently ignoring the UI tests failing).

amulware:
As a consequence, we would not notice other things breaking, making the tests pointless.

simonrozsival:
Good point. I will skip the test.

amulware:
But is there now a test that will fail once the API fix goes live?

That is much more important than the skip.

simonrozsival:
Now there is the test.

simonrozsival:
Now this PR is ready for reviews after a bug in tests was removed. I should also mention that I had to also "skip" the tests that are inherited from the base class. I didn't know how to do that in an elegant way, so I just commented out some pieces of code and added two TODO comments so we don't forget about this once we add the skipped fact back. I suppose we will get back to this in the near future, so I think this might be an OK thing to do for now. If anyone has a better idea how to skip those tests without much effort, I will gladly do it in a different way.

Toggl.Ultrawave.Tests.Integration/TimeEntriesApiTests.cs
```diff
@@ -103,6 +103,136 @@ private IObservable<ITimeEntry> CallEndpointWith(ITogglApi togglApi, TimeEntry c
...
+                updatedTimeEntry.Id.Should().Be(persistedTimeEntry.Id);
+                updatedTimeEntry.Description.Should().Be(timeEntryWithUpdates.Description);
+                updatedTimeEntry.WorkspaceId.Should().Be(persistedTimeEntry.WorkspaceId);
+                updatedTimeEntry.Billable.Should().Be(false);
+                updatedTimeEntry.ProjectId.Should().BeNull();
+                updatedTimeEntry.TaskId.Should().BeNull();
+            }
+
+            [Fact, LogTestInfo]
+            public async Task UpdatingFailsDueToApiBug()
```
amulware: Let's be more specific? `UpdatingFailsWhenWorkspaceHasNoTags` or something?

Toggl.Ultrawave.Tests.Integration/TimeEntriesApiTests.cs
```diff
@@ -103,6 +103,136 @@ private IObservable<ITimeEntry> CallEndpointWith(ITogglApi togglApi, TimeEntry c
...
+                        Billable = persistedTimeEntry.Billable,
+                        Start = persistedTimeEntry.Start,
+                        Stop = persistedTimeEntry.Stop,
+                        TagIds = persistedTimeEntry.TagIds,
+                        UserId = persistedTimeEntry.UserId,
+                        CreatedWith = persistedTimeEntry.CreatedWith
+                    };
+
+                    // return CallEndpointWith(togglApi, timeEntryWithUpdates);
+                    return Observable.Return<TimeEntry>(null); // TODO: remove when the
```
amulware: incomplete comment?

amulware: And can this method not create a tag, to make sure the regular call works?

Toggl.Ultrawave.Tests.Integration/TimeEntriesApiTests.cs
```diff
@@ -103,6 +103,136 @@ private IObservable<ITimeEntry> CallEndpointWith(ITogglApi togglApi, TimeEntry c
                 => togglApi.TimeEntries.Create(client);
         }
 
+        public sealed class TheUpdateMethod : AuthenticatedPutEndpointBaseTests<ITimeEntry>
+        {
+            [Fact(Skip = "A bug in staging API"), LogTestInfo] // TODO: when removing this test skip, also fix "CallEndpointWith"
```
amulware: I think we should be more specific in the skip text.

willsb:
I am okay with using a `//TODO` for remind us later, yes

Toggl.Ultrawave.Tests.Integration/TimeEntriesApiTests.cs
simonrozsival: I am not sure what you mean. This method does not create any tag.

amulware: But if it _did_ create a tag we wouldn't have to return a dummy observable, right? Because the endpoint would actually work?

simonrozsival: There is a test `AddTagsToAnExistingTimeEntry` for that. I don't think this basic test should do any extra actions. I still think this is a very temporary hack.

PR #537, opened 2017-09-07
Add edit duration view

### LINKED ISSUE #446: Duration edit screen

arturs105:
Closes #446 

This is just the UI - the value is not returned to edit view and is not persisted. For that we need the new version of MvvmCross, so it will come a bit later.

There are be some issues with start and end times not displaying the correct value due to time zones. That will be fixed in another PR.

The times are currently displayed only in 24h format. Displaying in user’s preferred format will come in a later PR.


Toggl.Daneel/Converters/TimeSpanToViewHeightConverter.cs
```diff
@@ -0,0 +1,41 @@
...
+        {
+            this.minViewHeight = minViewHeight;
+            this.maxViewHeight = maxViewHeight;
+
+            functionSlope = (nfloat)((maxViewHeight - minViewHeight) / (MaxDuration.TotalSeconds - minDuration.TotalSeconds));
+        }
+        
+        protected override nfloat Convert(TimeSpan value, Type targetType, object parameter, CultureInfo culture)
+        {
+            //y = mx + b
```
willsb: This comment doesn't seem to help much

Toggl.Daneel/Converters/TimeSpanToViewHeightConverter.cs
```diff
@@ -0,0 +1,41 @@
...
+            var returnValue = functionSlope * value.TotalSeconds + minViewHeight;
+            return (nfloat)returnValue;
+        }
+
+        protected override TimeSpan ConvertBack(nfloat value, Type targetType, object parameter, CultureInfo culture)
+        {
+            //x = (y - b) / m
+            var seconds = (value - minViewHeight) / functionSlope;
+            var rounding = 5 * 60; //Round to 5min
```
willsb: Should be a constant instead

Toggl.Daneel/ViewControllers/EditDurationViewController.cs
```diff
@@ -0,0 +1,97 @@
...
+namespace Toggl.Daneel.ViewControllers
+{
+    [ModalDialogPresentation]
+    public sealed partial class EditDurationViewController : MvxViewController<EditDurationViewModel>
+    {
+        private const int maxDurationViewHeight = 424;
+        private const int minDurationViewHeight = 72;
+
+        private readonly TimeSpanToViewHeightConverter testConverter
+                = new TimeSpanToViewHeightConverter(minDurationViewHeight, maxDurationViewHeight);
```
willsb: Overindenting

Toggl.Daneel/ViewControllers/EditDurationViewController.cs
```diff
@@ -0,0 +1,97 @@
...
+                Height = 480
+            };
+
+            EndTimeLabel.Font = EndTimeLabel.Font.GetMonospacedDigitFont();
+            DurationLabel.Font = DurationLabel.Font.GetMonospacedDigitFont();
+            StartTimeLabel.Font = StartTimeLabel.Font.GetMonospacedDigitFont();
+        }
+
+        nfloat previous;
+        private void onDurationViewDrag()
```
willsb: I'd encapsulate this behavior in a target binding instead

Toggl.Foundation.MvvmCross/Parameters/DurationParameter.cs
```diff
@@ -0,0 +1,14 @@
...
+
+namespace Toggl.Foundation.MvvmCross.Parameters
+{
+    public sealed class DurationParameter
+    {
+        public TimeSpan Duration { get; set; }
+
+        public bool IsRunning { get; set; }
+
+        public EditDurationBehaviourBase EditBehaviour { get; set; }
```
willsb: This won't be serialized properly

Toggl.Foundation.MvvmCross/ViewModels/EditDurationViewModel.cs
```diff
@@ -0,0 +1,67 @@
...
+        public override async Task Initialize(DurationParameter parameter)
+        {
+            await base.Initialize();
+
+            if (parameter.IsRunning)
+                timeService.CurrentDateTimeObservable
+                           .Subscribe(currentTime => EndTime = currentTime);
+
+            editBehavior = parameter.EditBehaviour;
+            editBehavior.Start.Subscribe(start => StartTime = start);
```
willsb: I think this edit behavior is a bit of over-engineering. I'd much rather pass something like a `FieldToEdit` enum in the `DurationParameter` and set the correct property according to that.

willsb: Even simpler would be checking if the end date exists. If it does, change it. Otherwise, change the Start date

Toggl.Daneel/ViewControllers/EditDurationViewController.cs
willsb: Just realized that you're using a converter here. This should *definitely* be a value converter

arturs105:
The binding is still kinda ugly (with the converter being used in `onDurationViewDrag`), but I can't seem to get the binding work in both ways (with `MvxBindingMode.TwoWay`) - the duration doesn't get updated, when the view's height changes - the converter's `ConvertBack` and `Convert` methods don't get invoked. Could this be because MvvmCross doesn't know, when the height is updated?

Any suggestions welcome.

Toggl.Daneel/Converters/TimeSpanToViewHeightConverter.cs
```diff
@@ -0,0 +1,40 @@
...
+using System.Globalization;
+using MvvmCross.Platform.Converters;
+
+namespace Toggl.Foundation.MvvmCross.Converters
+{
+    public sealed class TimeSpanToViewHeightConverter : MvxValueConverter<TimeSpan, nfloat>
+    {
+        private const int rounding = 5 * 60; //Round to 5min intervals
+
+        private readonly TimeSpan minDuration = new TimeSpan(0);
```
willsb: `TimeSpan.Zero` is clearer here

Toggl.Daneel/Views/ResizableView.cs
```diff
@@ -0,0 +1,69 @@
...
+        public event EventHandler HeightConstantChanged;
+
+        public nfloat HeightConstant
+        {
+            get => heightConstraint.Constant;
+            set
+            {
+                if (HeightConstant == value) return;
+                heightConstraint.Constant = value;
+                HeightConstantChanged?.Invoke(this, EventArgs.Empty);
```
willsb: Use the extension `HeightConstantChanged.Raise(this);`

Toggl.Daneel/Views/ResizableView.cs
```diff
@@ -0,0 +1,69 @@
...
+        {
+            heightConstraint = findOrCreateHeightConstraint();
+
+            panGestureRecognizer = new UIPanGestureRecognizer(onDrag);
+            AddGestureRecognizer(panGestureRecognizer );
+        }
+
+        private NSLayoutConstraint findOrCreateHeightConstraint()
+        {
+            foreach (var constraint in Constraints)
```
willsb: `Constraints.FirstOrDefault(c => c.FirstAttribute == NSLayoutAttribute.Height)`

Toggl.Daneel/Views/ResizableView.cs
```diff
@@ -0,0 +1,69 @@
...
+                NSLayoutAttribute.Height,
+                NSLayoutRelation.Equal,
+                1,
+                0
+            );
+            AddConstraint(createdConstraint);
+            return createdConstraint;
+        }
+
+        private nfloat previous;
```
willsb: Fields should be at the beginning of the file

Toggl.Daneel/Views/ResizableView.cs
```diff
@@ -0,0 +1,69 @@
...
+        private void onDrag()
+        {
+            var point = panGestureRecognizer.TranslationInView(this);
+            var newHeight = HeightConstant + point.Y - previous;
+
+            HeightConstant = newHeight.Clamp(MinHeight, MaxHeight);
+
+            previous = panGestureRecognizer.State == UIGestureRecognizerState.Ended
+                     ? previous = 0
```
willsb: This attribution is not needed. Also, can't this be inline?

Toggl.Foundation.MvvmCross/ViewModels/EditDurationViewModel.cs
```diff
@@ -0,0 +1,69 @@
...
+            SaveCommand = new MvxAsyncCommand(save);
+        }
+
+        public override async Task Initialize(DurationParameter parameter)
+        {
+            await base.Initialize();
+
+            isRunning = parameter.Stop == null;
+            if (isRunning)
+                timeService.CurrentDateTimeObservable
```
willsb: Even though this is a single statement, I'd still use `{ }` here to make things clearer

Toggl.Foundation.MvvmCross/ViewModels/EditDurationViewModel.cs
```diff
@@ -0,0 +1,69 @@
...
+                           .Subscribe(currentTime => StopTime = currentTime);
```
willsb: If the user keeps the phone on this screen for too long, StopTime will change but `onDurationChanged` never gets called, resulting into an UI that does not reflect the actual duration

Toggl.Daneel/Views/ResizableView.cs
arturs105: What do you mean by "this attribution is not needed"?

Toggl.Foundation.MvvmCross/ViewModels/EditDurationViewModel.cs
arturs105: What do you mean by this? `onDurationChanged` is meant to be called when user drags the resizable view to update the duration. The stop time and duration labels gets updated correctly. The only thing that could be missing is `DependsOnAttribute`s on the Duration, but it works properly even without those.

willsb:
I only now realized you didn't add any tests. You need to test the newly added `EditDurationCommand` and add tests to the new `EditDurationViewModel` as well

Toggl.Foundation.Tests/MvvmCross/ViewModels/EditDurationViewModelTests.cs
```diff
@@ -0,0 +1,101 @@
...
+
+                tryingToConstructWithEmptyParameters.ShouldThrow<ArgumentNullException>();
+            }
+
+        }
+
+        public sealed class TheDurationProperty : EditDurationViewModelTest
+        {
+            [Fact]
+            public async Task ChangingDurationForRunningTimeEntryUpdatesStartTime()
```
willsb: This test, and the one below it, has weird grammar. Consider changing it to `The duration when changed while updating the running time entry triggers the update of the start time` (and similar for the test below this one)

PR #538, opened 2017-09-07
Push update state (PUT)

### LINKED ISSUE #498: Sync state: push time entry

simonrozsival:
Closes #498 

This PR makes a PUT request and updates the locally stored data in the database with the data received from the server unless the user changes the data during the round-trip of the HTTP request.

~WIP: I am still not sure if I covered all cases with unit tests, I will try adding more tomorrow.~


Toggl.Foundation/Sync/States/Push/BaseUpdateEntityState.cs
```diff
@@ -0,0 +1,61 @@
...
+        public BaseUpdateEntityState(ITogglApi api, ITogglDatabase database)
+        {
+            this.api = api;
+            this.database = database;
+        }
+
+        public IObservable<ITransition> Start(TModel entity)
+            => update(entity)
+                .Select(updatedEntity => hasChanged(entity)
+                    .Select(changed => changed ? entityChanged(entity) : overwrite(entity.Id, updatedEntity))
```
amulware: The hasChanged and overwrite are two different transactions. This would allow user changes to get lost. Can they not be done in the same transaction using, for example, the batch updating logic with a special conflict resolver that reports back here what it decided?

simonrozsival:
Wating on #542

Toggl.Foundation/Sync/States/Push/BaseUpdateEntityState.cs
```diff
@@ -0,0 +1,65 @@
...
+        {
+            this.api = api;
+            this.database = database;
+        }
+
+        public IObservable<ITransition> Start(TModel entity)
+            => update(entity)
+                .Select(tryOverwrite(entity))
+                .SelectMany(updatedEntity => updatedEntity)
+                .Select(updatedEntity => updatedEntity == entity
```
amulware: Are you sure this equality is guaranteed to work? How do we know that the database will return the same object? I'd rather inject something into the tryOverwrite and overwriteIfLocalEntityDidNotChange that makes sure we know what was decided there, even if we don't get the same object back.

simonrozsival: Maybe the return of the batch update should contain the conflict resolution mode - something we already discussed, but haven't implemented yet. We would then explicitly have "ignore" or "update" and we won't have to compare anything here.

simonrozsival:
This now waits for #547

willsb:
Will review this after 547 is merged into develop

simonrozsival:
This is now ready for a review.

Toggl.Foundation.Tests/Sync/States/Push/BaseUpdateEntityStateTests.cs
```diff
@@ -0,0 +1,183 @@
...
+                GetRepository(database)
+                    .BatchUpdate(Arg.Any<IEnumerable<(long, TModel)>>(), Arg.Any<Func<TModel, TModel, ConflictResolutionMode>>())
+                    .Returns(Observable.Return(new[] { (ConflictResolutionMode.Update, entity) }));
+
+                state.Start(entity).SingleAsync().Wait();
+
+                AssertUpdateReceived(api, entity);
+            }
+
+            public void ReturnsEntityChangedTransitionWhenIfEntityChangesLocally()
```
willsb: Name is a bit weird. The If is not needed

Toggl.Foundation.Tests/Sync/States/Push/BaseUpdateEntityStateTests.cs
```diff
@@ -0,0 +1,183 @@
...
+            {
+                var state = CreateState(api, database);
+                var transition = state.Start(null).SingleAsync().Wait();
+                var parameter = ((Transition<(Exception Reason, TModel)>)transition).Parameter;
+
+                transition.Result.Should().Be(state.UpdatingFailed);
+                parameter.Reason.Should().BeOfType<ArgumentNullException>();
+            }
+
+            public void ReturnsFailTransitionWhenHttpFails()
```
willsb: Shouldn't this be returns *the* fail transition?

Toggl.Foundation/Sync/States/Push/BaseUpdateEntityState.cs
```diff
@@ -0,0 +1,64 @@
...
+        }
+
+        public IObservable<ITransition> Start(TModel entity)
+            => update(entity)
+                .Select(tryOverwrite(entity))
+                .SelectMany(result => result.Select(pair => ((ConflictResolutionMode mode, TModel updatedEntity))pair))
+                .Select(result => result.mode == ConflictResolutionMode.Ignore
+                    ? entityChanged(entity)
+                    : succeeded(result.updatedEntity))
+                .SelectMany(transition => transition)
```
willsb: Can't you simplify this by changing the previous Select into a SelectMany?

Toggl.Foundation.Tests/Sync/States/Push/BaseUpdateEntityStateTests.cs
simonrozsival: Oh I thought I already changed this weird name...

simonrozsival:
I changed it, it works.

simonrozsival:
I believe this is now ready to be reviewed and approved.

PR #541, opened 2017-09-08
Push unsyncable state

### LINKED ISSUE #496: Allow marking of syncables as invalid and keep track of error
If the API errors with 400 (and possibly 402 and others? - let's maybe assume all 400 errors until we are adding logic to deal with specific ones), it means that the entity cannot be synced in its current form.

We should store this fact (as a bool flag?) on the entity so that we never try to sync it again, until the user changes it, which should remove this flag again.

Backend also gives us a localised, human readable error message that could be something like "duration is too long" or "tasks not supported for this workspace". We should keep track of this as well so that we can show it to the user (in the details of the entity) and they know what to fix.

simonrozsival:
Closes #496 

This PR includes a state which follows after the update entity state or the create entity state when there is some error in the process and marks the entity as unsyncable when the error originated from the API request (for now we consider both 4xx and 5xx errors).

Squash right after approval!


Toggl.Foundation/Sync/States/Push/BaseUnsyncableEntityState.cs
```diff
@@ -0,0 +1,47 @@
...
+                failedPush.Reason == null
+                    ? nameof(failedPush.Reason)
+                    : nameof(failedPush.Entity)));
+
+        private IObservable<ITransition> failBecauseOfUnexpectedError(Exception reason)
+            => Observable.Throw<Transition<TModel>>(reason);
+
+        private IObservable<ITransition> markAsUnsyncable(TModel entity, string reason)
+            => GetRepository(database)
+                .Update(entity.Id, CreateUnsyncableFrom(entity, reason))
```
amulware: But what if the user made changes to the model in the meantime? This should not be overwritten, similar to the other state.

simonrozsival: You are right. I can use the BatchUpdate with conflict resolution in a similar way as in the update state. But that is not ideal.. this would be very easy to implement if the `doModifyingTransaction` function of the RealmAdapter was somehow exposed, but it isn't and I'm not sure we would agree on changing the `IRepository` interface. 

On the other hand if the sync of this entity failed and it changed in the meantime, we should try syncing it again, because the reason why the request failed might be removed by the user. I will rewrite this to BatchUpdate.

amulware: Yes, if the user has updated the entity in the meantime, it should be sent again, so we don't want to overwrite it here. I think using the batch update (which is really a batch _conditional_ update) is fine. You may need access to the original entity from the database though to check if it was modified or not?

simonrozsival: I think the same I did in the Update state could be applied here - I compare the `At` of the entity from the database with the `At` of the current representation in the database before overwriting it.

simonrozsival:
When #538 is merged, I will change `BatchUpdate` to `UpdateWithConflictResolution`. Then this will be ready for reviewing (also after the failing build is resolved).

willsb:
Please put this on-hold if it depends on another PR

simonrozsival:
This PR is now ready for a review.

simonrozsival:
This PR is now ready for a review.

Toggl.Foundation/Sync/States/Push/BaseUnsyncableEntityState.cs
```diff
@@ -0,0 +1,55 @@
...
+
+        public BaseUnsyncableEntityState(ITogglDatabase database)
+        {
+            this.database = database;
+        }
+
+        public IObservable<ITransition> Start((Exception Reason, TModel Entity) failedPush)
+            => failedPush.Reason == null || failedPush.Entity == null
+                ? failBecauseOfNullArguments(failedPush)
+                : failedPush.Reason is ApiException
```
willsb: Isn't it too simplistic to assume that `ApiException == Unsyncable`? What about network errors/lack of internet connections/ server error? Wouldn't those throw api exceptions?

Should there be a more complex check here, or does that check happen elsewhere and here we're left only with the truly unsyncable entities?

simonrozsival: I agree that there should be more logic here. But if I follow the issue #496, then this is the right way to do it.

PR #542, opened 2017-09-08
Change batch update to take a specific Id for the updated entities

simonrozsival:
We decided to update this method after an internal discussion.


Toggl.PrimeRadiant/IRepository.cs
```diff
@@ -11,7 +11,7 @@ public interface IRepository<TModel>
         IObservable<TModel> GetById(long id);
         IObservable<TModel> Create(TModel entity);
         IObservable<TModel> Update(long id, TModel entity);
-        IObservable<IEnumerable<TModel>> BatchUpdate(IEnumerable<TModel> entities, Func<TModel, TModel, ConflictResolutionMode> conflictResolution);
+        IObservable<IEnumerable<TModel>> BatchUpdate(IEnumerable<(long, TModel)> entities, Func<TModel, TModel, ConflictResolutionMode> conflictResolution);
```
amulware: Since this is a public interface, do we want to give names to the tuple items?

simonrozsival: I wasn't sure about if it is a good idea to name the values of the tuple here - but it will make it more clear, so I will name it.

PR #547, opened 2017-09-11
Change the return value of BatchUpdate

simonrozsival:
This PR is needed because of #538 

The return value of BatchUpdate wasn't specified before - now we need to know the resolution mode for the entities.


simonrozsival:
I am not sure what you mean by "public" here - the `IRepository` has the same return value - so whenever you use BatchUpdate from within Foundation, you get the enumerable of tuples.

willsb:
Then feel free to merge it :)

PR #549, opened 2017-09-11
Limit project space in log

willsb:
Fixes the log UI so that project/task/client names are truncated when they are too long. I used the opportunity to reduce the amount of views and rename things as needed.

This also now uses the same combiner as the Edit view controller


Toggl.Daneel/Views/TimeEntriesLogViewCell.cs
```diff
@@ -35,51 +35,39 @@ public override void AwakeFromNib()
             base.AwakeFromNib();
 
             TimeLabel.Font = TimeLabel.Font.GetMonospacedDigitFont();
+            var combinerBindingDefinition =
+                $"AttributedText ProjectTaskClient({nameof(TimeEntryViewModel.ProjectName)}, {nameof(TimeEntryViewModel.TaskName)}, {nameof(TimeEntryViewModel.ClientName)}, {ProjectTaskClientLabel.Font.CapHeight}, {nameof(TimeEntryViewModel.ProjectColor)})";
```
arturs105: This can now be removed and we should use fluent bindings instead

Toggl.Daneel/Views/TimeEntriesLogViewCell.cs
```diff
@@ -35,51 +35,39 @@ public override void AwakeFromNib()
...
                 bindingSet.Bind(TimeLabel)
                           .To(vm => vm.Duration)
                           .WithConversion(timeSpanConverter);
 
                 //Color
-                bindingSet.Bind(ProjectLabel)
+                bindingSet.Bind(ProjectTaskClientLabel)
                           .For(v => v.TextColor)
                           .To(vm => vm.ProjectColor)
                           .WithConversion(colorConverter);
```
arturs105: Won't this change the color for the whole label?

willsb: Nope. `TextColor` only applies to the ranges of the `AttributedText` that don't have the `BackgroundColor` set

Toggl.Daneel/Views/TimeEntriesLogViewCell.cs
willsb: 👌

Toggl.Daneel/Views/TimeEntriesLogViewCell.cs
arturs105: But where and why would the background color be changed? Shouldn't we only set the foreground color?

willsb: I meant foreground, my bad 😄

arturs105: I think you should add a comment about this, since it isn't obvious.

willsb: Done

Issue #545, opened 2017-09-08
Investigate syncing observable


amulware:
It appears that the syncing observables currently does not appear to be sending updates correctly.

I took a quick look at the code and it seems fine, so probably needs some more detailed debugging.

simonrozsival:
So the problem is that the `SyncManager` locks itself - the private `isRunningSync` variable is set to `true`, although it should be `false` after sync operation finishes. Also the `StateMachine` locks itself - the `currentState` is not set tu null.

PR #553, opened 2017-09-12
Fix sync failure when API returns a null when an empty list is expected

simonrozsival:
Ref #545 but does not close it - this is only a fix of s sub-problem.

This happens quite a lot when user has either no time entry at all or when nothing changes between two "get since" API calls.

Squash whenever.


Toggl.Foundation.Tests/Sync/States/Pull/PersistStateTests.cs
```diff
@@ -171,6 +177,18 @@ public void UpdatesSinceParametersOfTheFetchedEntity()
                     newSinceParameters => OtherSinceDatesDidntChange(oldSinceParameters, newSinceParameters, newAt)));
             }
 
+            public void HandlesNullValueReceivedFromTheServerAsAnEmptyList()
+            {
+                var oldSinceParameters = new SinceParameters(null);
+                var entities = (List<TInterface>)null;
```
willsb: I think `List<TInterface> entities = null` is a lot clearer

simonrozsival:
I don't think that that was caused by deserialization - I tried putting the JSON you sent me into deserialization tests and it worked just fine. But I will try to simulate it again.

simonrozsival:
Could you please create a separate issue for the deserialization error and add it to Sprint III? I think it would be better to deal with it in a separate PR once this is resolved.

willsb:
Yeah, I definitely think that it should be handled by another PR. I'm just wondering if it's an issue or not. You just said that the error didn't happen when you tried the faulty json. Should I open the issue or not?

simonrozsival:
I think we should be sure that there are no errors and if there was some problem which we can't identify, then I think it is an issue. The worst thing that can happen is that I won't be able to reproduce it anyhow and then we will close it until it surfaces again.

PR #555, opened 2017-09-12
Update Log empty State view

### LINKED ISSUE #514: Update log view empty state
See zeplin for details.

willsb:
Closes #514 

~This is waiting from further clarification from @pe0ny~

As clarified in the issue and on Slack, this is all that's needed for the issue to be closed


arturs105:
The screen for a user without any recent time entries matches the design, however, the same screen is show to a new user.

willsb:
We have no defined what is a "new user". I believe a new user is someone who signed up using the app, something that's not yet doable, therefore this shouldn't be taken into account yet. Correct me if I'm wrong, @pe0ny

pe0ny:
@willsb New user can also be a user who signed up on web but has not created any TE. But that's a _really_ edge case i would say.
So yes, we define new user = user who signs up using the app.

willsb:
@arturs105 As clarified by @pe0ny, I think this is ready for merging 🚂

PR #561, opened 2017-09-13
Fix sync deadlocks

### LINKED ISSUE #545: Investigate syncing observable

simonrozsival:
Closes #545, Ref #544 

There are two deadlocks: in `SyncManager` and in `StateMachineOrchestrator`. Both are caused by "locking" the entity (either by setting a boolean or a `IDisposable`) by a return value of a function, which runs the _whole state chain_ until reaching a dead end and "unlocks" the object in some "on complete" callback/subscribed method before the original method returns its value and "locks" the object due to its inconsistent internal state.

I simplified the `SyncStateQueue`: it must return the state before the orchestrator starts processing the state. Therefore I decided to remove the dependency of the "queue" on the orchestrator and leave the managing of the orchestrator up to the `SyncManager` which already has a depencency on orchestrator anyway. ✅ were adapted too. This in fact effectively solves the `SyncManager` deadlock. Now I also exposed the `IsRunningSync` property as a public-get private-set to make it easier to test. I know this might be considered leaky, but I think it makes sense to do it that way, because the manager does not return any failure when it skips the sync because it is already running and so the caller should have the option of somehow decide whether to start a sync or not when it is already running (and yes, I know that the value of `IsRunningSync` might change between checking the condition and initiating the synchronization.

The other problem is within `StateMachine`: the object holds its "is running" state by assigning the disposable of the subscription to an observable. The problem with this is that there are several times we assign to this variable (and override whatever was there before), but we set it to `null` to mark that the processing has already finished just once. The problem is that we assign `null` before all of the disposables are assigned, which leaves the internal state inconsistent. I made a simple boolean `isRunning` variable which has the same semantics as the one in `SyncManager` does and is set to true only once at the start of the process and set to false when it finishes (either by reaching dead end or catching an error). This fixes the other problems.

I added tests to cover all of these scenarios. The individual classes are well-tested, but it was their coupling which caused the problems. Therefore I didn't use substitutions but used the instances of our implementation.

~WIP: I yet have to add tests for the deadlock situations. I will also write a better description of the deadlocks.~


simonrozsival:
For reviewers: I created a dummy branch `deadlock-failing-tests` which shows how the tests were failing on the `develop` branch for comparison: see https://www.bitrise.io/build/7e8fd4728342741f

Toggl.Foundation.Tests/Sync/StateMachineOrchestratorTests.cs
```diff
@@ -43,7 +44,7 @@ protected void SendStateMachineEvent(StateMachineEvent @event)
 
         public sealed class TheConstructor
         {
-            [Theory]
+            [Xunit.Theory]
```
arturs105: I don't think it's needed

Toggl.Foundation/Sync/SyncStateQueue.cs
```diff
@@ -1,3 +1,4 @@
+using System;
```
arturs105: 🔥

Toggl.Foundation.Tests/Sync/DeadlockTests.cs
```diff
@@ -0,0 +1,237 @@
...
+        }
+
+        public sealed class TheSyncManager : BaseDeadlockTests
+        {
+            public TheSyncManager() : base(Substitute.For<IScheduler>())
+            {
+            }
+
+            [Fact]
+            public void DoNotGetStuckInADeadlockWhenThereAreNoTransitionHandlersForFullSync()
```
willsb: `DoesNot`*. Similar below

PR #564, opened 2017-09-13
Allow changing start time of time entries

### LINKED ISSUE #451: Choose start time when creating time entry
~~Requires #448~~ Done

- [x] Add the option to select start time when starting a time entry
- [ ] Limit the available range (Chosen date/time must be within X hours, and must be in the past. (X to be decided))

willsb:
Ref #451 

The UI hooks will only work properly after #448 is closed, but the tests show that, given the other VM behaves, things will work just fine.

Squash whenever


Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
```diff
@@ -141,6 +146,20 @@ private void toggleProjectSuggestions()
 
         private void toggleBillable() => IsBillable = !IsBillable;
 
+        private async Task changeStartTime()
+        {
+            IsEditingStartDate = true;
+
+            var currentDate = DateParameter.WithDate(StartDate);
```
arturs105: `currentlySelectedDate`. `currentDate` sounds more like `DateTimeOffset.Now`

Toggl.Foundation.Tests/MvvmCross/ViewModels/StartTimeEntryViewModelTests.cs
```diff
@@ -92,6 +92,24 @@ public void TogglesTheIsBillableProperty()
             }
         }
 
+        public sealed class TheChangeStartTimeCommand : StartTimeEntryViewModelTest
```
arturs105: Maybe add a test testing, whether the command sets `IsEditingStartDate ` to true and then to false?

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
willsb: 👌

PR #565, opened 2017-09-13
Allow editing duration of time entries

### LINKED ISSUE #450: Choose duration when creating time entry
Requires: #446

Note that the duration dialog has to work slightly different compared to the one in the edit dialog.

Obviously (I hope) choosing a different duration modifies the start-time of the entry to be created.

willsb:
Closes #450 

Similar to #564, it needs the other ViewModel to behave, but works as tests show.


Toggl.Foundation.Tests/MvvmCross/ViewModels/StartTimeEntryViewModelTests.cs
```diff
@@ -187,6 +187,24 @@ public void SetsTheIsSuggestingProjectsPropertyToFalseIfInProjectSuggestionMode(
             }
         }
 
+        public sealed class TheChangeDurationCommandCommand : StartTimeEntryViewModelTest
```
arturs105: I'd like the same test as when selecting start time - whether `IsEditingDuration ` is set.

PR #566, opened 2017-09-13
Return selected value from SelectDateTimeDialogViewModel

### LINKED ISSUE #448: Start/stop time edit dialog

arturs105:
Ref #448


Toggl.Foundation.MvvmCross/ViewModels/SelectDateTimeDialogViewModel.cs
```diff
@@ -40,7 +40,6 @@ public override void Prepare(DateParameter parameter)
 
         private Task close() => navigationService.Close(this);
```
willsb: Please make this return the value provided when navigating, to ensure nothing changed, otherwise we get undefined behavior

willsb:
Not only does it need to test the added behavior, but the whole ViewModel needs to be tested

PR #572, opened 2017-09-14
Add target binding for UIDatePicker to DateTimeOffset

arturs105:
`UIDatePicker` can now be bound to `DateTimeOffset` (instead of just `DateTime`).

We don't need to worry about time zones, since `NSDate` represents a single point in time and `UIDatePicker` automatically converts the time to user's time zone. It... just works 🍎


Toggl.Daneel/Toggl.Daneel.csproj
```diff
@@ -19,7 +19,7 @@
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
     <CodeAnalysisRuleSet>..\Toggl.ruleset</CodeAnalysisRuleSet>
-    <CodesignKey>iPhone Developer</CodesignKey>
+    <CodesignKey>iPhone Developer: Artūrs Eimanis (H7QQ87FRDS)</CodesignKey>
```
willsb: 🤔

arturs105: I've been trying out IntelliJ Rider and it sometimes makes some changes to Toggl.Daneel.csproj. I usually catch that and remove all changes, but since this time i added files to Daneel, i didn't check the .csproj for unrelated changes.

PR #571, opened 2017-09-14
Push sync: state machine setup

simonrozsival:
Closes #452 

WIP: Waiting for #541 and issue #569 to be resolved first.

PR #575, opened 2017-09-15
Push sync: state machine factory

### LINKED ISSUE #452: Push time entries to backend
At first, pushing time entries will be enough since that is the only thing the app can create for now.

The general steps are:

- [x] checking if any time entries have to be synced, and returning the oldest of them #497
- [x] try syncing this entry #498
  - [x] on success, use conflict resolution to persist return value from API #498
  - [x] on failure, mark entry as unsyncable (needs changes to database models) and don't try syncing it again until user changes it #496
  - [ ] (optional) ~~if failure is server/connection error, enter proper retry loops (refer to internal designs and documentation)~~  (we'll do this later)

simonrozsival:
Closes #452, replaces #571 

In the future, it will be simple to add pushing of other types of data using the `configurePush<T>` function.

Squash whenever


Toggl.Foundation/TogglSyncManagerFactory.cs
```diff
@@ -42,5 +45,44 @@ private static void configurePullTransitions(TransitionHandlerProvider transitio
             transitions.ConfigureTransition(persistClients.FinishedPersisting, persistProjects.Start);
             transitions.ConfigureTransition(persistProjects.FinishedPersisting, persistTimeEntries.Start);
         }
+        
+        private static void configurePushTransitions(TransitionHandlerProvider transitions, ITogglDatabase database, ITogglApi api, StateResult entryPoint)
+        {
+            var noMoreTimeEntriesToPush =
+                configurePushTransitionsForTimeEntries(transitions, database, api, entryPoint);
```
arturs105: `noMoreTimeEntriesToPush ` isn't used.

arturs105: Also, code guidelines say: "when splitting statements into multiple lines, do so by starting new lines with operators, and using an additional level of indentation." So the assignment would look like this (if `noMoreTimeEntriesToPush ` variable would be necessary).

```
var noMoreTimeEntriesToPush
    = configurePushTransitionsForTimeEntries(transitions, database, api, entryPoint);
```

Toggl.Foundation/TogglSyncManagerFactory.cs
```diff
@@ -42,5 +45,44 @@ private static void configurePullTransitions(TransitionHandlerProvider transitio
...
+            var push = new PushTimeEntriesState(database);
+            var pushOne = new PushOneEntityState<IDatabaseTimeEntry>();
+            var create = new CreateTimeEntryState(api, database);
+            var update = new UpdateTimeEntryState(api, database);
+            var unsyncable = new UnsyncableTimeEntryState(database);
+
+            return configurePush(transitions, entryPoint, push, pushOne, create, update, unsyncable);
+        }
+
+        private static IStateResult configurePush<T>(
```
arturs105: If i understand correctly this could be used (in the future) to push other entities besides time entries by creating a method like `configurePushTransitionsForTimeEntries` for other entities and then configuring the states for all entities in `configurePushTransitions`, right?

simonrozsival: Yes, exactly.

Toggl.Foundation/TogglSyncManagerFactory.cs
simonrozsival: I know the variable isn't used. I leaved it there intentionally to make it obvious why the `configurePushTransitions*` have a return value and so the person who will create push sync of some other type of entities knows right away what the entry point of the other push loop should be. I think it is reasonable to leave the variable here - the compiler will throw it away and it documents the code well. But if you insist or others think it shouldn't be here right now, I can remove it.

I will definitelly change the formatting.

PR #583, opened 2017-09-15
Add the duration property to time entries

simonrozsival:
The duration property is needed at least because of the webapp - previously it showed the duration of time entries created in the mobile app to be zero.

Squash whenever.


Toggl.Ultrawave/Models/TimeEntry.cs
```diff
@@ -20,6 +20,12 @@ internal sealed partial class TimeEntry : ITimeEntry
 
         public DateTimeOffset? Stop { get; set; }
 
+        public int Duration
+        {
+            get => Stop.HasValue ? (int)(Stop.Value - Start).TotalSeconds : 0;
```
willsb: Did you test this? I thought this had to be negative instead...

simonrozsival: When I put negative number into duration, I get 500 error. Webapp sends a negative duration, but in v8. Maybe let's keep this open over the weekend (sorry :/) and let's discuss it with the backend team then.

simonrozsival: OK so today negative values work in the staging env, so I fixed this. I also tested it against the webapp and now it works correctly - the webapp shows the time entry as running after pushing.

Since this was the only remark you had and I addressed it, I take it as your approval still holds and I will squash it.

PR #584, opened 2017-09-15
Update UI when syncing

### LINKED ISSUE #510: Update UI after fetching data
We should reload the time entry log and running time entry whenever sync finishes.

willsb:
Closes #510 

Injecting the DataSource and making it be used as the Repository will cause the Create and Update observables to be fired as the Sync Manager runs, making the UI responsive.

Squash whenever


simonrozsival:
This is very interesting. In fact I think we could simplify the whole `BaseCreateEntityState` (and similar states) - we could inject the repository directly and get rid of the `GetRepository` method althogether... because without this refactoring the code looks really weird (e.g., the `database` field is completely useless both to the base and concrete classes).

I can do that in a follow-up refactoring PR, so you don't have to deal this in a UI-related PR, so we can merge this before. Or maybe I can do it before you get here and we could simplify this by passing the data instead of the PrimeRadiant repository.

willsb:
Waiting for #590

Toggl.Foundation/DataSources/TimeEntriesDataSource.cs
```diff
@@ -137,5 +163,17 @@ private TimeEntry createUpdatedTimeEntry(IDatabaseTimeEntry timeEntry, EditTimeE
                         .SetAt(timeService.CurrentDateTime)
                         .SetSyncStatus(SyncStatus.SyncNeeded)
                         .Build();
+
+        private IDatabaseTimeEntry runningTimeEntry(IDatabaseTimeEntry timeEntry)
+        {
+            if (timeEntry == null || timeEntry.Stop != null)
+            {
+                currentlyRunningTimeEntryId = null;
```
arturs105: I don't like that this variable is updated in a method that is used in `Select()`. It selects the running entry **and** sets the `currentlyRunningTimeEntryId ` field. Is it possible to internally subscribe to `CurrentlyRunningTimeEntry` and set `currentlyRunningTimeEntryId ` when a new time entry is emitted?

Toggl.Foundation/TogglSyncManagerFactory.cs
```diff
@@ -6,37 +6,41 @@
...
 using Toggl.PrimeRadiant.Models;
+using Toggl.Foundation.DataSources;
 
 namespace Toggl.Foundation
 {
     public static class TogglSyncManager
     {
-        public static ISyncManager CreateSyncManager(ITogglDatabase database, ITogglApi api, IScheduler scheduler)
+        public static ISyncManager CreateSyncManager(ITogglDatabase database, ITogglApi api, 
+            ITogglDataSource dataSource, IScheduler scheduler)
```
arturs105: I think one of the following would be better

```
public static ISyncManager CreateSyncManager(
    ITogglDatabase database, ITogglApi api, ITogglDataSource dataSource, IScheduler scheduler)

```
```
public static ISyncManager CreateSyncManager(
    ITogglDatabase database, ITogglApi api,
    ITogglDataSource dataSource, IScheduler scheduler)
```

```
public static ISyncManager CreateSyncManager(
    ITogglDatabase database,
    ITogglApi api,
    ITogglDataSource dataSource,
    IScheduler scheduler)
```

arturs105: Applies to the following methods aswell.

PR #587, opened 2017-09-18
Implement persisting WorkspaceFeatureCollection to the database

### LINKED ISSUE #397: Implement persisting WorkspaceFeatureCollection to the database
Follow-up of #214

simonrozsival:
Closes #397 

This PR changes the `RealmAdapter` and `IRepository` to remove the dependency on `IBaseModel` and `IDatabaseSyncable`. The `IDatabaseSyncable` is never needed, so this can be removed easily. The `IBaseModel` is needed to access the `Id` property. This is overcome by injecting a `matchById` function, which is already used by the `BatchUpdate` function and the use of this function is widened. Therefore the `IBaseModel` is not needed in the constraints and the adapter can be used for the `WorkspaceFeatureCollection` models (matching by `WorkspaceId` instead of `Id`).

I also added the `Get(long id)` to the `RealmAdapter` so that the `IRepository` can implement `GetById(long id)` without knowing which property of the entity is the "id". This does not change the public interface of the `IRepository`.

~~WIP: The individual workspace features are persisted through the relation with the collection - the `RealmConstructors.tt` generator was updated according to https://realm.io/docs/xamarin/latest/#to-many-relationships. This should work, but I don't think I am doing it the right way. This would probably lead to accumulating many unnecessary entities which would never be accessed again. I should probably add the `WorkspaceId` property to the `RealmWorkspaceFeature` and in `RealmWorkspaceFeatureCollection.SetPropertiesFrom` first fetch (or just delete?) all features related to the workspace and only then trying to add or update the new ones.~~

~~I would be very glad if somebody gave me some comments to these ideas, so I can finish the PR tomorrow.~~


Toggl.PrimeRadiant.Realm/Models/RealmWorkspaceFeature.cs
```diff
@@ -16,5 +17,18 @@ public WorkspaceFeatureId FeatureId
         public int FeatureIdInt { get; set; }
         
         public bool Enabled { get; set; }
+
+        public static RealmWorkspaceFeature FindOrCreate(IDatabaseWorkspaceFeature entity, Realms.Realm realm)
+        {
+            var feature = realm.All<RealmWorkspaceFeature>()
+                .SingleOrDefault(x => x.FeatureIdInt == (int)entity.FeatureId && x.Enabled == entity.Enabled);
+
+            if (feature == null)
```
willsb: Null coalesce makes sense here, IMO

PR #594, opened 2017-09-19
Select projects and clients from edit view

### LINKED ISSUE #445: Edit view project selection and auto completion
I hope this can reuse the completion from the creation dialog.

It should split the entered string into words and match those against client, project, and task names.

Of course, the user should then be able to choose a project, and upon confirmation, the change has to be persisted.

arturs105:
Ref #445 

This PR adds the functionality to select and filter projects and clients and also return them to the edit time entry view. It does not, however, persist those changes after pressing "Confirm changes".

Do a functional test before approving.


Toggl.Daneel/Combiners/ProjectTaskClientValueCombiner.cs
```diff
@@ -6,10 +6,7 @@
 using System.Text;
 using UIKit;
 using Foundation;
-using System.Globalization;
```
willsb: Unrelated

Toggl.Daneel/ViewSources/SelectProjectTableViewSource.cs
```diff
@@ -0,0 +1,103 @@
...
+            => tableView.DequeueReusableCell(cellIdentifier, indexPath);
+        
+        public override nfloat GetHeightForRow(UITableView tableView, NSIndexPath indexPath) => 48;
+
+        public override nfloat GetHeightForHeader(UITableView tableView, nint section) => 40;
+
+        public override void HeaderViewDisplayingEnded(UITableView tableView, UIView headerView, nint section)
+        {
+            var firstVisible = TableView.IndexPathsForVisibleRows.First();
+            if (firstVisible.Section == section + 1)
```
willsb: Early return here instead

Toggl.Daneel/ViewSources/SelectProjectTableViewSource.cs
```diff
@@ -0,0 +1,103 @@
...
+
+        public override void WillDisplayHeaderView(UITableView tableView, UIView headerView, nint section)
+        {
+            var firstVisibleIndexPath = TableView.IndexPathsForVisibleRows.First();
+            if (firstVisibleIndexPath.Section == section)
+            {
+                var nextHeader = (SelectProjectHeaderViewCell)TableView.GetHeaderView(section + 1);
+                if (nextHeader == null) return;
+                nextHeader.TopSeparatorView.Hidden = false;
+                ((SelectProjectHeaderViewCell)headerView).TopSeparatorView.Hidden = true;
```
willsb: Maybe cast this at the top, assign `TopSeparatorView.Hidden` and early return if needed?

Toggl.Daneel/Views/SelectProjectHeaderViewCell.cs
```diff
@@ -0,0 +1,41 @@
...
+using UIKit;
+
+namespace Toggl.Daneel.Views
+{
+    public sealed partial class SelectProjectHeaderViewCell : MvxTableViewHeaderFooterView
+    {
+        public static readonly NSString Key = new NSString(nameof(SelectProjectHeaderViewCell));
+        public static readonly UINib Nib;
+
+        public UIView TopSeparatorView => TopSeparator;
```
willsb: Is `TopSeparator` protected?

Toggl.Foundation.MvvmCross/ViewModels/ProjectSuggestionCollection.cs
```diff
@@ -0,0 +1,23 @@
...
+    {
+        public string Workspace { get; }
+
+        public ProjectSuggestionCollection(string workspace, IEnumerable<ProjectSuggestion> suggestions)
+        {
+            Ensure.Argument.IsNotNull(workspace, nameof(workspace));
+            Ensure.Argument.IsNotNull(suggestions, nameof(suggestions));
+
+            Workspace = workspace;
+            Add(new ProjectSuggestion());
```
willsb: I don't like this to be made public. There's no need to instantiate multiple objects; expose this as `ProjectSuggestion.NoProject` instead

willsb: Also, do we really need the `No Project` suggestion to appear on every workspace?

Toggl.Foundation.MvvmCross/ViewModels/EditTimeEntryViewModel.cs
```diff
@@ -164,6 +170,31 @@ private Task close()
...
+            var selectedProjectId = await navigationService.Navigate<SelectProjectViewModel, IdParameter, IdParameter>(parameter);
+
+            if (selectedProjectId.Id == projectId)
+                return;
+
+            projectId = selectedProjectId.Id;
+
+            if (projectId == 0)
+            {
+                Project = "";
```
willsb: Make all these assignments inline

Toggl.Foundation.MvvmCross/ViewModels/EditTimeEntryViewModel.cs
```diff
@@ -164,6 +170,31 @@ private Task close()
             StartTime = selectedDate.GetDate();
         }
 
+        private async Task selectProject()
+        {
+            var parameter = IdParameter.WithId(projectId);
+            var selectedProjectId = await navigationService.Navigate<SelectProjectViewModel, IdParameter, IdParameter>(parameter);
+
+            if (selectedProjectId.Id == projectId)
+                return;
```
willsb: No need for a newline here

Toggl.Foundation.MvvmCross/ViewModels/SelectProjectViewModel.cs
```diff
@@ -0,0 +1,82 @@
...
+
+            OnTextChanged();
+        }
+
+        private void OnTextChanged()
+        {
+            infoSubject.OnNext(Text);
+        }
+
+        private void onSuggestions(IEnumerable<AutocompleteSuggestion> suggestions)
```
willsb: Since you can only ever have one type of suggestions, why not cast this on the observable declaration?

Toggl.Foundation.MvvmCross/ViewModels/SelectProjectViewModel.cs
```diff
@@ -0,0 +1,82 @@
...
+
+        public override void Prepare(IdParameter parameter)
+            => projectId = parameter.Id;
+
+        public override async Task Initialize()
+        {
+            await base.Initialize();
+
+            infoSubject.AsObservable()
+                .SelectMany(t => dataSource.AutocompleteProvider.Query(t, AutocompleteSuggestionType.Projects))
```
willsb: Give a meaningful name to this `t`

Toggl.Foundation.MvvmCross/ViewModels/SelectProjectViewModel.cs
```diff
@@ -0,0 +1,82 @@
...
+                .Subscribe(onSuggestions);
+
+            OnTextChanged();
```
willsb: Use `dataSource.AutocompleteProvider.Query(t, AutocompleteSuggestionType.Projects).StartWith()` or something like that to combine observables and making your intentions clearer instead of calling this method here

Toggl.Foundation.Tests/MvvmCross/ViewModels/SelectProjectViewModelTests.cs
```diff
@@ -0,0 +1,206 @@
...
+            }
+
+            [Fact]
+            public void ReturnsTheSelectedProjectId()
+            {
+                var selectedProject = new ProjectSuggestion();
+                ViewModel.SelectProjectCommand.ExecuteAsync(selectedProject).Wait();
+
+                NavigationService.Received()
+                    .Close(Arg.Is(ViewModel), Arg.Is<IdParameter>(p => p.Id == selectedProject.ProjectId));
```
willsb: This test is kinda bad because you're not initializing the VM with a different project id and the default value for long is 0 (the same value an empty `ProjectSuggestion` returns

Toggl.Foundation.Tests/MvvmCross/ViewModels/SelectProjectViewModelTests.cs
```diff
@@ -0,0 +1,206 @@
...
+            }
+        }
+
+        public sealed class TheTextProperty : SelectProjectViewModelTest
+        {
+            [Fact]
+            public async Task WhenChangedTheAutocompleteProviderIsQueried()
```
willsb: `WhenChangedQueriesTheAutocompleteProvider`

Toggl.Foundation/Autocomplete/Suggestions/ProjectSuggestion.cs
```diff
@@ -44,9 +46,11 @@ public ProjectSuggestion(IDatabaseProject project)
             ProjectName = project.Name;
             ProjectColor = project.Color;
 
-            if (project.Client == null) return;
-
-            ClientName = project.Client.Name;
+            if (project.Workspace != null)
+                Workspace = project.Workspace.Name;
```
willsb: Since you're only accessing Workspace once, why not use the elvis operator? (similar for Client)

Toggl.Daneel/Views/SelectProjectHeaderViewCell.cs
arturs105: It has no access modifier in the `SelectProjectHeaderViewCell.designer.cs`, so i think it's private.

Toggl.Foundation.MvvmCross/ViewModels/ProjectSuggestionCollection.cs
arturs105: Yes, Heidi confirmed, that the empty project should be in every workspace.

Toggl.Daneel/ViewSources/SelectProjectTableViewSource.cs
arturs105: I agree about moving the cast to the top, but i don't see how i can early return, since the both parts of the `if` statement are needed.

Toggl.Foundation.MvvmCross/ViewModels/EditTimeEntryViewModel.cs
arturs105: Do you mean something like this?

```
if (projectId == 0)
{
    Project = ""; Task = ""; Client = ""; ProjectColor = ""; return;
}
```

If so, then i disagree. After seeing this, i had the urge to remove the braces, since it looks like a single statement.

Toggl.Daneel/ViewSources/SelectProjectTableViewSource.cs
```diff
@@ -0,0 +1,104 @@
...
+            if (nextHeader == null) return;
+
+            nextHeader.TopSeparatorView.Hidden = true;
+        }
+
+        public override void WillDisplayHeaderView(UITableView tableView, UIView headerView, nint section)
+        {
+            var selectProjectHeaderViewCell = (SelectProjectHeaderViewCell)headerView;
+            var firstVisibleIndexPath = TableView.IndexPathsForVisibleRows.First();
+            if (firstVisibleIndexPath.Section == section)
```
willsb:
```
var topIsHidden = firstVisibleIndexPath.Section == section;
selectProjectHeaderViewCell.TopSeparatorView.Hidden = topIsHidden;
if (!topIsHidden) return;
```

Toggl.Foundation.MvvmCross/ViewModels/EditTimeEntryViewModel.cs
willsb:
```csharp
if (projectId == 0)
{
    Project = Task = Client = ProjectColor = ""; 
    return;
}
```

Toggl.Foundation.MvvmCross/ViewModels/EditTimeEntryViewModel.cs
```diff
@@ -162,6 +168,29 @@ private Task close()
...
 
+        private async Task selectProject()
+        {
+            var selectedProjectId = await navigationService.Navigate<SelectProjectViewModel, long, long>(projectId);
+
+            if (selectedProjectId == projectId) return;
+
+            projectId = selectedProjectId;
+
+            if (projectId == 0)
```
willsb:
```csharp
if (projectId == 0)
{
    Project = Task = Client = ProjectColor = ""; 
    return;
}
```

Toggl.Foundation.Tests/MvvmCross/ViewModels/SelectProjectViewModelTests.cs
```diff
@@ -0,0 +1,206 @@
...
+            {
+                await ViewModel.SelectProjectCommand.ExecuteAsync(ProjectSuggestion.NoProject);
+
+                await NavigationService.Received().Close(Arg.Is(ViewModel), Arg.Any<long>());
+            }
+
+            [Fact]
+            public void ReturnsTheSelectedProjectId()
+            {
+                var selectedProject = ProjectSuggestion.NoProject;
```
willsb: Like I said: Please do initialize this test with a value other than 0 so we can see that the selected value is indeed pumped

willsb: Or maybe pass a different selected project here, whichever works best

Toggl.Foundation.MvvmCross/ViewModels/EditTimeEntryViewModel.cs
arturs105: Yes, hadn't used this for a long time, so had completely forgotten about it.

arturs105:
The cursor color was changed [here](https://github.com/toggl/mobileapp/pull/594/commits/3cb464aac2af70bb0549c700c6278417a2317cef) by setting the tint color of the text field.

Toggl.Daneel/ViewSources/SelectProjectTableViewSource.cs
arturs105: So the whole method looks something like this?

```
var selectProjectHeaderViewCell = (SelectProjectHeaderViewCell)headerView;
var firstVisibleIndexPath = TableView.IndexPathsForVisibleRows.First();
var appearingFromTop = firstVisibleIndexPath.Section == section;

selectProjectHeaderViewCell.TopSeparatorView.Hidden = appearingFromTop;

if (!appearingFromTop) return;

var nextHeader = (SelectProjectHeaderViewCell)TableView.GetHeaderView(section + 1);
if (nextHeader == null) return;
nextHeader.TopSeparatorView.Hidden = false;
```

willsb: Exactly

willsb: Actually, nevermind, leave it as is

PR #602, opened 2017-09-20
Show exception message when login fails

arturs105:
This could be useful for bitrise builds, when other people can't login. Though we have to remember to change this back to normal before release. Maybe add another build configuration for bitrise and use preprocessor directives? Then we wouldn't have to remember to change this back. Thoughts?


willsb:
A simple:

```csharp

            #if DEBUG
            ErrorText = ex.Message;
            #endif
```

Fixes it for us

PR #606, opened 2017-09-21
Persist selected project for time entry

### LINKED ISSUE #445: Edit view project selection and auto completion

arturs105:
Closes #445


willsb:
So it'll set previously set to `null` projectIds to 0?

Not sure I like this

arturs105:
Yeah, sorry about that, will fix.

PR #615, opened 2017-09-22
Fix duration editing for running time entries

### LINKED ISSUE #447: Edit view duration editing
The user should be able to change the duration, which is translated into a change of the stop time strictly speaking. See #446

- [x] Fix the height glitch for long (>12h) time entries
- [ ] Fix duration editting for running time entries
- [x] Persist duration edits

arturs105:
Closes #447.

Editing running time entry used to set the stop time for the entry, making it not running anymore.


Toggl.Foundation.MvvmCross/ViewModels/EditDurationViewModel.cs
```diff
@@ -62,7 +62,14 @@ private Task close()
             => navigationService.Close(this, defaultResult);
 
         private Task save()
-            => navigationService.Close(this, DurationParameter.WithStartAndStop(StartTime, StopTime));
+        {
+            DateTimeOffset? stopTimeToBeReturned;
+            if (isRunning)
```
willsb: Can't this benefit from using a ternary operator?

arturs105: Not very neatly, because of casting issues. The following would work, but i prefer an `if` over this:
```
var stopTime = isRunning ? null : (DateTimeOffset?)StopTime;
```

PR #617, opened 2017-09-22
Serialization based on workspace features

### LINKED ISSUE #207: Decide what to do with premium attributes
Some properties are accessible only to pro users (`Project.Billable`) and must not be sent over to api, but our current serialisation model does not support that.

One possible solution would be a `IgnoreIfNotPremiumAttribute` similarly as `IgnoreWhenPostingAttribute`.

Any other solutions?

### LINKED ISSUE #522: Figure out POSTing/PUTting based on workspace features
Related to #207

Essentially, properties related to paid features may not be included when pushing entities, if the specific feature is not enabled.

We need some way (probably attributes) for the API models to know what properties require what features. We then have to inject the workspace features present when pushing, and the serialiser has to be smart about including only allowed properties.

simonrozsival:
Closes #207, closes #522 

This PR does not contain application of this attribute on the the existing models, because that is not part of neither of the issues.

Squash whenever.


Toggl.Ultrawave/Serialization/PropertiesFilters/RequiresFeatureAttributeFilter.cs
```diff
@@ -0,0 +1,36 @@
...
+        private readonly IWorkspaceFeatureCollection featuresCollection;
+
+        public RequiresFeatureAttributeFilter(IWorkspaceFeatureCollection featuresCollection)
+        {
+            this.featuresCollection = featuresCollection;
+        }
+
+        public IList<JsonProperty> Filter(IList<JsonProperty> properties)
+        {
+            if (featuresCollection != null)
```
willsb: Could be an early return

Toggl.Ultrawave/Serialization/PropertiesFilters/RequiresFeatureAttributeFilter.cs
```diff
@@ -0,0 +1,36 @@
...
+            {
+                foreach (JsonProperty property in properties)
+                {
+                    var attributes = property.AttributeProvider.GetAttributes(typeof(RequiresFeatureAttribute), false);
+                    if (attributes.OfType<RequiresFeatureAttribute>().Any(isNotEnabled))
+                        property.ShouldSerialize = _ => false;
```
willsb: Did you consider using the entity itself here (the `_` field) so we don't have to pass the feature collection around, or is that not doable?

Toggl.Ultrawave/Serialization/PropertiesFilters/IgnoreAttributeFilter.cs
```diff
@@ -1,17 +1,14 @@
...
-using System.Collections.Generic;
+﻿using System.Collections.Generic;
 using System.Linq;
-using Newtonsoft.Json;
 using Newtonsoft.Json.Serialization;
 
 namespace Toggl.Ultrawave.Serialization
 {
-    internal sealed class IgnoreAttributeContractResolver<TIgnoredAttribute> : DefaultContractResolver
+    internal class IgnoreAttributeFilter<TIgnoredAttribute> : IPropertiesFilter
```
willsb: Why did you unseal this class?

simonrozsival: I tried implementing this with inheritance first, so I unsealed it, but forgot to put it back when I changed my mind.

Toggl.Ultrawave/Serialization/PropertiesFilters/RequiresFeatureAttributeFilter.cs
simonrozsival: I am not sure how you mean this - the entity itself does not have any reference to the feature collection. Could you be more specific?

PR #622, opened 2017-09-22
Limit since to two months

### LINKED ISSUE #405: Never ask for more than two months of history with `since` parameters
Instead, consider just getting all entities instead in that case.

We also have to check how much data we actually get in the case of time entries and consider instead of 'all' getting the last ~6 weeks to make sure we have data to show, and to complete and suggest from.

simonrozsival:
Closes #405 

Squash whenever.


Toggl.Foundation/Sync/States/Pull/FetchAllSinceState.cs
```diff
@@ -43,6 +44,9 @@ public FetchAllSinceState(ITogglDatabase database, ITogglApi api)
 
         private static IObservable<T> getSinceOrAll<T>(DateTimeOffset? threshold,
             Func<DateTimeOffset, IObservable<T>> since, Func<IObservable<T>> all)
-            => threshold.HasValue ? since(threshold.Value) : all();
+            => threshold.HasValue && isWithinLimit(threshold.Value) ? since(threshold.Value) : all();
+
+        private static bool isWithinLimit(DateTimeOffset threshold)
+            => threshold > DateTimeOffset.Now.AddMonths(-SinceDateLimitMonths);
```
amulware: Do we not want to use our time service here?

Toggl.Foundation.Tests/Sync/States/Pull/FetchAllSinceStateTests.cs
```diff
@@ -76,6 +76,31 @@ public void MakesCorrectCallsWithSinceThresholds()
             }
 
             [Fact]
+            public void MakesCorrectCallsWithoutThresholdsWhenSinceTooFarInThePast()
```
amulware: Naming. This tests for a specific time interval, and that a specific call is received, so the name should be specific about that. General names like 'correct call' and 'too far in the past' are not very explanatory.

Toggl.Foundation.Tests/Sync/States/Pull/FetchAllSinceStateTests.cs
```diff
@@ -76,6 +79,31 @@ public void MakesCorrectCallsWithSinceThresholds()
             }
 
             [Fact]
+            public void MakesApiCallsWithoutTheSinceParameterWhenTheThresholdIsMoreThanTwoMonthsInThePast()
```
amulware: Should we have a test that checks dates just less than two months?

Toggl.Foundation/TogglSyncManagerFactory.cs
```diff
@@ -16,12 +16,13 @@ public static class TogglSyncManager
             ITogglDatabase database,
             ITogglApi api,
             ITogglDataSource dataSource,
+            ITimeService timeService,
```
amulware: Isn't this called anywhere that needs a change?

Toggl.Foundation.Tests/Sync/States/Pull/FetchAllSinceStateTests.cs
simonrozsival: I can make a property test that will give a random date between 0 - 2 months for that.

Toggl.Foundation/TogglSyncManagerFactory.cs
simonrozsival: I am not sure I understand you. I changed all places where I had to to use `ITimeService` for creating a new `FetchAllSince` state.

amulware: Yeah, nevermind. I was blind. All good here.

Toggl.Foundation.Tests/Sync/States/Pull/FetchAllSinceStateTests.cs
```diff
@@ -53,29 +54,36 @@ public void MakesNoApiCallsBeforeSubscription()
...
-            public void MakesCorrectCallsWithSinceThresholds()
+            [Property]
+            public void MakesCorrectCallsWithSinceThresholdsWhenSinceIsLessThanTwoMonthsInThePast(double val)
             {
+                if (double.IsNaN(val) || double.IsInfinity(val))
+                    return;
+
                 var now = timeService.CurrentDateTime;
+                var twoMonths = (now.AddMonths(2) - now);
+                int ticks = (int)(twoMonths.Ticks * Math.Abs(val) / double.MaxValue);
```
amulware: This has to be a long, not an int. By my calculation, two months is `5.256e+13` ticks, which does not fit inside an integer.

Toggl.Foundation.Tests/Sync/States/Pull/FetchAllSinceStateTests.cs
```diff
@@ -50,29 +54,61 @@ public void MakesNoApiCallsBeforeSubscription()
...
 
+            [Property]
+            public void MakesCorrectCallsWithSinceThresholdsWhenSinceIsLessThanTwoMonthsInThePast(double val)
+            {
+                if (double.IsNaN(val) || double.IsInfinity(val))
+                    return;
+
+                var now = timeService.CurrentDateTime;
+                var twoMonths = (now.AddMonths(2) - now);
+                int ticks = (int)(twoMonths.Ticks * Math.Abs(val) / double.MaxValue);
```
amulware: Als, please do the double division before the multiplication: `ticks * (abs / max)` to ensure nothing funny happens with large values.

Toggl.Foundation.Tests/Sync/States/Pull/FetchAllSinceStateTests.cs
```diff
@@ -62,8 +62,8 @@ public void MakesCorrectCallsWithSinceThresholdsWhenSinceIsLessThanTwoMonthsInTh
 
                 var now = timeService.CurrentDateTime;
                 var twoMonths = (now.AddMonths(2) - now);
-                int ticks = (int)(twoMonths.Ticks * Math.Abs(val) / double.MaxValue);
-                var since = now.AddTicks(-ticks);
+                var seconds = twoMonths.Seconds * (long)(Math.Abs(val) / double.MaxValue);
```
amulware: This needs to be TotalSeconds instead of Seconds. The latter merely returns the seconds component, not the full duration. Also, TotalSeconds returns a double, so the cast to long is no longer needed then.

Toggl.Foundation.Tests/Sync/States/Pull/FetchAllSinceStateTests.cs
```diff
@@ -62,7 +62,7 @@ public void MakesCorrectCallsWithSinceThresholdsWhenSinceIsLessThanTwoMonthsInTh
 
                 var now = timeService.CurrentDateTime;
                 var twoMonths = (now.AddMonths(2) - now);
-                var seconds = twoMonths.Seconds * (long)(Math.Abs(val) / double.MaxValue);
+                var seconds = twoMonths.TotalSeconds * (long)(Math.Abs(val) / double.MaxValue);
```
amulware: Actually, the cast to long is not only not needed but also significantly changing behaviour. The double will be in the range of 0 to 1, and will be rounded to one of those values, which I doubt is the intent.

simonrozsival: Yes, that was not intended. Now the semantics and the math should be OK.

PR #623, opened 2017-09-22
Fix combiner glitch

### LINKED ISSUE #631: Clients shouldn't be colored after the project
Currently clients (Toggl; Teamweek) are colored after the project though they shouldn't be. Clients should always be #a3a3a3
Note: tasks on the other hand should be colored after the project

screen shot 2017-09-26 at 08 21 48

willsb:
- The binding context might run on a background thread for :rocket: 
- We were using functions that required running on the UI thread
- :boom: 

Fix:

I now cache the little dot and require the font size when constructing the combiner. This way stuff doesn’t break.

This also now requires the client color, since this was broken previously.

Squash whenever

Collaterally closes #631


Toggl.Daneel/Combiners/ProjectTaskClientValueCombiner.cs
```diff
@@ -6,91 +6,89 @@
...
-using System.Globalization;
 using MvvmCross.Platform.UI;
-using CoreGraphics;
-using MvvmCross.Plugins.Color;
 using Toggl.Foundation.MvvmCross.Helper;
 using MvvmCross.Plugins.Color.iOS;
 using Toggl.Daneel.Extensions;
+using MvvmCross.Platform;
+using MvvmCross.Platform.Core;
+using System.Threading.Tasks;
```
arturs105: 10, 13, 14 and 15 🙅‍♂️

Toggl.Daneel/Combiners/ProjectTaskClientValueCombiner.cs
```diff
@@ -6,91 +6,89 @@
...
 
 namespace Toggl.Daneel.Combiners
 {
     public sealed class ProjectTaskClientValueCombiner : MvxFormatValueCombiner
     {
         private const string projectDotImageResource = "icProjectDot";
+        private readonly UIColor clientColor;
+        private readonly NSMutableAttributedString cachedDotString;
+
+        public ProjectTaskClientValueCombiner(nfloat fontHeight, UIColor clientColor)
```
arturs105: Check the color for null?

Toggl.Daneel/Combiners/ProjectTaskClientValueCombiner.cs
```diff
@@ -6,91 +6,89 @@
...
 
-        private static NSAttributedString tryAddColorToClient(string text, string client)
-        {
+            var text = builder.ToString();
+
             var result = new NSMutableAttributedString(text);
-            if (string.IsNullOrEmpty(client))
-                return result;
+            var clientIndex = text.Length - client.Length;
+            var projectNameRange = new NSRange(0, clientIndex);
```
arturs105: Doesn't this also include the task, since it's "Project: Task Client"?

arturs105:
Also looks like this fixes #631. But at the same time breaks the project/task/client row in the edit view  - project shouldn't be colored. Maybe we should add a `bool` to indicate, whether the project text needs to be colored.

Toggl.Daneel/Combiners/ProjectTaskClientValueCombiner.cs
willsb: Text includes the task as well

PR #626, opened 2017-09-25
Implement retry loop for push sync

### LINKED ISSUE #525: Implement retry loop for push sync
On non-400 failure when pushing, we should enter a retry loop according to internal documents before trying to push anything again.

simonrozsival:
Closes #525 

Squash whenever


Toggl.Foundation.Tests/Sync/States/Push/BaseUnsyncableEntityStateTests.cs
```diff
@@ -163,6 +183,33 @@ public void TheOnlyThingThatChangesInTheUnsyncableEntityIsTheSyncStatusAndLastSy
...
+
+                var transition = state.Start((reason, entity)).Wait();
+                var parameter = ((Transition<(TModel Entity, double LastWaitingTime)>)transition).Parameter;
+
+                transition.Result.Should().Be(state.SlowRetry);
+                parameter.Entity.Should().Be(entity);
+                parameter.LastWaitingTime.Should().Be(0);
+            }
+
+            public void TheSlowRetryIsEnteredWhenTheReasonIsAServerErrorExceptionOtherThanInternalServerErrorException(ServerErrorException reason)
```
willsb: The**Fast**Retry

Toggl.Foundation.Tests/Sync/States/Push/BaseUnsyncableEntityStateTests.cs
```diff
@@ -60,6 +78,8 @@ public interface IStartMethodTestHelper
             void ThrowsWhenTheReasonExceptionIsNotAnApiException();
             void TheUpdatedEntityHasTheSameIdAsTheOriginalEntity();
             void TheOnlyThingThatChangesInTheUnsyncableEntityIsTheSyncStatusAndLastSyncErrorMessage();
+            void TheSlowRetryIsEnteredWhenTheReasonIsAnInternalServerErrorException();
+            void TheSlowRetryIsEnteredWhenTheReasonIsAServerErrorExceptionOtherThanInternalServerErrorException(ServerErrorException reason);
```
willsb: The**Fast**Retry

Toggl.Foundation/Sync/States/Push/BaseRetryPushState.cs
```diff
@@ -0,0 +1,55 @@
+﻿using System;
+using System.Reactive.Concurrency;
+using System.Reactive.Linq;
+using Toggl.Ultrawave;
+
+namespace Toggl.Foundation.Sync.States.Push
+{
+    public abstract class BaseRetryPushState<TModel>
```
amulware: Why is this generic? I don't see anything in the design/specs that requires any knowledge of any models for retrying.

Toggl.Foundation/Sync/States/Push/BaseRetryPushState.cs
```diff
@@ -0,0 +1,55 @@
...
+            this.api = api;
+            this.scheduler = scheduler;
+            this.rnd = rnd;
+        }
+
+        public IObservable<ITransition> Start((TModel Entity, double LastWaitingTime) failedPush)
+            => api.Status.Get()
+                .SelectMany(isAvailable => isAvailable
+                    ? proceed(failedPush.Entity)
+                    : retry(failedPush));
```
amulware: Depending on the kind of status failure, this has to respond differently, which I don't believe it does.

Toggl.Foundation/Sync/States/Push/BaseRetryPushState.cs
willsb: I believe it's generic so it knows what entity to retry next, but it could simply go to the point where it looks at the DB instead, so yeah

Toggl.Foundation/Sync/States/Push/BaseRetryPushState.cs
simonrozsival: I don't understand what you mean by this - so far I relied on the implementation of the `StatusApi` which returns a boolean. I will look more into this to see all the scenarios.

Toggl.Foundation/Sync/States/Push/BaseRetryPushState.cs
simonrozsival: Yes the reason was to resume the loop directly in the "PushOne" state, but you are right that the entity can be forgotten and looked up from the database again.

Toggl.Foundation/Sync/States/Push/CheckServerStatusState.cs
```diff
@@ -0,0 +1,48 @@
...
+        }
+
+        public IObservable<ITransition> Start()
+            => api.Status.Get()
+                .SelectMany(proceed)
+                .Catch((InternalServerErrorException e) => delayedRetry(delay.NextSlowDelay()))
+                .Catch((ServerErrorException e) => delayedRetry(delay.NextFastDelay()));
+
+        private IObservable<ITransition> proceed(bool isAvaialbe)
+            => isAvaialbe
```
amulware: typo

Toggl.Foundation/Sync/States/Push/CheckServerStatusState.cs
```diff
@@ -0,0 +1,48 @@
...
+            this.api = api;
+            this.scheduler = scheduler;
+            this.delay = delay;
+        }
+
+        public IObservable<ITransition> Start()
+            => api.Status.Get()
+                .SelectMany(proceed)
+                .Catch((InternalServerErrorException e) => delayedRetry(delay.NextSlowDelay()))
+                .Catch((ServerErrorException e) => delayedRetry(delay.NextFastDelay()));
```
amulware: Are you sure this exception covers all other cases?

amulware: Unless there is some argument for catching this specifically, I might rather prefer catching anything here.

Toggl.Foundation/Sync/States/Push/CheckServerStatusState.cs
```diff
@@ -0,0 +1,48 @@
...
+
+        private IObservable<ITransition> proceed(bool isAvaialbe)
+            => isAvaialbe
+                ? Observable.Return(ServerIsAvailable.Transition())
+                : delayedRetry(delay.NextFastDelay());
```
amulware: Two different ways to get into the fast delay loop? That seems smelly.

Toggl.Foundation/Sync/States/Push/CheckServerStatusState.cs
```diff
@@ -0,0 +1,48 @@
...
+        {
+            this.api = api;
+            this.scheduler = scheduler;
+            this.delay = delay;
+        }
+
+        public IObservable<ITransition> Start()
+            => api.Status.Get()
+                .SelectMany(proceed)
+                .Catch((InternalServerErrorException e) => delayedRetry(delay.NextSlowDelay()))
```
amulware: Looking at the implementation and tests of the status API, I don't think this case will ever happen. See: https://github.com/toggl/mobileapp/blob/develop/Toggl.Ultrawave.Tests/ApiClients/StatusApiTests.cs#L60

simonrozsival: I need to change the `StatusApi` class in a different way. I made a pull request yesterday, but that doesn't in fact do exactly what I wanted. I will make a separate PR for the status API and then update this code afterwards.

Toggl.Foundation/TogglSyncManagerFactory.cs
```diff
@@ -73,25 +76,30 @@ public static class TogglSyncManager
...
+            var delay = new RetryDelayService(rnd);
             var push = new PushTimeEntriesState(database);
             var pushOne = new PushOneEntityState<IDatabaseTimeEntry>();
-            var create = new CreateTimeEntryState(api, dataSource.TimeEntries);
-            var update = new UpdateTimeEntryState(api, dataSource.TimeEntries);
-            var unsyncable = new UnsyncableTimeEntryState(dataSource.TimeEntries);
+            var create = new CreateTimeEntryState(api, dataSource.TimeEntries, delay);
+            var update = new UpdateTimeEntryState(api, dataSource.TimeEntries, delay);
+            var unsyncable = new UnsyncableTimeEntryState(dataSource.TimeEntries, delay);
+            var checkServerStatus = new CheckServerStatusState(api, scheduler, delay);
```
amulware: Instead of injecting the delay everywhere, would it not be easier and clearer, if there was an additional step afterwards that did the reset?

Toggl.Foundation/TogglSyncManagerFactory.cs
```diff
@@ -113,6 +122,10 @@ public static class TogglSyncManager
             transitions.ConfigureTransition(update.UpdatingSucceeded, push.Start);
             transitions.ConfigureTransition(update.UpdatingFailed, markUnsyncable.Start);
 
+            transitions.ConfigureTransition(markUnsyncable.CheckServerStatus, checkServerStatus.Start);
+            transitions.ConfigureTransition(checkServerStatus.Retry, checkServerStatus.Start);
+            transitions.ConfigureTransition(checkServerStatus.ServerIsAvailable, push.Start);
```
amulware: If the status is fine, it still has to do wait for the api retry delay before moving on here. I don't see that being done.

Toggl.Foundation/TogglSyncManagerFactory.cs
amulware: On that note, I don't see any tests for the reset.

Toggl.Foundation/TogglSyncManagerFactory.cs
```diff
@@ -113,6 +122,10 @@ public static class TogglSyncManager
             transitions.ConfigureTransition(update.UpdatingSucceeded, push.Start);
             transitions.ConfigureTransition(update.UpdatingFailed, markUnsyncable.Start);
 
+            transitions.ConfigureTransition(markUnsyncable.CheckServerStatus, checkServerStatus.Start);
```
amulware: I don't really understand what the unsyncable state does or why it is hooked up to the retry loop.

I wonder if it might be good if we start having another flowchart with our actual implementation, including annotations on all the transitions. I'm even thinking this could be auto-generated. Will think about that...

Toggl.Foundation/TogglSyncManagerFactory.cs
simonrozsival: Making one state for reseting makes sense. Although I am not sure if it will make it that much clearer and simpler. But at least it will be explicit - I will give it a try.

Toggl.Foundation/TogglSyncManagerFactory.cs
simonrozsival: You are right that it seems strange. I will change this.

Having a diagram of the whole process would be nice. And it probably wouldn't even be that hard to generate it with DOT and GraphViz.

Toggl.Foundation/TogglSyncManagerFactory.cs
simonrozsival: Oh, really? OK, I will add a delay.

amulware: It might be that I'm missing something, but what's the point of resetting the api retry delay, if we are not using (and increasing) it anywhere?

Toggl.Foundation/Sync/RetryDelayService.cs
```diff
@@ -0,0 +1,45 @@
+﻿using System;
+
+namespace Toggl.Foundation.Sync
+{
+    internal sealed class RetryDelayService : IRetryDelayService
+    {
+        private readonly Random rnd;
```
amulware: Let's call this `random` (also the parameter name in the constructor).

Toggl.Foundation/Sync/States/Push/BaseCreateEntityState.cs
```diff
@@ -36,7 +39,12 @@ private IObservable<TModel> create(TModel entity)
             => createdEntity => repository.Update(entity.Id, createdEntity).Select(CopyFrom);
 
         private Func<Exception, IObservable<ITransition>> fail(TModel entity)
-            => e => Observable.Return(CreatingFailed.Transition((e, entity)));
+            => exception =>
+                exception is ApiException
+                    ? exception is ServerErrorException
+                        ? Observable.Return(ServerError.Transition((exception, entity)))
+                        : Observable.Return(ClientError.Transition((exception, entity)))
+                    : Observable.Return(UnknownError.Transition((exception, entity)));
```
amulware: Can we be sure that `ApiException && !ServerErrorException -> ClientErrorException`?

Toggl.Foundation/Sync/States/Push/BaseUnsyncableEntityState.cs
```diff
@@ -13,6 +13,7 @@ internal abstract class BaseUnsyncableEntityState<TModel>
         private IRepository<TModel> repository;
 
         public StateResult<TModel> MarkedAsUnsyncable { get; } = new StateResult<TModel>();
+        public StateResult CheckServerStatus { get; } = new StateResult();
```
amulware: This is not used as far as I can tell.

Toggl.Foundation/Sync/States/Push/BaseUpdateEntityState.cs
```diff
@@ -47,7 +50,12 @@ private IObservable<ITransition> entityChanged(TModel entity)
                 : ConflictResolutionMode.Update;
 
         private Func<Exception, IObservable<ITransition>> fail(TModel entity)
-            => exception => Observable.Return((ITransition)UpdatingFailed.Transition((exception, entity)));
+            => exception =>
+                exception is ApiException
+                    ? exception is ServerErrorException
+                        ? Observable.Return(ServerError.Transition((exception, entity)))
+                        : Observable.Return(ClientError.Transition((exception, entity)))
+                    : Observable.Return(UnknownError.Transition((exception, entity)));
```
amulware: Same question as above.

Toggl.Foundation/Sync/States/Push/PushOneEntityFinishedState.cs
```diff
@@ -0,0 +1,23 @@
+﻿using System;
+using System.Reactive.Linq;
+
+namespace Toggl.Foundation.Sync.States
+{
+    internal sealed class PushOneEntityFinishedState
```
amulware: The name of this describes when it is supposed to be used, instead of what it does. I think it's better to call this something like `ResetAPIDelayState`.

The states should not know how they fit into a bigger picture.

simonrozsival: OK, why not. I thought of it more as a "steps that cleans up after one entity was pushed". But since there is currently only the resetting of delay duration service, I can make it more transparent about what it does.

Toggl.Foundation/Sync/RetryDelayService.cs
simonrozsival: OK.

Toggl.Foundation/Sync/States/Push/CheckServerStatusState.cs
```diff
@@ -0,0 +1,42 @@
...
+        public CheckServerStatusState(ITogglApi api, IScheduler scheduler, IRetryDelayService delay)
+        {
+            this.api = api;
+            this.scheduler = scheduler;
+            this.delay = delay;
+        }
+
+        public IObservable<ITransition> Start()
+            => api.Status.IsAvailable()
+                .Delay(delay.NextSlowDelay(), scheduler)
```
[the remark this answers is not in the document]
simonrozsival: I thought it was intended to do this way - because it only says it should do "slow retry". Making a separate delay timer for this is easy and it makes sense, I would expect the delay between the "ok" and next trial to be quite short (even though it says "slow"; originally I thought there wouldn't be any delay after we discover the server is already OK).

If we reset the delay timer here, we basically never increase the delay of this timer - because if the status endpoint does not return 200, the timer is not used and therefore the delay does not increase, or it returns 200 and it is used, so the delay is increased, but then immediately it would be reset. This makes sense, why not, if the repeated request fails, then there would be some delay anyway before trying the status endpoint. But then we don't need the complex timer here, we just need to generate a random number within the range.

I need to put more thought into this second delay object though, it is a bit confusing.

Toggl.Foundation/Sync/States/Push/BaseCreateEntityState.cs
simonrozsival: `ApiException` is not abstract, so in theory, there could be some other exception, which could be an `ApiException` but not server or client one. Maybe the first condition should not be `is ApiException` but rather `exception is ServerErrorException || exception is ClientErrorException` - that would make it maybe more bulletproof (although mayb.

Toggl.Foundation/Sync/States/Push/BaseUnsyncableEntityState.cs
simonrozsival: Yes, I forgot to :fire: it after I removed the usage of it, thanks.

Toggl.Foundation/Sync/States/Push/PushOneEntityFinishedState.cs
```diff
@@ -0,0 +1,23 @@
...
+
+        public PushOneEntityFinishedState(IRetryDelayService delay)
+        {
+            this.delay = delay;
+        }
+
+        public IObservable<ITransition> Start()
+        {
+            delay.Reset();
+            return Observable.Return(PushNext.Transition());
```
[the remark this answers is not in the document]
simonrozsival: I don't know the answer to your question in general, but you are right that this could easily be a part of the observable process and it would make it probably a bit more readable as well.

simonrozsival:
I am not aware of any guidelines for test names we use. I agree we should stick to some pattern. I will rename the tests.

amulware:
Given the circumstances, let's merge this without Will's approval.

PR #629, opened 2017-09-25
Ensure there can only ever be one running time entry when pulling data

### LINKED ISSUE #476: Ensure there can only ever be one running time entry when pulling data
Otherwise, :boom:

Followup to #437.

How to do it (roughly?), according to frontend:

> We have 2 running time entries after a pull. We compare the ats and determine the later one - that one will remain running. The other one needs to be stopped. The stop time of the other one becomes **the start of the "next" time entry (going forward in time) or now if there isn't one**.
>
> This should work for most cases where the user has started a TE on an offline-device, then started another one on another device and eventually these have both ended up in the webapp (one coming from our store and the other from the server) regardless of which whether is "ours" or "theirs" is newer.
>
> Additionally, since the automatically calculated stop time may still not be correct we could try to be polite and display a notification that tells the user to double check the time entry.

(emphasis mine)

Apart from the bold part, this could be implemented by adding a more complicated batch update method, something like this:

```
public IEnumerable<TModel> BatchUpdate(
	IEnumerable<TModel> entities,
	Func<TModel, Expression<Func<TModel, bool>>> matchEntity,
	Func<TModel, TModel, ConflictResolutionMode> conflictResolution,
	Func<TModel, Expression<Func<TModel, bool>>> areRivals,
	Func<TModel, TModel, (TModel, TModel)> fixRivals)
{
	Ensure.Argument.IsNotNull(entities, nameof(entities));
	Ensure.Argument.IsNotNull(matchEntity, nameof(matchEntity));
	Ensure.Argument.IsNotNull(conflictResolution, nameof(conflictResolution));

	var resolvedEntities = new List<TRealmEntity>();

	var realm = Realms.Realm.GetInstance();
	using (var transaction = realm.BeginWrite())
	{
		var realmEntities = realm.All<TRealmEntity>();

		foreach (var entity in entities)
		{
			var oldEntity = (TRealmEntity)realmEntities.SingleOrDefault(matchEntity(entity));
			var resolveMode = conflictResolution(oldEntity, entity);
			var resolvedEntity = resolveEntity(realm, oldEntity, entity, resolveMode);

			if (resolvedEntity != null)
			{
				resolvedEntities.Add(resolvedEntity);
			}

			if (resolveMode == ConflictResolutionMode.Create || resolveMode == ConflictResolutionMode.Update)
			{
				var rival = (TRealmEntity)realm.All<TRealmEntity>().SingleOrDefault(areRivals(entity));
				if (rival != null)
				{
					(TModel fixedEntity, TModel fixedRival) = fixRivals(resolvedEntity, rival);
					resolvedEntity.SetPropertiesFrom(fixedEntity, realm);
					rival.SetPropertiesFrom(fixedRival, realm);
				}
			}
		}

		transaction.Commit();
	}

	return resolvedEntities;
}
```

Except that the bold part needs another query, and things become a bit of a mess potentially.

Also, we really don't wanna run the `areRivals` query for every entity persisted, but only for those that can potentially have rivals (i.e. running time entries), so that adds a bit more complication.

simonrozsival:
Closes #476 

The implementation is general, so we can use it for use cases such as "make sure there aren't multiple projects with the same name".

~WIP: there are no tests yet.~

Squash whenever.


Toggl.PrimeRadiant.Realm/RealmAdapter.cs
```diff
@@ -72,13 +74,28 @@ public TModel Update(long id, TModel entity)
             using (var transaction = realm.BeginWrite())
             {
                 var realmEntities = realm.All<TRealmEntity>();
+                var maybeHaveRivals = new List<TRealmEntity>();
```
amulware: Not sure if the name is clear enough? It sounds more like a bool then a collection to me.

Toggl.Foundation/Sync/ConflictResolution/TimeEntryRivalsResolver.cs
```diff
@@ -0,0 +1,29 @@
...
+        public (IDatabaseTimeEntry FixedEntity, IDatabaseTimeEntry FixedRival) FixRivals(IDatabaseTimeEntry entity, IDatabaseTimeEntry rival, IQueryable<IDatabaseTimeEntry> allTimeEntries)
+            => entity.At > rival.At ? (entity, stop(rival, allTimeEntries)) : (stop(entity, allTimeEntries), rival);
+
+        private IDatabaseTimeEntry stop(IDatabaseTimeEntry toBeStopped, IQueryable<IDatabaseTimeEntry> allTimeEntries)
+        {
+            var next = allTimeEntries
+                .OrderBy(te => te.Stop)
+                .FirstOrDefault(other => other.Stop > toBeStopped.Start);
+            var stop = next?.Start ?? DateTimeOffset.Now;
+            return new TimeEntry(toBeStopped, stop);
```
amulware: I think that the stopped entry should be marked as needing sync, right? This does not seem to happen here.

simonrozsival: Good point.

Toggl.PrimeRadiant.Realm/RealmAdapter.cs
simonrozsival: I will try to think of something more descriptive then. Something like `entitiesWithPotentialRivals`?

amulware: Sure.

Toggl.Foundation/Sync/ConflictResolution/TimeEntryRivalsResolver.cs
simonrozsival: In fact the constructor sets the sync status to `SyncNeeded`. I will add a test for this behavior instead of changing this code.

amulware: I'm not happy with that implicit behaviour, but that's not part of this PR.

Toggl.Foundation.Tests/Sync/ConflictResolution/TimeEntryRivalsResolverTests.cs
```diff
@@ -0,0 +1,132 @@
...
+
+        private readonly ITimeService timeService;
+
+        private readonly IQueryable<IDatabaseTimeEntry> timeEntries = new EnumerableQuery<IDatabaseTimeEntry>(new[]
+        {
+            TimeEntry.Clean(new Ultrawave.Models.TimeEntry { Start = new DateTimeOffset(2017, 9, 10, 12, 0, 0, TimeSpan.Zero) }),
+            TimeEntry.Clean(new Ultrawave.Models.TimeEntry { Start = new DateTimeOffset(2017, 9, 15, 12, 0, 0, TimeSpan.Zero) }),
+            TimeEntry.Clean(new Ultrawave.Models.TimeEntry { Start = new DateTimeOffset(2017, 9, 20, 12, 0, 0, TimeSpan.Zero) }),
+            TimeEntry.Clean(new Ultrawave.Models.TimeEntry { Start = new DateTimeOffset(2017, 9, 25, 12, 0, 0, TimeSpan.Zero) })
+        });
```
amulware: Should these not have stop times to represent real data better?

simonrozsival: The stop times are not needed anywhere in these tests. But I can of course add some stop times.

amulware: They are not needed because we know the implementation of the resolver. For all the tests know it might get very upset if given a list of several apparently running time entries ;)

simonrozsival: That makes sense, I will add the stop times so there is no doubt.

simonrozsival:
This PR is now ready for a review.

Toggl.Foundation/Sync/ConflictResolution/TimeEntryRivalsResolver.cs
```diff
@@ -0,0 +1,42 @@
...
+            => CanHaveRival(entity)
+                ? potentialRival => potentialRival.Stop == null && potentialRival.Id != entity.Id
+                : (Expression<Func<IDatabaseTimeEntry, bool>>)(_ => false);
+
+        public (IDatabaseTimeEntry FixedEntity, IDatabaseTimeEntry FixedRival) FixRivals(IDatabaseTimeEntry entity, IDatabaseTimeEntry rival)
+            => rival.At < entity.At ? (entity, stop(rival)) : (stop(entity), rival);
+
+        private IDatabaseTimeEntry stop(IDatabaseTimeEntry toBeStopped)
+        {
+            var next = timeEntries
```
amulware: Does using the repo here work well with realm? We are in effect starting a read in the middle of a write, right? Even if it works with Realm, I'm worried it might not work with other databases that do not allow reading while writing. Why didn't injecting the queryable here work?

Toggl.Foundation/Sync/ConflictResolution/TimeEntryRivalsResolver.cs
```diff
@@ -0,0 +1,42 @@
...
+            this.timeService = timeService;
+            this.timeEntries = timeEntries;
+        }
+
+        public bool CanHaveRival(IDatabaseTimeEntry entity) => !entity.Stop.HasValue;
+
+        public Expression<Func<IDatabaseTimeEntry, bool>> AreRivals(IDatabaseTimeEntry entity)
+            => CanHaveRival(entity)
+                ? potentialRival => potentialRival.Stop == null && potentialRival.Id != entity.Id
+                : (Expression<Func<IDatabaseTimeEntry, bool>>)(_ => false);
```
amulware: Should this case ever happen? If not, I'd maybe rather see an exception here.

simonrozsival: It can happen when this class is used incorrectly. I agree with the exception.

Toggl.Foundation/Sync/ConflictResolution/TimeEntryRivalsResolver.cs
simonrozsival: The second transaction is a read-only transaction, I don't think there would be a reason why this won't work in a different database, but maybe it won't be possible... it definitely works well in Realm the way I wrote it. I am not really sure if we should optimize for potential future databases we may or may not use at this moment.

The main problem with doing it the old way was it threw `{System.NotSupportedException: The class ITimeEntry is not in the limited set of classes for this Realm, so sorting by its properties is not allowed.   at Realms.RealmResultsVisitor.TraverseSort (System.Linq.Expressions.MemberExpression expression) [0x0003c] in…}`.

Now when I look at it after some time, it seems to me that the `OrderBy` is not needed at all. I can do it more efficiently using `Aggregate` and maybe this whole problem would go away after this change.

amulware: I see. This is another issue with our abstraction and how it doesn't allow us to query on Realm directly. I believe this could also be fixed by using the previous implementation and running the query in memory (on IEnumerable, instead of IQueriable). Would that be better (I believe it will run in memory either way)?

Toggl.Foundation/Sync/ConflictResolution/TimeEntryRivalsResolver.cs
```diff
@@ -0,0 +1,43 @@
...
+            => rival.At < entity.At ? (entity, stop(rival, allTimeEntries)) : (stop(entity, allTimeEntries), rival);
+
+        private IDatabaseTimeEntry stop(IDatabaseTimeEntry toBeStopped, IQueryable<IDatabaseTimeEntry> allTimeEntries)
+        {
+            var stop = allTimeEntries.Where(other => other.Start > toBeStopped.Start)
+                .ToList()
+                .Select(te => te.Start)
+                .Where(start => start != default(DateTimeOffset))
+                .DefaultIfEmpty(timeService.CurrentDateTime)
+                .Min();
```
amulware: I like this query much better than the previous cod, thanks!

Toggl.Foundation/Sync/ConflictResolution/TimeEntryRivalsResolver.cs
```diff
@@ -0,0 +1,43 @@
...
+            return potentialRival => potentialRival.Stop == null && potentialRival.Id != entity.Id;
+        }
+
+        public (IDatabaseTimeEntry FixedEntity, IDatabaseTimeEntry FixedRival) FixRivals(IDatabaseTimeEntry entity, IDatabaseTimeEntry rival, IQueryable<IDatabaseTimeEntry> allTimeEntries)
+            => rival.At < entity.At ? (entity, stop(rival, allTimeEntries)) : (stop(entity, allTimeEntries), rival);
+
+        private IDatabaseTimeEntry stop(IDatabaseTimeEntry toBeStopped, IQueryable<IDatabaseTimeEntry> allTimeEntries)
+        {
+            var stop = allTimeEntries.Where(other => other.Start > toBeStopped.Start)
+                .ToList()
```
amulware: Why this ToList()? To make the following run in memory? Isn't it enough to cast to `IEnumerable`?

Toggl.Foundation/Sync/ConflictResolution/TimeEntryRivalsResolver.cs
```diff
@@ -0,0 +1,43 @@
...
+        {
+            this.timeService = timeService;
+        }
+
+        public bool CanHaveRival(IDatabaseTimeEntry entity) => !entity.Stop.HasValue;
+
+        public Expression<Func<IDatabaseTimeEntry, bool>> AreRivals(IDatabaseTimeEntry entity)
+        {
+            if (!CanHaveRival(entity))
+                throw new InvalidOperationException("The entity cannot have any rivals.");
```
amulware: If you prefer the if statement here, I'm fine with that, but just to make sure: Are you aware that with C#7 throwing an exception is an expression? So you can `condition ? expression : throw` for example.

simonrozsival: There were a lot of red squiggles when I did that, but I didn't spend very much time on it, so maybe there was some different bug which caused them. I am OK with the if statement here, but if you want me to use the ternary here, I can spend some time on it.

Toggl.Foundation/Sync/ConflictResolution/TimeEntryRivalsResolver.cs
simonrozsival: I need to execute the query in Realm and then do the rest in memory. This seemed like the most straightforward way of doing it. I am not sure if casting to `IEnumerable` executes the Realm query, I can have a look in the documentation.

amulware: To the best of my knowledge, IEnumerable will always run in memory, being pure C#. IQueryable is the one that does magic.

PR #635, opened 2017-09-26
Add tokens to Start Time Entry

### LINKED ISSUE #518: Allow selecting tags when creating time entries
This needs auto completion, suggestions, and complex UI. Refer to zeplin. Some info here: #315

### LINKED ISSUE #620: Change project tag
Change project tag, so there is not border and background is 12% of the project color:

toggl - ios - add new entry project added 1x

willsb:
Ref #518 
Closes #620 

NSAttributedString covered most of our needs, but I still needed to use CoreText and CoreGraphics because of the Rounded corners.

Not that I’m using custom attributes for each type of token. This is so we don’t rely on the default attributes, which can them be changed freely at any given time.


Toggl.Daneel/Binding/TextViewTextInfoTargetBinding.cs
```diff
@@ -96,25 +107,47 @@ private void queueValueChange()
...
-            var actualCursorPosition = descriptionLength + (isMovingForward ? paddedProjectLength : 0);
 
-            textFieldInfoSubject.OnNext(textFieldInfo.WithTextAndCursor(newDescription, actualCursorPosition));
+            var paddedProjectLength = textFieldInfo.PaddedProjectName().Length;
+            var cursorIsInsideProjectToken = newCursorPosition <= descriptionLength + paddedProjectLength;
+            if (cursorIsInsideProjectToken)
+            {
+                var actualCursorPosition = descriptionLength + (isMovingForward ? paddedProjectLength : 0);
+
+                onTextFieldInfoChanged(textFieldInfo.WithTextAndCursor(newDescription, actualCursorPosition));
```
simonrozsival: I am not sure which line of code it is exactly, but I can't write any text after the project "token", which should be possible according to https://zpl.io/25KDM3a though.

Toggl.Daneel/Binding/TextViewTextInfoTargetBinding.cs
```diff
@@ -96,25 +107,47 @@ private void queueValueChange()
...
+            foreach (var tag in textFieldInfo.Tags)
+            {
+                var tagTextLength = tag.Name.Length + 6;
+                currentIndex += tagTextLength;
+                var cursorIsInsideTag = newCursorPosition <= currentIndex;
+                if (!cursorIsInsideTag) continue;
+
+                var actualCursorPosition = currentIndex - (isMovingForward ? 0 : tagTextLength);
+
+                onTextFieldInfoChanged(textFieldInfo.WithTextAndCursor(newDescription, actualCursorPosition));
```
simonrozsival: The same problem as with project happens for tags. A more interesting thing happens when I try to type between two tags - if I have for example two tags "Tag A" and "Tag B" and I write the letter "X" with the cursor between the tokens, I get this: http://prntscr.com/gq7kti

Toggl.Daneel/Binding/TextViewTextInfoTargetBinding.cs
willsb: This will have to be changed in another PR

Toggl.Daneel/Binding/TextViewTextInfoTargetBinding.cs
willsb: Second part was addressed

arturs105:
Just did a functional test and this looks awesome, but before my actual review i'd like to point out that, when there are multiple tags selected and i delete one from the end (or the middle) the cursor jumps to the start of the text field. I think it should stay at the place that i just deleted the tag from. Probably an issue for another PR.

willsb:
@arturs105 I addressed your comment, it makes sense :)

Toggl.Foundation/Autocomplete/TextFieldInfo.cs
```diff
@@ -1,13 +1,14 @@
...
+using System.Linq;
 using Toggl.Foundation.Autocomplete.Suggestions;
 using Toggl.Multivac.Models;
 
 namespace Toggl.Foundation.Autocomplete
 {
     public struct TextFieldInfo : IEquatable<TextFieldInfo>
     {
-        public static TextFieldInfo Empty { get; } = new TextFieldInfo("", 0, null, "", "", new TagSuggestion[0]);
+        public static TextFieldInfo Empty { get; } = new TextFieldInfo(" ", 0, null, "", "", new TagSuggestion[0]);
```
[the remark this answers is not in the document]
willsb: 🤷‍♂️

PR #638, opened 2017-09-26
Do not hide the exceptions thrown by the request in StatusApi

simonrozsival:
This is needed because of #525 - I need to know what was the exact reason why the status check failed, so the thrown exception must not be hidden and translated into a simple `false` value.

Do not forget to run integration tests when reviewing.

Squash whenever.


simonrozsival:
I will use it like this: `api.Status.Get().Select(isAvailable => ...).Catch((Exception e) => ...);`

PR #639, opened 2017-09-26
Billable flag behavior when starting time entries

### LINKED ISSUE #528: Support billable flag in creation view
The button should only be shown if either:
- the user has selected a project that has a workspace with the Premium feature enabled PR
- the user has not selected a project and the default workspace has said feature enabled

i.e.

```
billableButton.Visible = (project?.Workspace ?? user.DefaultWorkspace).HasFeature(Premium)
```

Further, when the user selects a project the billable button state should be set to the projects DefaultBillable (or something like that) value. (Basically, users can have projects that will always default to billable time entries for convenience)

Of course, if the workspace in question does not have the feature enabled, billable should never be set to true.

willsb:
Closes #528 

Squash whenever


Toggl.Foundation.Tests/MvvmCross/ViewModels/StartTimeEntryViewModelTests.cs
```diff
@@ -502,6 +531,43 @@ public void SetsTheProjectColorToTheSuggestedProjectColor()
...
+                public async Task DisablesBillableIfTheWorkspaceOfTheSelectedProjectDoesNotAllowIt(bool? billableValue)
+                {
+                    Project.Billable.Returns(billableValue);
+                    DataSource.Projects.GetById(ProjectId).Returns(Observable.Return(Project));
+                    DataSource.Workspaces.GetById(WorkspaceId).Returns(Observable.Return(Workspace));
+                    DataSource.Workspaces.WorkspaceHasFeature(WorkspaceId, WorkspaceFeatureId.Pro)
+                        .Returns(Observable.Return(false));
+
+                    ViewModel.SelectSuggestionCommand.Execute(Suggestion);
+                    await Task.Delay(1000);
```
arturs105: Is this here because of concurrency problems. If so, isn't there a better way to ensure that everything runs fine?

Toggl.Foundation.Tests/DataSources/WorkspacesDataSourceTests.cs
```diff
@@ -0,0 +1,91 @@
...
+                var defaultWorkspace = await WorkspacesSource.GetDefault();
+
+                defaultWorkspace.Should().BeSameAs(DefaultWorkspace);
+            }
+        }
+
+        public sealed class TheGetByIdMethod : WorkspaceDataSourceTest
+        {
+            [Fact]
+            public async Task ReturnsTheDefaultWorkspaceForPassedId()
```
arturs105: I don't think this name is correct, because `GetById` returns whatever database returns for that id.

Toggl.Foundation.Tests/DataSources/WorkspacesDataSourceTests.cs
```diff
@@ -0,0 +1,91 @@
...
+                var defaultWorkspace = await WorkspacesSource.GetById(DefaultWorkspaceId);
+
+                defaultWorkspace.Should().BeSameAs(DefaultWorkspace);
+            }
+        }
+
+        public sealed class TheWorkspaceHasFeatureMethod : WorkspaceDataSourceTest
+        {
+            [Fact]
+            public async Task ReturnsTheDefaultWorkspaceForTheCurrentUser()
```
arturs105: Naming could also be improved.

PR #641, opened 2017-09-27
Fix modal controller counting in presenter

willsb:
Right now, dismissing via tapping on the background wasn’t updating the number of ModalViewControllers on the presenter. This caused some unexpected issues.

To test the problem: 
- Open the app on develop
- Login
- Create time entry
- Stop it
- Open edit view
- Dismiss it by tapping the background
- Try opening settings
- :boom: 

Squash whenever


Toggl.Daneel/Presentation/Transition/FromBottomTransitionDelegate.cs
```diff
@@ -1,11 +1,21 @@
...
+using Toggl.Multivac;
 using UIKit;
 
 namespace Toggl.Daneel.Presentation.Transition
 {
     public sealed class FromBottomTransitionDelegate : NSObject, IUIViewControllerTransitioningDelegate
     {
+        private readonly Action onCompletedCallback;
+
+        public FromBottomTransitionDelegate(Action onCompletedCallback)
```
arturs105: Is the `onCompletedCallback` invoked, when the presented view controller is dismissed? If so then the name is misleading, since the current name might give the idea that the action is invoked when the from bottom transition (animation) completes.

PR #644, opened 2017-09-27
Fix Collection namespaces

willsb:
They kinda messy

I also reduced the number of collections so that we’re only using the absolute must. This will be needed for a future PR

Squash whenever


Toggl.Daneel/ViewSources/StartTimeEntryTableViewSource.cs
```diff
@@ -28,6 +31,13 @@ public StartTimeEntryTableViewSource(UITableView tableView)
             tableView.RegisterNibForCellReuse(StartTimeEntryEmptyViewCell.Nib, emptySuggestionIdentifier);
         }
 
+        public override UIView GetViewForHeader(UITableView tableView, nint section)
+        {
+            if (!UseGrouping) return null;
+
+            return null;
```
arturs105: wat? I assume this for the grouping in the future. I think this method should be added, when the grouping will be introduced.

Toggl.Foundation.MvvmCross/Collections/WorkspaceGroupedCollection.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System.Collections.Generic;
+using MvvmCross.Core.ViewModels;
+using Toggl.Multivac;
+
+namespace Toggl.Foundation.MvvmCross.Collections
+{
+    public class WorkspaceGroupedCollection<T> : MvxObservableCollection<T>
```
arturs105: Shouldn't this be sealed?

Toggl.Daneel/ViewSources/StartTimeEntryTableViewSource.cs
willsb: Agreed

Toggl.Foundation.MvvmCross/Collections/WorkspaceGroupedCollection.cs
willsb: The idea is that on the next PR we'll inherit from it

PR #647, opened 2017-09-28
Make sure suggestion count isn't too high

### LINKED ISSUE #633: UI fixes: suggestions view
- [x] Suggestion view misses top separator
- [x] Suggestions view should end with the last suggested TE (the white are is redundant)
- [x] Determine suggestion count by the size of the screen

screen shot 2017-09-26 at 08 41 51

arturs105:
Closes #633 

Problem was that `SuggestionsTableViewSource ` needs to know the table view's height, but it isn't determined until `ViewDidLayoutSubviews`, so we reload the data, when the height is determined.


arturs105:
Actually, i'll split this into two PRs

PR #648, opened 2017-09-28
Remove white area below suggestions

arturs105:
Ref #633


Toggl.Daneel/ViewSources/SuggestionsTableViewSource.cs
```diff
@@ -26,14 +26,16 @@ public SuggestionsTableViewSource(UITableView tableView)
...
-            return new UIView(new CGRect(0, 0, totalWidth, 45))
+            var header = new UIView(new CGRect(0, 0, totalWidth, 45))
             {
                 new UILabel(new CGRect(16, 16, totalWidth - 16, 19))
                 {
                     Font = UIFont.SystemFontOfSize(16, UIFontWeight.Medium),
                     Text = Resources.SuggestionsHeader
                 }
             };
+            header.BackgroundColor = UIColor.White;
```
willsb: You could still set this with the object initializer syntax, any reason why you didn't?

arturs105: Tried, but didn't know exactly how.

willsb: Holy moly. Now that I look at it again, this is the collection initializer syntax. Nevermind, it's not possible

PR #649, opened 2017-09-28
Calculate suggestion count based on screen size

arturs105:
Ref #633

Problem was that `SuggestionsTableViewSource` needs to know the table view's height, but it isn't determined until `ViewDidLayoutSubviews`, so we reload the data, when the height is determined.


Toggl.Daneel/ViewSources/SuggestionsTableViewSource.cs
```diff
@@ -14,6 +15,11 @@ public sealed class SuggestionsTableViewSource : MvxTableViewSource
     {
         private const string cellIdentifier = nameof(SuggestionsViewCell);
         private const string emptyCellIdentifier = nameof(SuggestionsEmptyViewCell);
+        private const int headerHeight = 45;
+        private const int rowHeight = 64;
+
```
willsb: One extra newline (also, I like to keep the consts ordered by width, but that's my OCD)

Toggl.Daneel/ViewControllers/SuggestionsViewController.cs
```diff
@@ -25,19 +27,25 @@ public override void ViewDidLoad()
...
                       .To(vm => vm.IsEmpty)
                       .WithConversion(new MvxVisibilityValueConverter());
             
             bindingSet.Apply();
         }
+
+        public override void ViewDidLayoutSubviews()
+        {
+            base.ViewDidLayoutSubviews();
+            tableViewSource.ReloadTableData();
```
willsb: Is this needed? Reload is called whenever the collection changes

Toggl.Foundation/Suggestions/MostUsedTimeEntrySuggestionProvider.cs
```diff
@@ -12,7 +12,8 @@ namespace Toggl.Foundation.Suggestions
     public sealed class MostUsedTimeEntrySuggestionProvider : ISuggestionProvider
     {
         private const int daysToQuery = 42;
-        private const int maxNumberOfSuggestions = 5;
+        //Because biggest iphone(8+) can show this many
```
willsb: You can remove this comment, since this is platform agnostic code

Toggl.Daneel/ViewControllers/SuggestionsViewController.cs
arturs105: Yes. We need to reload data, when the actual size of the table view is known. When the data is first loaded (during the first calls to `SuggestionsTableViewSource.ReloadTableData`), the table view's size is 667, which is full screen.

willsb: I see. This approach feel hacky, but it's needed. I wonder, however: Isn't there a way to do this calculation inside the ViewSource? If there isn't we can leave it as is, but otherwise I'd prefer all the resizing logic to be in the TableViewSource

arturs105: The calculation is already done in the ViewSource, but it needs to be done done at the right time of the view controllers life cycle. This just tells it: "recalculate and repopulate it **now**".

I tried to see what method overrides in the view source might be useful for this, but didn't find any.

PR #650, opened 2017-09-28
Refactor CollectionViewSources

willsb:
Needed for a future PR

Refactored a lot of copy pasta into a single class. Much better now.

Squash ASAP


Toggl.Daneel/ViewSources/SelectProjectTableViewSource.cs
```diff
@@ -11,60 +11,30 @@
```
arturs105: `using`s on lines 5 and 5 have become unnecessary.

Toggl.Daneel/ViewSources/SelectProjectTableViewSource.cs
```diff
@@ -11,60 +11,30 @@
...
-            => getGroupAt(section).Count();
-
-        private WorkspaceGroupedCollection<ProjectSuggestion> getGroupAt(nint section)
-            => groupedItems.ElementAt((int)section);
-
-        protected override object GetItemAt(NSIndexPath indexPath)
-            => groupedItems.ElementAt(indexPath.Section).ElementAt((int)indexPath.Item);
-
-        private UITableViewHeaderFooterView getOrCreateHeaderViewFor(UITableView tableView)
+        protected override UITableViewHeaderFooterView GetOrCreateHeaderViewFor(UITableView tableView)
```
arturs105: Could the `GetOrCreateCellFor` and `GetOrCreateHeaderViewFor` methods be moved to the base class if `headerCellIdentifier` and `cellidentifier` are moved to the base class and made abstract?

Toggl.Daneel/ViewSources/SelectTagsTableViewSource.cs
```diff
@@ -11,60 +11,28 @@
```
arturs105: Some `using`s are also unnecessary.

Toggl.Daneel/ViewSources/SelectProjectTableViewSource.cs
willsb: Good point. Some might need overriding, but most won't. WIll do

PR #656, opened 2017-09-29
Persist tag edits

### LINKED ISSUE #519: Allow editing of tags in edit view
Refer to zeplin.

- [x] Create tag selection view
- [x] Persist tag changes
- [ ] Display tags in the edit view

arturs105:
Closes #519


Toggl.Foundation.Tests/MvvmCross/ViewModels/EditTimeEntryViewModelTests.cs
```diff
@@ -204,5 +207,75 @@ public void SetsTheStopTimeToTheValueReturnedByTheSelectDateTimeDialogViewModelW
...
+                    .Returns(Observable.Return(tags));
+                NavigationService
+                    .Navigate<long[], long[]>(Arg.Is(typeof(SelectTagsViewModel)), Arg.Any<long[]>())
+                    .Returns(Task.FromResult(tagIds));
+                await ViewModel.Initialize();
+
+                await ViewModel.SelectTagsCommand.ExecuteAsync();
+
+                await DataSource.Tags.Received()
+                    .GetAll(Arg.Any<Func<IDatabaseTag, bool>>());
```
simonrozsival: I think you should test the predicate here (`Arg.Is`) - that for tags with the given Ids it returns true and for different ids it returns false. Because if you don't do this, then the title of the test is misleading (the second part `ForReturnedTagIds` is not tested).

Toggl.Foundation.Tests/MvvmCross/ViewModels/EditTimeEntryViewModelTests.cs
```diff
@@ -204,5 +207,75 @@ public void SetsTheStopTimeToTheValueReturnedByTheSelectDateTimeDialogViewModelW
...
             }
         }
+
+        public sealed class TheSelectTagsCommand : EditTimeEntryViewModelTest
+        {
+            [Fact]
+            public async Task NavigatesToTheSelectTagsViewModelPassingCurrentTagIds()
+            {
+                long id = 13;
+                var tagIds = new long[] { 1, 43, 42, 10 };
```
simonrozsival: You could make this a property test with `long[] tagIds` as the parameter of the test.

Toggl.Foundation.Tests/MvvmCross/ViewModels/EditTimeEntryViewModelTests.cs
```diff
@@ -204,5 +207,75 @@ public void SetsTheStopTimeToTheValueReturnedByTheSelectDateTimeDialogViewModelW
...
+                    .Received()
+                    .Navigate<long[], long[]>(
+                        Arg.Is(typeof(SelectTagsViewModel)),
+                        Arg.Is<long[]>(ids => ids.SequenceEqual(tagIds)));
+            }
+
+            [Fact]
+            public async Task QueriesTheDataSourceForReturnedTagIds()
+            {
+                var tagIds = new long[] { 1, 42, 67, 21 };
```
simonrozsival: This could be a property test as well.

Toggl.Foundation.Tests/MvvmCross/ViewModels/EditTimeEntryViewModelTests.cs
simonrozsival: Maybe you could even make this part of the property test - get another array of longs and remove the items which are also in the `tagIds` array and test these for returning false.

Toggl.Foundation.Tests/MvvmCross/ViewModels/EditTimeEntryViewModelTests.cs
```diff
@@ -204,5 +207,75 @@ public void SetsTheStopTimeToTheValueReturnedByTheSelectDateTimeDialogViewModelW
...
+                await ViewModel.SelectTagsCommand.ExecuteAsync();
+
+                await DataSource.Tags.Received()
+                    .GetAll(Arg.Any<Func<IDatabaseTag, bool>>());
+            }
+
+            [Fact]
+            public async Task SetsTheReturnedTags()
+            {
+                var tagIds = new long[] { 1, 42, 67, 21 };
```
simonrozsival: Could be property test as well.

Toggl.Foundation.Tests/MvvmCross/ViewModels/EditTimeEntryViewModelTests.cs
arturs105: I tried to do this, but the problem is that we need to compare a `Func<IDatabaseTag, bool>` and the one in the view model uses the private field `tagIds`.

I could initialize the view model to some tag ids and check, whether the `Func` in `Arg.Is` behaves correctly according to those tags.

arturs105: I mean i can test, whether the `func` returns true for all the right values and false for some other values. Or is there a better way to do this?

simonrozsival: Yes, that is what I meant. I don't think there is a better way. If you don't hardcode the values, but use property testing, then it should be perfectly OK.

PR #665, opened 2017-10-02
Support tasks in autocomplete

### LINKED ISSUE #526: Support Tasks in creation view

### LINKED ISSUE #527: Support tasks in edit view

willsb:
Closes #526


Toggl.Foundation.Tests/Autocomplete/AutocompleteProviderTests.cs
```diff
@@ -183,9 +206,22 @@ public sealed class QueriesTheDatabaseForTimeEntries : AutocompleteProviderTest
...
                 [Fact]
+                public async Task SearchesTheTaskName()
+                {
+                    const string description = "21";
+                    var textFieldInfo = TextFieldInfo.Empty.WithTextAndCursor(description, 0);
+
+                    var suggestions = await Provider.Query(textFieldInfo);
+
+                    await Database.TimeEntries.Received().GetAll();
+                    suggestions.Should().HaveCount(1)
```
arturs105: I think we should check, whether the suggestion has a task with description "21".

arturs105: Upon closer inspection: `TimeEntrySuggestion` does not have any info about tasks, so disregard this.

arturs105:
> according to Zeplin, the down arrow, which is shown next to the number of tasks of a project suggestion, should change to up arrow (or rotate 180 deg) when the list of tasks is expanded

This would require quite a lot of refactoring (making this PR even bigger), so i think it would be better if it was done in another PR.

> When I write the task name the time entry which has this task is show as suggestion. If I click on it, the project and the text is used, but if I save it, the new time entry does not have the task set, only the project

Yes, will fix this

pe0ny:
> When I search by project ("@..."), I cannot search by the task name

Yes, it should search tasks as well.

> When I select some task, the token shows only the project although I would expect it to be "project: task". When I save it, it is OK though.

Yes. It should be 'Project: task'.

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
```diff
@@ -114,22 +117,31 @@ private void selectSuggestion(AutocompleteSuggestion suggestion)
...
+                        timeEntrySuggestion.ProjectId.Value,
+                        timeEntrySuggestion.ProjectName,
+                        timeEntrySuggestion.ProjectColor,
+                        timeEntrySuggestion.TaskId.Value,
+                        timeEntrySuggestion.TaskName);
                     break;
 
                 case ProjectSuggestion projectSuggestion:
-
+
```
simonrozsival: These whitespaces shouldn't be here.

PR #668, opened 2017-10-03
Fix pushing running time entry

simonrozsival:
After discussion with Eric from backend I found out, that the duration of a running time entry is the negative "current timestamp" (the toleration is something more than 40 days).

I added a property test into the integration tests. This makes them run longer, but not dramatically longer.

Squash whenever.


Toggl.Ultrawave.Tests.Integration/TimeEntriesApiTests.cs
```diff
@@ -92,6 +92,57 @@ public sealed class TheCreateMethod : AuthenticatedPostEndpointBaseTests<ITimeEn
...
+                };
+
+                var persistedTimeEntry = CallEndpointWith(togglApi, timeEntry).Wait();
+
+                persistedTimeEntry.Id.Should().BePositive();
+            }
+
+            [Theory]
+            [InlineData(1000)]
+            [InlineData(-1000)]
```
amulware: So a start date more than 999 hours in the future is forbidden too? That seems really arbitrary. Do I understand this correctly?

simonrozsival: This is what I observe. I would like to stick to some manual and do it consistently, but I have to do this by trial and error.

For our purposes this test is sufficient I believe. The backend has to have its own tests to test this "feature".

PR #680, opened 2017-10-04
Do not show old time entries in log after pushing if ID changed

### LINKED ISSUE #676: Fix pushed time entry appearing twice in log
This is caused by the ID change of the entity not being communicated to the UI.

I think the best solution is to instead of a tuple with a conflict resolution enum, send a property typed value that has different fields depending on whether it's a Delete (id), Create (entity) or Update (old id, new entity).

simonrozsival:
Closes #676.

This PR contains changes in the repository and Realm adapter and also in the view models.

The BatchUpdate now does not return tuples, but rather the `IResult<T>` interface which is implemented by these classes: `UpdateResult<T>`, `CreateResult<T>`, `DeleteResult<T>`, `IgnoreResult<T>`. 

The view model is changed to reflect these changes. Thanks to this the glitches in the UI (old unsynced items after push) are gone.

Squash whenever.


Toggl.Foundation/DataSources/TimeEntriesDataSource.cs
```diff
@@ -121,32 +122,32 @@ public IObservable<IDatabaseTimeEntry> Create(IDatabaseTimeEntry entity)
...
-                case ConflictResolutionMode.Create:
-                    timeEntryCreatedSubject.OnNext(tuple.Entity);
+                case CreateResult<IDatabaseTimeEntry> c:
+                    timeEntryCreatedSubject.OnNext(c.Entity);
                     return;
 
-                case ConflictResolutionMode.Update:
-                    timeEntryUpdatedSubject.OnNext(tuple.Entity);
+                case UpdateResult<IDatabaseTimeEntry> u:
+                    timeEntryUpdatedSubject.OnNext((u.OriginalId, u.Entity));
```
amulware: What is the reasoning being passing on these lower level fields instead of the result object?

Toggl.Foundation/Sync/States/Pull/BasePersistState.cs
```diff
@@ -43,7 +42,7 @@ public IObservable<ITransition> Start(FetchObservables fetch)
                 .Select(entities => entities.Select(ConvertToDatabaseEntity).ToList())
                 .SelectMany(databaseEntities =>
                     repository.BatchUpdate(databaseEntities.Select(entity => (GetId(entity), entity)), conflictResolver.Resolve, rivalsResolver)
-                        .Select(results => results.Select(result => result.Item2))
```
amulware: Is this Select here only for the cast? If so, is there no `Cast<>()` extension that could be used after IgnoreElements?

Toggl.PrimeRadiant.Realm/Database.cs
```diff
@@ -18,7 +18,10 @@ public Database()
             Projects = Repository<IDatabaseProject>.For((project, realm) => new RealmProject(project, realm));
             TimeEntries = Repository<IDatabaseTimeEntry>.For((timeEntry, realm) => new RealmTimeEntry(timeEntry, realm));
             Workspaces = Repository<IDatabaseWorkspace>.For((workspace, realm) => new RealmWorkspace(workspace, realm));
-            WorkspaceFeatures = Repository<IDatabaseWorkspaceFeatureCollection>.For((collection, realm) => new RealmWorkspaceFeatureCollection(collection, realm), id => x => x.WorkspaceId == id);
+            WorkspaceFeatures = Repository<IDatabaseWorkspaceFeatureCollection>.For(
+                (collection, realm) => new RealmWorkspaceFeatureCollection(collection, realm),
+                id => x => x.WorkspaceId == id,
+                features => features.WorkspaceId);
```
amulware: What does this change do? Does it belong here?

Toggl.PrimeRadiant.Realm/RealmAdapter.cs
```diff
@@ -75,29 +83,30 @@ public TModel Update(long id, TModel entity)
...
 
                 foreach (var entityWithPotentialRival in entitiesWithPotentialRival)
                 {
-                    resolvePotentialRivals(realm, entityWithPotentialRival, rivalsResolver, resolvedEntities);
+                    resolvePotentialRivals(realm, entityWithPotentialRival, rivalsResolver, results);
                 }
 
                 transaction.Commit();
-                return resolvedEntities;
+                return results.Cast<IResult<TModel>>();
```
amulware: Is this cast needed?

Toggl.PrimeRadiant/UpdateResult/IResult.cs
```diff
@@ -0,0 +1,6 @@
+﻿namespace Toggl.PrimeRadiant.UpdateResult
+{
+    public interface IResult<T>
```
amulware: Is IResult as good a name as possible? It seems very general, but I'm not sure I can think of anything more specific that isn't UpdateResult which would get confusing. (Already, having namespace and type of the same name is maybe not the best.)

Toggl.PrimeRadiant.Realm/Database.cs
simonrozsival: Yes, the signature of the `For` method changed (the last parameter was added), so this has to change as well (`RealmWorkspaceFeatureCollection` does not implement `IBaseModel`).

Toggl.PrimeRadiant/UpdateResult/IResult.cs
simonrozsival: It was originally called `IUpdateResult`, but then this was confusing. I will think of a better name for the interface.

I will give the namespace name a thought as well.

Toggl.Foundation.MvvmCross/ViewModels/TimeEntriesLogViewModel.cs
```diff
@@ -106,14 +107,14 @@ private void addTimeEntries(TimeEntryViewModelCollection collection)
             RaisePropertyChanged(nameof(IsEmpty));
         }
 
-        private void onTimeEntryUpdated(IDatabaseTimeEntry timeEntry)
+        private void onTimeEntryUpdated((long Id, IDatabaseTimeEntry Entity) timeEntry)
```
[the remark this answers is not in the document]
simonrozsival: The original thought was that I will do the `RealmAdapter` part and William will do the rest, but I had to do the whole thing because we didn't manage to finish it in Tallinn. I tried to change as little code in the view model to avoid breaking many things by accident. But I can 🔥 this now if it is not used and it can be brought back when needed later. I will refactor this.

Toggl.Foundation/DataSources/TimeEntriesDataSource.cs
simonrozsival: OK, so I gave this a thought and I think that it is not a very good idea to drag the `UpdateResult` into the observers - in this case the `TimeEntryLogViewModel`. The update method needs the ID of the original entity and then the new value of the entity, which is what this tuple holds and it is the same what the `UpdateResult<T>` holds.

The main reason is that these _results_ are meant for conflict resolution, but it would now have to be used in other places (e.g., `TimeEntriesDataSource.Update`). We use tuples all around the source, so I don't see any problem using the tuple here.

amulware:
Now that I see why the cast removed in `a0166f59a5e3b518d43d890f0f3dbe5db5d248ea` was necessary, I'm not sure I agree with that commit. It now partly works in a more abstract type, but then hast to cast back down in one place, which seems odd. Do you have an argument for why this is better? Otherwise I'd propose reverting it, but instead of casting at the end, casting when constructing the list in the first place. I think you can simply do `.ToList<IResult<TModel>>()`, or otherwise put the cast before the `ToList()`.

Toggl.Foundation.MvvmCross/ViewModels/TimeEntriesLogViewModel.cs
```diff
@@ -106,17 +107,8 @@ private void addTimeEntries(TimeEntryViewModelCollection collection)
...
-            if (collection == null) return;
-
-            var viewModel = collection.SingleOrDefault(vm => vm.Id == timeEntry.Id);
-            if (viewModel == null) return;
-
-            collection.Remove(viewModel);
-            RaisePropertyChanged(nameof(IsEmpty));
-        }
+        private void onTimeEntryUpdated((long Id, IDatabaseTimeEntry Entity) timeEntry)
+            => onTimeEntryDeleted(timeEntry.Id);
```
amulware: I'm confused about this now. The method says it updates, but it only deletes and ignores the other argument. If that is on purpose, I don't see any reason to have this method in the first place. But shouldn't it also insert the `Entity` back into the list? Or in the best case actually replace the old one?

simonrozsival: This is on purpose. This is just a slightly updated view model logic written by William. I first thought about removing the `onTimeEntryUpdated` method altogether and calling the `onTimeEntryDeleted` directly, but then it seemed even more confusing to me. I would suggest keeping this the way it is right now and creating an issue for refactoring this view model, because I think this is already out of the scope of this PR.

simonrozsival:
I will revert the commit.

Issue #322, opened 2017-07-31
Animations: Card - stopping a running entry


pe0ny:
**Stopping a running time entry**
Stop button scales down and entry moves away with the card.

![card-running-stop](https://user-images.githubusercontent.com/5064445/28776927-967e6d36-7601-11e7-95bd-46a4ce0f51ee.gif)

```javascript
toggl.stop.onTap ->
	hideIcon(toggl.stop)
	hideEntry()
	hideCard()

hideIcon = (icon) ->
	icon.animate
		scale: 0
		options:
			time: LeaveTimingFaster
			curve: EaseIn

hideEntry = () ->
	toggl.entry.animate
		opacity: 0
		y: 568
		options:
			time: LeaveTiming
			curve: EaseIn

hideCard = () ->
	toggl.card.animate
		y: 568
		options:
			time: LeaveTiming
			curve: cardOutCurve
```

willsb:
I'd like this to be on hold until I return from vacation, for I have a WIP branch [here](https://github.com/toggl/mobileapp/tree/stop-te-animation-wip). If it can't wait, please try and base your solution on what I have written.

PR #684, opened 2017-10-04
Use new iOS 10 curves

willsb:
Self-explanatory. This is needed for #322, otherwise the animations would conflict.

Squash whenever

### DIFF
--- a/Toggl.Daneel/Binding/BarButtonAnimatedEnabledTargetBinding.cs
+++ b/Toggl.Daneel/Binding/BarButtonAnimatedEnabledTargetBinding.cs
@@ -20,19 +20,18 @@ namespace Toggl.Daneel.Binding
 
         protected override void SetValue(bool value)
         {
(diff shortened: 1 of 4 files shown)


Toggl.Daneel/Binding/BarButtonAnimatedEnabledTargetBinding.cs
```diff
@@ -20,19 +20,18 @@ public BarButtonAnimatedEnabledTargetBinding(UIBarButtonItem target)
...
-
-            CATransaction.Commit();
+            AnimationExtensions.Animate(
+                Animation.Timings.EnterTiming,
+                Animation.Curves.SharpCurve,
+                () =>
+                {
+                    //This slight color change is needed for the fade animation to work.
+                    //Animating only the Enabled property has no effect.
+                    //Blame Apple.
```
amulware: I like spaces after `//`

PR #686, opened 2017-10-09
Add sync error message to edit time entry view

### LINKED ISSUE #531: Show failed sync in time entry edit view
Also show the message returned from backend for items that failed to sync to allow the user to fix the problematic data. Ask @pe0ny for design.

arturs105:
Closes #531 

This adds the the error view without any animations.


Toggl.Foundation.Tests/MvvmCross/ViewModels/EditTimeEntryViewModelTests.cs
```diff
@@ -310,5 +310,27 @@ private IDatabaseTag createTag(long id)
                 return tag;
             }
         }
+
+        public sealed class TheDismissSyncErrorMessageCommand : EditTimeEntryViewModelTest
```
simonrozsival: Please add also tests which will check that the `SyncErrorMessageVisible` is true when there is some error message and that it equals what it should equal and false when it is null or empty.

PR #699, opened 2017-10-09
Avoid updating running time entries which are never in the log view

### LINKED ISSUE #687: TE disappearing
Reproducing steps:
1) Install the app from scratch. use test.mobile.1 account to log in
2) Start a new TE, type in ‘test’ and pick a suggestion, tap green check
3) Open running TE edit view
4) change start time, save; confirm changes
5) Open running TE edit view
6) add tags, save; confirm changes
7) you should be on timer page now; stop TE (it disappears)

simonrozsival:
Closes #687 

This bugfix contains only one line. Editing a running time entry lead to freezing an observable in the time entries log, where there are no items for the running time entry. Filtering the running time entries out solves the problem. Please do a functional test and try reproducing the bug @pe0ny reported in the issue and also try different scenarios. The item should always appear in the log when it is stopped and changing other stopped items should update their items in the log as well.

Squash whenever.


simonrozsival:
I don't think the observable breaks when you _stop_ the time entry, but it rather breaks after the first _updated_ of the running time entry. The `Where(isNotRunning)` only does the thing that it filters out the time entries with `Stop == null`. So that is what is needed. We can add an _if_ to the `onTimeEntryCreated` and that will do the job equally well. What do you think is better? I would probably stick to this solution with observables filtering the invalid items.

PR #701, opened 2017-10-09
Fix CollectionToStringValueConverter

arturs105:
Edit view was previously silently throwing exceptions if there were no tags. Fixed now.


Toggl.Foundation.MvvmCross/Converters/CollectionToStringValueConverter.cs
```diff
@@ -10,10 +10,11 @@ namespace Toggl.Foundation.MvvmCross.Converters
     public sealed class CollectionToStringValueConverter<T> : MvxValueConverter<IEnumerable<T>, string>
     {
         protected override string Convert(IEnumerable<T> value, Type targetType, object parameter, CultureInfo culture)
-            => value.Aggregate(
```
amulware: Since it's an IEnumerable, please use `!Any()` instead of `Count() == 0` to prevent enumeration of the whole thing. Alternatively, consider turning this into a list instead to prevent double enumeration. Also, consider using `??` and `?.` in case it makes the syntax clearer.

amulware: Oh, actually, if we are gonna turn it into a list, might as well not check the count at all instead and use the string builder as before (it starts out and will result in an empty string for an empty collection).

PR #703, opened 2017-10-11
Change the generic type to avoid bad casts

### LINKED ISSUE #697: Trying to edit project in edit view crashes app

simonrozsival:
Closes #697 

The problem was that the cast in `GroupedCollectionTableViewSource<T>` on line 20 was invalid before this fix (but the `as` keyword does not throw, it rather returns `null`, so the origin of the bug wasn't very obvious).

Squash whenever.


arturs105:
Looks like this is already solved in `develop`.

simonrozsival:
Interesting, if it is so, then I will close this and the issue. I will check it out.

simonrozsival:
I confirm this was already solved by some other pull requests, even the "Files changed" now shows 0. Therefore I close this pull request and this issue.

Issue #708, opened 2017-10-12
App crashes when stopping a TE


pe0ny:
1) Start a new time entry or continue existing one
2) Tap stop
3) App crashes

simonrozsival:
Do you see the time entry in the log with some weird large negative duration? I just want to confirm that this is the very same bug I am observing.

simonrozsival:
The bug was fixed by the backend team yesterday. There is a related bug though (issue #714) which also happens after clicking the stop button, so I will leave this issue open until the other is closed.

PR #713, opened 2017-10-12
Add integration tests for validation of the data returned from the API for running time entries

simonrozsival:
Ref #708 

Run integration tests locally when reviewing. Squash whenever.


amulware:
Please keep the squashed commit message a bit shorter. We've had a number of messages recently that were too long for github's summary page.

PR #720, opened 2017-10-13
Enlarge tappable area

### LINKED ISSUE #636: Make tappable area in the edit view bigger
Currently projects, duration, start, tags and billable can only be edited by tapping exactly on the respective label/control, however the whole row in the edit view should be tappable.

kornelijepetak:
Closes #636 

However, two things should be noted:

1. When _Billable_ switch is tapped, it toggles with animation. If the row is clicked, but not on the switch itself, the switch button changes state but without animation. Couldn't find out how to do it with animation since the switch is bound to Billable property and changes instantly with prop change. I don't know if that's a problem at this stage (maybe a new issue?)...

2. The only (non-programmatic) way I found for two labels to be centered as a group, regardless of their size is to place same-width paddings to the left and right of them. I've seen this done this way in lots of tutorials and haven't seen any other way yet. I don't think having two or three empty views will degrade performance.

Any comments?


simonrozsival:
The lack of animation is very noticeable and annoying - I think you need to use this method to change the state of the `UISwitch` instead of just binding the boolean value of `On`: https://developer.xamarin.com/api/member/MonoTouch.UIKit.UISwitch.SetState/p/System.Boolean/System.Boolean/ - this would probably require writing a custom binding. But definitely discuss with @arturs105, he may have a better idea.

arturs105:
Yes, what Simon said. I quickly tried out the `SetState(bool, bool)` method and it works with an animation. You should create the target binding.

kornelijepetak:
I have addressed the binding issue, now it works as it should, it correctly animates.

I'll try to push another change with simpler UI.

kornelijepetak:
The UI was simplified (no more hackish padding elements).

Toggl.Daneel/Binding/SwitchAnimatedOnTargetBinding.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System;
```
arturs105: Unnecessary `using`

Toggl.Daneel/Binding/SwitchAnimatedOnTargetBinding.cs
```diff
@@ -0,0 +1,24 @@
...
+using MvvmCross.Binding;
+using MvvmCross.Binding.Bindings.Target;
+using UIKit;
+
+namespace Toggl.Daneel.Binding
+{
+    public class SwitchAnimatedOnTargetBinding : MvxTargetBinding<UISwitch, bool>
```
arturs105: Class not sealed

Toggl.Foundation.MvvmCross/ViewModels/EditTimeEntryViewModel.cs
```diff
@@ -276,5 +279,10 @@ private void onTags(IEnumerable<IDatabaseTag> tags)
 
         private void dismissSyncErrorMessageCommand()
             => SyncErrorMessageVisible = false;
+
+        private void toggleBillable() {
```
arturs105: Egyptian style brackets

Toggl.Foundation.MvvmCross/ViewModels/EditTimeEntryViewModel.cs
```diff
@@ -276,5 +279,10 @@ private void onTags(IEnumerable<IDatabaseTag> tags)
...
+            Billable = !Billable;
+        }
+
```
arturs105: Empty line

PR #727, opened 2017-10-13
Create base push entity state

simonrozsival:
I want to put the shared code into `BasePushEntityState<TModel>` to prevent more code duplication: Create and Update states are PushEntity states - they have the `Overwrite` method and most notably the `Fail` method which results in 3 different transitions which are necessary for the retry loop.

I want to use this for the `Delete` state, which will have the same code - ~~overwrite and~~ fail. This could prevent some errors in the future.

Squash whenever.


amulware:
Does/Will the overwrite support deletion?

simonrozsival:
Deletion won't need overwrite.

amulware:
The description above makes it sound like it does, but no matter. All in due time.

simonrozsival:
The PR #680 must be merged before this one - a(n ugly) manual merge will be necessary.

Toggl.Foundation/Sync/States/Push/BaseCreateEntityState.cs
```diff
@@ -7,47 +7,27 @@
...
         {
-            this.api = api;
-            this.repository = repository;
         }
 
-        public IObservable<ITransition> Start(TModel entity)
+        public override IObservable<ITransition> Start(TModel entity)
             => create(entity)
-                .SelectMany(overwrite(entity))
+                .SelectMany(Overwrite(entity))
```
amulware: I note that this overwrites without conflict resolution. Are we sure that is correct and that conflict resolution is never needed?

simonrozsival: So this PR is just a refactoring of the old code - there is no new logic, this is how it works from the start.

But to answer your question, this is correct - whatever is returned as the response to the POST request is written to the database without even looking at what is underneath - POST is law. No conflict resolution was ever intended to be here.

simonrozsival:
All these changes are done only to change the structure of the code and should not introduce any new behavior or changes in the current behavior or the business logic.

simonrozsival:
In fact, the changes in the tests are rather "unnecessary" and the original tests should pass as well. Although I don't think that having many duplicate tests is good. More so when there should be another set of tests which would be basically just a copy/paste of these tests.

Toggl.Foundation/Sync/States/Push/BaseCreateEntityState.cs
amulware: Makes sense. Thanks.

amulware:
That merge is up to you. If you like I can review again afterwards.

simonrozsival:
There was nothing dramatic about the merge after all, just two conflicts, so I don't think that there is a reason for a repeated review. Once the tests finish with an OK, I will squash this.

Issue #688, opened 2017-10-09
Push time entry deletion


amulware:
We use a local bool flag to indicate that a time entry was deleted. If this is pushed successfully, the entity can be deleted from the database entirely.

simonrozsival:
This is now waiting for PR #727 to be merged.

simonrozsival:
Now this is waiting for PR #734

simonrozsival:
There are issues for the remaining imperfections of the whole deletion - #689 and #751, so the PR #752 should have closed this issue, instead of just referencing it. Therefore I close this issue manually now.

PR #734, opened 2017-10-16
Add deleting of time entries to the API client

simonrozsival:
Ref #29, Needed for #688, Ref #736

_A note on integration tests:_ The test in the `TheGetAllMethod`, which might seem that it does not belong to this PR, was added because of the assertion in the same file on line `332`. On the first iteration I assumed that I will get an empty list, but I got a `null` instead. The backend team works on replacing these `null`s from the API and as this might change in the future, we should not only get a failing test for the `Delete` method, but also one for the `GetAll` method, which will be the real cause of that failed test.

Squash whenever.


Toggl.Ultrawave.Tests.Integration/TimeEntriesApiTests.cs
```diff
@@ -304,6 +315,79 @@ private IObservable<ITimeEntry> CallEndpointWith(ITogglApi togglApi, TimeEntry t
...
+            [Fact, LogTestInfo]
+            public async Task FailsIfDeletingANonExistingTimeEntryIsAttempted()
+            {
+                var (togglApi, user) = await SetupTestUser();
+                var timeEntry = createTimeEntry(user);
+                var persistedTimeEntry = await togglApi.TimeEntries.Create(timeEntry);
+
+                Action deleteNonExistingTimeEntry = () => togglApi.TimeEntries.Delete(new TimeEntry { Id = persistedTimeEntry.Id == 1 ? 2 : 1 }).Wait();
+
+                deleteNonExistingTimeEntry.ShouldThrow<ForbiddenException>();
```
amulware: Will this always be Forbidden? Or only if you specify an ID in a different workspace? What happens if we have it try to delete a time entry with the same id as the persisted one, plus 100000 or so?

simonrozsival: I will make separate tests for a time entry with the correct `Id`, but wrong `WorkspaceId`, the other way round and both incorrect.

I don't know what would we test if we did the + 100..00 - if backend handles deserialization and int64 overflow well? I don't think such a test is necessary/usefull.

amulware: Backend hands out time entry ids incrementally. +some large number will basically guarantee that the id is not an existing time entry, while otherwise, we cannot know for certain. I suppose another way would be to test with negative numbers? Though I'm not sure if that tests exactly the same.

My main point is that there is a big difference between a time entry in a different workspace, and a time entry that does not exist at all.

simonrozsival: I can now see your point. But I don't think that there is a case where I deleted a time entry with some specific `Id` and I wasn't sure if there was some other time entry which could accidentally have this id and could be affected by this operation.

I added more tests in `64d0138` which test both incorrect workspace and non-existing id. Isn't that sufficient?

simonrozsival:
Hmm. I see. I will then do the +100000 thing, it makes sense. And I will also add a test for deleting an existing time entry but of different user in a workspace which is inaccessible to the first user.

PR #738, opened 2017-10-17
Fix a cast which causes InvalidCastException

simonrozsival:
Sync does not work right now. This fixes it. Ref #724.

Squash whenever.


amulware:
What exactly was the problem?

simonrozsival:
Line 98 throws.

amulware:
Could this alternatively be fixed by adding an `out` to the interface's generic parameter?

simonrozsival:
I think the casts between `TModel` and `TRealmEntity` are now justifiable because they only happen in the private methods and we know exactly what type is in each field.

amulware:
One thing I note though: We didn't notice this failing before. Is this entirely because we do not have tests for realm? If so... not much we can do right now, but still, dangerous.

simonrozsival:
We don't have tests for RealmAdapter. In fact I don't know how that would be possible to do. We discussed it with @willsb and there is some way of doing integration tests for realm but we haven't done it yet.

This bug was introduced yesterday in #724, I did a poor job of testing the PR's functionality in the simulator - the sync indicator shows success, but it is a lie.

I am happy with this fix, so I will squash it.

Issue #751, opened 2017-10-19
Better handle 4xx http errors


amulware:
There could be many reasons for that, and we should try and handle all we can detect appropriately.

(list to be compiled)

amulware:
The errors we need to handle are:

410: Stops sync, and shows a special screen preventing the user from interacting with the app until they update
418: Same as 410, but a different screen
401: Same as 410/418 but the screen allows the user to re-login (with the same email). If that login succeeds, we can overwrite the user object with the new api key, reinitialise the api and maybe data sources, and let the user back into the app. Otherwise, the user also has the option to log out which may cause data to not be synced.

In all three cases, if the user closes and reopens the app, it should NOT start sync, and instead immediately show the respective screen.

We need to:
- figure out how to perform these three checks for every single api call (maybe using a custom rx extension?), and then stop the sync man each time (hopefully without needing extra transitions everywhere)
- figure out how to get this data out of the sync man, probably using one or several observables (injected into the states?), so we can show the respective screens
- store this data locally so it can be considered when the user relaunches the app
- figure out how the stored 410 and 418 can be invalidated by updating the app (maybe store the offending version number and compare against the current one?)

It's quite possible I'm missing further complications, but this should cover most of it.

amulware:
Before implementing this, I'd like a design here that covers all the above, and any other (corner) cases that we can find for discussion.

simonrozsival:
> figure out how to perform these three checks for every single api call

- I'm not sure if it is a good idea to create Rx extensions. All of our HTTP calls go through the `BaseApi` class and so we should catch these errors (client exceptions) there and persist the information about the problem and then somehow notify the Foundation.MvvmCross layer so it navigates the user to the desired view

    - this way we don't even have to do anything in the sync manager. These errors shouldn't be caught by any state and so the sync manager will error out and will stop syncing (but maybe we should freeze it automatically then?)

amulware:
If I'm not mistaken, right now any 400 error will cause the entity to be marked as unsyncable and the sync to proceed, so we still have to catch these errors here separately there anyways to make sure the sync is aborted, right?

If so, I see no reason to complicate the API. In general, I don't like that either, it's already passing the exception on correctly, what more can it do? Everything else should be up to whoever calls it.

Once we make these errors abort and get out of the sync man though, maybe they can be exposed as the 'reason' for the Failed event in #661, and foundation/ui can check that reason and determine how to react to it?

simonrozsival:
[...]

PR #752, opened 2017-10-19
Delete time entry state

simonrozsival:
Ref #688 

There will be another PR which will change the behavior of the way `ClientError` transition is handled by the `TogglSyncManagerFactory` to address 4xx errors in compliance with #751 

Squash whenever


Toggl.Foundation.Tests/Sync/States/Push/BaseDeleteEntityStateTests.cs
```diff
@@ -0,0 +1,162 @@
...
+        [Fact]
+        public void ReturnsFailTransitionWhenDatabaseOperationFails()
+            => helper.ReturnsFailTransitionWhenDatabaseOperationFails();
+
+        [Fact]
+        public void ReturnsSuccessfulTransitionWhenEverythingWorks()
+            => helper.ReturnsSuccessfulTransitionWhenEverythingWorks();
+
+        [Fact]
+        public void CallsDatabaseDeleteIsWithCorrectParameter()
```
amulware: Typo? Grammar? Remove 'Is'? (also in the called method)

simonrozsival:
I noticed a problem with deserialization of the "" body returned by the API to the DELETE request which doesn't happen when testing it by running unit tests, but only in simulator. I would like you to look at the two commits I did after your approval.

amulware:
The second commit looks fine. The first confuses me a bit. I didn't realise we were using data sources instead of the database directly anywhere. How come? What is the reasoning here?

I rather have the syncing not depend on the data sources if possible (it's sort of a circular dependency).

simonrozsival:
Well the reasoning is simple - when we use the repository directly, the UI does not update after changes, when we use it, it does. We came to this solution with @willsb and I think it is a good one. Deleting in the datasource however works differently - it only sets the flag. With syncing the repository must be used to work properly, moreover the UI does not need any updating because the time entry should already be hidden (although this does not work right now, but that is a different issue).

amulware:
I see. I originally thought the data sources would subscribe to changes from the sync manager, instead of having dependencies both ways.

I'd still like to go that way at some point, but right now that's not part of this.

Feel free to squash.

PR #753, opened 2017-10-19
Resolve delete client errors

### LINKED ISSUE #751: Better handle 4xx http errors

simonrozsival:
Closes #751 

This PR implements the "bottom left box" of the internal flowchart for the delete operation.

On-hold: Waitng for more information in the issue.

### DIFF
--- a/Toggl.Foundation.Tests/Sync/States/Push/BaseResolveDeleteClientErrorStateTests.cs
+++ b/Toggl.Foundation.Tests/Sync/States/Push/BaseResolveDeleteClientErrorStateTests.cs
@@ -1,0 +1,53 @@
+﻿using System;
+using System.Linq;
+using System.Reactive.Linq;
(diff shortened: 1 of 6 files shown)


amulware:
I don't see how we can implement #751 until we have workspace preferences implemented which will take weeks at least. The best we can do is assume that 4xx errors are fine after all.

As for 429, I think we should think about how to handle that in general, not just in this specific case.

I'm not sure if we want to merge this as is before considering how to handle client errors with the api in general

simonrozsival:
I agree that we don't have to rush this and keep this on-hold until the specification is complete.

simonrozsival:
Well this is in fact so old, that I entirely forgot about this. I think it is superseded by #1048 and we can close this PR.

PR #754, opened 2017-10-19
Delete time entries only locally state

simonrozsival:
This handles time entries which were created in the app, weren't pushed to the server and were deleted later. Also will be used as a step after #753.

Squash whenever.


Toggl.Foundation/Sync/States/Push/BaseDeleteLocalEntityState.cs
```diff
@@ -0,0 +1,33 @@
...
+
+        public BaseDeleteLocalEntityState(IRepository<TModel> repository)
+        {
+            Repository = repository;
+        }
+
+        public IObservable<ITransition> Start(TModel entity)
+            => delete(entity)
+                .Select(_ => Deleted.Transition())
+                .Catch(Observable.Return(DeletingFailed.Transition()));
```
amulware: This will create the fail-observable and transition even if it isn't needed, right? Not sure if that is something we should worry about, but it seems like a waste.

PR #761, opened 2017-10-20
Ensure one running time entry when starting

### LINKED ISSUE #690: Ensure single running time entry when starting timer
This has to happen inside a database transaction to ensure there are no race conditions.

If necessary we have to make sure this also works when continuing or starting a timer from suggestions, but ideally, we only have to have this logic in one place.

simonrozsival:
Closes #690 

The `Create` method of the DataSource is internally used in both starting and continuing a time entry and will be used also by continuing from suggestions.

Squash whenever.


Toggl.Foundation.Tests/DataSources/TimeEntriesDataSourceTests.cs
```diff
@@ -106,20 +106,35 @@ public void ThrowsIfAnyOfTheArgumentsIsNull(bool useIdProvider, bool useReposito
...
             [Fact]
             public async ThreadingTask CreatesANewTimeEntryInTheDatabase()
             {
                 await TimeEntriesSource.Start(CreateDto(ValidTime, ValidDescription, true, ProjectId));
 
-                await Repository.Received().Create(Arg.Any<IDatabaseTimeEntry>());
+                await Repository.Received().BatchUpdate(
+                    Arg.Any<IEnumerable<(long, IDatabaseTimeEntry)>>(),
+                    Arg.Any<Func<IDatabaseTimeEntry, IDatabaseTimeEntry, ConflictResolutionMode>>(),
+                    Arg.Any<IRivalsResolver<IDatabaseTimeEntry>>());
```
amulware: There is so much duplication in these tests. Maybe this can be abstracted to read nicer (mostly using default values for the things that are just `Arg.Any()`)?

PR #763, opened 2017-10-20
Clean models

### LINKED ISSUE #709: Verify all models have correct types
This is important to do now because once we release the app publicly, we will have to start implementing database migrations for each change to our models.

amulware:
Closes #709

This is to compile all changes we still want to make to the models before the first public release. Afterwards we will be less flexible and have to write migrations for all changes to the realm models.

For now I’m only editing the interfaces so we can discuss those changes easier before regenerating and fixing implementations and tests.


amulware:
It seems to me like the models are mostly looking quite good. I'd like at least two other people to look over the model interfaces and see if there is anything that does not make sense or could be improved.

arturs105:
It looks like the tests are failing because `ModelConstructors.cs` wasn't updated.

amulware:
Please see the original comment. ;)

arturs105:
I couldn't find any other properties that should be removed. But i think someone who knows business logic and mobile app's planned feature set better should take look at the models, so @pe0ny.

pe0ny:
Which models and from where I should check?

arturs105:
@pe0ny Look at files in `Toggl.Multivac/Models` and see if all properties will be needed and if it makes sense to have them. You can probably skip `IBaseModel` and workspace feature stuff. Note that Paul has already removed some properties in [this PR](https://github.com/toggl/mobileapp/pull/763/files).

simonrozsival:
What about these project properties:
```csharp
bool? Template { get; }

bool? AutoEstimates { get; }

int? EstimatedHours { get; }

double? Rate { get; }

string Currency { get; }

int? ActualHours { get; }
```
Are/will these fields be supported by the app?

simonrozsival:
Same question, `ITask`:
```csharp
long EstimatedSeconds { get; }

long TrackedSeconds { get; }
```

simonrozsival:
Same question, `IWorkspace`:
```csharp
bool Admin { get; }

DateTimeOffset? SuspendedAt { get; }

double? DefaultHourlyRate { get; }

string DefaultCurrency { get; }

bool OnlyAdminsMayCreateProjects { get; }

bool OnlyAdminsSeeBillableRates { get; }

bool OnlyAdminsSeeTeamDashboard { get; }

bool ProjectsBillableByDefault { get; }

int Rounding { get; }

int RoundingMinutes { get; }

string LogoUrl { get; }
```

simonrozsival:
Other than that the models seem good to me, although I didn't check them against the JSON responses from the API for any missing fields.

amulware:
None of the mentioned fields are used right now, but I can see us wanting to perhaps use them in the future. A lot of them relate to reports in some way or another, and might not be relevant, though if ever get a screen where one can edit workspaces/projects, they might become relevant quickly.

```
bool OnlyAdminsMayCreateProjects { get; }

bool OnlyAdminsSeeBillableRates { get; }

bool OnlyAdminsSeeTeamDashboard { get; }

bool ProjectsBillableByDefault { get; }
```

These fields will definitely be relevant once we allow project creation, and add reports.

pe0ny:
@arturs105 I checked it out but i'm afraid you overestimated my tech skills a bit 😛 
but i'm up for questions about business logic if you have any.

arturs105:
Are project templates a planned feature?

Will `ITimeEntry.CreatedWith` be ever used?

Should `IProject.EstimatedHours` be an int? (`ITask` has `long EstimatedHours`)

Will `IUser.StoreStartAndStopTime` be necessary?

pe0ny:
>Are project templates a planned feature?

Yes, user should be able to use templates when creating new project. Check #798 
Set project as a template is not needed on mobile.

For other points, i think @amulware is better for answering.

amulware:
`TimeEntry.CreatedWith` NEEDS to be pushed to the API when creating time entries, but that is the only use case. Maybe we can handle this in ultrawave only? I'm right now not sure if it should have the user agent of the client when posting, or the user agent of the client when creating the local data (could be different because app updates).

I'll ask backend for the other things.

simonrozsival:
It would be great to have `CreatedWith` only in Ultrawave. The correct value is in fact set in the `TimeEntriesApi.pushTimeEntry`. The API server does not return this value in the API response and therefore we can't even display it anywhere even if we wanted to.

amulware:
This is ready for review now! The big :fire: commit only implements the interface changes from earlier. Tests run fine and the app works fine too (after re-install of course).

PR #765, opened 2017-10-20
Update the UI after deleting a time entry

### LINKED ISSUE #689: Support time entry deletion in UI
- [x] Make delete button set deleted flag
- [ ] Filter deleted entries from all queries

simonrozsival:
Closes #689

Squash whenever.


amulware:
Do I understand correctly that it removes and re-inserts (if not empty) the whole day in the log?

Can't you notify it that the element at a specific index changed in the re-insertion case, instead of having two do it in two steps which might result in glitchy UI?

simonrozsival:
Well this is the way it works when adding a new time entry, so I used the same approach here. Maybe @willsb could refactor this when he gets back from his vacation or maybe he will tell us why this is the correct way of doing this.

amulware:
Oh.. no way of testing this?

simonrozsival:
I will add some tests.

simonrozsival:
I added tests, is it OK like this?

PR #767, opened 2017-10-23
Add more information to API exceptions

### LINKED ISSUE #625: Log whole API response on failed api test
As the title says. Not sure how to implement this best, but it would help us a lot with analysing errors, especially since the response contains a unique id that backend could grep their logs with, as well as the full error information of course.

simonrozsival:
Closes #625 

Brace yourself for a ton of changed files! Most of these changes are quite simple and there is a lot of code which is repeated (mostly changes in the signature of constructors in the exception classes and in unit tests).

There is no `FsCheck` NuGet package in the Ultrawave tests yet and I thought it is not necessary to add it for these tests, though it is possible if the reviewer thinks it would improve the quality of the unit tests of the serialization.

Squash whenever.


simonrozsival:
Example of a log of a(n artificially) failed HTTP request in the output of the integration tests:
```
 Toggl.Ultrawave.Tests.Integration.StatusApiTests+TheGetMethod.ShouldSucceedWithoutCredentials [FAIL]
      FluentAssertions.Execution.AssertionFailedException : Expected object to be <null>, but found Toggl.Ultrawave.Exceptions.InternalServerErrorException with message "ApiException for request GET https://toggl.space/api/v9/status: Response: (Status: [500 InternalServerError]) (Headers: ['Server': ['nginx'], 'Date': ['Mon, 23 Oct 2017 12:09:25 GMT'], 'Cache-Control': ['no-cache, private, no-store, must-revalidate, max-stale=0, post-check=0, pre-check=0'], 'X-Toggl-Request-Id': ['948cfc697583017f090b1db817dba0f8'], 'X-We-Are-Hiring': ['http://jobs.toggl.com/'], 'Instance': ['staging-web'], 'Strict-Transport-Security': ['max-age=15552000; includeSubDomains'], 'X-Content-Type-Options': ['nosniff'], 'X-Request-ID': ['948cfc697583017f090b1db817dba0f8'], 'Via': ['1.1 google'], 'Alt-Svc': ['clear'], 'Transfer-Encoding': ['chunked']]) (Body: {"status":"OK"}) (Message: Internal server error.)"
      .
      Stack Trace:
          at FluentAssertions.Execution.FallbackTestFramework.Throw (System.String message) [0x00006] in <bad8a3e2d9724123b0c81c480a3beb4a>:0 
```

Toggl.Ultrawave/Exceptions/ApiException.cs
```diff
@@ -1,12 +1,36 @@
...
+
+        public override string ToString()
+            => $"ApiException for request {request.HttpMethod} {request.Endpoint}: "
+                + $"Response: "
+                + $"(Status: [{(int)response.StatusCode} {response.StatusCode}]) "
+                + $"(Headers: [{SerializeHeaders(response.Headers)}]) "
+                + $"(Body: {response.RawData}) "
+                + $"(Message: {message})";
+
+        public static string SerializeHeaders(IEnumerable<KeyValuePair<string, IEnumerable<string>>> headers)
```
amulware: Why is this public?

Toggl.Ultrawave/Network/Response.cs
```diff
@@ -1,24 +1,27 @@
...
+{
+    internal sealed class Response : IResponse
+    {
+        public string RawData { get; }
+        public bool IsSuccess { get; }
+        public string ContentType { get; }
+        public HttpStatusCode StatusCode { get; }
+        public IEnumerable<KeyValuePair<string, IEnumerable<string>>> Headers { get; }
+
+        public Response(string rawData, bool isSuccess, string contentType, IEnumerable<KeyValuePair<string, IEnumerable<string>>> headers, HttpStatusCode statusCode)
```
amulware: Should the headers be checked for nullability?

simonrozsival:
You are right that the message is what is now shown to the user. But FluentAssertions do not use `ToString`, but explicitly use the `.Message` property. Maybe we should create a `HumanReadableMessage` in the exception class which would be used in the unsyncable state.

amulware:
I think added a new property for this explicit purpose, instead of hijacking the existing one makes sense, yes. I vote for `LocalizedApiErrorMessage`, but pretty much anything is fine.

simonrozsival:
OK. I will get back to this in the afternoon and I will also reply to the remaining comments then.

PR #771, opened 2017-10-23
Fix crashing when editing running time entry right after starting it

### LINKED ISSUE #759: App crashes when trying to open running TE
1) Continue already existing TE in the log list; Timer starts running
2) Tap running TE to open it; App crashes

When opening the app again, the same TE is running and now I can open it without the crash.

simonrozsival:
Closes #759 

Squash whenever.


amulware:
The change seems counterintuitive to me. Can you explain how this interacts with the id change when syncing?

simonrozsival:
If the ID changes, then the `currentlyRunningTimeEntryId` is already changed in the `maybeUpdateCurrently...` function and so this comparison would be false and the the `currentlyRunningTimrleEntry` will not change and the next time user will open the edit view, it will query rhe database with the old negative ID and it will crash.

PR #772, opened 2017-10-23
Show tasks in log and suggestions view

### LINKED ISSUE #743: Suggestions/log view don't display task name
When selecting project with a task, it displays task name correctly in edit and time entry creation view but when stopping the time entry, it won't display task name in suggestion/log list.

arturs105:
Closes #743


Toggl.Foundation/Suggestions/Suggestion.cs
```diff
@@ -16,12 +16,15 @@ public sealed class Suggestion
...
 
+        public string ClientName { get; } = "";
+
         public bool HasProject { get; } = false;
 
         public long WorkspaceId { get; }
 
         internal Suggestion(IDatabaseTimeEntry timeEntry)
         {
+            TaskName = "Some task";
```
simonrozsival: I don't understand why you put this dummy text here - this is not a test case to have mock values in there. Shouldn't this line be 🔥 ?

simonrozsival: In fact I can see it in the log: http://prntscr.com/h18d77 - you probably had that in that place for testing, but forgot to remove afterwards?

PR #775, opened 2017-10-24
Add fade view

arturs105:
Fade-outs are used in multiple places, so i created a view that fades out left and right sides of it's children.

I didn't change existing fades, because the fade widths aren't known yet. I will do it in another PR.


amulware:
Will this work for views that change their size?

arturs105:
I think it should, since the fade is added in `LayoutSubviews`.

amulware:
I just noticed that it creates a new gradient layer every time, which seems wasteful. First I thought it might be adding more and more of them, but it replaces instead, so it's probably fine for resizing views. Though if the layer could be reused, that would be great.

arturs105:
I modified the `FadeView` so it updates the mask only when the size of it changes. Size changes of child views don't require mask updates.

Toggl.Daneel/Views/FadeView.cs
```diff
@@ -0,0 +1,66 @@
...
+            base.LayoutSubviews();
+
+            if (fadeLayer?.Bounds == Bounds) return;
+
+            addFading();
+        }
+
+        private void addFading()
+        {
+            fadeLayer = new CAGradientLayer();
```
amulware: Why not reuse the same layer, instead of creating multiple ones?

Issue #692, opened 2017-10-09
Cancel Sync on Logout


amulware:
We have to figure out how to stop the sync machine in such a way that it is guaranteed to make no changes to the database anymore.

This is needed so that on logout/in no data from the previous user is visible to the new one.

I propose we find some way of telling the sync machine to stop once the next state finishes, and prevent it from entering any other state from that point on - i.e. freezing it. I do not think there is any need to ever unfreeze it, so this should be relatively easy?

Note that the logout has to wait until the sync machine is confirmed to have stopped all states before proceeding to clear the database and navigate to the login screen.

simonrozsival:
This should be easy. I don't think we ever discussed any case where we would want to stop the machine manually and resume later on.

PR #780, opened 2017-10-24
Cancel sync

simonrozsival:
Ref #692 

This PR does not close the issue - it only introduces the mechanism which should be then used after the user clicks the button, that will be handled in a different PR to make this a little bit more compact.

The idea of the cancellation is to set internal state of the `SyncManager` and `StateMachine` in a such way, that _it won't do any more transitions_. The currently executed state is not interrupted, instead after the state finishes with a result, the `StateMachine` announces a dead end and the `SyncManager` emits a `Sleep` state. This means that the sync process finished and the user can be logged out (given that the local state is completely wiped out, we don't have to deal with any inconsistencies caused by the interrupted sync process). The `SyncManager` does not allow starting any new sync process after it is frozen - a new instance must be created. Which is OK for us since it will be created as part of the new `DataSource` after the next login.

Feel free to suggest any more test cases for all of the three classes `SyncManager`, `StateMachine`, `StateMachineOrchestrator`.

This PR does not deal with cancelling `Observable.Delay`, that would be part of a separate follow-up PR.

Squash whenever.


Toggl.Foundation.Tests/Sync/SyncManagerTests.cs
```diff
@@ -338,5 +338,119 @@ public void TellsQueueToStartSyncAfterQueingPull()
...
+            }
+
+            [Fact]
+            public void RunningPushSyncOnFrozenSyncManagerGoesDirectlyToSleepState()
+            {
+                SyncManager.Freeze();
+                
+                SyncManager.PushSync();
+
+                Orchestrator.Received().Start(Arg.Is(Sleep));
```
amulware: Would this not succeed even if there were non-sleep Start calls?

Toggl.Foundation.Tests/Sync/SyncManagerTests.cs
```diff
@@ -338,5 +338,119 @@ public void TellsQueueToStartSyncAfterQueingPull()
...
+            }
+
+            [Fact]
+            public void RunningFullSyncOnFrozenSyncManagerGoesDirectlyToSleepState()
+            {
+                SyncManager.Freeze();
+
+                SyncManager.ForceFullSync();
+
+                Orchestrator.Received().Start(Arg.Is(Sleep));
```
amulware: ditto

PR #781, opened 2017-10-24
Display project, task and client for running time entries

### LINKED ISSUE #756: Add project, task and client to currently running time entry
Currently running time entry card displays only description. Project task and client could be added as a separate label so `ProjectTaskClientValueCombiner` could be reused for that.

arturs105:
Closes #756 

Do a functional test and try to break stuff.


Toggl.Foundation.MvvmCross/ViewModels/MainViewModel.cs
```diff
@@ -92,6 +106,32 @@ public override void ViewAppeared()
             navigationService.Navigate<TimeEntriesLogViewModel>();
         }
 
+        private void setRunningEntry(IDatabaseTimeEntry timeEntry)
+        {
+            if (timeEntry == null)
```
kornelijepetak: Any reason why not collapse the `if-statement`?

```c#
CurrentTimeEntryId = timeEntry?.Id;
CurrentTimeEntryProjectColor = timeEntry?.Project?.Color ?? "";
// ... the same for all 7 of them
```

PR #784, opened 2017-10-24
Cancel delays when canceling sync

simonrozsival:
Ref #692

On-hold: this must be reviewed and closed after #780 is merged.

Squash whenever.


Toggl.Foundation.Tests/Sync/States/Push/CheckServerStatusStateTests.cs
```diff
@@ -172,6 +175,20 @@ public void DelaysTheTransitionAtLeastByTheNextFastDelayTimeFromTheRetryDelaySer
             hasCompleted.Should().BeFalse();
         }
 
+        [Fact]
+        public void CompletesEvenThoughTheDelayIsNotOverButWhenDelayCancelationCompleted()
```
amulware: But the delay cancellation doesn't actually complete, it just calls OnNext?

Toggl.Foundation/Sync/States/Push/CheckServerStatusState.cs
```diff
@@ -37,6 +39,12 @@ private IObservable<ITransition> proceed
         private IObservable<ITransition> delayedRetry(TimeSpan period)
             => Observable.Return(Retry.Transition()).Delay(period, scheduler);
 
+        private IObservable<Unit> delay(IObservable<Unit> observable)
+            => observable
+                .Delay(apiDelay.NextSlowDelay(), scheduler)
+                .Merge(delayCancelation)
+                .FirstAsync();
```
amulware: I think I like the idea of using Merge and First for this. I don't know if there is maybe a more 'accepted' pattern to cancelling observables, but this seems reasonably elegant.

Toggl.Foundation/Sync/SyncManager.cs
```diff
@@ -19,14 +22,16 @@ public sealed class SyncManager : ISyncManager
         public SyncState State => orchestrator.State;
         public IObservable<SyncState> StateObservable => orchestrator.StateObservable;
 
-        public SyncManager(ISyncStateQueue queue, IStateMachineOrchestrator orchestrator)
```
amulware: What is your reasoning for including the cancellation observable here, instead of in the orchestrator or machine?

Toggl.Foundation.Tests/Sync/States/Push/CheckServerStatusStateTests.cs
simonrozsival: It doesn't "complete" in the way that the `OnComplete()` would be called. It emits a `Unit` value. I can rename this to be less confusing.

amulware: Please do. :)

Toggl.Foundation/Sync/SyncManager.cs
simonrozsival: It seemed obvious to put it into the `SyncManager`. It would also make sense to put it in the `StateMachine`, but I don't think there is a fundamental difference between putting it in these two classes.

amulware: Fair enough, I can't really make up my mind either about where it fits best and why.

amulware: Except maybe that the state machine is the one dealing with 'low level' observables (states) already, while the sync manager knows nothing about that sort of stuff. If you find that a convincing argument, change it, otherwise leave it.

PR #788, opened 2017-10-25
Use fade view in existing views

arturs105:


Toggl.Daneel/ViewControllers/MainViewController.cs
```diff
@@ -158,7 +158,8 @@ private void prepareViews()
             reportsButton.SetImage(UIImage.FromBundle("icReports"), UIControlState.Normal);
             settingsButton.SetImage(UIImage.FromBundle("icSettings"), UIControlState.Normal);
 
-            addFading();
+            RunningEntryDescriptionFadeView.FadeLeft = true;
+            RunningEntryDescriptionFadeView.FadeRight = true;
```
amulware: No way to set these in the designer?

amulware:
Otherwise, it looks good in the app.

Toggl.Daneel/ViewControllers/MainViewController.cs
arturs105: In XCode Interface Builder - no
In VS iOS Designer - yes (at least documentation said that).

I think most of us use IB (since iOS designer is painfully slow) so setting those properties in code makes more sense than to use 2 UI building tools.

amulware: Ok.

Issue #790, opened 2017-10-25
Fix resizing the container view on content size change in Edit View


kornelijepetak:
Currently, when the Edit View content grows or shrinks, its container view is resized accordingly, but only on second layout update.

Steps to reproduce:

* Open TE Edit View
* Write some text into Description field
* Write enough (but one by one) characters and look what happens when it goes to the next line
* The moment it jumps to the next line, the modal dialog should resize, but *it doesn't*
* Any other layout-inducing change at this point (e.g. adding more characters) will correctly resize the container

simonrozsival:
This issue was resolved and is not present in 1.5 anymore (and it was probably already fixed in some earlier version).

Issue #783, opened 2017-10-24
Create a custom UITextView with placeholder


kornelijepetak:
The title says it all.

Possible inspiration: https://github.com/devxoul/UITextView-Placeholder

Needed wherever we have a multiline TextView and a need for a placeholder.
Currently:

* TE Edit View > `Description` field

amulware:
Please add where we need this, thanks!

kornelijepetak:
I have. If anyone else knows some other places, feel free to add a comment or edit the OP.

amulware:
@kornelijepetak What is the status on this?

kornelijepetak:
I will see that this gets merged today. I just need to integrate with #944 which shouldn't be that much of a problem. Or so it seems. :)

PR #791, opened 2017-10-25
Improve EditView scrolling

### LINKED ISSUE #637: Disable initial scrolling in the edit view
Currently edit view can be scrolled upon opening, however scrolling should become active only when the edit view has grown to the maximum size (due to too many tags or something else) and still needs more space.

kornelijepetak:
Closes #637.

Scrolling now works only if it is needed and the screen resizes itself according to the content.

This introduces #783 abd #790.


Toggl.Daneel/ViewControllers/EditTimeEntryViewController.cs
```diff
@@ -172,6 +174,17 @@ public override void ViewDidLoad()
...
         }
 
+        public override void ViewWillLayoutSubviews()
+        {
+            var newSize = new CGSize(0, nonScrollableContentHeight + ScrollViewContent.Bounds.Height);
+            if (newSize != PreferredContentSize)
+            {
+                PreferredContentSize = newSize;
+                PresentationController.ContainerViewWillLayoutSubviews();
+                ScrollView.Bounces = ScrollViewContent.Bounds.Height > ScrollView.Bounds.Height;
```
arturs105: I think you can set `.Bounces` only once and it will bounce only, when scrolling is enabled.

arturs105:
Also the dismissal of the view controller is messed up in the following case:
1. Edit description by entering several newlines so that size increases
2. Delete one of the newlines
3. Dismissal is broken

Here's a gif:
oct-26-2017 10-09-40

kornelijepetak:
wee...
That's not good.
Does this happen on phone too or in sim only?

Cause if it's consistent, I don't think we can use this method at all.
@arturs105 Would you have any idea as to why that happens?

amulware:
Any estimate on how long this will take to fix?

kornelijepetak:
I have no idea. I have just opened it again to see what's wrong. I'd give it a go for an hour or two at most, but wouldn't like this to take another 2 days.

amulware:
Agreed, otherwise let's merge it with that bug remaining, especially since on actual devices you can't enter new lines. If the issue also happens for regular linebreaks and can be reproduced on a device, create an issue for it.

kornelijepetak:
But the problem is that this happens whenever a resize happens and not only on entering new lines. As long as you do something that resizes the window, the bug will happen on dismissal. I'll see if I can take care of that during the day.

kornelijepetak:
Actually, even if you don't change anything, and just try to pull the view up, it will do a locking motion (go a little bit upwards and then snap back to original position.)

For this behavior I am not sure if that was happening before. Gotta check that.

amulware:
I'm having a hard time reproducing the error above. As far as I can tell it only happens with actual new-line characters in the description. That is the only way I could cause the bug.

So, let's ignore that glitch and get this approved so we can merge this for now so that other work can go on and we can deal with the glitch later.

kornelijepetak:
I can repro this even if I just write enough words so that it switches to multiline (I don't have to enter any newline characters). At least in simulator. I don't know how it behaves on a real device.

I'll wait for @arturs105 to approve so we can move on.

arturs105:
You still haven't made the other requested changes.

kornelijepetak:
Done with requested changes.

Issue #786, opened 2017-10-25
UI: Cancel Sync on Logout


pe0ny:
1) Indicate syncing in Settings view even before user taps 'Sign Out'
![toggl - ios - settings loading v1 1x](https://user-images.githubusercontent.com/5064445/31984260-8e597f46-b969-11e7-85c5-c636bf4355f3.png)

2) When user taps 'Sign Out' while we're syncing, show confirmation dialog:
![toggl - ios - settings sign out confirmation 1x](https://user-images.githubusercontent.com/5064445/31984312-c64579a0-b969-11e7-9d9a-3a005113563a.png)

a) When user still decides to continue and taps 'Sign Out' again, disable 'back' button and show 'Logging you out securely…':
![toggl - ios - settings loading v3 1x](https://user-images.githubusercontent.com/5064445/31984466-5587d1b2-b96a-11e7-8c97-fdb25e1b88b7.png)

b) When user taps 'Cancel', close the dialog and go back to Settings page where we display 'Syncing' and 'Synced' when syncing is complete:
![toggl - ios - settings loading v1 1x](https://user-images.githubusercontent.com/5064445/31984729-3f24d216-b96b-11e7-8c4e-f4c4548d88d8.png) ![toggl - ios - settings synced 1x](https://user-images.githubusercontent.com/5064445/31984733-41702c5a-b96b-11e7-9802-dcf255ebf41c.png)

simonrozsival:
I have a question about the "Synced [tick]" status message - when and how exactly should it be shown? I think this message should only be displayed for a short while after the sync process finishes (like the pull-to-refresh status message is hidden after few seconds), because it can become outdated if the user edits something and the sync doesn't start for some reason (e.g., user is offline).

amulware:
But the user cannot edit data on this screen. I think it should always show it, just as a nice confirmation to the user that they can log out safely.

simonrozsival:
That is true. As long as the state of the screen is not preserved when the user goes back and forth, which I believe is not, then it doesn't have to be hidden.

simonrozsival:
Since we changed the behavior of the dialog in #835 with @amulware - the dialog shows not only if the sync is running at the moment, but also when there is some unsynced time entry in the local database - the title and the message of the dialog should be changed to reflect this. It doesn't make to show "We're syncing your data" to a user, who is offline.

PR #794, opened 2017-10-25
Cancel sync after the logout button is clicked

### LINKED ISSUE #692: Cancel Sync on Logout

simonrozsival:
Closes #692, Ref #786 

This PR awaits cancellation of the sync process before the user is logged out.

Clicking on the logout button disables that button and hides the back button. The sync status message and the yes/no dialogue will be part of a different PR.

On-hold: The PR #784 must be merged first.

Squash whenever


Toggl.Foundation.MvvmCross/ViewModels/SettingsViewModel.cs
```diff
@@ -28,6 +34,12 @@ public SettingsViewModel(ITogglDataSource dataSource, IMvxNavigationService navi
 
             this.dataSource = dataSource;
             this.navigationService = navigationService;
+            this.IsLoggingOut = false;
+            RaisePropertyChanged(nameof(IsLoggingOut));
```
amulware: Why is this event raise necessary?

Toggl.Foundation.MvvmCross/ViewModels/SettingsViewModel.cs
```diff
@@ -37,6 +49,10 @@ public SettingsViewModel(ITogglDataSource dataSource, IMvxNavigationService navi
 
         private async Task logout()
         {
+            IsLoggingOut = true;
+            RaisePropertyChanged(nameof(IsLoggingOut));
```
amulware: ditto

Toggl.Foundation.MvvmCross/ViewModels/SettingsViewModel.cs
```diff
@@ -15,6 +17,10 @@ public sealed class SettingsViewModel : MvxViewModel
 
         public string Title { get; private set; }
 
+        public bool IsLoggingOut { get; private set; }
+
+        public IObservable<bool> IsRunningSync { get; private set; }
```
amulware: Is this used anywhere right now?

Toggl.Foundation.Tests/MvvmCross/ViewModels/SettingsViewModelTests.cs
```diff
@@ -33,14 +39,60 @@ public void ThrowsIfAnyOfTheArgumentsIsNull(bool useDataSource, bool useNavigati
...
+                }
+
+                log.Should().HaveCount(expectedBooleans.Count).And.BeEquivalentTo(expectedBooleans);
+            }
+        }
+
         public sealed class TheLogoutCommand : SettingsViewModelTest
         {
             [Fact]
+            public async Task SetsTheIsLoggingOutFlagToTrue()
```
amulware: Should we also test that the flag, freeze and logout are called in the correct order and awaited before proceeding?

amulware:
Requested review by @arturs105 since he understands bindings much better than me.

Toggl.Foundation.MvvmCross/ViewModels/SettingsViewModel.cs
simonrozsival: It is not right now. Though it is tested and it should be later possible to implement the UI part without touching the VM logic.

simonrozsival:
I will add such test.

Toggl.Foundation.Tests/MvvmCross/ViewModels/SettingsViewModelTests.cs
simonrozsival: I will add such test.

simonrozsival: Test was added.

Toggl.Foundation.Tests/MvvmCross/ViewModels/SettingsViewModelTests.cs
```diff
@@ -50,6 +104,58 @@ public sealed class TheLogoutCommand : SettingsViewModelTest
...
+                int callingNavigate = -1;
+                int awaitingNavigation = -1;
+                var syncManager = Substitute.For<ISyncManager>();
+                syncManager.Freeze().Returns(Observable.Create<SyncState>(async observer =>
+                {
+                    if (ViewModel.IsLoggingOut == true)
+                        isLoggingOutFlag = operationCounter++;
+
+                    callingSyncManagerFreeze = operationCounter++;
+                    await Task.Delay(100);
```
amulware: I'm not a big fan of forcing the test to run slowly with these delays, but I also don't want to spend too much time on this.

Toggl.Daneel/Binding/NavigationItemHidesBackButtonTargetBinding.cs
```diff
@@ -0,0 +1,29 @@
...
+            : base(target)
+        {
+        }
+
+        protected override void SetValue(bool value)
+        {
+            Target.SetHidesBackButton(value, true);
+        }
+
+        protected override void Dispose(bool isDisposing)
```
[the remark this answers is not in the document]
simonrozsival: Probably not, combination of copy&paste and my laziness.. I will 🔥 it.

Toggl.Foundation.Tests/MvvmCross/ViewModels/SettingsViewModelTests.cs
simonrozsival: Maybe this could be reduced to a few ms. I just wanted to somehow simulate that the task runs for a while and that if there weren't `await`s, then they would overlap and the order wouldn't be the same. I will change this to 2ms, that should do the job and not slow down the tests in any (noticable) way.

Toggl.Foundation.MvvmCross/ViewModels/SettingsViewModel.cs
simonrozsival: OK, I can remove the observable.

PR #795, opened 2017-10-25
Hide timeline temporarily

### LINKED ISSUE #779: Remove timeline's grey bar
Remove timeline's grey bar at the top of the suggestions view until we have timeline ready.

kornelijepetak:
Closes #779 .

Resized to 0 height and hidden.

Should be easy to revert when timeline is implemented.

If there's a better way, please tell.


simonrozsival:
The spider should be definitely moved a bit up. It looks quite strange that the web isn't attached to anything now though, but I'm not sure what to do about that (@pe0ny?).

spider_hanging

pe0ny:
I agree. Let's attach it to the toggl's 'o'

PS. it feels like Nav bar's background is full white #ffffff though all the background should be #FAFBFC

toggl - ios - timer empty state temp 1x

simonrozsival:
@kornelijepetak Maybe you could also 🔥 the whole view. It can be added back once we implement the timeline.

kornelijepetak:
@pe0ny I'm not sure this is possible with the current setup. The Toggl logo is in a non-scrollable view. I am not an expert in iOS UI, but I think that if I want the spider to go away by swiping up, I must not remove it from the scrollable content view. However, if it remains in that view, it can't go outside of it (i.e. position it as far up as the toggl logo).

I would like @arturs105 to confirm this or advise on how to structure the view hierarchy so that the spider would be attached to the logo and still go away on swipe.

pe0ny:
Ok, i see your point. Let's not overkill it and use this one instead:

toggl - ios - timer empty state temp 1x

pe0ny:
@kornelijepetak Simon is actually right. The handler is centered but spider isn't:

screen shot 2017-10-31 at 15 54 16

kornelijepetak:
I think it's ok now. It can be merged after approval.

image

PR #807, opened 2017-10-26
Remove Stop and use Duration everywhere

### LINKED ISSUE #747: Represent time entries with start+duration, instead of start+stop
This better fits with how backend and the API represent data. If the API will be simplified in this area, stop time is also the field most likely to be removed, instead of duration.

Unfortunately, we still have to deal with negative-timestamp-in-duration-field encoding for running time entries for the time being.

What makes it worse is that in principle the API allows for the start time of a time entry to be basically arbitrary, even for running time entries. For example: The start time could be two months ago, and the negative duration encodes a running-time-duration of 2 hours.

Pending approval from @pe0ny, I would like to ignore such 'bad' data, even though it could occur in rare cases. I simply cannot find a way that we can have that display and edit in a sensible way in the app.

As for implementing this change:

Internally to the app I would like to have the duration field be `null` for running entries, and the proper duration in seconds for stopped ones.

When communicating with the API we will do the negative-duration conversion there, in both directions.

We will also serialise the stop time as `null` or `start+duration` when pushing time entries, but when pulling we can entirely ignore it.

simonrozsival:
Closes #747, Ref #709

Please look especially at the Ultrawave model - I use two fields for the duration property: `Duration` and `ApiDuration`. One of them is ignored for serialization. I decided to do it this way because it makes it easy to make sure that the internal `Duration` is `null` when it is running and when serialized the negative start timestamp is used and similarly for deserialization.

~~WIP: I still have to go through the code once more and make sure all of the `.Stop` usages are gone + do a functional test in the simulator.~~

Functional test shows that it works well after the refactoring (pushing and pulling data, duration edit view, ...).

There should be a follow-up issue to refactor the edit duration view, which still uses the stop time and then converts it to duration which is then passed to the entity. This PR is already quite big, I didn't want to include that in this PR which would make it a lot bigger.

Squash whenever.


amulware:
@simonrozsival since you know best what still has to be changed to remove all the `stop`s, can you create the followup issue?

amulware:
@arturs105 @kornelijepetak +1 review as sanity check please?

simonrozsival:
I already did - #810. Should I add more details to the issue or is it OK like this?

amulware:
Didn't see that, looks good!

Toggl.Ultrawave/Models/TimeEntry.cs
```diff
@@ -20,14 +22,22 @@ internal sealed partial class TimeEntry : ITimeEntry
...
         public DateTimeOffset Start { get; set; }
 
-        public DateTimeOffset? Stop { get; set; }
-
-        public long Duration
+        public DateTimeOffset? Stop
         {
-            get => Stop.HasValue ? (int)(Stop.Value - Start).TotalSeconds : -Start.ToUnixTimeSeconds();
+            get => Duration.HasValue ? Start.AddSeconds(Duration.Value) : null as DateTimeOffset?;
             set { }
```
kornelijepetak: Why do we allow an empty setter? I'm confused at the moment. It seems that there are lots of call sites that set the property to all sorts of values. But if it's just swallowed, what's it use? Or did I miss something?

simonrozsival: This is because of deserialization - it needs a setter for that value to work, but we don't need the value, so we simply ignore it.

Toggl.Foundation.Tests/Sync/ConflictResolution/TimeEntryRivalsResolverTests.cs
```diff
@@ -106,25 +109,25 @@ public void TheStoppedTimeEntryMustBeMarkedAsSyncNeededAndTheStatusOfTheOtherOne
...
             timeService.CurrentDateTime.Returns(now);
-            var a = TimeEntry.Clean(new Ultrawave.Models.TimeEntry { Id = 1, Stop = null, At = arbitraryTime.AddDays(21), Start = arbitraryTime.AddDays(20) });
-            var b = TimeEntry.Clean(new Ultrawave.Models.TimeEntry { Id = 2, Stop = null, At = arbitraryTime.AddDays(22), Start = arbitraryTime.AddDays(21) });
+            var a = TimeEntry.Clean(new Ultrawave.Models.TimeEntry { Id = 1, Duration = null, At = arbitraryTime.AddDays(21), Start = arbitraryTime.AddDays(20) });
+            var b = TimeEntry.Clean(new Ultrawave.Models.TimeEntry { Id = 2, Duration = null, At = arbitraryTime.AddDays(22), Start = arbitraryTime.AddDays(21) });
 
             var (fixedA, _) = resolver.FixRivals(a, b, timeEntries);
 
-            fixedA.Stop.Should().Be(now);
+            fixedA.Duration.Should().Be(5 * 24 * 3600);
```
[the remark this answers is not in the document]
simonrozsival: Yes, that is a good idea, I will change that.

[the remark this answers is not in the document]
amulware: :+1:

Toggl.Ultrawave/Models/TimeEntry.cs
kornelijepetak: That's what I thought. However, that does not explain why there are other callsites (except the serialization) where this property gets assigned a specific value.

simonrozsival: There should be none. This property is not in the `Toggl.Multivac.Models.ITimeEntry`, it is just in `Toggl.Ultrawave.Models.TimeEntry`. I found some leftovers in the tests, I will :fire: these.

kornelijepetak: True, those were only in tests.

simonrozsival: Thanks for pointing this out!

simonrozsival:
I addressed all comments, there is no review to dismiss, there were no drastic changes, so I take it that your reviews still hold, when the tests run successfully, I will squash this.

amulware:
:shipit:

PR #808, opened 2017-10-26
Improved task integration tests

### LINKED ISSUE #658: Fix integration tests timing out
They have been timing out on bitrise for over a week. No good. Fix pls.

amulware:
In essence, I added a helper class for subscriptions that remembers what it did in the past and avoids unnecessary API calls.

This makes the tests no longer freeze on bitrise, though currently, some of them are failing. But I believe that to be unrelated.

I also fixed some typos/copy paste errors.

Closes #658


amulware:
To be clear, I don't know why it was freezing before, except that it probably has something to do with the kind or amount of API requests, which at the moment is unique to these tests. That this fix works seems to confirm that hypothesis.

PR #815, opened 2017-10-26
Initiate push after deleting a time entry

### LINKED ISSUE #764: Deleting a time entry does not initiate push sync
The code contains the code for initiating push synchronization - in `EditTimeEntryViewModel.cs` on line 164. And there is also a unit test for the initiation of push sync. But it does not work for some reason.

simonrozsival:
Closes #764 

To be honest I am not really sure why it didn't work before. The action in the `Do` observable was never called when running in the simulator, but it worked when running unit tests. Moving the push to the `onDeleteCompleted` works as it should. Unit tests don't need any change.

Please do a functional test in simulator when reviewing.

Squash whenever


amulware:
That's a bit disturbing. Pinging @willsb in case he has insights, or would just like to know about this.

PR #817, opened 2017-10-26
Fix deleting running time entry

### LINKED ISSUE #764: Deleting a time entry does not initiate push sync

### LINKED ISSUE #816: Deleting running time entry leaves timer running
I get an EntityNotFoundException when tapping the running timer again.

The deleted event seems to not be passed to the timer.

simonrozsival:
Closes #816, in fact this would also close #764 and #815 wouldn't be needed, but I don't think we should "revert" that merge request.

Do a functional test in simulator when reviewing.

Squash whenever.


amulware:
I'm not sure how I feel about this. But that's more of an issue with how observables work I suppose. Still, it seems safer to depend on the completed, not the on-next for no-data/unit observables - except of course if it is an observable that might ping multiple times... I'm not being helpful.

PR #823, opened 2017-10-27
Include HTTP request body in the ApiException.ToString

simonrozsival:
This makes debugging integration tests much easier.

I am only thinking about security implications - if login fails, the user's password or the "almost correct password" will be included in the serialized log, if we then use some bug reporting service which would upload these values somewhere, we would expose our users' plaintext passwords. What should we do about this? A solution might be to check if it is a production or testing build and not include this in the production build.

Squash whenever.


amulware:
We indeed cannot expose ANY user data like that, especially to third parties.

simonrozsival:
We can probably close this without merging then.

amulware:
Let's do so for now and think about it again in a few weeks (we might never need it).

PR #827, opened 2017-10-27
Conditional serialization of paid features attributes

### LINKED ISSUE #796: Only POST/PUT paid feature fields if workspace has relevant feature enabled
Currently, this affects these two properties of time entries:

```csharp
public long? TaskId { get; set; }
public bool Billable { get; set; }
```

My idea of how this might work is:

- we inject the feature set of the relevant workspace when POST/PUTting (I don't see a way for the time entry itself to know this information about itself)
- we have an attribute on the above properties, something like `[RequiresFeature(Pro)]`
- update the serialiser to take into account the injected features and the attribute

simonrozsival:
Closes #796 

See the issue to see some reasoning behind these changes.

Squash whenever

### DIFF
--- a/Toggl.Daneel/Startup/Setup.cs
+++ b/Toggl.Daneel/Startup/Setup.cs
@@ -63,7 +63,7 @@ namespace Toggl.Daneel
             var version = NSBundle.MainBundle.InfoDictionary["CFBundleShortVersionString"];
             var userAgent = new UserAgent("Daneel", version.ToString());
 
-            var apiFactory = new ApiFactory(environment, userAgent);
+            var apiFactory = new ApiFactory(environment, userAgent, database.WorkspaceFeatures.GetById);
             var loginManager = new LoginManager(apiFactory, database, timeService, TaskPoolScheduler.Default);
 
             Mvx.RegisterSingleton<ITimeService>(timeService);
--- a/Toggl.Foundation/Login/ApiFactory.cs
+++ b/Toggl.Foundation/Login/ApiFactory.cs
@@ -1,4 +1,6 @@
-﻿using Toggl.Multivac;
+﻿using System;
+using Toggl.Multivac;
+using Toggl.Multivac.Models;
 using Toggl.Ultrawave;
 using Toggl.Ultrawave.Network;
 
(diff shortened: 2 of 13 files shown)


simonrozsival:
The issue is on-hold, so this is on-hold as well.

simonrozsival:
I originally thought about passing the features with each call, but that didn't seem very practical. You would have to manually retrieve the data from the database and then pass it as a parameter to the API client method, which is, in fact, the same that happens in the `TimeEntriesApi` class.

I'm not sure I see the race condition you mention. Can you elaborate on that maybe?

amulware:
I mean the race condition of a workspace features changing between getting the model for pushing, and getting the workspace features to serialise the model. But I'm not sure now that this ever matters. I can't really come up with a scenario where it breaks. In general, if we send it with the wrong features (maybe user just changed them in the web interface), and backend starts enforcing it, it will either cause an error or it will work, either should be fine without any harmful side effects.

I'd like @willsb's opinion on injecting this function into the API as well though, from an architectural standpoint. It sort of couples the API to the database in a tighter way than I'm sure I'm comfortable with.

simonrozsival:
The features can change only during pull sync and we don't allow multiple sync transitions at one time. So this could never happen.

amulware:
Until the day we allow users to sign up to a plan within the app, and then we have to maybe think about this and A LOT of things. But today is not that day.

I still want @willsb comments on this way of injecting the function into the API from a general architectural standpoint, but my points above are otherwise addressed.

willsb:
As long as it does not create a dependency on PrimeRadiant, I'm okay with injecting a named delegate or an interface (a `Func` on its own is rather hard to understand).

However, I'd still prefer us to pass parameters as needed. This is Foundation's job after all: To make PrimeRadiant/Ultrawave nicer to consume. All fetching would happen in the Foundation layer (more specifically, inside the sync algo, maybe even having a dedicated state for fetching this) and ofc not every method call would require the conditional serialization. Why do you think this would be bad? We already do a lot of fetching from the DB in order to create things in the ViewModels, and this would all happen behind the scenes in Sync code. Doesn't sound problematic to me at all.

amulware:
There neither seems to be a consensus, nor any impetus for implementing this in either case for the moment. Closing it until further notice.

PR #828, opened 2017-10-27
Fix serialization when negative duration is set explicitly

simonrozsival:
Run integration tests on develop and on this branch locally to see the difference.

Squash whenever.


amulware:
I don't like adding code that will never be relevant in production to make tests pass.

Let's instead replace the affected tests by new ones that that the creation of TEs with start times around the boundaries, (now +- ~999h)

PR #832, opened 2017-10-30
Re-add AdHoc building to cake

amulware:
This was accidentally removed in #822, while we are using it in bitwise for internal daily builds (oops :ghost: ).

I renamed it slightly and added the comment to group the iOS builds a bit.

:squid: at your pleasure.


amulware:
Also renamed the cake target on bitrise.

PR #835, opened 2017-10-30
Show confirmation dialog when user clicks on the logout when sync is running

simonrozsival:
Ref #786 

Do a functional test in simulator or on a device when reviewing. I recommend adding a `.Delay(TimeSpan.FromSeconds(10))` to `BasePersistState` on line 41 to make pull sync take much longer. ✅ of the view model included.

Squash whenever


Toggl.Foundation.MvvmCross/ViewModels/SettingsViewModel.cs
```diff
@@ -25,27 +27,49 @@ public sealed class SettingsViewModel : MvxViewModel
...
             BackCommand = new MvxAsyncCommand(back);
-            LogoutCommand = new MvxAsyncCommand(logout);
+            LogoutCommand = new MvxAsyncCommand(maybeLogout);
         }
 
         private Task back() => navigationService.Close(this);
 
+        private async Task maybeLogout()
+        {
+            if (!IsRunningSync)
```
amulware: While this condition is in accordance with the issue, it is not what I think we need.

Instead of checking whether we are syncing, should we not rather query the database to see if there is anything that needs syncing? (For example, if some data failed to sync, that would not be caught here right now.)

simonrozsival: You are right.

PR #837, opened 2017-10-30
Make return key dismiss keyboard in edit view

arturs105:
Entering a newline in description of time entry now dismisses the keyboard.


amulware:
How does this interact with copy/paste?

arturs105:
When pasting something with a newline, the whole pasted text gets ignored and keyboard is dismissed. Tested on simulator.

amulware:
The question is if that is what we want - or if we even care...  Any comments from @pe0ny?

pe0ny:
I don't see a reason why user should be add a new line in description at all. So if it misses the keyboard, i'm good with that.

I also tried to paste text with multiple lines and it pasted everything in one line which seems ok to me.

arturs105:
Tried this out in webapp - the newline character is converted to a space and the text is still added. We could do that as well.

amulware:
That sounds reasonable. I still like the 'enter to dismiss' though as well. (can't we do the same thing we did in the login, change the button text to 'done' maybe, hook it up nicely? - or is this done the same way in the login flow?)

pe0ny:
> the newline character is converted to a space and the text is still added. We could do that as well.

I think we're already doing this - at least that's how it worked for me

pe0ny:
And i like Paul's idea to change the button to 'Done'. it makes sense

arturs105:
So, to sum up, the changes are:

- Change the button text to "Done"
- Remove any newlines in pasted text and add that to description

But do we dismiss the keyboard, when user pastes stuff with newlines, or only when pressing "Done" (it will still try to insert a newline).

amulware:
* replace new lines by spaces (just in case)

I'd say keep the keyboard there, when pasting, the user may want to still edit the text or add more

PR #838, opened 2017-10-30
Handle tags for different workspaces in start view

### LINKED ISSUE #745: Handle tags for different workspaces in start view
When selecting a project in the start view that causes the time entry to change workspace (starting with the default WS of course), tags have to be cleared.

Similar to #716

Whether there should be a dialog for confirmation or something similar depends on @pe0ny.

Also consider that since the time entry without project will start out in the default workspace, and what to do when the user selects a tag from there, and then selects a project from a different workspace.

arturs105:
Closes #745


Toggl.Foundation.Tests/MvvmCross/ViewModels/StartTimeEntryViewModelTests.cs
```diff
@@ -40,23 +41,37 @@ protected StartTimeEntryViewModelTest()
                 DataSource.AutocompleteProvider.Returns(AutocompleteProvider);
             }
 
+            protected override void AdditionalSetup()
+            {
+                DialogService.Confirm(
+                       Arg.Any<string>(),
```
simonrozsival: The indentation should be decreased here by 4 spaces.

PR #839, opened 2017-10-30
Make sure time entries are grouped by local times

### LINKED ISSUE #809: Investigate/fix grouping by days
The grouping should group time entries by the date of the start time, in the current system time zone.

If this is not the case, it should be fixed. If it is the case, I'd like to know why the webapp shows different groupings at times.

screen shot 2017-10-26 at 14 11 37

arturs105:
Closes #809 

Time entries were grouped by whatever date was stored in the local database, which was utc.


arturs105:
I tried writing a test for checking, whether time entries are grouped by local time, but there is a problem: calling `.LocalDateTime` (which is done in the view model) uses local system's settings to convert UTC `DateTimeOffset` to a local date time. I was able to create a test that works for me, but there is no guarantee, that it would work in other time zones.

We could try to get the system's time zone and construct test datetimes based on that, but i'm not sure, it's worth the effort.

simonrozsival:
I see. I think we would find a way how to do it, but maybe it is an overkill. Let's keep it like this.

Issue #825, opened 2017-10-27
Signup UI


amulware:
Most of it is already there I think, but it needs to be finished and hooked up.

willsb:
Where can I get the links for the ToS and Privacy Policy?

pe0ny:
ToS: https://toggl.com/legal/terms
PP: https://toggl.com/legal/privacy

willsb:
This is on hold until #824 is closed

simonrozsival:
PR #873 is waiting for reviews.

PR #840, opened 2017-10-30
Add Browser View

willsb:
Ref #825

This add a browser view, making it possible for us to display web view content easily in the app. Will be needed for displaying both the ToS and PP pages in the Sign Up step.

Note that I have to do everything in code because of a bug in the way XCode parses XIB when there's a WKWebView in it. Since I didn't want to use the now deprecated UIWebView, this is the outcome. Not a huge problem, since there's only very little code.

Squash whenever


Toggl.Foundation.MvvmCross/ViewModels/BrowserViewModel.cs
```diff
@@ -0,0 +1,37 @@
...
+    [Preserve(AllMembers = true)]
+    public sealed class BrowserViewModel : MvxViewModel<BrowserParameters>
+    {
+        private readonly IMvxNavigationService navigationService;
+
+        public string Url { get; private set; }
+
+        public string Title { get; private set; }
+
+        public IMvxCommand CloseCommand { get; }
```
arturs105: `CloseCommand` isn't used anywhere, is it on purpose?

willsb: Some platforms might need it, but you're right, it should be added when needed

PR #841, opened 2017-10-30
Sync and logout status messages in the settings view

### LINKED ISSUE #786: UI: Cancel Sync on Logout

simonrozsival:
Closes #786 

~On-hold: waiting for #835 to be closed first.~

Question: It seems I am not able to match the size of the spinners in Zeplin and the MainViewModel - I set the size to 14x14 (image size, constraints), but it is still bigger than that (24x24) - _Scale to fit doesn't_ apply here for some reason. Also this is my first UI-related PR, I will be glad for any suggestions and comments regarding conventions and other best-practices used by our team which I'm not familiar to.

There are no animations for showing/hiding the status messages. If some animations will be requested, they should be part of a different PR.

Squash whenever.


simonrozsival:
WIP: Every time the user goes to the Settings view, it shows him "Synced [tick]" due to weird observables behavior I have to tackle yet - this might be misleading, it only means the sync is not running, not that everything is synced.

arturs105:
Just did a functional test and everything seems to be in order. However, i did notice a small delay before the message "Logging you out securely" appears. After i pressed "Log out" in the dialog, "Syncing" message was still shown for a little while and then switched to "Logging out securely". I wonder, if it is possible to make the "Logging you out securely" message appear immediately after the dialog closes.

Toggl.Foundation.MvvmCross/ViewModels/SettingsViewModel.cs
```diff
@@ -1,4 +1,6 @@
 ﻿using System;
+using System.Reactive.Concurrency;
```
arturs105: 🔥

Toggl.Foundation.MvvmCross/ViewModels/SettingsViewModel.cs
```diff
@@ -40,13 +46,21 @@ public SettingsViewModel(ITogglDataSource dataSource, IMvxNavigationService navi
             this.dialogService = dialogService;
 
             this.IsLoggingOut = false;
+            this.IsSynced = false;
+            this.IsRunningSync = false;
```
arturs105: `this.` prefix isn't needed for these 3.

Toggl.Foundation.MvvmCross/ViewModels/SettingsViewModel.cs
```diff
@@ -72,17 +86,19 @@ public SettingsViewModel(ITogglDataSource dataSource, IMvxNavigationService navi
...
 
         private async Task<bool> isSynced()
-        {
-            var everythingIsSynced = await dataSource.TimeEntries.GetAll(te => te.SyncStatus != PrimeRadiant.SyncStatus.InSync).SelectMany(te => te).IsEmpty();
-            return everythingIsSynced && !IsRunningSync;
-        }
+            => IsRunningSync ? false : await thereIsNothingToSync();
+
+        private async Task<bool> thereIsNothingToSync()
+            => await dataSource.TimeEntries.GetAll(te => te.SyncStatus != PrimeRadiant.SyncStatus.InSync).SelectMany(te => te).IsEmpty();
```
arturs105: Why not `Using PrimeRadiant`?

arturs105: And is the `SelectMany` call really necessary?

Toggl.Foundation.Tests/MvvmCross/ViewModels/SettingsViewModelTests.cs
```diff
@@ -139,6 +136,9 @@ public sealed class TheLogoutCommand : SettingsViewModelTest
                 DataSource.SyncManager.Returns(syncManager);
                 DataSource.Logout().Returns(Observable.Create<Unit>(async observer =>
                 {
+                    if (ViewModel.IsLoggingOut == true && ViewModel.IsSynced == false)
```
arturs105: The `== true` part can be dropped.

arturs105: Actually, the whole thing can be changed to `ViewModel.IsLoggingOut && !ViewModel.IsSynced`.

Toggl.Foundation.Tests/MvvmCross/ViewModels/SettingsViewModelTests.cs
```diff
@@ -116,7 +116,7 @@ public sealed class TheLogoutCommand : SettingsViewModelTest
             {
                 doNotShowConfirmationDialog();
                 int operationCounter = 0;
-                int isLoggingOutFlag = -1;
+                int isLoggingOutFlagIsSetAndIsSyncedFlagIsUnset = -1;
```
arturs105: Can't this be separated into two variables?

Toggl.Foundation.Tests/MvvmCross/ViewModels/SettingsViewModelTests.cs
simonrozsival: I know, but I like the long versions - at least the `== false` - it is much easier to read. From my experience, it is quite easy to overlook the `!` sometimes. But I can definitely 🔥 the `== true`.

Toggl.Foundation.Tests/MvvmCross/ViewModels/SettingsViewModelTests.cs
simonrozsival: The problem here is that I can't determine the order of setting the `IsLoggingOut` and `IsSynced` flags, so I thought it would be best to merge them into a single variable in this test.

Toggl.Foundation.MvvmCross/ViewModels/SettingsViewModel.cs
simonrozsival: `GetAll` returns `IObservable<List<ITimeEntry>>` and this always returns some list, I need to know if this list is empty or not. Maybe there is a better way, I'm not really happy with this either.

Toggl.Daneel/ViewControllers/SettingsViewController.cs
```diff
@@ -37,9 +43,47 @@ public override void ViewDidLoad()
...
 
+            bindingSet.Bind(SyncingView)
+                      .For(view => view.BindVisibility())
+                      .To(vm => vm.IsRunningSync)
+                      .WithConversion(visibilityConverter);
+
+            bindingSet.Bind(SyncedView)
+                      .For(view => view.BindVisibility())
+                      .To(vm => vm.IsSynced)
+                      .WithConversion(visibilityConverter);
```
amulware: Is it ok to bind two things to the same visibility thing? It seems like these would interfere.

amulware: Nevermind. I can't read.

PR #850, opened 2017-10-31
Remove DateTimeOffset.Now/UtcNow from unit tests

### LINKED ISSUE #805: Remove DateTimeOffset.Now and DateTimeOffset.UtcNow from tests unless really needed
There are quite a few tests which rely on the current time. Some of them even combine `UtcNow` and `Now`, which might cause problems for developers in different time zones.

Property tests or some fixed date times should be used in tests to prevent unexpected behavior - tests should be predictable.

simonrozsival:
Closes #805 

This has low priority. Removes some nondeterminism from the unit tests (does not have any effect on integration tests).

Squash whenever.


simonrozsival:
Yes, those dates are completely arbitrary. I tried to use different dates everywhere.

PR #851, opened 2017-10-31
Hide hinge while spider is hidden

### LINKED ISSUE #848: Show spider hinge only when spider is visible
This is not right.

image

kornelijepetak:
Closes #848.


kornelijepetak:
@simonrozsival I have added the missing border. It seems to me that it was never actually there. Instead, the timeline was serving as a visual separator. When we return the timeline, this border will probably not be needed anymore.

However, I was unable to fix the spacing between the border and the suggestions. Don't know if that's a problem at this point, but it's not as in Zeplin. Although, there is no suggestion screen without the suggestions there so I don't know how it should be.

amulware:
Is the review still valid?

kornelijepetak:
The review is still valid. The hinge is shown only when the spider is shown. However, I have accidentaly pushed another change into the same branch and that is the border below the navbar and above the suggestion. However, now that the timeline view is hidden temporarily, there is somewhat large paddding between that border and the suggestions.

I would like someone (perhaps @amulware) to check this functionally and decide whether to merge it like this or to revert the border changes (f9784b3) and put that into a separate PR (as it probably should have happend in the first place).

Anyways, feel free to squash this at your convenience.

amulware:
If @simonrozsival is fine with it, I'm fine with merging this now.

simonrozsival:
I'm fine with this.

PR #858, opened 2017-11-01
Test that backend stops previous running time entry when pushing a new running time entry

### LINKED ISSUE #856: Test backend stopping running TEs when pushing a second running one
Backend ensures that there will never be two running time entries by stopping any existing ones, anytime we push another running entry.

It would be great to test this.

simonrozsival:
Closes #856 

Run the integration test locally when reviewing.

Squash whenever.


Toggl.Ultrawave.Tests.Integration/TimeEntriesApiTests.cs
```diff
@@ -192,6 +192,26 @@ public sealed class TheCreateMethod : AuthenticatedPostEndpointBaseTests<ITimeEn
...
+
+                var postedFirstTimeEntry = await togglApi.TimeEntries.Create(firstTimeEntry);
+                var postedSecondTimeEntry = await togglApi.TimeEntries.Create(secondTimeEntry);
+                var stoppedFirstTimeEntry =
+                    await togglApi.TimeEntries.GetAll().SelectMany(te => te).Where(te => te.Id == postedFirstTimeEntry.Id).FirstAsync();
+
+                postedFirstTimeEntry.Duration.Should().BeNull();
+                postedSecondTimeEntry.Duration.Should().BeNull();
+                stoppedFirstTimeEntry.Duration.Should().NotBeNull();
+                postedFirstTimeEntry.Id.Should().Be(stoppedFirstTimeEntry.Id);
```
amulware: This is ensured by how the stopped entry is gotten in the query above. Wat would be nice to test is that the stop-date of this entry is the same as the At date, which should be the same as the At of the second posted one, and not the same as the first posted (2 second delay between the two postings to make sure?)

simonrozsival: You are right, I forgot to remove this assertion - I originally used a different query with `GetAllSince`.

Checking what the `At`s are makes a lot of sense, I will add that too.

PR #865, opened 2017-11-01
Remove ProjectSuggestion.FromProjectsPrependingEmpty

### LINKED ISSUE #818: Remove ProjectSuggestion.FromProjectsPrependingEmpty
Now that we care what workspace the "No project" project belongs to, this method makes little sense. Furthermore, empty projects are always added by the view models.

Removing this method also renders `ProjectSuggestion.NoProject` useless. This change might require updating/removing some tests.

arturs105:
Closes #818


arturs105:
I renamed it because the arguments of the method imply the `WithWorkspace` part and it was originally named that way, because at that time, `NoProject` property still was present and i couldn't use that.

PR #874, opened 2017-11-01
Page tracking

### LINKED ISSUE #868: Tracking view controllers/pages
So we can see where people go in the app, and what they look at.

willsb:
Closes #868


willsb:
I have figured a better way of dealing with this, when the times to deal with this comes

PR #877, opened 2017-11-01
Refactor the DialogService

willsb:
No reason for it not to be awaitable.

Squash whenever.


Toggl.Daneel/Services/DialogService.cs
```diff
@@ -1,36 +1,42 @@
 ﻿using System;
```
arturs105: 🔥

Toggl.Daneel/Services/DialogService.cs
```diff
@@ -1,36 +1,42 @@
...
+            var alert = UIAlertController.Create(title, message, UIAlertControllerStyle.Alert);
+            alert.AddAction(UIAlertAction.Create(dismissButtonText, UIAlertActionStyle.Cancel, _ => tcs.SetResult(null)));
             getPresentationController().PresentViewController(alert, true, null);
+
+            return tcs.Task;
         }
 
-        public void ShowMessage(string title, string message, string dismissButtonTitle, Action dismissAction)
+        public Task<bool> Confirm(
+            string title, string message,
```
arturs105: I'd like these to be on separate lines

Toggl.Foundation.MvvmCross/Services/IDialogService.cs
```diff
@@ -1,21 +1,20 @@
 ﻿using System;
```
arturs105: 🚫

Toggl.Daneel/Services/DialogService.cs
```diff
@@ -1,36 +1,42 @@
...
     {
-        public void Confirm(
-            string title,
-            string message,
-            string confirmButtonTitle,
-            string dismissButtonTitle,
-            Action confirmAction,
-            Action dismissAction,
-            bool makeConfirmActionBold)
+        public Task Alert(string title, string message, string dismissButtonText)
```
simonrozsival: I can't see any dialog like this in Zeplin and you didn't use it anywhere in this PR either. Do we need this at all? Can you think of an use case?

willsb: There are use cases, but YAGNI wins :P

PR #882, opened 2017-11-02
Disable all buttons while logging in

### LINKED ISSUE #880: Disable all buttons while logging in
All the buttons should be disabled during the login process.
Disabled buttons should be 50% transparent.

arturs105:
Closes #880 

Squash whenever.


Toggl.Daneel/ViewControllers/LoginViewController.cs
```diff
@@ -187,7 +210,7 @@ private void prepareTextFields()
 
         private void prepareNavigationBar()
         {
-            var attributes = new UITextAttributes { Font = UIFont.SystemFontOfSize(14, UIFontWeight.Medium) };
+            var attributes = new UITextAttributes { Font = UIFont.SystemFontOfSize(14, UIFontWeight.Medium)};
```
simonrozsival: I think this whitespace should stay there.

simonrozsival: I think this whitespace should stay there.

Toggl.Daneel/ViewControllers/LoginViewController.cs
```diff
@@ -83,6 +86,16 @@ public override void ViewDidLoad()
                       .For(v => v.BindAnimatedEnabled())
                       .To(vm => vm.NextIsEnabled);
 
+            bindingSet.Bind(backButton)
+                      .For(v => v.BindAnimatedEnabled())
+                      .To(vm => vm.IsLoading)
+                      .WithConversion(invertedBoolConverter);
+
+            bindingSet.Bind(ForgotPasswordButton)
+                      .For(v => v.Enabled)
```
simonrozsival: Why is one animated and the other not? (Moreover I can't see any visual difference in the simulator)

arturs105: Yeah, i couldn't see the animation as well. @willsb probably knows more about the animated enabled binding.

willsb: The `BindAnimatedEnabled ` binding is just for `UIBarButton`s

arturs105: `BindAnimatedEnabled` looks the same as binding `Enabled`, even did a test with slow animations on simulator, is that custom binding even needed, or should it be fixed to actually have an animation?

PR #884, opened 2017-11-02
Round to second

### LINKED ISSUE #705: verify / enforce time entry time && duration
please validate/ensure that:
  a) start time is not sub second precise (round/truncate it to seconds only part).
  b) stop time is not sub second precise (round/truncate it to seconds only part).
  c) duration is non-decimal number of full seconds between start and stop for stopped time 
entries. 
  d) duration field for running time entries encodes same moment of time as start field.

simonrozsival:
Closes #705 

Squash whenever


Toggl.Foundation/TimeService.cs
```diff
@@ -9,19 +9,31 @@ public sealed class TimeService : ITimeService
...
 
-        public DateTimeOffset CurrentDateTime => DateTimeOffset.UtcNow;
+        private readonly Func<DateTimeOffset> now;
+
+        public DateTimeOffset CurrentDateTime => now();
 
         public IConnectableObservable<DateTimeOffset> CurrentDateTimeObservable { get; }
 
         public TimeService(IScheduler scheduler)
+            : this(scheduler, () => DateTimeOffset.UtcNow)
```
willsb: Allowing mocked time inside the time service a massive overkill. Just check that the returned time doesn't have millis and that's enough.

Toggl.Foundation.Tests/TimeServiceTests.cs
```diff
@@ -35,6 +31,20 @@ public void ReturnsTheSameObjectsRegardlessOfTheTimeTheObserversSubscribed()
                              .Should().HaveCount(expectedNumberOfMessages)
                              .And.BeEquivalentTo(firstObserver.Messages);
             }
+
+            [Property]
+            public void ReturnsCurrentTimeFlooredToTheCurrentSecond(DateTimeOffset now)
+            {
+                var timeService = new TimeService(testScheduler, () => now);
+                DateTimeOffset roundedNow;
```
kornelijepetak: Even though mono compiler allows this, MS C# doesn't want to compile like this so I think it would be wise to initialize this to a `default(DateTimeOffset)`.

Otherwise, it's: `Error CS0165 Use of unassigned local variable 'roundedNow'`

simonrozsival: OK, thanks, I didn't try this in Windows.

Toggl.Foundation/TimeService.cs
```diff
@@ -18,10 +18,13 @@ public TimeService(IScheduler scheduler)
...
                 Observable
                     .Interval(TimeSpan.FromSeconds(1), scheduler)
-                    .Select(_ => DateTimeOffset.UtcNow)
+                    .Select(_ => floor(CurrentDateTime))
                     .Publish();
 
             disposable = CurrentDateTimeObservable.Connect();
         }
+
+        private DateTimeOffset floor(DateTimeOffset t)
```
kornelijepetak: Not that it matters for this PR, but such a function is maybe a good candidate for an application-wide extension method. Could be useful maybe in some other places in the future. Just a thought for consideration.

simonrozsival: All `DateTimeOffset` values in the app come either from the server, where they are rounded, or from this service, where they are rounded. It should not be needed anywhere else.

kornelijepetak: Allrighty then.

kornelijepetak:
At @willsb's convenience.

PR #894, opened 2017-11-03
Indicate selected project

### LINKED ISSUE #619: Selected project indicator
When user edits a TE which already has a project, then highlight this project in a projects list:
- Background color is 12% of the project color

toggl - ios - projects selected 1x

arturs105:
Closes #619

Squash whenever


Toggl.Foundation.MvvmCross/ViewModels/SelectProjectViewModel.cs
```diff
@@ -71,9 +71,13 @@ public override void Prepare(SelectProjectParameter parameter)
                        .StartWith(Text)
                        .SelectMany(text => dataSource.AutocompleteProvider.Query(text, AutocompleteSuggestionType.Projects))
                        .Select(suggestions => suggestions.Cast<ProjectSuggestion>())
+                       .Select(setSelectedProect)
                        .Subscribe(onSuggestions);
         }
 
+        private IEnumerable<ProjectSuggestion> setSelectedProect(IEnumerable<ProjectSuggestion> suggestions)
```
simonrozsival: Typo (`Proect`)

Toggl.Foundation.MvvmCross/ViewModels/SelectProjectViewModel.cs
```diff
@@ -71,9 +71,13 @@ public override void Prepare(SelectProjectParameter parameter)
...
+            => suggestions.Select(s => { s.Selected = s.ProjectId == projectId; return s; });
```
simonrozsival: I believe you could use `=> suggestions.Do(s => s.Selected = s.ProjectId == projectId);` here.

Toggl.Daneel/Views/ProjectSuggestionViewCell.cs
```diff
@@ -48,6 +49,7 @@ public override void AwakeFromNib()
                 var colorConverter = new MvxRGBValueConverter();
                 var taskCountConverter = new TaskCountConverter();
                 var visibilityConverter = new MvxVisibilityValueConverter();
+                var projectSelectedColorCombiner = new ProjectSelectedColorValueCombiner(0.12f);
```
simonrozsival: Please make `0.12f` a constant.

Toggl.Foundation.MvvmCross/ViewModels/SelectProjectViewModel.cs
arturs105: There is no `Do` method on `IEnumerable` (even with Linq).  However, there is a `ForEach` method, which (i guess) does more or less the same, so  I tried the following in the `Initialize` method:

```
infoSubject.AsObservable()
           .StartWith(Text)
           .SelectMany(text => dataSource.AutocompleteProvider.Query(text, AutocompleteSuggestionType.Projects))
           .Select(suggestions => suggestions.Cast<ProjectSuggestion>())
           .Do(suggestions => suggestions.ForEach(s => s.Selected = s.ProjectId == projectId))
           .Subscribe(onSuggestions);
```

But that doesn't work.

simonrozsival: You are right, that is only for observables. Then keep it this way.

PR #896, opened 2017-11-03
Trim long tag names in edit view

### LINKED ISSUE #859: Long tag names
Cut tags that are longer than 30 characters and add three dots at the end.

arturs105:
Closes #859


Toggl.Foundation.MvvmCross/ViewModels/EditTimeEntryViewModel.cs
```diff
@@ -298,5 +305,13 @@ private void clearTagsIfNeeded(long currenctWorkspaceId, long newWorkspaceId)
             RaisePropertyChanged(nameof(Tags));
             RaisePropertyChanged(nameof(HasTags));
         }
+
+        private string trimTag(string tag)
+        {
+            if (tag.Length <= maxTagLength)
```
simonrozsival: This is not a correct way of counting UTF-16 characters, see https://csharp.2000things.com/2014/01/09/1007-getting-length-of-string-that-contains-surrogate-pairs/

Toggl.Foundation.MvvmCross/ViewModels/EditTimeEntryViewModel.cs
```diff
@@ -298,5 +305,13 @@ private void clearTagsIfNeeded(long currenctWorkspaceId, long newWorkspaceId)
...
+                return tag;
+
+            return $"{tag.Substring(0, maxTagLength)}...";
```
simonrozsival: ; You should use `StringInfo` here as well not to damage strings with surrogate characters: https://msdn.microsoft.com/en-us/library/bx3zb6dh(v=vs.110).aspx

Toggl.Foundation.Tests/MvvmCross/ViewModels/EditTimeEntryViewModelTests.cs
```diff
@@ -690,5 +690,52 @@ private List<IDatabaseTag> createTags(int count)
                 ViewModel.Tags.Should().HaveCount(0);
             }
         }
+
+        public sealed class TheTagsProperty : EditTimeEntryViewModelTest
+        {
+            [Theory]
+            [InlineData(31)]
+            [InlineData(50)]
+            public async Task CutsLongTagNames(int tagLength)
```
simonrozsival: Generate also tags with weird symbols and test if these are trimmed correctly.

Toggl.Foundation.MvvmCross/ViewModels/EditTimeEntryViewModel.cs
arturs105: There is no such method in `Toggl.Foundation.MvvmCross`, however  in `Toggl.Foundation.Tests` there is. Could it be because of .NET portable?

PR #910, opened 2017-11-06
Fix UI jumps when toggling tasks in suggestions view and select project view

### LINKED ISSUE #864: Tapping a task's arrow takes me to the top of the projects list
1) Start new TE
2) Tap @
3) Scroll down and find a project with a task(s)
4) Tap a task's arrow
5) It takes user back to the top of the list (thought it shouldn't)

*Same applies when using a edit to change a project view*

simonrozsival:
Closes #864 

This is 🍎  UI fix. Do a functional test when when reviewing.

Squash whenever


Toggl.Daneel/Extensions/TableViewExtensions.cs
```diff
@@ -0,0 +1,34 @@
...
+using CoreGraphics;
+using UIKit;
+
+namespace Toggl.Daneel.Extensions
+{
+    public static class TableViewExtensions
+    {
+        public static void CorrectOffset(this UITableView table, CGPoint originalOffset, nfloat originalFrameHeight)
+        {
+            table.SetContentOffset(originalOffset, false);
```
arturs105: Is this line needed here? In some cases this causes the offset to be set twice.

Toggl.Daneel/ViewControllers/SelectProjectViewController.cs
```diff
@@ -1,7 +1,12 @@
-﻿using MvvmCross.Binding.BindingContext;
+﻿using System;
+using CoreGraphics;
```
arturs105: 🔥

Toggl.Daneel/ViewControllers/StartTimeEntryViewController.cs
```diff
@@ -1,12 +1,15 @@
 ﻿using System;
 using System.Collections.Generic;
+using CoreGraphics;
```
arturs105: 🔥

Toggl.Daneel/Extensions/TableViewExtensions.cs
simonrozsival: Yes, it is needed - the command is handled in a such way, that the whole group is removed and re-inserted. This causes the whole view to jump to the top. So on this line, I restore the offset which was used before the group was removed. Then, if the conditions below are met, I adjust the offset to the new content with an animation. If this line wasn't present, then it would jump to the top and then animate scrolling down.

arturs105: Oh, i see.

PR #913, opened 2017-11-06
Push tags

### LINKED ISSUE #907: Implement tag sync
To support tag creation, we need to mark them as needing-sync just like time entries.

During push sync we then need to check if there are any tags to be synced, and if so, sync them. Tag sync has a higher priority than TE sync, given the dependency.

In case tag creation fails, they have to be marked as unable to sync and the returned reason has to be saved, just like with TEs.

simonrozsival:
Closes #907 

Most of changed files are just renamed files (changed directory structure).

Squash whenever


kornelijepetak:
On an unrelated note, I've just noticed that `BaseUnsyncableEntityState.CreateUnsyncableFrom` has a parameter incorrectly named `reson` instead of `reason`. Maybe a new PR to fix a typo.

simonrozsival:
I will make a separate PR for this 💄 problem.

PR #915, opened 2017-11-06
Remove deleted tags during pull sync

### LINKED ISSUE #900: Implement tag deletion during pull sync
Tags can be deleted on the server. The response looks like this:
```json
[
    {
        "id": 3727054,
        "workspace_id": 2175975,
        "name": "tag1",
        "at": "2017-11-03T15:01:47.423741Z"
    },
    {
        "id": 3727052,
        "workspace_id": 2175975,
        "name": "tag1",
        "at": "2017-11-03T14:58:29.662409Z",
        "deleted_at": "2017-11-03T14:58:29.662409Z"
    },
    {
        "id": 3727051,
        "workspace_id": 2175975,
        "name": "tag2",
        "at": "2017-11-03T14:56:47.011079Z",
        "deleted_at": "2017-11-03T14:56:47.011079Z"
    },
    {
        "id": 3727053,
        "workspace_id": 2175975,
        "name": "tag3",
        "at": "2017-11-03T15:00:02.545934Z"
    },
    {
        "id": 3727055,
        "workspace_id": 2175975,
        "name": "tag4",
        "at": "2017-11-03T15:03:05.586949Z"
    }
]
```

We need to change the tag model and the `TagSyncSelector` so that it can be automatically deleted during pull sync.

simonrozsival:
Closes #900 

This PR contains the first Realm "migration" - it is in fact only schema version bump, but with explicit comment in the `MigrationCallback` which describes that there is nothing to do when updating from version 0 to version 1. I followed Realm documentation (https://realm.io/docs/xamarin/1.5.0/#migrations), but I would definitely like @willsb to review this part if it is OK.

Do a functional test when reviewing:

1. create some tags in the web app
2. pull-sync in the web app
3. create a time entry and add few tags
4. delete some of the tags you added to the TE in the web app
5. force pull-sync
6. the TE doesn't have the tags you deleted shown when viewed in the edit time entry view
7. none of the tags you deleted is in the list of available tags when editing tags of a time entry
8. push sync (by forcing sync)
9. none of the deleted tags re-appears in the web app

Squash whenever


Toggl.PrimeRadiant.Realm/Database.cs
```diff
@@ -38,13 +39,29 @@ public Database()
...
+                var realm = GetRealmInstance();
 
                 using (var transaction = realm.BeginWrite())
                 {
                     realm.RemoveAll();
                     transaction.Commit();
                 }
             });
+
+        public static Realms.Realm GetRealmInstance()
```
amulware: Instead of exposing this publicly, wouldn't it be best to call the migration code only once when the `Database` is created? Though we always have to pass in the config? Maybe hand that down to all repositories?

I'm really not a fan of this static stuff (and allocating the whole configuration for every realm access). :/

Toggl.PrimeRadiant.Realm/RealmAdapter.cs
```diff
@@ -189,5 +187,7 @@ private TRealmEntity getEntityFromResult(IConflictResolutionResult<TModel> resul
                 results.Add(new UpdateResult<TModel>(originalRivalId, rival));
             }
         }
+
+        private Realms.Realm realm => Database.GetRealmInstance();
```
amulware: Does this even work? It will now get multiple realm instances for a single transaction, each time creating a configuration and doing whatever realm does internally.

amulware:
I did not do a functional review, and would also like to have @willsb look this over - maybe after the above is addressed.

Toggl.PrimeRadiant.Realm/Database.cs
simonrozsival: We always have to pass the config, that is why I made all those other changes to the Realm related code. The config has to have the correct version ID, if it has a higher one, then the MigrationCallback is invoked, if it has a lower one (defaults to 0), then it should fail.

I agree with the remark on the static method. I will change this so we pass it to the repositories and realm adapters directly.

Toggl.PrimeRadiant.Realm/RealmAdapter.cs
simonrozsival: This is used in such a way so it uses only one instance for single transaction.

simonrozsival: But I will make it more obvious.

Issue #891, opened 2017-11-03
Limit time entry description to 3000 characters


amulware:
Otherwise, backend might refuse.

We'll need some UI message showing when this happens.

I don't think we should actually prevent the user from entering too many characters though, in case they copy paste and want to edit in the app. (?) Or maybe this limit is so ridiculously high that we don't really care, and can just clip the text when it gets too long? Having some message seems nice to me though.

It should be limited in:
- [x] Start time entry view
- [ ] Edit time entry view

pe0ny:
I think we just should prevent entering more than 3000 chars in UI and show a message when the limit is exceeded (when user tries to enter more). 
Also, i believe we need this error message both for TE creation view and edit view.

pe0ny:
- [ ] Display exceeded number of characters 
- [ ] Disable save button
- [ ] When user deletes exceeded characters, remove the number and enable save button again

![toggl - ios - add new entry char limit 1x](https://user-images.githubusercontent.com/5064445/32367143-ca66652c-c089-11e7-8098-e822c0e7fc57.png) ![toggl - ios - edit char limit 1x](https://user-images.githubusercontent.com/5064445/32367144-ca90260a-c089-11e7-91c6-f04f70f10a7d.png)

PR #916, opened 2017-11-06
Limit description in start time entry view

arturs105:
Ref #891


Toggl.Foundation.Tests/MvvmCross/ViewModels/StartTimeEntryViewModelTests.cs
```diff
@@ -918,5 +919,73 @@ public void IsClearedWhenThereAreNoWordsToQuery()
...
+
+                ViewModel.DescriptionRemainingBytes.Should()
+                    .Be(MaxTimeEntryDescriptionLengthInBytes);
+            }
+
+            [Theory]
+            [InlineData("Hello fox")]
+            [InlineData("Some emojis: 🔥😳👻")]
+            [InlineData("Some weird characters: āčēļķīņš")]
+            [InlineData("Some arabic bois: ظۓڰڿڀ")]
```
simonrozsival: This is the only thing in the PR I am not very sure about. I had to google what this means and I got many different results. Please consider if some people might consider this offensive and maybe change the value, since this is a open source project.

Issue #798, opened 2017-10-26
'New project' view


pe0ny:
Needed fields:
- [x] Project name (inline)
- [x] Workspace (opens WS list)
- [x] Client (opens Clients list)
- [x] ~~Template (opens Templates list)~~
- [x] Public/private switch
- [x] Color (opens Color picker popup)

Other notes:
- [x] When user taps 'Create', save and select project and direct user back to
  - [x] time entry edit view
  - [x] time entry creation view
- [x] When user taps 'X', close 'New project' view and direct user back to
  - [x] projects list view
  - [x] time entry creation view

![toggl - ios - add new project 1x](https://user-images.githubusercontent.com/5064445/32043256-e3ca8fcc-ba42-11e7-9313-b533f8fa2910.png)

amulware:
I suggest we implement this without any of the child screens so that those can be implemented in parallel and one by one for easier review.

willsb:
What's the default color for projects?

pe0ny:
As far as I know there isn't default color. Webapp assigns different color every time and perhaps we can do the same - randomly pick one color from the list?

amulware:
I think picking a random colour from the pre-set ones is a decent solution.

But question: Do we want to allow the user to set the 'default billable' flag? If so, workspaces have a 'projects default billable flag default' that should probably be respected when selecting a workspace (and as default from the default workspace)?

amulware:
Also: How do we prevent duplicate project names?

pe0ny:
Let's not add billable flag atm because then it makes sense to let user set billable rate as well. But yes, when it's set in ws settings, we should take this into account.

amulware:
So we will use the default settings from the workspace and just take that value 'secretely' behind the scenes?

amulware:
Note that there are several fields that have to be set as well, even though they are not exposed in the UI:

Since they are not shown in the UI, they can probably just be computed when the project is finally created.

Note that these properties also only should be set if the workspace has the Pro feature enabled.

- `Billable` if there is a template, take its value for this field, otherwise, take the workspace's `ProjectsBillableByDefault `
- `AutoEstimates` is taken from template - otherwise null

@pe0ny please check if these are the only fields that have to be set next to the list in the original comment.


Also note that templates can only be selected when the Pro feature is enabled. @pe0ny What does the UI show for free workspaces - is the templates field hidden or disabled?

amulware:
[...]

PR #917, opened 2017-11-06
Add project creation to DataSource

willsb:
Ref #798 

Squash whenever


amulware:
This is not handling several properties of the project model. Why is that?

willsb:
This is handling the properties that appear in the create view. More will be added as (and if) needed

amulware:
We need at least the billable field, and maybe also some related to estimataes (see issue).

amulware:
Only the billable field is needed and should be taken from the selected workspace default field.

amulware:
Or be left at `null` for workspaces missingthe Pro feature.

willsb:
So only pro workspaces have the `ProjectsBillableByDefault` flag?

amulware:
Any workspace 'has' the field, but it should only be considered if they are Pro, yes.

Toggl.Foundation/Models/Project.cs
```diff
@@ -0,0 +1,137 @@
...
+
+            public DateTimeOffset? At { get; private set; }
+
+            public DateTimeOffset? ServerDeletedAt { get; private set; }
+
+            public bool IsDeleted { get; private set; }
+
+            public long? ClientId { get; private set; }
+
+            public long? TemplateId { get; private set; }
```
amulware: Let's remove this until we need it.

Toggl.Foundation/DTOs/CreateProjectDTO.cs
```diff
@@ -0,0 +1,19 @@
...
+    {
+        public string Name { get; set; }
+
+        public string Color { get; set; }
+
+        public bool IsPrivate { get; set; }
+
+        public long? ClientId { get; set; }
+
+        public long? TemplateId { get; set; }
```
amulware: :fire: until needed

PR #919, opened 2017-11-07
Invalid transition state

simonrozsival:
Squash whenever


Toggl.Foundation.Tests/Sync/States/InvalidTransitionStateTests.cs
```diff
@@ -0,0 +1,55 @@
...
+        [Fact]
+        public void ThrowsInvalidOperationException()
+        {
+            Exception caughtException = null;
+            var state = new InvalidTransitionState("Some message");
+
+            var observable = state.Start();
+            observable.Subscribe(_ => { }, (Exception exception) => caughtException = exception);
+
+            caughtException.Should().NotBeNull();
```
kornelijepetak: Is `Should().NotBeNull()` necessary when followed by `BeOfType()`? I mean, if it is `null`, then it will not be of a specific exception type.

Toggl.Foundation.Tests/Sync/States/InvalidTransitionStateTests.cs
```diff
@@ -0,0 +1,55 @@
...
+        [Property]
+        public void ThrowsInvalidOperationExceptionWithASpecificMessage(NonEmptyString message)
+        {
+            Exception caughtException = null;
+            var state = new InvalidTransitionState(message.Get);
+
+            var observable = state.Start();
+            observable.Subscribe(_ => { }, (Exception exception) => caughtException = exception);
+
+            caughtException.Message.Should().Be($"Invalid transition: {message}");
```
kornelijepetak: Is there a better way to handle this string so we don't repeat it twice, or is it enought that we use this test as a safeguard?

simonrozsival: I need to use it at least twice - once in the actual implementation and once in the test - these should (must?) be independent so we know that the message is correct and somebody doesn't change it by accident.

Toggl.Foundation.Tests/Sync/States/InvalidTransitionStateTests.cs
kornelijepetak: I get what you mean, I just believe that functionally, you would get the same outcome for the same input. In case of the first problem you are testing, that no exception would be thrown, the `BeOfType` would also fail the test. But if you'd like the intention to be more explicit, which I can also argue for, that is completely fine. If you think no changes are needed, then do not make them 😄

Toggl.Foundation.Tests/Sync/States/InvalidTransitionStateTests.cs
kornelijepetak: If you'd pull them from the same source, you would not be able to change it by accident cause it would be able to be changed only in that one location. But if one would change the call-site, the test would still detect it. 

Whether there IS a location where this could reasonably be put so it'd be accessible to both call sites, that is an entirely another story.

PR #925, opened 2017-11-08
Push projects

### LINKED ISSUE #909: Implement project sync
To support project creation, we need to mark them as needing-sync just like time entries.

During push sync we then need to check if there are any projects to be synced, and if so, sync them. Project sync has a higher priority than TE Sync, given the dependency.

In case project creation fails, they have to be marked as unable to sync and the returned reason has to be saved, just like with TEs.

simonrozsival:
Closes #909 

This PR is in fact the same as #921, but with each mention of `Client` replaced with `Project`.

Squash whenever


Toggl.Foundation/Toggl.Foundation.csproj
```diff
@@ -211,11 +214,6 @@
   </ItemGroup>
   <ItemGroup />
   <ItemGroup />
-  <ItemGroup>
-    <Folder Include="Sync\States\Push\TimeEntries\" />
-    <Folder Include="Sync\States\Push\BaseStates\" />
-    <Folder Include="Sync\States\Push\Tags\" />
-    <Folder Include="Helper\" />
-  </ItemGroup>
+  <ItemGroup />
```
kornelijepetak: Oh, Visual Sudio and its beauties.

PR #927, opened 2017-11-08
Remove fixed test account from UI tests

### LINKED ISSUE #749: Remove fixed test account from UI tests
I'd rather have it create a more predictable new user per test. We could simply call the appropriate sign up api in the test code, and then login to that user and run the ui test.

simonrozsival:
Closes #749 

Squash whenever


simonrozsival:
It works, but I found out that in newer versions of NUnit it won't be supported (https://github.com/nunit/nunit/blob/d922adae5cd30ad5544ee693f6ae6177722e3569/src/NUnitFramework/framework/Internal/AsyncInvocationRegion.cs#L76), so I will change it.

PR #928, opened 2017-11-08
Add code owners

willsb:
More info on code owners can be found [here](https://help.github.com/articles/about-codeowners/)

I need everyone to review and agree on this one. Please submit files you want to own so I can add them.

Squash as you please once approved by everyone, but not before.


arturs105:
Looks good, however, there is one thing bugging me - inconsistencies in commented lines:

```
# Some have spaces after the number sign
#Others don't
```

willsb:
> I don't want certain files to become 'off limits' only being changed by a few certain people. Redundancy is important.

This is not what code owners is about, I don't get this concern.

> This also introduces a certain amount of inflexibility, since we will be unlikely to update this file often. Not to mention that it is very rough the way it is, but if we make it more detailed it will grow state and outdated even faster.

How so? This will notify owners for a review, but their review is not mandatory by default. This is a way to have file owners notified instantly (regardless of whether the PR maker requests their review or not). By no means we have to make owner's review mandatory.

> If I understand correctly, code owners will always be assigned to PRs that change the files they own, but it is not reasonable to expect all code owners to always be available for review. Do we manually unassign them? Or is reviewing when assigned no longer obligatory? Code owners seems to muddle this process, and it is not at all clear to me how you see this working. And if you can make it work in some clear way, it should be documented.

The value I see in this is having the resident expert notified of changes so he/she can catch critical errors. It won't prevent anyone from merging MvvmCross changes while I'm on vacation; but will give me a list of every file change after I return. It'll also always make sure that I get shown as a possible reviewer, even when GitHub's algorithm wouldn't suggest me. I find this all very valuable.

amulware:
I know it's not what code owners is about, but I'm worried that people will interpret it that way in the end, maybe even subconsciously to some degree.

And, as I said above, right now we do not simply request/dismiss reviewers without actually meaning it. With this change, the automatically assigned reviews will be sort of optional, and how can we tell that apart from specifically requested ones?

How do you see that working?

amulware:
After more research, my understanding is now that code owner reviews are not optional, or in other words: If code owners are set up, one code owner always has to approve a PR. I do not see that working for anything that has only a single owner. How do we handle people being on vacation, or similar?

Also, my points above have not been addressed.

willsb:
From the docs linked in the PR description:

> When someone with admin or owner permissions has enabled required reviews, they also can _optionally require approval from a code owner_ before the author can merge a pull request in the repository.
Emphasis mine. So no, it's not mandatory and it can be ignored in extreme cases.

So if a code owner is on vacation, someone else can still review and the code owner gets a notification that the file he owns was changed (because of the review request). 

You say that the code owner review will be "Optional", but I only see it being optional when the reviewer is on vacation. Let me explain why: Right now we already sort of prefer leaving the reviewing process to someone who has a better expertise at a subject. Code owners will make it simpler for us to pick reviewers. At the end of the day, code owners would eventually have to take a look at that file anyways, so this simply automates the process and creates a map of who's the resident expert in what.

amulware:
So you would leave code owner review as optional in the system, but within the team consider them non-optional, except for special circumstances?

What happens in the case where two people are assigned as code owners for the same files, so both get requested for a PR making changes to them. If one of them approves the PR, is this enough?

If yes: Can the PR author dismiss the other review?
  - If yes: how do we prevent accidental dismissal or dismissal of reviewers who actually wanted to review?
  - If no: do we then allow merging of PRs without all requested reviews having approved to be allowed? Again, how does this handle reviewers who still wanted to review, and how do we then prevent the closing of PRs where not all requested have approved in general?

If no: Will that not delay us, maybe for no good reason?

Also what happens in case a code owner is automatically requested but hasn't reviewed, yet there are two or more reviews from other people who are regarded as knowledgeable on the topic? Do we have to wait for code owner review? Or can we just dismiss it? Or just merge without all requested having approved (basically same questions as above).

amulware:
Since the points above remain unaddressed, I'm closing this for now. Feel free to reopen this or another PR in the future.

PR #929, opened 2017-11-08
Add Project view

willsb:
Ref #798 

This adds the project view. Best reviewed on a per commit basis.

Squash at will


Toggl.Daneel/ViewControllers/EditProjectViewController.cs
```diff
@@ -0,0 +1,74 @@
...
+                      .For(v => v.AttributedText)
+                      .To(vm => vm.ClientName)
+                      .WithConversion(new AddNewAttributedStringConverter(Resources.AddClient, ClientLabel.Font.CapHeight));
+
+            bindingSet.Apply();
+        }
+
+        private void prepareViews()
+        {
+            // Resize the switch
```
arturs105: I don't think this comment is necessary.

Toggl.Foundation.MvvmCross/ViewModels/EditProjectViewModel.cs
```diff
@@ -0,0 +1,120 @@
...
+        private readonly ITogglDataSource dataSource;
+        private readonly IMvxNavigationService navigationService;
+
+        private long? clientId;
+        private long workspaceId;
+
+        public bool IsPrivate { get; set; }
+
+        [DependsOn(nameof(Name))]
+        public bool SaveEnabled => !string.IsNullOrEmpty(Name);
```
simonrozsival: So `"             "` is a valid project name? Shouldn't it be `string.isNullOrWhiteSpace(name)` and later trim the name? What are the valid names accepted by backend?

Toggl.Foundation.Tests/MvvmCross/ViewModels/EditProjectViewModelTests.cs
```diff
@@ -0,0 +1,243 @@
...
+            public async Task ReturnsNull()
+            {
+                ViewModel.Prepare();
+
+                await ViewModel.CloseCommand.ExecuteAsync();
+
+                await NavigationService.Received()
+                    .Close(Arg.Is(ViewModel), Arg.Is<long?>(result => result == null));
+            }
+        }
```
simonrozsival: Maybe add test that this command does not call `Create` on the data source.

Toggl.Foundation.Tests/MvvmCross/ViewModels/EditProjectViewModelTests.cs
```diff
@@ -0,0 +1,243 @@
...
+                ViewModel.Prepare();
+
+                await ViewModel.PickColorCommand.ExecuteAsync();
+
+                await NavigationService.Received()
+                    .Navigate<MvxColor, MvxColor>(typeof(SelectColorViewModel), Arg.Any<MvxColor>());
+            }
+
+            [Fact]
+            public async Task SetsTheReturnedColorAsTheColorProperty()
```
simonrozsival: This could be a nice property test with two colors as parameters.

Toggl.Foundation.MvvmCross/ViewModels/EditProjectViewModel.cs
amulware: This is a very important point. Please make sure to clarify with product/backend. Also make sure this is enforced not just in the UI, but in Foundation itself.

Toggl.Foundation.MvvmCross/ViewModels/EditProjectViewModel.cs
```diff
@@ -0,0 +1,120 @@
...
+            DoneCommand = new MvxAsyncCommand(done);
+            CloseCommand = new MvxAsyncCommand(close);
+            PickColorCommand = new MvxAsyncCommand(pickColor);
+            PickClientCommand = new MvxAsyncCommand(pickClient);
+            PickWorkspaceCommand = new MvxAsyncCommand(pickWorkspace);
+        }
+
+        public override void Prepare()
+        {
+            var random = new Random();
```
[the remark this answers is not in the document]
willsb: Literally two lines below 😄 
No need to hold onto this instance, since we're only using it once and it gets instantiated when users open the view.

Toggl.Foundation.Tests/MvvmCross/ViewModels/EditProjectViewModelTests.cs
willsb: 👍

Toggl.Foundation.Tests/MvvmCross/ViewModels/EditProjectViewModelTests.cs
willsb: I'd have to create my own generators. The added value wouldn't be worth the problem

Toggl.Foundation.MvvmCross/ViewModels/EditProjectViewModel.cs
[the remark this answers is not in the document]
amulware: Sorry, I meant when the method is called, not when the random is used, but you answered that too. Since it happens across time, it is most likely alright. We'll see if the colours aren't very random.

PR #933, opened 2017-11-09
Ask confirmation before deleting a time entry

### LINKED ISSUE #911: Ask confirmation when deleting TE
Ask confirmation 'Delete' or 'Cancel' when user taps 'Delete' in edit view.

arturs105:
Closes #911


Toggl.Foundation.MvvmCross/Services/IDialogService.cs
```diff
@@ -9,5 +9,11 @@ public interface IDialogService
             string message,
             string confirmButtonText,
             string dismissButtonText);
+
+        Task<string> ShowActionSheet(
+            string title,
+            string message,
+            string cancelText,
+            params (string text, bool destructive)[] actions);
```
willsb: This is rather leaky, since `ActionSheet` is an iOS term. I'd change the naming. I'd also create a struct so we are not using Tuples across projects.

Toggl.Foundation.MvvmCross/ViewModels/EditTimeEntryViewModel.cs
```diff
@@ -159,8 +167,18 @@ private void subscribeToTimeServiceTicks()
                 .Subscribe(_ => RaisePropertyChanged(nameof(Duration)));
         }
 
-        private void delete()
+        private async Task delete()
         {
+            var result = await dialogService.ShowActionSheet(
+                title: null,
```
willsb: If this is only used here and this place ignores title and message, why add it?

Toggl.Foundation.MvvmCross/ViewModels/EditTimeEntryViewModel.cs
```diff
@@ -159,8 +167,18 @@ private void subscribeToTimeServiceTicks()
...
+                message: null,
+                cancelText: Resources.Cancel,
```
willsb: Also why is `cancelText` simply treated like an action?

Toggl.Foundation.Tests/MvvmCross/ViewModels/EditTimeEntryViewModelTests.cs
```diff
@@ -77,33 +79,90 @@ public sealed class TheCloseCommand : EditTimeEntryViewModelTest
...
+
+                    await ViewModel.DeleteCommand.ExecuteAsync();
+
+                    await DataSource.SyncManager.DidNotReceive().PushSync();
+                }
+            }
+
+            public sealed class WhenUserCnacels : TheDeleteCommand
+            {
+                public WhenUserCnacels()
```
willsb: Typo

Toggl.Foundation.MvvmCross/Services/IDialogService.cs
arturs105: Do you have any suggestions for the method name? `ShowMultipleChoiceDialog`? `ShowMultipleChoiceAlert`?

willsb: `ShowMultipleChoiceDialog` is a good one :)

Toggl.Foundation.MvvmCross/ViewModels/EditTimeEntryViewModel.cs
arturs105: Do you mean that we should return null, if user cancels? Because that makes sense and i didn't think of that (obviously).

Issue #797, opened 2017-10-26
Project creation


pe0ny:
- [x] User can create new project in time entry creation flow
- [x] User can create new project in projects view (when editing time entry)
- [x] Show '+ Create project "new-project-name"' at the top of the list when user has typed in project name that doesn't exist yet.
- [x] By default assign new project under default workspace
- [x] When user taps '+ Create project "new-project-name"', open 'New project' view

amulware:
This is only about showing the "+add new project" at the top of the list in both creation and edit views.

The screen itself is in #798 and its details are in further issues.

PR #937, opened 2017-11-09
Create project in StartTimeEntryView

willsb:
Ref #797 
Ref #798

Squash whenever


arturs105:
Code-wise all changes look good to go, however, the close icon in the create project view does not match zeplin. Also, i don't think current animation (sliding up) makes much sense for a close icon like the one in zeplin. I think it implies horizontal navigation like the default one in iOS, but a confirmation from @pe0ny is needed.

willsb:
Horizontal navigation makes little to no sense for a card, so dunno

Toggl.Daneel/ViewSources/StartTimeEntryTableViewSource.cs
```diff
@@ -63,6 +73,26 @@ public override UITableViewCell GetCell(UITableView tableView, NSIndexPath index
...
+        public override nint NumberOfSections(UITableView tableView)
+        {
+            var sections = base.NumberOfSections(tableView);
+            return sections == 0 && IsSuggestingProjects ? 1 : sections;
+        }
+
+        protected override object GetItemAt(NSIndexPath indexPath)
+        {
+            if (ItemsSource.Count() == 0 && IsSuggestingProjects)
+                return $"Create project \"{CurrentQuery}\"";
```
simonrozsival: I thing we should either show this if `CurrentQuery.Length > 0` or we should display something else than `+ Create project ""`. I didn't find anything related to this edge case in the issues, but I think it behaves really strange to see this message. In this case I would expect to see something like `+ Create project` without the quotes.

Also I think that this PR does not do exactly what is in the specification:
> Show '+ Create project "new-project-name"' at the top of the list when user has typed in project name that doesn't exist yet.

I interpret that that you should show the + button even if there are some suggestions, but none of them _exactly_ matchest the query (i.e.,  `ItemsSource.All(x => x.Name != CurrentQuery)`). It makes sense, because you might have a project with a name which would be substring of a different project.

pe0ny:
I would still use slide up animation, not left-right.

PR #938, opened 2017-11-10
Fix color serialization

willsb:
Squash whenever


willsb:
I thought about Foundation because that's the only place where we can manually create a project, but you are correct

Toggl.Ultrawave.Tests/Models/ProjectTests.cs
```diff
@@ -45,5 +46,15 @@ public void CanBeSerialized()
...
         }
+
+        [Fact]
+        public void ColorIsAlwaysLowercaseWhenSerialized()
+        {
+            var serializer = new Ultrawave.Serialization.JsonSerializer();
+            var json = serializer.Serialize(validProject);
+            var newProject = serializer.Deserialize<Project>(json);
+
+            newProject.Color.Should().Be(newProject.Color.ToLower());
```
simonrozsival: This test is not correct - you test that the color value is lowercase after _deserialization_, not after serialization. Use regexp instead:
```csharp
json.Should().MatchRegex("^.*\"color\":\"#f61d38\".*$");
json.Should().NotMatchRegex("^.*\"color\":\"#F61D38\".*$");
```

Toggl.Ultrawave.Tests/Models/ProjectTests.cs
```diff
@@ -1,5 +1,6 @@
 ﻿using System;
 using FluentAssertions;
+using Newtonsoft.Json;
```
simonrozsival: 🔥

Toggl.Ultrawave/Models/Project.cs
```diff
@@ -1,4 +1,5 @@
 ﻿using System;
+using Newtonsoft.Json;
```
simonrozsival: 🔥

PR #942, opened 2017-11-10
Select Client View

### LINKED ISSUE #799: Clients view
- [x] User can access clients list when creating new project (from 'New project' view)
- [x] Show clients only from WS that was selected from New project view
- [x] Show 'No client' option as well so user can deselect a client if needed
- [x] '<' arrow takes user back to 'New project' view without saving anything
- [x] When user selects a client from a list, save and take user back to 'New project' view

toggl - ios - add client 1x

### LINKED ISSUE #800: Client creation
- [x] User can create new client in Clients view
- [x] Show '+ Create client "new-client-name"' at the top of the list when user has typed in client name that doesn't exist yet.
- [x] Assign new client under the workspace that was selected in 'New project' view
- [x] When user taps 'Create', save and select client, and direct user back to 'New project' view
- [x] When user taps '<' arrow, take user back to 'New project' view without saving anything
- [x] When user taps 'x' clean the input field and show list of existing clients

toggl - ios - add new client 1x

willsb:
Closes #799 
Closes #800 

Do not approve without a functional test


Toggl.Foundation.MvvmCross/ViewModels/SelectClientViewModel.cs
```diff
@@ -0,0 +1,99 @@
...
+
+        private async Task selectClient(string clientName)
+        {
+            var clientId = allClients.FirstOrDefault(c => c.Name == clientName)?.Id ?? 0;
+            await navigationService.Close(this, clientId);
+        }
+
+        private async Task createClient()
+        {
+            var client = await dataSource.Clients.Create(Text.Trim(), workspaceId);
```
simonrozsival: There should be some validation here - if you put several whitespaces in the `Text`, then this throws an exception (Builder validation).

Toggl.Foundation/Models/Client.cs
```diff
@@ -0,0 +1,99 @@
...
+
+            public Builder SetIsDeleted(bool isDeleted)
+            {
+                IsDeleted = isDeleted;
+                return this;
+            }
+
+            private void ensureValidity()
+            {
+                if (string.IsNullOrEmpty(Name))
```
simonrozsival: Here should be better validation - name length must be < 255 bytes.

Toggl.Foundation.MvvmCross/ViewModels/SelectClientViewModel.cs
```diff
@@ -0,0 +1,99 @@
...
+        private IEnumerable<IDatabaseClient> allClients;
+
+        public string Text { get; set; } = "";
+
+        public bool SuggestCreation
+        {
+            get
+            {
+                var text = Text.Trim();
+                return !string.IsNullOrEmpty(Text) && !Suggestions.Any(s => s == text);
```
simonrozsival: That should be `!string.IsNullOrEmpty(text)`, right? Now you allow creating a whitespace-only client name.

PR #943, opened 2017-11-10
Push user during sync

### LINKED ISSUE #931: Support pushing of user object
This is needed for #860

I'm not sure if this is as easy as it sounds or if we want to be more careful to not accidentally change/mess up any settings. Please take care to investigate that and maybe post any relevant considerations below.

simonrozsival:
Closes #931 

Squash whenever


amulware:
I didn't review in detail, but it looks fine at a glace.

PR #944, opened 2017-11-13
Add character counter to description editing

### LINKED ISSUE #891: Limit time entry description to 3000 characters

arturs105:
Closes #891

### DIFF
--- a/Toggl.Daneel/Binding/TextViewWithCounterCountedTextTargetBinding.cs
+++ b/Toggl.Daneel/Binding/TextViewWithCounterCountedTextTargetBinding.cs
@@ -22,6 +22,14 @@ namespace Toggl.Daneel.Binding
             Target.CountedText = value;
         }
 
+        protected override void Dispose(bool isDisposing)
+        {
+            base.Dispose(isDisposing);
+
+            if (!isDisposing) return;
+            Target.CountedTextChanged -= onCountedTextChanged;
+        }
+
         private void onCountedTextChanged(object sender, EventArgs e)
         {
             FireValueChanged(Target.CountedText);
--- a/Toggl.Daneel/Binding/TextViewWithCounterCountedTextTargetBinding.cs
+++ b/Toggl.Daneel/Binding/TextViewWithCounterCountedTextTargetBinding.cs
@@ -1,0 +1,30 @@
+﻿using System;
+using MvvmCross.Binding;
+using MvvmCross.Binding.Bindings.Target;
+using Toggl.Daneel.Views;
+
+namespace Toggl.Daneel.Binding
+{
+    public sealed class TextViewWithCounterCountedTextTargetBinding : MvxTargetBinding<TextViewWithCounter, string>
+    {
+        public const string BindingName = "CountedText";
+
+        public override MvxBindingMode DefaultMode => MvxBindingMode.TwoWay;
+
+        public TextViewWithCounterCountedTextTargetBinding(TextViewWithCounter target)
+            : base(target)
+        {
+            target.CountedTextChanged += onCountedTextChanged;
+        }
+
+        protected override void SetValue(string value)
+        {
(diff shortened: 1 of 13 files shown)


kornelijepetak:
Does VC still need a `IUITextViewDelegate` since the `TextViewWithCounter` is now a delegate to itself?

Toggl.Daneel/Binding/TextViewWithCounterCountedTextTargetBinding.cs
```diff
@@ -0,0 +1,30 @@
...
+    public sealed class TextViewWithCounterCountedTextTargetBinding : MvxTargetBinding<TextViewWithCounter, string>
+    {
+        public const string BindingName = "CountedText";
+
+        public override MvxBindingMode DefaultMode => MvxBindingMode.TwoWay;
+
+        public TextViewWithCounterCountedTextTargetBinding(TextViewWithCounter target)
+            : base(target)
+        {
+            target.CountedTextChanged += onCountedTextChanged;
```
willsb: Please clear this subscription upon disposal to avoid ze leaks :non-potable_water:

Toggl.Daneel/Views/TextViewWithCounter.cs
```diff
@@ -0,0 +1,137 @@
...
+using Toggl.Foundation.MvvmCross.Helper;
+using UIKit;
+using static Toggl.Multivac.Extensions.StringExtensions;
+
+namespace Toggl.Daneel.Views
+{
+    [Register(nameof(TextViewWithCounter))]
+    public sealed class TextViewWithCounter : UITextView, IUITextViewDelegate
+    {
+        public event EventHandler CountedTextChanged;
```
willsb: Why do you need both the TargetBinding and a property with an event? That looks redundant

Toggl.Daneel/Views/TextViewWithCounter.cs
```diff
@@ -0,0 +1,137 @@
...
+        private string countedText;
+        public string CountedText
+        {
+            get => countedText;
+            set
+            {
+                if (countedText == value) return;
+                countedText = getTextWithoutCounter(value);
+                setAttributedText(countedText);
+                CountedTextChanged?.Invoke(this, EventArgs.Empty);
```
willsb: `CountextTextChanged.Raise(this);`

Toggl.Daneel/Views/TextViewWithCounter.cs
```diff
@@ -0,0 +1,137 @@
...
+            {
+                selectedRange.Location += text.LengthInGraphemes();
+            }
+            return true;
+        }
+
+        [Export("textViewDidChange:")]
+        public new void Changed(UITextView textView)
+        {
+            CountedText = Text.Replace('\n', ' ');
```
willsb: A comment here would be nice, since this is not very obvious

Toggl.Daneel/Views/TextViewWithCounter.cs
arturs105: Because the binding needs to know, when the value is updated from inside the view to update the view model. And i can't use any events from `UITextView`, since i'm already using a delegate.

arturs105:
Putting on hold until #783 is closed.

kornelijepetak:
@arturs105, #783 has been closed and merged. What is the status of this?

arturs105:
Closing this due to merge conflicts in .xib file. Will start from a clean branch.

PR #945, opened 2017-11-13
Project creation from edit and start time entry

### LINKED ISSUE #797: Project creation

### LINKED ISSUE #798: 'New project' view

### LINKED ISSUE #939: Create new project from suggestions needs revision
This is a follow-up of issue #797, which was already partially implemented in PR #937 .

1. When the query after the `@` symbol is empty, we should display something else than `+ Create project ""`. I didn't find anything related to this edge case in the previous issue, but I think it is really strange to see this message. In this case I would expect to see something like `+ Create project` without the quotes.

2. Also I think that PR #937 did not do exactly what was in the specification:

    > Show '+ Create project "new-project-name"' at the top of the list when user has typed in project name that doesn't exist yet.

    I interpret that in a such way that you should show the `+` button even if there are some suggestions, but none of them __exactly__ matchest the query (i.e., `ItemsSource.All(x => x.Name != CurrentQuery)`). It makes sense, because you might have a project with a name which would be substring of a different project.

I would like to ask @pe0ny or @amulware for confirmation or for any further comments to my suggestions.

willsb:
Closes #797 
Closes #798 
Closes #939 

This refactors the mechanism for suggesting a project creation. Allows one to create a project from both start time entry and edit view, as per specs.

Functional test this and try to find edge cases.


simonrozsival:
Other than ☝️ everything looks good to me.

Issue #926, opened 2017-11-08
Make sure timeline displays correctly over night


amulware:
If the app is left open (or in the background) over midnight, the timeline may not display correctly the next day.

- At the very least the 'today' and 'yesterday' labels will be incorrect.

- It also may be that newly started time entries on the next day show more than once or not at all in the timeline until the app is restarted.

- Further, the date might not always be displayed correctly in the edit view in this case.

More information and screenshots available internally.

The above points should be investigated, and whatever can be reproduced should be fixed.

simonrozsival:
I tried implementing this with an observable, which emits a value on every midnight (local time). The time entries log could subscribe to this observable and every time it produces a new value, it would update the UI. My progress can be seen in branch `midnight-ticking`.

For some reason there is a problem with unit tests - they run well in Visual Studio (Windows), but fail in console and the whole process somehow crashes. I believe this is a problem in the TestScheduler, but I'm not sure.

I will now unasign myself from this issue and do something else, because I'm stuck here. I will try to get back to this issue later this week, unless somebody implements in the meantime.

simonrozsival:
This wasn't closed automatically, so I'm closing it manually now.

PR #951, opened 2017-11-14
Midnight ticking

simonrozsival:
Ref #926 

This PR introduces an observable in the `TimeService` which publishes a value at _local_ midnight. Also this PR changes the `CurrentDateTime` so it does not return `DateTimeOffset.UtcNow`, but rather `scheduler.Now`. This is equivalent (see unit test).

Squash whenever.


simonrozsival:
WIP: For some reason tests works well in Windows, but not in macOS :-/

PR #958, opened 2017-11-14
Handle changing ids during entity creation

### LINKED ISSUE #950: Handle changing ids during entity creation
Projects depending on other projects that are being synced at the same time could try and use outdated ids and cause crashes.

Here are the cases I see right now:

- projects when creating TE
- tags when creating TE
- clients when creating project

and in the near future:

- workspace when creating project
- workspace when creating client
- workspace when creating tag
- workspace when creating TE

I do not know how to best handle this, but we have to, soon.

simonrozsival:
Closes #950 

This PR introduces a new DB field in nearly all models. A migration is thus needed. The `OriginalId` field is nullable and so the default value is `null`, which is OK and doesn't need any other change other than schema version bump. User's of the public release have schema version 0 and we already did a change in the DB, so `develop` branch uses currently schema version 1. You will have to uninstall the previous version for testing, but user's won't need to. If we decide to increase the schema version every time we make a change in the database, not at most 1 per release, then I would bump the version to 2, but I don't think it is needed.

_Naming issue:_ I wasn't able to come up with anyghing better than `IModifiableId` interface. Suggestions for better names are welcome.

Please pay close attention to `StartTimeEntryViewModel.done()` method. I would like to have the changed logic confirmed by @amulware or @pe0ny (when user removes the project using backspace, the workspace of the project is kept - this is different from the previous behavior (caused by lines 300 and 396)). I believe previous behavior would lead to problems with tags and maybe other problems as well.

Make a functional test when reviewing. I recommend adding `dataSource.SyncManager.PushSync();` on line 238 of `StartTimeEntryViewModel.cs`. Be aware that there are currently no unit tests and no way of testing `RealmAdapter` class other than through a thorough functional test.

Squash whenever


Toggl.PrimeRadiant.Realm/Models/RealmConstructors.cs
```diff
@@ -205,33 +216,35 @@ public void SetPropertiesFrom(IDatabaseTimeEntry entity, Realms.Realm realm)
...
             var skipWorkspaceFetch = entity?.WorkspaceId == null || entity.WorkspaceId == 0;
-            RealmWorkspace = skipWorkspaceFetch ? null : realm.All<RealmWorkspace>().Single(x => x.Id == entity.WorkspaceId);
+            RealmWorkspace = skipWorkspaceFetch ? null : realm.All<RealmWorkspace>().Single(x => x.Id == entity.WorkspaceId || x.OriginalId == entity.WorkspaceId);
             var skipProjectFetch = entity?.ProjectId == null || entity.ProjectId == 0;
-            RealmProject = skipProjectFetch ? null : realm.All<RealmProject>().SingleOrDefault(x => x.Id == entity.ProjectId);
-            var skipTaskFetch = RealmProject == null || entity?.TaskId == null || entity.TaskId == 0;
-            RealmTask = skipTaskFetch ? null : realm.All<RealmTask>().SingleOrDefault(x => x.Id == entity.TaskId);
+            RealmProject = skipProjectFetch ? null : realm.All<RealmProject>().Single(x => x.Id == entity.ProjectId || x.OriginalId == entity.ProjectId);
+            var skipTaskFetch = entity?.TaskId == null || entity.TaskId == 0;
+            RealmTask = skipTaskFetch ? null : realm.All<RealmTask>().Single(x => x.Id == entity.TaskId || x.OriginalId == entity.TaskId);
```
amulware: These should be `SingleOrDefault`, right?

amulware:
And for the record, the workspace should indeed be kept when removing the project.

Toggl.PrimeRadiant.Realm/Models/RealmConstructors.cs
simonrozsival: I think this will have to be done in a separate PR - the first hotfix we did for `beta-1` was just in the `.cs` file, not in the template.

simonrozsival: OK

simonrozsival:
Ref #901

Toggl.PrimeRadiant.Realm/Models/IModifiableId.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
```
arturs105: None of these are necessary.

amulware:
@arturs105 is the above behaviour reproducible consistently? And is it present outside this branch?

If the answer to the first question is yes, no matter the second, please create a `bug` issue for this.

amulware:
:poop:

PR #959, opened 2017-11-14
Add clear text button to tags view

arturs105:


simonrozsival:
What should this PR do? :) It's hard to review, when it is not specified what the outcome should look like. This is not related to any issue?

arturs105:
Sorry about that. There was no issue for this and i had nothing to do. This adds a small "x" button next to the text in select tags view. When clicked, the button clears the entered text.

Toggl.Daneel/ViewControllers/SelectTagsViewController.designer.cs
```diff
@@ -1,55 +1,54 @@
...
+// actions made in the UI designer. If it is removed, they will be lost.
+// Manual changes to this file may not be handled correctly.
 //
 using Foundation;
-using System;
 using System.CodeDom.Compiler;
 
 namespace Toggl.Daneel.ViewControllers
 {
-    [Register ("SelectTagsViewController")]
```
willsb: Gross auto-generated tabs :(

arturs105: I couldn't find any setting that would change this. Closest was "Convert tabs to spaces", but that has been on since the beginning. I also noticed that several other .designer.cs files have tabs in them.

PR #962, opened 2017-11-14
Hsv Color Picker

### LINKED ISSUE #803: Color picker
- [x] User can access color picker popup when creating new project (from 'New project' view)
- [x] 'x' takes user back to 'New project' view without saving anything
- [x] 'Save' takes user back to 'New project' view with saving selected color
- [x] User checkbox to indicate selected color
- [x] Show default colors to all users and picker only to starter, premium and enterprise workspaces
- [x] When user uses picker then fill empty circle with this color and add checkbox

toggl - ios - color picker 1x toggl - ios - color picker pro 1x

willsb:
:rainbow: :rainbow: :rainbow: :rainbow: :rainbow: :rainbow: :rainbow: :rainbow: :rainbow: :rainbow: :rainbow: :rainbow: :rainbow: :rainbow: :rainbow: 

Closes #803 

Do a functional test. Try verifying workspaces with different access levels. This should only show workspaces with the Pro feature.


Toggl.Foundation.MvvmCross/Helper/Color.cs
```diff
@@ -75,6 +76,59 @@ public static class StartTimeEntry
             public static readonly MvxColor AddIconColor = new MvxColor(75, 200, 0);
         }
 
+        internal static MvxColor FromHSV(float hue, float saturation, float value)
+        {
+            int r = 0, g = 0, b = 0;
+
+            if (saturation == 0)
+            {
+                r = g = b = (int)(value * 255.0f + 0.5f);
```
amulware: Why this 0.5 offset here and below?

Toggl.Foundation.MvvmCross/Parameters/ColorParameters.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System;
+using MvvmCross.Platform.UI;
+
+namespace Toggl.Foundation.MvvmCross.Parameters
+{
+    public sealed class ColorParameters
+    {
+        public MvxColor Color { get; set; }
+
+        public bool IsPremium { get; set; }
```
amulware: Should we at this point still talk about 'features' or 'subscription plans'? Since the condition for when to show the colour picker might change, shouldn't we call this `AllowCustomColors` or something along those lines?

kornelijepetak: I second that motion. With @amulware's example, `AllowCustomColor` does not require understanding of relationship between premium and this feature at this place, but leaves it to be setup where it belongs. Is also more resistant to changes.

Toggl.Foundation.MvvmCross/Helper/Color.cs
willsb: ©️ 🍝

Toggl.Daneel/Converters/ComplementValueConverter.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+using System.Globalization;
+using MvvmCross.Platform.Converters;
+
+namespace Toggl.Daneel.Converters
+{
+    public class ComplementValueConverter : MvxValueConverter<float, float>
+    {
+        protected override float Convert(float value, Type targetType, object parameter, CultureInfo culture)
+            => Math.Abs(((float)(value / 1) - 1));
```
kornelijepetak: Isn't `value / 1` the same as `value`? Am I missing some sci-fi math here?

Toggl.Daneel/Converters/ComplementValueConverter.cs
```diff
@@ -0,0 +1,15 @@
...
+
+        protected override float ConvertBack(float value, Type targetType, object parameter, CultureInfo culture)
+            => Math.Abs(((float)(value / 1) - 1));
```
kornelijepetak: as above.

Toggl.Foundation.MvvmCross/Helper/Color.cs
kornelijepetak: Yeah, knowing that `value` should go from 0 to 1 (probably), this 0.5 really does look scary. 😄

Toggl.Foundation.MvvmCross/ViewModels/SelectColorViewModel.cs
```diff
@@ -35,16 +47,57 @@ public SelectColorViewModel(IMvxNavigationService navigationService)
             SelectColorCommand = new MvxCommand<SelectableColorViewModel>(selectColor);
         }
 
-        public override void Prepare(MvxColor parameter)
+        public override void Prepare(ColorParameters parameter)
```
kornelijepetak: Somehow I feel this _parameter_ should be _parameters_... (pun intended).
It represents something in plural.

Toggl.Foundation.MvvmCross/Helper/Color.cs
amulware: It just causes different rounding basically. (the (int) cast always rounds down)
I probably wouldn't have that in an implementation I wrote because I only want 1 -> 255, not 0.999 - or maybe I would? No idea, it most likely doesn't really matter.

Toggl.Foundation.MvvmCross/ViewModels/SelectColorViewModel.cs
willsb: I don't like changing the name of arguments for an method that has been overridden

amulware: I think, the issue here really is that the type and argument name both contain 'parameter' - which arguably is not the best name for an argument in the first place, but maybe should then be changed as a whole, and I think this isn't our interface?

willsb: The interface the VM implements is not ours, but the Parameters are. We are simply lego naming the `Parameter` suffix for clarity. I don't see that as an issue, but if we wanna change it here, we'd have to change it everywhere, which is not in this PRs scope

PR #964, opened 2017-11-14
User sign up

### LINKED ISSUE #825: Signup UI

willsb:
Closes #825 

Squash whenever


Toggl.Foundation/Login/LoginManager.cs
```diff
@@ -51,6 +51,25 @@ public IObservable<ITogglDataSource> Login(Email email, string password)
                     });
         }
 
+        public IObservable<ITogglDataSource> SignUp(Email email, string password)
+        {
+            if (!email.IsValid)
+                throw new ArgumentException("A valid email must be provided when trying to Login");
```
amulware: 'signup', not 'login'

Toggl.Ultrawave/ApiClients/UserApi.cs
```diff
@@ -54,6 +54,7 @@ public IObservable<IUser> SignUp(Email email, string password)
             return CreateObservable<User>(endPoints.Post, new HttpHeader[0], json);
         }
 
+        [Preserve(AllMembers = true)]
```
amulware: Why is this necessary? The classes are used just above.

Also, is there a reason that we are not simply skipping this whole project for link-stripping purposes?

willsb: The default constructor was being stripped away, dunno why

We have discussed this skipping process and opted for make the libraries linker safe instead. This approach is better in the long run

amulware: Noted.

PR #965, opened 2017-11-15
Textview placeholder

### LINKED ISSUE #783: Create a custom UITextView with placeholder

kornelijepetak:
Closes #783.

Adds a placeholder to a TextView by inheriting from TextView and using attributed strings.
Placeholder shows only when its `Text` is empty and the control does not have a focus.
The `Text` property is bindable through custom `TargetBinding`.

Please do a thorough functional testing!

Supports changing (in a non-bindable manner):

* `PlaceholderText`
* `PlaceholderColor`
* `PlaceholderSize`


Toggl.Daneel/Views/TextViewWithPlaceholder.cs
```diff
@@ -0,0 +1,108 @@
...
+        {
+            get => text;
+            set
+            {
+                if (text == value)
+                    return;
+                
+                text = value;
+                updateAttributedText(value);
+                TextChanged.Raise(this);
```
arturs105: Would it be possible to achieve the same result by not overriding `Text`, but executing this setter code in `Changed(..)`?

arturs105: If it works, the custom binding may not even be needed.

Toggl.Daneel/Views/TextViewWithPlaceholder.cs
```diff
@@ -0,0 +1,108 @@
...
+            }
+        }
+
+        public string PlaceholderText { get; set; }
+
+        private UIStringAttributes defaultTextAttributes => new UIStringAttributes
```
arturs105: This could be a field instead of a property, that way attributes could be reused.

Toggl.Daneel/Views/TextViewWithPlaceholder.cs
```diff
@@ -0,0 +1,108 @@
...
+        {
+            ParagraphStyle = new NSMutableParagraphStyle { Alignment = base.TextAlignment },
+            Font = UIFont.SystemFontOfSize(Font.PointSize)
+        };
+
+        private UIStringAttributes placeholderAttributes => new UIStringAttributes
```
arturs105: Same as above.

Toggl.Daneel/Views/TextViewWithPlaceholder.cs
```diff
@@ -0,0 +1,108 @@
...
+            Delegate = this;
+        }
+
+        protected void updateAttributedText(string text)
+        {
+            if (string.IsNullOrEmpty(text))
+            {
+                AttributedText = isFocused
+                    ? new NSAttributedString("")
+                    : new NSMutableAttributedString(PlaceholderText ?? "", placeholderAttributes);
```
arturs105: This doesn't have to be mutable

Toggl.Daneel/ViewControllers/EditTimeEntryViewController.cs
```diff
@@ -201,29 +202,14 @@ private void prepareViews()
...
             PreferredContentSize = View.Frame.Size;
-            DescriptionTextView.Delegate = this;
             BillableSwitch.Resize(switchHeight);
             prepareDescriptionField();
         }
 
         private void prepareDescriptionField()
         {
             DescriptionTextView.TintColor = Color.StartTimeEntry.Cursor.ToNativeColor();
+            DescriptionTextView.PlaceholderText = "Add description";
```
arturs105: The string should be a resource.

Toggl.Foundation.MvvmCross/Helper/Color.cs
```diff
@@ -96,7 +96,13 @@ public static class Settings
             public static readonly MvxColor SyncStatusText = new MvxColor(144, 146, 147);
         }
 
-        public static readonly MvxColor[] DefaultProjectColors = 
+        public static class Common
+        {
+            public static readonly MvxColor Placeholder = pinkishGrey;
```
arturs105: I think `TextPlaceHolder` would be a better name.

Toggl.Daneel/ViewControllers/EditTimeEntryViewController.cs
kornelijepetak: Totally true, I just forgot about that. Will change.

Toggl.Daneel/Views/TextViewWithPlaceholder.cs
```diff
@@ -0,0 +1,108 @@
...
+using Toggl.Foundation.MvvmCross.Helper;
+using UIKit;
+
+namespace Toggl.Daneel.Views
+{
+    [Register(nameof(TextViewWithPlaceholder))]
+    public class TextViewWithPlaceholder : UITextView, IUITextViewDelegate
+    {
+        protected readonly int defaultPlaceholderSize = 14;
+        protected readonly UIColor defaultPlaceholderColor = Color.Common.Placeholder.ToNativeColor();
```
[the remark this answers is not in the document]
kornelijepetak: I was thinking along the lines of being able to change the defaults in inherited classes. If you don't think this is necessary, I will switch to `private`.

Toggl.Daneel/Views/TextViewWithPlaceholder.cs
kornelijepetak: This is how I did it first, but it has a flaw. And that is the fact that setting the 'AttributedText' also changed the `Text`, which means that once the placeholder was shown, the `Text` property changed to 'Add description'. Then when you tap on the textview to add a description, the text is already set and you first have to delete 'Add description' before writing your own. I was trying to find a solution around that, but it seems I must not let the underlying textview do its own thing with the Text.

Toggl.Daneel/Views/TextViewWithPlaceholder.cs
kornelijepetak: I thought about it before doing it this way, but I wanted to enable this to change if underlying properties change (alignment, color, size). I agree that we rarely need to change those once they are setup, so I can accept the argument. In light of my explanation, do you still think it would be better to change it to cached despite losing this functionality?

Toggl.Daneel/Views/TextViewWithPlaceholder.cs
kornelijepetak: Same as above.

Toggl.Foundation.MvvmCross/Helper/Color.cs
kornelijepetak: What about `PlaceholderText`? It would be more consistent with the names above.

Toggl.Daneel/Views/TextViewWithPlaceholder.cs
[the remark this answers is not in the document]
arturs105: We can make them private for now and change it in the future, if needed.

Toggl.Daneel/Views/TextViewWithPlaceholder.cs
```diff
@@ -0,0 +1,108 @@
...
+
+        public TextViewWithPlaceholder(IntPtr handle) : base(handle)
+        {
+            PlaceholderSize = defaultPlaceholderSize;
+            PlaceholderColor = defaultPlaceholderColor;
+
+            Delegate = this;
+        }
+
+        protected void updateAttributedText(string text)
```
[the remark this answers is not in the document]
kornelijepetak: I believe this will have to be overridden to enable some other features in the future. Or at least some parts of it would need to be copied.

Toggl.Daneel/Views/TextViewWithPlaceholder.cs
arturs105: I think we should still change it to a field, since changing those properties isn't needed now. We can change this to a property in the future if the need arises.

Toggl.Daneel/Views/TextViewWithPlaceholder.cs
arturs105: Same as above.

Toggl.Daneel/Views/TextViewWithPlaceholder.cs
arturs105: Okay. Makes sense, then we can keep it this way.

Toggl.Daneel/Views/TextViewWithPlaceholder.cs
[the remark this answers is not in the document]
arturs105: Let's make it private for now and change if needed.

kornelijepetak:
@amulware and @willsb After talking to @arturs105 we decided to call you in. :)
I'd like you to review this, especially the part about the cached property.

1. Are we using a naming convention when implementing private properties that have the same-ly named backing field? I used `_` prefix, but feel free to correct me if we are.

2. At first I did those as normal properties that create `UIStringAttributes` every time they are needed, but since this is happening on every character entered, it may be better to cache them. We are losing the ability to change some styling, but we probably won't need those, so it's ok.

3. The reason the initialization can't be moved to a constructor is the fact that `base.Font` will be `null` at that time and so the initial value can't be properly set. The effect of this is that it will ignore such settings on TextView in IB, because those changes will come in too late.

Is this ok, or do you want anything done differently?

Toggl.Daneel/Views/TextViewWithPlaceholder.cs
```diff
@@ -0,0 +1,111 @@
...
+                updateAttributedText(value);
+                TextChanged.Raise(this);
+            }
+        }
+
+        public string PlaceholderText { get; set; }
+
+        private UIStringAttributes _defaultTextAttributes;
+        private UIStringAttributes defaultTextAttributes 
+            => _defaultTextAttributes ?? (_defaultTextAttributes = new UIStringAttributes
```
willsb: Why not make this a `Lazy<T>` instead?

Toggl.Daneel/Views/TextViewWithPlaceholder.cs
```diff
@@ -0,0 +1,111 @@
...
+            => _placeholderAttributes ?? (_placeholderAttributes = new UIStringAttributes
+            {
+                ParagraphStyle = new NSMutableParagraphStyle { Alignment = TextAlignment },
+                ForegroundColor = PlaceholderColor,
+                Font = UIFont.SystemFontOfSize(PlaceholderSize)
+            });
+
+        public TextViewWithPlaceholder(IntPtr handle) : base(handle)
+        {
+            PlaceholderSize = defaultPlaceholderSize;
```
willsb: No init shall be made in `IntPtr` constructors

kornelijepetak: shall not or should not? In any case, why?

kornelijepetak:
I tried with `Lazy<T>` but this won't work, because it's called before the user has a chance to update `PlaceholderColor` and `PlaceholderSize` from outside. It would work with `defaultTextAttributes` but not with `placeholderAttributes`.

So what I did is initialized the `defaultTextAttributes` in `AwakeFromNib` where all require properties are already available (i.e. the font setup through IB). However, the `placeholderAttributes` will need to be refreshed when `PlaceholderColor` or `PlaceholderSize` change, so I have set it up so that it updates the style accordingly. Now, they are cached, but with correct values.

Please look again at the code and correct if there's anything else that's a problem.

Issue #947, opened 2017-11-13
API token has been reset UI


pe0ny:
- [x] When user has reset the API token:
a) ask user to log back in by entering his/her password
b) sign out (possible data loss)
Don't allow user to go back and access any data.
 
![toggl - ios - api token reset 1x](https://user-images.githubusercontent.com/5064445/33199984-5e037640-d0fb-11e7-8592-96a618940944.png)

- [x] When user decides to sign out and has some unsynced data, show similar warning we show on 'regular' sign out while having unsynced data.

![toggl - ios - api token reset confirmation 1x](https://user-images.githubusercontent.com/5064445/33199977-5905a9c4-d0fb-11e7-87ab-fe76d8438a97.png)

amulware:
@pe0ny What should happen if the user logs out, has no unsynced time entries, but has unsynced projects?
Do we only check time entries? Or do we check all things? Should the copy then be changed?

It's a bit of an edge case, but let's make sure we make a conscious decision.

amulware:
Also, in accordance with #751 make sure that this is the first thing people see if restarting the app after the bad token was detected. This is important so that our users can lock other people out from seeing their data if they lose their device - at least superficially.

simonrozsival:
Just FYI settings view currently only checks for unsynced TEs.

amulware:
Good point, we _probably_ want the same logic for both. (?)

pe0ny:
@amulware Yes, we want the same logic for both (settings and when API token has been reset). I would say we should check all the data (time entries, projects, clients, tags etc).

And let's change copy to:
Are you sure?
You have some unsynced data. Any unsynced data will be lost

willsb:
This is not on Zeplin. It'd be good to have these there

pe0ny:
Updated Zeplin files:
API token reset UI: https://zpl.io/aRmXYgp
And confirmation: https://zpl.io/bWQXYrx

willsb:
How should we display error when the users types in the wrong password?

pe0ny:
We'll display the same error message that we display in 'normal' log in view and remove the API part: https://zpl.io/25Nxpg8

willsb:
Okay, but where? Below the help text? Above it? Should we replace the help text with the error message?

pe0ny:
[...]

PR #969, opened 2017-11-15
Data source has unsynced data

simonrozsival:
Ref #947 

This PR adds a single place where it is checked, whether there are unsynced data. This should be used in the Settings view (that is also implemented in this PR) and later in the screen where user can re-login after the API token was reset.

Also from now on we do not check only for unsynced TEs, but all kinds of entities.

Squash whenever.


Toggl.Foundation/DataSources/TogglDataSource.cs
```diff
@@ -45,6 +48,22 @@ public TogglDataSource(ITogglDatabase database, ITogglApi api, ITimeService time
...
         public IAutocompleteProvider AutocompleteProvider { get; }
 
+        public async Task<bool> HasUnsyncedData()
+            => await hasUnsyncedData(database.TimeEntries)
+               || await hasUnsyncedData(database.Projects)
+               || await hasUnsyncedData(database.User)
+               || await hasUnsyncedData(database.Tasks)
+               || await hasUnsyncedData(database.Clients)
+               || await hasUnsyncedData(database.Tags)
+               || await hasUnsyncedData(database.Workspaces);
```
amulware: This will run the queries one after the other - do we have an idea how this behaves performance wise compared to running them at the same time?

Toggl.Foundation/DataSources/TogglDataSource.cs
```diff
@@ -45,6 +48,22 @@ public TogglDataSource(ITogglDatabase database, ITogglApi api, ITimeService time
         public ISyncManager SyncManager { get; }
         public IAutocompleteProvider AutocompleteProvider { get; }
 
+        public async Task<bool> HasUnsyncedData()
```
amulware: Not having this be an observable feels a bit out of place. Any particular reasoning for this to be a task instead?

Toggl.Foundation/DataSources/TogglDataSource.cs
simonrozsival: I was aware that this runs one after another, but Realm queries are very fast and I don't think this would be a problem. I can rewrite it to run in parallel, if you think we could benefit from that.

Toggl.Foundation/DataSources/TogglDataSource.cs
simonrozsival: This was because of the way I implemented it - it is simpler to return directly the `Task<bool>` and not wrap it in `Observable.Return(...);` Maybe I can think of some way I can merge the observables so they run in parallel and this could return a single observable.

amulware:
If there is an easy and nice way to turn it into observables that would be neat. If not, I'm alright with this too.

Issue #661, opened 2017-10-02
The sync manager should expose failures of the sync process


simonrozsival:
Right now there is no way of showing the user that the sync failed (for any reason). Every sync process ends with showing the user a green tick and a "Synced" message even if there was an error.

There are several distinct scenarios of failure and they should be presented in different ways:
- There is no Internet connection
- The client or the API is outdated (HTTP 410, 418)
- There was some other error

When the sync is triggered automatically (automatic push after user changed some time entry), we shouldn't show any errors so we don't confuse the user.

The reason I created this issue was me thinking about the way to implement #579 - namely whether to even initiate the push sync when there is no Internet connection. We discussed this with @willsb and we figured out that it would be best to expose the errors with the same observable we use today, but add more sync states besides `Pull`, `Push`, `Sleep` - something like `NoInternetConnection`, `ClientOutdated`, `ApiOutdated`, `Error`, but only when the full sync was initiated by the user.

amulware:
Ref #751

amulware:
I'm no longer sure if we should use the sync manager's observable. It significantly complicates the class and makes it has to handle a lot of different things, while it's really only meant to handle the thread safety and correct order of operations.

Instead, I'm leaning towards custom observables (maybe use a container with several of them for different events? unless we thing a single one is enough), injected into the states, and have the states push these events out as side effects.

Comments and other ideas are very welcome.

This may have to be solved together with #751, or separately for the issues not covered there.

In both cases, I'd like to have a design proposed here first and only fully implemented if we have a consensus.

simonrozsival:
I agree that we should have a different observable for the UI which would be easier to consume - it could emit events like `Syncing`, `Synced`, `Offline`, `Failed`. I think that for simplicity we should treat both of these issues separately.

simonrozsival:
So my suggestion is:

- add new observable `SyncProgressObservable` to `StateMachineOrchestrator`
- in the the `startSync` method emit `Syncing`
- in the `onStateEvent` handler look for relevant state machine events (`StateMachineError`, StateMachineDeadEnd`) and emit corresponding events (error => `Offline` or `Failed`, deadend => `Synced`)
- the observable will be accessible through the `SyncManager` in the same way the `StateObservable` is accessible currently
- UI can subscribe to this new observable instead of the `StateObservable` and have more detailed status information (now we show "Synced" even when the whole process failed)
[...]

PR #975, opened 2017-11-16
Clear queue when an error occurs during sync

simonrozsival:
Ref #661 

This PR makes sure we don't immediately continue with synchronization when an error is reported. Please review the thread-safety test twice, just for sure.. 😃 

Squash whenever


Toggl.Foundation.Tests/Sync/SyncStateQueueTests.cs
```diff
@@ -31,119 +33,166 @@ private void queueSyncs(params SyncState[] states)
...
+
+                Clear();
+                var returnValues = DequeueUntilSleep();
+
+                returnValues.Should().HaveCount(1);
+                returnValues.ShouldBeSameEventsAs(Sleep);
+            }
+
+            [Property]
+            public void RetursSleepAfterClearNoMatterWhatWasQueuedPreviously(bool[] pushPull)
```
willsb: `NonEmptyArray<bool>`, maybe?

amulware:
Given that neither of us reviewed the tests well, I'll go over those again in a minute.

simonrozsival:
@amulware That can happen and I don't think it would be incorrect behavior. We might temporarily block further syncing, but what should be the reason to enable it again? A timeout? Force sync?

amulware:
Well, the thing is that the UI causes the sync not knowing that it shouldn't anymore. It's up to the UI to figure out if it thinks syncing works again, but because of this race condition, I don't see that working. Though I'm not sure if this can actually be avoided at all.

amulware:
Unless we can somehow enforce that no sync is scheduled until the error is handled on the ui thread or something. Or maybe we need to temporarily freeze the state machine, and the unfreeze it manually in the ui.

But I'm not sure it's worth all these complications. Let's maybe leave it how it is and see if we notice anything bad happening because of this. (Maybe graphical glitches, or weird double navigation to the error screens?)

simonrozsival:
Then maybe besides clearing the queue, we can block further syncing for the next `n` seconds, or until user forces sync manually.

The error might be being offline (probably most frequent one), timeout (server is not available for some reason and retry loop reaches a delay > 60s), or if there is a bug in the app. 410, 418 and 401 will destroy (freeze) the `SyncManager`, so that is not a case which we need to consider.

I think we should leave this as is right now, and maybe create an issue to consider possible solutions for this problem later.

amulware:
In the case of offline and timeout, I dont really care, only 401, 410 and 418 are problematic to me. But if you say that you will freeze in that case, then maybe it's mostly alright? Or is the UI going to decide on that freeze? If so there is still a race condition... But let's leave this as is for now and we'll see.

simonrozsival:
If we run into these 3 cases, the user has to either update the app and start it again or log in again (=> new data source with new sync manager). Therefore we should freeze it in those cases - it is already prepared in #978.

Toggl.Foundation.Tests/Sync/SyncStateQueueTests.cs
```diff
@@ -31,119 +34,166 @@ private void queueSyncs(params SyncState[] states)
...
+            [Property]
+            public void RetursSleepAfterClearNoMatterWhatWasQueuedPreviously(NonEmptyArray<bool> pushPull)
+            {
+                DequeuedStates.Clear();
+                QueueSyncs(BoolToEnumValues(pushPull.Get));
+
+                Clear();
+                var returnValues = DequeueUntilSleep();
+
+                returnValues.Should().HaveCount(1);
```
amulware: Are these count-checks necessary?

Issue #306, opened 2017-07-25
Handle 410 and 418 status codes


amulware:
410 indicates that the app is using a deprecated version of the API. 418 indicates that the current client itself has been deprecated.

Both should only happen for people who do not use the app in a very long time, or if backend detects an acute problem with a specific app version and has to block it.

In either case, we should prevent the user from entering further data (since it cannot be synced) and prompt them to update to the newest version of the app.

Optionally, we could allow users to ignore the warning, and continue tracking locally, at the risk of data loss. *dramatic finger wagging*

pe0ny:
Link to the design: https://zpl.io/VxYdoRa
User can't close the popup or do anything before he/she has updated the app.

amulware:
Closing in favour of #751

PR #978, opened 2017-11-16
Sync progress observable

simonrozsival:
Ref #661, ref #306 

I decided to move the logic from `StateMachineOrchestrator` to `SyncManager`, because the progress of the whole sync process is too high level for the orchestrator - e.g., it wouldn't be possible to emit `Syncing`, do `Pull`, (no `Synced` and `Syncing`), do `Push`, and `Synced` afterwards.

~Please note that currently orchestrator does not emit errors, that will come in a separate PR.~

Squash whenever


Toggl.Foundation/Sync/SyncManager.cs
```diff
@@ -14,10 +16,13 @@ public sealed class SyncManager : ISyncManager
 
         private bool isFrozen;
 
+        private readonly ISubject<SyncProgress> progress;
+
         public bool IsRunningSync { get; private set; }
 
         public SyncState State => orchestrator.State;
         public IObservable<SyncState> StateObservable => orchestrator.StateObservable;
```
willsb: Maybe mark this guy as obsolete in the interface to prevent people from consuming it before we make the change suggested in #981

Toggl.Foundation.Tests/Sync/SyncManagerTests.cs
```diff
@@ -457,5 +459,120 @@ public void CompletesWhenSleepStateOccursAfterPushSync()
...
+                Queue.Dequeue().Returns(queuedState);
+
+                OrchestratorSyncComplete.OnNext(Pull);
+
+                Orchestrator.Received().Start(queuedState);
+                emitted.Should().BeNull();
+            }
+
+            [Fact]
+            public void EmitsSyncedWhenSyncCompletsAndQueueIsEmpty()
```
amulware: `CompletEs`

Toggl.Foundation/Sync/SyncManager.cs
```diff
@@ -27,7 +32,10 @@ public SyncManager(ISyncStateQueue queue, IStateMachineOrchestrator orchestrator
             this.queue = queue;
             this.orchestrator = orchestrator;
 
-            orchestrator.SyncCompleteObservable.Subscribe(syncOperationCompleted);
```
amulware: Should this maybe be a BehaviourSubject, so that subscribers can get the 'current' state immediately.

Toggl.Foundation/Sync/SyncManager.cs
```diff
@@ -27,7 +32,10 @@ public SyncManager(ISyncStateQueue queue, IStateMachineOrchestrator orchestrator
...
+            progress = new Subject<SyncProgress>();
+            ProgressObservable = progress.AsObservable();
+
+            orchestrator.SyncCompleteObservable.Subscribe(syncOperationCompleted, syncOperationFailed);
```
amulware: I am under the impression that observables cannot continue emitting values after they report an error. If that is so, it seems that if this observable ever errors, we can no longer use it afterwards, and sync will be broken.

Toggl.Foundation/Sync/SyncManager.cs
```diff
@@ -70,24 +78,59 @@ private void syncOperationCompleted(SyncState operation)
...
+                switch (error)
+                {
+                    case OfflineException offline:
+                        progress.OnNext(SyncProgress.OfflineModeDetected);
+                        break;
+
+                    case ClientDeprecatedException clientDeprecated:
+                    case ApiDeprecatedException apiDeprecated:
+                    case UnauthorizedException unauthorized:
+                        progress.OnNext(SyncProgress.ClientErrorAborted);
```
amulware: One of these three errors has to be handled differently, and maybe the other two in the future as well. If we do not expose them differently, I'm not sure how we could treat them appropriately in the UI.

Toggl.Foundation/Sync/SyncManager.cs
```diff
@@ -70,24 +78,59 @@ private void syncOperationCompleted(SyncState operation)
...
+                        break;
+
+                    default:
+                        progress.OnNext(SyncProgress.Failed);
+                        break;
+                }
```
amulware: In general, why should this switch be here, instead of in the view models, where we will definitely have to treat different errors differently already?

Toggl.Foundation/Sync/SyncManager.cs
```diff
@@ -70,24 +78,59 @@ private void syncOperationCompleted(SyncState operation)
...
         {
-            if (IsRunningSync) return;
+            if (IsRunningSync) return false;
 
             var state = isFrozen ? Sleep : queue.Dequeue();
             IsRunningSync = state != Sleep;
+
             orchestrator.Start(state);
+
+            return IsRunningSync;
```
amulware: Why would we return this instead of checking the value afterwards at the calling point when needed?

Toggl.Foundation/Sync/SyncProgress.cs
```diff
@@ -0,0 +1,11 @@
+﻿namespace Toggl.Foundation.Sync
+{
+    public enum SyncProgress
+    {
+        Syncing,
+        Synced,
+        OfflineModeDetected,
+        Failed,
+        ClientErrorAborted
```
amulware: This name really isn't clear to me. Maybe something like 'OutdatedClientError'? Though from the comments above, I'm not sure if the distinction here is needed at all since I might rather have the UI handle different errors.

Toggl.Ultrawave/ApiClients/BaseApi.cs
```diff
@@ -71,24 +72,31 @@ private IObservable<T> createObservable<T>(Endpoint endpoint, IEnumerable<HttpHe
             var request = new Request(body, endpoint.Url, headers, endpoint.Method);
             return Observable.Create<T>(async observer =>
             {
-                var response = await apiClient.Send(request).ConfigureAwait(false);
-                if (response.IsSuccess)
+                try
```
amulware: This try makes the already sub-optimal nesting even worse. Consider refactoring.

Toggl.Foundation/Sync/SyncManager.cs
simonrozsival: You mean we should have an observable for caught exceptions? Or that here we should use the `OnError` because sync won't continue afterwards anyway? (and therefore calling `Freeze` as well)

Toggl.Foundation/Sync/SyncManager.cs
```diff
@@ -74,23 +83,53 @@ private void syncOperationCompleted(SyncResult result)
...
+            {
+                progress.OnNext(SyncProgress.Failed);
+            }
+
+            if (error is ClientDeprecatedException
+                || error is ApiDeprecatedException
+                || error is UnauthorizedException)
+            {
+                progress.OnError(error);
+                Freeze();
```
amulware: What is the reasoning for reporting the error before freezing instead of afterwards?

Toggl.sln
```diff
@@ -515,4 +515,7 @@ Global
 		{B3FDFBF3-8D73-478F-B10E-9843C8613C87} = {B55DCFC1-9EA2-4289-B3EE-8979E0A9AB2E}
 		{DEAB771F-ED46-4A27-AD25-F26723820F93} = {128F799A-3060-4D72-9F14-93620F25E5E3}
 	EndGlobalSection
+	GlobalSection(ExtensibilityGlobals) = postSolution
+		SolutionGuid = {2DF8FE0D-778C-477F-992A-F2D2264D20E0}
+	EndGlobalSection
```
amulware: What's this?

Toggl.Foundation.Tests/Sync/SyncManagerTests.cs
```diff
@@ -467,6 +470,156 @@ public void CompletesWhenSleepStateOccursAfterPushSync()
...
+
+                OrchestratorSyncComplete.OnNext(new Error(exception));
+
+                caughtException.Should().NotBeNull();
+                caughtException.Should().Be(exception);
+            }
+
+            [Theory]
+            [MemberData(nameof(clientErrorAbortedExceptions))]
+            public void FreezesTheSyncManagerhenAKnownClientErrorExceptionIsReported(ClientErrorException exception)
```
amulware: 🐔 Typo

Toggl.sln
simonrozsival: That must have gotten there by mistake, I will 🔥 it.

Toggl.Foundation/Sync/SyncManager.cs
simonrozsival: The UI must be notified with the exception so it can distinguish what to do - it must navigate the user to the correct screen.

amulware: But how does that dictate the order here?

simonrozsival: Oh, my bad, I didn't read it properly. That is something I didn't consider - it is done inside the critical section of the lock, so the order shouldn't matter - even if somebody tried to invoke some method of the `SyncManager` in the handler of the error, it would wait until this method finishes and the manager would already be frozen. Do you see a problem with the ordering?

simonrozsival: You cannot check from the outside.

PR #979, opened 2017-11-16
Orchestrator emits errors

simonrozsival:
Ref #661 

This PR stops the Orchestrator from hiding the errors from the observers.

Squash whenever


Toggl.Foundation/Sync/StateMachineOrchestrator.cs
```diff
@@ -65,6 +68,13 @@ private void completeCurrentSync()
             syncComplete.OnNext(State);
         }
 
+        private void reportError(Exception exception)
+        {
+            syncing = false;
+            syncComplete.OnError(exception);
+            goToSleep();
```
kornelijepetak: `goToSleep()` will call `ensureNotSyncing()` which should return without throwing because of prior `syncing = false`. But for some unknown reason, what happens if the subscriber of `syncComplete` (on `onError`) restarts something or something else happens in-between (:racehorse: :racehorse:) that changes `syncing` (if it can at all, correct me)?
In that case, the message in the exception would not match the scenario (no sync was started at that point, but an error occured).
So it'd be better if you just called `OnNext` and that's it.

Toggl.Foundation/Sync/StateMachineOrchestrator.cs
```diff
@@ -65,6 +68,13 @@ private void completeCurrentSync()
             syncComplete.OnNext(State);
         }
 
+        private void reportError(Exception exception)
+        {
+            syncing = false;
+            syncComplete.OnError(exception);
```
amulware: I might be mistaken something, but if we report an error to an observable, we won't be able to report anything else in the future, right? How is that going to work? Or am I missing something?

simonrozsival: You are right, I checked the documentation and OnError signalizes the end of sequence. I will change this on Monday.

Toggl.Foundation.Tests/Sync/StateMachineOrchestratorTests.cs
```diff
@@ -208,14 +218,14 @@ public void ShouldCauseExpectedCompletedEventWhenSyncingCompletes()
             }
 
             [Fact]
-            public void ShouldCauseExpectedCompletedEventWhenSyncingFails()
+            public void ShouldCauseExpectedReportErrorWhenSyncingFails()
```
amulware: Something about the new name doesn't make sense to me. 'ReportError'?

Toggl.Foundation/Sync/StateMachineOrchestrator.cs
```diff
@@ -55,14 +55,24 @@ public void Freeze()
...
             syncing = false;
-            syncComplete.OnNext(State);
+            syncComplete.OnNext(new SyncResultSuccessful(State));
+        }
+
+        private void reportError(Exception error)
+        {
+            syncing = false;
+            syncComplete.OnNext(new SyncResultError(error));
+            stateEntered.OnNext(SyncState.Sleep);
```
amulware: This seems to be the only place where this class decides itself to enter a new state. Will that not cause problems?

amulware: In either case, should this not be tested?

Toggl.Foundation/Sync/SyncResult.cs
```diff
@@ -0,0 +1,30 @@
...
+    {
+        public SyncState Operation { get; }
+
+        public SyncResultSuccessful(SyncState operation)
+        {
+            Operation = operation;
+        }
+    }
+
+    public sealed class SyncResultError : SyncResult
```
amulware: One of these is called `Successful` the other `Error`: one adjective, and one noun. If possible, please unify. Also, if possible, can we maybe skip the `SyncResult` part? It would read nicer when checking for these types.

Toggl.Foundation/Sync/StateMachineOrchestrator.cs
simonrozsival: I will think of way how to put this logic in the `SyncManager`.

amulware: Unless you think having it here makes more sense. I'm open to an argument. :D

Toggl.Foundation/Sync/SyncManager.cs
```diff
@@ -65,12 +65,17 @@ public IObservable<SyncState> Freeze()
...
             lock (stateLock)
             {
                 IsRunningSync = false;
-                startSyncIfNeeded();
+
+                if (result is Success)
+                    startSyncIfNeeded();
+
+                if (result is Error)
+                    orchestrator.Start(Sleep);
```
amulware: It's not necessarily an issue, but I want to observe that we are not clearing the queue or preventing further syncing at this point. I believe those will come in future PRs, but in general, we should pay good attention to handle this correctly.

Toggl.Foundation/Sync/SyncManager.cs
```diff
@@ -65,12 +65,17 @@ public IObservable<SyncState> Freeze()
...
         {
             lock (stateLock)
             {
                 IsRunningSync = false;
-                startSyncIfNeeded();
+
+                if (result is Success)
+                    startSyncIfNeeded();
+
+                if (result is Error)
```
amulware: With these two type checks, there is the possibility of someone adding a third type which will cause this method to set the bool above but not do anything else. Could this cause invalid state? Is it maybe worth throwing here in case the result is of an unknown type?

Toggl.Foundation/Sync/SyncManager.cs
simonrozsival: Well this is exactly the point where we will clear the queue in #975, I'm not sure if we are on the same page here.

I agree about adding a test for not preventing further sync.

Toggl.Foundation/Sync/SyncManager.cs
simonrozsival: Throwing sounds good. This error handling is a stub at this moment, more logic will come in the other PRs which are on hold untill this gets merged.

Toggl.Foundation/Sync/SyncManager.cs
amulware: I think we are on the same page. I just wanted to make sure we are. :) If you want to add a test, feel free.

simonrozsival: OK.

PR #980, opened 2017-11-16
Do not catch certain errors in sync states

simonrozsival:
Ref #661, ref #306 

~This is on hold, until #978 is merged - it introduces `OfflineException`, which should be also rethrown.~

~Also PR #995 should be merged before this one.~


Toggl.Foundation.Tests/Sync/States/Push/BaseStates/BasePushEntityStateTests.cs
```diff
@@ -92,6 +94,34 @@ public void ReturnsFailTransitionWhenDatabaseOperationFails()
...
+            try
+            {
+                state.Start(Substitute.For<TModel>());
+            }
+            catch (Exception e)
+            {
+                caughtException = e;
+            }
+
+            caughtException.Should().ShouldBeEquivalentTo(exception);
```
amulware: Why not use `.ShouldThrow<>`?

Toggl.Foundation.Tests/Sync/States/Push/BaseStates/BaseCreateEntityStateTests.cs
```diff
@@ -58,13 +58,10 @@ public static object[] ClientExceptions()
...
                 new object[] { new BadRequestException(request, response) },
-                new object[] { new UnauthorizedException(request, response) },
                 new object[] { new PaymentRequiredException(request, response) },
                 new object[] { new ForbiddenException(request, response) },
                 new object[] { new NotFoundException(request, response) },
-                new object[] { new ApiDeprecatedException(request, response) },
                 new object[] { new RequestEntityTooLargeException(request, response) },
-                new object[] { new ClientDeprecatedException(request, response) },
                 new object[] { new TooManyRequestsException(request, response) }
             };
```
amulware: We seem to have this same list in three different places. Maybe this can be refactored into a single list somewhere to make sure we change it correctly in the future? (Similar with other similar lists if they exist.)

If so, then as a separate PR though. :)

Toggl.Foundation/Sync/States/Push/BaseStates/BasePushEntityState.cs
```diff
@@ -27,12 +27,19 @@ public BasePushEntityState(ITogglApi api, IRepository<TModel> repository)
...
-                        ? Observable.Return(ClientError.Transition((exception, entity)))
-                        : Observable.Return(UnknownError.Transition((exception, entity)));
+            => exception => shouldRethrow(exception)
+                ? Observable.Throw<ITransition>(exception)
+                : Observable.Return(fail(entity, exception));
+
+        private bool shouldRethrow(Exception e)
+            => e is ApiDeprecatedException || e is ClientDeprecatedException || e is UnauthorizedException;
+
+        private ITransition fail(TModel entity, Exception e)
```
amulware: Since this returns a transition, maybe rename to `failTransition`? I had to look twice before I got how this worked as is.

Toggl.Foundation.Tests/Sync/States/Push/BaseStates/BasePushEntityStateTests.cs
simonrozsival: That didn't work for me - I want to test that the type of `caughtException` is the same as `exception`, but I cannot use the type of `exception` in `.ShouldThrow<>`. Or if there is some way which I'm not aware of, please point me in the right direction :)

amulware: Nevermind on this, I didn't realise you needed the type statically for that to work.

PR #982, opened 2017-11-16
Fix tag displaying

### LINKED ISSUE #666: Fix tag displaying
Currently tag displaying in the start time entry view is broken if the tags span multiple lines.

screen shot 2017-10-03 at 11 33 55

### LINKED ISSUE #890: Selecting text not working properly
Both the start view text field and the edit view description text field (maybe others as well?) behave strangely with selections. Sometimes they just deselect or don't allow me to select the letters I want, instead selecting whole words or the entire text. It seems very finicky and I'm not sure what the pattern is, but it doesn't seem right.

willsb:
:warning: Even though this is a WIP, please do a functional test (preferably on Physical devices) so you can point out possible flaws and the like :warning: 

:dragon: :dragon: BEAST SLAYER :dragon: :dragon: 

Closes #666 
Closes #890  

// TODO
- [ ] Fix emoji tags
- [ ] Fix scroll offset size
- [ ] Fix text selection behaviour for ranged selections
- [ ] General refactor for more maintainable code
- [ ] If there is no text, only two lines of tags, the tag borders overlap.
- [ ] Looks like the the second, third etc. rows have tag positioned closer to the bottom of the border
- [ ] 3D touch messes things up

### DIFF
--- a/Toggl.Daneel/Binding/TextViewTextInfoTargetBinding.cs
+++ b/Toggl.Daneel/Binding/TextViewTextInfoTargetBinding.cs
@@ -6,6 +6,7 @@ using MvvmCross.Platform.Core;
 using Toggl.Daneel.Extensions;
 using Toggl.Daneel.Views;
 using Toggl.Foundation.Autocomplete;
+using Toggl.Multivac.Extensions;
 using UIKit;
 
 namespace Toggl.Daneel.Binding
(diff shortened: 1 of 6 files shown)


arturs105:
Did a functional test on an iPhone:
- Wrapping and scrolling work fine 👌 
- When using 3D touch to move the cursor over a tag, some [glitchy behaviour happens](https://youtu.be/qxtgaJ6sJ60)
- Looks like the the second, third etc. rows have tag positioned closer to the bottom of the border
- If there is no text, only two lines of tags, the tag borders overlap. Adding some text or project fixes it
- If i add tags, then a project and then try to type some text between the project and tags, stuff breaks - the first tag is displayed repeatedly as text in front of the project for each 2 typed characters. The tag with borders still stays after the project

Full code review coming a bit later.

willsb:
The beast won this battle, but the war is far from over!

image

PR #993, opened 2017-11-20
Fix incorrect condition for selecting suggestions vs. creating new project/tag

### LINKED ISSUE #972: Start time entry project suggestions do not work as expected
Whenever the `+ Create project "..."` is visible, selecting any suggested existing project navigates the user to the Create project view _when there is only one workspace_. It works well when there are multiple workspaces separated by headers (bug is located at `CreateSuggestionGroupedTableViewSource.cs` on line 64).

simonrozsival:
Closes #972 

Whenever "create" suggestion is shown, the "coordinates" will be section: 0, row: 0. Checking just the section doesn't work when the grouping is not used (section 0 contains also other suggestions than the "create" one, but the row wasn't checked).

Squash whenever

_Sidenote: I find it confusing when the workspace isn't shown at all. I think the condition for not using workspaces should not be "all the suggested projects are from one workspace", but "the user has access to only one workspace". It is not a part of this issue and maybe @pe0ny could consider changing the behavior in this way._


amulware:
Please redo to be merged from/into `beta-2`

PR #996, opened 2017-11-20
Refactor edit duration view

### LINKED ISSUE #810: Refactor edit duration view
This is a follow-up of #747 

The edit duration view uses start and stop times internally, although now it is not necessary any more and the edit view could be simplified to use the start time and the duration itself.

simonrozsival:
Closes #810 

Do a functional test during reviewing (both start and edit time entry views).

Squash whenever


Toggl.Foundation.MvvmCross/ViewModels/EditDurationViewModel.cs
```diff
@@ -46,37 +50,32 @@ public override void Prepare(DurationParameter parameter)
...
 
         private Task save()
         {
-            DateTimeOffset? stopTimeToBeReturned;
-            if (isRunning)
-                stopTimeToBeReturned = null;
-            else
-                stopTimeToBeReturned = StopTime;
-            return navigationService.Close(this, DurationParameter.WithStartAndStop(StartTime, stopTimeToBeReturned));
+            var stopTimeToBeReturned = isRunning ? (TimeSpan?)null : Duration;
```
kornelijepetak: You have switched from using `Stop`s to using `Duration`s but the naming is now awry. Is it not?

Toggl.sln
```diff
@@ -1,6 +1,8 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 2012
+# Visual Studio 15
```
willsb: Someone got a shiny new VS, but I don't think this should be included **here**

simonrozsival:
I updated the solution file, it is not included any more.

Issue #960, opened 2017-11-14
Trim TE description before saving changes


simonrozsival:
Currently we allow the user to insert several white spaces and take it as a valid description. Webapp trims the description, we should as well.

- [x] Start time entry view
- [x] Edit time entry view
- [x] Project name
- [x] Client name
- [x] Tag name

amulware:
The same should happen for projects, clients and tags if it doesn't already. (And in those cases a whitespace-only entity should not be allowed to be created at all.)

simonrozsival:
> also important to check for existing names after trimming, not before

simonrozsival:
All the PRs for all the entity types were merged, I will close this now manually.

PR #999, opened 2017-11-20
Trim project and client names before creating them

simonrozsival:
Ref #960 

Squash whenever


Toggl.Foundation.MvvmCross/ViewModels/EditProjectViewModel.cs
```diff
@@ -35,6 +35,8 @@ public sealed class EditProjectViewModel : MvxViewModel<string, long?>
 
         public string Name { get; set; } = "";
 
+        public string TrimmedName => Name.Trim();
```
kornelijepetak: Since this is called only once, is this really required?

kornelijepetak: Future-proof solution?

willsb: I have the same question

PR #953, opened 2017-11-14
Tag creation from edit view

arturs105:
Ref #820

PR #1000, opened 2017-11-20
Feature/tag empty state

### LINKED ISSUE #821: Tags view: empty state
- [ ] Show this when user doesn't have any tag
- [ ] Show placeholder: 'Add a tag'
- [ ] When user starts typing, show '+ Create tag "new-tag-name"', and keep the empty state illustration + copy

toggl - ios - tags empty state 1x

kornelijepetak:
Closes #821 (1st and 2nd point).

3rd point will be closed by #953.

I had to add another collection of tags because I couldn't compare to `Tags` as it is filtered by the autocompletion feature. Basically what would happen is that when you typed in some characters that yield no tag suggestions, the empty state would show as if the user has no tags, which is obviously not correct.

Additionally, when #953 is merged, reaction to tag adding (without leaving the tags view) must be implemented to trigger the `IsEmpty` again.

Please do a functional test.
 
Also if something else should be tested, please note. I wasn't sure about that.


Toggl.Foundation.MvvmCross/ViewModels/SelectTagsViewModel.cs
```diff
@@ -20,6 +22,7 @@ public sealed class SelectTagsViewModel : MvxViewModel<(long[] tagIds, long work
         private readonly IMvxNavigationService navigationService;
         private readonly ITogglDataSource dataSource;
         private readonly Subject<string> textSubject = new Subject<string>();
+        private readonly ReplaySubject<IEnumerable<IDatabaseTag>> allTagsSubject = new ReplaySubject<IEnumerable<IDatabaseTag>>(1);
```
simonrozsival: I think you should use rather `BehaviorSubject` here - it works the same way, but you don't have to have the magical constant `1` in there.

Toggl.Foundation.MvvmCross/ViewModels/SelectTagsViewModel.cs
```diff
@@ -56,10 +64,18 @@ public override void Prepare((long[] tagIds, long workspaceId) parameter)
             selectedTagIds.AddRange(parameter.tagIds);
         }
 
+        private List<IDatabaseTag> allTags = new List<IDatabaseTag>();
```
simonrozsival: This should be read-only.

Toggl.Foundation.MvvmCross/ViewModels/SelectTagsViewModel.cs
simonrozsival: Also I don't understand how you use this subject? I can't see any `OnNext` in this class. This means that on line 73 it is completely useless and you get only the data from the merged observable. Or am I missing something?

Toggl.Foundation.Tests/MvvmCross/ViewModels/SelectTagsViewModelTests.cs
```diff
@@ -156,6 +157,54 @@ await autocompleteProvider.Received()
             }
         }
 
+        public sealed class TheIsEmptyProperty : SelectTagsViewModelTest
```
simonrozsival: You should add test for the case when there are no tags, but then user adds some and then it shouldn't be empty any more.

Toggl.Foundation.MvvmCross/ViewModels/SelectTagsViewModel.cs
simonrozsival: Is this the thing you mention that should be done after #953 is merged? If that is the case, then please add the `on-hold` and `wip` labels to this PR and I'll get back to it when you finish it.

kornelijepetak: Yes, you have guessed correctly. I just forgot to add the labels. This PR does not support adding tags yet (that was done in #953)

Toggl.Foundation.Tests/MvvmCross/ViewModels/SelectTagsViewModelTests.cs
kornelijepetak: As soon as #953 is merged. :)

kornelijepetak:
I have also added two tests. One tests that the real tags are observed and not the filtered collection, and the other proves that `IsEmpty` will change accordingly when a new tag is created.

Toggl.Daneel/Binding/TextFieldPlaceholderTargetBinding.cs
```diff
@@ -0,0 +1,23 @@
+﻿
+
```
simonrozsival: Empty lines

Toggl.Daneel/Binding/TextFieldPlaceholderTargetBinding.cs
```diff
@@ -0,0 +1,23 @@
...
+    public TextFieldPlaceholderTargetBinding(UITextField target)
+        : base(target)
+    {
+    }
+
+    protected override void SetValue(string value)
+    {
+        Target.Placeholder = value;
+    }
+}
```
simonrozsival: Missing new line

simonrozsival:
Also please make the same changes you did to `IsEmpty` in #1001 for the sake of consistency.

PR #1001, opened 2017-11-21
Feature/project empty state

### LINKED ISSUE #811: Projects view: empty state
- [ ] Show this when user doesn't have any project
- [ ] Show placeholder: 'Add a project'
- [ ] When user starts typing, show '+ Create project "new-project-name"', and keep the empty state illustration + copy

toggl - ios - projects empty state 1x

kornelijepetak:
Closes #811 .

The image/copy is now shown only when user has no projects whatsoever.
Since upon project creation, the user is returned to the TE edit view with the new project already selected, the selection screen is not shown again. That's why it's enough to only test for existing projects in `Initialize`.

The tests try to prove 
- that IsEmpty is correctly set depending on whether user has projects or not
- that no matter what is displayed as a result of filtering, the IsEmpty will take only the real project existence into account

Do a functional, and also say if something else should be tested for too.


Toggl.Daneel/Binding/TextFieldPlaceholderTargetBinding.cs
```diff
@@ -0,0 +1,23 @@
+﻿
```
willsb: Empty lines

Toggl.Foundation.MvvmCross/ViewModels/SelectProjectViewModel.cs
```diff
@@ -85,6 +95,11 @@ public override void Prepare(SelectProjectParameter parameter)
         {
             await base.Initialize();
 
+            dataSource.Projects
```
willsb: This should happen in the Initialize method, which is async and allows for you to `await` this guy

Toggl.Foundation.MvvmCross/ViewModels/SelectProjectViewModel.cs
```diff
@@ -46,6 +49,13 @@ public bool SuggestCreation
             }
         }
 
+        public bool IsEmpty => !(hasProjects ?? true);
```
willsb: `hasProjects.HasValue ? hasProjects.Value : true;`

willsb: Moreover: Why make hasProjects nullable? One either has projects or doesn't 🤔

Toggl.Foundation.MvvmCross/ViewModels/SelectProjectViewModel.cs
```diff
@@ -112,11 +134,13 @@ private void onSuggestions(IEnumerable<ProjectSuggestion> suggestions)
 
         private async Task createProject()
         {
-            if (!SuggestCreation) return;
-            
+            if (!SuggestCreation)
+                return;
```
willsb: Guidelines say this should be inline

Toggl.Foundation.MvvmCross/ViewModels/SelectProjectViewModel.cs
```diff
@@ -112,11 +134,13 @@ private void onSuggestions(IEnumerable<ProjectSuggestion> suggestions)
...
+
             var createdProjectId = await navigationService.Navigate<EditProjectViewModel, string, long?>(Text.Trim());
-            if (createdProjectId == null) return;
-            
+            if (createdProjectId == null)
```
willsb: Guidelines say this should be inline

Toggl.Foundation.MvvmCross/ViewModels/SelectProjectViewModel.cs
```diff
@@ -142,7 +166,8 @@ private Task close()
                 Resources.Cancel
             );
 
-            if (!shouldSetProject) return;
+            if (!shouldSetProject)
```
willsb: Guidelines say this should be inline

Toggl.Foundation.MvvmCross/ViewModels/SelectProjectViewModel.cs
```diff
@@ -175,10 +200,12 @@ private void setProject(AutocompleteSuggestion suggestion)
         private void toggleTaskSuggestions(ProjectSuggestion projectSuggestion)
         {
             var grouping = Suggestions.FirstOrDefault(s => s.WorkspaceName == projectSuggestion.WorkspaceName);
-            if (grouping == null) return;
+            if (grouping == null)
+                return;
 
             var suggestionIndex = grouping.IndexOf(projectSuggestion);
-            if (suggestionIndex < 0) return;
+            if (suggestionIndex < 0)
```
willsb: Guidelines say this should be inline

Toggl.Foundation.MvvmCross/ViewModels/SelectProjectViewModel.cs
```diff
@@ -196,7 +223,8 @@ private void toggleTaskSuggestions(ProjectSuggestion projectSuggestion)
         {
             foreach (var suggestion in suggestions)
             {
-                if (suggestion is TaskSuggestion) continue;
+                if (suggestion is TaskSuggestion)
```
willsb: Guidelines say this should be inline

Toggl.Foundation.MvvmCross/ViewModels/SelectProjectViewModel.cs
kornelijepetak: Regarding the first comment, did you require change because it looks convoluted to you? I have changed it none the less, just interested in what you find problematic.

Now to address the second question. It is true that one either has projects or doesn't, but there is a time when we still don't know whether one does. Ok, so it can be done with a regular `bool`, but then I have to initially set `hasProjects` to `true`. Why? Because until we for sure know that the user has or has not any projects, the image should not appear. The worst thing is that the image appears and then flickers away once we have retrieved some projects from the database.

So the question here is whether to:
- leave it as it is (which conveys the intention better)
- switch to bool and initialize it to `true`
- switch to bool, initialize it to `true` and write a comment explaining why do we initialize it to an unintuitive value
- switch to bool, but add another bool variable called `weHaveNoIdeaWhetherUserHasProjectsWeAreStillWaitingOnDatabaseToTellUs` and then combine them in `IsEmpty` correctly.

Whatcha want me to do?

Toggl.Foundation.MvvmCross/ViewModels/SelectProjectViewModel.cs
kornelijepetak: The way I see it, it is inside `Initialize`. I think you saw something wrong. :) 
Also, isn't a handler in `Subscribe` executed as async (making this call non-blocking)? With await, I would prolong the next subscription (of `infoSubject`), would I not?

Toggl.Foundation.MvvmCross/ViewModels/SelectProjectViewModel.cs
kornelijepetak: I have fixed all requested code style changes. VSMac's auto-formatter doesn't like our code style guidelines. :p

kornelijepetak:
I have added additional test that checks whether the initial value is correct, so please take that into consideration when rereviewing.

PR #1003, opened 2017-11-21
Get rid of useless functions in StateMachine

simonrozsival:
This is just a small refactoring, no issue is related to this PR.

Squash whenever.


simonrozsival:
I personally think that it is not very useful to have a method which wraps one line of code, which is called from only one place and which describes what it does quite well, it is not some complicated formula or magic happening there, so it would deserve extra attention. But after all, this is a very minor thing and we shouldn't spend too much time on it. Unless somebody agrees with me and approves this by the time I finish working today, I will close it and forget about it :)

amulware:
I think the levels of abstractions/concerns are mixed a bit this way which is not ideal, but it is more consistent in its inconsistency.

PR #1004, opened 2017-11-21
Add string length extensions

arturs105:
This adds `string.LengthInBytes()` and `string.LengthInGraphemes()` extension methods.


Toggl.Foundation.MvvmCross/ViewModels/EditTimeEntryViewModel.cs
```diff
@@ -42,7 +43,7 @@ public sealed class EditTimeEntryViewModel : MvxViewModel<long>
         public long Id { get; set; }
 
         private int descriptionByteCount
```
simonrozsival: You could maybe also 🔥 this, as it is not necessary, I think.

PR #1007, opened 2017-11-21
Fix suggestion checking

willsb:
This wasn’t taking into account the fact that one can click the the Project button to start project suggestion even after having selected a project.

Squash whenever


Toggl.Foundation.Tests/MvvmCross/ViewModels/StartTimeEntryViewModelTests.cs
```diff
@@ -184,6 +184,7 @@ public TheSuggestCreationProperty()
                 await ViewModel.Initialize();
                 ViewModel.TextFieldInfo = TextFieldInfo.Empty
                     .WithProjectInfo(WorkspaceId, ProjectId, ProjectName, ProjectColor);
+                ViewModel.ToggleProjectSuggestionsCommand.Execute();
```
simonrozsival: I don't think this test is correct. Shouldn't there be a similar test, where there won't be any change to the `TextFieldInfo` after this command is executed?

PR #1009, opened 2017-11-22
Access restriction storage for iOS

simonrozsival:
Ref #751 

Squash whenever


willsb:
Why not make this allow for arbitrary safe storage using an Indexer and keys in constants instead?

simonrozsival:
I'm not sure how exactly you mean that. I like the fact that user (Foundation) doesn't have access to any key through an indexer, but can modify and read only certain values.

Toggl.Daneel/Services/UserDataAccessRestrictionStorage.cs
```diff
@@ -0,0 +1,60 @@
...
+            => NSUserDefaults.StandardUserDefaults.SetBool(true, UnauthorizedAccess);
+
+        public void ClearRestrictions()
+        {
+            NSUserDefaults.StandardUserDefaults.SetString("0.0.0", OutdatedClient);
+            NSUserDefaults.StandardUserDefaults.SetString("0.0.0", OutdatedApi);
+            NSUserDefaults.StandardUserDefaults.SetBool(false, UnauthorizedAccess);
+        }
+
+        public bool IsAccessRestricted()
```
willsb: I feel this would be nicer as a property

Toggl.Daneel/Services/UserDataAccessRestrictionStorage.cs
```diff
@@ -0,0 +1,60 @@
...
+            => NSUserDefaults.StandardUserDefaults.SetString(version.ToString(), OutdatedApi);
+
+        public void SetUnauthorizedAccess()
+            => NSUserDefaults.StandardUserDefaults.SetBool(true, UnauthorizedAccess);
+
+        public void ClearRestrictions()
+        {
+            NSUserDefaults.StandardUserDefaults.SetString("0.0.0", OutdatedClient);
+            NSUserDefaults.StandardUserDefaults.SetString("0.0.0", OutdatedApi);
+            NSUserDefaults.StandardUserDefaults.SetBool(false, UnauthorizedAccess);
```
amulware: If all of these are cleared at once, how will we handle the case when we detect two errors? or can that never happen?

simonrozsival: Either we block the app because it is outdated - 2/3 possibilities are resolved by updating the app, the third could not happen (user might revoke the token while the app is blocked, but we cant detect that untill user updates the app) or when user revokes the token, waits really long to re-login, we detect outdated API. In this case the user must first update the app, only then can he re-login. If we clear all restrictions then the first HTTP request would restore the restriction and show the login screen. That seems error prone so I will split this into two clear methods.

simonrozsival: In fact we never have to clear the `OutdatedClient` or `OutdatedApi` - once it is updated to a higher version of the app, restrictions will disappear implicitly. I suppose it will be always updated to the newest version, so the user shouldn't see this screen more than once in a row.

PR #1013, opened 2017-11-23
Fix incorrect deletion of local data

### LINKED ISSUE #1011: Check sync of deleted time entries
We saw some crashing that we couldn't explain. The only anomaly we found so far is that the respective database had unsynced deleted time entries.

This needs more investigation, but as of right now I'm not convinced that deleted time entries are synced correctly.

It appears like we inject the data source instead of the database into the `DeleteLocalTimeEntryState`, which when deleting, it will only set the delete flag, not actually delete the entity from the database.

I may or may not have understood this, but someone should definitely take a look and make sure deletion syncs correctly.

Cases to cover:

- time entry with negative ID is deleted locally (flag is set): when the syncing finds this, delete it in the database for good (instead of doing any api calls)
- time entry with non-negative ID is deleted locally: do the delete call to backend, on success delete in database, on fail... I do not know, might depend on error

simonrozsival:
Closes #1011 

Squash whenever


simonrozsival:
I was able to reproduce the bug by:

- disabling wi-fi
- creating a TE
- stopping it
- deleting it
- connecting back to the network

App then crashes right after starting. This PR fixes it.

simonrozsival:
The cause of the crash was that the data source `Delete` - this doesn't really remove the entity, so next time an entity for syncing should be picked, it is the same one. This goes on infinitely, until the stack overflows and app crashes.

PR #1014, opened 2017-11-23
Add character counter to description in edit view

### LINKED ISSUE #891: Limit time entry description to 3000 characters

arturs105:
Closes #891


Toggl.Daneel/Views/TextViewWithPlaceholder.cs
```diff
@@ -14,8 +14,8 @@ public class TextViewWithPlaceholder : UITextView, IUITextViewDelegate
         private readonly UIColor defaultPlaceholderColor = Color.Common.PlaceholderText.ToNativeColor();
 
         private bool isFocused;
-        private UIStringAttributes defaultTextAttributes;
         private UIStringAttributes placeholderAttributes;
+        protected UIStringAttributes DefaultTextAttributes;
```
willsb: No protected fields 🚫 
Please make this a property

Toggl.Daneel/Views/TextViewWithCharacterCounter.cs
```diff
@@ -0,0 +1,128 @@
...
+
+        private void setAttributedText(string value)
+        {
+            if (string.IsNullOrEmpty(value))
+            {
+                AttributedText = new NSAttributedString("");
+                return;
+            }
+
+            var charactersRemaining = MaxLengthInBytes - CountedText.LengthInBytes();
```
willsb: I feel like all those checks could both test this logic and later reuse it on Android

arturs105: I'm not sure i understand, what you mean.

Issue #1015, opened 2017-11-23
Investigate failing time entry creation integration tests


amulware:
The following two tests are failing.

```
TimeEntriesApiTests+TheCreateMethod
TheTimeEntryStoredInBackendIsARunningTimeEntryWhenFetchingItAfterPostingANewRunningTimeEntry

BadRequestException : ApiException for request POST https://toggl.space/api/v9/workspaces/2410694/time_entries: Response: 
(Status: [400 BadRequest]) 
(Body: Max allowed duration per 1 time entry is 999 hours) 
(Message: The data is not valid or acceptable.)


TimeEntriesApiTests+TheCreateMethod
TheTimeEntryReturnedByBackendIsARunningTimeEntryWhenPostingANewRunningTimeEntry

BadRequestException : ApiException for request POST https://toggl.space/api/v9/workspaces/2410695/time_entries: Response: 
(Status: [400 BadRequest]) 
(Body: Max allowed duration per 1 time entry is 999 hours) 
(Message: The data is not valid or acceptable.)
 ```

I'd like to know why these started failing, and if the mistake is on our side, fix it. Otherwise, talk to backend and see what's what.

My intuition says that this either started failing because the tests use hardcoded dates of some sort that are now longer than 999 hours in the past, or that backend changed their duration check subtly.

Please request my review if there will be a PR here.

simonrozsival:
It is because of hardcoded dates. I will change the tests.

PR #1016, opened 2017-11-23
Reduce integration test parallelism

amulware:
I’ve been talking to backend about our secure channel failures, and their guess was that we are making so many requests in a short time that some load balancing or attack prevention system will kill some of them.

It is strange to me that usually only a few fail, and they fail in a row, and then the rest run usually fine. Maybe the HTTP implementation automatically throttles for us or something, I don’t know.

I’ve been trying several times since then to decrease the amount of parallelism that our tests run with, with the `parallelizeTestCollections` xunit setting being the appropriate one to use as far as I can tell from the documentation.

However, that setting appears to have absolutely no effect as far as I can tell. My best guess is that this is because of our tests being asynchronous, but I don't know for sure.

What has an effect, however, is the `maxParallelThreads` setting, which is changed here.

Note that I set this to `2`. A value of `1` causes the tests to freeze up at some point and not continue, while a value of `3` causes secure channel failures. At a value of `2`, the tests always complete, and **most of the time** do not error.

I'm not necessarily happy with my understanding of what is going on, or with how reliable this fix is, but empirically, this seems to at least improve our problem at minimal cost: Without these changes the tests run in 95 seconds, with them in 105 seconds on my machine.

Additionally, I've added a 1-second delay before every single test by means of the `LogTestInfoAttribute`, which in my tests prevents the last few rare secure channel failures. Note that only having the delay, without limiting threads, does not prevent the errors. I assume this is because we still kick off too many requests at the same time, we just have some delay between groups of too many requests.

Unfortunately, this change makes the tests run significantly longer, currently about 175 seconds on my machine.

I think this tradeoff is worth it for the time being, especially since we run them only very rarely.

Thoughts and opinions welcome.

PS: All my tests were run locally, so I cannot say if it will behave the exact same on bitrise, but given that the errors occurred very similarly, I am reasonably hopeful.

For the tests running on bitrise, see: https://www.bitrise.io/build/be087f3c21d3874d
The only failures there are handled in #1015
Of course feel free to rerun yourself either there or locally.


willsb:
Did you fine tune the one second delay or is it a magic number? 

Decreasing it until you reach a minimum value will speed up the tests a little

amulware:
Ah, my apologies for not elaborating. I tested 0.5, 1 and 2 seconds, the latter two working fine, and the first not seeming to make any difference - of course, I only ran a handful of times each.

I'm rather on the safe side and keep it a little bit higher than needed though.

amulware:
Should I add a comment to the code, or will this PR be documentation enough (git blame will point here) for future reference?

willsb:
A comment in code would be nice to have, pretty please

simonrozsival:
@amulware Maybe we could use a global lock to make sure there is always only one test running at a time? Then maybe we wouldn't need the 1s delay.

amulware:
@simonrozsival That's a possibility, but I'm not sure if it's a good idea... I'm afraid it might freak out xunit. And right now we still have a small degree of parallelism which speeds up the tests. That being said, I'll give it a try. But if it doesn't work (better) I say we go with this.

amulware:
Ok, I tried using an `AutoResetEvent` which as far as I can tell should do what @simonrozsival suggests trivially. (initialise as signalled, WaitOne before the test, Set after the test) but it seems to get stuck in the first test somewhere and I'm not sure why. It shows the starting console message, but not the end one (despite both being between the WaitOne and the Set, so I'm really not sure how that can happen - unless maybe it gets stuck using all threads of the threadpool to kick off tests, and has none left to do the async work inside the tests, but I don't know what to do if that is the case).

TL;DR: Can I get a ✅ ? :D

Toggl.Ultrawave.Tests.Integration/LogTestInfoAttribute.cs
```diff
@@ -2,6 +2,7 @@
 using System.Diagnostics;
 using System.Reflection;
 using System.Text.RegularExpressions;
+using System.Threading;
```
simonrozsival: 🔥

amulware: How so? I'm using System.Threading.Thread

amulware:
I think this is fine now. If you both agree, feel free to 🦑

Toggl.Ultrawave.Tests.Integration/LogTestInfoAttribute.cs
```diff
@@ -18,6 +18,9 @@ public override void Before(MethodInfo methodUnderTest)
             formattedTestName = getFormattedName(methodUnderTest.ReflectedType, methodName);
             Console.WriteLine(formattedTestName.Replace("`1", ""));
 
+            // This is to make integration tests run slightly slower to prevent SecureChannelFailure
+            // errors in most cases (compined with limiting the XUnit max max parallel threads)
```
willsb: `max max parallel threads`

Toggl.Ultrawave.Tests.Integration/LogTestInfoAttribute.cs
```diff
@@ -19,7 +19,8 @@ public override void Before(MethodInfo methodUnderTest)
             Console.WriteLine(formattedTestName.Replace("`1", ""));
 
             // This is to make integration tests run slightly slower to prevent SecureChannelFailure
-            // errors in most cases (compined with limiting the XUnit max max parallel threads)
+            // errors caused by too many HTTP calls in quick succession in most cases
+            // (compined with limiting the XUnit max parallel threads)
```
willsb: `compined`

Toggl.Ultrawave.Tests.Integration/LogTestInfoAttribute.cs
simonrozsival: My bad.

PR #1018, opened 2017-11-23
Signup password min length

### LINKED ISSUE #988: Sign up UI: add additional copy below the password field
When user signs up and starts entering his/her password, show this text below the input field:
'Password should be at least 6 characters.'

toggl - ios - sign up step2 1x

simonrozsival:
Closes #988 

I reused the `ErrorText` property of the `LoginViewModel`. I thought about renaming it to `HelpText` or something like that, but that would also transitively mean renaming the label in the controller and the `.xib` which would mean a looot of changes. If the reviewer thinks it is worth it, I can do it.

Squash whenever.


willsb:
Are we validating the password size in the LoginViewModel?

simonrozsival:
I don't think there is any local validation (except for the email validation). It would be worth creating an issue for that.

willsb:
Why not include this here and:

- Clear the error text based on password size
- Change the condition for validating the password

simonrozsival:
But what would be the behavior? Show "Password should be at least 6 characters." and when the password is changed and the length is >= 6 chars. we just change it to ""? Form validations normally show some kind of explicit visual feedback that it is now valid. This is up to @pe0ny to decide.

Issue #977, opened 2017-11-16
Optimize slow running tests


willsb:
Ideally we should aim for having all tests running in less than half second, but PRs can focus on each of those concern lists:

## CRITICAL - >= 5 seconds
- [ ] 30.540s - Toggl Foundation Tests MvvmCross ViewModels EditTimeEntryViewModelTests TheSelectTagsCommand (4 tests) 
- [ ] 8.745s - Toggl Foundation Tests Autocomplete AutocompleteProviderTests TheQueryMethod QueriesTheDatabaseForTimeEntries (10 tests) 
- [ ] 6.375s - Toggl Foundation Tests Sync States FetchAllSinceStateTests TheStartMethod (8 tests) 
- [ ] 6.326s - Toggl Foundation Tests Sync SyncManagerTests TheOrchestratorCompleteObservable (11 tests) 

## BAD - >= 2 seconds
- [ ] 5.704s - Toggl Foundation Tests Sync SyncManagerTests TheForceFullSyncMethod (10 tests) 
- [ ] 5.130s - Toggl Foundation Tests MvvmCross ViewModels SettingsViewModelTests TheLogoutCommand (15 tests) 
- [ ] 4.548s - Toggl Foundation Tests MvvmCross ViewModels TimeEntriesLogViewModelTests TheTimeEntriesProperty (3 tests) 
- [ ] 4.423s - Toggl Foundation Tests Autocomplete AutocompleteProviderTests TheQueryMethod QueriesTheDatabaseForProjects (6 tests) 
- [ ] 4.153s - Toggl Foundation Tests Sync SyncManagerTests ThePushSyncMethod (10 tests) 
- [ ] 3.476s - Toggl Foundation Tests MvvmCross ViewModels SuggestionsViewModelTests TheStartTimeEntryCommand (3 tests) 
- [ ] 3.134s - Toggl Foundation Tests MvvmCross ViewModels TimeEntriesLogViewModelTests TheContinueTimeEntryCommand (4 tests) 
- [ ] 3.005s - Toggl Foundation Tests Sync ConflictResolution TimeEntryRivalsResolverTests (8 tests) 
- [ ] 2.859s - Toggl Foundation Tests MvvmCross ViewModels SettingsViewModelTests TheFlags (6 tests) 
- [ ] 2.772s - Toggl Foundation Tests MvvmCross ViewModels EditTimeEntryViewModelTests TheInitializeMethod (4 tests) 
[...]

PR #1022, opened 2017-11-23
Log slow tests during runs

willsb:
Ref #977 

This outputs slow tests on the console. Will help us become more aware of what’s going on.

Also important to note that Properties ignore Before/After attributes. Once they support that we can add this attribute to properties as well


amulware:
You may want to make this build first ;)

PR #1035, opened 2017-11-27
Add first reports client for projects summary

### LINKED ISSUE #970: Create first reports api endpoint and api client
Relevant information:

`POST reports/v3/workspaces/{workspace_id}/projects/summary`

```
// Input:
// {
// 	-- required --
// 	"start_date": "2006-01-02",
//
// 	-- optional(can be null) --
// 	"end_date": "2006-01-02", // Default is 7 days from start_date
// }
//
// Output:
// [array of {
// 	"user_id": int,
// 	"project_id": int || null,
// 	"tracked_seconds": int,
// 	"billed_seconds": int, // Only for non-free workspaces
// }]
//
// Output headers:
// 	X-Range-Start = Actual start date used for query
// 	X-Range-End = Actual end date used for query
```

- The interval can be at most one year, we should enforce this on our side.
- Internally, the start and end date should be strongly typed.
- All integers in the output are 64bit.
- We can probably ignore the output headers for now (or ever).

The endpoint is live and should be working fine. Let's do some testing so we understand exactly how it works, and how it errors under different conditions and encapsulate that in integration tests.

simonrozsival:
Closes #970 

~WIP: the `ReturnsTheSumOfBillableSecondsForTheGivenIntervalWhenTheWorkspaceIsPaid` integration tests does not work right now. I will have to investigate how to change the test to work as expected.~ ✅ 

~Also when #1039 is merged (or some similar PR), we can use `AuthenticatedEndpointBaseTests` for reports.~

Integration tests (commit `c2eec25`): https://www.bitrise.io/build/74d1f1719529fc29

Squash whenever


Toggl.Multivac/ReportsModels/IProjectSummary.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace Toggl.Multivac.ReportsModels
```
willsb: I'd prefer `Toggl.Multivac.Models.Reports`, what do you think?

Toggl.Ultrawave/ReportsApiClients/ProjectsSummaryApi.cs
```diff
@@ -0,0 +1,62 @@
...
+                var summary = new ProjectsSummary { StartDate = startDate, EndDate = endDate, ProjectsSummaries = projectsSummaries };
+
+                observer.OnNext(summary);
+                observer.OnCompleted();
+
+                return () => { };
+            });
+        }
+
+        private sealed class ProjectsSummaryParameters
```
willsb: Do `[Preserve(AllMembers = true)]` here

simonrozsival: I use the constructor on line 32, I thought I don't need the preservation attribute in such case. Are you sure it is needed?

Issue #1005, opened 2017-11-21
Outdated App UI


amulware:
Design: https://zpl.io/VxYdoRa

<img width="350" alt="screen shot 2017-11-21 at 16 50 02" src="https://user-images.githubusercontent.com/3987892/33082026-0fc4a6be-cedc-11e7-8dcd-bbe3b41f48cb.png">

- [ ] show this overlay when we receive 410 or 418 codes from the api
- [ ] if the user restarts the app, this overlay still has to be visible, and no sync may be triggered
- [ ] if the user restarts the app after updating it, the overlay will no longer be shown, and sync is triggered

amulware:
@pe0ny please confirm:

- the update button leads to the app store page of the app?
- the link leads to http://toggl.com ?

pe0ny:
@amulware yes 👍

PR #1039, opened 2017-11-27
Detect unauthorized access when API returns 403 Forbidden

simonrozsival:
Ref #1005 

Squash whenever


Toggl.Ultrawave/ApiClients/BaseApi.cs
```diff
@@ -18,6 +18,8 @@ internal abstract class BaseApi
         private readonly IApiClient apiClient;
         private readonly IJsonSerializer serializer;
 
+        private readonly string loggedUri = "me/logged";
```
amulware: Why is this and the changes to the Endpoint type necessary? Can we not just add this as an actual Endpoint instance?

Toggl.Ultrawave/ApiClients/BaseApi.cs
```diff
@@ -105,5 +107,33 @@ private IObservable<T> createObservable<T>(Endpoint endpoint, IEnumerable<HttpHe
...
             });
         }
+
+        private async Task<Exception> getExceptionFor(
+            IRequest request,
+            IResponse response,
+            Endpoint endpoint,
+            IEnumerable<HttpHeader> headers)
+        {
+            Exception exception = ApiExceptions.For(request, response);
```
amulware: I don't really like that this is called no matter what happens below. I rather have it only be called when needed.

Toggl.Ultrawave/ApiClients/BaseApi.cs
```diff
@@ -105,5 +107,33 @@ private IObservable<T> createObservable<T>(Endpoint endpoint, IEnumerable<HttpHe
...
+            }
+
+            return exception;
+        }
+
+        private async Task<bool> isLoggedIn(Uri baseUrl, IEnumerable<HttpHeader> headers)
+        {
+            var request = new Request(String.Empty, new Uri(baseUrl, loggedUri), headers, HttpMethod.Get);
+            var response = await apiClient.Send(request).ConfigureAwait(false);
+            return response.IsSuccess;
```
amulware: This can fail for other reasons than not being logged out, right? In that case, we tell users that they reset the API token, maybe even if they didn't. Is there no more specific check we can make here?

Toggl.Ultrawave/ApiClients/BaseApi.cs
amulware: `??` at the return statement would be fine for me.

Toggl.Ultrawave/ApiClients/BaseApi.cs
simonrozsival: Margus wrote:
> fire a 2nd call on 403 - v9/me/logged, if that one gives 4xx - log the user out.

Therefore you are right, there should be a more complex statement.

Toggl.Ultrawave/TogglApi.cs
```diff
@@ -23,15 +23,15 @@ public TogglApi(ApiConfiguration configuration, HttpClientHandler handler = null
             var apiClient = new ApiClient(httpClient, userAgent);
             var endpoints = new Endpoints(configuration.Environment);
 
-            Tags = new TagsApi(endpoints.Tags, apiClient, serializer, credentials);
-            User = new UserApi(endpoints.User, apiClient, serializer, credentials);
-            Tasks = new TasksApi(endpoints.Tasks, apiClient, serializer, credentials);
+            Tags = new TagsApi(endpoints.Tags, apiClient, serializer, credentials, endpoints.LoggedIn);
```
amulware: At this point, why not simply give it the whole `endpoints` object?

simonrozsival:
I'm not sure which tests to add to further test this. It is in fact tested by the `AuthenticatedEndpointBaseTests` and then in some specific integration tests, which expect `ForbiddenException` to be thrown (e.g., `WorkspacesApiTests`.`FailsForWrongWorkspaceId`).

PR #1048, opened 2017-11-28
Handle 429 HTTP errors during push sync

### LINKED ISSUE #1037: Handle HTTP 429 responses
We don't implement actual rate limiting right now. While we could do that, I rather am optimistic and let the API tell us when we are pushing too much.

In that case (we get a 429) we should either simply back off for a bit, or enter the retry loop. I think using the retry loop sort of makes sense? Or will the status endpoint also return 429?

simonrozsival:
Closes #1037 

This PR adds a new state, which is the very basic implementation of the "bottom left part" of the internal push sync diagram. It cannot resolve any error, so it handles everything as unsyncable - except for the case whent he `TooManyRequestsException` is handled, in which case the retry loop should be entered.

Squash whenever


Toggl.Foundation.Tests/Sync/States/Push/Projects/TryResolveClientErrorForProjectStateTestscs.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System;
+using Toggl.Foundation.Models;
+using Toggl.Foundation.Sync.States.Push;
+using Toggl.PrimeRadiant.Models;
+
+namespace Toggl.Foundation.Tests.Sync.States.Push
+{
+    public sealed class TryResolveClientErrorForProjectStateTests : BaseTryResolveClientErrorStateTests<IDatabaseProject>
```
amulware: What's up with the filename? `*Testscs.cs`?

Toggl.Foundation/Sync/States/Push/Clients/TryResolveClientErrorForClientState.cs
```diff
@@ -0,0 +1,9 @@
+﻿using Toggl.PrimeRadiant.Models;
+
+namespace Toggl.Foundation.Sync.States.Push
+{
+    public sealed class TryResolveClientErrorForClientState
```
amulware: Why even have these implementations if they don't do anything instead of just using the generic class?

Issue #990, opened 2017-11-20
Log view: add padding at the bottom


pe0ny:
Currently when user scrolls at the bottom of the log list, the last time entry is hidden behind the running time entry (or play button). To avoid that happening, padding between the last TE and bottom of the screen should be 64pt.

Zeplin: https://zpl.io/a309jlQ

Current:
![img_4593](https://user-images.githubusercontent.com/5064445/33007800-04757922-cdd9-11e7-9329-19c345efcaa5.png)

Better solution:
![screen shot 2017-11-20 at 09 51 48](https://user-images.githubusercontent.com/5064445/33007753-d74aa9c2-cdd8-11e7-8d8f-8e5f97fea8b9.png)

pe0ny:
@kornelijepetak Is it related that currently there is some padding always visible at the bottom though it shouldn't be?

<img width="407" alt="screen shot 2017-11-30 at 09 04 41" src="https://user-images.githubusercontent.com/5064445/33418220-874c1668-d5ad-11e7-929d-d0e8d496a979.png">

kornelijepetak:
I haven't push anything to `develop` related to this and I think @willsb hasn't either. It might have creeped in with another push. Maybe we can rush @willsb to finish his WIP so we can see if the padding is still there afterwards. What do you think?

arturs105:
That padding at the bottom is actually a sticky footer (the gray area between the sections). That is fixed in the iPhone X PR for all devices.

willsb:
This will be collaterally fixed by the PR to fix #322

PR #1051, opened 2017-11-29
Adapt the app for iPhone X

### LINKED ISSUE #968: Adapt the app for iPhone X
- Notice, how, when scrolling, some stuff appears behind the status bar.
- The separators between sections (days in the log) look like sticky footers at the bottom, though this can't be seen in the gif, so see for yourself in the simulator.
- Probably some other stuff as well.

nov-15-2017 13-58-42

arturs105:
Closes #968 

Ref #990 (Removes the sticky footers)

Please do a functional test on iPhone X simulator and try to find something wrong, that i may have overlooked.

If you're feeling generous try to find something wrong on another iPhones/simulators/iOS versions, since there was some extra work to make things work on iOS 10.


Toggl.Daneel/Presentation/Transition/ModalPresentationController.cs
```diff
@@ -32,6 +32,18 @@ public sealed class ModalPresentationController : UIPresentationController
 
             var recognizer = new UITapGestureRecognizer(dismiss);
             AdditionalContentView.AddGestureRecognizer(recognizer);
+
+            nfloat distanceFromTop;
+            if (UIDevice.CurrentDevice.CheckSystemVersion(11, 0))
```
kornelijepetak: I see you are calling this on several locations. Maybe it would good to extract that into a common place and change it to something like `Device.IsIPhoneX` (just an example) for readability. Especially in case we need something similar in the future for other versions so that it's all in the same place.

Toggl.Daneel/Presentation/Transition/ModalPresentationController.cs
```diff
@@ -32,6 +32,18 @@ public sealed class ModalPresentationController : UIPresentationController
...
+            {
+                distanceFromTop = UIApplication.SharedApplication.KeyWindow.SafeAreaInsets.Top + 20;
```
kornelijepetak: Please extract the number to a constant (or a readonly field).

Toggl.Daneel/Presentation/Transition/ModalPresentationController.cs
arturs105: This isn't for iPhone X. This checks, whether the iOS version is >= 11, because the safe area APIs were added in iOS 11.

Toggl.Daneel/ViewControllers/OnboardingViewController.cs
```diff
@@ -29,6 +29,12 @@ public override void ViewDidLoad()
 
             preparePlaceholders();
 
+            if (UIDevice.CurrentDevice.CheckSystemVersion(11, 0))
+            {
+                var navigationBarHeight = NavigationController.NavigationBar.Frame.Height;
+                AdditionalSafeAreaInsets = new UIEdgeInsets(-navigationBarHeight, 0, 0, 0);
```
willsb: How does this play out with other devices running iOS11? 🤔

arturs105: The `if (UIDevice.CurrentDevice.CheckSystemVersion(11, 0))` makes sure that the code runs only on iOS 11 and later

Issue #981, opened 2017-11-16
Ditch StateObservable in favor of ProgressObservable 


willsb:
#978 introduces a nicer public API for interacting with the syncing algorithm. This makes the previous method (the one exposing `SyncState`) obsolete. 

We should remove the `StateObservable` from the `ISyncManager` interface and make consumers use the new observable instead, making sure to turn public members into internal while at it.

- [ ] The main view/suggestions view
- [ ] Settings view

amulware:
This is a bit more complicated than just switching the observable, since the events contain a lot more information.

We have to handle at least the following cases separetely:

- Syncing
- Synced
- Offline
- 410/418
- 401

simonrozsival:
The `IObservable<SyncProgress> SyncManager.ProgressObservable` works this way:
- it emits five types of progress:
    - `Unknown` (before the first real value is emitted, which should be just right after the app start) - this value can be ignored
    - `Syncing` - syncing started or continues (this value can be emitted multiple times during sync)
    - `Synced` - syncing finished without any error
    - `OfflineModeDetected` - the user is offline and sync cannot continue - in this case no `Synced` or `Failed` will be emitted
    - `Failed` - in case _any_ exception was thrown during syncing (including 401, 410, 418)

In case one of the 401, 410, 418 exceptions was caught, not only will the `Failed` value be emitted, but the syncing process will throw an error with the exception (`Toggl.Ultrawave.Exceptions.ClientDeprecatedException`, `Toggl.Ultrawave.Exceptions.ApiDeprecatedException`, `Toggl.Ultrawave.Exceptions.UmauthorizedException`). This must be caught and user must be navigated to the proper screen (`OutdatedAppViewModel`, `TokenResetViewModel`) and the reason must be saved to `Toggl.PrimeRadiant.IAccessRestrictionStorage`. When the error is thrown, the observable completes and won't emit any further values. The `SyncManager` will also be frozen and won't be able to sync anything, until user resolves the issue (use loggs in again or he/she updates the app).

simonrozsival:
The PR #1010 should be merged first, because it contains the logic how to handle the errors.

PR #1052, opened 2017-11-29
API Error handling service

simonrozsival:
Ref #981, Ref #1010 

I first thoguht about creating a custom observer, but that didn't seem to be a good idea, so I went with ~two~ _one_ reusable service~s~, which abstracts the handling of 401, 410, 418.

Squash whenever


simonrozsival:
I am not sure what did I do wrong in git, that all the old commits are visible (I am quite sure I branched from up-to-date develop, but it seems I didn't...) but at least these will go away when it is squashed.

Toggl.Foundation.MvvmCross/Services/ApiErrorHandlingService.cs
```diff
@@ -0,0 +1,55 @@
...
+            IAccessRestrictionStorage accessRestrictionStorage)
+        {
+            Ensure.Argument.IsNotNull(navigationService, nameof(navigationService));
+            Ensure.Argument.IsNotNull(accessRestrictionStorage, nameof(accessRestrictionStorage));
+
+            this.navigationService = navigationService;
+            this.accessRestrictionStorage = accessRestrictionStorage;
+        }
+
+        public bool MaybeHandleDeprecationError(Exception error)
```
willsb: It'll always handle the error, but won't always navigate. I think the `Maybe` is too much of legoing

Toggl.Foundation.MvvmCross/Services/ApiErrorHandlingService.cs
```diff
@@ -0,0 +1,55 @@
...
+                case ClientDeprecatedException _:
+                    accessRestrictionStorage.SetClientOutdated();
+                    navigationService.Navigate<OutdatedAppViewModel>();
+                    return true;
+            }
+
+            return false;
+        }
+
+        public bool MaybeHandleUnauthorizedError(Exception error)
```
willsb: Similar to above

willsb: Do you have a good reason why we should keep those two methods instead of a single unified method?

Toggl.Foundation.MvvmCross/ViewModels/MainViewModel.cs
```diff
@@ -151,22 +148,10 @@ private Task editTimeEntry()
...
-                    accessRestrictionStorage.SetClientOutdated();
-                    navigationService.Navigate<OutdatedAppViewModel>();
-                    return;
-                case UnauthorizedException unauthorized:
-                    accessRestrictionStorage.SetUnauthorizedAccess();
-                    navigationService.Navigate<TokenResetViewModel>();
-                    return;
-                default:
-                    throw new ArgumentException(nameof(exception));
+                throw new ArgumentException(nameof(exception));
```
willsb: This looks hella dangerous! How can you be sure that the API won't throw something else? This could crash the app for the final users

Toggl.Foundation.MvvmCross/Services/ApiErrorHandlingService.cs
amulware: .Net lingo would say this should be called `TryHandleDeprecationError` (and similar below)

Toggl.Foundation.MvvmCross/Services/ApiErrorHandlingService.cs
simonrozsival: `Unauthorized` cannot be handled during login like this, that is the main reason why.

Toggl.Foundation.MvvmCross/ViewModels/MainViewModel.cs
simonrozsival: The API might throw anything, but it won't make it into the `ProgressObservable` `OnError` - there are exactly 3 possible exceptions atm which can be observed here.

amulware: It's a good question though: How will we remember to add another case here if we ever add a 4th error?

amulware: Ideally, the type system would somehow force us to consider all cases... oh F#...

willsb: Weak type system require unit tests that break... Ensure this guys is tested against all exceptions that can escape the ProgressObservable so that if new exceptions are added we get a red light

Toggl.Foundation.MvvmCross/Services/ApiErrorHandlingService.cs
willsb: understandable have a nice day

PR #1054, opened 2017-11-29
Fix re-login

simonrozsival:
The access restriction wasn't cleared after login. Also it must be cleared when logging out on this screen, because the next time you logged in, you the flag would still be set and it would require you to re-login even though you didn't reset your API token.

Also I disabled awaiting the `ForceFullSync` after successful login - the push sync can be started here (unlike after normal login when there is no local data to be synced) and this might lead into the retry loop and take minutes to finish.

Squash whenever


simonrozsival:
The updates include moving the logic from the view model to the data source. What makes this more complex than the previous version is that I moved also the `Freeze` from the VM to the data source. If you think I should split this in two RPs, I will do so in the morning.

Toggl.Foundation.MvvmCross/ViewModels/TokenResetViewModel.cs
```diff
@@ -106,7 +105,9 @@ private void done()
         {
             Mvx.RegisterSingleton(newDataSource);
 
-            await newDataSource.SyncManager.ForceFullSync();
+            #pragma warning disable CS4014
```
willsb: Since you are no longer awaiting this, you can make this method sync and also not await the navigation

Toggl.Foundation/DataSources/TogglDataSource.cs
```diff
@@ -65,6 +70,13 @@ private IObservable<bool> hasUnsyncedData<TModel>(IRepository<TModel> repository
                 .Select(unsynced => unsynced.Any())
                 .SingleAsync();
 
-        public IObservable<Unit> Logout() => database.Clear();
+        public IObservable<Unit> Logout()
+            => Observable.Create<Unit>(async observer =>
+            {
+                await SyncManager.Freeze();
```
willsb: Since database.Clear() already returns Unit, I think a clearer way to express this would be:
```csharp
public IObservable<Unit> Logout() => 
   SyncManager.Freeze()
        .SelectMany(_ => database.Clear())
        .Do(_ => accessRestrictionStorage.ClearUnauthorizedAccess());
```

willsb: Also, why doesn't freeze return Unit as well? 🤔

simonrozsival: No, Freeze returns `SyncState`, although it could return `Unit` and there would be no difference, we don't really use the returned values.. :/ But I won't change that in this PR, but I might change that in the near future in a separate PR.

Toggl.Foundation/DataSources/TogglDataSource.cs
```diff
@@ -1,28 +1,33 @@
...
-            Ensure.Argument.IsNotNull(api, nameof(api));
             Ensure.Argument.IsNotNull(database, nameof(database));
             Ensure.Argument.IsNotNull(timeService, nameof(timeService));
-            Ensure.Argument.IsNotNull(scheduler, nameof(scheduler));
+            Ensure.Argument.IsNotNull(accessRestrictionStorage, nameof(accessRestrictionStorage));
 
             this.database = database;
+            this.accessRestrictionStorage = accessRestrictionStorage;
+
+            accessRestrictionStorage.ClearUnauthorizedAccess();
```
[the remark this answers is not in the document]
simonrozsival: That seems reasonable. I will try implementing it.

amulware:
Thanks for the list. I'm not completely happy with this, especially how the flag is cleared in the data source on logout just bothers me somehow... But I don't have a better solution right now, and I certainly don't want this PR to get more complicated.

simonrozsival:
Theoretically, we don't have to clear it on logout. When the user then logs in, it will be cleared then.

A different way how to do it could be to remember the API token which doesn't work - when the current user has a different API token, the restriction won't apply. This way it is unset implicitly, similarly to the app version restrictions.

I kinda like this idea. I would do that in a different PR. This PR would be reduced only to the refactoring of freezing the sync manager on logout and not awaiting sync after repeated login.

amulware:
Actually, I think we have to reset on logout, in case the user closes the app before coming back and logging in, right?

simonrozsival:
If the user is not logged in, he/she will be navigated to onboarding when the app starts -> login/signup, not to the token reset VM.

amulware:
I see. Still a weird edge case that I'm not sure I'm comfortable having. Clearly this needs some thought if we are gonna refactor it in any major way.

amulware:
The commit log here is getting long, let's revisit these things with a fresh start in the future. :D

PR #1059, opened 2017-11-30
Use progress observable

### LINKED ISSUE #981: Ditch StateObservable in favor of ProgressObservable

simonrozsival:
Closes #981 

There are a two things I'm still thinking of:
- The `SyncState` values are still available through the return values of the `PushSync`, `ForceFullSync`, `Freeze`, although I'm pretty sure we don't use the actual values anywhere, but we just use it for awaiting completion
- ~The errors which happen during the syncing progress are caught in `MainViewModel` and in `SettingsViewModel` - is it possible, that sync is running, but these view models aren't active and these errors won't be caught? I can think of the awaited PULL sync after login and the PULL+PUSH sync (which won't be awaited soon after PR #1054 is merged) after repeated login after the API token was reset. I will have to check the documentation of MvvmCross and see how exactly it works. I will also check how this plays with other VMs (edit view..). If this might be a problem, then maybe we would have to trasnfer the logic into the data source or the sync manager itself.~ The catching part is now only in the `TogglDataSource`.

I would also like to point out that this PR introduces only a very limited UI changes to the notification area. I believe it should be fixed in a separate PR (e.g., the offline mode bar has an icon and an "x" button to the right in Zeplin).


simonrozsival:
This is now waiting for #1054 to be merged first.

simonrozsival:
This PR is now ready for reviews.

Toggl.Foundation.MvvmCross/ViewModels/SettingsViewModel.cs
```diff
@@ -76,11 +76,11 @@ public SettingsViewModel(ITogglDataSource dataSource, IMvxNavigationService navi
...
             disposeBag.Add(dataSource.SyncManager
-                .StateObservable
-                .Subscribe(async state =>
+                .ProgressObservable
+                .Subscribe(async progress =>
                 {
-                    IsRunningSync = IsLoggingOut == false && state != SyncState.Sleep;
-                    IsSynced = IsLoggingOut == false && await isSynced();
+                    IsRunningSync = IsLoggingOut == false && progress == SyncProgress.Syncing;
+                    IsSynced = IsLoggingOut == false && progress == SyncProgress.Synced && await isSynced();
```
amulware: The change to this line should significantly improve performance, right? (ref @willsb's high CPU usage in the settings view)

On the other hand, what if all data is synced, but the app is offline, user forced a sync, sync reports offline and 'IsSynced' will be false?
Is this a real case? Is it problematic?

Toggl.Foundation/DataSources/TogglDataSource.cs
```diff
@@ -61,18 +68,27 @@ public IObservable<bool> HasUnsyncedData()
...
-                .FirstAsync()
-                .SelectMany(_ => database.Clear())
-                .Do(_ => accessRestrictionStorage.ClearUnauthorizedAccess())
-                .FirstAsync();
+        private void onSyncError(Exception exception)
+        {
+            if (apiErrorHandlingService.TryHandleDeprecationError(exception) == false
+                && apiErrorHandlingService.TryHandleUnauthorizedError(exception) == false)
+            {
+                throw new ArgumentException(nameof(exception));
```
amulware: This should pass the `exception` as inner exception and include a better message.

Toggl.Foundation.MvvmCross/ViewModels/SettingsViewModel.cs
simonrozsival: I hope this will improve the issue.

This is used by the UI only to show the "Synced [tick]" message above the Logout button. When the user is offline, it should be blank. When the user clicks the logout button, the `isSynced()` method is called (and awaited) to be sure that the state is correct.

amulware: Ah, I see. That should be good enough, thanks.

Toggl.Daneel/Startup/Setup.cs
```diff
@@ -74,8 +79,17 @@ protected override void InitializeApp(IMvxPluginManager pluginManager, IMvxAppli
...
             var apiFactory = new ApiFactory(environment, userAgent);
             var accessRestrictionStorage = new UserDataAccessRestrictionStorage(Version.Parse(version.ToString()));
-            var loginManager = new LoginManager(apiFactory, database, timeService, googleService, TaskPoolScheduler.Default, accessRestrictionStorage);
-            var deprecationHandlingService = new ApiErrorHandlingService(navigationService, accessRestrictionStorage);
+            var apiErrorHandlingService = new ApiErrorHandlingService(navigationService, accessRestrictionStorage);
+
+            ITogglDataSource createDataSource(ITogglApi api)
+            {
+                ISyncManager createSyncManager(ITogglDataSource dataSource)
+                    => TogglSyncManager.CreateSyncManager(database, api, dataSource, timeService, scheduler);
```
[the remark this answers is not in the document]
simonrozsival: Once we remove the dependency of the sync states on `ITogglDataSource`, this could be even a bit nicer.

Toggl.Foundation/DataSources/TogglDataSource.cs
```diff
@@ -37,6 +43,7 @@ public sealed class TogglDataSource : ITogglDataSource
 
             AutocompleteProvider = new AutocompleteProvider(database);
             SyncManager = createSyncManager(this);
+            SyncManager.ProgressObservable.Subscribe(_ => { }, onSyncError);
```
willsb: Please capture a reference of this disposable

willsb: Also, don't we have an extension method to subscribe only onError? 🤔

simonrozsival: We have only `public static IDisposable Subscribe<T>(this IObservable<T> observable, Action<Exception> onError, Action onCompleted)`, I can add the extension, but I'm not sure if it will be used somewhere else and is worth adding the extension method.

Issue #1036, opened 2017-11-27
Implement endpoint to fetch unknown projects


amulware:
This will be necessary to show archived projects in our reports (see #970).

We may also want to use this to show archived projects in the log, but that will be more difficult to get right.

There should be an endpoint to request several projects by id. More info soon.

amulware:
The endpoint is: `POST /reports/api/v3/projects/search` wih as body the project ids: `{"ids":[array of ids]}`

simonrozsival:
Is this really the correct URL or is this feature already on staging server? I am getting 404.

simonrozsival:
I checked the source code and the correct route is `POST /reports/api/v3/workspace/{workspace_id}/search/projects`

PR #1061, opened 2017-11-30
Merge TogglApi and ReportsApi

simonrozsival:
Ref #1036 

This PR is very straightforward, almost exclusively renaming and merging classes and interfaces into one.

Integration tests: https://www.bitrise.io/build/cbb997b46b3366d5

Squash whenever


Toggl.Ultrawave/Network/ReportsEndpoints/ProjectsSummaryEndpoints.cs
```diff
@@ -1,5 +1,6 @@
 ﻿using System;
-namespace Toggl.Ultrawave.Network
+
+namespace Toggl.Ultrawave.Network.Reports
```
willsb: This should be `Toggl.Ultrawave.Network`

Toggl.Ultrawave/Network/ReportsEndpoints/ReportsEndpoints.cs
```diff
@@ -1,13 +1,13 @@
 ﻿using System;
 using Toggl.Ultrawave.Helpers;
 
-namespace Toggl.Ultrawave.Network
+namespace Toggl.Ultrawave.Network.Reports
```
willsb: This should be `Toggl.Ultrawave.Network`

willsb: And maybe remove this ReportsEndpoints folder altogether

simonrozsival: The endpoints for reports and v9 are different and the classes holding them might have similar names, I would prefer to keep the separate namespace. I could rename the folder so that it matches the namespace.

willsb: I'd prefer that, then

PR #1063, opened 2017-11-30
Multiple occurences of same tag

### LINKED ISSUE #1050: Multiple occurences of the same tag in new TE screen
Currently you can add a single tag multiple times when creating a new TE screen.
As far as I understand, they will not be created multiple times, but from UX perspective, this should probably not be allowed.

@pe0ny It's your call whether we should just silently ignore the tap when user clicks on a tag suggestion for a tag that is already added, or should it be hidden from suggestions altogether.

kornelijepetak:
Closes #1050


Toggl.Foundation.Tests/Autocomplete/TextFieldInfoTests.cs
```diff
@@ -207,6 +207,42 @@ public void DoesNotChangeAnyPropertyIfThereIsNoProjectQueryInTheDescription()
...
+            [Fact, LogIfTooSlow]
+            public void DoesNotAddTagIfAlreadyAdded()
+            {
+                var newDescription = $"{Description}@something";
+                var tag = createTagSuggestion(1);
+
+                var textFieldInfo = TextFieldInfo.Empty.AddTag(tag);
+
+                textFieldInfo.Tags.Should().HaveCount(1);
+
```
simonrozsival: 🔥 the extra empty lines, keep it AAA!

Toggl.Foundation.Tests/Autocomplete/TextFieldInfoTests.cs
```diff
@@ -207,6 +207,42 @@ public void DoesNotChangeAnyPropertyIfThereIsNoProjectQueryInTheDescription()
...
+        {
+            [Fact, LogIfTooSlow]
+            public void AddsTagsCorrectly()
+            {
+                var tag1 = createTagSuggestion(1);
+                var tag2 = createTagSuggestion(2);
+
+                var textFieldInfo = TextFieldInfo.Empty.AddTag(tag1);
+
+                textFieldInfo.Tags.Should().HaveCount(1);
```
simonrozsival: I don't think this assertion is necessary, also use the AAA spacing in this test.

PR #1065, opened 2017-11-30
Add reports empty state for selected period

### LINKED ISSUE #1027: Reports UI: empty state for selected time period
- Show this state for users who don't have no data for selected time period.

Zeplin: https://zpl.io/blrEjP1

toggl - ios - reports empty state selected time period 1x

arturs105:
Closes #1027


Toggl.Daneel/ViewControllers/MainViewController.cs
```diff
@@ -122,7 +123,9 @@ public override void ViewWillAppear(bool animated)
             {
                 new UIBarButtonItem(UIBarButtonSystemItem.FixedSpace) { Width = -10 },
                 new UIBarButtonItem(settingsButton),
-                //new UIBarButtonItem(reportsButton)
+                #if DEBUG
```
arturs105: This is for testing the reports while it is still in development, so we don't have to uncomment this line every time to test something out. Also this way we won't accidentally release unfinished reports (given that we won't push an update built in debug mode).

kornelijepetak: And I was just meaning to ask you to separate this into separate PR. But I'm not sure if it's really that important. ;)

Issue #812, opened 2017-10-26
Reports UI


pe0ny:
In the **first section** show:
- [x] Total duration for selected time period and static illustration below
- [x] Billable percentage and dynamic chart below

In the **second section** show:
- [x] Top projects pie chart
- [x] Top projects list with project (client) name, percentage and duration
- [x] Sort projects by duration (desc.)

Zeplin: https://zpl.io/bLLONJM

![toggl - ios - reports 1x](https://user-images.githubusercontent.com/5064445/32053359-c917471c-ba63-11e7-8287-8d2a5afafc8c.png)

amulware:
This needs some more information, specifically exactly how the data shown in this report gets compiled.

- What workspace(s) are included?
- Data of the single user only?
- What does 'top projects' mean exactly? I.e. how many are shown based on what criteria?
- Is 'other' shown in the list below?
- In the design the pie chart and projects below don't match up, I'm assuming they should match in the app?
- What is that green graph under the total time? What exactly does it show and where does that data come from?

pe0ny:
1) **What workspace(s) are included?** By default we're going to show data from default workspace. But keep in mind we want to add WS select in reports page as well if user has more than one workspace. For example, you can check these designs (last two screens): https://redpen.io/nb73cb82483e8dfce5
2) **Data of the single user only?** At first yes. Next step would be to add 'Me' and 'Team' tabs so user can see his own data and he can switch to team's data.
3) **What does 'top projects' mean exactly? I.e. how many are shown based on what criteria?** We show all the projects that duration exceed 10% of the total duration. All the projects that duration doesn't exceed 10% of the total duration will be generated under 'Other' category which may contain one or more projects (reason behind is to group smaller project so they don't clutter the pie chart and the list below).
4) **Is 'other' shown in the list below?** ☝️ 
5) **In the design the pie chart and projects below don't match up, I'm assuming they should match in the app?** Yes, these should match.
6) **What is that green graph under the total time? What exactly does it show and where does that data come from?** It's just an illustration and comes from zeplin ;)

amulware:
@pe0ny Loading the report may take up to a few seconds I imagine (no tests so far, if you like, whoever does #970 can investigate this with some real data). Do we want some sort of loading state while the user is waiting for data to download?

pe0ny:
Yes. Good point. We definitely need some loading state. I'll add this to design to-do list.

PR #1064, opened 2017-11-30
Search projects (reports API)

simonrozsival:
Closes #1036 

Integration tests (commit `71fd54e`): https://www.bitrise.io/build/f37de8b7c6e37b32

Squash whenever

PR #1073, opened 2017-12-01
Reports provider

willsb:
Ref #812 

This adds the possibility of fetching reports from the Foundation Layer

Squash whenever


Toggl.Foundation/Reports/Segment.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System;
+namespace Toggl.Foundation.Reports
+{
+    public struct Segment
+    {
+        public float Value { get; }
```
simonrozsival: Isn't this name too vague? A segment of what? What about something like `ChartSegment`?

Toggl.Foundation/Reports/ReportsProvider.cs
```diff
@@ -0,0 +1,53 @@
...
+
+            return new ProjectSummaryReport(segments, totalSeconds, billablePercentage); 
+        }
+
+        private async Task<Segment> segmentFromSummary(IProjectSummary summary)
+        {
+            if (summary.ProjectId == null)
+                return new Segment(Resources.NoProject, summary.TrackedSeconds, Color.NoProject);
+
+            var project = await projectsRepository.GetById(summary.ProjectId.Value);
```
simonrozsival: This might very easily not find anything (and throw?). The PR #1064 adds an API call which should be used in this case to fetch the projects, which aren't found. I'm not sure if this PR should wait for the other PR and implement this as well, or if it should at least collect the IDs of projects, which aren't stored in the local database, so we can easily implement it next week. Probably the latter one - or at least this should catch the case when it is not found and the rest will be done in a different PR next week.

Toggl.Foundation/Reports/ReportsProvider.cs
```diff
@@ -0,0 +1,53 @@
...
+                .Apply(Task.WhenAll);
+
+            var totalSeconds = response.ProjectsSummaries.Select(x => x.TrackedSeconds).Sum();
+            var billableSeconds = response.ProjectsSummaries.Select(x => x.BillableSeconds ?? 0).Sum();
+            var billablePercentage = (100 / totalSeconds) * billableSeconds;
+
+            return new ProjectSummaryReport(segments, totalSeconds, billablePercentage); 
+        }
+
+        private async Task<Segment> segmentFromSummary(IProjectSummary summary)
```
simonrozsival: Why doesn't this return `IObservable<Segment>`?

Toggl.Foundation/Reports/ReportsProvider.cs
```diff
@@ -0,0 +1,53 @@
...
+       
+        private async Task<ProjectSummaryReport> summaryReportFromResponse(IProjectsSummary response)
+        {
+            var segments = await response.ProjectsSummaries
+                .Select(segmentFromSummary)
+                .Apply(Task.WhenAll);
+
+            var totalSeconds = response.ProjectsSummaries.Select(x => x.TrackedSeconds).Sum();
+            var billableSeconds = response.ProjectsSummaries.Select(x => x.BillableSeconds ?? 0).Sum();
+            var billablePercentage = (100 / totalSeconds) * billableSeconds;
```
simonrozsival: I would personally prefer having this in the `ProjectSummaryReport` - that would also encapsulate the consistency of the data => it wouldn't be possible to have a different sum of seconds in the array of Segments and in the float variable.

willsb: Agreed

Toggl.Foundation/Reports/ReportsProvider.cs
```diff
@@ -0,0 +1,53 @@
...
+namespace Toggl.Foundation.Reports
+{
+    public sealed class ReportsProvider : IReportsProvider
+    {
+        private readonly IProjectsSummaryApi projectSummaryApi;
+        private readonly IRepository<IDatabaseProject> projectsRepository;
+
+        public ReportsProvider(ITogglApi api, ITogglDatabase database)
+        {
+            projectsRepository = database.Projects;
```
arturs105: Ensure not null.

Toggl.Foundation/Reports/ReportsProvider.cs
willsb: Since it's only private to the class, we'll never have the need combine this with anything, hence introducing Observable semantics would be an overkill

Toggl.Foundation/Reports/ReportsProvider.cs
```diff
@@ -0,0 +1,62 @@
...
+
+        private Func<IProjectsSummary, Task<ProjectSummaryReport>> summaryReportFromResponse(long workspaceId)
+            => async response =>
+            {
+                var projectIds = response.ProjectsSummaries
+                    .Where(s => s.ProjectId.HasValue)
+                    .Select(s => s.ProjectId.Value)
+                    .ToArray();
+
+                var projectsInReport = await projectsApi.Search(workspaceId, projectIds);
```
amulware: What is the reasoning behind fetching data that the user might already have on their device, especially considering that we're on mobile where data can be costly, and the user might be doing several reports calls after each other, always requesting the same projects we already have locally? (I would predict that for most cases we will have all, or the vast majority of needed projects already.)

Toggl.Foundation/Reports/ReportsProvider.cs
```diff
@@ -0,0 +1,62 @@
...
+            projectsApi = api.Projects;
+            projectSummaryApi = api.ProjectsSummary;
+        }
+
+        public IObservable<ProjectSummaryReport> GetProjectSummary(long workspaceId, DateTimeOffset startDate, DateTimeOffset? endDate)
+            => projectSummaryApi
+                .GetByWorkspace(workspaceId, startDate, endDate)
+                .SelectMany(summaryReportFromResponse(workspaceId));
+
+        private Func<IProjectsSummary, Task<ProjectSummaryReport>> summaryReportFromResponse(long workspaceId)
```
amulware: This method is big and should probably be split up.

Toggl.Foundation/Reports/ReportsProvider.cs
simonrozsival: It would be nice to have some lazy-loading service, which would fetch the projects which weren't present locally - but that might be in some way in conflict with the sync algorithm. But we should at least cache these projects in memory, so if the user returns to the page, we don't have to make the request. So there would be three layers, where the project will be searched: in memory -> in the database -> on the server.

amulware: I'm ok with caching projects in memory, maybe in the view model layer, so that they would not get outdated too much if the app is running for days. At while the screen remains open, the user might change the date range multiple times, and searching for projects every time seems like a waste of data and time.

I don't think we'll run into any issues with the sync if we are just caching things in memory for a bit.

amulware: TL;DR: What @simonrozsival says.

amulware:
I'm not going to address this in depth anymore today, so if other reviewers are happy, go on without me.

One question though: What happens when the user selects a different date range while a query is running? Will this kick off a second query? If so, this will potentially fail, since `List<T>` is not threadsafe but is iterated and modified without any checks/locks. Please consider and fix if necessary!

Toggl.Foundation.Tests/Reports/ReportsProviderTests.cs
```diff
@@ -0,0 +1,221 @@
...
+                var projectsInApi = actualProjectIds.Where((i, id) => i % 2 != 0).ToArray();
+                var summaries = getSummaryList(actualProjectIds);
+                apiProjectsSummary.ProjectsSummaries.Returns(summaries);
+                configureRepositoryToReturn(projectsInDb, projectsInApi);
+                configureApiToReturn(projectsInApi);
+
+                ReportsProvider.GetProjectSummary(workspaceId, DateTimeOffset.Now.AddDays(-7), DateTimeOffset.Now).Wait();
+                ReportsProvider.GetProjectSummary(workspaceId, DateTimeOffset.Now.AddDays(-7), DateTimeOffset.Now).Wait();
+
+                ProjectsApi.Received(1)
```
simonrozsival: I think this test should clear the received calls on line 128 and here it should test that the second `GetProjectSummary` didn't call the API at all.

Toggl.Foundation/Reports/ReportsProvider.cs
```diff
@@ -0,0 +1,112 @@
...
+
+        public IObservable<ProjectSummaryReport> GetProjectSummary(
+            long workspaceId, DateTimeOffset startDate, DateTimeOffset? endDate)
+            => projectSummaryApi
+                .GetByWorkspace(workspaceId, startDate, endDate)
+                .SelectMany(response => summaryReportFromResponse(response, workspaceId));
+
+        private IObservable<ProjectSummaryReport> summaryReportFromResponse(IProjectsSummary response, long workspaceId)
+            => response.ProjectsSummaries
+                    .Select(s => s.ProjectId)
```
simonrozsival: Too much indentation.

Toggl.Foundation/Reports/ReportsProvider.cs
```diff
@@ -0,0 +1,112 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reactive.Linq;
+using System.Threading.Tasks;
```
simonrozsival: 🔥 line 5

Toggl.Foundation/Reports/ReportsProvider.cs
```diff
@@ -106,7 +110,10 @@ private static ChartSegment chartFromSummary(IProjectSummary summary, IList<IPro
 
         private void persistInMemoryCache(List<IProject> apiProjects)
         {
-            apiProjects.ForEach(memoryCache.Add);
+            lock (listLock)
```
simonrozsival: 🔥 this lock is not needed

willsb: We are modifying the list here

simonrozsival: But you are already in the critical section when you get to this lock.

PR #1078, opened 2017-12-04
Add borders to tags in edit time entry view

### LINKED ISSUE #685: Edit view: multiple tags
When user has selected multiple tags, let's use only one row to display tags and if there are more to fit, use '...' to indicate. When user taps on '...' or on any other tag, open tag view.

toggl - ios - edit tags 1x

Zeplin: https://zpl.io/2vnPWw5

arturs105:
Closes #685


Toggl.Daneel/Binding/TextViewTagListTargetBinding.cs
```diff
@@ -0,0 +1,89 @@
...
+            }
+
+            Target.AttributedText = createAttributedString(
+                tagTokens.Take(tagCount)
+                .ToList()
+                .Append(elipsisTag)
+            );
+        }
+
+        private NSAttributedString createAttributedString(
```
willsb: Make this static, pl0x

Toggl.Daneel/Binding/TextViewTagListTargetBinding.cs
```diff
@@ -0,0 +1,89 @@
...
+                return;
+            }
+
+            var tagCount = cumulativeLengths
+                .FirstIndexOf(length => length + elipsisTagWidth >= Target.Bounds.Width);
+
+            if (tagCount == -1)
+            {
+                Target.AttributedText = createAttributedString(
+                    new TokenTextAttachment[] { elipsisTag });
```
willsb: You can probs omit `TokenTextAttachment`, methinks

Toggl.Daneel/Extensions/TokenExtensions.cs
```diff
@@ -82,21 +93,21 @@ private static void addProjectAttachmentsIfNeeded(TextFieldInfo info, NSMutableA
...
-
-                var tokenString = new NSMutableAttributedString(NSAttributedString.FromAttachment(textAttachment));
                 var attributes = new UIStringAttributes { ParagraphStyle = paragraphStyle };
                 attributes.Dictionary[TagIndex] = new NSNumber(i);
-                tokenString.AddAttributes(attributes, new NSRange(0, tokenString.Length));
-
-                finalString.Append(tokenString);
+                tagToken.AddAttributes(attributes, new NSRange(0, tagToken.Length));
+                finalString.Append(tagToken);
+                i++;
```
willsb: I find a for loop much more readable than a foreach loop that increments

Toggl.Daneel/TokenTextAttachment.cs
```diff
@@ -31,7 +31,7 @@ public sealed class TokenTextAttachment : NSTextAttachment
             using (var context = UIGraphics.GetCurrentContext())
             {
                 var tokenPath = UIBezierPath.FromRoundedRect(new CGRect(
-                    x: interTokenSpacing,
+                    x: interTokenSpacing / 2,
```
willsb: what are those? 👀

Toggl.Daneel/ViewControllers/EditTimeEntryViewController.cs
```diff
@@ -192,6 +187,19 @@ public override void ViewDidLoad()
...
         }
 
+        public override void ViewDidLayoutSubviews()
+        {
+            base.ViewDidLayoutSubviews();
+            
+            //This binding needs to be created, when TagsTextView has it's
+            //proper size. In ViewDidLoad() TagsTextView width isn't initialized
+            //yet, which results in displaying less tags than possible
+            this.CreateBinding(TagsTextView)
```
willsb: This is not ideal 🤔 
Can't you make it so® that you recalculate the amount of tags when the view layouts?

Toggl.Daneel/ViewControllers/EditTimeEntryViewController.cs
```diff
@@ -209,6 +217,7 @@ private void prepareViews()
             PreferredContentSize = View.Frame.Size;
             BillableSwitch.Resize();
             prepareDescriptionField();
+            TagsTextView.TextContainer.LineFragmentPadding = 0;
```
willsb: I'm 12 and what's this?

Toggl.Multivac/Extensions/EnumerableExtensions.cs
```diff
@@ -0,0 +1,21 @@
...
+using System.Collections.Generic;
+
+namespace Toggl.Multivac.Extensions
+{
+    public static class EnumerableExtensions
+    {
+        public static int FirstIndexOf<T>(
+            this IEnumerable<T> self, Func<T, bool> predicate)
+        {
+            int i = 0;
```
willsb: Use a for loop. They exist for a reason :P

amulware: And handle the IEnumerator<T> manually? Or is the nice solution I'm not seeing?

arturs105: I guess it will have to be a for loop that uses an `IEnumerable`

Toggl.Daneel/TokenTextAttachment.cs
arturs105: These were added so that there is a little whitespace at both ends of the token, instead of just the beginning.
Before: `__[tag]`
Now: `_[tag]_`
This was done because tags were slightly off center in edit view.
Now that, i think about this - distance between the description and text in start view has been decreased due to this.

Toggl.Daneel/ViewControllers/EditTimeEntryViewController.cs
arturs105: Is there a way to make MvvmCross do that from the view controller?

Toggl.Daneel/ViewControllers/EditTimeEntryViewController.cs
arturs105: A small padding between the left edge of the `UITextView` and the place, where text starts. Setting this to 0 allows displaying text (tags in this case) from edge to edge.

amulware:
I'm sorry, but I must protest the manual use of enumerators, there is really no need for that. Not to mention they are so easy to use incorrectly. For example, they should be disposed which was forgotten here. Unless @willsb has some alternative or strong case for doing it this way, please revert the loop change.

PR #1091, opened 2017-12-04
Prevent keyboard on select workspace in settings

### LINKED ISSUE #1082: Settings: switch default ws view opens a keyboard
Currently, when accessing WS view from settings, it opens a keyboard which it shouldn't. This allows user start typing and items disappearing from the list. A keyboard should be hidden/not triggered in this view. Note that a keyboard should be available in WS view when creating a new project.

kornelijepetak:
Closes #1082 

If there was a better way, please advise!


willsb:
`allowQuerying` and `shouldFocusOnStart` will both always have the same value, so there's no added benefit to adding another property

kornelijepetak:
Ah well. All the work for one line :)
Yeah, I misunderstood the `AllowQuerying` in this context.

PR #1102, opened 2017-12-06
Show client in TE autocomplete suggestion

### LINKED ISSUE #1058: TE creation view: show client name when suggesting time entries
Show client name in TE creation view when suggesting time entries to a user. Use the same order we use in the log view:

Description
Project Client

Zeplin: https://zpl.io/aMLGMZz

### LINKED ISSUE #1104: Fix small padding in front of the project dot in iOS
Ref PR #1102

simonrozsival:
Closes #1058, Closes #1104 

Squash whenever


kornelijepetak:
There's also a question of how tasks should be displayed... I don't see the design on Zeplin showing tasks... @simonrozsival? @pe0ny 

In this PR the project name and task are of the same color and look like this:

Project: "ABCDEFGH"
Task: "Do something different"

image

Should they be of the same color?

simonrozsival:
I believe it should behave the same way it does in the rest of the app - the task is always colored the same way the project is.

kornelijepetak:
Be that as it may, here's an example of my newly created project:

It is called: `TheProject: Shut the door`.

But by suggestion, I can't discern whether that's a project name or a Project+Task.

Just saying. However, that's not maybe that much related to this PR.

When padding is fixed, ✅ .

image

simonrozsival:
This is the very same padding that is unfortunately also present in the time entries log and suggestions. Fixing this component would be nice, but I think this would deserve a separate issue and PR.
```csharp
//There neeeds to be a space before the dot, otherwise the colors don't work
var result = new NSMutableAttributedString(" ");
var attachmentString = NSAttributedString.FromAttachment(attachment);
result.Append(attachmentString);
```

pe0ny:
@kornelijepetak That's correct behaviour how to display tasks. I see your point when a user creates a project name which includes a colon (:) but for now, I would still keep it that way.

kornelijepetak:
I don't mind it like this, I was just raising concern and since it's been addressed, I have nothing more to say about this :)

PR #1105, opened 2017-12-06
Continue time entry with tags

### LINKED ISSUE #1103: Continuing TEs does not copy tags
But it should. Fix pls.

arturs105:
Closes #1103 

Squash whenever


Toggl.Foundation.Tests/MvvmCross/ViewModels/TimeEntriesLogViewModelTests.cs
```diff
@@ -308,15 +312,17 @@ public sealed class TheContinueTimeEntryCommand : TimeEntriesLogViewModelTest
...
                 timeEntry.Project.Returns(project);
                 timeEntry.TaskId.Returns(taskId);
+                timeEntry.TagIds.Returns(tagIds);
                 timeEntry.Duration.Returns(100);
                 var timeEntryViewModel = new TimeEntryViewModel(timeEntry);
 
                 ViewModel.ContinueTimeEntryCommand.ExecuteAsync(timeEntryViewModel).Wait();
 
                 DataSource.TimeEntries.Received().Start(Arg.Is<StartTimeEntryDTO>(dto =>
+                    dto.TagIds.SequenceEqual(tagIds) &&
```
simonrozsival: This fails, when `tagIds` is null. Either avoid having null here in the first place, or change this assertion.

PR #1109, opened 2017-12-06
Fix DialogService

### LINKED ISSUE #1088: UI doesn't update when creating project into non-default WS
*I used test.account.2 to test that.*

1) Start a new TE
2) Tap @'add some text here' eg @testingprojectabc
3) Tap "+ Create project 'testingprojectabc'"
4) Project creation view opens: Tap on a WS field and change the WS to non-default one. (By default was selected 'AutoShop' and I changed it to 'Big Company Workspace')
5) Tap 'Create'
6) App directed me back to TE creation view where I saw plain text/description: '@testingprojectabc' and no token.
7) Tap green checkmark: Project is created in the DB (checked from the webapp) but mobile UI didn't update.

kornelijepetak:
Closes #1088 

- Extracted finding the top ViewController to a service.

_Similar is used with Google login and 1Password, but I it can be done as another PR._


kornelijepetak:
I get what you mean, will fix in a second.

Toggl.Daneel/Startup/Setup.cs
```diff
@@ -42,15 +42,15 @@ public Setup(MvxApplicationDelegate applicationDelegate, UIWindow window)
...
+        private Setup(MvxApplicationDelegate applicationDelegate, IMvxIosViewPresenter presenter)
             : base(applicationDelegate, presenter)
         {
         }
 
         protected override void InitializeLastChance()
         {
             base.InitializeLastChance();
-
+
```
kornelijepetak: Wait, what?

Issue #1076, opened 2017-12-04
Fix incorrect suggestions behavior


simonrozsival:
I noticed two weird ways the "Create project" and "Create tag" suggestions in the "Start time entry view" which should be fixed to make it more predictable:

1. Creating duplicate tags
    - create a tag "abcde"
    - go to start time entry view
    - type in "abcdef"
    - hit backspace
    - you will get a `+ Create tag "abcde"` suggestion
    - click on the suggestion
    - you now have two tags called "abcde"
2. Incorrect project name suggestion
    - go to start time entry view
    - type in something like "time entry @project #tag" without selecting any suggestions
    - you will get a suggestion `+ Create project "tag"` instead of "project" (or in fact I am not really sure what should be the correct suggestion, but definitely not "tag" - this should be discussed in the comments IMO)

simonrozsival:
@pe0ny @amulware What should happen, when user types something like "time entry @project #tag" without selecting any suggestion in the meantime and moving the cursor (so it is at the very end)?`
    

a) show  `+ Create project "project #tag"` (supposing there is no project with that name) 
b) show `+ Create project "tag"` (supposing there is no project with that name)
c) when typing #, switch from suggesting projectst to suggesting tags containing "tag"?
d) something else

amulware:
a) maybe. # is definitely allowed in project names
b) absolutely not
c) maybe, this is similar to what todoist does, which is pretty intuitive to me, but people can't use '#' for searching projects then
d) not sure

Unless there is another option I can't think of, I think c) is best.

pe0ny:
I agree with Paul: c) is the best and most intuitive. Also, that's how we already do it :)

pe0ny:
Oh, never mind me, we don't. We still suggest projects, not tags. So yes, we should suggest tags.

simonrozsival:
OK

simonrozsival:
Follow-up question: What should happen when user types `abc @def ghi @jkl mno` or `abc #def ghi #jkl mno`? Should we consider only the part from the second symbol, or the whole thing after the first symbol?

amulware:
From the second in my opinion.

PR #1118, opened 2017-12-07
Fix autocompletion bug 2/2

simonrozsival:
Ref #1076 

This fixes the first bug described in the issue. A related problem was that the "No project" wasn't displayed when only the "+ Create project" suggestion was shown, so that is fixed in this PR as well.

Squash whenever


Toggl.Daneel/ViewSources/CreateSuggestionGroupedTableViewSource.cs
```diff
@@ -14,7 +14,20 @@ public abstract class CreateSuggestionGroupedTableViewSource<T> : GroupedCollect
 
         public string Text { get; set; }
 
-        public bool SuggestCreation { get; set; }
+        private bool suggestCreation;
+
```
willsb: Remove this empty line pl0x

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
```diff
@@ -473,12 +473,21 @@ private void onSuggestions(IEnumerable<AutocompleteSuggestion> suggestions)
...
         private IEnumerable<WorkspaceGroupedCollection<AutocompleteSuggestion>> groupSuggestions(
             IEnumerable<AutocompleteSuggestion> suggestions)
         {
             var firstSuggestion = suggestions.FirstOrDefault();
+            if (firstSuggestion == null && IsSuggestingProjects && TextFieldInfo.WorkspaceId.HasValue)
+            {
+                var collection = new WorkspaceGroupedCollection<AutocompleteSuggestion>(
+                    "", new[] { ProjectSuggestion.NoProject(TextFieldInfo.WorkspaceId.Value, "") });
+
+                return new[] { collection };
```
willsb: A comment here would be nice. Something in the lines of:

`// This ensures that "No Project" is shown even when no suggestion matches the search query`

Issue #813, opened 2017-10-26
Reports UI: Calendar


pe0ny:
- [ ] Tap on the header 'This week' to show/hide the calendar
- [ ] By default show 'This week'
- [ ] Use left-right swipe to switch months
- [ ] To pick a date range: first tap - start date, second tap - end date
- [ ] User can scroll shortcuts left-right
- [ ] Close the calendar once date range has been selected (after second tap or selecting a range from shortcuts)

Shortcuts:
- This week
- Last week
- This month
- Last month
- This year

Zeplin: https://zpl.io/VYPXp0n & https://zpl.io/2ZqXD3X

![toggl - ios - reports calendar v1 1x](https://user-images.githubusercontent.com/5064445/32053593-7af4fde4-ba64-11e7-9958-81a6d8bf5917.png) ![toggl - ios - reports calendar v2 1x](https://user-images.githubusercontent.com/5064445/32053596-7dbbe934-ba64-11e7-93d0-ca0627e3b73c.png)

amulware:
In the design, it looks like the calendar makes the header expand, instead of being on top of it. Does it scroll with the rest of the page then? Or does it stick to the top when scrolling?

amulware:
@pe0ny The above questions.

Also, the report can show at most the data for one year. How do we limit the user to that? And in general, do we have some limit about how far back they can go? How far into the future?

pe0ny:
- Calendar expands the header; Content reveals the calendar.
- When user start scrolling down (swipes up), then content covers the calendar and in the end closes it.
- We should limit it by not allowing user to move back in calendar more than one year.
- Current month should be the max value.

amulware:
Also: How do 'this week', 'this month' and 'this year' work when in the middle of a week/month/year? Is it showing the last 7 days, or everything since monday (and similar for month, year)?

pe0ny:
- Even when in the middle of a week/month/year, it shows full week/month/year

amulware:
I'm not sure I understand. :D - so what range does it show on a Wednesday? Wed - Wed? If so, I also need an explanation of what 'last week' means.

And one other point, just to clarify: We CAN get data longer than one year ago, just not for more than one year at a time (i.e. start and end date can be no more than one year apart) - do you still want to limit it to the last year?

pe0ny:
On Wednesday it will show range from Monday to Sunday OR when user has set in settings that weeks starts with Sunday, then from Sunday to Saturday.

[...]

PR #1127, opened 2017-12-08
Add models for reports calendar

arturs105:
Ref #813 

I decided to split this up into several smaller PRs for easier reviews.


Toggl.Foundation.MvvmCross/ViewModels/CalendarPageViewModel.cs
```diff
@@ -0,0 +1,61 @@
...
+        public List<CalendarDayViewModel> Days { get; }
+            = new List<CalendarDayViewModel>();
+
+        public CalendarMonth CalendarMonth { get; }
+
+        public CalendarPageViewModel(CalendarMonth calendarMonth)
+        {
+            CalendarMonth = calendarMonth;
+
+            var firstDayOfMonth = (int)new DateTime(
```
simonrozsival: This would deserve its own function as you use the almost same 5 lines again in the constructor.

Toggl.Foundation.MvvmCross/ViewModels/CalendarPageViewModel.cs
```diff
@@ -0,0 +1,61 @@
...
+namespace Toggl.Foundation.MvvmCross.ViewModels
+{
+    public sealed class CalendarPageViewModel
+    {
+        public List<CalendarDayViewModel> Days { get; }
+            = new List<CalendarDayViewModel>();
+
+        public CalendarMonth CalendarMonth { get; }
+
+        public CalendarPageViewModel(CalendarMonth calendarMonth)
```
simonrozsival: This should probably take in mind the `Toggl.Multivac.IUser.BeginningOfWeek` property. Now we assume that the user prefers Monday as the first day of the week.

Toggl.Foundation.MvvmCross/ViewModels/CalendarPageViewModel.cs
```diff
@@ -0,0 +1,61 @@
...
+            CalendarMonth = calendarMonth;
+
+            var firstDayOfMonth = (int)new DateTime(
+                calendarMonth.Year,
+                calendarMonth.Month, 1)
+                    .DayOfWeek;
+            firstDayOfMonth = convertToMondayBased(firstDayOfMonth);
+
+            //Add a few days from previous month
+            if (firstDayOfMonth != 0)
```
simonrozsival: You don't need this if, you can do it the same way as with the last week.

Toggl.Foundation.MvvmCross/ViewModels/CalendarPageViewModel.cs
```diff
@@ -0,0 +1,61 @@
...
+                {
+                    Days.Add(new CalendarDayViewModel(
+                        daysInPreviousMonth - i,
+                        false));
+                }
+            }
+
+            //Add days from the target month
+            var daysInMonth = calendarMonth.DaysInMonth;
+            for (int i = 0; i < daysInMonth; i++)
```
simonrozsival: Maybe you could refactor this to a separate method, which would get a range of days + the boolean and add the days? This constructor is IMO too long and the for loops are used three times.

Toggl.Foundation.MvvmCross/ViewModels/ReportsCalendarViewModel.cs
```diff
@@ -0,0 +1,39 @@
...
+        public override void Prepare()
+        {
+            base.Prepare();
+
+            var now = timeService.CurrentDateTime;
+
+            CurrentMonth = new CalendarMonth(now.Year, now.Month);
+
+            var monthIterator
+                = new CalendarMonth(now.Year, now.Month).AddMonths(-11);
```
simonrozsival: Give this constant a name please.

Toggl.Foundation.Tests/MvvmCross/ViewModels/ReportsCalendarViewModelTests.cs
```diff
@@ -0,0 +1,77 @@
...
+
+                await ViewModel.Initialize();
+
+                ViewModel.Months.Should().HaveCount(12);
+                var firstDateTime = now.AddMonths(-11);
+                var month = new CalendarMonth(
+                    firstDateTime.Year, firstDateTime.Month);
+                for (int i = 0; i < 12; i++, month = month.Next())
+                {
+                    ViewModel.Months[i].CalendarMonth.Should().Be(month);
```
simonrozsival: Please also test that the beginning of week is the same as the user's preference.

Toggl.Foundation.MvvmCross/ViewModels/CalendarPageViewModel.cs
```diff
@@ -0,0 +1,60 @@
+﻿using System.Collections.Generic;
+using Toggl.Multivac;
+
+namespace Toggl.Foundation.MvvmCross.ViewModels
+{
+    public sealed class CalendarPageViewModel
```
simonrozsival: There are no tests for this class and I think this should be tested very carefully.

Toggl.Foundation.Tests/MvvmCross/ViewModels/ReportsCalendarViewModelTests.cs
arturs105: There isn't an easy way to test this since `CalendarPageViewModel` doesn't publicly expose the beginning of week. Though, i think, this will be covered in `CalendarPageViewModel` tests.

Toggl.Foundation.Tests/MvvmCross/ViewModels/CalendarPageViewModelTests.cs
```diff
@@ -0,0 +1,107 @@
...
+                    .Skip(calendarMonth.DaysInMonth);
+
+                if (expectedDayCount == 0)
+                    daysFromNextMonth.Should().BeEmpty();
+                else
+                    daysFromNextMonth
+                        .Should()
+                        .OnlyContain(day => !day.IsInCurrentMonth);
+            }
+        }
```
simonrozsival: Please add one more test for me - that the number of days % 7 is always zero.

Issue #1047, opened 2017-11-28
Consolidate Integration test attributes.


willsb:
Right now we have `LogTestInfoAttribute`, which is vital because of the `Thread.Sleep`, and `LogIfTooSlowAttribute`, which shows the slow tests. 

Log If too slow doesn't really apply for integration tests because its values for slow aren't configurable (and even if they were, the tests vary wildly due to internet connection speed/server availability). 

We should ditch the usage of `LogIfTooSlow` from the integration tests and instead make `LogTestInfo` nicer by Including duration before the name of the test. We also need to make sure that every method has the LogTestInfoAttribute.

Alternatively we could instead override `FactAttribute` and implement our own `TestDiscoverer` that does all that automagically, but I'll leave that decision to whoever picks this issue.

PR #1131, opened 2017-12-08
Stop using the LogIfTooSlow attribute in integration tests

simonrozsival:
Ref #1047 

Integration tests do not work well in command line locally (with current cake configuration). Also we don't need to measure which integration tests are slow.

> but do we actually care if integration tests run slow? I mean.. they all do
> -- @amulware 

Squash whenever


amulware:
Integration tests: https://www.bitrise.io/build/e2e6ec38b139600e

PR #1136, opened 2017-12-08
Reports UI

willsb:
Ref #812 

This adds all UI components up to the pie chart. 

There’s something wrong with the deserialisation of the reports api (my guess is that the error is linker related), so I’m using a stub for the time being. Unit Tests are to be added once the stub is replaced.

To functional test this, you can use mock values.

Squash whenever


Toggl.Daneel/Converters/ReportPercentageLabelValueConverter.cs
```diff
@@ -0,0 +1,41 @@
...
+using Foundation;
+using MvvmCross.Platform.Converters;
+using MvvmCross.Plugins.Color.iOS;
+using Toggl.Foundation.MvvmCross.Converters;
+using Toggl.Foundation.MvvmCross.Helper;
+using UIKit;
+
+namespace Toggl.Daneel.Converters
+{
+    public class ReportPercentageLabelValueConverter : MvxValueConverter<float, NSAttributedString>
```
arturs105: Class not sealed

Toggl.Daneel/Converters/TimeSpanReportLabelValueConverter.cs
```diff
@@ -0,0 +1,62 @@
...
+using Foundation;
+using MvvmCross.Platform.Converters;
+using MvvmCross.Plugins.Color.iOS;
+using Toggl.Foundation.MvvmCross.Converters;
+using Toggl.Foundation.MvvmCross.Helper;
+using UIKit;
+
+namespace Toggl.Daneel.Converters
+{
+    public class TimeSpanReportLabelValueConverter : MvxValueConverter<TimeSpan, NSAttributedString>
```
arturs105: Not sealed

Toggl.Daneel/Converters/ReportPercentageLabelValueConverter.cs
```diff
@@ -0,0 +1,41 @@
+﻿using System;
+using System.Globalization;
+using Foundation;
+using MvvmCross.Platform.Converters;
+using MvvmCross.Plugins.Color.iOS;
+using Toggl.Foundation.MvvmCross.Converters;
```
arturs105: 🔥

Toggl.Daneel/Converters/TimeSpanReportLabelValueConverter.cs
```diff
@@ -0,0 +1,62 @@
+﻿using System;
+using System.Globalization;
+using Foundation;
+using MvvmCross.Platform.Converters;
+using MvvmCross.Plugins.Color.iOS;
+using Toggl.Foundation.MvvmCross.Converters;
```
arturs105: 🔥

Toggl.Daneel/ViewControllers/ReportsViewController.cs
```diff
@@ -2,13 +2,19 @@
 using MvvmCross.Binding.BindingContext;
 using MvvmCross.Binding.iOS;
 using MvvmCross.iOS.Views;
+using MvvmCross.iOS.Views.Presenters.Attributes;
+using MvvmCross.Plugins.Color.iOS;
 using MvvmCross.Plugins.Visibility;
```
arturs105: 🔥

Toggl.Daneel/ViewControllers/ReportsViewController.cs
```diff
@@ -21,20 +27,44 @@ public override void ViewDidLoad()
...
+                      .For(v => v.AttributedText)
+                      .To(vm => vm.BillablePercentage)
+                      .WithConversion(new ReportPercentageLabelValueConverter());
+
+            bindingSet.Bind(TotalDurationLabel)
+                      .For(v => v.AttributedText)
+                      .To(vm => vm.TotalTime)
+                      .WithConversion(new TimeSpanReportLabelValueConverter());
+
+            // Pretty stuff
```
arturs105: Space at the start of commnet

pe0ny:
@arturs105 

> maybe we should round billable percentage to 2 decimal places? 

Yes, agreed.


> should we display the segment label, if the segment is really small (so the label is drawn outside the segment itself)

You're talking about pie chart segments, correct? Can you give me an example, a screenshot maybe? But I guess the answer is yes, we shouldn't display it then.
Also, keep in mind we show the projects that duration exceed 10% of the total duration. All the projects that duration doesn't exceed 10% of the total duration will be generated under 'Other' category which may contain one or more projects. It was also mentioned in this issue #812 Do you think it might be the case anyway?

arturs105:
I think i may have had a project with less than 10%. Gonna try with exactly 10% now.

But i think category "other" can still be really small and we should maybe hide the label in those cases.

arturs105:
The gray area in the screenshot is 10%, so that's more than enough space for the label. However, "other" segment is really small (i suppose it can be in real-life situations as well) and it overlaps the segment right next to it. We should hide the label (or display it outside the chart).

Also: long names need to be truncated

screen shot 2017-12-12 at 12 01 20

pe0ny:
@arturs105 Thanks 🙌 Let's do it so then: disable/hide labels when the segment is under 10%.

pe0ny:
About long names: can we limit it to 18 characters. If the project name is more than 18 char, add three dots at the end '...'

Toggl.Daneel/Converters/TimeSpanReportLabelValueConverter.cs
willsb: I'm afraid this is actually needed

Toggl.Daneel/ViewControllers/ReportsViewController.cs
willsb: Do we like spaces or not? 😄  I personally do not, but some people do.

kornelijepetak: IMHO it looks much better if there is some padding before the text.

arturs105: I have no preference for spaces vs no spaces, but we should stick to one version. Since there are 2 other comments without a space in the same file, i vote for removing the space.

Toggl.Daneel/Converters/ReportPercentageLabelValueConverter.cs
```diff
@@ -29,7 +29,7 @@ public sealed class ReportPercentageLabelValueConverter : MvxValueConverter<floa
 
         protected override NSAttributedString Convert(float value, Type targetType, object parameter, CultureInfo culture)
         {
-            var percentage = $"{value}%";
+            var percentage = $"{value.ToString("{0:0.00}")}%";
```
arturs105: I don't think this is correct.

willsb:
As discussed on slack, grouping of small segments into a "Others" category will be done in a different PR

Toggl.Daneel/ViewControllers/ReportsViewController.designer.cs
```diff
@@ -12,20 +12,42 @@ namespace Toggl.Daneel.ViewControllers
 	[Register ("ReportsViewController")]
 	partial class ReportsViewController
 	{
+		[Outlet]
+		Toggl.Daneel.Views.Reports.PercentageView BillablePercentageGraph { get; set; }
+
+		[Outlet]
+		UIKit.UILabel BillablePercentageView { get; set; }
```
arturs105: `BillablePercentageLabel`

Toggl.Daneel/ViewControllers/ReportsViewController.designer.cs
```diff
@@ -12,20 +12,42 @@ namespace Toggl.Daneel.ViewControllers
 	[Register ("ReportsViewController")]
 	partial class ReportsViewController
 	{
+		[Outlet]
+		Toggl.Daneel.Views.Reports.PercentageView BillablePercentageGraph { get; set; }
```
arturs105: I don't think `graph` is the appropriate term here. `BillablePercentageView` (or something else would be better).

Toggl.Daneel/ViewControllers/ReportsViewController.cs
```diff
@@ -43,27 +71,8 @@ private void prepareViews()
...
-            totalDuration.Append(new NSAttributedString(
-                ":00",
-                new UIStringAttributes
-                {
-                    Font = UIFont.SystemFontOfSize(fontSize, UIFontWeight.Light)
-                })
-            );
-            TotalDurationLabel.AttributedText = totalDuration;
+            TotalDurationGraph.Image = 
+                TotalDurationGraph.Image.ImageWithRenderingMode(UIImageRenderingMode.AlwaysTemplate);
```
arturs105: > When splitting statements into multiple lines, do so by starting new lines with operators, and using an additional level of indentation. 

https://github.com/toggl/mobileapp/blob/develop/docs/style-guide.md

Toggl.Daneel/Views/Reports/PercentageView.cs
```diff
@@ -0,0 +1,49 @@
...
+
+namespace Toggl.Daneel.Views.Reports
+{
+    [Register(nameof(PercentageView))]
+    public sealed class PercentageView : UIView
+    {
+        private static readonly UIColor disabledColor = Color.Reports.Disabled.ToNativeColor();
+        private static readonly UIColor activeColor = Color.Reports.PercentageDisabled.ToNativeColor();
+
+        private readonly CALayer layer = new CALayer();
```
arturs105: This could use a better name.

Toggl.Foundation.MvvmCross/ViewModels/ReportsViewModel.cs
```diff
@@ -1,21 +1,41 @@
 ﻿using System;
+using System.Linq;
+using System.Reactive;
+using System.Reactive.Linq;
+using System.Reactive.Subjects;
 using MvvmCross.Core.ViewModels;
+using Toggl.Foundation.DataSources;
```
arturs105: 🔥

Toggl.Foundation.Tests/MvvmCross/ViewModels/MainViewModelTests.cs
```diff
@@ -110,6 +108,22 @@ public sealed class TheOpenSettingsCommand : MainViewModelTest
             }
         }
 
+        public sealed class TheOpenReportsCommand : MainViewModelTest
+        {
+            [Fact, LogIfTooSlow]
+            public async Task NavigatesToTheSettingsViewModel()
```
arturs105: Wrong test name

arturs105:
Also: i think percentages inside the chart segments need to be rounded to 2 decimal places. Calling @pe0ny for confirmation.

pe0ny:
@arturs105 I actually vote for no decimal places since it's such a small text there.

Toggl.Daneel/Views/Reports/PieChartView.cs
```diff
@@ -0,0 +1,108 @@
...
+
+        public override CGRect Bounds
+        {
+            get
+            {
+                return base.Bounds;
+            }
+            set
+            {
+                viewCenterX = value.Size.Width * 0.5f;
```
kornelijepetak: Not at all important, but dividing by 2 would better convey the purpose, IMO.

Toggl.Daneel/Views/Reports/PieChartView.cs
```diff
@@ -0,0 +1,108 @@
...
+                // Calculate end angle
+                var endAngle = startAngle + 2 * pi * percent;
+
+                // Draw arc
+                ctx.MoveTo(viewCenterX, viewCenterY);
+                ctx.AddArc(viewCenterX, viewCenterY, radius, startAngle, endAngle, clockwise: false);
+                ctx.FillPath();
+
+                // Disable drawing on segments that are too small
+                if (percent > textDrawingThreshold)
```
kornelijepetak: The problem I have with this approach is this:

With dummy data where there are many projects, all with similar duration, the labels are not visible at all, and since there is no legend, this chart is almost unusable, unless you really can differentiate your projects by the color.

image

willsb: The 10% threshold was defined by @pe0ny, so if she feels this should be improved, it should happen in a different PR (we need to move forward with this one).

pe0ny: I don't think it's a problem since there will be a legend: list view below https://zpl.io/bLLONJM

PR #1138, opened 2017-12-11
Remove user caching

### LINKED ISSUE #1099: After changing default WS, new TEs won’t be tracked under default WS right away
*Used test.account.2 to test.*

1) I changed default WS on mobile
2) Checked - it was also changed on web
3) Started new TE on mobile - it wasn't tracked under default WS.
4) Restarted the app couple of times, then it worked again.

kornelijepetak:
Fixes #1099 

The issue is that the user is cached and the cache is never invalidated.
Now that I've said it, maybe `InvalidateCache` would be a better name :)

Please look at the code and tell me if this is OK or it should be done with some kind of `RX` construct (and why).

Also do a functional.

Since the user is no longer cached, `UserDataSource.Current()` no longer has any side effects so I have converted it to a property.


simonrozsival:
I am not sure we need some kind of cache here - Realm is fast and we don't use it frequently (start/stop/continue time entry, settings view, resetting API token) and we never use it in a loop. Is there something that I don't see which justifies this caching, @willsb?

kornelijepetak:
I was thinking the same thing but didn't yet have time to ask if caching was done with a particular reason.

willsb:
🔥  cache

simonrozsival:
We definitely need at least one more test: if you retrieve some data from the `Current()` function, then the data stored in the database changes, and you retrieve the data from `Current()` again, you should get the up-to-date data in both cases (the second one would not be up-to-date if there was some caching layer and it wasn't invalidated, as it was in the previous version). Did I describe it understandably?

kornelijepetak:
Of course this might be a meaningful test, from the perspective of the app that **once had** a cached user. But if it never had caching, we wouldn't even think about this test. For instance, we don't test whether a Tag returned is a cached version, cause we never intended tags to be cached. I don't know if I'm describing this clearly. If we follow the same reasoning, we should test all other entities.

All that being true, **I will** create a test for it, only because the `User` is somewhat different from other entities (in the following aspects: it had caching which we removed, there is only one User in the app). 

Also it's interesting to note that there were no caching tests before... When I removed cache, none of the tests failed...

PR #1139, opened 2017-12-11
Start full sync after app resumes from background 

### LINKED ISSUE #1112: Start full sync after app resumes from background
Of course only if the user is logged in (and none of the access restrictions are active). Also, I think we should add some minimum time the app has to have been in the background, say.. 5 minutes?

I think it would be useful for future things to implement this with an 'app became active after X time inactivity' observable, which we could use for metrics and other features in the future.

I'm also not sure if this should be entirely UI code, or if it makes sense to have logic like this in Foundation.

simonrozsival:
Closes #1112 

The `IBackgroundService` service should be called whenever the app goes into the background and is resumed later. The observable can be configured to emit values only after the app was suspended for at least some minimum time period.

The iOS integration is implemented (see `AppDelegate`) in this PR. I also refactored and extended the TogglDataSource and SyncManager to work well with this new service.

Please pay close attention to how this behaves when user starts the app logged in/logged out/with invalid token. And in cases when user loggs in/signs up/resets the token. I hope I didn't forget about any corner case.

Please also make a functional test when reviewing (tip: change the `TimeSpan.FromMinutes(5)` to something like `TimeSpan.FromSeconds(15)` for faster reviewing :-D ) 

Squash whenever


Toggl.Foundation/Sync/SyncManager.cs
```diff
@@ -56,6 +58,13 @@ public IObservable<SyncState> ForceFullSync()
             }
         }
 
+        public void ForceFullSyncOnSignal(IObservable<Unit> signalSource)
```
willsb: No me gusta having a subscription of outside stuff inside the SyncManager :(

Toggl.Foundation/Services/IBackgroundService.cs
```diff
@@ -1,13 +1,14 @@
...
 
-namespace Toggl.Foundation.MvvmCross.Services
+namespace Toggl.Foundation.Services
 {
     public interface IBackgroundService
     {
         void EnterBackground();
         void EnterForeground();
 
+        IObservable<Unit> AppBecameActive { get; }
```
willsb: I think that's the only observable we need. The method below seems pretty pointless IMO

willsb:
@simonrozsival Could you please update this branch so that it's in netstandard before I review again?

simonrozsival:
I will try to merge it :D

simonrozsival:
@willsb it is now ready for your review.

Toggl.Foundation.MvvmCross/FoundationExtensions.cs
```diff
@@ -73,6 +77,7 @@ public static class FoundationExtensions
             var settingsStorage = new SettingsStorage(self.Version, keyValueStorage);
             var apiErrorHandlingService = new ApiErrorHandlingService(navigationService, settingsStorage);
 
+            Mvx.RegisterSingleton<IBackgroundService>(self.BackgroundService);
```
willsb: I think the generic parameter here can be 🔥

amulware:
@willsb anything specific you'd like me to pay attention to?

willsb:
The changes in the DataSource require more attention

Toggl.Foundation/DataSources/TogglDataSource.cs
```diff
@@ -65,6 +73,21 @@ public sealed class TogglDataSource : ITogglDataSource
...
+
+            if (signalDisposable != null)
+                throw new InvalidOperationException("The StartSyncing method has already been called.");
+
+            signalDisposable = backgroundService.AppBecameActive
+                .Subscribe((Unit _) => SyncManager.ForceFullSync());
+
+            return SyncManager.ForceFullSync()
+                .Select(_ => Unit.Default);
+        }
```
amulware: I'm not sure if we always want to have the subscription to the background service and the first force sync linked like this, but that's something we can change when it's needed.

Toggl.Foundation/DataSources/TogglDataSource.cs
```diff
@@ -96,6 +121,11 @@ private void onSyncError(Exception exception)
             {
                 throw new ArgumentException($"{nameof(TogglDataSource)} could not handle unknown sync error {exception.GetType().FullName}.", exception);
             }
+
+            stopSyncingOnSignal();
```
amulware: This will be called on deprecation or unauthorized only, right? And all other errors here will cause the app to crash with the argument exception above?

If I got that right, I'd prefer this method to be refactored into:

```
if (apiErrorHandlingService.TryHandleDeprecationError(exception) ||
    apiErrorHandlingService.TryHandleUnauthorizedError(exception)
{
    stopSyncingOnSignal();
    return;
}
throw new ArgumentException($"{nameof(TogglDataSource)} could not handle unknown sync error {exception.GetType().FullName}.", exception);
```

I think that would be easier to read.

Toggl.Foundation/Services/BackgroundService.cs
```diff
@@ -0,0 +1,47 @@
...
+
+            var timeInBackground = timeService.CurrentDateTime - lastEnteredBackground.Value;
+            lastEnteredBackground = null;
+            appBecameActiveSubject.OnNext(timeInBackground);
+        }
+
+        public IObservable<Unit> AppBecameActive
+            => appBecameActiveSubject
+                .AsObservable()
+                .Where(actualTimeSpentInBackground => actualTimeSpentInBackground >= timeInBackground)
```
amulware: Are we sure we want to hide this logic inside this service? What if in the future we want to do something every time we become active? I think this implicit behaviour, which is not at all clear from the API of the class, could lead to confusion.

I think I might rather see this being an IObservable<TimeSpan> that emits how long the app was in the background.

Also, I think this properly should probably be called `AppResumedFromBackground` or something else that makes clear that the app is resuming, not starting.

Thoughts?

amulware: Unrelated: This property will create a new observable chain every time it is called. Could/should that not be cached? What are the arguments for doing it this way?

Toggl.Foundation/DataSources/TogglDataSource.cs
simonrozsival: At the moment, we do. At least in all places where we started sync after app start/login/re-login this should also subscribe to the signal. When we want it to behave differently, we can simply change it.

amulware:
Especially the last commit includes some extensive changes. Can you take another look, @willsb?

PR #1153, opened 2017-12-12
Remove business logic from UserDefaultStorage

### LINKED ISSUE #1150: Remove business logic from IAccessRestrictionStorage implementations
Right now the [Android](https://github.com/toggl/mobileapp/blob/develop/Toggl.Giskard/Services/SharedPreferencesAccessRestrictionStorage.cs) and [iOS](https://github.com/toggl/mobileapp/blob/develop/Toggl.Daneel/Services/UserDefaultsStorage.cs) implementations of the settings storage both have some Business logic to mark the revoke the status of `new user` when the app has not been used for too long.

While the storage should remain platform specific (and possibly moved to PrimeRadiant once it's made netstandard), we should move the logic elsewhere.

simonrozsival:
Closes #1150 

Squash whenever

### DIFF
--- a/Toggl.Daneel/Services/UserDefaultsStorage.cs
+++ b/Toggl.Daneel/Services/UserDefaultsStorage.cs
@@ -1,110 +1,22 @@
-﻿using System;
-using Foundation;
-using Toggl.Foundation;
+﻿using Foundation;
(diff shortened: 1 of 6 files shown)


willsb:
Obsolete

PR #1154, opened 2017-12-13
Start suggesting time entries when user has typed in at least 2 characters

simonrozsival:
Ref #315

Squash whenever


willsb:
Did you consider simply to `Throttle`/`Where` on the appropriate places? 

This seems a bit overkill, as this is a restriction of a specific screen.

simonrozsival:
It might be possible to do it with a `.Where(field => field.Text.Length >= minLenght || field.Text[0] == '#' || field.Text[0] == '@')` or something a bit cleaner but with the same semantics, but I am not sure if I like it. I think this is the correct place to implement it, but I am open to discussion.

willsb:
Shouldn't we show the shortcuts empty state while not suggesting anything? Or showing nothing is correct here?

simonrozsival:
I'm not sure what you mean, it shows the empty state with the shortcuts with this chage.

PR #1157, opened 2017-12-13
Fix build configs

amulware:
Currently, the configs are broken in several ways:

- integration tests build Giskard (but not Foundation) without ‘buildAll’ and thus fail locally
- Giskard in general is included in all configs, even though it really doesn’t have to be
- test projects are included in too many configs (like the Debug|IPhone(Simulator) ones) where they don’t need to be
- there are |IPhone and |IPhoneSimulator variants of all configs, which is not necessary

I basically went over everything and deleted as much as I could, and (un)checked as much as appropriate. We want to build as little as possible, but as much as necessary. The only exception is the Debug|Any CPU config which builds everything so we can easily ensure that everything is building.

As a followup, @willsb will submit a PR that adds proper android configs.

The new list of configs is:
screen shot 2017-12-13 at 13 09 29


amulware:
I'm also running integration tests to make sure I didn't break anything: https://www.bitrise.io/build/0533288d575ec09b

amulware:
The integration tests work fine on bitrise and locally for me.

Toggl.Giskard/Toggl.Giskard.csproj
```diff
@@ -17,7 +17,7 @@
     <MonoAndroidResourcePrefix>Resources</MonoAndroidResourcePrefix>
     <MonoAndroidAssetsPrefix>Assets</MonoAndroidAssetsPrefix>
     <AndroidUseLatestPlatformSdk>true</AndroidUseLatestPlatformSdk>
-    <AndroidLinkMode>Full</AndroidLinkMode>
+    <AndroidLinkMode>None</AndroidLinkMode>
```
amulware: @willsb It did this for me, should I change it back?

willsb: Nope. So far it's fine. I need to enable the linker manually as the project matures

amulware:
If you both approve, feel free to 🦑

simonrozsival:
I am not able to build it on Windows, which is a bit unfortunate, as I personally prefer to use VS on Windows, but maybe it shouldn't be that much of a priority when it works on OSX just fine. So please either fix Cake on Windows or let's agree that we don't care about Windows right now, and I will approve this PR.

simonrozsival:
Windows ✅

PR #1161, opened 2017-12-13
Share registration code

### LINKED ISSUE #1150: Remove business logic from IAccessRestrictionStorage implementations

willsb:
Closes #1150 
Supersedes  #1153 

This adds all registration code into a centralised point, allowing us to share all things in both platforms.


Toggl.Foundation.MvvmCross/FoundationExtensions.cs
```diff
@@ -0,0 +1,74 @@
...
+using Toggl.Ultrawave;
+
+namespace Toggl.Foundation.MvvmCross
+{
+    public static class FoundationExtensions
+    {
+        private const int newUserThreshold = 60;
+        private static readonly TimeSpan retryDelayLimit = TimeSpan.FromSeconds(60);
+
+        public static void Initialize(this Foundation self,
```
simonrozsival: Why is this an extension method and not a method of the Foundation class? And why are some objects stored in the container class and passed to the container and some are passed only to the `Initialize` method? What is the rule by which they are passed to one or the other?

Toggl.Foundation.MvvmCross/FoundationExtensions.cs
```diff
@@ -0,0 +1,74 @@
...
+                TogglSyncManager.CreateSyncManager(self.Database, api, dataSource, timeService, retryDelayLimit, scheduler);
+
+            ITogglDataSource createDataSource(ITogglApi api)
+                => new TogglDataSource(self.Database, timeService, apiErrorHandlingService, createSyncManager(api))
+                    .RegisterServices();
+
+            var loginManager = 
+                new LoginManager(self.ApiFactory, self.Database, self.GoogleService, settingsStorage, createDataSource);
+
+            app.Initialize(loginManager, navigationService, settingsStorage);
```
simonrozsival: This method does 3 things: stores some information to the local storage, registers services, and initializes the app. I think this should be split into 3 separate private methods.

Toggl.Foundation.MvvmCross/FoundationExtensions.cs
```diff
@@ -0,0 +1,74 @@
...
+            {
+                var lastUsedDate = DateTimeOffset.Parse(lastUsed);
+                var offset = now - lastUsedDate;
+                if (offset >= TimeSpan.FromDays(newUserThreshold))
+                {
+                    settingsStorage.SetIsNewUser(false);
+                }
+            }
+
+            Mvx.RegisterSingleton(dialogService);
```
simonrozsival: No `Ensure.Argument.IsNotNull`? Why?

Toggl.Foundation.MvvmCross/FoundationExtensions.cs
willsb: Because they are in completely different projects. Plus, we agreed not to take *any* hard dependencies; so this extension method requires what's needed *for MvvmCross* and not for Foundation to work.

PR #1163, opened 2017-12-14
Add Android Resources

willsb:
This adds all of the Droid resources in bulk, to avoid conflicts in the future.

Squash whenever


willsb:
The biggest impact in build times comes from layout files. Drawables are fine :)

PR #1166, opened 2017-12-18
Do not hide suggestions when moving the cursor

simonrozsival:
Ref #315 

These three features are implemented in this PR:

> When user types in description, hits @ (enters projects suggestion mode) and then moves cursor into description (text typed in before), then user is still in projects suggestion mode until he/she starts typing

> When user types in description, hits @ (enters projects suggestion mode), types in something, eg '@abc-project' and then moves cursor into description (text typed in before), then user is still in projects suggestion mode until he/she starts typing

> When user moves the cursor back to the @, he/she can continue projects search ☝️ Applies to both two above mentioned cases.

Squash whenever


Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
```diff
@@ -450,6 +452,19 @@ await dataSource.User.Current
...
         }
 
+        private bool shouldUpdateSuggestions(TextFieldInfo textFieldInfo)
+        {
+            bool shouldUpdate = false; 
+            
+            if (textFieldInfo.Text != previousTextFieldInfo.Text || textFieldInfo.CursorPosition >= previousTextFieldInfo.CursorPosition)
+            {
+                previousTextFieldInfo = TextFieldInfo;
+                shouldUpdate = true;
```
willsb: We tend to prefer early returns, and in this case I think early returning looks much nicer

PR #1167, opened 2017-12-18
Autocomplete requires space before the query symbol

simonrozsival:
Ref #315 

This PR implements this feature:
>When user is typing a description, moves cursor somewhere in the middle of the description text and hits @, then nothing happens 🙀 User can keep typing description.

I applied the same logic for the tags with the ` #` because it makes sense to me as well. If this is the wrong behavior, @pe0ny, I will implement this in a different way.

Squash whenever


pe0ny:
Yes, this is correct and should be implemented to both - projects and tags.

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
```diff
@@ -374,8 +373,15 @@ private void toggleProjectSuggestions()
...
             }
 
-            var newText = TextFieldInfo.Text.Insert(TextFieldInfo.CursorPosition, QuerySymbols.ProjectsString);
-            TextFieldInfo = TextFieldInfo.WithTextAndCursor(newText, TextFieldInfo.CursorPosition + 1);
+            appendSymbol(QuerySymbols.ProjectsString);
+        }
+
+        private void appendSymbol(string symbol)
+        {
+            var shouldAddWhitespace = TextFieldInfo.CursorPosition > 0 && Char.IsWhiteSpace(TextFieldInfo.Text[TextFieldInfo.CursorPosition - 1]) == false;
```
kornelijepetak: `!Char.IsWhiteSpace(TextFieldInfo.Text[TextFieldInfo.CursorPosition - 1])`

simonrozsival: Well, I don't like the use of the exclamation mark - one can easily overlook it. We discussed it with @amulware and there is no preferred way of doing it or at least there is nothing on that subject in the style guide.

Issue #1124, opened 2017-12-08
Chinese keyboard creates both Chinese and Latin characters in start view


amulware:
I don't know enough about international keyboards to give exact steps to reproduce or advise otherwise.

ApocalypticFish:
When using a Chinese Simplified (Pinyin) -I think-  keyboard or Japanese Romaji Keyboard and attempting to type a description in a new entry or editing an existing entry to type characters it is supposed to replace the Latin characters, and does this behaviour in other iOS apps (messenger/messages) but when attempting on the app it creates the Chinese characters and leaves the international/English characters.
To replicate this with the Chinese or Japanese keyboard enabled type an english character then attempt to enter the Japanese or Chinese character from the suggestions bar, it will create the Chinese or Japanese character but won't remove the Latin character from the description field.

amulware:
Possibly related to #1158

simonrozsival:
I tested this with Chinese (simplified) keyboard and it worked the same way as with the Japanese keyboard. There were definitely no Latin characters inserted into the text. I think we can close this for now and reopen it if we get another bug report after this fix is released.

willsb:
Tested this with a Chinese speaker and the problem seems to be gone, indeed

PR #1170, opened 2017-12-18
Do not report the change until it is committed

### LINKED ISSUE #1158: Japanese input bug keyboard and flick input
From Twitter:

>My env is iOS 11.2 on iPhone 6.
>There is flic-input bag in Japanese on new toggl iOS app, I hope it will be fixed as soon as possible.
>For example, when typing with ABC, it is output as AABABC.(Replace the alphabet with Japanese characters)
>As I confirmed it now, the same problem has occurred not only in flick input but also in keyboard input in Japanese characters." 

https://twitter.com/mtane0412/status/940784885521895424/video/1

Assuming it's something like this? https://www.imore.com/how-use-keyboard-flick-ipads-quicktype-keyboard-ios-11

simonrozsival:
Closes #1158, Ref #1124 

This PR needs a functional review from someone, who is familiar with typing in Japanese on a smartphone.


Toggl.Daneel/Binding/TextViewTextInfoTargetBinding.cs
```diff
@@ -138,8 +138,11 @@ public TagDeletedEventArgs(nint cursorPosition, nint tagIndex)
 
             [Export("textViewDidChange:")]
             public void DidChange(UITextView textView)
-                => TextChanged.Raise(this);
-           
+            {
+                if (textView.MarkedTextRange == null)
```
willsb: Please early return when not null of use brackets. This looks confusing.

amulware:
Also please add a comment for the reasoning behind the condition. This may not be obvious anymore in the future.

willsb:
Question: Even though we don't report the text change, isn't the position change being reported somehow? Just asking to make sure we are not doing the calculations anyway because of a `cursor position changed ` event.

simonrozsival:
I will test what happens when the the cursor moves.

simonrozsival:
Changing the cursor while typing => commit.
Typing before committing => no event.

Issue #1074, opened 2017-12-04
Change syncing bar in accordance with designs


amulware:
Check designs on zeplin, and change the bar accordingly.

amulware:
Specifically, when we reach 'offline', the bar should stay and have a dismiss button. Possibly the same for 'failed'?  ping @pe0ny

simonrozsival:
Also I noticed that the "Synced[tick]" seems strange - the icon is too close to the text.

willsb:
It looks weird after #1102 was merged because it relied on the padding added by the initial space when creating attachments. 

We simply need to add that space back at the end of the `Synced` string.

pe0ny:
Offline state with 'try again' and 'dismiss' buttons: https://zpl.io/bPrBM6D
When syncing fails because of no connection, it makes sense to show the same 'Offline' state ☝️ 
But I believe we should have separate 'Failed' state as well when for example backend won't respond? @amulware

amulware:
I'm not sure what you mean with 'won't respond'. Do you mean in case the servers are down, or otherwise entirely unreachable? I'm not sure how much we can detect technically here. I think those cases would currently look the same as 'offline' to us.

simonrozsival:
What happens when the servers are unreachable (but the user is connected to the Internet) is that the app will try repeating it several times which might take a few minutes (2-3) and eventually a `TimeoutException` will be thrown and the user will see a "Failed" msg and the bar will become red. There are no "failed" state designs in Zeplin though.

pe0ny:
@simonrozsival Thanks! That's what I was wondering about if we need 'Failed' state or not. I'll add this to Zeplin as well.

pe0ny:
Failed view: https://zpl.io/VKEYoeW
_I'm not really a big fan of big red error messages, so I would keep it grey but with different text on it._

simonrozsival:
The UI is finished, but the "retry button" and "dismiss button" don't actually have any logic assigned to them. When the user taps on the screen when the syncing bar is visible, it is dismissed (the scroll view resets its position). I wasn't able to prevent this default behavior and implement this explicitly and I will have to consult this with @willsb or @arturs105 so I don't waste too much time searching for a solution.

amulware:
Moving milestone as a bunch of this has been implemented now, and the rest can be done soon(tm).

arturs105:
[...]

Issue #1055, opened 2017-11-29
Tapping on the status bar shows green syncing view


arturs105:
This is probably due to the scrollview scrolling to the very top, so we have to somehow limit it so it doesn't scroll that far.

PR #1183, opened 2017-12-19
Change syncing bar with accordance with the designs

simonrozsival:
Ref #1074

This makes all the elements look as they should according to the designs. The only thing that is missing is that tapping the "Offline" and "Failed" labels doesn't initiate new sync, but rather hides the bar - ~~which is probably due to #1055~~ - I wasn't able to bind the tapping on `SyncStateLabel` (with the Button trait) to the refresh command so far. I think this is a minor issue and could be dealt with later.

Please make a functional test. You probably won't be able to reproduce failure, but the offline case should be in fact the exact same in terms of the UI.

Squash whenever


simonrozsival:
Yeah, that was probably unrelated to this PR. If you find a way how to reproduce it, create a new issue please.

PR #1186, opened 2017-12-20
Fix spacing above and below description in Edit TE view

### LINKED ISSUE #961: Fix margins of the description text field (and placeholder) in Edit TE view
The margins are too small, check Zeplin for the correct offsets.

Zeplin: https://zpl.io/Vxnrl7J

wrong_margins

kornelijepetak:
Fixes #961


kornelijepetak:
It happens to me too, but this is very strange as I have not touched anything but the constraints...

PR #1189, opened 2017-12-20
Keep the exception causing OfflineException

### LINKED ISSUE #1188: Keep the exception causing OfflineException
OfflineException hides all `HttpRequestException`s.

simonrozsival:
Closes #1188 

Squash whenever


amulware:
Oh, should there be a test that the thrown exception is passed on as the inner one?

simonrozsival:
Here comes your test.

Issue #1176, opened 2017-12-19
iPhone X Crashing, Sync slow down/freeze


nicktoggl:
1. Sync happens very slowly even when I'm on my fantastic work network.
I compare to the old app which syncs much faster
2. Because it syncs slowly, I pull the tab to refresh to update it. But
EVERY time I do this action the app crashes and sits there frozen

- Using Verizon iPhone X 256gb
- Running iOS 11.2.1
- I'm also playing music on Spotify through my Airpods if that might
also impact something

case/605107

UPDATE:"If it helps, I just tried on my home network and the app sycned correctly
along with NOT freezing when I pulled down to refresh. I will investigate
when I get to work again tomorrow." firewall/proxy issue/interference?

TheSoloJake:
case/602887 also experiencing freezing issues on iPhone X:

-Has app running in background and switches to it using iOS's App Switcher to start/stop time entries. 
-App randomly freezes/crashes when switching back to the app

nicktoggl:
Another iPhone X user reporting crashing

I downloaded the new app, and it crashes every time I try to use it. Now, I can't log time on the old version or the new version.  I am running iOS 11.2.1 on an iPhone X

case/605115

amulware:
I haven't heard any reports of this kind for some time, so I'll consider this closed. If there are more reports, please feel free to reopen or open a new issue.

PR #1190, opened 2017-12-20
Temporary fix for app freezing

simonrozsival:
Ref #1176 

The bug happened when the user used the app, then switched to a different app, so our app was suspended, and then resumed the Toggl app and started syncing.

This is probably related to the scheduler we were using in the `Timeout`. This PR removes the timeout, so the app works. It is only a temporary fix, because we want to have a timeout there, but one, which doesn't cause more problems than it solves.

Please test in the simulator and on a device several times.

Squash ASAP.


simonrozsival:
The issue is referenced but since this is just a tmp fix we shouldn't close it.

PR #1200, opened 2017-12-21
Fix reports bug with many "No project" slots

### LINKED ISSUE #1198: ReportsProvider memory caching layer doesn't work well
The bug is here:
```
return notInDatabaseIds.Length == 0
                    ? databaseProjectsObservable
                    : databaseProjectsObservable.Merge(searchMemoryAndApi(workspaceId, notInDatabaseIds));
```

This should always emit just ONE value - list of projects, but when there is something missing in the database, it emits multiple lists, which causes a weird looking report:

incorrect_graph

The individual segments are repeated, but shown as "No Project":

incorrect_legend

simonrozsival:
Closes #1198

This merges the lists into a single list. Also fixes a problem when the data from Realm are accessed from a different thread, which caused a crash of the app.

Squash whenever


Toggl.Foundation.Tests/Reports/ReportsProviderTests.cs
```diff
@@ -110,6 +110,43 @@ public void QueriesTheApiIfAnyProjectIsNotFoundInTheDatabase(NonEmptyArray<NonNe
                         calledIds => ensureExpectedIdsAreReturned(calledIds, projectsInApi)));
             }
 
+            [Property(MaxTest = 1)]
```
willsb: On a second thought, you might want to increase this test's `StartSize` and `EndSize`, just in case

PR #1212, opened 2017-12-28
Fix scroll to top

### LINKED ISSUE #1055: Tapping on the status bar shows green syncing view

arturs105:
Closes #1055 

Do a functional test before approving.

Since there are 2 scroll views in the main view (main scroll view and time entries log scroll view), i decided to scroll them both to the top, but animate only the outer scroll view (main scroll view).

Squash whenever.


willsb:
Could you please document what made you change your mind, @kornelijepetak ?

PR #1213, opened 2017-12-28
Show onboarding steps only once if non-skipped

### LINKED ISSUE #855: Show onboarding steps only once if non-skipped
Show onboarding steps only once if non-skipped. Meaning when user already went through all the steps (didn't skip), and logs out, opens the app again, show last onboarding step first (with login and signup options).

arturs105:
Closes #855 

Squash whenever.


Toggl.Foundation.MvvmCross/ViewModels/OnboardingViewModel.cs
```diff
@@ -28,8 +30,11 @@ public sealed class OnboardingViewModel : MvxViewModel
         };
 
         private readonly IMvxNavigationService navigationService;
+        private readonly IOnboardingStorage onboardingStorage;
 
-        private int currentPage = TrackPage;
+        private bool[] pagesVisited = new bool[PageInfo.Length];
```
kornelijepetak: I believe that instead of tracking which pages have been visited to ensure that pages were not skipped, a better solution would be to use `isSkipped` when a user skips the onboarding. And only if you reach the last page with `isSkipped` being `false` would you set the onboarding complete. I guess this works too, so I will approve it, but it seems my suggestion would better convey the purpose. Would also like @willsb  approval. The rest seems great.

willsb: I agree that this is a much nicer approach

kornelijepetak:
Scneario:

* User opens the app
* User skips the onboarding
* User goes back page by page until reaching the first page again (thus having seen all pages)
* User logs in and uses the app
* User logs out
* Next time, onboarding **is not** shown (because user has effectively seen every onboarding page, albeit in reverse order) 

@pe0ny could you please confirm this behaviour? I will merge this now assuming this makes sense. If this needs to be changed, please open a new issue and feel free to assign me.

pe0ny:
@kornelijepetak Yes, it's correct behaviour. 👍

kornelijepetak:
Great then! @willsb please ✔️  and :shipit:

Toggl.Foundation.MvvmCross/ViewModels/OnboardingViewModel.cs
```diff
@@ -91,11 +100,28 @@ public OnboardingViewModel(IMvxNavigationService navigationService)
...
 
+        public override void Prepare()
+        {
+            base.Prepare();
+
+            CurrentPage = onboardingStorage.CompletedOnboarding()
+                ? LoginPage
+                : TrackPage;
+
+            OnCurrentPageChanged();
```
willsb: Is this necessary? Could you test it?

PR #1214, opened 2017-12-29
Switch to Email and Password structs

### LINKED ISSUE #1117: Introduce the Password struct
Similar to the `Email` struct, which is responsible for its own validation etc, we shouldn't be passing strings around when dealing with password, but instead using a Password type.

arturs105:
Closes #1117 


Toggl.PrimeRadiant.Realm/Models/RealmUser.cs
```diff
@@ -24,8 +25,15 @@ public BeginningOfWeek BeginningOfWeek
...
-        public string Email { get; set; }
+
+        [Ignored]
+        public Email Email
+        {
+            get => EmailString.ToEmail();
+            set => EmailString = value.ToString();
+        }
+
+        public string EmailString { get; set; }
```
simonrozsival: This will definitely need a database migration.

simonrozsival: Maybe we could avoid a DB migration by using the `[MapTo("Email")]` attribute here.

Toggl.Multivac/Password.cs
```diff
@@ -0,0 +1,23 @@
...
+        public bool IsValid => password?.Length >= minimumLength;
+
+        private Password(string password)
+        {
+            this.password = password;
+        }
+
+        public override string ToString() => password;
+
+        public static Password FromString(string password)
```
kornelijepetak: I would rather this be called simply `From`. Sounds a bit like a hungarian notation. When *used in a sentence*, I think it reads better.

amulware: I agree with that change.

Toggl.Foundation.MvvmCross/Services/PasswordManagerResult.cs
```diff
@@ -1,13 +1,20 @@
...
     public sealed class PasswordManagerResult
     {
-        public string Email { get; }
+        public Email Email { get; }
+        public Password Password { get; }
 
-        public string Password { get; }
         public static PasswordManagerResult None { get; } = new PasswordManagerResult("", "");
 
         public PasswordManagerResult(string email, string password)
```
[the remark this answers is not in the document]
kornelijepetak: For simplicity, the constructor could remain, but it should be made private. The only instance that uses this constructor should be made to call the `typed` one.

Toggl.PrimeRadiant.Realm/Models/RealmUser.cs
willsb: Just tested `MapTo` locally and it indeed works :)

arturs105:
@willsb Regarding the email disappearing: what if we modify the `Email` struct, so it contains the email string even, when the string is an invalid email (kinda like `Password` does)? Currently, when constructing `Email` with an invalid address, it returns an `Email` struct that has it's internal `email` field empty.

willsb:
Exposing the underlying email, even if invalid, is a possible solution. If you think that's the best approach, I'm okay with it

Toggl.Multivac/Password.cs
```diff
@@ -0,0 +1,23 @@
+﻿namespace Toggl.Multivac
+{
+    public struct Password
+    {
+        private const int minimumLength = 6;
+
+        private readonly string password;
+
+        public bool IsValid => password?.Length >= minimumLength;
```
willsb: I'd rather have this calculated on construction, since this is all immutable.

Toggl.Ultrawave.Tests.Integration/UserApiTests.cs
```diff
@@ -312,5 +322,31 @@ protected override IObservable<IUser> CallUpdateEndpoint(ITogglApi api, IUser en
                 return api.User.Update(entityWithUpdates);
             }
         }
+
+        private sealed class InvalidEmailTestData : IEnumerable<object[]>
+        {
+            private List<object[]> emailStrings
+;
```
willsb: 🤔

Toggl.Ultrawave/Toggl.Ultrawave.csproj
```diff
@@ -8,6 +8,9 @@
   <ItemGroup>
     <PackageReference Include="Newtonsoft.Json" Version="10.0.3" />
   </ItemGroup>
+  <ItemGroup>
```
willsb: This can be undone

Toggl.Giskard/Resources/layout/LoginActivity.axml
```diff
@@ -15,7 +15,7 @@
         android:layout_height="?attr/actionBarSize" />
     <EditText
         android:id="@+id/LoginEmailTextField"
-        local:MvxBind="Text Email; Visibility Visibility(EmailFieldVisible); Focus IsEmailFocused"
+        local:MvxBind="Text Email, Converter=EmailToString; Visibility Visibility(EmailFieldVisible); Focus IsEmailFocused"
```
willsb: I prefer the `EmailToString(Email)` syntax. I think so does @kornelijepetak

arturs105: So it becomes ` local:MvxBind="Text EmailToString(Email);"`?

willsb: Precisely

kornelijepetak: Yes. William is completely and utterly correct about me in this instance. 😄

Issue #1031, opened 2017-11-24
TE creation view: no projects


pe0ny:
- [ ] Show this when user taps @ or project icon below and doesn't have any project
- [ ] When user starts typing, show '+ Create project "new-project-name"', and keep the empty state text visible as well

Zeplin: https://zpl.io/V0PpoEE & https://zpl.io/aXK84xM

![toggl - ios - add new entry first project 1x](https://user-images.githubusercontent.com/5064445/33311888-e32b8ba6-d42e-11e7-9033-261bcff444f0.png) ![toggl - ios - add new entry first project v2 1x](https://user-images.githubusercontent.com/5064445/33311892-e4a13080-d42e-11e7-9cf5-8d11d94476e0.png)

simonrozsival:
Most of the implementation is in the PR #1215, so we should wait until it is merged.

PR #1215, opened 2017-12-29
Show "no tags" message when user has no tags

### LINKED ISSUE #1032: TE creation view: no tags
- [ ] Show this when user taps # or tags icon below and doesn't have any tag
- [ ] When user starts typing, show '+ Create a tag "new-tag-name"', and keep the empty state text visible as well

Zeplin: https://zpl.io/2y1P3eo & https://zpl.io/VKE8lX8

toggl - ios - add new entry first tag 1x toggl - ios - add new entry first tag v2 1x

arturs105:
Closes #1032 

This was made with #1031 in mind, so most of the stuff can be reused for that issue.

Squash whenever and make any requested changes whenever, since i'm on vacation next week.


Toggl.Daneel/ViewSources/StartTimeEntryTableViewSource.cs
```diff
@@ -19,12 +20,22 @@ public sealed class StartTimeEntryTableViewSource : CreateSuggestionGroupedTable
...
         private const string projectCellIdentifier = nameof(ProjectSuggestionViewCell);
+        private const string noEntityInfoCellIdentifier = nameof(NoEntityInfoViewCell);
         private const string emptySuggestionIdentifier = nameof(StartTimeEntryEmptyViewCell);
+        private const string tagIconIdentifier = "icIllustrationTagsSmall";
+
+        private readonly NoEntityInfoMessage noTagsInfoMessage
+            = new NoEntityInfoMessage(
+                text: Resources.NoTagsInfoMessage,
+                imageResource: tagIconIdentifier,
+                imageResourceIndex: 39);
```
simonrozsival: I don't like this very much. When the copy changes, we would need to fix this - also there is no test for this (I am not sure how it would even work) so we might easily overlook this. I would prefer the resource to contain the `#` symbol and to replace that symbol with the image => you would pass the symbol to the `NoEntityInfoMessage` and it would do the string manipulation. This way it would be easy to use it for @ in projects as well.

willsb: Did you just say Platform specific unit tests? ®

https://github.com/toggl/mobileapp/pull/1205

simonrozsival:
Also, I noticed that when you Press the '#' symbol, you can see the "Add keywords..." view, and as you start typing, the "+ Create tag" view pushes it down - but that shouldn't happen according to Zeplin.

Issue #1119, opened 2017-12-07
Timer page UI


pe0ny:
- [x] For the MVP, show only suggestions and log. *Note: Timeline won't be needed for beta release.*
- [x] Show up to 5 suggestions.
- [x] Show suggestions count and number of currently displayed suggestions at the top right corner.
- [x] When there is only 1 suggestion available, hide suggestions count at the top right corner.
- [x] When there aren't any suggestions available, don't show suggestions section at all.
- [x] Show description, project, client and task names (if available) on a time entries.
Use the same format we use for iOS:
```
Description
Project: task Client
```
- [x] Show 'Add description' if there are no description or project available.
- [x] Mark unsynced time entries with grey syncing icon.
- [x] Mark time entries that cannot be synced with a red icon. Do not allow a user to continue these time entries.
- [x] Use fade for long description/project/client/task names both in suggestions and in log.

### Suggestions and log
![timer with out timeline](https://user-images.githubusercontent.com/5064445/33719210-af02b142-db68-11e7-8827-52d07dd83c09.png)

Zeplin: https://zpl.io/VxnpJRX

### Only one suggestion available
![timer with out timeline - one suggestion](https://user-images.githubusercontent.com/5064445/33717978-8eaaa048-db64-11e7-941f-c9ef9130b455.png)

Zeplin: https://zpl.io/2vnj50Y

### No suggestions available

![timer with out timeline - no suggestions](https://user-images.githubusercontent.com/5064445/33718058-d160efa0-db64-11e7-816f-8950901063cd.png)

Zeplin: https://zpl.io/bzwd9LM

willsb:
@pe0ny Shouldn't it say `Working on this?` when there's only one suggestion available?

pe0ny:
@willsb Yes, good point!

willsb:
This can be closed once #1253, #1258 and #1260 are merged

PR #1231, opened 2018-01-04
Droid Time Entries log

willsb:
Ref #1119 

This makes the log appear. yay.


Toggl.Giskard/Converters/BoolToConstantOverrides.cs
```diff
@@ -0,0 +1,11 @@
+﻿using Toggl.Foundation.MvvmCross.Converters;
+
+// Use this file to register overrides of BoolToConstantValueConverter.
+// They are all single line classes, but an override is needed to avoid using the Fluent syntax.
+namespace Toggl.Giskard.Converters
+{
+    public sealed class LogDescriptionTopOffsetValueConverter : BoolToConstantValueConverter<int>
+    {
+        public LogDescriptionTopOffsetValueConverter() : base(16, 27) { }
```
kornelijepetak: I am sad that we need to do it this way. I don't like this hard-coded at this place. I assume you did it because Tibet binding does not allow passing parameters to the constructor, and are we not able to utilize the `parameter` to specify this in any elegant way?

I guess handling margins in `Convert` while having to write it like this would be disgusting:
`MvxBind="TopMargin LogDescriptionTopOffset(HasProject, &quot;16,27&quot;)"`

willsb: > I assume you did it because Tibet binding does not allow passing parameters to the constructor, and are we not able to utilize the parameter to specify this in any elegant way?

Precisely

> I guess handling margins in Convert while having to write it like this would be disgusting:

We would still have to write a custom ValueCombiner. And it would be even worse, since it would be a bitch to cast the values accordingly (i.e.: There's the `If` value combiner, which doesn't suit our needs for this exact reason).

Issue #1056, opened 2017-11-29
Reports UI: loading state


pe0ny:
- Show this state while loading data
Zeplin: https://zpl.io/bzwZzME

![toggl - ios - reports loading state 1x](https://user-images.githubusercontent.com/5064445/33381854-d6204632-d527-11e7-8b5d-3a6f0b682cda.png)

- Colours that should change: #ECF0F2 and #FAFBFC
- Duration for one full circle: 4s
- Curve: Linear *check #172*

![ezgif-2-438229309a](https://user-images.githubusercontent.com/5064445/33381664-4a0dfec8-d527-11e7-9592-bce675810178.gif)

amulware:
What kind of transition is there when the data gets there @pe0ny? A simple fade?

amulware:
And if the user selects a different date range, does it go back to this loading state?

pe0ny:
> What kind of transition is there when the data gets there @pe0ny? A simple fade?
Yes, let's use just fade for now.
For future reference, it would be super nice to use the same animation but faster (duration 2s, instead of 4s). So when data is ready to display, the loader will move on and show actual data/ colours, not grey area anymore.

> And if the user selects a different date range, does it go back to this loading state?
Yes.

simonrozsival:
This issue is almost finished - what is missing is the fade transition.

PR #1233, opened 2018-01-04
Reports loading state - animated mock pie chart

simonrozsival:
Ref #1056 

This PR adds a spinning mock pie chart while the reports are loading. This does not completely close the issue because the fade-in transition isn't part of this PR and will be added later.

Squash whenever


Toggl.Daneel/Views/Reports/LoadingPieChartView.cs
```diff
@@ -0,0 +1,103 @@
...
+using UIKit;
+using static Toggl.Foundation.MvvmCross.Helper.Animation;
+
+namespace Toggl.Daneel.Views.Reports
+{
+    [Register(nameof(LoadingPieChartView))]
+    public sealed class LoadingPieChartView : UIView
+    {
+        private static readonly nfloat pi = (nfloat)(Math.PI);
+        private static readonly nfloat fullCircle = 2 * pi;
```
willsb: I'm starting to think that maybe we should have a common place with all those math constants, since we are starting to use them all the time. IMO it's better than defining them as private constants everywhere. OFC this is not in this PR's scope, but what do you think?

Toggl.Daneel/Views/Reports/LoadingPieChartView.cs
```diff
@@ -0,0 +1,103 @@
...
+        private static readonly nfloat startAngle = -0.5f * pi;
+        private const double fps = 60.0;
+
+        private CGColor lightColor = Color.Reports.Loading.LightColor.ToNativeColor().CGColor;
```
willsb: This and `darkColor` can be made `readonly`

Toggl.Daneel/Views/Reports/ReportsHeaderView.cs
```diff
@@ -49,6 +51,11 @@ public override void AwakeFromNib()
                           .To(vm => vm.TotalTime)
                           .WithConversion(new TimeSpanReportLabelValueConverter());
 
+                //Loading chart
+                bindingSet.Bind(LoadingPieChartView)
```
willsb: I assume that this is in front of the other pie chart and bound to have the same center/width/height/aspect, correct?

Toggl.Daneel/Views/Reports/LoadingPieChartView.cs
```diff
@@ -0,0 +1,103 @@
...
+            animation.TimingFunction = Curves.Linear.ToMediaTimingFunction();
+            animation.CalculationMode = CAAnimation.AnimationLinear;
+            animation.FillMode = CAFillMode.Forwards;
+            animation.RemovedOnCompletion = false;
+            animation.AnimationStopped += (sender, e) =>
+            {
+                (foregroundLayer, backgroundLayer) = (backgroundLayer, foregroundLayer);
+                foregroundLayer.RemoveFromSuperLayer();
+                Layer.AddSublayer(foregroundLayer);
+                foregroundLayer.AddAnimation(animation, null);
```
willsb: Can you please put a breakpoint here to verify that this is not leaking when you leave the reports view?

Toggl.Daneel/Views/Reports/LoadingPieChartView.cs
simonrozsival: I agree that having constants in one place would be better than having them scattered all over the place. Please make an issue so we don't forget about that!

PR #1238, opened 2018-01-05
Suggestions Fragment

willsb:
Ref #1119 

This adds basic functionality for the fragment. The scrolling will be fixed in a different PR; This one just puts things there.

Squash at Will


willsb:
<3

Issue #1223, opened 2018-01-02
Improve login screen invalid data feedback


amulware:
Here are a couple of pain points of mine:

- when no valid email is entered or the password field is empty, the 'next' button is grayed out, but the effect is so subtle I can only tell because I know. This could confuse users if they try tapping the button
- when I hit [return] on the keyboard with no valid password entered, the keyboard closes, but nothing else happens
- when I hit return with an empty password field, the keyboard closes

I think we should make it clearer to the user that a valid email is required, and definitely not close the keyboard. Also if possible, we should change the return-button text to 'next' or similar.

Maybe if the user tries to hit next/return when there is no password/an invalid email, we should show a message stating so under the text field? "Please enter a valid email address." "Please enter your password."(login)/"Please enter a password."(signup)

willsb:
Ref #1021

ApocalypticFish:
Not sure if this needs to be changed for the field on each page, but this occurs (or doesn't occur) on signup, login and forgot your password pages

amulware:
We probably have to rethink this with the upcoming changes to the login/signup screens anyways. Still leaving it open so that we take another look at the UX in the future.

amulware:
We should also show a specific error when we detect the user being offline.

amulware:
Shall be superseded given we have an entirely new design.

Issue #1021, opened 2017-11-23
Add validation to password length during sign up


willsb:
Follow up #1018, we need to disable sign up while the password is too short and clear the error message after the length is big enough. 

This needs further clarification from @pe0ny

pe0ny:
@willsb Can you clarify which one do you mean - sign up or log in? (since the name says 'sign up' and description 'log in')

willsb:
Sign up

pe0ny:
Approved from my side:

- [ ] Show 'Password should be at least 6 characters.' until user has entered a password that is at least 6 char. Disable 'Next' button (opacity 50%)
- [ ] When user can entered a password that is at least 6 char, remove 'Password should be at least 6 characters.' and show checkmark on the right. Enable and change 'Next' opacity to 100%

![toggl - ios - sign up step2 1x](https://user-images.githubusercontent.com/5064445/33213627-03ab69b0-d131-11e7-8e5b-a2e709a6e821.png) ![toggl - ios - sign up step2 confirmed 1x](https://user-images.githubusercontent.com/5064445/33213628-063a4106-d131-11e7-8c5e-9e2f49ba28c4.png)

willsb:
What about the "View password" button? I think it add much more value than this checkmark (after all, the error disappearing and the next being enabled should be enough evidence that the password is valid).

pe0ny:
I need coffee or something stronger...

simonrozsival:
And when user deletes some of the characters and it goes back to < 6, it should re-appear?

What happens when the API call fails and the error message is shown in the label and user changes the Password field - should the message appear again?

pe0ny:
> And when user deletes some of the characters and it goes back to < 6, it should re-appear?
yes

> What happens when the API call fails and the error message is shown in the label and user changes the Password field - should the message appear again?
you mean when user taps 'next' and then api call fails > we say 'Sorry, we’re having trouble signing you up. Please try again.' > and then user deletes some char > then yes, we show again 'password should be longer...'

willsb:
While at it, please try and check #1117

PR #1241, opened 2018-01-05
Do not close keyboard when return is hit and the password is invalid in Signup

simonrozsival:
Ref #1223

This PR addresses some of the points described in the issue. Namely, it doesn't close the keyboard if the input isn't valid and it changes the text of the return button to 'Next'.

The downside is that now we can't easily react to pressing the return key when the input isn't valid - we would have to add another binding. This shouldn't be a problem though I think - when #1021 is closed, we shouldn't need any other help messages in this view IMO.

Squash whenever


willsb:
This looks good and I even functional tested it, but does it close the issue or not? The description makes the PR's intentions quite ambiguous

simonrozsival:
I am convinced it closes the issue. The description isn't well written... also the issue isn't very specific. Maybe we could just remove the `Closes` from the PR for now and maybe close it later if we consider it OK this way. I think it is now ready to be merged, @willsb.

PR #1243, opened 2018-01-05
Hide suggestions when none is available

willsb:
Ref #1119 

Squash whenever


Toggl.Giskard/Fragments/SuggestionsFragment.cs
```diff
@@ -1,20 +1,44 @@
...
+    public sealed class SuggestionsFragment : MvxFragment<SuggestionsViewModel>
     {
+        private int containerHeightInPixels;
+
         public override View OnCreateView(LayoutInflater inflater, ViewGroup container, Bundle savedInstanceState)
         {
             base.OnCreateView(inflater, container, savedInstanceState);
             var view = this.BindingInflate(Resource.Layout.SuggestionsFragment, null);
+
+            containerHeightInPixels = (int)130.DpToPixels(Activity);
```
kornelijepetak: Would you mind moving this to a constant?

willsb: Not possible. We need a reference to a Context because the size in Pixel changes based on the screen size. Unless you mean the 130, which IMO will make things less readable, but can be done

kornelijepetak: I thought the only thing worse than singletons are magic constants 😆 
I was talking about `130`, but if you don't want to change that for one reason or another, just tell me. This can be approved in that case. I don't see why it would make things less readable, but I agree that this line is actually an assignment of a constant (plus conversion), so this is almost as if defining a constant here, so... up to you.

PR #1211, opened 2017-12-28
Add calendar tests

arturs105:
This adds tests that were missing from #1174 

Squash whenever.

**Edit:** since i'm on vacation next week, I hereby give permission to @willsb, @simonrozsival and @kornelijepetak to make all requested changes and merge this PR.

PR #1248, opened 2018-01-08
Fix reports "This year" shortcut

### LINKED ISSUE #1245: Selecting "This year" in reports calendar makes the app crash
Reports calendar tries to mark the days of the whole year as selected - but it tries to mark also the days of months which are in the future and which aren't shown in our calendar and this incorrect indexing causes a crash.

simonrozsival:
Closes #1245, Ref #1211 

This PR fixes a 🐛 which was caused by invalid indexing. This code should be more declarative and easier to read. Also, there are now two property tests to ensure that the app won't crash.

Squash whenever


Toggl.Foundation.MvvmCross/ViewModels/Calendar/CalendarDayViewModel.cs
```diff
@@ -15,30 +22,27 @@ public sealed class CalendarDayViewModel : INotifyPropertyChanged
...
-        public DateTimeOffset ToDateTimeOffset()
-            => new DateTimeOffset(
-                CalendarMonth.Year,
-                CalendarMonth.Month,
-                Day,
-                0,
-                0,
-                0,
-                TimeSpan.Zero);
+        public DateTimeOffset ToDateTimeOffset() => dateTime;
```
arturs105: This method could be changed to a property, since it's basically a getter now.

PR #1252, opened 2018-01-08
Main page scrolling

willsb:
Ref #1119 

This makes everything scroll in sync. Android can be nice sometimes.
This also adds a little padding at the end of the RecyclerView so that the last item is always visible behind the play button, similar to what happens in iOS.

Squash whenever


willsb:
If we used an enum, we'd have to cast. I'd rather not

PR #1253, opened 2018-01-08
Continue Time entries

### LINKED ISSUE #1119: Timer page UI

willsb:
Closes #1119 

Swipe to continue works too :heart_eyes:

The only reason why there’s no swipe to delete is because that’d require ViewModel infrastructure that wouldn’t even get used in iOS until much later.

Squash whenever


Toggl.Giskard/Resources/layout/TimeEntriesLogFragmentCell.axml
```diff
@@ -100,6 +100,13 @@
         android:layout_marginBottom="10dp"
         android:layout_alignParentBottom="true"
         android:layout_alignParentRight="true" />
+    <Button
+        local:MvxBind="Visibility Visibility(CanSync)"
+        android:id="@+id/TimeEntriesLogCellContinueButton"
```
kornelijepetak: I'd still like to see ID first, even if longer than the MvxBind. It's its identity.
No need to wait for Bitrise to retest everything so I will now approve it anyway as it is not really that important.

willsb: I'm really indifferent, so we can go through these for consistency later and have the id above 👌

PR #1254, opened 2018-01-08
Limit number of suggestions in Droid

willsb:
Ref #1119 

Squash whenever


Toggl.Daneel/Startup/Setup.cs
```diff
@@ -64,7 +66,8 @@ protected override void InitializeApp(IMvxPluginManager pluginManager, IMvxAppli
                 environment
             );
 
-            foundation.RegisterServices(new DialogService((ITopViewControllerProvider)Presenter),
+            foundation.RegisterServices(maxNumberOfSuggestions,
```
amulware: That's not really a service though, is it?

willsb: It's part of the service registration, however. As time passes, the creation of the suggestions will have to be split (since a bunch of different providers will be registered by each platform), but until then I'd like to keep it all in the same place.

Would you rather me split the construction of the provider container right now?

amulware: Leave it as it is for now then. Let's just make sure it'll turn out nice when we touch it next.

PR #1256, opened 2018-01-08
Adjust suggestion sizes and paddings

willsb:
Ref #1119 

This:

- Adds paddings to the suggestions
- Makes the cards snap
- Makes the card big if there’s only one suggestion

Squash whenever


kornelijepetak:
I don't know if that's just the different device screen size, but the suggestion title went below the toolbar...

image

kornelijepetak:
Also I noticed that the log lags when I scroll through it (and I have like 20 entries at most). My device is rather powerful, so it's serious. (Although I am not sure if this is the PR that introduced this problem). I will check and report.

Toggl.Giskard/Views/SuggestionsRecyclerViewSnapHelper.cs
```diff
@@ -0,0 +1,41 @@
...
+
+        private OrientationHelper getHorizontalHelper(RecyclerView.LayoutManager layoutManager)
+            => horizontalHelper ?? (horizontalHelper = OrientationHelper.CreateHorizontalHelper(layoutManager));
+
+        private View getStartView(RecyclerView.LayoutManager layoutManager, OrientationHelper helper)
+        {
+            var linearLayoutManager = (LinearLayoutManager)layoutManager;
+
+            var firstChildIndex = linearLayoutManager.FindFirstVisibleItemPosition();
+            if (firstChildIndex == RecyclerView.NoPosition) return null;
```
kornelijepetak: From formatting guidelines:
> Return early in single line, if the condition is short **and there is no return value.**

Toggl.Giskard/Views/SuggestionsRecyclerViewSnapHelper.cs
```diff
@@ -0,0 +1,41 @@
...
+
+            var lastItemIndex = linearLayoutManager.ItemCount - 1;
+            var isLastItem = linearLayoutManager.FindLastCompletelyVisibleItemPosition() == lastItemIndex;
+            if (isLastItem) return null;
```
kornelijepetak: same here.

Issue #1053, opened 2017-11-29
Duration and start/end times edit view


pe0ny:
- [x] This view is for both: editing start/end times and duration at the same time.
- [x] The wheel is always draggable from both ends in order to change the start and stop times. Note that duration changes as well when dragging either start or stop handler.
- [x] When user starts dragging in the dark grey area of the circle (the area between two handles), then it moves both start and end times but duration remains the same.
- [x] For entries longer than 12h, user can spin the circle over itself over and over again. In that case, we can allow 999h duration, too.
*When an entry is more than 12 hours long, then both handles are still draggable. A user can spin full circles forward and backwards to increase or decrease a duration.*
- [x] Both start and end times are tappable and will open default iOS date-time picker. The content below will be pushed out as the picker appears.
- [ ] User can scroll the card's content: Only picker and the wheel will scroll.
- [x] To close a picker, user have to tap again on a start/end time.
- [x] Changing the picker, updates the wheel and vice versa.
- [x] While editing start/stop time/date by using the picker, automatically updates date and time above as well.
- [x] When user is editing stopped time entries, allow editing both start and stop times.
- [x] When user is editing currently running time entry, display 'Set' in end time field.
- [x] When user is editing currently running time entry, show ticking timer inside the circle.
- [x] When user is editing currently running time entry, do not allow dragging end time handle or the dark grey area of the circle.
- [x] When user taps 'Set', set end time for that time entry. End time should be the current time.
- [x] Include vibration feedback in the wheel for every 5 mins.
- [x] When a time entry is so short that two circles are overlapping, then start time handler should be on top.

### 24hour and 12hour clock examples when editing already stopped time entry
![toggl - ios - start-end times 24hour 1x](https://user-images.githubusercontent.com/5064445/34677939-5ea88f28-f49a-11e7-95d5-466d186702ea.png) ![toggl - ios - start-end times 12hour 1x](https://user-images.githubusercontent.com/5064445/34677944-6171aa46-f49a-11e7-9573-6bc2c8f19db1.png)

Zeplin: https://zpl.io/aBw0d8L & https://zpl.io/bWQxmEM

### Editing running time entry
[...]

PR #1259, opened 2018-01-09
Add date time pickers logic to the Wheel view

simonrozsival:
Ref #1053 

This PR adds date picker component to the view and enables editing the start and stop time of the time entry.

~~WIP: the bindings of the minimum and maximum time constraints for the UIDatePicker don't work at the moment.~~

Squash whenever


simonrozsival:
The date validation issue is now fixed.

PR #1261, opened 2018-01-09
Add no projects info message

### LINKED ISSUE #1031: TE creation view: no projects

arturs105:
Closes #1031


Toggl.Daneel/ViewSources/StartTimeEntryTableViewSource.cs
```diff
@@ -184,5 +195,16 @@ protected override object GetCreateSuggestionItem()
...
+
+        private NoEntityInfoMessage GetNoEntityInfoMessage()
+        {
+            if (ShouldShowNoTagsInfoMessage)
+                return noTagsInfoMessage;
+
+            if (ShouldShowNoProjectsInfoMessage)
+                return noProjectsInfoMessge;
+
+            throw new InvalidOperationException("This method should not be called, when there is no info message to be shown");
```
arturs105: I'm not sure, whether this should throw. Would it be better to make this method something like `bool TryGetNoEntityInfoMessage(out NoEntityInfoMessage stuff)`?

Toggl.Daneel/ViewSources/StartTimeEntryTableViewSource.cs
```diff
@@ -25,19 +25,27 @@ public sealed class StartTimeEntryTableViewSource : CreateSuggestionGroupedTable
...
         private readonly NoEntityInfoMessage noTagsInfoMessage
             = new NoEntityInfoMessage(
                 text: Resources.NoTagsInfoMessage,
                 imageResource: tagIconIdentifier,
                 characterToReplace: '#');
+        private readonly NoEntityInfoMessage noProjectsInfoMessge
+            = new NoEntityInfoMessage(
+                text: Resources.NoProjectsInfoMessage,
+                imageResource: projectIconIdentifier,
+                characterToReplace: '#');
```
simonrozsival: We could use '@' for projects.

Toggl.Daneel/ViewSources/StartTimeEntryTableViewSource.cs
```diff
@@ -184,5 +195,16 @@ protected override object GetCreateSuggestionItem()
             => IsSuggestingProjects
                 ? $"{Resources.CreateProject} \"{Text}\""
                 : $"{Resources.CreateTag} \"{Text}\"";
+
+        private NoEntityInfoMessage GetNoEntityInfoMessage()
```
simonrozsival: Capital G

Toggl.Daneel/ViewSources/StartTimeEntryTableViewSource.cs
simonrozsival: This should never happen in production, so I think throwing is OK. If it ever happened, we would be able to see it in HockeyApp or something. Is there a way how to test that this never throws?

Toggl.Foundation/Resources.resx
```diff
@@ -247,4 +247,8 @@ time entries will appear here.</value>
         <value>Add keywords to your
 time entries with # tags.</value>
     </data>
+    <data name="NoProjectsInfoMessage">
+        <value>Create a project # and
```
simonrozsival: \# -> @

Toggl.Daneel/ViewSources/StartTimeEntryTableViewSource.cs
arturs105: I don't think there is an easy way to test this, since:
1. We are not testing table view sources
2. This method is private
3. The view model is involved in setting `ShouldShowNoProjectsInfoMessage` and `ShouldShowNoTagsInfoMessage`, so that adds extra complexity
4. It would be really hard to simulate real life behaviour of a table view source (which method gets called when)

simonrozsival: What does @willsb think?

willsb: What @arturs105 said. If this was in a more reliable point of the project, I'd be fine with having it throw. Those sources are not really under our control and behaviour is weird.

Now for this specific case it doesn't look like we can ever get into the throw state because that specific cell shouldn't be returned otherwise. I say we keep it throwing and observe its behaviour

PR #1263, opened 2018-01-10
Make calendar day headers adapt to different beginnings of week

### LINKED ISSUE #1203: Report calendar displaying wrong date
"Today, it's Friday, Dec 22 while the app calendar (in the date picker of reports) thinks it the 21...
<see screenshots attached>"

I can replicate but it appears correct on Jordan's and Rico's

image1 2

case/606543

arturs105:
Closes #1203 

I believe the issue was day headers because:
1. In the #1203 @simonrozsival pointed out that, the calendar did not (at that time) mark the current day as selected.
2. In the screenshot you can see, that the calendar is displaying correct dates for December, when Sunday is the start of the week, but the day headers are static and start with Monday.


Toggl.Daneel/ViewControllers/ReportsCalendarViewController.cs
```diff
@@ -18,6 +19,17 @@ public partial class ReportsCalendarViewController : MvxViewController<ReportsCa
         private const int estimatedQuickSelectShortcutHeight = 32;
         private const int estimatedQuickSelectShortcutWidth = 90;
 
+        private readonly string[] dayHeaders =
+        {
+            Resources.S,
```
willsb: Nope. `Resources.SundayInitial`, so this can be easily translatable

PR #1279, opened 2018-01-11
The Wheel

### LINKED ISSUE #1240: Refactor common mathematical constants into a class
As discussed [here](https://github.com/toggl/mobileapp/pull/1233#discussion_r159694217), we are repeating a lot of mathematical constants in different places as private constants. 

We should instead create a `Math` class in Multivac that we can simply `import static` for simplicity sake.

simonrozsival:
Ref #1053, closes #1240 

This PR implements the wheel rendering & touch functionality 🎉 

~~WIP: I would like to add more unit tests to the math functions, but as this PR does a lot of things and will take a long time to review, you can start reviewing the rendering and the touch input processing.~~

I suggest making several reviews to make it easier and less scary for the person reviewing, because doing a single review for everything should be quite scary for anyone:

1. `WheelBackgroundView` - how the clock background is built and displayed
2. `WheelForegroundView` - how the foreground UI is render
3. `WheelForegroundView` - how the touch events are handled
4. `Math` - how the math functions are implemented in Multivac
5. `EditDurationViewController` + `EditDurationViewModel` - how is it all put together


arturs105:
1. The dragging is quite "jumpy", and the scroll view interferes with the wheel as you can see in the gif.
2. The clock times look pixelated/low res (can be seen in the gif as well)
3. Can we make the iOS date picker appear/disappear with animation (when tapping on start/end time)?

jan-12-2018 12-21-59

simonrozsival:
1. I broke the angles in the last commit - I am aware of that (it is now probably rotated by 90 degrees), but I am currently not able to run the app in simulator to debug it. I will fix this soon. That is also why it scrolls instead of dragging sometimes - try commit `09a2d97` before I fix this.
2. The numbers are weird indeed. I will look into it more in-depth.
3. I would like to add an animation, but probably in a different PR.

simonrozsival:
I reverted the last two commits which broke the UI and I will return them once I fix the issue you pointed out, @arturs105.

arturs105:
Did not have time to do a full review, but found these things in a functional test.
1. The iOS date barrel thingy jumps up/down when switching to and from start time and end time editing. 
jan-22-2018 16-55-11
2. The z axis of draggable handles (the white circles) could be reversed. Currently the start time handle is "on top" of end time, which leads to little weird editing (see gif). Also, when dragging the end handle around the wheel multiple times, it goes under the start handle. 
jan-22-2018 17-00-35
3. Dragging the whole segment is kinda broken/sensitive (see gif). I think this is a really good feature that we should have (when works properly), but we could remove it for now for faster release of the wheel.
jan-22-2018 17-03-01

All of these things are more or less minor, whether they should be fixed is up to @pe0ny and @simonrozsival 

Code review coming tomorrow.

simonrozsival:
1. I will look into the binding to see if I can get rid of this
2. This is the indented behavior, I believe (start should be always on top - see the issue)
3. This is broken, I will look into it (probably something to do with changing this from 12 hour wheel to 1 hour wheel)

simonrozsival:
I addressed all comments.

Toggl.Daneel/Views/EditDuration/WheelBackgroundView.cs
```diff
@@ -0,0 +1,132 @@
...
+
+            return layer;
+        }
+
+        private CALayer createClockDial()
+        {
+            var dial = new CALayer();
+            var minuteSegmentsPerHourMark = MinutesInAnHour / HoursOnTheClock;
+
+            var scaledThickSegment = scaleRect(this.thickSegmentDimensions);
```
arturs105: `this.` not necessary

Toggl.Daneel/Binding/ViewAnimatedVisibilityTargetBinding.cs
```diff
@@ -21,6 +21,8 @@ public ViewAnimatedVisibilityTargetBinding(UIView target)
 
         protected override void SetValue(bool value)
         {
+            if (Target.Hidden != value) return;
```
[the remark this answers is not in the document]
simonrozsival: Without this the datepicker jumped up and down, as you noticed earlier. This doesn't run the animation if the state doesn't change.

Toggl.Daneel/Extensions/PointExtensions.cs
```diff
@@ -0,0 +1,14 @@
...
+using Toggl.Multivac;
+
+namespace Toggl.Daneel.Extensions
+{
+    public static class PointExtensions
+    {
+        public static CGPoint ToCGPoint(this Point point)
+            => new CGPoint(point.X, point.Y);
+
+        public static Point ToMultivacPoint(this CGPoint point)
```
kornelijepetak: The class is named `Point` and not `MultivacPoint` so I am not sure if this method should be named like this. I don't think it will introduce any confusion if it was just `ToPoint`.

Toggl.Daneel/Views/EditDuration/BaseWheelView.cs
```diff
@@ -0,0 +1,74 @@
...
+
+        protected CGAffineTransform CreateTranslationTransform(nfloat r, nfloat angle)
+        {
+            var tx = r * (nfloat)Math.Sin(angle);
+            var ty = r * (nfloat)Math.Cos(angle);
+            return CGAffineTransform.MakeTranslation(Center.X + tx, Center.Y - ty);
+        }
+
+        protected nfloat Resize(nfloat originalSize)
+            => originalSize * (Radius / 128f);
```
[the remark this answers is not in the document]
simonrozsival: The original sizes are taken from Zeplin and there the radius is 128 pt. I should probably normalize all the constants so it should be only multiplied by the `Radius` value.

Toggl.Daneel/ViewControllers/EditDurationViewController.cs
```diff
@@ -122,16 +122,49 @@ public override void ViewDidLoad()
...
+            bindingSet.Bind(WheelView)
+                      .For(v => v.MinimumEndTime)
+                      .To(vm => vm.MinimumStopDate);
+
+            bindingSet.Bind(WheelView)
+                      .For(v => v.StartTime)
+                      .To(vm => vm.StartTime);
+
+            bindingSet.Bind(WheelView)
+                      .For(v => v.EndTime)
```
kornelijepetak: `EndTime` vs `StopTime`. Can this be streamlined?

Toggl.Daneel/Views/EditDuration/BaseWheelView.cs
```diff
@@ -0,0 +1,74 @@
...
+
+namespace Toggl.Daneel.Views.EditDuration
+{
+    public abstract class BaseWheelView : UIView
+    {
+        protected new CGPoint Center;
+
+        protected nfloat Radius;
+
+        protected nfloat SmallRadius;
```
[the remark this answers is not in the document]
kornelijepetak: ☝️

Toggl.Daneel/ViewControllers/EditDurationViewController.cs
```diff
@@ -111,27 +111,60 @@ public override void ViewDidLoad()
...
+            bindingSet.Bind(WheelView)
+                      .For(v => v.MaximumStartTime)
+                      .To(vm => vm.MaximumStartTime);
+
+            bindingSet.Bind(WheelView)
+                      .For(v => v.MinimumStartTime)
+                      .To(vm => vm.MinimumStartTime);
+
+            bindingSet.Bind(WheelView)
+                      .For(v => v.MaximumEndTime)
```
kornelijepetak: `EndTime` vs `StopTime`. Can this be streamlined?

Toggl.Daneel/Views/EditDuration/BaseWheelView.cs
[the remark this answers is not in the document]
kornelijepetak: Yes, please do that.

Toggl.Daneel/Views/EditDuration/WheelBackgroundView.cs
```diff
@@ -0,0 +1,132 @@
...
+using CoreGraphics;
+using static Toggl.Multivac.Math;
+using Foundation;
+
+namespace Toggl.Daneel.Views.EditDuration
+{
+    [Register(nameof(WheelBackgroundView))]
+    public sealed class WheelBackgroundView : BaseWheelView
+    {
+        private readonly CGColor wheelColor = Color.EditDuration.Wheel.Background.ToNativeColor().CGColor;
```
kornelijepetak: I'd prefer this to be called `wheelBackgroundColor` because that's what it is. Wheel is a composition of all its parts, and this is only the color of its background.

Toggl.Daneel/Views/EditDuration/WheelBackgroundView.cs
```diff
@@ -0,0 +1,132 @@
...
+
+        private readonly CGColor thickSegmentColor = Color.EditDuration.Wheel.ThickMinuteSegment.ToNativeColor().CGColor;
+
+        private readonly CGColor thinSegmentColor = Color.EditDuration.Wheel.ThinMinuteSegment.ToNativeColor().CGColor;
+
+        private readonly CGSize thickSegmentDimensions = new CGSize(1.7f, 5f);
```
kornelijepetak: Not a thing that needs to be changed, but why not `Size` instead of `Dimensions`?

Toggl.Multivac/Math.cs
```diff
@@ -0,0 +1,54 @@
+﻿using System;
+
+namespace Toggl.Multivac
+{
+    public static class Math
+    {
+        public const float QuaterOfCircle = 0.5f * (float)System.Math.PI;
```
kornelijepetak: `Quarter`

Toggl.Multivac/Math.cs
```diff
@@ -0,0 +1,54 @@
...
+
+        public const float FullCircle = 2f * (float)System.Math.PI;
+
+        public const int HoursOnTheClock = 12;
+
+        public const int MinutesInAnHour = 60;
+
+        public const int SecondsInAMinute = 60;
+
+        public static double TimeToAngleOnTheDial(TimeSpan time)
```
kornelijepetak: I don't think `Time` is necessary at the start of this method name.

Toggl.Multivac/Math.cs
```diff
@@ -0,0 +1,54 @@
...
+            => TimeToAngle(time) - QuaterOfCircle;
+
+        public static double TimeToAngle(TimeSpan time)
+            => (time.Minutes * SecondsInAMinute + time.Seconds) / (double)(MinutesInAnHour * SecondsInAMinute) * FullCircle;
+
+        public static double PositiveAngle(double angle)
+        {
+            while (angle < 0) angle += FullCircle;
```
kornelijepetak: Maybe I got something wrong, but does angle ever go below -2*Pi?

Toggl.Multivac/Math.cs
```diff
@@ -0,0 +1,54 @@
...
+        public const int HoursOnTheClock = 12;
+
+        public const int MinutesInAnHour = 60;
+
+        public const int SecondsInAMinute = 60;
+
+        public static double TimeToAngleOnTheDial(TimeSpan time)
+            => TimeToAngle(time) - QuaterOfCircle;
+
+        public static double TimeToAngle(TimeSpan time)
```
kornelijepetak: Since this is a static class, I think extension methods would be much nicer here.

Also, then this:
```cs
private double startTimeAngle => PositiveAngle(TimeToAngleOnTheDial(startTime.LocalDateTime.TimeOfDay));
```
could be written as this:
```cs
private double startTimeAngle => startTime.LocalDateTime.TimeOfDay.ToAngleOnTheDial().ToPositiveAngle(
```

Toggl.Daneel/ViewControllers/EditDurationViewController.cs
simonrozsival: I don't think so. I gave it a thought when I started doing the wheel and there isn't really a stop date, there is an end date (which might be changing for a running time entry), but it is not a stop time. Maybe we could rename it in the VM too, but that would be a different PR.

Toggl.Daneel/ViewControllers/EditDurationViewController.cs
simonrozsival: Same as below.

Toggl.Daneel/Extensions/PointExtensions.cs
simonrozsival: There are a lot of points out there in the wild :-D This one is in fact `Toggl.Mutlivac.Point`, so I think the naming isn't that wrong, although I agree that it is not even close to ideal.

Toggl.Daneel/ViewControllers/EditDurationViewController.cs
kornelijepetak: Yeah, I think we should change the VM. As you suggested, in another PR. @amulware?
Cause in reality, we use the term "End time", are we not? Although we are stopping a time entry. Now this is confusing 😄

amulware: Different PR in either case. Freel free to create an issue and make the argument there.

Toggl.Multivac/Math.cs
simonrozsival: Strictly speaking, they can be anything. They go below 0 when I use the TimeToAngleOnTheDial. In some specific scenarios, I need it to be between 0 and 2PI to check if some angle is between two other angles.

Toggl.Multivac/Math.cs
simonrozsival: I don't want to add extension methods to doubles. I would have to create custom `Angle` value type, which would be just a wrapper around a double. Maybe it is not a bad idea.

kornelijepetak: Why? Just because of polluting autosuggestions?

Toggl.Multivac/Math.cs
kornelijepetak: Yeah, I was considering modulo (`(angle % FC + FC) % FC`) if numbers get far below 0, but if they don't then this `while` will not loop too much and in the end it will be more performant than modulo on doubles.

Toggl.Foundation.Tests/MvvmCross/ViewModels/CalendarPageViewModelTests.cs
```diff
@@ -8,6 +8,7 @@
 using Toggl.Foundation.MvvmCross.ViewModels.Calendar;
 using Toggl.Multivac;
 using Xunit;
+using Math = System.Math;
```
arturs105: :fire:

simonrozsival: This is unfortunately not possible - or at least it would not be so simple. There is now a conflict between `System.Math` and `Toggl.Multivac.Math`. Maybe I could rename our `Math` class to something else to avoid that. If you also have some naming suggestions, please post them here.

simonrozsival:
There is one known issue in this PR: dragging the gray area makes the duration oscillate between two values (like 17 minutes / 18 minutes) - this is due to rounding down to seconds. This will be fixed with #1361 later and we can merge this without that being fixed in this PR.

PR #1297, opened 2018-01-13
Improve animations

willsb:
It now looks like this:


Toggl.Giskard/Activities/MainActivity.cs
```diff
@@ -36,25 +38,56 @@ protected override void OnCreate(Bundle bundle)
...
+                    .Start();
+            }
         }
-    }
 
+        private sealed class FabAsyncHideListener : FloatingActionButton.OnVisibilityChangedListener
+        {
+            private readonly TaskCompletionSource<object> hideTaskCompletionSource = new TaskCompletionSource<object>();
+
+            public Task HideAsync => hideTaskCompletionSource.Task;
```
kornelijepetak: I'd rather this be a method, but this is not a cause for alarm.

PR #1302, opened 2018-01-22
The Spider

### LINKED ISSUE #1528: Finish new spider
#1302 is in need of attention. We have to make sure that there are no memory/other leaks and resolve merge conflicts and then start reviewing.

simonrozsival:
Closes #1528 

For now, the spider only appears when the suggestions are empty. It will be possible to place it in other views in the future as well. Also, there is currently no animation for hiding the spider - it will just disappear. We can add one later as well, but I think it is not needed for now.

Please pay close attention to possible leaks.

Squash whenever


simonrozsival:
I think there is some bug which causes the app to freeze after some time being suspended in the background. I will have to debug this somehow during next week.

tnebrs:
I just wanted to say thank you for this. Of all the wonderful features of Toggl and its views, this is my favourite. I may have glossed over logging the amount of time I spent admiring it.

PR #1304, opened 2018-01-22
Create expression tree manually in TE rivals resolver

### LINKED ISSUE #1301: Fix continue/suggestion crashing when TE is already running
This goes into the release branch please! 💚

simonrozsival:
Closes #1301 

The problem that this PR fixes was probably related to a recent update of Realm (1.4 -> 2.1) and the way it now handles expressions. The app would throw an exception in `TimeEntryRivalsResolver.cs` on line 36 because Realm couldn't work with type `IDatabaseTimeEntry` and it needed something derived from `RealmObject` - so that is why I added the generics so a `RealmTimeEntry` type could be passed to this method which is in Foundation. Another problem was that the expression in the `Where(...)` call wasn't compiled in a way which would be usable with Realm - the left-hand side of the expression must be the realm object directly (`other.Start`), but the compiler generated `Convert(other, ITimeEntry).Start` instead. That is why I needed to create the expression tree manually.

Please make a thorough functional test, at least test:
- continuing TE works
- creating a TE from a suggestion works
- starting/stopping TE works in the app
- pulling edited TEs from server works (stopped TE, edited description, changed workspace, added/removed tags)
- pulling deleted TEs from server works
- app resolves two running TEs well (go offline, create a running TE on the server, create a TE in the app, go online, force sync)


Toggl.Foundation/Sync/ConflictResolution/TimeEntryRivalsResolver.cs
```diff
@@ -28,18 +27,32 @@ public TimeEntryRivalsResolver(ITimeService timeService)
...
+            where TDatabaseObject : IDatabaseTimeEntry
             => rival.At < entity.At ? (entity, stop(rival, allTimeEntries)) : (stop(entity, allTimeEntries), rival);
 
-        private IDatabaseTimeEntry stop(IDatabaseTimeEntry toBeStopped, IQueryable<IDatabaseTimeEntry> allTimeEntries)
+        private IDatabaseTimeEntry stop<TDatabaseObject>(IDatabaseTimeEntry toBeStopped, IQueryable<TDatabaseObject> allTimeEntries)
+            where TDatabaseObject : IDatabaseTimeEntry
         {
-            var stop = ((IEnumerable<IDatabaseTimeEntry>)allTimeEntries.Where(other => other.Start > toBeStopped.Start))
+            var stop = allTimeEntries.Where(startsAfter<TDatabaseObject>(toBeStopped.Start))
+                .ToList()
```
amulware: Why is the ToList() necessary?

Toggl.PrimeRadiant/IRivalsResolver.cs
```diff
@@ -8,6 +8,7 @@ public interface IRivalsResolver<TModel>
     {
         bool CanHaveRival(TModel entity);
         Expression<Func<TModel, bool>> AreRivals(TModel entity);
-        (TModel FixedEntity, TModel FixedRival) FixRivals(TModel entity, TModel rival, IQueryable<TModel> allEntities);
+        (TModel FixedEntity, TModel FixedRival) FixRivals<TRealmObject>(TModel entity, TModel rival, IQueryable<TRealmObject> allEntities)
```
amulware: How come the generic parameter doesn't break the interface (or rather its usages)?

Toggl.Foundation/Sync/ConflictResolution/TimeEntryRivalsResolver.cs
simonrozsival: We need to cast it from IQueryable to IEnumerable, so we can use normal LINQ. If you want to use the previous way of doing it with the explicit cast (`(IEnumerable<TDatabaseObject>)`), I will change it.

amulware: I'd prefer that, yes.

Toggl.PrimeRadiant/IRivalsResolver.cs
simonrozsival: Because it is inferred from usage (the third param). The reason why the tests were failing (and what was fixed in `f2bcfc7`) is that it was used directly with `IDatabaseTimeEntry` and it wasn't possible to use reflection correctly on that type. Maybe we should add a `class` constraint here.

simonrozsival:
The class constraint is there because I need to access `typeof(TDatabaseObject).GetProperty("Start")` - the `IDatabaseTimeEntry` interface doesn't have such property - it is in `ITimeEntry`. If the `TDatabaseObject` is a class, then I can access the property thanks to inheritance.

arturs105:
Did a functional test, continuing and starting from suggestions is indeed fixed.

Though the app crashes with an `EntityNotFoundException` if you start randomly continuing time entries (abusing the app), which leaves the app in a broken state - running time entry isn't shown and starting a new one crashes the app. This might be unrelated, so what we do about this is up to @simonrozsival and @amulware

amulware:
Any idea on why/how that exception happens? What entity is not found?

Could it be that it tries to fetch a TE by its negative ID, just after it was synced and changed to a positive ID? Or do we take local-ids into account for that?

arturs105:
Only happens when you don't wait for the syncing to complete after starting/continuing a time entry and start several ones like this.

Actually found two different exceptions while doing this:
* `EntityNotFoundException`
    * `TimeEntriesLogViewModel.cs:199` (when continuing repeatedly)
    * `SuggestionsViewModel.cs:88` (when using suggestions repeatedly)
* `OverflowException` (`TimeSpan` duration too long)
    * `TimeEntryViewModelCollection.cs:27`

Toggl.Foundation/Sync/ConflictResolution/TimeEntryRivalsResolver.cs
```diff
@@ -28,18 +28,33 @@ public TimeEntryRivalsResolver(ITimeService timeService)
...
-            long duration = (long)(stop - toBeStopped.Start).TotalSeconds; // truncates towards zero (floor)
+            long duration = (long)(stopTime - toBeStopped.Start).TotalSeconds; // truncates towards zero (floor)
             return new TimeEntry(toBeStopped, duration);
         }
+
+        private Expression<Func<TDatabaseObject, bool>> startsAfter<TDatabaseObject>(DateTimeOffset start)
+            where TDatabaseObject : IDatabaseTimeEntry
+        {
+            var other = Expression.Parameter(typeof(TDatabaseObject), "other");
+            var otherStart = Expression.Property(other, typeof(TDatabaseObject).GetProperty("Start"));
```
willsb: You could have `nameof(ITimeEntry.Start)` here instead

amulware:
Can we squash this @simonrozsival?

PR #1305, opened 2018-01-23
Add haptic feedback to the pull-to-refresh gesture

simonrozsival:
This is a minor feature I want to add - many other iOS apps (facebook, twitter, ...) have this kind of feedback for the pull-to-refresh gesture where the user gets a light vibration when he can stop pulling and the app will refresh. I think it makes the app feel a lot better.

Make a functional test on a physical device while reviewing.

There is no issue or discussion for this feature, so I think this shouldn't be squashed before @amulware and @pe0ny approve it.


simonrozsival:
I created an issue and I am closing this for now so it doesn't distract us.

PR #1318, opened 2018-01-24
Remove warnings about suggestions

willsb:
We were getting warnings about missing properties. Some were due to copy pasta and others because of linker stripping. This fixes both.

While I was at it, I also made Suggestion a struct. There’s no reason for it not to be a ValueType.

I’d like a review (and functional test) from someone from both iOS and Android teams.


amulware:
What's the reasoning with turning this into a struct? Will that not involve a lot of copying of data anytime we work with it?

willsb:
@amulware No, it does not. It simply makes the process of creating those compiler safe (ensuring everything is properly initialized) and, in general, communicates the class' purpose better.

amulware:
How does it make that compiler safe? Or, how is that different from having it a reference type with a proper constructor?

I'm also not sure how the class' purpose dictates what kind of type it should be. In general, value types should be small atomic units of information. This is a large thing with a bunch of different things in it. All that will be copied around every time the struct is passed to a new function, whether it's all needed or not.

Issue #1306, opened 2018-01-23
`EntityNotFoundException` when starting multiple time entries quickly


amulware:
Followup of https://github.com/toggl/mobileapp/pull/1304

Happens for both continuing and suggestion-starting, not sure about the start button.

simonrozsival:
Apparently clicking the continue button repeatedly leads to a state when there are now two running TEs in the local database. My theory is that there are two parallel Realm write transactions which both add a running TE and both succeed. I am not 100% sure this is what happens, but I will keep debugging this and make sure we avoid this. Once this happens, the app will crash and it won't work properly anymore (no running TE will be displayed).

simonrozsival:
Interesting - syncing the app _fixes_ this inconsistency (backend probably stopped one of the TEs when pushing and then pulling). So at least it is easy to recover from this state if some user ever experiences this before we fix it/prevent this from happening.

simonrozsival:
I found out that user can "easily" tap the stop button twice with a double tap (the button is still active while it is shrinking) and the second stop, of course, cannot stop anything and it will cause a crash.

I think we should fix this in two different ways:
- introduce some "locking" mechanism into TimeEntriesDataSource to prevent race conditions
- disable buttons for a short period of time once they are clicked so the user can't accidentally click it twice (this could be easily done for the stop button - you can only click this button once and then it must be disabled until a new TE is started

amulware:
Disabling the button (without changing the way it's drawing) while it shouldn't be used makes perfect sense to me.

About the locking mechanism, I'm not really sure. I would like to hear some more specifics and other's opinions.

simonrozsival:
_Specifics_: we do several operations while user stops the TE, starts, updates, ... - we build an entity from a DTO, we make a database operation, we update the internal state of the data source (`long? currentlyRunningTimeEntryId`). This should be a sort of a transaction. We shouldn't run several of these transactions in parallel and each of them should wait until the previous one completes all of its steps. These operations happen quite rarely and they are mostly triggered by some of the user's action (clicking a button, pull-to-refresh).

willsb:
A lock mechanism **in the UI** is the cleanest, most testable and more cross platform solution. A simple implementation of `CanExecute` in the `StopCommand` will fix this.

simonrozsival:
The app still crashes when I try to quickly click the continue button several times in a row.

simonrozsival:
With push sync disabled, this does not happen.

[...]

PR #1321, opened 2018-01-25
Fix behavior of the Stop button and the Edit command

simonrozsival:
Ref #1306 

There was a problem that user could click the stop button twice before it would disappear after it was clicked the first time (and try to stop it for the second time => exception). When this was disabled then the user could click the stop button AND then edit the "currently running time entry" (which was already stopped => exception).

Just for sure, I made sure user cannot tap the start button if it was shown for some reason while there was already a running time entry.

Please make a functional test - first try to reproduce the crash in the develop branch and afterward test this fix. I would like someone from the Android subteam to try this as well.

Squash whenever


simonrozsival:
This PR doesn't close the issue because it only deals with the currently running TE in the MainViewModel. The fix for the continue button (in the TimeEntriesLogViewCell) is not included in this PR.

willsb:
Please note that I've approved the ViewModel portion of this; a review and functional test from @arturs105 is still needed

PR #1347, opened 2018-01-26
Add application shortcuts

arturs105:
Note that this only adds the shortcuts to the app - nothing happens when you tap on them. Shortcut  handling will come in another PR.


Toggl.Foundation/Login/LoginManager.cs
```diff
@@ -53,7 +58,8 @@ public IObservable<ITogglDataSource> Login(Email email, Password password)
                     .SelectMany(_ => apiFactory.CreateApiWith(credentials).User.Get())
                     .Select(User.Clean)
                     .SelectMany(database.User.Create)
-                    .Select(dataSourceFromUser);
+                    .Select(dataSourceFromUser)
+                    .Do(_ => shortcutCreator.OnLogin());
```
simonrozsival: I think the best place for calling `OnLogin` should be the `dataSourceFromUser` - as it should be called also after `GetDataSourceIfLoggedIn` and `RefreshToken` anyway - or am I missing something?

Toggl.Foundation/ApplicationUrls.cs
```diff
@@ -0,0 +1,14 @@
...
+{
+    public static class ApplicationUrls
+    {
+        public static string Reports = "toggl://main/reports";
+
+        public static string StartTimeEntry = "toggl://main/start";
+
+        public static string ContinueTimeEntryEntry(long id) => $"toggl://continue?id={id}";
+
+        public static string StartTimeEntryWith(string description, long? projectId, long? taskId)
```
simonrozsival: I think we should explicitly specify the workspace here.

Toggl.Foundation/Login/LoginManager.cs
arturs105: I disagree. I think `dataSourceFromUser` should do only the thing it says it does - create a data source from `IUser`. But i will add `OnLogin` to `GetDataSourceIfLoggedIn` and `RefreshToken`.

Toggl.Foundation/ApplicationUrls.cs
```diff
@@ -0,0 +1,14 @@
+﻿namespace Toggl.Foundation
+{
+    public static class ApplicationUrls
+    {
+        public static string Reports = "toggl://main/reports";
+
+        public static string StartTimeEntry = "toggl://main/start";
+
+        public static string ContinueTimeEntryEntry(long id) => $"toggl://continue?id={id}";
```
willsb: Why not add `main` here or simply drop `main` from the other urls?

Toggl.Foundation/Shortcuts/BaseApplicationShortcutCreator.cs
```diff
@@ -0,0 +1,82 @@
...
+        private IDisposable mostUsedTimeEntrySubscription;
+
+        public BaseApplicationShortcutCreator(MostUsedTimeEntrySuggestionProvider mostUsedTimeEntryProvider)
+        {
+            Ensure.Argument.IsNotNull(mostUsedTimeEntryProvider, nameof(mostUsedTimeEntryProvider));
+
+            this.mostUsedTimeEntryProvider = mostUsedTimeEntryProvider;
+        }
+
+        private readonly ApplicationShortcut reportsShortcut = new ApplicationShortcut(
```
willsb: These fields should be before the constructor

Toggl.Foundation/Shortcuts/BaseApplicationShortcutCreator.cs
```diff
@@ -0,0 +1,82 @@
...
+                    suggestion.Description,
+                    suggestion.WorkspaceId,
+                    suggestion.ProjectId,
+                    suggestion.TaskId),
+                title: Resources.MostUsedEntry,
+                subtitle: $"{suggestion.Description} {suggestion.ProjectName}",
+                type: ShortcutType.TimeEntrySuggestion
+            );
+
+        public abstract void ClearAllShortCuts();
```
willsb: Why not make these protected?

Toggl.Foundation/Shortcuts/IApplicationShortcutCreator.cs
```diff
@@ -0,0 +1,14 @@
...
+using Toggl.Multivac.Models;
+
+namespace Toggl.Foundation.Shortcuts
+{
+    public interface IApplicationShortcutCreator
+    {
+        void OnLogin();
+        void OnLogout();
+        void OnTimeEntryStarted(ITimeEntry timeEntry);
+        void ClearAllShortCuts();
```
willsb: I don't think this and `SetShortcuts` need to be part of the interface

Toggl.Foundation/Toggl.Foundation.csproj
```diff
@@ -29,4 +29,7 @@
     <ProjectReference Include="..\Toggl.PrimeRadiant\Toggl.PrimeRadiant.csproj" />
     <ProjectReference Include="..\Toggl.Ultrawave\Toggl.Ultrawave.csproj" />
   </ItemGroup>
+  <ItemGroup>
```
willsb: `git checkout develop -- Toggl.Foundation/Toggl.Foundation.csproj`

amulware:
Please note above what shortcuts exactly are being added.

Also, in general, it is best to keep `develop` in a release-able state. What's the reasoning for handing the shortcuts in a separate PR?

Toggl.Giskard/Startup/Setup.cs
```diff
@@ -49,17 +50,19 @@ protected override void InitializeApp(IMvxPluginManager pluginManager, IMvxAppli
...
                 clientName,
                 version,
-                new Database(),
-                new TimeService(Scheduler.Default),
+                database,
+                timeService,
                 new MailService(),
                 new GoogleService(),
                 environment,
+                new ApplicationShortcutCreator(new MostUsedTimeEntrySuggestionProvider(database, timeService, 1)),
```
willsb: It's kinda weird that we are creating another `MostUsedTImeEntrySuggestionProvider` instead of using the current provider holder that, in principle, has all the providers used by the platform. Any reason why you chose this approach?

arturs105: 1. Forgot that `ISuggestionProviderContainer` is registered as singleton
2. The shortcut creator needs just `MostUsedTimeEntrySuggestionProvider `.

I could switch to injecting `ISuggestionProviderContainer` and finding `MostUsedTimeEntrySuggestionProvider ` in the provider list, but there's no guarantee that finding it will always succeed  - in the future we might modify the provider list in `ISuggestionProviderContainer`.

arturs105:
@amulware The reasons for splitting this into 2 PRs were:
1. This PR is already quite big
2. Needed William to approve shortcut creation before working on shortcut handling

I can add the handling stuff to this PR as well, if you want.

amulware:
Understood. If you can temporarily disable the shortcuts from showing up, so that we can release at any moment, I don't mind if this gets merged unfinished. But I do not want `develop` to have a feature the user can discover that might break the app, or not work at all. :)

PR #1352, opened 2018-01-26
Fix invalid duration display

### LINKED ISSUE #1351: Invalid duration display
With current implementation of `TimeSpanToDurationWithUnitValueConverter`, when duration is exactly one minute*, the converter will return `00 sec`. Only when the duration skips to 1 minute and 1 second will the duration be displayed correctly. The same happens for hours.

kornelijepetak:
Closes #1351 

Squash whenever.


Toggl.Foundation.MvvmCross/Converters/TimeSpanToDurationWithUnitValueConverter.cs
```diff
@@ -11,30 +11,13 @@ public sealed class TimeSpanToDurationWithUnitValueConverter : MvxValueConverter
     {
         protected override string Convert(TimeSpan value, Type targetType, object parameter, CultureInfo culture)
         {
-            var unit = getUnitFor(value);
-            var durationBuilder = new StringBuilder();
+            if (value >= TimeSpan.FromHours(1))
+                return $"{(int)value.TotalHours:00}:{value.Minutes:00}:{value.Seconds:00}";
```
amulware: Given that the format is '00', is the cast to Int32 necessary?

amulware: What about replacing the whole line with
`return value.ToString("hh:mm:ss");`
and maybe below
`return $"{value:"mm:ss"} {Resources.UnitMin}";`
`return $"{value:"ss"} {Resources.UnitSecond}";`
?
Or does that not work?

amulware: Oh, scratch the first, hours has to go above 60. Right.

PR #1357, opened 2018-01-29
Add feedback to settings

### LINKED ISSUE #1181: Add feedback option to the Settings page
- [ ] Add feedback option so users can contact us directly from the app.
- [ ] Let's keep it simple and open an email-sending form when user taps 'Submit Feedback'. *(Like the one we used for Superday)*
- [ ] Send email to support@toggl.com
- [ ] Subject line: Toggl iOS feedback
- [ ] Include app version number, OS version and device in the email
- [ ] When a user doesn't have an email account set up, display this error message:
*Oops! Seems like your email account is not set up*
*Go to “Settings > Mail > Add Account” to set up an email account or send us your feedback to support@toggl.com*
- [ ] When a user is offline, display this error message:
*Sorry. Can’t send the email.*
*You’re offline. Please connect to the internet and try again.*

### Settings
toggl - ios - settings synced copy 1x

Zeplin: https://zpl.io/25NG94r

### Email account not set up
toggl - ios - settings error1 1x

Zeplin: https://zpl.io/VkKk7eg

### No connection
toggl - ios - settings error2 1x

Zeplin: https://zpl.io/bo056Mg

arturs105:
Closes #1181 

❗️ Please do a functional test **on an actual iPhone**. I couldn't do it, since i don't have my usb-c -> usb adapter with me (first world problems). Please check, that the correct iPhone model is added to the e-mail. On simulator it said "x86_64", but i hope it works on an actual iPhone (haven't tested it due to the reason above).


Toggl.Daneel/DeviceInfo.cs
```diff
@@ -0,0 +1,40 @@
...
+        public string OperatingSystem { get; }
+
+        public DeviceInfo()
+        {
+            var device = UIDevice.CurrentDevice;
+            OperatingSystem = $"{device.SystemName} {device.SystemVersion}";
+            PhoneModel = getDeviceModel();
+        }
+
+        private string getDeviceModel()
```
simonrozsival: Please add an explanation why and how exactly this works. There are also no try/catch blocks or null checks for the pointers - what if anything goes wrong? Is there some scenario, where the app crashes when it reads some memory it doesn't have access to?

arturs105:
I added some error handling, but i wasn't sure if wrapping `getDeviceModel` in `try catch(Exception)` was needed.

Toggl.Daneel/DeviceInfo.cs
```diff
@@ -0,0 +1,202 @@
...
+        }
+
+        [DllImport("libc", CallingConvention = CallingConvention.Cdecl)]
+        static internal extern int sysctlbyname([MarshalAs(UnmanagedType.LPStr)] string property, IntPtr output, IntPtr oldLen, IntPtr newp, uint newlen);
+
+        //Source: https://github.com/dannycabrera/Get-iOS-Model/blob/master/Xamarin.iOS.DeviceHardware.cs
+        private string getDeviceModel()
+        {
+            //Get length of phone identifier
+            var pointerToLength = Marshal.AllocHGlobal(sizeof(int));
```
simonrozsival: This can cause `OutOfMemoryException` - we should definitely use a try-catch here.

Issue #1143, opened 2017-12-11
New time entry UI


pe0ny:
- [x] Suggestion modes for time entries, projects, tasks and tags work the same way like on iOS.
- [x] Tapping projects or tags icon at the top right corner, opens projects/tags suggestion mode.
- [x] Tapping billable icon at the top right corner, sets (or removes) a billable flag to a time entry.
- [x] Back arrow takes the user back to the timer page and discards new time entry.
- [x] Tapping ticking duration opens expandable time/date edit. 
- ~Start/stop times and date use pickers like they do in edit view.~ On hold for the meantime
- [x] Green fab saves the changes and takes the user back to the timer page.

### New time entry
![new time entry](https://user-images.githubusercontent.com/5064445/33833361-1344f080-de88-11e7-913d-a00ad597edc3.png)

Zeplin: https://zpl.io/2Ed3ONx

### Time entry autocomplete
![new time entry autocomplete](https://user-images.githubusercontent.com/5064445/33833911-e0465a64-de89-11e7-9c18-4f398e1754df.png)

Zeplin: https://zpl.io/V4EXDmM

### Projects autocomplete
![new time entry - projects](https://user-images.githubusercontent.com/5064445/33833925-eaf2eb4e-de89-11e7-9dce-990bb9edeff0.png)

Zeplin: https://zpl.io/bAO38MO

### Tasks
![new time entry - projects and tasks](https://user-images.githubusercontent.com/5064445/33833947-fbd3352c-de89-11e7-89c2-2ef171d84fb5.png)

Zeplin: https://zpl.io/a30JpLl

### Expanded time/date edit
![date expanded](https://user-images.githubusercontent.com/5064445/33834159-c84013a0-de8a-11e7-89e1-a8cc8922bf0e.png)

Zeplin: https://zpl.io/VkKLDYg

PR #1367, opened 2018-01-29
Refactor text info target binding

willsb:
Ref #1143 

This adds the more complex target binding, allowing us to use the shortcuts. I ended up reusing a significant amount of iOS code which turned out to be shareable! Nice one. The refactor also allowed me to add tests to the behaviour, making it much more predictable and easier to understand.

For this reason I require functional tests and reviews from members of both teams.


Toggl.Giskard/Autocomplete/TokenSpan.cs
```diff
@@ -0,0 +1,78 @@
...
+using Java.Lang;
+using MvvmCross.Platform;
+using MvvmCross.Platform.Droid;
+using MvvmCross.Plugins.Color.Droid;
+using Toggl.Giskard.Extensions;
+using FoundationColor = Toggl.Foundation.MvvmCross.Helper.Color;
+
+namespace Toggl.Giskard
+{
+    public sealed class ProjectTokenSpan : TokenSpan
```
kornelijepetak: Not a big issue, but I would still like every class to live in its own file.

Toggl.Giskard/Autocomplete/AutocompleteTextWatcher.cs
```diff
@@ -0,0 +1,46 @@
...
+        public event EventHandler TextChanged;
+        public event EventHandler ProjectDeleted;
+        public event EventHandler CursorPositionChanged;
+        public event EventHandler<TagDeletedEventArgs> TagDeleted;
+
+        public void AfterTextChanged(IEditable s)
+        {
+        }
+
+        public void BeforeTextChanged(ICharSequence s, int start, int count, int after)
```
kornelijepetak: Would you mind naming the first parameter to something more meaningful, for example `sequence`?

Toggl.Giskard/Autocomplete/AutocompleteTextWatcher.cs
```diff
@@ -0,0 +1,46 @@
...
+using Toggl.Foundation.MvvmCross.Autocomplete;
+using Object = Java.Lang.Object;
+
+namespace Toggl.Giskard.Autocomplete
+{
+    public sealed class AutocompleteTextWatcher : Object, ITextWatcher, IAutocompleteEventProvider
+    {
+        public event EventHandler TextChanged;
+        public event EventHandler ProjectDeleted;
+        public event EventHandler CursorPositionChanged;
```
kornelijepetak: VS reports this event never used. I guess it's just implemented here to conform to `IAutocompleteEventProvider` interface but the implementor itself does not utilize it. Could you please confirm or check if it should be raised in at least one scenario.

willsb: Yes, it's implemented to conform with the interface. So far I haven't had the need to raise this event on Android, no

Toggl.Giskard/Autocomplete/AutocompleteTextWatcher.cs
willsb: Sure thing!

Toggl.Giskard/Autocomplete/TokenSpan.cs
willsb: 👌

PR #1368, opened 2018-01-29
Add Arrow back button

willsb:
Ref #1143 

Because the Toolbar was behaving weirdly, we added our own arrow manually. 

Squash whenever


willsb:
The problem is that the toolbar's back button was taking up space from the EditText

Issue #1365, opened 2018-01-29
TE creation view for manual entry


pe0ny:
We'll have two TE creation views: one for timer and one for manual mode.

- [ ] Trigger TE creation view for manual entry when a user has set so in the settings.
- [ ] Do not start the timer.
- [x] Set default duration value for 30min.
- [x] Change description placeholder to "What have you done?".

### TE creation view: manual entry
![toggl - ios - manual entry 1x](https://user-images.githubusercontent.com/5064445/35512002-d636d6a4-0506-11e8-8be6-d9202bc2cbf6.png)

Zeplin: https://zpl.io/2jkOBmm

simonrozsival:
What should be the default start time? Now or 30 minutes in the past or something completely different?

pe0ny:
@simonrozsival End time should be now and start should be 30min in the past.

PR #1369, opened 2018-01-29
Replace TextFieldInfo binding with a custom view

simonrozsival:
Ref #1365

I need this so that I can create a binding for the placeholder text. Also I never liked this binding - I think it is much cleaner as a separate view.

Squash whenever


simonrozsival:
I will have to redo this. I didn't remember that William refactored this recently and that PR hasn't been merged yet.

PR #1370, opened 2018-01-29
Track crashes using App Center

### LINKED ISSUE #1222: Set up HockeyApp to track crashes
Given how badly Firebase works with Xamarin let's use HockeyApp to get better stack traces instead. We also might not have any need for uploading symbol files there - it certainly wasn't necessary with Xamarin Insights.

willsb:
Closes #1222 

All needed variables were already added to Bitrise. You can already see crashes in the Droid project on app center.


amulware:
I note that this only activates crash tracking on production builds, which is indeed the correct behaviour. 👌

PR #1375, opened 2018-01-30
Manual mode settings

### LINKED ISSUE #1390: Settings view: Add manual switch
- [x] By default manual mode is off.
- [x] When a user turns on manual mode then play icon in the timer page will be replaced by the ‘+’ icon.

### Settings
toggl - ios - settings 1x

Zeplin: https://zpl.io/29EBEz2

simonrozsival:
Closes #1390

The preferences are reset when the user logs out, as was discussed in Slack.

Please review this PR, but let's not merge this until the manual mode PR #1394 is merged so that if we released then we don't have an option which doesn't have any effect.


Toggl.Daneel/ViewControllers/SettingsViewController.cs
```diff
@@ -111,6 +115,10 @@ public override void ViewDidLoad()
                       .For(v => v.BindAnimatedOn())
                       .To(vm => vm.UseTwentyFourHourClock);
 
+            bindingSet.Bind(ManualModeSwitch)
+                      .For(v => v.BindAnimatedOn())
+                      .To(vm => vm.IsManualModeEnabled);
```
willsb: What if the user taps the button? I don't think the command will be triggered in that case. 

Either disable the button for taps or move the calls to `userPreferences` to a  `OnIsManualModeEnabledChanged` method and let Fody handle the rest. I prefer the later

Issue #1121, opened 2017-12-07
Edit view UI


pe0ny:
- [x] Description and duration use inline editing.
- [x] Project and tags use a separate view.
- [x] Start/stop times and date use pickers.
- [x] Tapping back arrow discards changes.
- [x] Tapping 'Save' saves changes.
- [x] When time entry is running, the duration must be ticking as well.
- [x] Show 'Set' only when time entry is running.
- [x] Tapping 'Stop' stops a time entry and sets an end time.

### Edit view
![edit](https://user-images.githubusercontent.com/5064445/33831868-68657770-de82-11e7-950d-b4df8972864e.png)

Zeplin: https://zpl.io/V4E1oDz

### Description field active
![edit - description](https://user-images.githubusercontent.com/5064445/33722522-e1e663b0-db72-11e7-893b-72f761dce882.png)

Zeplin: https://zpl.io/bezYOnz

### All fields selected
![edit all fields selected](https://user-images.githubusercontent.com/5064445/33831874-70bdafd2-de82-11e7-8c31-d4bf5792423d.png)

Zeplin: https://zpl.io/VQMY1rm

amulware:
Implementations of specific features: #1122 #1123 #1143 #1144

kornelijepetak:
Closed by #1353 .

Redesigning UI will be part of a separate issue.

PR #1392, opened 2018-01-31
Droid chronology

kornelijepetak:
Adds Date and Time pickers to Droid project.

Currently the time mode (AM/FM vs military) defaults to `military`. Users should pull these from the user settings and provide to appropriate extension method.

Squash whenever.


willsb:
Ref #1143 and #1121

Toggl.Giskard/Dialogs/TogglDatePicker.cs
```diff
@@ -0,0 +1,54 @@
...
+
+            completionSource.SetResult(selectedDate);
+        }
+    }
+
+    public static class TogglDatePickerExtensions
+    {
+        public static void BindDatePickerToClick(this View view, DateTimeOffset? initialValue, Action<DateTimeOffset> onDateSelected)
+        {
+            view.Click += async (s, e) =>
```
willsb: Nopity nope. Turn this into a `WeakSubscribe` and return the disposable, otherwise we will get all them leaks 🚱

Toggl.Giskard/Dialogs/TogglDatePicker.cs
```diff
@@ -0,0 +1,54 @@
...
+using System.Threading.Tasks;
+using Android.App;
+using Android.Content;
+using Android.Views;
+
+namespace Toggl.Giskard.Dialogs
+{
+    public sealed class TogglDatePicker
+    {
+        private TaskCompletionSource<DateTimeOffset> completionSource;
```
willsb: All of these can be made readonly

Toggl.Giskard/Dialogs/TogglDatePicker.cs
```diff
@@ -0,0 +1,54 @@
...
+
+        public TogglDatePicker(Context context, DateTimeOffset? date)
+        {
+            this.context = context;
+            this.date = date?? DateTimeOffset.Now;
+        }
+
+        public Task<DateTimeOffset> Show(string title = "")
+        {
+            completionSource = new TaskCompletionSource<DateTimeOffset>();
```
willsb: This is meant to be a one use thing, so you should throw if it `Show` was already called

Toggl.Giskard/Toggl.Giskard.csproj
```diff
@@ -474,6 +476,7 @@
   </ItemGroup>
   <ItemGroup>
     <Folder Include="Presenters\" />
+    <Folder Include="Dialogs\" />
```
willsb: Not needed

Toggl.Giskard/Dialogs/TogglDatePicker.cs
kornelijepetak: This would of course be collected when the view is collected (when Edit TE activity is closed), but still, it's better to prevent this from happening at all.

kornelijepetak: but that is only my case, who knows who's gonna call this and where...

Toggl.Giskard/Toggl.Giskard.csproj
kornelijepetak: Not that I'd like to constantly battle Visual Studio .csproj generation. Can I please leave this the way it is? It's gonna get recreated anyway on every odd date...

willsb: Sure thing

PR #1393, opened 2018-01-31
Telemetry abstraction

### LINKED ISSUE #1128: Create crossplatform custom analytics abstraction
We should not expose firebase functionality but instead create a wrapper that can be implemented by other services in the future if needed.

Take a look at the tracking related issues to see what kind of events and metadata we want to track, or start a discussion if you like to determine how much/what kind of stuff we need. (We can of course always add more functionality to the interface later.)

Also see how to best translate this data into firebase terminology, so that we can run statistics as easily as possible.

Specific issues:
#616, #427, #868, #1116

willsb:
Closes #1128 

This includes, as expected, only the interface and a dummy implementation. Actual implementation and usage will come in later PRs that will solve the issues linked in #1128 

Squash whenever


simonrozsival:
I believe that this is so simple that I don't have to re-review after the conflicts are resolved.

amulware:
I don't think that this is a good long term solution. For everything that has to be tracked, we need to add new methods here, and then implement that in any implementations. I think it would be better to have a more generic method, or a few more generic methods that could track many different kinds of events.

Thoughts? @willsb @simonrozsival

simonrozsival:
I think it is good to have a specific method for a specific event.

amulware:
What about having to edit several different places every time we add a new event? It makes adding events a lot harder than it could be, and we'll have A LOT of different events in the future.

PR #1394, opened 2018-01-31
Manual mode

### LINKED ISSUE #1365: TE creation view for manual entry

simonrozsival:
Closes #1365

When the manual mode is enabled, the button image changes and also initiates the start view with different parameters, so that it behaves the way it should.

Squash whenever


Toggl.Foundation.MvvmCross/Parameters/StartTimeEntryParameters.cs
```diff
@@ -0,0 +1,19 @@
...
+{
+    public sealed class StartTimeEntryParameters
+    {
+        public DateTimeOffset StartTime { get; set; }
+
+        public string PlaceholderText { get; set; }
+
+        public TimeSpan? Duration { get; set; }
+
+        public StartTimeEntryParameters(DateTimeOffset startTime, string placeholderText, TimeSpan? duration)
```
willsb: Could you have explicit static methods so consumers are clear about which properties should be used for manual/auto mode?

Something in the lines of `StartTimeEntryParameters .ForManualMode()` and `StartTimeEntryParameters .ForAutomaticMode()`

Toggl.Foundation.MvvmCross/ViewModels/MainViewModel.cs
```diff
@@ -30,6 +31,8 @@ public sealed class MainViewModel : MvxViewModel
         private readonly IOnboardingStorage onboardingStorage;
         private readonly IMvxNavigationService navigationService;
 
+        private readonly TimeSpan defaultManualModeDuration = TimeSpan.FromMinutes(30);
```
willsb: Shouldn't these be configurable?

Toggl.Foundation.MvvmCross/ViewModels/MainViewModel.cs
```diff
@@ -165,8 +172,13 @@ private Task openSettings()
             await navigationService.Navigate<ReportsViewModel, long>(user.DefaultWorkspaceId);
         }
 
-        private Task startTimeEntry() =>
-            navigationService.Navigate<StartTimeEntryViewModel, DateTimeOffset>(timeService.CurrentDateTime);
+        private Task startTimeEntry()
+        {
+            var parameter = IsInManualMode
+                ? new StartTimeEntryParameters(timeService.CurrentDateTime.Subtract(defaultManualModeDuration), Resources.ManualTimeEntryPlaceholder, defaultManualModeDuration)
```
willsb: If you follow my advice and have the static methods, the string for the placeholder shouldn't be provided by consumers of the method

Toggl.Foundation.MvvmCross/ViewModels/MainViewModel.cs
simonrozsival: Not right now - the specs say exactly 30 minutes. I can put this in the constants class, but other than that I wouldn't put too much effort into this, as it's maybe not gonna be needed.

simonrozsival:
I noticed I made a mistake yesterday - I branched the original code from the settings branch, so this PR contained some files from the other PR (#1375)... I, therefore, cherry-picked only the commits that were intended to be in this PR (+1 commit I'm not sure how it got here, but it doesn't affect anything and it will be forgotten once we squash this) and force-pushed them here.

Toggl.Foundation.MvvmCross/Parameters/StartTimeEntryParameters.cs
```diff
@@ -0,0 +1,33 @@
...
+            Duration = duration;
+        }
+
+        public static StartTimeEntryParameters ForManualMode(DateTimeOffset now)
+            => new StartTimeEntryParameters(
+                now.Subtract(defaultManualModeDuration),
+                Resources.ManualTimeEntryPlaceholder,
+                defaultManualModeDuration);
+        
+        public static StartTimeEntryParameters ForAutomaticMode(DateTimeOffset now)
```
amulware: Please call this `ForTimerMode`, since that is the official name that we use.

willsb:
This is kinda blocking the progress for Droid, since it makes massive changes to the StartTimeEntryViewModel. 

In particular, the removal of the StopTime field, since, according to droid designs, users can select the stop date. This is not something impossible to circumvent (we simply calculate the duration when user selects the stop time) but now I'll have to handle it differently.

simonrozsival:
I can revert the `Duration` change, it's not a big deal. I changed it because we never use the value of the stop time, we use it only to calculate the duration. Are there any other problematic changes?

willsb:
No, I don't want it reverted. It's fine. I just need this merged :)

PR #1395, opened 2018-01-31
Move IDeviceInfo stuff into IPlatformConstants

arturs105:
Ref #1357 #1181 

Squash whenever


Toggl.Daneel/Startup/AppDelegate.cs
```diff
@@ -33,7 +33,7 @@ public override bool FinishedLaunching(UIApplication application, NSDictionary l
             #endif
             #if USE_ANALYTICS
             Microsoft.AppCenter.AppCenter.Start(
-                {TOGGL_APP_CENTER_ID_IOS},
```
willsb: y tho

PR #1398, opened 2018-01-31
Track onboarding skips

### LINKED ISSUE #427: Track of how many users skip walkthrough
Track of how many users click ‘Skip’ on walkthrough steps 1,2,3
= users who click ‘Skip’ / users who are on walkthrough steps 1, 2 or 3

willsb:
Closes #427 

I did that while waiting for Bitrise to build, hence why it’s done one week in advance.

Squash whenever


kornelijepetak:
I see you are doing that by sending a page index at the moment. What about sending the mnemonic for the page instead, cause IIRC we are missing one screen on the onboarding (timeline) so when we finally add it, the numbers will be off.

Comment on this or change it and I will approve it. No other objections.

willsb:
It won't be off because we currently removed the last page, but I like your idea :)

PR #1408, opened 2018-02-02
Fix dragging the wheel

### LINKED ISSUE #1361: Make dragging whole interval in wheel behave correctly
'Correctly' meaning: the way we arbitrarily define it as follows:

- [x] Dragging the interval changes both start and stop time
- [x] Dragging the interval NEVER changes the duration (not even the seconds)
- [x] Dragging the interval causes the start-time to round to full minutes
- [x] To keep duration the same, the end time has to be modified appropriately when the start time changes

simonrozsival:
Closes #1361 

Please make a functional test while reviewing. Try dragging by the start and stop handles, by the gray arc between them. Try while the TE is running and when it is stopped.

Squash whenever


amulware:
Functional test: Looks good, except I think users should not be able to drag the interval for running entries.

PR #1410, opened 2018-02-02
Prevent log from flickering

### LINKED ISSUE #1113: Stopping time entry causes brief flicker in log
This is especially visible if there are multiple days of data in the log. I think this may be related to how we remove a whole day from the list and reinsert it when time entries in the list are updated.

willsb:
Closes #1113 

Since there’s no support to what we are doing with nested observable collections, I had to pump my own events when the child collections get updated. 

We can support that in other views, if needed, but for the time being I’m defaulting to the previous behaviour in lists other than the log.


willsb:
I will wait until @kornelijepetak review this on Droid. I also don't wanna merge such big change before a release

PR #1412, opened 2018-02-02
Hide login fab when data is invalid

### LINKED ISSUE #1207: Make FAB and Back button states follow the VM state
Currently, the back button and the FAB do not get disabled or enabled according to the VM flow.

The calls to the handlers get swallowed if they should not be pressed at certain state, but the visuals don't change at all.

willsb:
Closes #1207 

Squash whenever


Toggl.Giskard/Bindings/FABVisibilityTargetBinding.cs
```diff
@@ -0,0 +1,30 @@
+﻿using Android.Support.Design.Widget;
+using MvvmCross.Binding;
+using MvvmCross.Binding.Droid.Target;
+
+namespace Toggl.Giskard.Bindings
+{
+    public sealed class FABVisibilityTargetBinding : MvxAndroidTargetBinding<FloatingActionButton, bool>
+    {
+        public const string BindingName = "FABVisibility";
```
arturs105: This is the only thing that's bothering me. My personal preference when using acronyms in pascal or camel case is to capitalise only the first letter: `FabVisibility`.

But i think our guidelines don't say anything about this and i'm not really sure how we do this in other places, so 🤷‍♂️

willsb: I agree with you, I will change that

PR #1414, opened 2018-02-05
Google signup basic integration tests

### LINKED ISSUE #1239: Create basic integration tests for Google signup endpoints
It isn't possible to test the whole process - we would need to create a new google account for each test, but we should at least test that the endpoints we use doesn't return 404 and maybe we will come up with more possible tests.

simonrozsival:
Closes #1239 

~This PR is useless at the moment. The idea was to test at least the fact that the signup endpoint is reachable and that we get some error because of validation. But there is probably some bug in the API and an invalid Google token leads to a 503 response, which isn't very helpful.~

~I will look into the backend repo if there is already an issue for the incorrect validation or I will create one and refer to this PR. Until this is resolved I don't think it makes sense to review this PR and I will set the issue #1239 to `on-hold`.~

Integration tests: https://www.bitrise.io/build/62708c170536d2b6

Squash whenever


willsb:
Any news on this one, @simonrozsival?

amulware:
The backend issue (see reference above) is still open.

simonrozsival:
The backend issue is closed now but hasn't been deployed yet.

PR #1416, opened 2018-02-05
Fix issue with visible placeholder when text changes programatically

simonrozsival:
When the user starts the new TE and clicks on the "toggle projects suggestions" button or "toggle tags suggestions" button, the placeholder isn't hidden, although the text field now should show just `#` or `@`. 

Squash whenever


amulware:
I don't understand the bug fixed. How exactly do I reproduce it?

simonrozsival:
If you start a new TE on `develop` and then you press the project button or the tags button in the bottom left corner, a `@` or `#` symbol is appended to the description text. If this text is empty, then it will now contain just the symbol. The problem is that it will not be visible, because it will be hidden under the placeholder (notice you can't even see the cursor).

PR #1417, opened 2018-02-05
Creation view: Ask confirmation when discarding TE

### LINKED ISSUE #1042: TE creation view: Ask confirmation when discarding TE
Ask confirmation: 'Discard' or 'Continue Editing' when user taps 'x' in TE creation view.

Zeplin: https://zpl.io/2jpx48Q

toggl - ios - add new entry discard confirmation 1x

simonrozsival:
Closes #1042 

Please make a functional test while reviewing.

Squash whenever


amulware:
According to the issue we only want to show the dialog if data is entered. However, if I type something into the field (or use project/tag buttons) and then clear it again, it still asks me.

Is this intended? Should it not actually check if there is data in the field?

amulware:
Maybe we need better definitions of what it means for the user to have 'entered data'.

pe0ny:
I would say if you clear it again, then the app shouldn't ask this confirmation.

However, we should also consider when a user changed:
- Billable flag
- Date
- Start/stop/duration
as 'entering data' and then ask confirmation even if description/project/tag field is empty.

simonrozsival:
That makes sense. I will update the PR to match this specification.

simonrozsival:
This is now updated and ready for another review

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
```diff
@@ -490,9 +503,28 @@ private void toggleTaskSuggestions(ProjectSuggestion projectSuggestion)
                 StartTime.Second,
                 StartTime.Offset)
                 .Clamp(minimum, maximum);
+
+            if (StartTime.Date != newStartTime.Date)
+            {
+                StartTime = newStartTime;
+            }
```
amulware: Is this condition necessary?

simonrozsival: It is not. 🔥

simonrozsival: It turned out that the condition is quite useful - I will keep it, otherwise I would have to change the way we compare datetimeoffsets (we would have to ignore sub-second time).

PR #1425, opened 2018-02-05
Status Bar colors

willsb:
This changes the status bar (and the notification icon) colors. 

I need @pe0ny to take a look at this in a pre-marshmallow device, since there's no way to change the notification icon tint in Lollipop. We need to validate how this looks before merging.


Toggl.Giskard/Extensions/ActivityExtensions.cs
```diff
@@ -0,0 +1,23 @@
...
+    public static class ActivityExtensions
+    {
+        public static void ChangeStatusBarColor(this Activity activity, Color color, bool useDarkIcons = false)
+        {
+            var window = activity.Window;
+            window.AddFlags(WindowManagerFlags.DrawsSystemBarBackgrounds);
+            window.ClearFlags(WindowManagerFlags.TranslucentStatus);
+            window.SetStatusBarColor(color);
+
+            if (Build.VERSION.SdkInt < BuildVersionCodes.M) return;
```
kornelijepetak: While you may have complete APIs from version 1 in your head, others may not, so I would add a comment here why do you make a distinction based on the API level. What is different from Marshmallow.

Toggl.Giskard/Activities/StartTimeEntryActivity.cs
```diff
@@ -25,6 +26,8 @@ public sealed class StartTimeEntryActivity : MvxAppCompatActivity<StartTimeEntry
 
         protected override void OnCreate(Bundle bundle)
         {
+            this.ChangeStatusBarColor(Color.White, true);
```
kornelijepetak: While I like flat surfaces, the dark status bars look different in comparison to this one, because this view is white as it is, and now the status bar is white too. I'd expect the same in other views, however other views have toolbar a bit brighter than the status bar. This may be intentional, but I feel it is inconsistent. Not something that invalidates this PR, but still, I'd like @pe0ny's look.

kornelijepetak: To illustrate:

image

image

willsb: Agreed. Material guidelines say the status bar should use the 700 color of the theme while the toolbar uses the 500 color, hence why you (and other Android users) are used to seeing the contrast

Toggl.Giskard/Extensions/ActivityExtensions.cs
willsb: Something I used to do is having static method for these kinds of checks, just for readability sake. But basically the flag that makes status bar icons darker is not available on pre-lollipop devices, as mentioned in the PR

kornelijepetak: Nobody will go to a closed PR to understand why this is here should the need arise in the future. So if you are disgusted by in-code comments, static method/property like  `Sdk.SupportsThisOrThat` or whatever would be nice I think.

Toggl.Giskard/Activities/StartTimeEntryActivity.cs
pe0ny: So the screenshot below illustrates how status bar will look on a pre-marshmallow device?
And if I understood correctly, you're happy with the dark one but think the white one should be a bit darker (light grey)?

kornelijepetak: Well, I don't mind it either way. I just feel it's inconsistent. I'd have to see how a very light gray would look instead of white, but in principle yes. I just wanted to say it so it doesn't pass unnoticed. If @pe0ny is alright with that, we should merge it.

kornelijepetak: @pe0ny confirmed this.

Issue #1389, opened 2018-01-31
Edit view updates


pe0ny:
- [ ] Tapping on a start/stop field, opens the wheel view
- [ ] Tapping on a duration field, opens the wheel view with the numpad opened
- [ ] Tapping on a date field, opens calendar barrel popup
- [ ] Tapping on a ‘Stop’, stops the time entry and assigns end time value

### Edit view - Running and stopped entry:
![toggl - ios - edit - running entry 1x](https://user-images.githubusercontent.com/5064445/35608982-3b1c9fd2-0664-11e8-9441-a40819d20158.png) ![toggl - ios - edit - stopped entry 1x](https://user-images.githubusercontent.com/5064445/35608984-3cdd0172-0664-11e8-9600-57c67417c958.png)

Zeplin: https://zpl.io/V0Mleex & https://zpl.io/2EyEqml

PR #1429, opened 2018-02-06
Edit view layout update

simonrozsival:
Ref #1389 

> Tapping on a duration field opens the wheel view with the numpad opened

☝️  this might change and so this is not implemented by this PR.

Don't forget to make a functional test when reviewing.

Squash whenever


Toggl.Daneel/ViewControllers/EditTimeEntryViewController.cs
```diff
@@ -22,7 +21,7 @@ namespace Toggl.Daneel.ViewControllers
     public partial class EditTimeEntryViewController : MvxViewController
     {
         private const int switchHeight = 24;
-        private const float nonScrollableContentHeight = 100;
+        private const float nonScrollableContentHeight = 116;
```
amulware: Two height constants, one being an integer, the other a float?

simonrozsival: One of them is not needed, so I 🔥 it since I am basically refactoring this class. The other should be a float.

Toggl.Daneel/ViewControllers/EditTimeEntryViewController.cs
```diff
@@ -119,14 +120,31 @@ public override void ViewDidLoad()
...
             bindingSet.Bind(ConfirmButton).To(vm => vm.ConfirmCommand);
-            bindingSet.Bind(DurationLabel)
+
+            bindingSet.Bind(DurationView)
                       .For(v => v.BindTap())
                       .To(vm => vm.EditDurationCommand);
 
+            bindingSet.Bind(StartTimeView)
+                      .For(v => v.BindTap())
+                      .To(vm => vm.EditDurationCommand);
```
kornelijepetak: Why does tapping `StartTimeView` lead to `EditDurationCommand`?

Toggl.Foundation.MvvmCross/ViewModels/EditTimeEntryViewModel.cs
```diff
@@ -93,6 +93,9 @@ public TimeSpan Duration
             }
         }
 
+        [DependsOn(nameof(StopTime))]
+        public bool IsRunning => stopTime.HasValue == false;
```
kornelijepetak: There is already a property called `IsTimeEntryRunning` with the same behaviour.

Toggl.Foundation.MvvmCross/ViewModels/EditTimeEntryViewModel.cs
```diff
@@ -149,7 +154,7 @@ public TimeSpan Duration
             ConfirmCommand = new MvxCommand(confirm);
             CloseCommand = new MvxAsyncCommand(close);
             EditDurationCommand = new MvxAsyncCommand(editDuration);
-            SelectStartDateTimeCommand = new MvxAsyncCommand(selectStartDateTime);
+            StopCommand = new MvxCommand(stopTimeEntry, () => IsRunning);
```
kornelijepetak: Why is this command really necessary? The `Stop` command is executed only if `() => IsRunning` but the `SelectEndTimeCommand` command checks the same condition anyway and does the same job. Seems superfluous to me.

Whether `SelectEndTimeCommand` is a good name, that can be discussed, but this `StopCommand` seems redundant.

Correct me if I'm wrong.

simonrozsival: I didn't want to bind `SelectEndTimeCommand` to the `Stop` button. Maybe you are right and this is unnecessary though 🤔

Toggl.Foundation.MvvmCross/ViewModels/EditTimeEntryViewModel.cs
simonrozsival: 👍

Toggl.Daneel/ViewControllers/EditTimeEntryViewController.cs
simonrozsival: Because it should open the Wheel view. The VM needs to be renamed, but that is will not be in this PR.

Toggl.Foundation.MvvmCross/ViewModels/EditTimeEntryViewModel.cs
simonrozsival: As I pointed out in Slack, the `StopCommand` is necessary due to specific needs of the iOS app (the `SelectEndTimeCommand` opens the time picker instead of the wheel).

PR #1432, opened 2018-02-07
Change duration while editing using numpad

simonrozsival:
There is a problem in the current version that when you use the numpad for editing the duration in the Wheel view and then you immediately tap the `save` button, the value isn't saved, because the input didn't lose focus. This PR fixes this.

As a side effect, we get a cool update of the wheel view during editing duration using the numpad.

Squash whenever


Toggl.Daneel/Views/EditDuration/DurationField.cs
```diff
@@ -93,29 +95,35 @@ private void startEditing(object sender, EventArgs e)
...
         }
 
         private void tryUpdate(DurationFieldInfo nextInput)
         {
             if (nextInput.Equals(input) == false)
             {
-                isDirty = true;
                 input = nextInput;
+
+                if (input.IsEmpty)
```
kornelijepetak: I'd write it this way, but I'll leave the choice with you (see, I also did the `== false`, just as you like it) 😄
```
isDirty = input.IsEmpty == false;
Duration = isDirty ? input.ToTimeSpan() : originalDuration;
```

PR #1446, opened 2018-02-07
Consolidate test attributes

### LINKED ISSUE #1047: Consolidate Integration test attributes.

willsb:
Closes #1047 

Squash whenever


Toggl.Ultrawave.Tests.Integration/LogTestInfoAttribute.cs
```diff
@@ -30,7 +30,7 @@ public override void Before(MethodInfo methodUnderTest)
         public override void After(MethodInfo methodUnderTest)
         {
             stopwatch.Stop();
-            Console.WriteLine($"{formattedTestName} ended after: {stopwatch.Elapsed}");
```
amulware: If we don't show what test this refers to, how will we know which test runs slow?

I believe that xunit will run many tests at once, so the console log will not be clear then.

amulware: I thought we were in the other attribute, so this isn't so much about tests running slow. Still, is it not useful to see which test completed, instead of just posting orphaned times?

willsb: I am showing the test name, what's the problem you see here?

amulware: The problem I see is that I can't read. ¯\_(ツ)_/¯

PR #1448, opened 2018-02-07
Track time entry starting events

### LINKED ISSUE #428: Track of how many users are using suggestions and log view
We want to know of how many users are using suggestions and log view.
To measure that, track of how many time entries are being made (both from 'play' and 'continue' button) on both views (suggestions and log view).

willsb:
Closes #428 

Squash whenever


Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
```diff
@@ -544,6 +549,7 @@ await dataSource.User.Current
                 .SelectMany(dataSource.TimeEntries.Start)
                 .Do(_ => dataSource.SyncManager.PushSync());
 
+            analyticsService.TrackStartedTimeEntry(TimeEntryStartOrigin.Main);
```
amulware: This is called when the start view is opened, not when the entry is actually confirmed started, right?

Toggl.Foundation.Tests/MvvmCross/ViewModels/StartTimeEntryViewModelTests.cs
```diff
@@ -1160,6 +1166,17 @@ public void CreatesARunningTimeEntryWhenDurationIsNull()
                     DataSource.TimeEntries.Received().Start(Arg.Is<StartTimeEntryDTO>(dto => dto.Duration.HasValue == false));
                 }
 
+                [Fact]
+                public void RegistersTheEventInTheAnalyticsService()
+                {
+                    var parameter = new StartTimeEntryParameters(DateTimeOffset.Now, "", null);
+
+                    ViewModel.Prepare(parameter);
+                    ViewModel.DoneCommand.ExecuteAsync().Wait();
```
amulware: Let's test that the done command is necessary and that the event is not reported when cancelling?

Toggl.Foundation/Analytics/BaseAnalyticsService.cs
```diff
@@ -1,19 +1,28 @@
...
+        private const string startTimeEntryEventName = "TimeEntryStarted";
 
         public void TrackOnboardingSkipEvent(string pageName)
         {
-            var dict = new Dictionary<string, string> { { onboardingSkipEventName, pageName } };
+            var dict = new Dictionary<string, string> { { pageParameter, pageName } };
             NativeTrackEvent(onboardingSkipEventName, dict);
         }
 
+        public void TrackStartedTimeEntry(TimeEntryStartOrigin origin)
```
amulware: Since we also have manual mode, maybe it would be better to call this `CreatedTimeEntry` or `TimeEntryCreated` and add another origin for manual mode?

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
willsb: Look at the lines on the left side. This is inside the `done` method.

Toggl.Foundation.Tests/MvvmCross/ViewModels/StartTimeEntryViewModelTests.cs
willsb: 👍

Toggl.Foundation/Analytics/BaseAnalyticsService.cs
willsb: Okay. My assertion was that we only cared about the button clicked, not the mode as well. This can be easily changed.

amulware: Let's track only actually created entries. Maybe we wanna track cancellations of dialogs/popups in the future as well though, let's keep that in mind for even more creepy stats. :)

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
amulware: Yap.

PR #1451, opened 2018-02-08
Fix device detection

### LINKED ISSUE #1441: Device type detection fails
The prefilled information in the feedbac email is incorrect. For my iPhone 7 it says “Failed to detect phone model”.

This might be possible to fix after i bit of a debugging, but I would suggest to include the model type onlywhen we can detect it and skip it if we can’t. It doesn’t look good...

simonrozsival:
Closes #1441 

Please make a functional test on a device

Squash whenever


Toggl.Foundation.MvvmCross/ViewModels/SettingsViewModel.cs
```diff
@@ -171,13 +172,20 @@ await navigationService
             var version = userAgent.ToString();
             var phone = platformConstants.PhoneModel;
             var os = platformConstants.OperatingSystem;
-            //2 leading newlines, so user user can type something above this info
-            var message = $"\n\nVersion: {version}\nPhone: {phone}\nOS: {os}";
+
+            var messageBuilder = new StringBuilder();
+            messageBuilder.Append("\n\n"); // 2 leading newlines, so user user can type something above this info
```
willsb: This is *very* OS specific. I don't really get the reason why you're doing it, but this should not be in the VM

simonrozsival: I don't think that is within the scope of this PR. We should create an issue to refactor this in some future milestone.

PR #1461, opened 2018-02-08
Change Wheel view height

### LINKED ISSUE #1455: Wheel view UI updates
- [ ] Do not to use a full-screen view instead, the view’s height should depend on the content - like the edit view.
- [x] When user taps on the duration, the view should move up to make a room for the keyboard.

### Wheel view
toggl - ios - wheel 1x toggl - ios - wheel - duration edit 1x

Zeplin: https://zpl.io/aRNr8Q0

simonrozsival:
Closes #1455 

The height of the Wheel view is now lower so it is easier to use. The position of the view is adapted when the keyboard is shown.

Please make a functional test and also test on iPhone X in the simulator.

Squash whenever


Toggl.Daneel/ViewControllers/EditDurationViewController.cs
```diff
@@ -15,16 +17,28 @@
...
+
+        public EditDurationViewController() : base(nameof(EditDurationViewController))
         {
         }
 
         public override void ViewDidLoad()
         {
             base.ViewDidLoad();
 
+            if (PresentationController is ModalPresentationController modalPresentationController)
```
willsb: Could you at least move this to a method so we can get what's going on?

amulware: Yeah, I agree with that, give it a name please. :)

PR #1473, opened 2018-02-09
Update crash packages

willsb:
After investigating the crashes further, we found out that the crash handlers were racing for registering the global fatal exception handler. That means we’ll get errors spread across the two platforms. Not cool.

This completely removes the Firebase crash package and updates the other packages.

Squash as you please


amulware:
Is there any reason to merge this into `saturn`?

PR #1484, opened 2018-02-10
Start time entry interactors

willsb:
In a great effort to clean our codebase, we should start using interactors! 

This first PR adds the factory boilerplate and the first interactors, used to start time entries. This will make our codebase clearer, easier to understand and will help us out of dependency hell/spread out business logic.

Shoutout to @Odrakir 

Squash whenever


Toggl.Foundation.MvvmCross/Interactors/StartTimeEntry/StartTimeEntryInteractor.cs
```diff
@@ -0,0 +1,30 @@
...
+        protected override TimeEntryStartOrigin Origin { get; }
+
+        public StartTimeEntryInteractor(
+            ITogglDataSource dataSource, 
+            IUserPreferences userPreferences,
+            IAnalyticsService analyticsService, 
+            Func<IDatabaseUser, StartTimeEntryDTO> dtoFunction)
+            : base(dataSource, analyticsService)
+        {
+            this.dtoFunction = dtoFunction;
```
Odrakir: Why is this function injected?

Odrakir: Oh... I think I get it, kind of...

I would probably skip all the inheritance and inject the `StartTimeEntryDTO` from the caller (the viewModel). But that's probably just me and I don't know what DTO is... 🤷‍♂️

Toggl.Foundation.MvvmCross/Interactors/Shared/BaseStartTimeEntry.cs
```diff
@@ -0,0 +1,33 @@
...
+        }
+
+        public IObservable<IDatabaseTimeEntry> Execute()
+            => dataSource.User.Current
+                .Select(dtoFromUser)
+                .SelectMany(dataSource.TimeEntries.Start)
+                .Do(_ => dataSource.SyncManager.PushSync())
+                .Do(_ => analyticsService.TrackStartedTimeEntry(Origin));
+
+        protected abstract StartTimeEntryDTO dtoFromUser(IDatabaseUser user);
```
simonrozsival: What about naming this `createDtoForUser` or something like that?

Toggl.Foundation.MvvmCross/Interactors/StartTimeEntry/StartTimeEntryInteractor.cs
simonrozsival: I don't like this function. Let's not use it this way. I think we should create the DTO directly in the view model and not to hide it here. All the VMs already have access to the data source, I don't see any benefit in hiding the conversion from e.g. a suggestion into the DTO in the interactor.

If we really wanted to do it this way (having more than one interactor which takes the actual DTO), then I think a better way would be to add two more specific interactors which would convert a suggestion or another TE into this DTO (they could easily have their own reference to the data source and fill in the user ID and the workspace ID) and we would use composition instead of inheritance to achieve the same functionality but without the template design pattern.

Toggl.Foundation.MvvmCross/Interactors/Shared/BaseStartTimeEntry.cs
arturs105: I think `dtoFromUser` is fine, because line 26 reads almost like human-speak.

kornelijepetak: I am with @arturs105 on this one.

Toggl.Foundation.MvvmCross/Interactors/Shared/BaseStartTimeEntry.cs
```diff
@@ -0,0 +1,33 @@
+﻿using System;
+using System.Reactive.Linq;
+using Toggl.Foundation.Analytics;
+using Toggl.Foundation.DataSources;
+using Toggl.Foundation.DTOs;
+using Toggl.PrimeRadiant.Models;
+
+namespace Toggl.Foundation.MvvmCross.Interactors
+{
+    public abstract class BaseStartTimeEntry : IInteractor<IObservable<IDatabaseTimeEntry>>
```
amulware: How about `CreateTimeEntry(Interactor)(Base)` or something more general like that?

Toggl.Foundation.MvvmCross/Interactors/StartTimeEntry/StartTimeEntryInteractor.cs
willsb: @simonrozsival There's no way for a time entry to be started without first fetching the user, I'm afraid. This is why I exposed this function instead of something else.

How do you propose to have the this composition? We inject what in which interactor?

Toggl.Foundation.MvvmCross/Interactors/IInteractor.cs
```diff
@@ -0,0 +1,9 @@
+﻿using System;
+
+namespace Toggl.Foundation.MvvmCross.Interactors
```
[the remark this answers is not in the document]
willsb: You are mostly correct, there currently is only one dependency in MvvmCross (the `TimeEntryViewModel` class) and it will most likely be changed anyways

Toggl.Foundation.MvvmCross/Interactors/Shared/BaseStartTimeEntry.cs
willsb: This is the exact reason for the naming

Toggl.Foundation.MvvmCross/Interactors/Shared/BaseStartTimeEntry.cs
```diff
@@ -0,0 +1,33 @@
...
+        private readonly ITogglDataSource dataSource;
+        private readonly IAnalyticsService analyticsService;
+
+        protected abstract TimeEntryStartOrigin Origin { get; }
+
+        protected BaseStartTimeEntry(ITogglDataSource dataSource, 
+                                     IAnalyticsService analyticsService)
+        {
+            this.dataSource = dataSource;
+            this.analyticsService = analyticsService;
```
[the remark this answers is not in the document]
willsb: _[sad piano music starts playing on the background]_

Toggl.Foundation.MvvmCross/Interactors/IInteractor.cs
```diff
@@ -0,0 +1,9 @@
...
+{
+    public interface IInteractor<out T>
+    {
+        T Execute();
```
amulware: What is the purpose of this interface? Why are we passing parameters in the constructor of the interactor, instead of having parameters in the Execute methods?

willsb: This is to enforce parameters in the constructor. Interactors are meant to be short living objects; Letting them be constructed earlier on might make people instantiate them and having them lying around until all params are available, something we do not want.

This also makes it simpler to enforce something in all interactors, if need be (like forcing all of them to implement `IDisposable`, for example)

Toggl.Foundation.MvvmCross/Interactors/Shared/BaseStartTimeEntry.cs
willsb: I don't understand what you meant with this comment

Toggl.Foundation.MvvmCross/Interactors/IInteractor.cs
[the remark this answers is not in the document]
simonrozsival: I think we should have a different DTO for this and this layer should be completely separate from the framework we use.

Toggl.Foundation.MvvmCross/Interactors/Shared/BaseStartTimeEntry.cs
amulware: I'm talking about renaming the class.

Toggl.Foundation.MvvmCross/Interactors/IInteractor.cs
amulware: Very well. Not 100% convinced yet, but I'll see.

Toggl.Foundation.MvvmCross/Interactors/Shared/BaseStartTimeEntry.cs
willsb: Yeah, I did rename it to StartTimeEntryInteractor. Create is a much better name, however

Toggl.Foundation/Models/IStartTimeEntry.cs
```diff
@@ -0,0 +1,66 @@
+﻿using System;
+
+namespace Toggl.Foundation.Models
+{
+    public interface ITimeEntryPrototype
```
arturs105: File name does not match interface name.

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
```diff
@@ -6,31 +6,29 @@
...
 using Toggl.Multivac;
 using Toggl.Multivac.Extensions;
-using Toggl.PrimeRadiant.Settings;
 using static Toggl.Foundation.Helper.Constants;
 
 namespace Toggl.Foundation.MvvmCross.ViewModels
 {
     [Preserve(AllMembers = true)]
-    public sealed class StartTimeEntryViewModel : MvxViewModel<StartTimeEntryParameters>
+    public sealed class StartTimeEntryViewModel : MvxViewModel<StartTimeEntryParameters>, ITimeEntryPrototype
```
arturs105: Why was this view model made an `ITimeEntryPrototype`?

Toggl.Foundation.Tests/Mocks/MockTimeEntryPrototype.cs
```diff
@@ -1,13 +1,10 @@
 ﻿using System;
+using Toggl.Foundation.Models;
 
-namespace Toggl.Foundation.DTOs
+namespace Toggl.Foundation.Tests.Mocks
 {
-    public sealed class StartTimeEntryDTO
+    public sealed class MockTimeEntryPrototype : ITimeEntryPrototype
```
arturs105: Why not use `Substotute.For<ITimeEntryPrototype>()`?

Toggl.Foundation.Tests/MvvmCross/ViewModels/SuggestionsViewModelTests.cs
```diff
@@ -139,10 +136,21 @@ private Recorded<Notification<Suggestion>> createRecorded(int ticks, Suggestion
...
+            {
+                var user = Substitute.For<IDatabaseUser>();
+                user.Id.Returns(10);
+                DataSource.User.Current.Returns(Observable.Return(user));
+
+                TimeService.CurrentDateTime.Returns(DateTimeOffset.Now);
+            }
+
             [Property]
             public void StarstATimeEntryWithTheSameValuesOfTheSelectedSuggestion(
```
arturs105: Can't we test all of this stuff in the interactor tests and then test that this view model actually calls that interactor? (applies to most of these tests)

Toggl.Foundation.Tests/Mocks/MockTimeEntryPrototype.cs
willsb: Because this is much simpler. Also, this is something we should be doing for a while now for our models; Sometime a simple Get/Set model is easier to read and write. Another PR will seal Ultrawave model and introduce mockable versions of our models instead

Toggl.Foundation.Tests/MvvmCross/ViewModels/SuggestionsViewModelTests.cs
willsb: I prefer this because it ensures that the interactor behaves according to what the VM expects. We might change something that breaks the interactor for VM X but not for VM Y

Toggl.Foundation.Tests/Mocks/MockTimeEntryPrototype.cs
arturs105: Isn't that the whole purpose of NSubstitute?

willsb: Nope. NSubstitute is meant to mock complex services when we just want to check calls. Our models are by no means complex

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
willsb: Because it can be used to create a time entry

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
```diff
@@ -529,27 +523,18 @@ private void toggleBillable()
...
             await navigationService.Close(this);
         }
 
+        public long? ProjectId => TextFieldInfo.ProjectId;
+
+        public long? TaskId => TextFieldInfo.TaskId;
+
+        public string Description => TextFieldInfo.Text?.Trim() ?? "";
+
+        public long WorkspaceId => TextFieldInfo.WorkspaceId;
```
amulware: These properties should be with the other ones above.

Toggl.Foundation.MvvmCross/ViewModels/TimeEntryViewModel.cs
```diff
@@ -45,6 +46,8 @@ public sealed class TimeEntryViewModel : MvxNotifyPropertyChanged
 
         public long[] TagIds { get; }
 
+        TimeSpan? ITimeEntryPrototype.Duration => null;
```
amulware: Why can this be null?

Toggl.Foundation.Tests/Mocks/MockTimeEntryPrototype.cs
amulware: I'm fine with what @willsb describes here.

Toggl.Foundation/Interactors/Shared/StartTimeEntryInteractor.cs
```diff
@@ -0,0 +1,72 @@
...
+using Toggl.Foundation.Models;
+using Toggl.Foundation.Shortcuts;
+using Toggl.Multivac;
+using Toggl.Multivac.Extensions;
+using Toggl.PrimeRadiant;
+using Toggl.PrimeRadiant.Models;
+
+namespace Toggl.Foundation.Interactors
+{
+    internal sealed class StartTimeEntryInteractor : IInteractor<IObservable<IDatabaseTimeEntry>>
```
amulware: Rename to `CreateTimeEntryInteractor` since this does not necessarily start running entries?

Toggl.Foundation/Interactors/InteractorFactory.cs
```diff
@@ -0,0 +1,77 @@
...
+
+        public IInteractor<IObservable<IDatabaseTimeEntry>> StartTimeEntry(ITimeEntryPrototype prototype)
+            => new StartTimeEntryInteractor(
+                idProvider,
+                timeService,
+                dataSource,
+                analyticsService,
+                shortcutCreator,
+                prototype,
+                userPreferences.IsManualModeEnabled() ? TimeEntryStartOrigin.Manual : TimeEntryStartOrigin.Timer);
```
amulware: I'm not sure I like there being logic inside the interactor factory. I much rather have sub-classed interactors that do these additional checks/conversions/etc. and are tested appropriately.

Toggl.Foundation/Models/ITimeEntryPrototype.cs
```diff
@@ -0,0 +1,66 @@
...
+        bool IsBillable { get; }
+
+        DateTimeOffset StartTime { get; }
+
+        TimeSpan? Duration { get; }
+    }
+
+    public static class TimeEntryPrototypeExtensions
+    {
+        internal struct TimeEntryPrototype : ITimeEntryPrototype
```
amulware: Is this ever used as a value type? If not, and it's always boxed, just turn it into a class in the first place.

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
```diff
@@ -529,27 +523,18 @@ private void toggleBillable()
...
-                    Description = TextFieldInfo.Text?.Trim() ?? "",
-                    ProjectId = TextFieldInfo.ProjectId,
-                    TagIds = TextFieldInfo.Tags.Select(t => t.TagId).Distinct().ToArray()
-                })
-                .SelectMany(dataSource.TimeEntries.Start)
-                .Do(_ => dataSource.SyncManager.PushSync());
-
-            var origin = userPreferences.IsManualModeEnabled() ? TimeEntryStartOrigin.Manual : TimeEntryStartOrigin.Timer;
-            analyticsService.TrackStartedTimeEntry(origin);
+            await interactorFactory.StartTimeEntry(this).Execute();
```
amulware: I'm not sure if I like using the VM as the prototype. Would it not be nicer to use a DTO?

Toggl.Foundation.MvvmCross/ViewModels/TimeEntryViewModel.cs
willsb: When you click the continue button, the time entry should be started (meaning it should be set without a duration, as the currently running time entry)

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
willsb: I don't think it would. We'd simply introduce another type that would require ugly initialization when we already have all the needed properties here. I think it'll just add more clutter

Toggl.Foundation/Interactors/InteractorFactory.cs
willsb: Me, on the other hand, don't see the problem in it. Those are being covered by tests in all the places where they are used. What do you mean with tested appropriately?

amulware: Nevermind the tested thing. But what bothers me is that as far as I understand, the whole point of interactors is to have all business logic in single points - the interactor - while this extracts tiny amounts of that logic into the factory, which really shouldn't worry about that stuff. I suppose the counter-argument is that no business logic is being executed here, it is merely configuring the interactor correctly... I'm still not really happy with that? What do you and others think?

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
amulware: I can let it slide as an implementation detail.

Toggl.Foundation/Interactors/InteractorFactory.cs
willsb: I agree that there should be no business logic in the factory; But this is really a tiny thing (for logging, which doesn't even affect our users directly). Now this opens my eyes for something much worse right below: I'm setting the start date/duration of the started time entry. This might be the case for subclassing. If you read the whole discussion, however, you'll see that some people were not happy with the idea of subclassing for small things. Maybe we can come to some middle ground, I'll wait to see what others have to weight in

amulware: Instead of inheritance... we could always use composition. :D

simonrozsival: I agree that the ternary operator is a bit weird, although it doesn't seem like a problem to me. The `StartTimeEntryViewModel` knows if it is creating a TE in `manual` mode or in `timer` mode (we can add a field and set it in the `Prepare` method) and there is no need to check for users preferences. That might not even be a good idea after all - if we add some force touch shortcut to the play button to start in the other mode then this would log an incorrect origin.

I haven't noticed the problem with setting the start time before - that is obviously a bug - good catch 👍 Maybe there is a similar problem with duration as well?

I don't see that as a reason for creating subclasses though. The `CreateTimeEntryInteractor` should just take the start time as an explicit parameter in the constructor and it can be filled correctly using the time service in the factory in the cases of starting a suggestion or continuing a TE (that doesn't seem like business logic to me, but like something that belongs into a factory).

I don't have any problem with subclassing, but I don't think it needs to be used in cases where a simpler and more flexible solution is available. If you think this should be implemented with inheritance then do it that way 😃

Toggl.Foundation/Interactors/InteractorFactory.cs
```diff
@@ -0,0 +1,82 @@
...
+
+            this.dataSource = dataSource;
+            this.idProvider = idProvider;
+            this.timeService = timeService;
+            this.userPreferences = userPreferences;
+            this.shortcutCreator = shortcutCreator;
+            this.analyticsService = analyticsService;
+        }
+
+        public IInteractor<IObservable<IDatabaseTimeEntry>> CreateTimeEntry(ITimeEntryPrototype prototype)
```
amulware: I don't like how `Continue` and `StartSuggestion` really are just special cases of this `Create`. But I cannot think of a better name for the `Create` method. `Start` doesn't cover it, and `StartOrManual` is silly. Maybe it is ok as it is?

kornelijepetak: The way I see it, the point is that in both cases, a TE is actually being created so `Create` makes sense IMO.

amulware: Sure, but suggestions and continuing also create, and the interface does not really make it 100% clear why they have special methods. /shrug

amulware:
For reference, with Arturs on vacation and his points above addressed, I think we can merge without his approval.

So only @kornelijepetak's and maybe @Odrakir's approval would be needed.

Toggl.Foundation/Interactors/IInteractorFactory.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System;
+using Toggl.Foundation.Models;
+using Toggl.Foundation.Suggestions;
+using Toggl.PrimeRadiant.Models;
+
+namespace Toggl.Foundation.Interactors
+{
+    public interface IInteractorFactory
```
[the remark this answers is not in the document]
willsb: I initially thought this would be a good idea, but this has been questioned above and, in the end, the line is quite thin and we are bound to reuse many of the interactor throughout the whole app, so this would end up in dependency hell all over again. This is easier to manage

Toggl.Giskard/Services/GoogleService.cs
```diff
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
```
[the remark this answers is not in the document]
willsb: BOM 💥 
This change will keep happening, so IMO should simply commit it

Toggl.Foundation.Tests/Mocks/MockTimeEntryPrototype.cs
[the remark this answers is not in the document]
willsb: Agreed. This is outside this PR's scope, but I 100% agree

Toggl.Foundation/Autocomplete/TextFieldInfo.cs
```diff
@@ -7,13 +7,14 @@ namespace Toggl.Foundation.Autocomplete
...
-        public static TextFieldInfo Empty { get; } = new TextFieldInfo("", 0, null, null, "", "", null, "", new TagSuggestion[0]);
+        public static TextFieldInfo Empty(long workspaceId) 
+            => new TextFieldInfo("", 0, workspaceId, null, "", "", null, "", new TagSuggestion[0]);
 
         public string Text { get; }
 
         public int CursorPosition { get; }
 
-        public long? WorkspaceId { get; }
+        public long WorkspaceId { get; }
```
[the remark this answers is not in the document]
willsb: To make initialisation easier, I guess. Very wrong

Toggl.Foundation/Interactors/Shared/CreateTimeEntryInteractor.cs
```diff
@@ -0,0 +1,90 @@
...
+            this.idProvider = idProvider;
+            this.dataSource = dataSource;
+            this.timeService = timeService;
+            this.shortcutCreator = shortcutCreator;
+            this.analyticsService = analyticsService;
+        }
+
+        public IObservable<IDatabaseTimeEntry> Execute()
+            => dataSource.User.Current
+                .Select(user =>
```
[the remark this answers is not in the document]
willsb: Agreed

PR #1488, opened 2018-02-12
Fix app crashing in the start view

### LINKED ISSUE #1483: App crashes in the Start view
How to reproduce:
- start a new time entry
- tap the tags or projects button
- select a tag/project
- tap the tags or projects button
- app crashes

Entering everything using @ and # shortcuts works well though.

simonrozsival:
Closes #1483 

This prevents the app from crashing. The text field logic will have to be rewritten to work properly though.

Please make a functional test when reviewing.

Squash whenever


simonrozsival:
The problem is that the `CursorPosition` property has invalid values for some reason when there is a project token and/or some tag tokens. The binding doesn't work well...

amulware:
This should go into a hotfix branch (branched from the release tag) please 👌

PR #1489, opened 2018-02-12
Fix reports calendar culture info

### LINKED ISSUE #1155: ReportsViewModel test fails because of localization issues
`InlineData` strings are hardcoded to English locale... Which fails the test on other cultures.

Currently the following test is failing: `ReportsViewModelTests.TheCurrentDateRangeStringProperty.ReturnsSelectedDateRangeAsStringIfTheSelectedPeriodIsNotTheCurrentWeek`

Example in Croatian `hr-HR` culture:

image

simonrozsival:
Closes #1155 

Squash whenever


amulware:
As far as I can see, this changes our calendar to always print dates in American English format.

Is this the intended behaviour? Should we not rather print in the user's format and adapt the tests for that?

Also ping @pe0ny

simonrozsival:
I would think that when the whole app is in English, it should be also displayed in English, and not in the device's language.

amulware:
I see your point. @pe0ny has the final call. (If we don't care that much right now, then let's go with what Simon says)

pe0ny:
I agree with Simon. Since for now, the whole app is in English only, dates should be also displayed in American English.

PR #1491, opened 2018-02-12
Track signup and login events

### LINKED ISSUE #616: Track of how many users sign up or log in
We want to know of how many users who download the app sign up or log in (or don't do anything at all).

simonrozsival:
Closes #616 

This is on hold until the issue is clarified.


amulware:
I believe the issue has been clarified. Can work continue or are there more open questions?

simonrozsival:
This PR is now ready for review.

Toggl.Foundation.MvvmCross/ViewModels/LoginViewModel.cs
```diff
@@ -303,22 +316,31 @@ private void signUp()
             loginDisposable =
                 loginManager
                     .SignUp(Email, Password)
+                    .Do(_ => analyticsService.TrackSignUpEvent(AuthenticationMethod.EmailAndPassword))
                     .Subscribe(onDataSource, onError, onCompleted);
         }
 
         private async Task startPasswordManager()
         {
+            analyticsService.TrackPasswordManagerButtonClicked();
```
willsb: In principle, you shouldn't be able to click the button if it's not available unless there's a bug. Any reason why we are tracking this?

simonrozsival: Good point. I will remove it.

PR #1503, opened 2018-02-13
Add swipe actions

### LINKED ISSUE #1396: Add swipe gesture to continue existing time entries
- [ ] Allow users to use swipe right gesture to continue existing time entries.
- [ ] Swipe to continue should work the same way like the continue button on the right side.
- [ ] Swipe is an additional option to continue time entries. This means users should be still able to use continue button on the right side as well.

### Timer page
toggl - ios - timer - swipe gestures 1x

Zeplin: https://zpl.io/VqOPJq5

arturs105:
Closes #1396 

I ended up using two APIs:
- swipe to continue - `UIContextualAction` (works only on iOS 11 and later, approved by Heidi)
- swipe to delete - `UITableViewRowAction` (works on all our supported devices)

I used 2 different APIs because `UITableViewRowAction`doesn't allow adding actions on the left (leading) side, but the other one doesn't work in pre iOS 11.

Squash whenever


Toggl.Foundation.MvvmCross/ViewModels/TimeEntriesLogViewModel.cs
```diff
@@ -208,6 +211,11 @@ private void OnIsWelcomeChanged()
         private Task edit(TimeEntryViewModel timeEntryViewModel)
             => navigationService.Navigate<EditTimeEntryViewModel, long>(timeEntryViewModel.Id);
 
+        private async Task delete(TimeEntryViewModel timeEntryViewModel)
+        {
+            await dataSource.TimeEntries.Delete(timeEntryViewModel.Id);
```
simonrozsival: Please add a unit test for this command.

Toggl.Daneel/ViewSources/TimeEntriesLogViewSource.cs
```diff
@@ -62,5 +74,44 @@ public override bool ShouldScrollToTop(UIScrollView scrollView)
             scrollView.SetContentOffset(CGPoint.Empty, false);
             return false;
         }
+
+        public override UITableViewRowAction[] EditActionsForRow(UITableView tableView, NSIndexPath indexPath)
+        {
+            return new[]
```
simonrozsival: Couldn't this be all in one line?

PR #1504, opened 2018-02-13
Fix incorrect tracking parameter

simonrozsival:
Squash whenever


simonrozsival:
I don't think it is vague in the context of starting a new time entry.

amulware:
@kornelijepetak Please approve or suggest another name. Both Simon and me think this is fine. :)

Issue #1323, opened 2018-01-25
Track successful/failed sync of entities


amulware:
On fail, we should try and include the reason for the failure, but we have to make sure that we do so in an anonymous way. This might need investigation whether we can include the error message returned from backend or not.

Related to #1116

simonrozsival:
Is there something like a list of all possible error messages or is there some way of extracting it from the v9 repo?

amulware:
Not to my knowledge, maybe backend can help.

kornelijepetak:
I had a talk with @amulware and it was decided that we will track:

1) Every entity (workspace, tag, client...) as it's own type for the sake of accuracy
2) Every occurence of an error as an event
3) Every successful sync of a single entity as an event

If at any time we realize there are sending too many of them, we will think about the solution then. Until that happens (if it does), it's more useful to have better numbers from the start.

kornelijepetak:
Also, it seems that backend takes responsibility to make sure that no server response will return any kind of identifying data...

PR #1506, opened 2018-02-13
Track sync errors

### LINKED ISSUE #1116: Report syncing errors to crash reporting tools
This will need a custom abstraction that allows us to report errors/warnings no matter what service we are using.

The reason behind this is that we should gather analytics about what kind of errors happen with what frequency to improve the syncing algorithms and detect problems quicker.

We probably want to report ALL errors that happen inside the syncing machine for the start. If some of them happen a lot and we don't actually care we can always filter them later.

However, we have to be extremely careful that these errors do not contain any actual user data.

simonrozsival:
Closes #1116 

This PR does not cover handling failed sync of entities - that is in the scope of PR #1323 and #1507 

Squash whenever


Toggl.Daneel/Services/AnalyticsService.cs
```diff
@@ -8,12 +9,21 @@ namespace Toggl.Daneel.Services
...
                 parameters.Values.ToArray(),
                 parameters.Keys.ToArray()
             ));
         }
+
+        protected override void NativeTrackException(Exception exception)
+        {
+            var eventName = exception.GetType().FullName;
+            var parameters = new Dictionary<string, string> { [exceptionMessageParameter] = exception.Message };
+            NativeTrackEvent(eventName, parameters);
```
amulware: I think the event name should be the same for all of these since the number of them is limited. Instead, I'd rather pass the exception type as a property.

simonrozsival:
I went through the scenarios and I am quite sure there won't be any user data in the tracked exceptions:

- it could be Realm errors, but we already dealt with that
- or it could be TimeoutException
- or it could be some exception not related to user's data (`StackOverflowException`, etc.) 
- it cannot be HTTP requests error exception because those are caught and processed (=> retry loop, unsyncable)

Have I forgotten/oversimplified this? Can anyone think of any exception which might contain our users' data?

amulware:
Your list sounds reasonable. We'll see about logging http errors down the line, and can then think about this again :D

Toggl.Foundation/Sync/SyncManager.cs
```diff
@@ -110,6 +117,7 @@ private void processError(Exception error)
             else
             {
                 progress.OnNext(SyncProgress.Failed);
+                analyticsService.TrackSyncError(error);
```
amulware: What is your reasoning for not tracking offline exceptions?

amulware:
Also please fix tests. :)

Toggl.Foundation/Sync/SyncManager.cs
simonrozsival: That exception doesn't specify an error and it will happen all the time for people I guess - is that something we want to track? If we do, I will change it ofc.

amulware: Ok, fair enough, I will create another issue for tracking this separate, and we can then decide after some time if we find having that data useful or not

PR #1511, opened 2018-02-14
Fix creating duplicates in the time entries log

### LINKED ISSUE #1493: Duplicate entries appearing
/case/643749: Duplicate entries appear after stopping timer.

/case/643152: Duplicate entries appear on sync. Attempting to delete duplicate entries crashes the app.

/case/643763: Duplicate (and even triplicate!) entries appear "after I adjust the duration wheel". Most likely this means duplicates appear after editing an entry.

/case/643764: Duplicates appear after entry is added, when the timer is started, or after the timer is stopped. Confirms that duplicate entries aren't being synced.

### LINKED ISSUE #1508: Duplicate entries after changing the start date
I'm not sure if it's related to #1493 or not

Steps to reproduce:
1) Tap to edit existing time entry (I opened an entry with today's date)
2) Change start date to yesterday and save
3) Going back to timer page you can see two entries: one in Today and one in Yesterday

Also, app crashes when deleting yesterday's entry:
1) Tap to edit yesterday's entry
2) Delete - App directs you back to the timer page
3) Both entries (one in Today and one in Yesterday) are still there
4) Tap to edit yesterday's entry - app crashes

simonrozsival:
Closes #1508, closes #1493

The problem was that if the date of the TE changed, we would never bother dealing with the old item in the log and we wouldn't remove it. This removes the old one and inserts the new one - both animations a preserved 👌 

Don't forget to make a thorough functional test.

Squash whenever


amulware:
Thank you for the rebase. Looking at the changeset before made my browser freeze :D

amulware:
After studying the code I believe this indeed fixes both issues. The 'random' duplication would appear more often for users in timezones further away from GMT+0 because of the mix of local/utc dates

PR #1517, opened 2018-02-14
Fix select date issues in the creation view and edit view

### LINKED ISSUE #1510: Time entry is created on a different day then the one selected in the date picker
The problem is with timezones - the datepicker returns a time on midnight and the -1 timezone (or any other timezone with negative offset) will change the selected date.

### LINKED ISSUE #1516: Incorrect behavior of selecting start date
There are problems in both the start time entry view and in the edit view:

- start date of running time entry can be set to any date since 2006 -> duration waaaay over 999h in the edit view
- the start time entry view doesn't distinguish between a running and stopped TE at all and the limits for a stopped time entry are incorrect
- the start time entry view doesn't preserve duration for a stopped time entry

simonrozsival:
Closes #1510, closes #1516 

Squash whenever


Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
```diff
@@ -487,27 +489,30 @@ private void toggleBillable()
 
         private async Task setStartDate()
         {
-            var minimum = StartTime + DisplayedTime - TimeSpan.FromHours(MaxTimeEntryDurationInHours);
-            var maximum = StartTime + DisplayedTime;
-            var parameters = DateTimePickerParameters.WithDates(DateTimePickerMode.DateTime, StartTime, minimum, maximum);
+            var (earliestAllowedStartTime, latestAllowedStartTime) = isRunning
+                ? (timeService.CurrentDateTime - MaxTimeEntryDuration, timeService.CurrentDateTime)
+                : (EarliestAllowedStartTime, LatestAllowedStartTime);
```
amulware: This looks to be the same code as in the edit VM. Can this be abstracted somewhere else?

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
```diff
@@ -487,27 +489,30 @@ private void toggleBillable()
...
 
-            var selectedDate = await navigationService
+            var parameters = DateTimePickerParameters.WithDates(
+                DateTimePickerMode.Date,
+                StartTime,
+                earliestAllowedStartTime,
+                latestAllowedStartTime);
```
amulware: Maybe together which this.

amulware: Or maybe the entire containing method?

amulware:
Also, let's ease up on the use of red labels :D

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
simonrozsival: I don't think it is a good idea to move the whole command method somewhere else - I refactored it a bit and now it is very clear what the command does. Maybe this could be abstracted with an interactor later, but I rather wouldn't do any drastic changes in this hotfix.

Toggl.Foundation.MvvmCross/ViewModels/EditTimeEntryViewModel.cs
```diff
@@ -285,15 +285,9 @@ private void stopTimeEntry()
...
-            var latestAllowedStartTime = IsTimeEntryRunning
-                ? timeService.CurrentDateTime
-                : LatestAllowedStartTime;
-
-            var parameters = DateTimePickerParameters.WithDates(
-                DateTimePickerMode.Date,
-                StartTime,
-                EarliestAllowedStartTime,
-                latestAllowedStartTime);
+            var parameters = IsTimeEntryRunning
```
amulware: This condition could still easily be abstracted away, but if you prefer it this way, I'm not gonna argue.

PR #1518, opened 2018-02-14
Enable bouncing in the reports view

### LINKED ISSUE #1355: Bring back bouncing in scroll views
Make iOS users happy and let the scroll views bounce freely.

That includes the bottom of the timeline, the reports page and the whole settings page.

The pull to refresh should also stretch freely and not get stuck as it does now.

- [X] Main view (with stretching pull-to-refresh)
- [X] Reports
- [ ] Settings
- [x] Start time entry suggestions
- [x] Project list
- [x] Workspace list
- [x] Tags list
- [x] Clients list
- [X] EditView
- [x] Onboarding

simonrozsival:
Ref #1355

Do a functional test when reviewing.

Squash whenever


pe0ny:
It should be grey as the nav bar, so #FAFBFC

simonrozsival:
A side note: we will have to make a big refactoring of the settings view to make it scrollable.

arturs105:
I think putting the whole thing in a `UIScrollView` should work.

PR #1527, opened 2018-02-15
Add tags selection screen

### LINKED ISSUE #1123: Tags view UI
- [x] User can access tags view from the edit view.
- [ ] Show tags only from the workspace that the time entry is assigned to.
- [x] Indicate selected tags with blue checkboxes.
- [x] Use fade for long tag names.
- [x] When a user has typed in tag name that doesn't exist yet, show 'Create tag' option.
- [x] Tapping 'Create tag' creates a new tag and selects it automatically.

### Tags view
tags

Zeplin: https://zpl.io/2Ed87e8

### Tags view: create tag
tags - create tag

Zeplin: https://zpl.io/VDpm8qv

kornelijepetak:
Closes #1123 

Certain things had to be changed in `Daneel` because it was using different kind of `IMvxCommand` which was causing binding errors in `Giskard`. The `SelectTagsViewModel.Text` is bound to the text field and will contain the current value of the tag that is being created. No need for this method to be parametereized.

I have checked the iOS build functionally and there are no functional changes to iOS app. Still, make sure to test both apps.


Toggl.Giskard/Activities/SelectTagsActivity.cs
```diff
@@ -0,0 +1,52 @@
...
+            if (keyCode == Keycode.Back)
+            {
+                ViewModel.CloseCommand.Execute();
+                return true;
+            }
+
+            return base.OnKeyDown(keyCode, e);
+        }
+
+        private void onNavigateBack(object sender, Toolbar.NavigationClickEventArgs e)
```
willsb: This is not used anywhere

Toggl.Giskard/Adapters/SelectTagsRecyclerAdapter.cs
```diff
@@ -0,0 +1,72 @@
...
+using Android.Support.V7.Widget;
+using Android.Views;
+using MvvmCross.Binding.Droid.BindingContext;
+using MvvmCross.Core.ViewModels;
+using MvvmCross.Droid.Support.V7.RecyclerView;
+using Toggl.Giskard.TemplateSelectors;
+
+namespace Toggl.Giskard.Adapters
+{
+    public sealed class SelectTagsRecyclerAdapter : MvxRecyclerAdapter
```
willsb: Is this needed? The code seems identical to an already existing adapter

Toggl.Giskard/Resources/layout/SelectTagsActivityCell.axml
```diff
@@ -19,18 +20,27 @@
...
     <View
         android:background="@android:color/white"
         android:layout_width="24dp"
         android:layout_height="match_parent"
         android:layout_gravity="center_vertical|right" />
-    <CheckBox
-        android:checked="true"
-        android:buttonTint="@color/checkboxTint"
+    <ImageView
+        local:MvxBind="Visibility Visibility(Selected)"
```
willsb: Why are you using ImageViews instead of the checkboxes? These won't animate smoothly

kornelijepetak: I was thinking at the time along the lines of tapping the whole row anyway, but you are right, I forgot about animations.

Issue #1534, opened 2018-02-15
Implement the v9 /me/preferences endpoint


simonrozsival:
We currently have no way of getting user's preferences - mainly regarding the date, time and duration formatting.

The endpoint is `/api/v9/me/preferences`, the response looks like this:
```json
{
    "timeofday_format": "h:mm A",
    "date_format": "MM/DD/YYYY",
    "duration_format": "improved",
    "record_timeline": false,
    "send_product_emails": true,
    "send_weekly_report": true,
    "send_timer_notifications": true,
    "CollapseTimeEntries": true
}
```

- [x] Add a model
- [x] Add singleton repository
- [x] Add the endpoint
- [x] Add HTTP client class (GET, POST)
- [x] Add persisting state (similar to workspace features - override unless the sync state is set to `SyncNeeded`)
- [ ] Add pushing state
- [x] Add reference to the repository to the database
- [x] Add reference to the data source

amulware:
Should we include properties that we don't care about?
Our policy so far has been not to add things we know we won't care about, so I propose we only consider:
```
{
    "timeofday_format": "h:mm A",
    "date_format": "MM/DD/YYYY",
    "duration_format": "improved",
    "CollapseTimeEntries": true
}
```

I believe if we don't include the other properties on pushing, it will not override them either, so that should be fine.

amulware:
There was an internal discussion about using `me/preferences/mobile` (which would have to be created first), with the conclusion that we won't.

amulware:
It occurs to me that we either need to ensure that fetching the preferences always works during signin/signup or we have to have some default settings in the app. Otherwise, the app might get into a state where it always crashes because it tries to access user settings to display data.

The same might be true for workspaces?

simonrozsival:
I think that we should use some default values (e.g., what we use now) in case we couldn't pull user's settings.

What exactly do you have in mind wrt workspaces?

amulware:
Well, we can retry at least, but I'm not sure how often this would even happen.

We can also for now just let it be caught by https://github.com/toggl/mobileapp/issues/1342
(for now == when that is implemented)

simonrozsival:
This could happen if the user was online at the time of logging in but then he lost the Internet connection during pull sync. Or if the endpoint is broken. Or if the server randomly dies for this single request or that request time-outs (but then it should be fixed in the retry loop automatically I believe).

amulware:
[...]

PR #1536, opened 2018-02-16
Preferences model

simonrozsival:
Ref #1534 

This PR removes unnecessary (= non-existing in API or with incorrect values) properties from the User entity and adds a Preferences model, which will be used to get and post user's preferences.

Squash whenever


Toggl.PrimeRadiant.Realm/Models/RealmConstructors.cs
```diff
@@ -275,8 +275,6 @@ public void SetPropertiesFrom(IDatabaseUser entity, Realms.Realm realm)
             DefaultWorkspaceId = entity.DefaultWorkspaceId;
             Email = entity.Email;
             Fullname = entity.Fullname;
-            TimeOfDayFormat = entity.TimeOfDayFormat;
-            DateFormat = entity.DateFormat;
```
amulware: Realm models (and maybe others?) should be updated as well, no? And then we'll need a migration?

Toggl.PrimeRadiant.Realm/Database.cs
```diff
@@ -59,12 +59,12 @@ private Realms.Realm getRealmInstance()
         private RealmConfiguration createRealmConfiguration()
             => new RealmConfiguration
             {
-                SchemaVersion = 1,
+                SchemaVersion = 2,
                 MigrationCallback = (migration, oldSchemaVersion) =>
                 {
-                    if (oldSchemaVersion < 1)
```
amulware: It bothers me that we don't have an easy way to make sure this will always be updated in the future. We could force to have a list of conversion delegates and ensure that it has one for each schema version maybe... but that might also be over-engineering it :D

simonrozsival: I don't understand what is the problem here :D

amulware: I like the code to ensure correctness on our behalf is all. :P

PR #1539, opened 2018-02-16
Add the date format struct

arturs105:
Squash whenever


Toggl.Multivac/DateFormat.cs
```diff
@@ -0,0 +1,49 @@
...
+    {
+        /// <summary>
+        /// Intended for displaying in the UI
+        /// </summary>
+        public string HumanReadableFormat { get; }
+
+        /// <summary>
+        /// Intended for  using in dateTime.ToString(useHere)
+        /// </summary>
+        public string DateToStringFormat { get; }
```
simonrozsival: I think this could be called just `DateFormat`

Toggl.Multivac/DateFormat.cs
```diff
@@ -0,0 +1,49 @@
...
+
+        /// <summary>
+        /// Same as DateToStringFormat, but without the year portion
+        /// </summary>
+        public string ShortTimeToStringFormat { get; }
```
simonrozsival: I think this could be called `ShortDateFormat`

Toggl.Multivac/DateFormat.cs
arturs105: Can't do that. The struct is already named that, so a property can't be named the same, and i don't want to rename the struct to something like `DateTimeFormat`, because construction works only for date formats.

Toggl.Multivac/DateFormat.cs
```diff
@@ -0,0 +1,50 @@
+﻿namespace Toggl.Multivac
+{
+    public struct DateFormat
+    {
+        /// <summary>
+        /// Intended for displaying in the UI
+        /// </summary>
+        public string HumanReadableFormat { get; }
```
simonrozsival: Also I would call this (and the related method) `DatePattern`

arturs105: Although i agree that `HumanReadableFormat` isn't the best name, i don't think that `DatePattern` would better indicate that this property is meant for displaying in UI and not for actual conversions.

What would you think about `LocalizedDateFormat`?

arturs105: Or maybe we could even drop the `DateFormat` part from the properties which would turn this
```
var dateFormat = new DateFormat();
dateFormat.LongDateFormat;
dateFormat.ShortDateFormat;
dateFormat.LocalizedDateFormat;
```
into this:
```
var dateFormat = new DateFormat();
dateFormat.Long;
dateFormat.Short;
dateFormat.Localized;
```

simonrozsival: I like this!

kornelijepetak: I agree, putting type name into the property name is superfluous and reminds me of a country: :hungary: 

I vote for `Long, Short, Localized`.

PR #1547, opened 2018-02-19
Prevent race conditions when updating grouped tableviews

### LINKED ISSUE #1523: Various GroupedCollectionTableViewSource crashes

`

willsb:
Closes #1523 

A thorough functional test by more than one person is strongly requested before we ship this.


simonrozsival:
I did a functional test and all I can say is that this doesn't break it. I can't reproduce the original bug and I definitely can't reproduce the bug with this change. Also, it is very clear what the code does and it can't break anything.

I think this is could be merged into some hotfix branch instead of `develop` and release it to the users - if this really fixes the bug, we will see it as the number of crashes goes down.

I don't want to approve this PR right now, I would like to have some discussion with the other devs if it makes sense to release this even though we probably cannot reproduce the original bug and 100% verify that this fixes the bug.

kornelijepetak:
I can't find anything wrong with the code here, but like I mentioned I was unable to produce the bug even without this change. I am OK with approving, but would also like to know what others think.

The only thing that comes to mind is to force some quick random item generation in underlying data source and see if we can break the thing in the branch before this fix. And then if we can do it consistently (by some subjective certainty though), we apply this fix and try again.
I am afraid I understand what's going on with this bug too little to be able to provide any meaningful way to consistently (as in 100% deterministic) induce this bug.

amulware:
This doesn't so much prevent the race condition but swallows it.

An alternative might be to have a lock around these two blocks of code?

If we can do what @kornelijepetak suggests, throwing maybe 10 threads at it calling these with different parameters and slightly random timing, that would be great.

Otherwise, let's release this as a hotfix this week and see what happens.

simonrozsival:
Please change the base to `hotfix-1.3.3`

Issue #1172, opened 2017-12-18
Add open source licenses to App


willsb:
We **must** add and `About` or `Open source` tab to our app with this info (in the settings, I think)

Please note that it should not rely on the internet, since this info must be available even when the app has no connectivity

pe0ny:
We need a list of licenses we have to add first before we can start with designing this.

willsb:
This is a list of all our projects. We'll hunt licenses (when applicable) tomorrow.

Common:
Fody
Microsoft.Reactive.Testing
MvvmCross
Newtonsoft.Json
PropertyChanged.Fody
Realm
StyleCop.Analyzers
System.Reactive.Core
System.Reactive.Interfaces
System.Reactive.Linq
System.Reactive.PlatformServices

Test projects:
FluentAssertions
FsCheck
FsCheck.Xunit
FSharp.Core
NUnit
NSubstitute
Xamarin.TestCloud.Agent
Xamarin.UITest
xunit

iOS:
Xamarin.1PasswordExtension
Xamarin.Firebase.iOS.Analytics
Xamarin.Firebase.iOS.Core
Xamarin.Firebase.iOS.CrashReporting
Xamarin.Firebase.iOS.InstanceID
Xamarin.Google.iOS.Core
Xamarin.Google.iOS.SignIn

Android: 
Xamarin.Android.Support.Animated.Vector.Drawable
Xamarin.Android.Support.Annotations
Xamarin.Android.Support.Design
Xamarin.Android.Support.Fragment
Xamarin.Android.Support.v7.AppCompat
Xamarin.Android.Support.v7.RecyclerView
Xamarin.Android.Support.Vector.Drawable

pe0ny:
@willsb Should we link those to somewhere as well?

willsb:
Nope. They simply have a plain text license that we need to display somewhere in our app, that's all

amulware:
@pe0ny This still needs a design. I figure having it somewhere in the settings is good. Doesn't have to be prominent, just so that users can find it if they insist. Maybe just behind a button called 'Licenses', or maybe we can find smarter/funnier copy.

amulware:
And @willsb was gonna hunt licenses. ;) Let's see if we can group things so it doesn't look like some crazy long list? I imagine most of the things above will have the same few licenses.

pe0ny:
I would do it so then: List all the projects that have the same license and display the license below it.

## Design Preview
### Licenses
[...]

PR #1549, opened 2018-02-19
Add software licenses provider

willsb:
Ref #1172 

This adds the licenses as a json file. Next step, once this is merged, is to have some kind of service that:

- Reads the common licenses file
- Combines it with the platform specific ones
- Returns the whole thing as a `Dictionary<string, string>` that can be consumed in the UI properly.

Please do this as an interaction, if the initial interactor PR is merged then.

Squash whenever


simonrozsival:
If the JSON file isn't downloaded from the internet but is bundled with the app at compile time, why not just skip the parsing and have directly the `Dictionary<string, string>` (platform specific, of course)? What is the advantage of the JSON in this use case?

Toggl.Foundation/Services/BaseLicenseProvider.cs
```diff
@@ -0,0 +1,40 @@
...
+        };
+
+
```
simonrozsival: Extra new line.

PR #1540, opened 2018-02-16
Api client for /me/preferences

simonrozsival:
Ref #1534 

This PR adds an API client and integration tests for this endpoint.

Please run integration tests when reviewing.

Integration tests in bitrise: https://www.bitrise.io/build/8ebd928ae8712d1b

Squash whenever

PR #1558, opened 2018-02-19
Add preferences repository and data source

simonrozsival:
Ref #1540 

I updated the single object storage so that it doesn't require the entity to have an ID (i.e.,  implement `IBaseModel`). It doesn't really make any sense to use the ID in this repository and the preferences entity doesn't have any at all.

Squash whenever


Toggl.PrimeRadiant.Tests/Realm/RealmSingleObjectStorageTests.cs
```diff
@@ -5,7 +5,7 @@ namespace Toggl.PrimeRadiant.Tests.Realm
     public sealed class RealmSingleObjectStorageTests : SingleObjectStorageTests<TestModel>
     {
         protected override ISingleObjectStorage<TestModel> Storage { get; } 
-            = new SingleObjectStorage<TestModel>(new TestAdapter());
+            = new SingleObjectStorage<TestModel>(new TestAdapter(_ => __ => true));
```
kornelijepetak: Inception. :p

simonrozsival:
We store the ID of the user, nothing changes in that aspect. The change is that whenever we want to access the stored entity, we now won't look at its ID - if there is some object, return it - if there is none, throw. 

The reason is that the Preferences object doesn't have any ID. If this repository should be used (and I am persuaded it should) for storing this type of object, it must not consider any IDs. There is still the `RealmAdapter` class underneath and its methods require passing some `long` value as an "id". It might be possible to edit `RealmAdapter` and add methods like `First`, `UpdateFirst`, `DeleteFirst`, but I didn't want to touch `RealmAdapter` and break things.

If somebody has any opinions, they are welcome. I will leave this PR open for now and merge it tomorrow unless somebody requests changes by then.

Issue #1380, opened 2018-01-30
Add date format setting


amulware:
Part of #1331

Also see #1349

pe0ny:
## Design Preview
### Date format
![toggl - ios - date format 1x](https://user-images.githubusercontent.com/5064445/35963751-61be0260-0cbe-11e8-8478-8a69ab572639.png)

Zeplin: https://zpl.io/VDDGBAe

pe0ny:
- Log titles (today, yesterday etc) is not affected by this instead, we use this format: Tue, 13 Feb
- Edit view and the wheel view are affected by this setting.
- Also, calendar in reports will be affected by this setting (and let's do what webapp does and cut the year part so it will be shorter)

arturs105:
Is the year cut only for the calendar in reports or in other places as well?

pe0ny:
Let's cut the year for the calendar in reports and dates in the wheel view. in the edit view, let's keep the year.

PR #1566, opened 2018-02-21
Add date format setting

arturs105:
Ref #1380

The setting is implemented, but hidden for now, since it's not working properly without syncing.

Squash whenever


simonrozsival:
In principle, this works well and gets the job done. My concern is that we need to add more similar selection views (duration, first day of week, timezone) and we will add another ~700 LOC for each of these, when this could be done as a generic VM which would take a parameter with a title + a `Dictionary<T, string>` where the key would be returned when the user selects the row with the value displayed (like any other select view) + the currently selected item.

Do you think we should keep making specific VMs + views + table sources + table cells which will be almost the same or will we make a generic view from this (we could then use it for selecting default workspace as well)?

PR #1586, opened 2018-02-23
Extract task expandable project cell view holder

kornelijepetak:
I have noticed that both Start TE and Project selection in EditView need this view holder so I have pulled it out to be reused. Basically the only difference is their View ID.

One file (project cell) is required for this to work (because of Resource IDs). It is a part of #1584 which should be merged first (and soon).

Squash whenever.


Toggl.Giskard/Views/SelectProjectRecyclerViewHolder.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System;
+using Android.Runtime;
+using Android.Views;
+using MvvmCross.Binding.Droid.BindingContext;
+
+namespace Toggl.Giskard.Views
+{
+    public sealed class SelectProjectRecyclerViewHolder : SelectProjectWithExpandableTasksRecyclerViewHolder
+    {
+        public override int ToggleTasksExpansionButtonId => Resource.Id.SelectProjectToggleTasksButton;
```
willsb: Why add inheritance when this could be made mutable (`{ get; set; }`) and then set during construction, like the Commands? (See example in the StartTimeEntryRecyclerAdapter)

kornelijepetak: Yeah makes sense. I was thinking along the lines of being able to extend specific viewholder when the need arises, but I agree it's better to think about it when we need it.

PR #1589, opened 2018-02-26
Add workspace list to settings

### LINKED ISSUE #1382: Allow switching default workspace
Part of #1268

willsb:
Closes  #1382 

Squash whenever


Toggl.Giskard/Resources/layout/SettingsActivity.axml
```diff
@@ -36,6 +36,28 @@
                 android:layout_marginBottom="40dp"
                 android:layout_width="match_parent"
                 android:layout_height="wrap_content" />
+            <TextView
+                android:textSize="12sp"
+                android:paddingLeft="16dp"
+                android:textAllCaps="true"
+                android:letterSpacing="0.04"
+                android:text="@string/DefaultWorkspace"
```
kornelijepetak: Williamize this order! 😄

Toggl.Foundation.MvvmCross/ViewModels/SettingsViewModel.cs
```diff
@@ -290,5 +302,17 @@ private void OnIsManualModeEnabledChanged()
             var newPreferences = await dataSource.Preferences.Update(preferencesDto);
             DateFormat = newPreferences.DateFormat;
         }
+
+        private async Task changeWorkspace(long selectedWorkspaceId)
```
kornelijepetak: I think it's better if you move this method rigth below `selectWorkspace` method so the related functions are close to each other. More fluent readability.

kornelijepetak: Also it would be a bit clearer if this method was called `changeDefaultWorkspace`. The same with other things (`editDefaultWorkspace`, `EditDefaultWorkspaceCommand`).

`EditWorkspaceCommand` makes you think it's the workspace you are editing, and not the user settings. `Select` would IMO be better than `Edit` in this case.

Maybe I'm overreacting, but I think it would be much clearer. Think about it and act appropriately. Whatever you choose, it will no longer cause PR change requests. Other remarks still hold.

Issue #1378, opened 2018-01-30
Add duration display format to settings


amulware:
Part of #1331

Also see #1349

pe0ny:
## Design Preview
### Duration display format
![toggl - ios - duration display format 1x](https://user-images.githubusercontent.com/5064445/35963788-8dfb5bca-0cbe-11e8-87c2-bd1c5299eee7.png)

Zeplin: https://zpl.io/adA4pzl

PR #1597, opened 2018-02-28
Parametric duration

simonrozsival:
Ref #1378 

This PR does multiple things:
- introducing the _decimal_ format
- moving conversion of time span to string to an extension method
- time span converters accept parameters
- introducing a value combiner to bind duration + duration format
- inject format into time entries log table and into reports legend table

I don't think it would be easy to split this into multiple PRs, but suggestions are welcome.

WIP: I need to make sure this does not break Android.

Squash whenever


simonrozsival:
I will split this PR into several smaller ones.

PR #1601, opened 2018-03-01
Parametric duration formatting

simonrozsival:
Ref #1378 

This PR adds an extension method which converts a `TimeSpan` to string with the given format. This extension method is then used in a new converter and new combiner for easy usage in bindings.

Squash whenever


Toggl.Foundation.MvvmCross/Combiners/DurationValueCombiner.cs
```diff
@@ -0,0 +1,53 @@
...
+using Toggl.Multivac;
+
+namespace Toggl.Foundation.MvvmCross.Converters
+{
+    [Preserve(AllMembers = true)]
+    public sealed class DurationValueCombiner : MvxValueCombiner
+    {
+        public override bool TryGetValue(IEnumerable<IMvxSourceStep> steps, out object value)
+        {
+            TimeSpan duration;
```
kornelijepetak: Explicit access modifiers.

Toggl.Foundation.MvvmCross/Combiners/DurationValueCombiner.cs
```diff
@@ -0,0 +1,53 @@
...
+        private (TimeSpan, DurationFormat) getParameters(IEnumerable<IMvxSourceStep> steps)
+        {
+            var parameters = steps.ToArray();
+
+            if (parameters.Length != 2)
+                throw new ArgumentException(
+                    $"The {nameof(DurationValueCombiner)} must be used with 2 parameters: a TimeSpan and a DurationFormat. It was instead used with {parameters.Length} parameters.");
+
+            var duration = parameters[0].GetValue() as TimeSpan?;
+            if (duration.HasValue == false)
```
kornelijepetak: If you don't want to use `!`, I think `duration == null` would be more readable/intuitive.

Toggl.Foundation.MvvmCross/Combiners/DurationValueCombiner.cs
```diff
@@ -0,0 +1,53 @@
...
+                throw new ArgumentException($"The first parameter of the {nameof(DurationValueCombiner)} must be a {nameof(TimeSpan)}, but the given parameter is of type {parameters[0].SourceType.FullName}");
+
+            var format = parameters[1].GetValue() as DurationFormat?;
+            if (format.HasValue == false)
```
kornelijepetak: ibid.

Toggl.Foundation.MvvmCross/Converters/ParametricTimeSpanToDurationValueConverter.cs
```diff
@@ -0,0 +1,21 @@
...
+using Toggl.Foundation.Extensions;
+using Toggl.Multivac;
+
+namespace Toggl.Foundation.MvvmCross.Converters
+{
+    public sealed class ParametricTimeSpanToDurationValueConverter : MvxValueConverter<TimeSpan, string>
+    {
+        protected override string Convert(TimeSpan value, Type targetType, object parameter, CultureInfo culture)
+        {
+            if (parameter == null || parameter is DurationFormat == false)
```
kornelijepetak: `parameter == null` is not needed here. If `parameter` is `null`, `parameter is DurationFormat` will yield `false` anyway.

Toggl.Foundation/Extensions/TimeSpanExtensions.cs
```diff
@@ -0,0 +1,41 @@
...
+namespace Toggl.Foundation.Extensions
+{
+    public static class TimeSpanExtensions
+    {
+        public static string ToFormattedString(this TimeSpan duration, DurationFormat format)
+        {
+            switch (format)
+            {
+                case DurationFormat.Classic:
+                    return convertToClassicFormat(duration);
```
kornelijepetak: For those methods that have a single statement body, why not inline?

Toggl.Foundation/Extensions/TimeSpanExtensions.cs
```diff
@@ -0,0 +1,41 @@
...
+                case DurationFormat.Decimal:
+                    return convertToDecimalFormat(duration);
+                default:
+                    throw new ArgumentOutOfRangeException(
+                        $"The duration converter parameter '{format}' is not of the supported formats.");
+            }
+        }
+
+        private static string convertToDecimalFormat(TimeSpan value)
+            => $@"{value.TotalHours:00.00} {Resources.UnitHour}";
```
kornelijepetak: `@` is not needed here.

Toggl.Foundation.MvvmCross/Combiners/DurationValueCombiner.cs
simonrozsival: This is a declaration inside of a method.

Toggl.Foundation/Extensions/TimeSpanExtensions.cs
simonrozsival: I think it is better to name the conversions and put them in separate functions - it makes the code easier to understand I think. Also, since they are very similar, it should be harder to break it with future changes this way.

kornelijepetak: That is alright.

Toggl.Foundation.MvvmCross/Combiners/DurationValueCombiner.cs
kornelijepetak: Man, I am blind.

PR #1604, opened 2018-03-01
Hotfix 1.3.4

simonrozsival:
Change log:

- Fixes a problem with the time entries' log which caused a crash.

### DIFF
--- a/Toggl.Daneel/Info.plist
+++ b/Toggl.Daneel/Info.plist
@@ -34,7 +34,7 @@
 	<key>CFBundleName</key>
 	<string>Toggl</string>
 	<key>CFBundleShortVersionString</key>
-	<string>1.3.3</string>
+	<string>1.3.4</string>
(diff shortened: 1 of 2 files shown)


simonrozsival:
- this hotfix doesn't fix the issue (I experienced a crash caused by `UITableView.g.cs`)
- there is a different issue that this hotfix introduces; steps to reproduce:
    - sync the app with the server
    - put the device into flight mode
    - stop the TE in the mobile app, start a different TE A
    - stop the TE in the web app, start a different TE B
    - disable flight mode
    - pull to refresh
    - now the correct behavior would be to see the TE B shown in both the web app and the mobile app, but instead, the mobile doesn't show any running TE and the play button is semi-transparent and cannot be clicked (a TE behind it will be continued)
    - another pull-to-refresh gesture fixes the state of the app

Due to these two reasons, I don't think we should release this version to the users.

amulware:
Closing for the above reasons.

Issue #1349, opened 2018-01-26
Use user settings for formatting times and dates


amulware:
We may want to cache this or something for performance when rendering time entries, etc.?

The setting should be respected everywhere in the app. Maybe a list should be made.

simonrozsival:
These are the places where we display a time (not duration) I am aware of:

- The edit time entry view (start time, in the future maybe the stop time)
- ~The duration view~
- The wheel view (start and stop time + the date picker)
- The select start time view (date picker)

All of these places show the date and time side by side (e.g., "Mon, 22 Jan 16:00").

simonrozsival:
Dates are also shown in the headers of the time entries log.

simonrozsival:
Technically, what we need to do is to inject the user's settings into the `DateToTitleStringValueConverter` class and the `DateTimeToTimeValueConverter` class.

Both of these have the formats hardcoded in: `$"{value:ddd, dd MMM}"` and `.ToString("HH:mm")` respectively.

simonrozsival:
☝️ this means that we don't need two tasks for this - all of the code will be shared by the platforms.

PR #1605, opened 2018-03-01
Current preferences observable

simonrozsival:
Ref #1378, Ref #1349 

This PR adds an observable which provides current user's preferences (emits new values when it was changed in the app or when new preferences were pulled from the server).

Squash whenever


Toggl.Foundation/DataSources/PreferencesDataSource.cs
```diff
@@ -10,13 +14,30 @@ namespace Toggl.Foundation.DataSources
 {
     public sealed class PreferencesDataSource : IPreferencesSource
     {
-        private ISingleObjectStorage<IDatabasePreferences> databaseStorage;
+        private readonly ISingleObjectStorage<IDatabasePreferences> databaseStorage;
+        private readonly ISubject<IDatabasePreferences> currentPreferencesSubject;
+    
+        public IObservable<IDatabasePreferences> CurrentPreferences { get; }
```
kornelijepetak: This will be accessed mostly through the `dataSource.Preferences.*` so it would be easier on the eye if it was only called `Current`, similar to the `User`.

Toggl.Foundation/DataSources/Interfaces/IPreferencesSource.cs
```diff
@@ -1,11 +1,14 @@
 ﻿using System;
 using Toggl.Foundation.DTOs;
+using Toggl.PrimeRadiant;
 using Toggl.PrimeRadiant.Models;
 
 namespace Toggl.Foundation.DataSources
 {
-    public interface IPreferencesSource
+    public interface IPreferencesSource : IRepository<IDatabasePreferences>
```
arturs105: Why does this inherit from `IRepository`? I don't think it makes sense for preferences to have `GetById(id)` or `GetAll()`

simonrozsival: This is because of sync - when new preferences are pulled, the UI can be updated. This is similar to what we do with time entries. Also we need `GetAll(pred)` for push sync. The other methods could probably throw `NotImplementedException`... but I am not sure if that would be a good thing.

PR #1608, opened 2018-03-01
Refactor custom modal dialogs

### LINKED ISSUE #1269: Ask confirmation when deleting TE
- [x] Ask confirmation when user wants to delete existing time entry (taps on a "Delete this entry") from the edit view.
- [x] 'Cancel' will simply close the dialog.
- [x] 'Delete' will delete the time entry and direct user to the timer page.

### Delete confirmation
edit delete

Zeplin: https://zpl.io/VxKXlAw

### LINKED ISSUE #1270: Ask confirmation when discarding TE
- [x] Ask confirmation when user wants to discard a time entry in TE creation view (taps on a back icon at the top).
- [x] Show this only when user has entered something in the field.
- [x] 'Cancel' will simply close the dialog.
- [x] 'Discard' will discard the time entry and direct user to the timer page.

### Discard confirmation
new time entry - discard

Zeplin: https://zpl.io/VYW7R9L

willsb:
Closes #1269
Closes #1270 

The current way wasn’t playing nice with droid, so now instead we are no longer sharing the text (since it’s different for each platform).

The change is minimal since the whole logic was already implemented by the VM; I just had to implement the methods on the Droid service.

Squash whenever


Toggl.Daneel/Services/DialogService.cs
```diff
@@ -41,27 +42,26 @@ public DialogService(ITopViewControllerProvider topViewControllerProvider)
             return tcs.Task;
         }
 
-        public Task<string> ShowMultipleChoiceDialog(
-            string cancelText,
-            params MultipleChoiceDialogAction[] actions)
+        public Task<bool> ConfirmDestructiveAction(ActionType type)
```
simonrozsival: I think it was better when the confirm and cancel buttons' text was passed directly. I think we don't need any new enum, this constructor might simply take two string params.

willsb: The text on both platforms is completely different

PR #1634, opened 2018-03-02
Convert from realm object to foundation object in preferences data source

simonrozsival:
Squash ASAP


simonrozsival:
The problem is that if you get an object from the Realm database, you can't use it from a different thread than the one which did the read transaction. I forgot to copy the values from the Realm object in several methods of the data source, now it can be returned to a different thread and it won't throw an exception.

PR #1642, opened 2018-03-05
Push date format change

simonrozsival:
Ref #1380 

This PR initiates push sync whenever the user changes the date format.

Squash whenever


Toggl.Foundation.MvvmCross/ViewModels/SettingsViewModel.cs
```diff
@@ -331,6 +331,7 @@ private void OnIsManualModeEnabledChanged()
             
             var preferencesDto = new EditPreferencesDTO { DateFormat = newDateFormat };
             var newPreferences = await dataSource.Preferences.Update(preferencesDto);
+            await dataSource.SyncManager.PushSync();
```
amulware: Awaiting this will wait for the push sync to complete, which might take a long time. Should this not be fire-and-forget?

simonrozsival: Yeah, you are right, this might not be a good idea. I will change it.

amulware:
I note that the lines
```
var newPreferences = await dataSource.Preferences.Update(preferencesDto);
await dataSource.SyncManager.PushSync();
```
repeat a lot in this file.

I'd suggest extracting them into a method, and/or removing the `await` from all of them.

amulware:
Maybe you even want that function to take a mutator `Action<EditPreferencesDTO>` or `Func< EditPreferencesDTO, EditPreferencesDTO>` instead of taking the DTO itself, to also share the DTO code.

PR #1645, opened 2018-03-05
New droid timer view

willsb:
:shipit:


willsb:
Note that this introduces some initial friction when scrolling. This will have to be handled in a different PR. Optimization, yay

PR #1648, opened 2018-03-05
Clear backstack when logging in/out

willsb:
Before: You can hit the back button and see the login page/onboarding after logging in. You can see the settings/log after logging out.

Now: The backstack behaves appropriately

Squash whenever.


Toggl.Giskard/Presenters/TogglPresenter.cs
```diff
@@ -1,20 +1,41 @@
...
+            var intent = base.CreateIntentForRequest(request);
+
+            if (clearBackStackTypes.Contains(request.ViewModelType))
+            {
+                intent.SetFlags(ActivityFlags.ClearTop | ActivityFlags.ClearTask | ActivityFlags.NewTask);
+            }
+
+            return intent;
+        }
+
```
kornelijepetak: One `\n` too much

Toggl.Giskard/Presenters/TogglPresenter.cs
```diff
@@ -1,20 +1,41 @@
...
 using MvvmCross.Droid.Support.V7.AppCompat;
+using Toggl.Foundation.MvvmCross.ViewModels;
 using Toggl.Foundation.MvvmCross.ViewModels.Hints;
 using Toggl.Giskard.Activities;
 
 namespace Toggl.Giskard.Presenters
 {
     public sealed class TogglPresenter : MvxAppCompatViewPresenter
     {
+        private readonly HashSet<Type> clearBackStackTypes = new HashSet<Type>
```
kornelijepetak: This is a number of types small enough that the array would suffice, and arrays are IMO a bit easier on the eyes.
No need to change, just saying.

willsb: I like my O(1), so I'm keeping this 💃

kornelijepetak: Sure. Like I said, for small number it's irrelevant. Not only that. This happens so rarely that we already talked too much about it. 😄

PR #1657, opened 2018-03-06
Detect that user already has an account during sign up 

### LINKED ISSUE #1221: Detect that user already has an account during sign up
When user tries to create a new account for an email address which is already used by some account, we receive a 200 response from the backend with the response `{"has_account":"true"}`.

We should display some better copy for the error msg in this case - please provide some text @pe0ny or @amulware.

simonrozsival:
Closes #1221 

This PR introduces a new exception which will be thrown when backend returns 400 response with a specific error message.

This PR shouldn't break Android - it will show the `User with this email already exists.` message, but it won't show the `Try logging in instead.` button. This will have to be done by someone from the Android team.

Squash whenever


Toggl.Ultrawave/ApiClients/UserApi.cs
```diff
@@ -55,7 +58,11 @@ public IObservable<IUser> SignUp(Email email, Password password)
                 }
             };
             var json = serializer.Serialize(dto, SerializationReason.Post, null);
-            return CreateObservable<User>(endPoints.Post, new HttpHeader[0], json);
+            return CreateObservable<User>(endPoints.Post, new HttpHeader[0], json)
+                .Catch<IUser, BadRequestException>(badRequestException
+                    => badRequestException.LocalizedApiErrorMessage == userAlreadyExistsApiErrorMessage
```
amulware: Is this the only way of checking this? If so, I don't know if I can be ok with that. In the very least this needs to be tested in the integration test.

simonrozsival: I couldn't find a better way of detecting this. The response is 400 and it has plaintext body with this string constant.

There is already an integration test for this (`UserApiTests.cs:241`).

amulware: Ah I see, I didn't understand that the change to the test actually includes testing that the string comparison works.

Toggl.Ultrawave/Exceptions/EmailIsAlreadyUsedException.cs
```diff
@@ -0,0 +1,13 @@
+﻿namespace Toggl.Ultrawave.Exceptions
+{
+    public sealed class EmailIsAlreadyUsedException : ClientErrorException
+    {
+        private readonly BadRequestException actualError;
+
+        public EmailIsAlreadyUsedException(BadRequestException badRequestException)
+            : base(badRequestException.Request, badRequestException.Response, badRequestException.Message)
+        {
+            actualError = badRequestException;
```
amulware: What is it keeping this for? Consider assigning to `InnerException` or removing the field.

Toggl.Ultrawave/Exceptions/ApiException.cs
```diff
@@ -7,28 +7,28 @@ namespace Toggl.Ultrawave.Exceptions
 {
     public class ApiException : Exception
     {
-        private readonly IRequest request;
+        internal IRequest Request { get; }
```
willsb: 🤮

simonrozsival:
The title changes from "Sign up" to "Log in" - I didn't have to change anything, it just worked.

PR #1660, opened 2018-03-06
Auto-bump versions for AdHoc/AppStore releases

### LINKED ISSUE #1475: Have cake fill in number of commits as build number
If we do this, we no longer have to increase build numbers manually, which makes releasing easier, and allows us to have automatically scheduled testflight builds as well.

For help, consider asking the superday devs how they did it.

willsb:
Closes #1475 

We can’t use `agvtool` because this is not an Xcode project, but it’s not that useful anyways so this is fine.

We are currently parsing the XML, transforming it and then saving the changes. Since the XML itself has a weird structure, I’m just getting the last node and changing its inner text to be the number of commits. Pretty straight forward.

Since the file is already being transformed to include env var secrets, it’ll only work in flows that ignore the teardown. Which is fine, since this is not something we want to commit (and the flows that ship something need to ignore teardown because they use the env vars).

Squash whenever


simonrozsival:
Why not just do it every single time and use a simple substitution? This seems unnecessarily complicated.

amulware:
What bothers me is that the xml of that file changes sometimes. I.e. VS or a different program might decide to reorder the file. In that case, just taking the last node will break. Can't we instead have a replacement string in the file?

PR #1661, opened 2018-03-06
Check for api tokens existing in integration tests

### LINKED ISSUE #1652: Throw in integration tests if signup does not return valid API token
This is a thing that used to happen too much in v8, and might still be happening in v9. At least that would explain why we are getting some random 403 errors. Given how we use the api, the signup endpoint not returning an API token is a major issue, so let's catch that and throw with info from the signup request (ideally the response headers if we can get them) for easier debugging.

amulware:
Closes #1652

~~Except it's not printing request ids. But I don't think that's possible without hacking it or making a hardcoded http call instead of our api call to fetch the user.~~

~~Ideas welcome. Otherwise, let's merge this since it is better than nothing and will improve debugging at least a tiny bit.~~

EDIT: See updated info below: https://github.com/toggl/mobileapp/pull/1661#issuecomment-371413410


simonrozsival:
I think we might want to throw directly in the API client so that we can handle this in the app if this happens (we know user's email and password, so we can automatically perform login in the background without the user noticing).

I am not really sure what the app does when this happens... it probably shows an error because initial sync failed and from then on it shows just the error message, because the email is already taken. When #1221 is merged then it will show a better error message and if the user logged in with the email and the pwd it would work, but we could do that without him/her noticing as I mentioned earlier.

amulware:
That actually makes a lot of sense. I'll make those changes and have the API throw a custom exception.

amulware:
I did a pretty big refactor of the base api, because the main method got really really long (see https://github.com/toggl/mobileapp/pull/1661/commits/a67380521fee2c2d6a404ac4fbc947cc8ef2ff6b#diff-fa0df3bd95aad82131dd342cb51a958bR78 )

Tests still pass (excluding one of the new ones right now - I'll fix that), including integration tests: https://www.bitrise.io/build/afbf93e951e9370b

amulware:
Things should be working now. 🤞

What it's doing:
- api clients can now inject a data validation action into the base client which is called after deserialisation
- if that action throws, it's exception is wrapped in an `InvalidDataException` which is passed on to the observable
- the user api uses this to throw a `UserIsMissingApiTokenException` if the api ever returns a user object without an api token

To check the functional addition of this PR, see these commits: a673805 (validator), dad7bb7 (checking api token)
To check out the base api refactor, see this commit: 4eb3684

Otherwise, I recommend looking at the complete changeset.

If you rather have me split this up into a refactor and a feature PR, I can do so, but I don't think that that will help much with making it more readable. Up to you though.

Possible point of improvement:
- we could skip the rethrow with the `InvalidDataException`, which might be nice since that one does not inherit from our own `ApiException`. Or maybe we want to replace it with an `InvalidDataApiException` or something?

TODO (future PRs, will create issues):
- catch this exception in all login/signup flows, and retry fetching the user once or twice with a short delay to try and get an API token
- if an API token really cannot be gotten, the login/signup has to be considered failed, since without the token we can't do anything

amulware:
Another integration tests run for good measure: https://www.bitrise.io/build/5153f9115bac0387

simonrozsival:
We currently don't use this in any sync API request but if we would, how would this `InvalidDataException` be handled? It is not an `ApiException`, so it will be rethrown by the state and caught by the `StateMachine` and we will display the "Sync failed" status to the user. Is that the correct behavior? Shouldn't it be handled like a server error and enter retry loop when the server response is invalid?

Toggl.Ultrawave.Tests/ApiClients/UserApiTests.cs
```diff
@@ -0,0 +1,139 @@
...
+            private readonly UserApi api;
+            
+            public Base()
+            {
+                api = new UserApi(
+                    new Endpoints(ApiEnvironment.Staging),
+                    apiClient,
+                    jsonSerializer,
+                    credentials
+                    );
```
simonrozsival: Bad `)` placement.

Toggl.Ultrawave/Exceptions/UserIsMissingApiTokenException.cs
```diff
@@ -0,0 +1,14 @@
+﻿using Toggl.Ultrawave.Network;
+
+namespace Toggl.Ultrawave.Exceptions
+{
+    public class UserIsMissingApiTokenException : ApiException
```
simonrozsival: Shouldn't this inherit from `ServerErrorException`?

amulware:
@simonrozsival Should I completely skip the `InvalidDataException` rethrow?

It sort of makes sense to me to use an exception like that since it represents a category of errors.
But it might make our exception handling a bit more complicated.

On the other hand, not having that exception means that we have to also think carefully what invalid data cases (if we'll ever have more than this single one) fall under client/server/other errors.

Toggl.Ultrawave/Exceptions/UserIsMissingApiTokenException.cs
amulware: Note to self: This should also be sealed.

simonrozsival:
I would prefer skipping the rethrow. It would be even easier for the consumer - he would know that he wants to catch the exception he threw, not some `InvalidDataException`.

amulware:
It would be easier, true. But the grouping would no longer be possible, and it's perfectly possible to catch an exception only if it has a given inner exception. But I digress. We can solve the grouping with inheritance as well.

My key concern:

If we make this a server exception, and it enters the retry loop, it will find the status returning 200, and will then retry the api after a minute, correct? That would make login/signup take a whole minute, no? 

That doesn't seem ideal. Though it's probably better than failing, and we can make a special case for this exception soon? Hmm...

Can you confirm that I understand correctly that we'll enter a slow api retry and will delay signup/whatever by one minute if we make it a server exception, @simonrozsival?

amulware:
I confused multiple things. There is no retry loop doing signup. What I said above applies to other syncs, which is maybe not ideal, but it should also never happen. And if it does, waiting a minute is a minor evil compared to saving a user without api token which will make all future requests blow up/might log the user out.

Toggl.Ultrawave.Tests/ApiClients/UserApiTests.cs
simonrozsival: We normally place the closing bracket on the same line as the last argument of the function. This doesn't seem consistent with the rest of the codebase.

Toggl.Ultrawave/ApiClients/BaseApi.cs
```diff
@@ -55,54 +57,52 @@ protected BaseApi(IApiClient apiClient, IJsonSerializer serializer, Credentials
...
-                }
+                    await throwIfRequestFailed(request, response, headerList).ConfigureAwait(false);
 
-                T data;
-                try
-                {
-                    data = await process(response.RawData);
+                    data = await processResponseData(request, response, process).ConfigureAwait(false);
+
+                    validateResponse(request, response, data, responseValidator);
```
simonrozsival: I think invoking the validator directly here would be better - there is no try-catch in the static method, so I don't see a need for any extra method.

amulware: I think there is a lot going on with the `?.` and `Invoke()`. I much rather give this a name and forget about the details.

Toggl.Ultrawave.Tests/ApiClients/UserApiTests.cs
amulware: I think that it is much nicer this way, but I'm gonna change it for consistency.

Toggl.Ultrawave/ApiClients/BaseApi.cs
simonrozsival: OK

PR #1675, opened 2018-03-07
Support for multiple bundle ids

### LINKED ISSUE #1476: Have separate app-ids for debug/adhoc/production builds
Maybe separate into three different app ids?

We just need to figure out how to have a different app id based on the build config.

I propose:
```
com.toggl.daneel.adhoc
com.toggl.daneel.debug
```

Ideally, they have different app icons as well (@Odrakir has amazing shopping skills for this). And maybe we can even show a different Toggl logo in the app, so that we always know what we are looking at?

Just some ideas.

willsb:
Closes #1476 

TODO:
- [ ] Add pretty icons -> @pe0ny 
- [x] Create provisioning profiles -> @amulware 

However, the bulk of the code won’t change, so feel free to review in the meantime to speed things up.


amulware:
I generated both new app ids, and a provisioning profile for the adhoc one. That profile is already on Bitrise, so adhoc builds there should fail until this is merged.

amulware:
I think I prefer this implementation. It's a lot less code, pretty straightforward and by not using templates nothing can go wrong with those. It's still not 'nice' but I'm don't think there is a nice way of doing this, given that we use Xamarin.

amulware:
Maybe @simonrozsival wants to take a look as well since he was arguing for doing this quite strongly?

amulware:
I say merge this and make an issue for the icons. It should work just fine the way it is, and the different names will be clear enough.

PR #1678, opened 2018-03-08
Queue updated user data for push sync

simonrozsival:
We must change the sync status of the user so it can be pushed.


amulware:
What about the other syncing related properties? What about the `At` property?

Toggl.Foundation.Tests/DataSources/UserDataSourceTests.cs
```diff
@@ -72,11 +72,15 @@ public void DoesNotUseCachedVersion()
 
         public sealed class TheConstructor
         {
-            [Fact, LogIfTooSlow]
-            public void ThrowsIfTheArgumentIsNull()
+            [Theory, LogIfTooSlow]
+            [ClassData(typeof(TwoParameterConstructorTestData))]
+            public void ThrowsIfTheArgumentIsNull(bool useRepository, bool useTimeServide)
```
amulware: rename to `ThrowIfAnyArgumentIsNull`?

Issue #1529, opened 2018-02-15
Finish 3d touch shortcuts


amulware:
They are not yet handled properly. Let's do that.

PR #1689, opened 2018-03-09
Shortcuts handler

simonrozsival:
Ref #1529 

This PR adds basics of handling shortcuts or any links with the `toggl://` scheme (this can be later easily used for implementing `x-calblack-url`). Handling actual links will be part of a different PR.

_How to test:_ Set a breakpoint in `TogglApplicationShortcutsHandlingService.cs` on the line `21`, run the app, open Safari on the device/simulator, type `toggl://whatever?x=y` into the address field, you should see the parsed shortcut in the debugger.

Squash whenever


Toggl.Daneel/Info.plist
```diff
@@ -29,6 +29,7 @@
 	<true/>
 	<key>LSApplicationQueriesSchemes</key>
 	<array>
+		<string>toggl</string>
```
kornelijepetak: Tabz here. 
Could you please check whether manually changing it to tabs will make VS respect the whitespace after reediting the `.plist` from the editor. If not, I propose we ignore whitespace in these files when reviewing.

Toggl.Foundation/Shortcuts/ShortcutUrl.cs
```diff
@@ -0,0 +1,93 @@
...
+        public const string Scheme = "toggl";
+
+        public string Action { get; }
+
+        public Dictionary<string, string> Parameters { get; }
+
+        public static ShortcutUrl Invalid { get; } = FromAction(String.Empty);
+
+        private string parametersString
+            => Parameters == null || Parameters.Count == 0
```
kornelijepetak: What do you think about adding `IEnumerable<T>.IsNullOrEmpty` extension for situations like these? This types of check happen rather often IMO.

Toggl.Foundation/Shortcuts/ShortcutUrl.cs
```diff
@@ -0,0 +1,93 @@
...
+                ? String.Empty
+                : "?" + Parameters
+                    .Select(pair => $"{pair.Key}={pair.Value}")
+                    .Aggregate("", (acc, pair) => $"{acc}&{pair}")
```
kornelijepetak: Prefer `string.Join("&", pairs)` in this case. It is faster, and more readable than `Aggregate` and does not require Trim at the end.

I would also strongly advise creating `Join(this IEnumerable<string>, string separator)` extension method that would enable **`fluent`** continuation of the expression to avoid deep function compositions.

Toggl.Foundation/Shortcuts/ShortcutUrl.cs
```diff
@@ -0,0 +1,93 @@
...
+            var parameters = new Dictionary<string, string>();
+
+            try
+            {
+                urlWithoutScheme
+                    .Substring(questionMarkIndex + 1)
+                    .Split('&')
+                    .Select(pair => pair.Split('='))
+                    .Do(throwIfNotAPair)
+                    .ForEach(pair => parameters.Add(pair[0], pair[1]));
```
kornelijepetak: I'd rather go with `parameters[pair[0]] = pair[1]` unless you explicitly want this to throw if there are multiple occurrences of parameters.

Toggl.Foundation/Shortcuts/ShortcutUrl.cs
```diff
@@ -0,0 +1,93 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using Toggl.Multivac.Extensions;
+
+namespace Toggl.Foundation.Shortcuts
+{
+    public struct ShortcutUrl
+    {
+        public const string Scheme = "toggl";
```
arturs105: This isn't used anywhere outside this class, so it could be made private.

Toggl.Foundation/Shortcuts/ShortcutUrl.cs
```diff
@@ -0,0 +1,93 @@
...
+                    .Select(pair => $"{pair.Key}={pair.Value}")
+                    .Aggregate("", (acc, pair) => $"{acc}&{pair}")
+                    .Trim('&');
+
+        private const string prefix = Scheme + "://";
+
+        private ShortcutUrl(string action, Dictionary<string, string> parameters)
+        {
+            this.Action = action;
+            this.Parameters = parameters;
```
arturs105: `this` not necessary.

Toggl.Foundation/Shortcuts/ShortcutUrl.cs
```diff
@@ -0,0 +1,93 @@
...
+
+        private string parametersString
+            => Parameters == null || Parameters.Count == 0
+                ? String.Empty
+                : "?" + Parameters
+                    .Select(pair => $"{pair.Key}={pair.Value}")
+                    .Aggregate("", (acc, pair) => $"{acc}&{pair}")
+                    .Trim('&');
+
+        private const string prefix = Scheme + "://";
```
arturs105: `$"{Scheme}://"`

Toggl.Foundation/Shortcuts/ShortcutUrl.cs
```diff
@@ -0,0 +1,93 @@
...
+
+        private ShortcutUrl(string action, Dictionary<string, string> parameters)
+        {
+            this.Action = action;
+            this.Parameters = parameters;
+        }
+
+        public override string ToString() => prefix + Action + parametersString;
```
arturs105: `$"{prefix}{action}{parametersString}"`

Toggl.Daneel/Startup/AppDelegate.cs
```diff
@@ -63,5 +66,10 @@ public override void DidEnterBackground(UIApplication application)
             base.DidEnterBackground(application);
             backgroundService.EnterBackground();
         }
+
+        public override bool HandleOpenURL(UIApplication application, NSUrl url)
```
willsb: What's the difference between this method and the `OpenUrl` method? We already override `OpenUrl`, but it's currently not compiled because of pre-processor directives. It's used by the google login.

Toggl.Foundation.Tests/Services/TogglApplicationShortcutsHandlingServiceTests.cs
```diff
@@ -0,0 +1,83 @@
...
+
+                var result = handlingService.TryHandle(ShortcutUrl.Invalid);
+
+                result.Should().BeFalse();
+            }
+
+            [Property]
+            public void ReturnsTrueIfAnyOfTheInnerHandlersSucceds(byte numberOfHandlers)
+            {
+                if (numberOfHandlers == 0) return;
```
willsb: You can configure this with the `StartSize` property of the `PropertyAttribute`

Toggl.Foundation/Shortcuts/ShortcutUrl.cs
```diff
@@ -0,0 +1,93 @@
...
+
+        public static ShortcutUrl FromAction(string action)
+            => FromActionWithParams(action, null);
+
+        public static ShortcutUrl FromActionWithParams(string action, Dictionary<string, string> parameters)
+        {
+            if (action.Contains("://"))
```
willsb: So I can't have a time entry whose description is `Fixing http://toggl.com`? :P

simonrozsival:
I briefly read the docs of MVVM Uri routing which I wasn't aware of and it seems like all of the parsing in the PR could be removed + the service is also not relevant. Maybe even the `ShortcutUrl` type could be removed completely for generating the URLs. I will think about it and either close this PR as a whole or I will 🔥 most of the stuff and keep just the generating of URLs.

simonrozsival:
I will close this PR. If some parts of the code seem useful in the future, we can cherry-pick it from here.

PR #1694, opened 2018-03-09
Fix weird ui behavior when current time entry changes twice fast

### LINKED ISSUE #1691: Conflict resolution problem
Steps to reproduce:
- start a TE in mobile app
- stop the TE in the web app
- start another TE in the web app
- force sync in the mobile app
- there will be a weird animation and the "play" button will be semitransparent

One user reports that after doing this whenever he then opens the app, it crashes.

simonrozsival:
Closes #1691

If you have an idea for a better solution, please share it here in the comments.

Squash whenever


amulware:
Squash away.

Issue #1271, opened 2018-01-11
Edit view: error message UI


pe0ny:
- [x] Add the error message at the top of the input fields.
- [x] Allow scrolling if the content doesn't fit on the screen anymore.
- [ ] User can't close or hide the error message.
- [ ] Error message will disappear when the error is fixed (eg user changed the project).

### Error message
![edit error](https://user-images.githubusercontent.com/5064445/34825406-0bb19768-f6db-11e7-94cb-3337cff82d9e.png)

Zeplin: https://zpl.io/be1WR3W

kornelijepetak:
A comment from @amulware 

> yeah, the last can’t be done. the error might not be shown if the entry was edited in any way (since the sync state will be different), and if it fails to sync again, the error will come back. if it does sync, the error wouldn’t be shown

@pe0ny This flow is a bit strange, requires a lot of fiddling with validation, and I am not sure if we can detect that all errors are fixed unless trying to resync the TE and backend validates it.

So, how do you want to proceed? Should we still do what the Toggl iOS does: mark as invalid, dismiss it by clicking the message and wait for another attempt?

amulware:
Note: Syncing with backend is currently the ONLY way to validate, and for the foreseeable future that'll be all we have. Workspace settings like required fields and locked time entries are important, but just not doable right now.

pe0ny:
@kornelijepetak I see. Obviously, we can remove the last point in the list then. And yes, I would go with the iOS way and let user dismiss it by tapping on the error message and waiting for another attempt.

PR #1695, opened 2018-03-09
Add error view to Edit TE screen

kornelijepetak:
Ref #1271 

Currently, the error message is shown upon error.

Having things run like described in the issue requires more fiddling with the view model so I will leave this for another PR for simplicity sake.

I am sorry about the `.axml`. Sometimes a few changes make the diff unrecognizable.


Toggl.Giskard/Resources/layout/EditLayout.axml
```diff
@@ -7,304 +7,358 @@
...
-        android:layout_width="0.5dp"
-        android:layout_height="72dp"
-        android:layout_below="@id/EditBillableRow"
-        android:layout_centerHorizontal="true" />
-    <RelativeLayout
-        android:id="@+id/EditTimeLeftPart"
-        local:MvxBind="Click SelectStartTimeCommand"
-        android:background="?attr/selectableItemBackground"
-        android:layout_weight="1"
+    <android.support.v4.widget.NestedScrollView
```
willsb: Why use a `NestedScrollView` instead of a plain old `ScrollView`?

kornelijepetak: Here I need an advice. The way I understood `NestedScrollViews`, it will enable me to scroll the container and the text inside a scrollable textedit independently. Haven't tried it, but would it work correctly even with the normal `ScrollView` if I had to scroll another control inside it?

PR #1700, opened 2018-03-12
Update reports calendar behavior

### LINKED ISSUE #1275: Fix how calendar appears in reports view
Temporarily we decided to open calendar at the top of the content. Let's fix it now and make it how it should be:
- [x] Calendar expands the header; Content reveals the calendar.
- [x] When user start scrolling down (swipes up), then content covers the calendar and in the end closes it.

simonrozsival:
Closes #1275 

Don't forget about a functional test when reviewing.

Squash whenever


Toggl.Daneel/ViewControllers/ReportsViewController.cs
```diff
@@ -56,24 +59,52 @@ public override void ViewDidLoad()
...
+
+                if (topConstant == 0) return;
+
+                // we need to adjust the offset of the scroll view so that it doesn't fold
+                // under the calendar while scrolling up
+                var adjustedOffset = new CGPoint(offset.X, ReportsTableView.ContentOffset.Y - offset.Y);
+                ReportsTableView.SetContentOffset(adjustedOffset, false);
+                View.LayoutIfNeeded();
+
+                if (topConstant == calendarHeight) HideCalendar();
```
kornelijepetak: I thought we are writing if-block in the same line as the condition only for returns with simple conditions.

Toggl.Daneel/ViewControllers/ReportsViewController.cs
```diff
@@ -56,24 +59,52 @@ public override void ViewDidLoad()
...
+        private void onReportsTableScrolled(object sender, CGPoint offset)
+        {
+            if (CalendarIsVisible)
+            {
+                var topConstant = (TopCalendarConstraint.Constant + offset.Y).Clamp(0, calendarHeight);
+                TopCalendarConstraint.Constant = topConstant;
+
+                if (topConstant == 0) return;
+
+                // we need to adjust the offset of the scroll view so that it doesn't fold
```
kornelijepetak: Somebody's maybe gonna be mad, but I like that you have included a comment here as it describes why this calculation was done.

Toggl.Daneel/ViewControllers/ReportsViewController.cs
simonrozsival: This seems like a simple condition to me - it is just one comparison and all the names are short. If you still want me to change the formatting here, please confirm it in another comment.

Toggl.Daneel/ViewControllers/ReportsViewController.cs
simonrozsival: I thought this correction of the offset isn't very obvious so there should be a comment. Thanks for the confirmation 👍

Toggl.Daneel/ViewControllers/ReportsViewController.cs
kornelijepetak: No need to change it. :shipit: I didn't click approve accidentally :)

It's just that I personally find this inconsistent with the usual procedural flow and how blocks are formatted and indented, and I just noted that I thought we are doing this only for returns.

Toggl.Daneel/ViewControllers/ReportsViewController.cs
kornelijepetak: In general I think our codebase would benefit much with adding a little bit of comments. Names of variables and functions often convey correctly `what` the code is doing, but almost never `why`. I'd like to see those more often. Easier reviewing, easier reading, easier changing (if ever needed).

Toggl.Daneel/ViewControllers/ReportsViewController.cs
amulware: It is inconsistent with the guidelines, which specify that we only use this syntax for early returns. /shrug

simonrozsival: I will change it then.

kornelijepetak: I think this single exception makes things worse. Just like the fact that in JavaScript you can omit the semicolon... except when you can't.

Personally, I even find this form better
```
if(isDisabled)
    return;
```
than the same line, because I somehow naturally expect that `what happens if condition satisfies` is **below** and **indented**. That's what I expect from all other blocks (namespaces, classes, functions, loops, etc.)

But this PR is not a guide in codestyling.

amulware: I agree with you @kornelijepetak. I think the main person to convince is @willsb. Or if you can get a majority on your side, you could just submit a PR and change it. :D

PR #1692, opened 2018-03-09
Fix Android build errors

willsb:
This fixes the daily builds.

What happened is that we are now using Proguard to prevent some of the google lib methods from being stripped in the Java layer. The heap increase in size is needed. 

UPDATE: The new Xamarin tooling also made it necessary to add a new `-nowarn` to our proguard file. I also bumped the Xamarin build tools.

[AdHoc build here](https://www.bitrise.io/build/baef29655cc624ed)

Squash whenever

PR #1702, opened 2018-03-12
Allow time entry deletion in the log

### LINKED ISSUE #1341: Swipe to delete/continue
## Swipe to continue
quickstart phase 4 1

Zeplin: https://zpl.io/2jlRLAr

## Swipe to delete
quickstart phase 4 1

Zeplin: https://zpl.io/agGBqlM

willsb:
Closes #1341 

~~If you wanna test locally, please merge #1692 in this branch if it hasn’t been merged into develop (in which case, just merge develop).~~ Merged!

I’d like to point out to @pe0ny that in the Inbox app the delete action is in the other side, so I particularly find it confusing to swipe left to delete. Maybe it’s just me, but I’m still voicing this concern, just in case.

Squash whenever.


kornelijepetak:
@pe0ny @willsb 
Interesting read: https://ux.stackexchange.com/questions/64659/swipe-left-to-delete-or-swipe-right-to-delete

In my opinion, swiping to left is kind of negative (going back, deleting, etc.). But I agree, this should be consistent with the rest of the platform. Not so much with specific apps, which may or may not conform to platform conventions. (There is probably lots of users that do not use Inbox app at all). Just sayin'

pe0ny:
In the Inbox app (for Android) there isn't actually no swipe to delete option. There's only Done which archives the email. I also think since we introduce these gestures in the onboarding it should be covered. We can change it when we start getting feedback from users.

Toggl.Giskard/Views/MainRecyclerViewTouchCallback.cs
```diff
@@ -1,38 +1,55 @@
...
-            : base(0, ItemTouchHelper.Left)
+            : base(0, ItemTouchHelper.Left | ItemTouchHelper.Right)
         {
             this.recyclerView = recyclerView;
 
-            var color = new Color(ContextCompat.GetColor(context, Resource.Color.playButtonGreen));
-            background = new ColorDrawable(color);
-            continueDrawable = ContextCompat.GetDrawable(context, Resource.Drawable.play_white);
-            continueDrawableMargin = (int)16.DpToPixels(context);
+            background = new[] { Resource.Color.playButtonRed, Resource.Color.playButtonGreen }
```
kornelijepetak: I'd say this variable should be in plural as it is a collection of `backgrounds`.

willsb:
I was measuring the wrong text🤦‍♂️

kornelijepetak:
@willsb I have merged `develop` into this branch, and since then, the tests fail. I have restarted bitrise 2 times after the initial build, both times failure. The tests don't seem related in any way to this PR. Would you check a look?

PR #1704, opened 2018-03-13
Update broken integration test

### LINKED ISSUE #1703: Update integration test
It looks like we forgot to update an integration test in #1657.

https://www.bitrise.io/build/791abf3580433551
```
Failed   Toggl.Ultrawave.Tests.Integration.UserApiTests+TheSignUpMethod.FailsWhenTheEmailIsAlreadyTaken
Error Message:
 Expected a <Toggl.Ultrawave.Exceptions.BadRequestException> to be thrown, but found a <Toggl.Ultrawave.Exceptions.EmailIsAlreadyUsedException>
```

amulware:
Closes #1703 

This was probably forgotten because there are two integration tests testing almost the same thing. Except one of them tries signing up again with the same password, while the other does not.

🦑


amulware:
Your question is incorrect. The real question is: Why wasn't there one?

PR #1706, opened 2018-03-13
Improve GroupCollectionViewSource

### LINKED ISSUE #1523: Various GroupedCollectionTableViewSource crashes

willsb:
Closes #1523 

This PR:

- Wraps the animations in Begin/EndUpdates blocks, which is recommended by Apple
- Handlles all cases of NotifyCollectionChangedAction and defaults to TableView.ReloadData()
- Removes the try/catch blocks from the animations, since they aren’t really catching anything (the throw happens asynchronously in the TableView, no in the Source)
- Puts a lock around the switch that tries to do the animations 

REPRO STEPS FOR TESTING:
- Open app
- Start new time entry
- Ensure it’s synced and running
- Put the app into the background
- Go to the web app, stop the running time entry
- Create a bunch of tes in the web app
- Start a new te in the web app and leave it running
- Wait for 5 minutes, the time it takes for the background service to kick in (alternatively, reduce the time it takes for it to kick in by changing the value in FoundationExtensions.cs, line 139)
- Open the app and it should not crash

Test with new users (welcome view), old users with no TEs (spider view) and also with users that have recent time entries.

:bug: :gun:


simonrozsival:
I am not able to reproduce the original bug using the steps in the simulator. I will try it also on my device.

simonrozsival:
I can't reproduce it even on a device. Is @arturs105  or @kornelijepetak  able to reproduce the bug on `1.4.1` (without this PR's fix)?

Toggl.Daneel/ViewSources/GroupedCollectionTableViewSource.cs
```diff
@@ -168,63 +155,101 @@ protected void OnChildCollectionChanged(object sender, ChildCollectionChangedEve
...
+                switch (args.Action)
+                {
+                    case NotifyCollectionChangedAction.Add:
+                        var indexToAdd = NSIndexSet.FromIndex(args.NewStartingIndex);
+                        OnSectionAdded(indexToAdd);
+                        break;
+
+                    case NotifyCollectionChangedAction.Remove:
+                        var indexToRemove = NSIndexSet.FromIndex(args.OldStartingIndex);
+                        OnSectionRemoved(indexToRemove);
```
amulware: Why is this (and the above) deserving of a method, and the others don't? I'd say let's be consistent here. And maybe consider using a shared helper method that accepts an `Action` or an `Action<TableView>` that does the begin/end-updates for us, instead of repeating it everywhere.

For example:
```
update(v => v.DeleteSections(indexToRemove, UITableViewRowAnimation.Automatic));
```
Or generic:
```
update(v => v.DeleteSections, indexToRemove, UITableViewRowAnimation.Automatic);
```
(ok, maybe I'm done here now xD)

Toggl.Daneel/ViewSources/GroupedCollectionTableViewSource.cs
```diff
@@ -168,63 +155,101 @@ protected void OnChildCollectionChanged(object sender, ChildCollectionChangedEve
...
-                    break;
-            }
+            TableView.BeginUpdates();
+            TableView.DeleteSections(indexToRemove, UITableViewRowAnimation.Automatic);
+            TableView.EndUpdates();
         }
 
         protected virtual void OnSectionAdded(NSIndexSet indexToAdd)
         {
+            TableView.BeginUpdates();
```
simonrozsival: This does not require a lock? Isn't it possible that this could crash as well?

Toggl.Daneel/ViewSources/GroupedCollectionTableViewSource.cs
```diff
@@ -168,63 +155,101 @@ protected void OnChildCollectionChanged(object sender, ChildCollectionChangedEve
...
-                case NotifyCollectionChangedAction.Add:
-                    OnSectionAdded(NSIndexSet.FromIndex(args.NewStartingIndex));
-                    break;
-
-                case NotifyCollectionChangedAction.Remove:
-                    var indexToRemove = NSIndexSet.FromIndex(args.OldStartingIndex);
-                    TableView.DeleteSections(indexToRemove, UITableViewRowAnimation.Automatic);
-                    break;
-            }
+            TableView.BeginUpdates();
```
simonrozsival: Lock?

Toggl.Daneel/ViewSources/GroupedCollectionTableViewSource.cs
willsb: Those methods deserve a method of its own because someone needed to override them somewhere down the line. I didn't create methods for the other actions because YAGNI.

I find your solution with an action quite cumbersome and not really needed tbh. The code is clear enough as is

Toggl.Daneel/ViewSources/GroupedCollectionTableViewSource.cs
willsb: This method is called inside a lock

Toggl.Daneel/ViewSources/GroupedCollectionTableViewSource.cs
willsb: OnSeconRemoved is called from within a lock

Toggl.Daneel/ViewSources/GroupedCollectionTableViewSource.cs
simonrozsival: But is it guaranteed to be called from the locked section? It is a protected method - the base class will never call it?

Toggl.Daneel/ViewSources/GroupedCollectionTableViewSource.cs
amulware: OnSectionRemoved isn't overriden anywhere though, and wasn't even used before these changes?

amulware: Regarding clarity, if others are ok with it, I can approve it, but to me, it feels repetitive and maybe inconsistent.

kornelijepetak: Otherwise, while being a bit cumbersome, I do vote for extracting it into a method for reuse/consistency/clarity.

Alternatively, and probably an overkill, I was considering something like this:

```csharp
using(TableView.WithSafeUpdates())
    TableView.MoveSection(args.OldStartingIndex, args.NewStartingIndex);
```
Requires additional class (that implements `IDisposable` and an extension method). Calls `Begin` in `WithSafeUpdates` extension method and `EndUpdates` in `Dispose`.

amulware: Note, the disposable pattern here is super useful, and can be implemented with a struct. The way `using` is compiled, I don't think it will box and allocate but just keep it on the stack. Super neat!

kornelijepetak: But @willsb hated it. So, it's for you to decide. 😄 Suggestions are mine, the choice in this case is not.

willsb: I'm kind of against all of that 😄 

We'll need more LoC for the using. If you want me to remove the OnRowDeleted method, I can do that. I can also investigate if the OnRowAdded method can be removed as well. But I'm against extracting this just to make it more .netty.

amulware: Please remove the method. Otherwise, I might disagree but I'm fine merging this.

Issue #1681, opened 2018-03-08
Fix foundation model builders for good


amulware:
We've had at least two bugs because of forgetting to copy fields in our model builders in 1.4 alone (and several before then).

Ref: #1678, #1679 

I want us to find a permanent solution to problems of that kind, ideally a compile-time one.

amulware:
Proposal: Remove the hand-written constructors that generate models from builders inside the builder code. Instead, auto-generate constructors that include all properties and call those from the builder. Then the builder has no choice but to pass all parameters. It's not foolproof though (parameters can be set to constants or other random properties of compatible type). I'd prefer something foolproof.

amulware:
Proposal: Write tests for the builders. Have a test pass in a fully populated model (how do we know it's fully populated though?) and see if the model build from the builder is equivalent.

amulware:
The proposal to use `With()` methods was accepted in #1711, which also showcases the syntax we will use in the future.

Note that in addition to what is seen there we will move model validation to a manually written method called from the generated constructor. Ideally, we could somehow ensure that validation is always complete, even when models are changed though. So consider generating a validation method call for each property, so that one has to add a validation method (even if just an empty one) for each of them.

Note that we may want to do this as part of #1891

amulware:
Yes, but down the line as part of the prime radiant refactor.

PR #1711, opened 2018-03-13
Replace builders by With() methods

amulware:
Ref #1681

Consider this a proposal. We were talking internally about using `With()` methods to mutate models. This is a first draft of how that could be done. Comments are very welcome.

Note that I edited the generated files directly. If we agree on how we'd like to do things, I would, of course, change the generators and use the same kind of logic everywhere. I would also generate the `With()` extension methods because that would be trivial to do.


amulware:
Note that we could also remove the DTOs easily, if we exposed the `With()` extension method to the UI, and changed calls like:

```
var preferencesDto = new EditPreferencesDTO { DurationFormat = newDurationFormat };
var newPreferences = await updatePreferences(preferencesDto);
```

into:

```
var newPreferences = await updatePreferences(p => p.With(durationFormat: DurationFormat));
```

But I'm not sure if we feel comfortable with that. It gives a lot of power to the UI.

Toggl.Foundation/DataSources/PreferencesDataSource.cs
```diff
@@ -95,10 +95,12 @@ private void processConflictResultionResult(IEnumerable<IConflictResolutionResul
...
-                .FromExisting(existing)
-                .SetFrom(dto)
-                .SetSyncStatus(SyncStatus.SyncNeeded)
-                .Build();
+            => existing.With(
+                dateFormat: dto.DateFormat,
+                durationFormat: dto.DurationFormat,
+                timeOfDayFormat: dto.TimeOfDayFormat,
+                collapseTimeEntries: dto.CollapseTimeEntries,
+                syncStatus: SyncStatus.SyncNeeded
```
amulware: The only thing I don't like is that it's easy to forget to specify the sync status here. However, that was the same before as well I suppose.

simonrozsival:
This is really nice 👍 The only thing I don't like is the way you place the closing bracket ⚔️

Toggl.Foundation/Models/Preferences.cs
```diff
@@ -1,128 +0,0 @@
...
-                return this;
-            }
-
-            public Builder SetSyncStatus(SyncStatus syncStatus)
-            {
-                SyncStatus = syncStatus;
-                return this;
-            }
-
-            private void ensureValidity()
```
amulware: Oh yes, one more important thing: We lose validation with this!

Should we try and stick it back in somewhere? Or rely on the With() method to never have bugs?  (validation has caught bugs for us in the past, by causing crashes instead of bad data that would be much harder to debug)

willsb: Not sure how I feel about making those fields mandatory. Might turn the whole thing into an unusable mess...

amulware: Making them mandatory where? Anywhere it's using `New<>`? I definitely wouldn't do that. The main danger is that someone might pass default/invalid/null/0 values for `New<>` ending up with an invalid object (like a TE with user-id 0, which happened recently and made the app crash which is better than causing bad data imo).

simonrozsival: I don't think we need validation anymore - the main purpose of the validation we had was to check if we filled all the necessary fields. We now either have to use an existing entity and change some of the fields or use the constructor which has no optional fields. Isn't that enough to check that the new entity has all the values set explicitly?

If we want some validation, then maybe it should be in the constructor..

amulware: Well, the user-id==0 was caused by a field not set, and that can't happen anymore with this. But it we somehow did get a user-id==0, TEs would still end up invalid, and I'm not sure what to do about that then. :/

I'm undecided. But yes, we could call a Validate() method in the constructor that we then implement in a hand-written partial class.

Other's thoughts?

Toggl.Foundation/DataSources/PreferencesDataSource.cs
kornelijepetak: If `syncStatus` is always needed, why not make it non-optional?

amulware: Because you might not want to change it? Maybe that's unlikely, but if it always changes to `SyncNeeded` we could hardcode that in the With() method, but that seems strange to me. Then the `At` date comparison should also go in there, and where does it end? Thoughts?

kornelijepetak: But then I don't think you can really enforce this by any language construct. I believe we just need to be mindful and not forget important things. Also, tests should help too 😄

amulware: I agree. Unless we encode business logic in the models themselves, this will be tricky. And that would be tricky too, so better to just move to well-tested, and carefully written and reviewed, interactors.

amulware:
@willsb Still waiting for an answer from you.

In general, if there are no more comments on this soon, I will proceed at my own discretion.

willsb:
Why can't you run the very same validation function at the end of the `With` method?

amulware:
@willsb because With is generated and an extension method. The solution would be having Validate in a handwritten partial class, and be public. Is that what you would propose? Or is there a different solution?

Honestly, I rather have it in the constructor, so that ANY model created has to be valid by necessity, even if we call the constructor ourselves.

amulware:
Also, I suppose the extension method argument is silly since I could stick the With methods into a partial class as well.

amulware:
Yap. You're correct. I forgot we're calling it on the interface, not on the model.

amulware:
I will move validation into the constructor then. I think with that, all open questions are solved. I'll see about getting around to completing this asap. :)

amulware:
We'll go ahead as discussed above, but I'm closing this since it may be out of date. Anything else will happen in #1681 and future PRs.

PR #1712, opened 2018-03-13
Add app shortcut handling

### LINKED ISSUE #1529: Finish 3d touch shortcuts

arturs105:
Closes #1529


Toggl.Foundation.MvvmCross/ViewModels/MainViewModel.cs
```diff
@@ -114,6 +123,22 @@ public bool ShouldShowWelcomeBack
...
 
+        public void Init(string id)
+        {
+            if (string.IsNullOrEmpty(id)) return;
+
+            long parsedId = long.Parse(id);
+
+            var subscription = dataSource
+                .TimeEntries
+                .GetById(parsedId)
```
simonrozsival: What happens if a TE with the ID does not exist? Does the app crash? Is it even possible that there would be a shortcut with an ID of a TE which does not exist?

Toggl.Foundation/Shortcuts/BaseApplicationShortcutCreator.cs
```diff
@@ -24,59 +24,50 @@ public abstract class BaseApplicationShortcutCreator : IApplicationShortcutCreat
...
-                },
-                () => timeEntrySuggestionsSubscription.Dispose());
-        }
+                });
+
+        private IObservable<ApplicationShortcut> createMostRecentEntryShortcut()
+            => dataSource
+                .TimeEntries
+                .GetAll()
+                .Select(timeEntries => timeEntries.MaxBy(timeEntry => timeEntry.Start))
```
simonrozsival: This throws if the user doesn't have any TEs - which happens e.g. right after sign up - will the app crash in that case?

Toggl.Foundation/Sync/SyncManager.cs
```diff
@@ -91,6 +96,7 @@ private void syncOperationCompleted(SyncResult result)
                     if (IsRunningSync == false)
                     {
                         progress.OnNext(SyncProgress.Synced);
+                        shortcutCreator.OnSynced();
```
simonrozsival: I don't like the fact that `SyncManager` depends on shortcut creator - wouldn't it be better to just subscribe to the `SyncManager.ProgressObservable.Where(progress => progress == SyncProgress.Sync)` in the data source or somewhere else?

Toggl.Foundation.MvvmCross/ViewModels/ReportsViewModel.cs
```diff
@@ -7,14 +7,17 @@
...
 using Toggl.Foundation.DataSources;
 using Toggl.Foundation.MvvmCross.Parameters;
+using Toggl.Foundation.MvvmCross.ViewModels;
 using Toggl.Foundation.MvvmCross.ViewModels.Calendar;
 using Toggl.Foundation.MvvmCross.ViewModels.Hints;
 using Toggl.Foundation.Reports;
 using Toggl.Multivac;
 using Toggl.PrimeRadiant.Models;
 
+[assembly: MvxNavigation(typeof(ReportsViewModel), ApplicationUrls.Reports)]
```
simonrozsival: `ReportsViewModel` needs workspace ID as a param for the `Prepare` method to set up all the subscriptions. As this method is not called, the reports page is broken when opened using the shortcut.

I suppose this should take a regexp with `workspace_id=\d+` and the class should have an explicit `Init` method.

Toggl.Foundation/Shortcuts/BaseApplicationShortcutCreator.cs
simonrozsival: It seems the app does not crash, but shouldn't it be handled in some better/more explicit way?

Toggl.Foundation.MvvmCross/ViewModels/MainViewModel.cs
arturs105: The shortcut is updated after TE creation and sync, so if the user deletes the most recent time entry and doesn't sync, then the shortcut becomes invalid.

However, since we decided not to show user data in the shortcuts, we can just get the latest time entry after the shortcut has been pressed.

Toggl.Foundation.MvvmCross/ViewModels/MainViewModel.cs
```diff
@@ -136,7 +151,7 @@ public bool ShouldShowWelcomeBack
             var currentlyRunningTimeEntryDisposable = dataSource
                 .TimeEntries
                 .CurrentlyRunningTimeEntry
-                .Throttle(currentTimeEntryDueTime, scheduler) // avoid overwhelming the UI with frequent updates
```
simonrozsival: Why did you remove this comment? It was there on purpose.

Toggl.Foundation.MvvmCross/ViewModels/MainViewModel.cs
```diff
@@ -159,6 +174,29 @@ public bool ShouldShowWelcomeBack
             disposeBag.Add(syncManagerDisposable);
             disposeBag.Add(isEmptyChangedDisposable);
             disposeBag.Add(currentlyRunningTimeEntryDisposable);
+
```
simonrozsival: Extra \n

Toggl.Foundation.MvvmCross/ViewModels/MainViewModel.cs
```diff
@@ -159,6 +174,29 @@ public bool ShouldShowWelcomeBack
...
+                    await continueCurrentEntry();
+                    break;
+
+                case ApplicationUrls.Main.Action.Stop:
+                    await stopTimeEntry();
+                    break;
+            }
+        }
+
+        private async Task continueCurrentEntry()
```
simonrozsival: This does not necessarily continue the "current" time entry, but the "last TE" - maybe the name of the method should be `continueMostRecentTimeEntry()` so it is less confusing/it describes the action better.

Toggl.Ultrawave.Tests.Integration/UserApiTests.cs
```diff
@@ -226,7 +226,7 @@ public void FailsWhenThePasswordIsTooShort(string empty)
                     .SignUp(email, "thePasswordIsNotImportant".ToPassword())
                     .Wait();
 
-                secondSigningUp.ShouldThrow<EmailIsAlreadyUsedException>();
```
simonrozsival: This exception type shouldn't be changed.

Toggl.Ultrawave.Tests.Integration/UserApiTests.cs
```diff
@@ -238,7 +238,7 @@ public void FailsWhenThePasswordIsTooShort(string empty)
 
                 Action secondSigningUp = () => unauthenticatedTogglApi.User.SignUp(email, password).Wait();
 
-                secondSigningUp.ShouldThrow<EmailIsAlreadyUsedException>();
```
simonrozsival: This exception type shouldn't be changed.

Toggl.Ultrawave.Tests.Integration/UserApiTests.cs
```diff
@@ -329,4 +329,4 @@ public IEnumerator<object[]> GetEnumerator()
             IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();
         }
     }
-}
```
simonrozsival: Return the \n back please.

Toggl.Foundation.MvvmCross/ViewModels/MainViewModel.cs
arturs105: Oh, yeah. This was due to merging issues. I re-added the line, but i forgot to bring back the comment. Will do now.

Toggl.Foundation.Tests/MvvmCross/ViewModels/MainViewModelTests.cs
```diff
@@ -264,8 +270,13 @@ public TheStopTimeEntryCommand()
                 Scheduler.AdvanceBy(TimeSpan.FromMilliseconds(50).Ticks);
             }
 
+            protected override void AdditionalSetup()
```
simonrozsival: 🔥 ?

Toggl.Foundation/Shortcuts/BaseApplicationShortcutCreator.cs
```diff
@@ -24,61 +22,51 @@ public abstract class BaseApplicationShortcutCreator : IApplicationShortcutCreat
...
+        private readonly ApplicationShortcut stopTimeEntryShortcut = new ApplicationShortcut(
+            ApplicationUrls.Main.StopTimeEntry,
+            "Stop",
+            ShortcutType.StopTimeEntry
+        );
 
-        public void OnLogin()
+        public void OnLogin(ITogglDataSource dataSource)
         {
+            if (dataSource == null) return;
```
simonrozsival: Is it even possible that you would get a `null` here?

arturs105: Yeah. This method get called from `LoginManager.GetDataSourceIfLoggedIn()` and it is null, when the user isn't logged in.

The app did actually crash whenever it was opened without a user signed in. This null check fixes that crash.

simonrozsival: OK

Toggl.Foundation.MvvmCross/ViewModels/MainViewModel.cs
```diff
@@ -159,6 +174,28 @@ public bool ShouldShowWelcomeBack
...
+
+                case ApplicationUrls.Main.Action.Stop:
+                    await stopTimeEntry();
+                    break;
+            }
+        }
+
+        private async Task continueMostRecentEntry()
+        {
+            var interactor = await dataSource
```
willsb: This whole query should be an interactor of its own.

Toggl.Foundation.MvvmCross/ViewModels/ReportsViewModel.cs
```diff
@@ -29,7 +33,7 @@ public sealed class ReportsViewModel : MvxViewModel<long>
 
         private DateTimeOffset startDate;
         private DateTimeOffset endDate;
-        private long workspaceId;
+        private long? workspaceId;
```
willsb: I vote against making it nullable and check for 0 instead. 0 is never valid

Toggl.Foundation.Tests/Toggl.Foundation.Tests.csproj
```diff
@@ -29,4 +29,9 @@
     <PackageReference Include="xunit.runner.visualstudio" Version="2.3.1" />
     <DotNetCliToolReference Include="dotnet-xunit" Version="2.3.1" />
   </ItemGroup>
+  <ItemGroup>
```
willsb: Please 🔥

Toggl.Foundation/Models/TimeEntryPrototype.cs
```diff
@@ -0,0 +1,42 @@
+﻿using System;
+using System.Linq;
+using Toggl.Multivac.Models;
+
+namespace Toggl.Foundation.Models
+{
+    public sealed class TimeEntryPrototype : ITimeEntryPrototype
```
willsb: Please make this internal once that whole query is changed into an interactor

amulware:
Yes, that would make sense. I'm not a big fan of the stop shortcut (right now) though, since it also relies on data we don't have (we don't know if a TE is actually running or not). What was @pe0ny latest call on this?

pe0ny:
@amulware Afaik, we decided to remove it because we don't know if the TE is actually running or not.

PR #1713, opened 2018-03-14
Fix main view for iOS 10

arturs105:
**Previously:** 
img_1166

**Currently:** 
screen shot 2018-03-14 at 10 41 53

❗️  Please do a functional test.

Issue #1680, opened 2018-03-08
Main view navigation bar dissappears


amulware:
Well. It sort of disappears. Here are some screenshots:

![image uploaded from ios](https://user-images.githubusercontent.com/3987892/37141336-94e988a4-22ac-11e8-9e6c-c9d2e370e862.png)
![image uploaded from ios 1](https://user-images.githubusercontent.com/3987892/37141334-94ab4918-22ac-11e8-882a-b10adbaa9882.png)
![image uploaded from ios 2](https://user-images.githubusercontent.com/3987892/37141335-94c9dbc6-22ac-11e8-811f-4d6e59606eb6.png)

It might have to do with going to the settings during sync or something, but I'm not sure how to reproduce it right now.

simonrozsival:
We still don’t know how to reproduce this issue?

amulware:
Nope, I can't even reproduce it randomly. I'll remove it from the milestone. Maybe we'll figure it out in the future. Otherwise, if we haven't by end of April, anyone feel free to close this.

PR #1714, opened 2018-03-14
Fix main view for iOS 10

arturs105:
**Previously:** 
img_1166

**Currently:** 
screen shot 2018-03-14 at 10 41 53

❗️  Please do a functional test.

Squash whenever


amulware:
Follow up on #1713: #1680 was not caused by this issue. It occurred on 11.2

PR #1718, opened 2018-03-14
Fix settings crash

### LINKED ISSUE #1688: Fix settings scrollview layout
There is definitely a glitch on iPhoneX where the scrollview does not reach all the way to the bottom of the screen (you have to drag down the scrollview to see this):

screen shot 2018-03-08 at 15 50 05

We should also look into how it looks on iPhone SE, where if scrolled all the way down, the signout button touches the bottom of the screen, while on larger phones it does not. Ping @pe0ny for this one.

### LINKED ISSUE #1715: App crashes, when opening settings
This happens on iOS 10.x (works fine on iOS 11), because of `TopConstraint.AdaptForIos10(navbar)`, but `TopConstraint` is null.

arturs105:
Closes #1715 
Ref #1688 

The problem was an outlet that was null. The settings view had some weird constraints as well, so i reworked it and organised it better.

Squash whenever


simonrozsival:
I don't think we should deal with #1688 in this PR and push that into the hotfix. Also I don't think that it actually fixes it:
>Padding between the Sign out button and the bottom of the screen should always be 24pt like shown in here: https://zpl.io/29EBEz2

On SE there is a way too large gap between the buttons and on Plus and X it is far from the bottom edge.

arturs105:
Although i agree that we shouldn't deal with #1688 for the hotfix (more on that in the next paragraph), fixing the crash required some layout modification, since, after fixing the crash, the whole scroll view had too big initial offset. I couldn't easily solve that, so i just restructured the scroll view better.

Actually yes, this PR doesn't close the issue, since the sign out button isn't positioned correctly on big screens (though i think it looks correct on SE).

Issue #1519, opened 2018-02-14
[onboarding] phase 1: TE create view


fonorobert:
Phase 1 of onboarding flow.
Phase 2: #1520 
Phase 3: #1521 
Phase 4: #1522 

Add:

- [x] tooltip for start button in log view
- [x] modified TE create view
- [x] modified checkmark button in TE create view
- [x] tooltip for stop button in log view

**Only show this phase to users that are new users and signed up in the app!**

# Steps & designs
## 1. Tooltip in log view pointing user to start button
https://app.zeplin.io/project/5728ce853dd1b182475f25e9/screen/5a8301c783a1a15b4f6e5f6f
![screenshot from 2018-02-14 13-41-02](https://user-images.githubusercontent.com/6963710/36208034-0412cd60-1198-11e8-8593-74cf3587321d.png)

**Show:**
- when user first signs up and sees log view

**Hide:**
- If user has tapped the tooltip to dismiss it
- If user tapped the start button.

Copy: “Tap to start timer”

## 2. New TE view, once user taps “start” for the first time
https://app.zeplin.io/project/5728ce853dd1b182475f25e9/screen/5a84129c0bcd0643489d675a

![screenshot from 2018-02-14 13-42-48](https://user-images.githubusercontent.com/6963710/36208069-1d995b5a-1198-11e8-8309-894829a430db.png)

Secondary input fields for project and tag adding are removed.
Checkmark button in bottom right is greyed out.

**Show:**
- first time the user starts a TE in the app

## 3. Checkmark button highlighted
https://app.zeplin.io/project/5728ce853dd1b182475f25e9/screen/5a84129b74b47303abb16429
![screenshot from 2018-02-14 13-46-28](https://user-images.githubusercontent.com/6963710/36208082-29e8bc66-1198-11e8-9c0c-abe237ef3bcb.png)

Checkmark is white, green solid circle around it.

**Show:**
- After step 2, when user starts typing in description field

## 4. Second tooltip in log view pointing to stop button
https://app.zeplin.io/project/5728ce853dd1b182475f25e9/screen/5a8301c63467b6111e837b8d
![screenshot from 2018-02-14 13-48-00](https://user-images.githubusercontent.com/6963710/36208091-339cd27e-1198-11e8-907a-55bf3085f8bb.png)

**Show:**
- After step 3
- first time user has a running TE in the app

**Hide:**
- If user has tapped the tooltip to dismiss it
- If user tapped the stop button.

[...]

PR #1720, opened 2018-03-14
Onboarding services

simonrozsival:
Ref #1519 

This PR adds platform specific services which will eventually contain individual onboarding "steps" which will be used by the UI to determine if some tooltip should be visible or not + dismissing it.

A lot of the code is just related to registering the services - it might be easier to review by individual commits.

Squash whenever


simonrozsival:
The logic needs to be in the ViewController - the `IOnboardingService` will be injected into the VM and then the VC has to have correct bindings.

Toggl.Daneel/Services/DaneelOnboardingService.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System;
+using Toggl.Foundation.Services;
+using Toggl.Multivac;
+using Toggl.PrimeRadiant.Settings;
+
+namespace Toggl.Daneel.Services
+{
+    public sealed class DaneelOnboardingService : IOnboardingService
```
arturs105: I like this naming scheme. I think we should do the same for other services as well (`DialogService`, `MailService`, `GoogleService` etc.). That way we wouldn't have 2 classes of the same name and fuzzy search would work better. What do others think?

Toggl.PrimeRadiant/Onboarding/DismissableOnboardingStep.cs
```diff
@@ -0,0 +1,34 @@
+﻿using System;
+using Toggl.Multivac;
+using Toggl.PrimeRadiant.Settings;
+
+namespace Toggl.PrimeRadiant.Onboarding
+{
+    public sealed class DismissableOnboardingStep : IDismissableOnboardingStep
+    {
+        private readonly IOnboardingStep onboardingStep;
```
arturs105: I don't think this makes much sense, because this class is an `IOnboardingStep` itself, so there's no need to hold a reference on some other `IOnboardingStep`, because all occurences of `onboardingStep.SomeMethod()` could be replaced by `this.SomeMethod()`.

Toggl.PrimeRadiant/Onboarding/IOnboardingStep.cs
```diff
@@ -0,0 +1,8 @@
+﻿using System;
+namespace Toggl.PrimeRadiant.Onboarding
+{
+    public interface IOnboardingStep
+    {
+        bool ShouldBeVisible();
```
arturs105: Why not property?

Toggl.PrimeRadiant/Settings/IOnboardingStorage.cs
```diff
@@ -11,5 +11,8 @@ public interface IOnboardingStorage
         bool IsNewUser();
         string GetLastOpened();
         bool CompletedOnboarding();
+
+        bool WasDismissed(string key);
+        void Dismiss(string key);
```
arturs105: I think naming can be improved here. Either `DismissOnboardingStep(string key)` or `Dismiss(IOnboardingStep step)` (or something even better)

Toggl.PrimeRadiant/Onboarding/DismissableOnboardingStep.cs
simonrozsival: The idea is that the actual onboarding states don't have to know anything about them being dismissable. They should just check if the conditions for them to be displayed are met. This single class can then wrap any onboarding step and some string key and implement the dismissing logic - I would prefer using composition over inheritance here but I could give it another thought if you really think it would be beneficial in this case.

arturs105: Using composition makes sense i guess, but currently it seems like we're mixing composition with inheritance. I suggestion changing this:
```
interface IDismissableOnboardingStep : IOnboardingStep
{
}
```
to:
```
interface IDismissableOnboardingStep
{
    IOnboardingStep Step { get; }
}
```
Could that work?

Toggl.Daneel/Services/DaneelOnboardingService.cs
arturs105: Since in slack discussions we decided not to prepend project name to all services, this shouldn't have it either.

PR #1722, opened 2018-03-15
Add TransparentColor value converter

kornelijepetak:
Well, the title says it all.

This is needed to specify transparency in project selection tokens in `Giskard`.

~~However, since it's not dependent on anything related to `Giskard`, I have put it into `foundation`.~~

Squash whenever.


Toggl.Foundation.Tests/MvvmCross/Converters/TransparentColorValueConverterTests.cs
```diff
@@ -0,0 +1,89 @@
...
+            [InlineData("#FF004412")]
+            [InlineData("FF004412")]
+            [InlineData("FF00441")]
+            public void ThrowsForTooLongValues(string value)
+            {
+                var result = convert(value);
+
+                result.Should().Be(MvxBindingConstant.UnsetValue);
+            }
+
```
kornelijepetak: This line intentionally left blank 😛 Will fix.

Toggl.Foundation.MvvmCross/Converters/TransparentColorValueConverter.cs
```diff
@@ -0,0 +1,34 @@
...
+    public sealed class TransparentColorValueConverter : MvxValueConverter<string, string>
+    {
+        private const long defaultOpacity = 255;
+
+        protected override string Convert(string value, Type targetType, object parameter, CultureInfo culture)
+        {
+            if (string.IsNullOrEmpty(value))
+                throw new ArgumentException("Value must be specified.", nameof(value));
+
+            value = value.Trim().Trim('#');
```
amulware: Changing argument variables is not nice.

Toggl.Foundation.MvvmCross/Converters/TransparentColorValueConverter.cs
```diff
@@ -0,0 +1,34 @@
...
+            var isHexString = int.TryParse(value, NumberStyles.HexNumber, CultureInfo.InvariantCulture, out var _);
+
+            if (!isHexString || value.Length != 6)
+                throw new ArgumentException("The value must be a 6 digit hex color string!", nameof(value));
+
+            var opacity = parameter is long opacityParameter
+                ? (long)Math.Round(2.55 * opacityParameter.Clamp(0, 100))
+                : defaultOpacity;
+
+            var result = $"#{opacity:X2}{value.Trim('#')}".ToUpper();
```
amulware: `value` was already trimmed of # above.

Toggl.Foundation.MvvmCross/Converters/TransparentColorValueConverter.cs
```diff
@@ -0,0 +1,34 @@
...
+        {
+            if (string.IsNullOrEmpty(value))
+                throw new ArgumentException("Value must be specified.", nameof(value));
+
+            value = value.Trim().Trim('#');
+
+            var isHexString = int.TryParse(value, NumberStyles.HexNumber, CultureInfo.InvariantCulture, out var _);
+
+            if (!isHexString || value.Length != 6)
+                throw new ArgumentException("The value must be a 6 digit hex color string!", nameof(value));
```
amulware: Why bother doing the parsing and throwing here if this is never used? Wouldn't a simple length check be enough?

Toggl.Foundation.Tests/MvvmCross/Converters/TransparentColorValueConverterTests.cs
```diff
@@ -0,0 +1,89 @@
...
+                convertedValues.Should().Be(converted);
+            }
+
+            [Theory]
+            [InlineData("FADE12", 0, "#00FADE12")]
+            [InlineData("#ABCDEF", 20, "#33ABCDEF")]
+            [InlineData("ABCDEF", 40, "#66ABCDEF")]
+            [InlineData("#ABCDEF", 60, "#99ABCDEF")]
+            [InlineData("136789", 80, "#CC136789")]
+            [InlineData("#CAB755", 100, "#FFCAB755")]
```
amulware: Property test? Or at least also test with a value that doesn't end up with double letters for the alpha part.

Toggl.Foundation.MvvmCross/Converters/TransparentColorValueConverter.cs
kornelijepetak: If I get `#AMULWARE` as input, I want to throw, because I want:
* 6 digit string
* valid hex color

kornelijepetak: Checking only length will allow things like `#XYZTUV` which is not a valid `hex color`

amulware: But is this the correct place to check that? Shouldn't this have been checked much earlier (or later when this colour is consumed)? It just feels strange to me to do relatively complex string parsing just to satisfy a condition that doesn't actually affect the code run here.

kornelijepetak: I need to check whether I have a valid hex color and I have to throw in that case. But you are right that 6-digit check can be done earlier to prevent this parsing if length is not valid. Would that work for you?

amulware: If you are still going with this manual approach... I don't think that it matters much, no. I'm bothered by the parsing in the first place. But it's up to you guys.

kornelijepetak:
I have chosen to make `DrawableColor` work with `Color` too. Makes more sense.

PR #1729, opened 2018-03-15
Set properties from fixed entity only when it is different

### LINKED ISSUE #1103: Continuing TEs does not copy tags

### LINKED ISSUE #1725: Continuing a TE doesn't copy tags when another TE is running at that moment
How to reproduce:

1. make sure all TEs are stopped
2. continue a TE A which has at least one tag
3. check the newly created TE B - it has the same tags as A
4. continue the TE A again (while B is still running)
5. B will be stopped, a new TE C will be running
6. check the TE C - it doesn't have any tags, expected behavior would be to have the same tags as A

simonrozsival:
Closes #1725, ref #1103

in `SetPropertiesFrom(entity)` there is this code:
```
...
RealmTags.Clear();
if (entity.TagIds != null)
{
    var allRealmTags = entity.TagIds.Select(id => realm.All<RealmTag>().Single(x => x.Id == id || x.OriginalId == id));
    foreach (var oneOfRealmTags in allRealmTags)
        RealmTags.Add(oneOfRealmTags);
}
...
```
The problem is that it tries to set properties from itself and `RealmTags.Clear();` destroys the information which should be later copied

Squash whenever


simonrozsival:
The other variant is `Object.ReferenceEquals(entity, fixedEntity) == false`. I think that using `Equals` is better in this case - for some entities it might be useful to override it, but normally it behaves like `ReferenceEquals`.

PR #1732, opened 2018-03-15
Fix setting up initial state of preferences

### LINKED ISSUE #1721: Problem with date and time formats
*I'm currently using TF build 1.4 (28) which should be the same as the one that is currently up in the App Store.*

When edit running or existing TE, then date/time formats are messed up. Check the screenshot below:

image uploaded from ios

Also, when I open settings to check out what happens there, then date format is "missing" (nothing isn't selected) and 24-hour format is turned off, though I haven't turned it off. When I turn 24-format back on, then all is good again. Though, the issue appears again when i restart the app.

image uploaded from ios 1

That's the best way how can I describe the issue. Ping me for more questions.

simonrozsival:
Closes #1721

There was a problem that the preferences weren't loaded because there was no subscription to the observable on line 32. Also just to be sure the observable sequence is initialized with the default preferences.

Squash whenever


Toggl.Foundation/DataSources/PreferencesDataSource.cs
```diff
@@ -26,10 +28,10 @@ public PreferencesDataSource(ISingleObjectStorage<IDatabasePreferences> database
 
             this.databaseStorage = databaseStorage;
 
-            currentPreferencesSubject = new ReplaySubject<IDatabasePreferences>(1);
+            currentPreferencesSubject = new BehaviorSubject<IDatabasePreferences>(Preferences.DefaultPreferences);
 
             Current = currentPreferencesSubject.AsObservable();
-            Get().Do(currentPreferencesSubject.OnNext);
+            initializationDisposable = Get().Do(currentPreferencesSubject.OnNext).Subscribe();
```
amulware: Why not just use `onNext` when subscribing?

PR #1734, opened 2018-03-15
Make Interactor Aware tests use mocks

willsb:
As discussed on slack, this is needed for #1712 

I removed tests that no longer made sense and transformed the tests that did make sense. 

More than one pair of eyeballs reviewing this is mandatory.


Toggl.Foundation.Tests/Interactors/CreateTimeEntryInteractorTests.cs
```diff
@@ -0,0 +1,300 @@
...
+            }
+
+            protected abstract IObservable<IDatabaseTimeEntry> CallInteractor(ITimeEntryPrototype prototype);
+
+            [Fact, LogIfTooSlow]
+            public async Task CreatesANewTimeEntryInTheDatabase()
+            {
+                await CallInteractor(CreatePrototype(ValidTime, ValidDescription, true, ProjectId));
+
+                await DataSource.TimeEntries.ReceivedWithAnyArgs().Create(null);
```
arturs105: Why not `Arg.Any<IDatabaseTimeEntry>()` instead of null?

willsb: Because of `ReceivedWithAnyArgs`

PR #1735, opened 2018-03-15
Refactor the reset method

willsb:
This PR:

- Improves the method name
- Only calls clear if the database is indeed empty
- Uses the [ReplaceWith](https://github.com/MvvmCross/MvvmCross/blob/f80160ac1588c90ce2cd45d9487d744712c62e2a/MvvmCross/ViewModels/MvxObservableCollection.cs#L105-L119) method, which does event suppression and only sends a single reset signal (which won’t be animated on iOS, and that’s fine).

Squash ASAP


Toggl.Foundation.MvvmCross/ViewModels/TimeEntriesLogViewModel.cs
```diff
@@ -1,4 +1,6 @@
 ﻿using System;
+using System.Collections.Generic;
+using System.Diagnostics.Contracts;
```
simonrozsival: 🔥 ?

Toggl.Foundation.MvvmCross/ViewModels/TimeEntriesLogViewModel.cs
```diff
@@ -108,30 +108,23 @@ public sealed class TimeEntriesLogViewModel : MvxViewModel
             disposeBag.Add(preferencesDisposable);
         }
 
-        private async Task reset()
+        private async Task fetchGroupedTimeEntries()
```
amulware: Please use a different name. Toggl has a grouped time entry feature, and that's bound to be confusing. I suggest `fetchSectionedTimeEntries` or `fetchTimeEntriesGroupedByDay`

amulware:
Please explain what bug this fixes. :)

Toggl.Foundation.MvvmCross/ViewModels/TimeEntriesLogViewModel.cs
willsb: 👍

willsb:
Me and @simonrozsival have found two points that might be causing #1523. This PR fixes one of them (the reset method) and he'll submit another PR fixing the other point (the empty time entry cells).

amulware:
This seems to now add them all at once, instead of one by one, and it only does the clear if it needs to, sure. In fact, the clear will never happen now, will it? The datasource should NEVER return null there, that would be very bad practice. Always return an empty collection.

But yeah sure, I don't see how this is fixing anything, but it seems like a fine change.

PR #1738, opened 2018-03-16
Fix datepicker time format

### LINKED ISSUE #1737: Datepicker not respecting 12/24h setting
The one in the wheel view. It should be following user settings, not system settings.

simonrozsival:
Closes #1737 

There is no property for specifying 12/24 hour format - the locale needs to be changed - that is why the British format is used for the 24-hour format and the US locale for 12-hour format.

Don't forget about a functional test when reviewing.

Squash whenever


amulware:
Please merge this into develop, it's really not that important.

simonrozsival:
OK

PR #1739, opened 2018-03-16
Fix 1password crash

### LINKED ISSUE #1736: Tapping 1Password icon on login causes app to crash
As reported by [this social post](https://app.agorapulse.com/inbox/twitter/51704?view=inbox&filter=any&item=974502055959285760) (https://twitter.com/peeja/status/974502055959285760) and reproduced by yours truly.

1. Open Toggl app (If logged into Toggl app, log out through settings screen)
2. Progress through welcome screen
3. Tap on Log in button
4. Tap on 1Password button beside the Email textbox 
5. App crashes to home screen

Confirmed on iOS 11.2.6 (15D100) for the following devices:
* iPhone 8
* iPhone 8 Plus

simonrozsival:
Closes #1736 

For some reason creating the `NSDictionary` for the Firebase analytics threw an exception - I just wonder why it doesn't crash for other tracked events, like sign up or every page change?

Also I thought we already ditched Firebase - is there a reason we still log Firebase? Is it ok to get rid of it?

I don't use 1pwd so I can't verify that the whole process now works - somebody with 1pwd must verify this.

Squash whenever


amulware:
This removes additional data from all events tracked on firebase though. While I have no idea why it would crash in that on instant (maybe the superdevs have some idea), I don't think we can just remove this alltogether.

amulware:
To answer your question: We did not ditch Firebase at all. It's still very much open to what we are gonna use in the long term.

PR #1742, opened 2018-03-16
Simplify time entries view source

simonrozsival:
Ref #1523 

This PR simplifies the time entries log view source and removes the empty state from it and instead moves it into separate components which are used by the main view directly.

There might be a problem with the consistency of the rows in the table source where for an empty state it contains 3 rows and after 1 row is added, it now has 1 row - Apple might not like this (3 + 1 = 1) and crash our app.

Squash whenever


Toggl.Daneel/ViewControllers/MainViewController.cs
```diff
@@ -317,6 +320,20 @@ private void prepareSpiderViews()
             spiderBroView.CenterXAnchor.ConstraintEqualTo(spiderContainerView.CenterXAnchor).Active = true;
         }
 
+        private void prepareEmptyStateView()
+        {
+            emptyStateView.BackgroundColor = UIColor.Clear;
+            // emptyStateView.Frame = TimeEntriesLogTableView.Frame;
```
willsb: Comment 👀

Toggl.Daneel/ViewSources/TimeEntriesLogViewSource.cs
```diff
@@ -90,40 +80,13 @@ protected override IEnumerable<TimeEntryViewModel> GetGroupAt(nint section)
...
-
-        public override nint RowsInSection(UITableView tableview, nint section)
-        {
-            if (shouldUseEmptyStateCells)
-                return emptyStateCellCount;
-
-            return base.RowsInSection(tableview, section);
-        }
-
         public new object GetItemAt(NSIndexPath indexPath)
```
willsb: Why shadow it in first place if you're only calling base?

Toggl.Daneel/ViewSources/TimeEntriesLogViewSource.cs
```diff
@@ -140,10 +103,7 @@ public override UISwipeActionsConfiguration GetLeadingSwipeActionsConfiguration(
 
         protected override void OnSectionAdded(NSIndexSet indexToAdd)
         {
-            if (IsEmptyState)
-                TableView.ReloadSections(indexToAdd, UITableViewRowAnimation.Automatic);
-            else
-                TableView.InsertSections(indexToAdd, UITableViewRowAnimation.Automatic);
+            TableView.InsertSections(indexToAdd, UITableViewRowAnimation.Automatic);
```
willsb: I think we can now officially 🔥 The `OnSecionAdded` method, right?
Because iirc this is the only place of usage (and it's now only calling the same as base, only without the begin/end updates)

Toggl.Daneel/ViewSources/TimeEntriesLogViewSource.cs
simonrozsival: Base method is `protected`, this is a way of exposing this publicly. I don't like it either, but I don't want to do any unnecessary refactoring in this PR.

PR #1760, opened 2018-03-19
Show confirmation dialog if there are changes in the edit TE view

### LINKED ISSUE #1543: Edit view: Ask confirmation when discarding changes
- [x] Ask confirmation: "Discard" or "Keep Editing" when a user taps 'X' in the edit view.
- [x] Ask confirmation only when a user has changed anything. Use the same logic we use in iOS #1042 
- [x] Don't ask anything if a user hasn't made any changes.
- [x] Changes to consider:
- Description
- Project/task
- Tags
- Start/end time/date
- Duration
- Billable flag
- [x] When a user changes the field (eg description) and then changes it back, then the app shouldn't ask this confirmation, and consider this as no changes were made.

### Edit view: confirmation
edit cancel edit

Zeplin: https://zpl.io/VDDzzR3

### LINKED ISSUE #1612: Edit view: Ask confirmation when discarding changes
- [x] Ask confirmation: 'Discard' or 'Continue Editing' when a user has made any changes and taps 'x' in the edit view. (Like in the Start view #1042)
- [x] Don't ask anything if a user hasn't made any changes.
- [x] Changes to consider:
- Description
- Project/task
- Tags
- Start/end time/date
- Duration
- Billable flag
- [x] When a user changes the field (eg description) and then changes it back, then the app shouldn't ask this confirmation, and consider this as no changes were made.

simonrozsival:
Closes #1612, closes #1543 

I am not sure if the behavior on Android will be correct (the back button?) - please test it and request changes if it breaks the Android app.

Squash whenever


kornelijepetak:
I have two observations:

1) The wording is now very confusing. When I start a new time entry, `discard time entry` makes sense, but once the TE exists, this is completely off. I am not discarding time entry in that case, but discarding `changes` to the TE. The same message cannot be used in both places. @pe0ny opinions?

2) Android breaks like this:
* Tap the TE (Edit View opens)
* Change the project
* Press back button (on toolbar)
* Confirmation prompt is shown
* Instead of clicking any of those buttons, click outside of the dialog
* :bug: Clicking on toolbar's back button no longer works

kornelijepetak:
Note that this also fixes #1543 if done right.

pe0ny:
@kornelijepetak 
Yes, in the edit view we should ask: "Discard changes?" https://zpl.io/VDDzzR3
And in the Start view, we should ask: "Discard this time entry?" https://zpl.io/2ZPM4Od

simonrozsival:
Also, I don't think that 2) is related to this PR - it is probably something you need to fix in Android in general (dismissing modals by tapping on the background - not calling the closing command correctly).

PR #1762, opened 2018-03-19
Fix top margin for sign out button

### LINKED ISSUE #1751: Settings view: Move Sign out button at the bottom
Since it's quite easy to sign out by accident, move Sign out button at the bottom of the screen, further from the other list like we do in iOS.

### Settings
settings
*Sign out cell should have top and bottom borders which doesn't stand out here in GH issue but are visible in Zeplin*

Zeplin: https://zpl.io/VD73lje

daividssilverio:
Closes #1751 

squash whenever


kornelijepetak:
As per this design, I think it should be `16dp` top margin.
https://zpl.io/VD73lje

However, I've now realized that the `synced` status row is missing there. I will open an issue for that.

kornelijepetak:
@lizaPlotnikova I see you have updated above mentioned Zeplin design. It is reverted to the old one. Was that intentional? Because it is no longer in sync with what is specified on this issue? 

@pe0ny can you check too?

pe0ny:
@kornelijepetak needs design?

kornelijepetak:
Not anymore. I was just pointing to the error in design. Which you promptly corrected. 😄

daividssilverio:
Now it's fixed. o/

PR #1765, opened 2018-03-20
Stop crashes once and for all

### LINKED ISSUE #1523: Various GroupedCollectionTableViewSource crashes

willsb:
Closes #1523 AND LEAVES IT CLOSED

Test thoroughly and merge it when you feel like it


amulware:
Requested two reviews. Others are VERY welcome.

Toggl.Foundation/DataSources/TimeEntriesDataSource.cs
```diff
@@ -56,7 +56,7 @@ internal sealed class TimeEntriesDataSource : ITimeEntriesSource
             this.timeService = timeService;
             this.shortcutCreator = shortcutCreator;
 
-            TimeEntryCreated = timeEntryCreatedSubject.AsObservable();
+            TimeEntryCreated = timeEntryCreatedSubject.Select(TimeEntry.From).AsObservable();
```
simonrozsival: I don't think this is necessary - you already call this on every occasion where you call `OnNext` on this subject. Is there some reason I don't see?

Toggl.Foundation.MvvmCross/ViewModels/TimeEntriesLogViewModel.cs
```diff
@@ -82,16 +83,18 @@ public sealed class TimeEntriesLogViewModel : MvxViewModel
...
             var updateDisposable =
                 dataSource.TimeEntries.TimeEntryUpdated
-                          .Subscribe(onTimeEntryUpdated);
+                    .Subscribe(onTimeEntryUpdated);
 
             var createDisposable =
                 dataSource.TimeEntries.TimeEntryCreated
                     .Where(isNotRunning)
-                    .Subscribe(safeInsertTimeEntry);
+                    .Buffer(TimeSpan.FromMilliseconds(500))
```
simonrozsival: Magic constant! And `500ms` is the final value? Wouldn't `100ms` or even less be enough?

Toggl.Foundation.MvvmCross/ViewModels/TimeEntriesLogViewModel.cs
```diff
@@ -153,29 +156,39 @@ private void onTimeEntryUpdated((long Id, IDatabaseTimeEntry Entity) tuple)
                     return;
                 }
 
-                safeInsertTimeEntry(timeEntry);
+                safeInsertTimeEntries(new[] { timeEntry }.ToList());
             }
         }
 
-        private void safeInsertTimeEntry(IDatabaseTimeEntry timeEntry)
+        private void safeInsertTimeEntries(IList<IDatabaseTimeEntry> timeEntries)
```
simonrozsival: Wouldn't `IEnumerable` be enough? You enumerate it only once.. There wouldn't be a need for such an ugly conversion at line `159`.

simonrozsival:
It seems to work fine so far.

Toggl.Foundation/DataSources/TimeEntriesDataSource.cs
willsb: Nope, you are correct

Toggl.Foundation.MvvmCross/ViewModels/TimeEntriesLogViewModel.cs
willsb: Reduced it to 100. Should be enough

Toggl.Foundation.MvvmCross/ViewModels/TimeEntriesLogViewModel.cs
willsb: 👍

Toggl.Foundation.MvvmCross/ViewModels/TimeEntriesLogViewModel.cs
```diff
@@ -21,6 +22,7 @@ namespace Toggl.Foundation.MvvmCross.ViewModels
     [Preserve(AllMembers = true)]
     public sealed class TimeEntriesLogViewModel : MvxViewModel
     {
+        private const int timeEntryCreationBuffer = 100;
```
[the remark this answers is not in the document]
amulware: I would prefer this being a timespan and having a name change as well.

Toggl.Foundation.MvvmCross/ViewModels/TimeEntriesLogViewModel.cs
```diff
@@ -82,16 +84,18 @@ public sealed class TimeEntriesLogViewModel : MvxViewModel
...
                 dataSource.TimeEntries.TimeEntryUpdated
-                          .Subscribe(onTimeEntryUpdated);
+                    .Subscribe(onTimeEntryUpdated);
 
             var createDisposable =
                 dataSource.TimeEntries.TimeEntryCreated
                     .Where(isNotRunning)
-                    .Subscribe(safeInsertTimeEntry);
+                    .Buffer(TimeSpan.FromMilliseconds(timeEntryCreationBuffer))
+                    .DistinctUntilChanged()
```
amulware: Why the `DistinctUntilChanged`?

Toggl.Foundation.MvvmCross/ViewModels/TimeEntriesLogViewModel.cs
```diff
@@ -82,16 +84,18 @@ public sealed class TimeEntriesLogViewModel : MvxViewModel
...
             var updateDisposable =
                 dataSource.TimeEntries.TimeEntryUpdated
-                          .Subscribe(onTimeEntryUpdated);
+                    .Subscribe(onTimeEntryUpdated);
 
             var createDisposable =
                 dataSource.TimeEntries.TimeEntryCreated
                     .Where(isNotRunning)
-                    .Subscribe(safeInsertTimeEntry);
+                    .Buffer(TimeSpan.FromMilliseconds(timeEntryCreationBuffer))
```
amulware: Does this lead to time entries showing up 100ms late after a sync/manual creation?

Toggl.Foundation.MvvmCross/ViewModels/TimeEntriesLogViewModel.cs
```diff
@@ -153,29 +157,39 @@ private void onTimeEntryUpdated((long Id, IDatabaseTimeEntry Entity) tuple)
...
+                        TimeEntries.InsertInChildCollection(collectionIndex, timeEntryViewModel);
+                    }
+                }
+                else
+                {
+                    var timeEntriesToAdd = group.Select(te => new TimeEntryViewModel(te, durationFormat)).ToArray();
+                    var newCollection = new TimeEntryViewModelCollection(indexDate, timeEntriesToAdd, durationFormat);
+                    var foundIndex = TimeEntries.IndexOf(TimeEntries.FirstOrDefault(x => x.Date < indexDate));
+                    var indexToInsert = foundIndex == -1 ? TimeEntries.Count : foundIndex;
+                    TimeEntries.Insert(indexToInsert, newCollection);
```
amulware: I wouldn't mind having the two blocks here extracted as named methods for a bit better overview.

Toggl.Foundation/DataSources/TimeEntriesDataSource.cs
```diff
@@ -135,12 +135,13 @@ public IObservable<IDatabaseTimeEntry> Create(IDatabaseTimeEntry entity)
             => repository
                 .UpdateWithConflictResolution(entity.Id, entity, alwaysCreate, rivalsResolver)
                 .Select(result => ((CreateResult<IDatabaseTimeEntry>)result).Entity)
+                .Select(TimeEntry.From)
```
amulware: Who dis?

amulware: Also all the ones below. Wrapping for thread safety because they come from realm? Is that necessary? (If so, why? Was this causing crashes?)

amulware:
Two more questions:

1. I understand what the code does, but can you explain how this fixes the crash?
2. Why ~~500~~ 100ms?

Toggl.Foundation/DataSources/TimeEntriesDataSource.cs
willsb: This starts causing crashes once we `Buffer` the time entries

amulware: Ah, the buffer causes it to escape the current thread since it's async, yes?

Toggl.Foundation.MvvmCross/ViewModels/TimeEntriesLogViewModel.cs
willsb: 🔥

Toggl.Foundation.MvvmCross/ViewModels/TimeEntriesLogViewModel.cs
willsb: Yes

willsb:
It fixes the crash because we were sending two different add signals (one for adding sections and one for adding items) and before the section was animated the number of items would change again and that would explode. Now we add all the buffered items in bulk and all is well again.

100ms is enough time to buffer and get all the changes in the same batch that will update the UI

Toggl.Foundation/DataSources/TimeEntriesDataSource.cs
amulware: Got that confirmed internally.

Toggl.Foundation.MvvmCross/ViewModels/TimeEntriesLogViewModel.cs
```diff
@@ -153,30 +156,45 @@ private void onTimeEntryUpdated((long Id, IDatabaseTimeEntry Entity) tuple)
...
+                var indexDate = group.Key;
+                var collectionIndex = TimeEntries.IndexOf(x => x.Date.LocalDateTime == indexDate);
+                var groupExists = collectionIndex >= 0;
+                if (groupExists)
+                {
+                    foreach (var timeEntry in group)
+                    {
+                        var timeEntryViewModel = new TimeEntryViewModel(timeEntry, durationFormat);
+                        TimeEntries.InsertInChildCollection(collectionIndex, timeEntryViewModel);
+                    }
```
amulware: I'd like this loop in a method as well since it is on the same level as the method in the else block below unless you have a good reason why this should be here.

Toggl.Foundation.Tests/MvvmCross/ViewModels/TimeEntriesLogViewModelTests.cs
```diff
@@ -170,6 +171,7 @@ public sealed class WhenReceivingAnEventFromTheTimeEntryCreatedObservable : Time
                     await ViewModel.Initialize();
 
                     TimeEntryCreatedSubject.OnNext(NewTimeEntry);
+                    await ThreadingTask.Delay(200);
```
amulware: You made our tests last a whole second longer. :D
Should we refactor this to use a scheduler in a follow-up pr?

amulware:
Oh, one more question: These changes are in shared code, for a pretty iOS specific bug. Are we ok with that?

willsb:
Regarding both the possible refactor and the "solving iOS bugs in shared code":

This will all change in the future. Me and @simonrozsival are very much hating the current state of the StartTimeEntryViewModel. With Interactors coming, we'll eventually kill all of the DataSources, at which point the way we handle emission of updates will also change. So no, I don't think this will be a problem at all; this is bound to be refactored anyway.

amulware:
Alright, that sounds good.

PR #1772, opened 2018-03-21
Fix settings layout

simonrozsival:
This PR fixes the positioning of the logout button + fixes bounding in the settings view.

Squash whenever


arturs105:
Does this close #1688?

simonrozsival:
- bouncing
- it does not, but it could - I will change it to close that issue

PR #1774, opened 2018-03-21
Fail integration tests on failed subscription

amulware:
Previously (i.e. currently) tests that require a subscription plan change fail on the first api call that requires that change.

The bug is in the actual subscriptions endpoint of backend, so I much rather fail on that call, for better error messages.

This does that (and the same for other helper api calls).

In the future, we really should consider getting rid of these helpers as much as possible, and just using properly implemented API calls which don't require the ugly hardcoding done here and will error properly automatically. But for now, I think this is alright.


amulware:
🦑

amulware:
For reference, here are the tests now failing with nicer errors: https://www.bitrise.io/build/bb2a55e06a60bc1f

Issue #1574, opened 2018-02-22
Edit view: Remove billable field for free WS


pe0ny:
Do not show billable field to users who are under the free WS.

PR #1779, opened 2018-03-21
Refactor WorkspaceSource into Interactors

willsb:
Ref #1574 

The code for checking feature availability was a mess. So I refactored it!

In the great effort of removing data sources, this PR removes the Workspaces data source and refactors its methods into small interactors. Note that we are no longer checking for the pro feature, but rather asking whether a feature is enabled or not. This is much much safer for the future, when the business logic for feature availability might change.

Please review each commit individually and don’t be daunted by the amount of changed files; Most of them are basic changes.

Given the size of this PR and the fact that it prevents the current milestone from being closed, I’ll mark it as blocking and ask you to review it ASAP to prevent big conflicts.

The usual “At least two pairs of eyeballs need to review this” holds true.


amulware:
Only had a quick glance, but I gotta think a bit about this exact set of interactors. There are probably different ways of doing it and arguments for each. We should make sure we are all on the same page what scope interactors should have, and how they should be implemented exactly.

Looking forward to getting this merged though!

willsb:
Note to reviewers: Functional tests of the affected ViewModels are very appreciated.

Toggl.Foundation/Interactors/Abstract/WorkspaceHasFeatureInteractor.cs
```diff
@@ -0,0 +1,26 @@
+﻿using System;
+using System.Reactive.Linq;
+using Toggl.Multivac;
+using Toggl.PrimeRadiant;
+
+namespace Toggl.Foundation.Interactors
+{
+    internal abstract class WorkspaceHasFeatureInteractor<TValue> : IInteractor<IObservable<TValue>>
+    {
+        private readonly ITogglDatabase database;
```
kornelijepetak: Not sure if I don't understand something that I should, but every inheritor has its own:
* private `IDatabase` property
* constructor that propagates the `IDatabase` instance to base constructor
* `Ensure.Argument.IsNotNull` for that same instance.

I don't think inheritors should have their own `IDatabase` member because `WorkspaceHasFeatureInteractor.Database` could be `protected readonly`. Also, `Ensure.Argument.IsNotNull` is thus unnecessarily called multiple times.

I'd go with protected database.

Toggl.Foundation/Interactors/IInteractorFactory.cs
```diff
@@ -7,12 +8,42 @@ namespace Toggl.Foundation.Interactors
 {
     public interface IInteractorFactory
     {
+        #region Time Entries
```
kornelijepetak: I'm still somehow not satisfied with the sheer amount of methods in that same interface. I don't mind this interface declaration, but I kind of mind the list at the call-site.

But for the time being, I will not get too vocal about this. We can easily refactor this if it gets too bloated.

Toggl.Foundation.MvvmCross/ViewModels/EditProjectViewModel.cs
```diff
@@ -26,9 +26,10 @@ public sealed class EditProjectViewModel : MvxViewModel<string, long?>
         private readonly Random random = new Random();
         private readonly ITogglDataSource dataSource;
         private readonly IDialogService dialogService;
+        private readonly IInteractorFactory interactorFactory;
```
kornelijepetak: I'm also thinking about how to properly name this variable.
It feels a bit 🇭🇺naming this according only to the two used patterns' names, without any other references.

I can't say I have much better solutions, but I'd like us to think about it.

* **`interactorFactory`**`.AreProjectsBillableByDefault(workspaceId).Execute();`

* **`domain`**`.AreProjectsBillableByDefault(workspaceId).Execute();`
* **`data`**`.AreProjectsBillableByDefault(workspaceId).Execute();`
* **`logic`**`.AreProjectsBillableByDefault(workspaceId).Execute();`
* **`business`**`.AreProjectsBillableByDefault(workspaceId).Execute();`
* **`businessLogic`**`.AreProjectsBillableByDefault(workspaceId).Execute();`
* **`toggl`**`.AreProjectsBillableByDefault(workspaceId).Execute();`

I don't know... any opinions? @amulware

simonrozsival: It would sort of make sense to have functions starting with `CreateXYZ` in an object which is a _factory_. But writing (and reading) `interactorFactory.CreateAreProjectsBillableByDefault(workspaceId).Execute()` seems nuts. But maybe we could then have a second object `IInteractorsCollection interactors` and use it as "create + execute": `interactors.AreProjectsBillableByDefault(workspaceId)` which would internally do the above. But this is probably overthinking this. I am fine with the way it is right now.

Toggl.Foundation.MvvmCross/ViewModels/EditTimeEntryViewModel.cs
```diff
@@ -475,7 +479,7 @@ private void onPreferencesChanged(IDatabasePreferences preferences)
 
         private async Task updateFeaturesAvailability()
         {
-            AllowsBillableRates = await dataSource.Workspaces.WorkspaceHasFeature(workspaceId, WorkspaceFeatureId.Pro);
+            AllowsBillableRates = await interactorFactory.WorkspaceAllowsBillableRates(workspaceId).Execute();
```
simonrozsival: I really like how the business logic details are hidden 👍

Toggl.Foundation/Interactors/Workspace/AreCustomColorsEnabledForWorkspaceInteractor.cs
```diff
@@ -0,0 +1,25 @@
...
+using Toggl.Multivac;
+using Toggl.PrimeRadiant;
+using static Toggl.Multivac.WorkspaceFeatureId;
+
+namespace Toggl.Foundation.Interactors
+{
+    internal sealed class AreCustomColorsEnabledForWorkspaceInteractor : WorkspaceHasFeatureInteractor<bool>
+    {
+        private readonly long workspaceId;
+        private readonly ITogglDatabase database;
```
simonrozsival: 🔥 - this interactor doesn't need the database, just the base needs it.

Toggl.Foundation/Interactors/Workspace/AreProjectsBillableByDefaultInteractor.cs
```diff
@@ -0,0 +1,36 @@
...
+
+            this.database = database;
+            this.workspaceId = workspaceId;
+        }
+
+        public override IObservable<bool?> Execute()
+            => CheckIfFeatureIsEnabled(workspaceId, WorkspaceFeatureId.Pro)
+                .SelectMany(isPro =>
+                {
+                    if (!isPro)
```
simonrozsival: I would myself use a ternary operator instead of this if and two `return`s, but this is also fine by me.

Toggl.Foundation/Interactors/Workspace/WorkspaceAllowsBillableRatesInteractor.cs
```diff
@@ -0,0 +1,25 @@
...
+using Toggl.Multivac;
+using Toggl.PrimeRadiant;
+using static Toggl.Multivac.WorkspaceFeatureId;
+
+namespace Toggl.Foundation.Interactors
+{
+    internal sealed class WorkspaceAllowsBillableRatesInteractor : WorkspaceHasFeatureInteractor<bool>
+    {
+        private readonly long workspaceId;
+        private readonly ITogglDatabase database;
```
simonrozsival: :fire: - this interactor doesn't need reference to the database, just its base needs it

Toggl.Foundation/Interactors/Abstract/WorkspaceHasFeatureInteractor.cs
willsb: Agreed

Toggl.Foundation/Interactors/Workspace/GetAllWorkspacesInteractor.cs
```diff
@@ -0,0 +1,23 @@
...
+using Toggl.PrimeRadiant;
+using Toggl.PrimeRadiant.Models;
+
+namespace Toggl.Foundation.Interactors
+{
+    internal sealed class GetAllWorkspacesInteractor : IInteractor<IObservable<IEnumerable<IDatabaseWorkspace>>>
+    {
+        private readonly ITogglDatabase database;
+
+        public GetAllWorkspacesInteractor(ITogglDatabase database)
```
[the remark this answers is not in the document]
willsb: I prefer this to receive a database for simplicity sake 😄

Toggl.Foundation.MvvmCross/ViewModels/SelectProjectViewModel.cs
```diff
@@ -112,13 +118,14 @@ public override void Prepare(SelectProjectParameter parameter)
...
             await base.Initialize();
 
-            dataSource.Workspaces
-                      .GetAll()
-                      .Subscribe(workspaces =>
-                      {
-                          allWorkspaces = workspaces.ToList();
-                          UseGrouping = allWorkspaces.Count > 1;
-                      });
+            interactorFactory
```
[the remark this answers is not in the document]
willsb: Agreed

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
```diff
@@ -252,11 +252,11 @@ public override void Prepare(StartTimeEntryParameters parameter)
         {
             await base.Initialize();
 
-            await setBillableValues(0);
-
             TextFieldInfo =
                 await dataSource.User.Current.Select(user => TextFieldInfo.Empty(user.DefaultWorkspaceId));
-            
+
+            await setBillableValues(lastProjectId);
```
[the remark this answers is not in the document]
willsb: That will not get called. It will early return because the project did not change

Toggl.Foundation/Interactors/Workspace/AreProjectsBillableByDefaultInteractor.cs
willsb: Because the expression in the later return is quite long, I prefer to leave this as is

PR #1781, opened 2018-03-21
Forgot password

kornelijepetak:
Closes #1267 

About time.

Squash whenever.

PR #1782, opened 2018-03-22
Add try logging in on android sign up

### LINKED ISSUE #1667: Show correct UI if existing user tries to sign up
The Android part of #1221

daividssilverio:
Closes #1667 

Android part of #1221 

I updated the layout so it won't look weird when the "Try logging in" message appears. The Terms of Service and Privacy Policy texts were getting in the way.

squash whenever


kornelijepetak:
Basically this seems alright, however the toolbar was ditched in #1781.

What I did there is I've introduced normal UI elements that look like toolbar so that binding does not require weak subscriptions for the title. This is what we do in other activities where we kind of _extend_ the toolbar so that we can do what we want with it and that it still looks like a toolbar.

However, in this case I was following this route for convenience, but I would like @willsb to provide his insight. Should I revert to the built-in toolbar (more boilerplate, inconvenient binding in code-behind) or should @daividssilverio wait for #1781 to get merged and then reapply only the needed changes?

daividssilverio:
I wonder if that won't make stuff more difficult later on, if we want to make use of the design library, for  animations and other stuff together with the AppBarLayout, for example.
Also having to handle the up navigation by ourselves.

Couldn't we create custom bindings for toolbars that would handle that kind of stuff?

Toggl.Giskard/Activities/LoginActivity.cs
```diff
@@ -32,15 +37,23 @@ protected override void OnCreate(Bundle bundle)
...
             SupportActionBar.SetDisplayHomeAsUpEnabled(true);
             SupportActionBar.SetDisplayShowHomeEnabled(true);
 
             toolbar.NavigationClick += navigationClick;
+            disposable = ViewModel.WeakSubscribe<PropertyChangedEventArgs>(nameof(ViewModel.Title), onTitleChanged);
+        }
+
+        private void onTitleChanged(object sender, PropertyChangedEventArgs args)
+        {
+            var toolbar = FindViewById<Toolbar>(Resource.Id.LoginToolbar);
```
willsb: Cache the call to `FindViewById`, pls

Toggl.Giskard/Activities/LoginActivity.cs
```diff
@@ -32,15 +37,23 @@ protected override void OnCreate(Bundle bundle)
...
+            if (toolbar != null)
+            {
+                toolbar.Title = ViewModel.Title;
```
willsb: I would not check for nulls here. Let the app crash so we know in development time that there's no toolbar in the xml.

willsb:
Regarding toolbars: I prefer using the system provided one unless it's impossible not to (e.g.: In the start time entry view model). Is this the case, @kornelijepetak? If not, I'm with @daividssilverio on this one.

kornelijepetak:
Ok, I will revert my thing to the built-in. Proceed with the merge as soon as @willsb approves.

Toggl.Giskard/Activities/LoginActivity.cs
daividssilverio: Hmm, that's fair. I'll remove the null check then.

PR #1783, opened 2018-03-22
Add manual switch

### LINKED ISSUE #1766: Settings view: Add manual switch
- [ ] By default manual mode is off.
- [ ] When a user turns on manual mode then play icon in the timer page will be replaced by the ‘+’ icon.

### Settings
settings

Zeplin: https://zpl.io/VD73lje

'+' Icon is available here: https://zpl.io/2EdyAwM

daividssilverio:
Closes #1766 

It still needs a more appropriate '+' icon for the play button, but it is ok to check the code.

squash whenever


Toggl.Giskard/Activities/MainActivity.cs
```diff
@@ -51,10 +52,23 @@ protected override void OnCreate(Bundle bundle)
...
+            manualModeDisposable =
+                ViewModel.WeakSubscribe<PropertyChangedEventArgs>(nameof(ViewModel.IsInManualMode), onManualModeToggle);
+        }
+
+        private void onManualModeToggle(object sender, PropertyChangedEventArgs args)
+        {
+            updatePlayButtonIcon();
+        }
+
+        private void updatePlayButtonIcon()
```
kornelijepetak: Since you are not calling this method at any other place, it's body can be moved into the handler itself and method can be removed.

Toggl.Giskard/Activities/MainActivity.cs
```diff
@@ -51,10 +52,23 @@ protected override void OnCreate(Bundle bundle)
...
             stopButton = FindViewById<FloatingActionButton>(Resource.Id.MainStopButton);
+            updatePlayButtonIcon();
             coordinatorLayout = FindViewById<CoordinatorLayout>(Resource.Id.MainCoordinatorLayout);
-
-            disposable =
+            
+            syncProgressDisposable =
                 ViewModel.WeakSubscribe<PropertyChangedEventArgs>(nameof(ViewModel.SyncingProgress), onSyncChanged);
+            manualModeDisposable =
+                ViewModel.WeakSubscribe<PropertyChangedEventArgs>(nameof(ViewModel.IsInManualMode), onManualModeToggle);
```
kornelijepetak: Is the handler always executed on UI thread? Can `SetImageResource` make problems if it's not?

Toggl.Giskard/Resources/layout/SettingsActivity.axml
```diff
@@ -225,6 +225,50 @@
                 android:background="@color/separator"
                 android:layout_height="0.5dp"
                 android:layout_width="match_parent" />
+            <View
+                    android:layout_marginTop="24dp"
```
kornelijepetak: This is just FYI, not something you need to change at this time, but there's a consensus between me and @willsb about the order of attributes:

* `android:id`
* `local:MvxBind`
* `tools:`
* all others
* `android:layout_`

In each of those categories, the attributes are sorted by ascending length (attribute name + value).

Now, you don't need to change this now because I intend to submit a PR with all AXML formatted that way (by a tool) cause there are other places where this is not respected. It's just FYI.

kornelijepetak: And if you are asking yourself why, @willsb likes aesthetics 😄

Toggl.Giskard/Activities/MainActivity.cs
daividssilverio: I'm calling it in the constructor to update the the button right after the FindViewById

Toggl.Giskard/Resources/layout/SettingsActivity.axml
daividssilverio: Hmm, that's nice. I like this kind of stuff too (more specifically on our classes). 
Never really cared about the xml aesthetics though. I usually like to use the auto refactor for this. It gives us some consistency.

Toggl.Giskard/Activities/MainActivity.cs
daividssilverio: Hmm, SetImageResource does coding and decoding of the drawable in the Ui thread: 
I didn't get any errors while testing. I know that ui updates should be done on the Ui thread, it might cause problems if run on other threads.
https://developer.android.com/reference/android/widget/ImageView.html#setImageResource%28int%29

Not sure how this WeakSubscribe works. With Kotlin/Java, when I used Rx I had to define where that would run using: subscribeOn, observeOn. Is there something similar here?

Toggl.Giskard/Activities/MainActivity.cs
```diff
@@ -51,10 +52,23 @@ protected override void OnCreate(Bundle bundle)
...
+        }
+
+        private void onManualModeToggle(object sender, PropertyChangedEventArgs args)
+        {
+            updatePlayButtonIcon();
+        }
+
+        private void updatePlayButtonIcon()
+        {
+            playButton?.SetImageResource(ViewModel.IsInManualMode ? Resource.Drawable.play_white : Resource.Drawable.add_new);
```
willsb: MvvmCross provides a `DrawableId` icon. Create a converter to convert from bool to int and do this in a xml binding instead

Toggl.Giskard/Resources/layout/SettingsActivity.axml
willsb: A E S T H E T I C S

daividssilverio: I tried to make it look better

Issue #1622, opened 2018-03-02
Reports: Pie chart


pe0ny:
Reports view includes three different sections: **Summary**, **Bar chart** and **Pie chart** with a list. In this issue implement the **Pie chart section with a list**.

- [x] Pie chart shows top projects with percentages.
- [ ] We show all the projects that duration exceed 10% of the total duration. All the projects that duration doesn't exceed 10% of the total duration will be generated under 'Other' category which may contain one or more projects (the reason behind is to group smaller project so they don't clutter the pie chart and the list below).
- [x] The list view includes: a project name, client (if any), percentage and duration
- [x] Sort projects by duration (desc.)

### Reports
![reports](https://user-images.githubusercontent.com/5064445/36897665-330502fc-1e20-11e8-9453-400746f664f1.png)

Zeplin: https://zpl.io/2yBEpQL

simonrozsival:
Wouldn't it be better to group the projects just in the pie chart, but display all of them in the list?

amulware:
I would agree with that (I actually thought that's how it was designed). Up to @pe0ny

pe0ny:
I like Simon's idea - let's do it so 🙌

PR #1789, opened 2018-03-22
Pie Chart

willsb:
Ref #1622

This only adds the pie chart. Quick reviews are appreciated so I can continue work on the item list.


Toggl.Giskard/Views/PieChartView.cs
```diff
@@ -0,0 +1,121 @@
...
+
+                    // Restore the original coordinate system.
+                    canvas.Restore();
+                }
+
+                startDegrees = endDegrees;
+            }
+        }
+    }
+}
```
simonrozsival: \n

Toggl.Giskard/Views/PieChartView.cs
```diff
@@ -0,0 +1,121 @@
...
+        }
+
+        public override void Draw(Canvas canvas)
+        {
+            var viewCenterX = Width * 0.5f;
+            var viewCenterY = Height * 0.5f;
+            var radius = viewCenterX;
+            var totalSeconds = (float)Segments.Select(x => x.TrackedTime.TotalSeconds).Sum();
+
+            var startDegrees = 270.0f;
```
simonrozsival: 270 degrees?? That is normally "six o'clock" - on Android this corresponds to "twelve o'clock" (normally 90deg)?

Toggl.Giskard/Views/PieChartView.cs
```diff
@@ -0,0 +1,121 @@
...
+
+            var oval = new RectF(0, 0, Width, Height);
+
+            foreach (var segment in Segments)
+            {
+                var segmentPaint = new Paint();
+                segmentPaint.Color = MvxColor.ParseHexString(segment.Color).ToAndroidColor();
+
+                var percent = (float)segment.TrackedTime.TotalSeconds / totalSeconds;
+                var sweepDegrees = 360.0f * percent;
```
simonrozsival: `360.0f` would deserve its own constant.

Toggl.Daneel/Views/Reports/PieChartView.cs
```diff
@@ -47,36 +45,35 @@ public override void Draw(CGRect rect)
...
 
             foreach (var segment in Segments)
             {
                 ctx.SetFillColor(MvxColor.ParseHexString(segment.Color).ToNativeColor().CGColor);
 
-                var percent = (nfloat)(segment.TrackedTime.TotalSeconds / totalValue);
-                // Calculate end angle
-                var endAngle = startAngle + 2 * pi * percent;
+                var percent = (float)segment.TrackedTime.TotalSeconds / totalSeconds;
+                var endAngle = startAngle + 2 * FloatPi * percent;
```
simonrozsival: `2 * FloatPi == (float)FullCircle` - maybe you could add a float variant of the `FullCircle` constant?

Toggl.Multivac/Math.cs
```diff
@@ -8,6 +8,8 @@ public static class Math
 
         public const double FullCircle = 2 * System.Math.PI;
 
+        public const float FloatPi = (float)System.Math.PI;
```
kornelijepetak: I am really disturbed by the 🇭🇺 for this constant.
Since there is no `Pi` constant in `Toggl.Multivac.Math` I would go with the simple `Pi`.

Additionally, even though it probably won't do us any problems here, I am a strong believer in casting things down only at the last step so that the precision errors are not accumulated. Since this is not calculated thousands of times per second to create performance issues, I would do all things in doubles and then cast down to float only when API requires us to, i.e. at
`ctx.AddArc(viewCenterX, viewCenterY, radius, startAngle, endAngle, clockwise: false);`

Toggl.Daneel/Views/Reports/PieChartView.cs
kornelijepetak: No 🇭🇺, please! 😛

kornelijepetak:
I have also noted that when I open the report screen, it takes between 2 and 4 seconds until the pie chart appears. The containing card is visible from the get go. If it really needs that much time to calculate stuff, some kind of `loading` should be shown until data is ready.

willsb:
It needs this time to fetch stuff from backend. Not a biggie, the loading will probably come in the future

Toggl.Giskard/Views/PieChartView.cs
willsb: Clarified on Slack :smile:

PR #1793, opened 2018-03-23
Fix the URL of the subscription endpoint

simonrozsival:
Integration tests: https://www.bitrise.io/build/eed2c3a7fd5007fb

Squash whenever


amulware:
The new endpoint returns `500`, let's wait and see before we make any changes.

amulware:
The relevant tests now pass.

PR #1795, opened 2018-03-23
Polish the edit view to adapt size when billable is hidden

### LINKED ISSUE #1794: Polish the hidden billable field
This is a follow-up of #1574 after #1786 was merged.

When the billable control is hidden, the row is still there, it is just empty:

missing_billable

simonrozsival:
Closes #1794 

Squash whenever


kornelijepetak:
@simonrozsival I have noticed that when I switch from the project which is in a free WS to a project in workspace that allows billable, the billable goes down in UI and I am not able to scroll. I see it only partially.

image

simonrozsival:
Scrolling is disabled until the edit view fills the whole screen (a very long description in case of iPhone X) - maybe we should consider enabling it all the time.

I am not able to reproduce the state you have in the screenshot - it is true that content size is adjusted a bit later and it "jumps" into the correct place, but it always does adjust the size.

Does this happen consistently whenever you change free/pro workspace for the TE?

arturs105:
Tested the same thing as @kornelijepetak, however i was able to scroll, but we should still do something about this.

Also, when changing from project with billable available to one without, the billable row goes away, but the edit view remains the same size, which leads to a bit of unused space.

Toggl.Daneel/ViewControllers/EditTimeEntryViewController.cs
```diff
@@ -272,5 +277,15 @@ private void setupDismissingByTappingOnBackground()
                 modalPresentationController.AdditionalContentView.AddGestureRecognizer(tapToDismiss);
             }
         }
+
+        private async void onPropertyChanged(object sender, PropertyChangedEventArgs e)
+        {
+            if (e.PropertyName == "IsBillableAvailable")
```
arturs105: Can't you use `nameof(EditTimeEntryViewModel.IsBillableAvailable)`?

Toggl.Daneel/ViewControllers/EditTimeEntryViewController.cs
```diff
@@ -272,5 +277,15 @@ private void setupDismissingByTappingOnBackground()
...
+            {
+                await Task.Delay(10);
```
arturs105: Why the delay?

simonrozsival: I would really like to implement it in a different way - I just need this to happen the next time when the layout is rendered and the new size of the stack view is known. If you have any tip how to do that without the delay I will gladly change this.

Toggl.Daneel/ViewControllers/EditTimeEntryViewController.cs
simonrozsival: I tried that and it didn't work that way.

arturs105: I just tested this out and it's actually working. Could you try once more?

simonrozsival: Oh, I might messed it up then. I will try it.

Toggl.Daneel/ViewControllers/EditTimeEntryViewController.cs
arturs105: You could set some flag here, then override some lifecycle method and check the flag in that method, though i'm not entirely sure about this.

Or you could bind the height to `IsBillableAvailable` with the bool converter.

Toggl.Daneel/ViewControllers/EditTimeEntryViewController.cs
willsb: Actually, can you weak subscribe instead? 😄

Toggl.Daneel/ViewControllers/EditTimeEntryViewController.cs
willsb: Does [this](https://developer.apple.com/documentation/quartzcore/cadisplaylink) help?

willsb: Artur's Binding idea sounds better, however

simonrozsival: I can't use the binding with a converter - the height of the whole view is dynamic, it depends on how many lines the description has and on if the billable switch is visible or not.

simonrozsival:
I updated this PR. It should now work well and the code should be clean enough.

PR #1796, opened 2018-03-23
Handle UserIsMissingApiTokenException on Login & Signup

### LINKED ISSUE #1701: Catch `UserIsMissingApiTokenException` on login and signup
When this exception is encountered on login/signup, let's retry the endpoint say... 2? more times, after a couple of seconds each. Maybe once after 2 seconds, and then after 10, just to make sure?

That should give our replicas enough time to catch up and return us a valid API token.

If the retries still fail to get us a token, then we'll fail as we would on any server error.

Follow up to #1652

daividssilverio:
Closes #1701
~WIP! Still trying to do it. Until now I've create an extension to retry an operation based on an exception and number of retries.
I added some tests just to try it out, right now it's not working.~

It's working. I've used the new retryWhen operator introduced in #2226.


willsb:
Ref #1701

Toggl.Foundation/Login/LoginManager.cs
```diff
@@ -54,12 +61,16 @@ public IObservable<ITogglDataSource> Login(Email email, Password password)
...
-                    .SelectMany(database.User.Create)
-                    .Select(dataSourceFromUser)
-                    .Do(shortcutCreator.OnLogin);
+                .Clear()
+                .SelectMany(_ => apiFactory.CreateApiWith(credentials).User.Get())
+                .Select(User.Clean)
+                .SelectMany(database.User.Create)
+                .Select(dataSourceFromUser)
+                .Do(shortcutCreator.OnLogin)
+                .DelayedConditionalRetry(2,
```
simonrozsival: What does `2` mean? It isn't clear to me without looking at the definition of the function - let's give it a name and make it a private const of this class.

Toggl.Foundation/Login/LoginManager.cs
```diff
@@ -54,12 +61,16 @@ public IObservable<ITogglDataSource> Login(Email email, Password password)
...
-                    .Select(dataSourceFromUser)
-                    .Do(shortcutCreator.OnLogin);
+                .Clear()
+                .SelectMany(_ => apiFactory.CreateApiWith(credentials).User.Get())
+                .Select(User.Clean)
+                .SelectMany(database.User.Create)
+                .Select(dataSourceFromUser)
+                .Do(shortcutCreator.OnLogin)
+                .DelayedConditionalRetry(2, 
+                    attempt => TimeSpan.FromSeconds(attempt == 1 ? 2 : 10),
```
simonrozsival: This would deserve two private readonly properties as well (2s, 10s) and maybe a function so it is obvious that for the attempt `1` the delay (is it a delay?) is different than for the other.

Toggl.Foundation/Login/LoginManager.cs
```diff
@@ -87,7 +98,12 @@ public IObservable<ITogglDataSource> SignUp(Email email, Password password)
                     .Select(User.Clean)
                     .SelectMany(database.User.Create)
                     .Select(dataSourceFromUser)
-                    .Do(shortcutCreator.OnLogin);
+                    .Do(shortcutCreator.OnLogin)
+                    .DelayedConditionalRetry(2,
```
simonrozsival: Same as above. Since the whole `DelayedConditionalRetry` call is quite complex and is repeated twice for the same use case, we could move it into a separate method.

Toggl.Foundation/Extensions/RxExtensions.cs
```diff
@@ -0,0 +1,36 @@
...
+                    shouldRetryOn(exception)
+                    ? Observable.Throw<(bool succeeded, T t, Exception e)>(exception)
+                    : Observable.Return((succeeded: false, t: default(T), e: exception))
+                ).Retry(maxRetries + 1)
+                .SelectMany(result =>
+                    result.succeeded ? Observable.Return(result.t) : Observable.Throw<T>(result.e)
+                );
+        }
+    }
+}
```
simonrozsival: `\n`

Toggl.Foundation/Login/LoginManager.cs
simonrozsival: Does resubscribing to the login observable really repeat the HTTP request? (see the `BaseApi` class)

Toggl.Foundation/Login/LoginManager.cs
willsb: `attempt => TimeSpan.FromSeconds(attempt * 8 - 6)`

Toggl.Foundation/Login/LoginManager.cs
daividssilverio: Hey, I looked into the `BaseApi` and I didn't find anything to make sure if resubscribing to the login really repeats the HTTP request.

What I'm doing in the tests is to check if the `ApiFactory`'s `CreateApiWith` is being called as much as I would expect. Is it enough to confirm that the HTTP request will be made again?

I had trouble writing theses tests and I'm not completely confident in them.

I'm also not sure if I should've tested this here. Maybe this DelayedConditionalRetry extension should have it's own tests instead.

daividssilverio: It is number of times to retry the observable when the UserIsMissingApiTokenException is thrown.
Named it NumberOfRetriesBeforeGivingUpOnFetchingAUserWithApiToken. Too long?

Toggl.Foundation/Login/LoginManager.cs
daividssilverio: I'm choosing will's idea, there is no special reason for not using a simple function for this. But I did name it.

Those numbers were suggest adhoc by Paul on the description of this issue.

Toggl.Foundation/Login/LoginManager.cs
daividssilverio: Moved it. I created a method for this and it looks a little odd to me now. What do you think?

Toggl.Foundation/Extensions/RxExtensions.cs
daividssilverio: done.

daividssilverio:
Updating on this: @simonrozsival It does retry the HTTP request. It is going all the way up to when ` var response = await sendRequest(request).ConfigureAwait(false);`(BasiApi.cs:95) for every try.

Toggl.Foundation/Extensions/RxExtensions.cs
```diff
@@ -0,0 +1,55 @@
+﻿using System;
+using System.Reactive.Concurrency;
+using System.Reactive.Linq;
+using Toggl.Ultrawave.Exceptions;
+
+namespace Toggl.Foundation.Extensions
+{
+    public static class RxExtensions
+    {
+        public static IObservable<T> DelayedConditionalRetry<T>(
```
Odrakir: I think this method does too much for an extension. Ideally you would use the `retryWhen` operator for this, but .Net doesn't have it yet (although there is a [pull request](https://github.com/dotnet/reactive/pull/486) pending).

I say we add our own implementation (basically copy other's) and use it to create this operator, which will be much simpler. 

Do you mind if I do a new PR you can use to clean this one a bit?

daividssilverio: Not at all! Less code, better; prettier, even better.

Toggl.Multivac/Extensions/ObservableExtensions.cs
```diff
@@ -69,5 +70,28 @@ public static IObservable<T> DelayIf<T>(this IObservable<T> observable, Predicat
...
         }
+
+        public static IObservable<T> ConditionalRetryWithBackoffStrategy<T>(
+            this IObservable<T> source,
+            int maxRetries,
+            Func<int, TimeSpan> backOffStrategy,
+            Func<Exception, bool> shouldRetryOn,
+            IScheduler scheduler)
+        {
+            return source.RetryWhen(v =>
```
Odrakir: I kind of like better the version using the override of `SelectMany` which includes a 0 based index. You have to take the 0 into account, but it's cleaner and more functional. Not a bid deal though, your call.

daividssilverio: How does it look?

simonrozsival:
Is this still `on-hold`, @daividssilverio?

Toggl.Foundation/Login/LoginManager.cs
```diff
@@ -148,5 +161,20 @@ private IObservable<IUser> signUp(Email email, Password password, bool termsAcce
...
+            return observable.ConditionalRetryWithBackoffStrategy(
+                NumberOfRetriesBeforeGivingUpOnFetchingAUserWithApiToken,
+                backOffStrategyForDelayedRetry(),
+                exception => exception is UserIsMissingApiTokenException,
+                scheduler);
+        }
+
+        private Func<int, TimeSpan> backOffStrategyForDelayedRetry()
+        {
+            return attempt => TimeSpan.FromSeconds(attempt * 8 - 6);
```
simonrozsival: looks like magic - why times 8 and minus 6? What is it? 2, 10, 18, 26, ... Maybe we could have an explicit `List<TimeSpan>` which would specify the delays + also implicitly hold the number of retries?

Toggl.Multivac/Extensions/ObservableExtensions.cs
```diff
@@ -69,5 +70,27 @@ public static IObservable<T> DelayIf<T>(this IObservable<T> observable, Predicat
...
```
simonrozsival: I don't like the name `v` - what does it stand for?

Toggl.Multivac/Extensions/ObservableExtensions.cs
```diff
@@ -69,5 +70,27 @@ public static IObservable<T> DelayIf<T>(this IObservable<T> observable, Predicat
...
+            {
+                return v.SelectMany((error, retryCount) =>
+                {
+                    var currentTry = retryCount + 1;
+                    if (!shouldRetryOn(error) || currentTry > maxRetries)
+                    {
+                        throw error;
+                    }
+
+                    return Observable.Return(1).Delay(backOffStrategy(currentTry), scheduler);
```
simonrozsival: Why return `1`? I don't think it has any meaning, am I right? If it is the case, it would be nicer to return `Unit.Default`, which is exactly for that purpose (the value doesn't matter, just call the `onNext` hander).

Toggl.Foundation/Login/LoginManager.cs
daividssilverio: Or maybe we should ditch the expression and just use an exponential delay 2^retry. I guess it doesn't have to be that formula. I made like this because of the description on #1701 and one old suggestion because I was using an If 1st retry = 2 seconds, if 2nd retry = 10 seconds.

simonrozsival: Yeah it doesn't really matter. But there should be some limit then for the case when somebody changes the max retries to 20 and doesn't realize that it might get stuck "forever" - max should be like 1 minute.

Toggl.Foundation/Login/LoginManager.cs
```diff
@@ -20,28 +24,33 @@ public sealed class LoginManager : ILoginManager
         private readonly IApplicationShortcutCreator shortcutCreator;
         private readonly IAccessRestrictionStorage accessRestrictionStorage;
         private readonly Func<ITogglApi, ITogglDataSource> createDataSource;
+        private readonly IScheduler scheduler;
+        private const int NumberOfRetriesBeforeGivingUpOnFetchingAUserWithApiToken = 2;
```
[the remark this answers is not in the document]
daividssilverio: 💄 👌

Toggl.Foundation/Login/LoginManager.cs
```diff
@@ -20,28 +24,33 @@ public sealed class LoginManager : ILoginManager
...
         {
             Ensure.Argument.IsNotNull(database, nameof(database));
             Ensure.Argument.IsNotNull(apiFactory, nameof(apiFactory));
             Ensure.Argument.IsNotNull(accessRestrictionStorage, nameof(accessRestrictionStorage));
             Ensure.Argument.IsNotNull(googleService, nameof(googleService));
             Ensure.Argument.IsNotNull(shortcutCreator, nameof(shortcutCreator));
             Ensure.Argument.IsNotNull(createDataSource, nameof(createDataSource));
-
+            Ensure.Argument.IsNotNull(scheduler, nameof(scheduler));
             this.database = database;
```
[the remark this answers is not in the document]
daividssilverio: 👌

Toggl.Foundation/Login/LoginManager.cs
daividssilverio: I've added a "safety cap" to the conditional retry operator

Toggl.Multivac/Extensions/ObservableExtensions.cs
daividssilverio: I stands for value, but I've changed for errorSignal for clarity.

Toggl.Multivac/Extensions/ObservableExtensions.cs
daividssilverio: Yeah, it looks nicer. And you are right, no meaning.

Toggl.Multivac/Extensions/ObservableExtensions.cs
```diff
@@ -78,19 +79,26 @@ public static IObservable<T> DelayIf<T>(this IObservable<T> observable, Predicat
...
+                    var delay = backOffStrategyDelayCap(backOffStrategy(currentTry));
+                    return Observable.Return(Unit.Default).Delay(delay, scheduler);
                 });
             });
         }
+
+        private static TimeSpan backOffStrategyDelayCap(TimeSpan delay)
+        {
+            var oneMinute = TimeSpan.FromMinutes(1);
+            return delay >= oneMinute ? oneMinute : delay;
```
simonrozsival: I would put the cap directly to the `backOffStrategyForDelayedRetry` - what if somebody later has a use case for this operator where he/she wants > 1 minute?

daividssilverio: Makes sense! I've moved the safety cap to the LoginManager.

Toggl.Foundation/Login/LoginManager.cs
```diff
@@ -81,13 +94,15 @@ public IObservable<ITogglDataSource> SignUp(Email email, Password password, bool
...
                 throw new ArgumentException($"A valid {nameof(password)} must be provided when trying to signup");
 
-            return database
-                    .Clear()
-                    .SelectMany(_ => signUp(email, password, termsAccepted, countryId))
-                    .Select(User.Clean)
-                    .SelectMany(database.User.Create)
-                    .Select(dataSourceFromUser)
-                    .Do(shortcutCreator.OnLogin);
+            return retryWhenUserIsMissingApiTokenException(database
```
amulware: Inconsistent indentation compared to the Login method.

Toggl.Foundation/Login/LoginManager.cs
```diff
@@ -148,5 +163,22 @@ private IObservable<IUser> signUp(Email email, Password password, bool termsAcce
                 .User
                 .SignUp(email, password, termsAccepted, countryId);
         }
+
+        private IObservable<ITogglDataSource> retryWhenUserIsMissingApiTokenException(
```
amulware: Consider renaming to:
`retryWhenUserIsMissingApiToken`
(the exception is an implementation detail)
or if you must mention the exception:
`retryOnUserIsMissingApiTokenException`

Toggl.Foundation/Login/LoginManager.cs
```diff
@@ -81,13 +94,15 @@ public IObservable<ITogglDataSource> SignUp(Email email, Password password, bool
...
-            return database
-                    .Clear()
-                    .SelectMany(_ => signUp(email, password, termsAccepted, countryId))
-                    .Select(User.Clean)
-                    .SelectMany(database.User.Create)
-                    .Select(dataSourceFromUser)
-                    .Do(shortcutCreator.OnLogin);
+            return retryWhenUserIsMissingApiTokenException(database
+                .Clear()
+                .SelectMany(_ => signUp(email, password, termsAccepted, countryId))
```
amulware: Will this sign up multiple times when the first fails to return an api token? That is not good. It should try simply fetching the user (logging in) on the following attempts.

daividssilverio:
@amulware Regarding the tests, the test `RetriesAfterAWhileWhenTheApiThrowsUserIsMissingApiTokenException`-`[Theory][InlineData(1 second, 1 api call)]` covers your concern about the first request being sent without delays. I'll add another case to cover the expected first call without delays using [InlineData(0 seconds, 1 api call)].

I'll address your other comments soon.

daividssilverio:
🤔 @amulware  Is doing executing the full Login flow when receiving UserIsMissingApiTokenException on SignUp?
What I mean is:
Try to sign up, get UserIsMissingApiTokenException, then just execute with the Login method from the `LoginManager` (after 2 seconds, maybe), and it would behave just like logging in, trying again up to 3 times when receiving `UserIsMissingApiTokenException`.

It would run like this: 
SignUp -> Fail -> wait 2 secs -> 
Login -> Fail -> wait 2 secs ->
Login -> Fail -> wait 4 secs ->
Login -> Fail -> 💥

amulware:
About Google Login/Signup: I'm actually thinking: Can we even repeat those? Or are the magic tokens only valid once?

If we can't repeat them, we obviously shouldn't try, and in that case, there might be nothing we can do if the API token is missing, leaving people who signup through google possibly quite confused (unless we communicate to them to please try LOGGING IN after). :/

We definitely need to track if this happens: #2400

daividssilverio:
The tokens given to us by the google apis are usually valid for a short period of time, so I guess we are safe trying again, but I guess it's better to track it to make sure.

Toggl.Foundation.Tests/Login/LoginManagerTests.cs
```diff
@@ -164,6 +178,19 @@ await LoginManager
 
                 ApplicationShortcutCreator.Received().OnLogin(Arg.Any<ITogglDataSource>());
             }
+
+            [Fact, LogIfTooSlow]
+            public void DoesNotRetryWhenTheApiThrowsSomethingOtherThanUserIsMissingApiTokenException()
+            {
+                var serverErrorException = Substitute.For<ServerErrorException>(Substitute.For<IRequest>(), Substitute.For<IResponse>(), "Some Exception");
```
amulware: Huh. This is new syntax for me. Interesting.

Toggl.Foundation.Tests/Login/LoginManagerTests.cs
```diff
@@ -434,5 +474,269 @@ await LoginManager
...
+                TestScheduler.Start();
+
+                LoginManager.Login(Email, Password).Subscribe(observer);
+                TestScheduler.AdvanceBy(TimeSpan.FromSeconds(seconds).Ticks);
+
+                Api.User.Received(apiCalls).Get();
+            }
+
+            [Fact, LogIfTooSlow]
+            public void ToLoginTwoTimesWhenReceivingUserIsMissingApiTokenExceptionAndThenThrowIt()
```
amulware: Doesn't actually check the 'two times' part.

Toggl.Foundation.Tests/Login/LoginManagerTests.cs
```diff
@@ -434,5 +474,269 @@ await LoginManager
...
+
+                observer.Messages.Single().Value.Exception.Should().BeOfType<UserIsMissingApiTokenException>();
+            }
+
+            [Fact, LogIfTooSlow]
+            public void AndStopRetryingAfterASuccessFullLoginApiCall()
+            {
+                var observer = TestScheduler.CreateObserver<ITogglDataSource>();
+                var userIsMissingApiTokenException = new UserIsMissingApiTokenException(Substitute.For<IRequest>(), Substitute.For<IResponse>());
+                Api.User.Get().Returns(Observable.Throw<IUser>(userIsMissingApiTokenException), Observable.Return(User));
```
amulware: Also new syntax to me, I'm learning a lot. :D

Toggl.Foundation.Tests/Login/LoginManagerTests.cs
```diff
@@ -434,5 +474,269 @@ await LoginManager
...
+                TestScheduler.Start();
+
+                LoginManager.LoginWithGoogle().Subscribe(observer);
+                TestScheduler.AdvanceBy(TimeSpan.FromSeconds(seconds).Ticks);
+
+                Api.User.Received(apiCalls).GetWithGoogle();
+            }
+
+            [Fact, LogIfTooSlow]
+            public void ToLoginTwoTimesWhenReceivingUserIsMissingApiTokenExceptionAndThenThrowIt()
```
amulware: 'two times'

Toggl.Foundation.Tests/Login/LoginManagerTests.cs
amulware: I'd change the name.

Toggl.Foundation/Login/LoginManager.cs
```diff
@@ -53,26 +64,36 @@ public IObservable<ITogglDataSource> Login(Email email, Password password)
...
-                .SelectMany(api => api.User.GetWithGoogle())
-                .Select(User.Clean)
-                .SelectMany(database.User.Create)
-                .Select(dataSourceFromUser)
-                .Do(shortcutCreator.OnLogin);
+                .SelectMany(loginWithGoogle);
+
+        private IObservable<ITogglDataSource> loginWithGoogle(string googleToken)
+        {
+            return Observable.Return(Credentials.WithGoogleToken(googleToken))
```
amulware: Why not reuse the credentials like in the login method? That seems clearer/simpler to me.

Toggl.Foundation/Login/LoginManager.cs
```diff
@@ -87,18 +108,25 @@ public IObservable<ITogglDataSource> SignUp(Email email, Password password, bool
...
         public IObservable<ITogglDataSource> SignUpWithGoogle()
             => database
                 .Clear()
                 .SelectMany(_ => googleService.GetAuthToken())
-                .SelectMany(apiFactory.CreateApiWith(Credentials.None).User.SignUpWithGoogle)
-                .Select(User.Clean)
-                .SelectMany(database.User.Create)
-                .Select(dataSourceFromUser)
-                .Do(shortcutCreator.OnLogin);
+                .SelectMany(googleToken => Observable.Return(googleToken)
```
amulware: Maybe stick this into another method to be similar to the google login above? consistency and all that.

Toggl.Foundation/Login/LoginManager.cs
daividssilverio: It does look nicer too.

Toggl.Foundation/Login/LoginManager.cs
```diff
@@ -53,26 +64,37 @@ public IObservable<ITogglDataSource> Login(Email email, Password password)
...
-                .SelectMany(database.User.Create)
-                .Select(dataSourceFromUser)
-                .Do(shortcutCreator.OnLogin);
+                .SelectMany(loginWithGoogle);
+
+        private IObservable<ITogglDataSource> loginWithGoogle(string googleToken)
+        {
+            var credentials = Credentials.WithGoogleToken(googleToken);
+
+            return retryWhenUserIsMissingApiToken(
```
Odrakir: I feel this breaks the nice rx operator syntax, in which every line is a command. I would add the `ConditionalRetryWithBackoffStrategy` right here (as the last command), I know it looks cleaner with just one line, but indirection is also harder to read.

Another option, if possible, create a local Observable extension for it. But I don't feel it's necessary.

Odrakir: If this has already been discussed, you can ignore me.

daividssilverio: This was requested here: https://github.com/toggl/mobileapp/pull/1796#discussion_r177015931 . But the local observable seems interesting as well, but we can refactor that later, right?

Toggl.Foundation/Login/LoginManager.cs
```diff
@@ -87,18 +109,31 @@ public IObservable<ITogglDataSource> SignUp(Email email, Password password, bool
...
+            return Observable
+                .Return(googleToken)
                 .SelectMany(apiFactory.CreateApiWith(Credentials.None).User.SignUpWithGoogle)
                 .Select(User.Clean)
                 .SelectMany(database.User.Create)
                 .Select(dataSourceFromUser)
-                .Do(shortcutCreator.OnLogin);
+                .Do(shortcutCreator.OnLogin)
+                .Catch<ITogglDataSource, UserIsMissingApiTokenException>(
+                    _ => loginWithGoogle(googleToken).DelaySubscription(TimeSpan.FromSeconds(2), scheduler)
```
simonrozsival: This deserves its own method + the two seconds would deserve a constant.

Toggl.Foundation/Login/LoginManager.cs
```diff
@@ -87,18 +109,31 @@ public IObservable<ITogglDataSource> SignUp(Email email, Password password, bool
                     .Select(User.Clean)
                     .SelectMany(database.User.Create)
                     .Select(dataSourceFromUser)
-                    .Do(shortcutCreator.OnLogin);
+                    .Do(shortcutCreator.OnLogin)
+                    .Catch<ITogglDataSource, UserIsMissingApiTokenException>(
+                        _ => Login(email, password).DelaySubscription(TimeSpan.FromSeconds(2), scheduler)
```
simonrozsival: This also deserves a separate method.

Toggl.Foundation/Login/LoginManager.cs
simonrozsival: Ricardo is right - it could easily be a private extension method (just the scheduler needs to be passed as an argument from the instance).

PR #1799, opened 2018-03-23
Reports legend

### LINKED ISSUE #1622: Reports: Pie chart

willsb:
Closes #1622 

I noticed that we are not doing *any* kind of grouping for the segments < 10%. I will open a different issue to fix that for both iOS and Android.

Squash whenever


Toggl.Giskard/Views/ReportsRecyclerViewHolder.cs
```diff
@@ -0,0 +1,49 @@
...
+using MvvmCross.Droid.Support.V7.RecyclerView;
+using MvvmCross.Platform;
+using MvvmCross.Platform.Droid;
+using Toggl.Giskard.Extensions;
+
+namespace Toggl.Giskard.Views
+{
+    public sealed class ReportsRecyclerViewHolder : MvxRecyclerViewHolder
+    {
+        private static readonly int defaultCardSize;
```
kornelijepetak: Where is this used?

Toggl.Giskard/Views/ReportsRecyclerViewHolder.cs
```diff
@@ -0,0 +1,49 @@
...
+            : base(handle, ownership)
+        {
+        }
+
+        public void RecalculateSize()
+        {
+            var height = IsLastItem ? lastContainerHeight : normalContainerHeight;
+            var layoutParameters = ItemView.LayoutParameters;
+            layoutParameters.Height = height;
+            ItemView.LayoutParameters = layoutParameters;
```
kornelijepetak: Can you please remind me when does this need to be in `Post`? Is this not in `Post` because `RecalculateSize` is called only from `OnBindViewHolder` which is always executed on UI thread?

Toggl.Giskard/Startup/TogglBindingBuilder.cs
```diff
@@ -0,0 +1,14 @@
+﻿using Toggl.Foundation.MvvmCross.Combiners;
+using MvvmCross.Binding.Droid;
+
+namespace Toggl.Giskard
+{
+    public sealed class TogglBindingBuilder : MvxAndroidBindingBuilder
+    {
+        protected override void FillValueCombiners(MvvmCross.Binding.Combiners.IMvxValueCombinerRegistry registry)
+        {
+            registry.AddOrOverwrite("Duration", new DurationValueCombiner());
```
kornelijepetak: Can you explain in a few words why is this hacking of the binding process needed? Which part could not have been done without the custom binding builder?

willsb: This is no hack; this is the correct way of registering combiners

Toggl.Giskard/Views/ReportsRecyclerViewHolder.cs
willsb: :eyes:

Toggl.Giskard/Views/ReportsRecyclerViewHolder.cs
willsb: Precisely. Post == please execute this, in the UI thread, the next time the views are being redrawn

PR #1801, opened 2018-03-26
Add fade animation to reports

### LINKED ISSUE #1056: Reports UI: loading state

arturs105:
Closes #1056


arturs105:
1. The loading card is a way to implement the transition with less code. Otherwise i would need to create several new bindings and bind to the existing card.
2. Will fix the empty state thing.

Issue #1616, opened 2018-03-02
Client name colour


pe0ny:
Make sure that Client name text colour is #5E5B5B where we display it next to the project name:
- [x] Timer page
- [x] Start view
- [x] Edit view
- [x] Projects view
- [x] Reports

PR #1802, opened 2018-03-26
Reports client name

### LINKED ISSUE #1598: Reports UI updates
- [x] Overiview section: display data on a card
- [x] Overiview section: Total and Billable titles - font size 12pt; letter spacing -0,2; color #5E5B5B
- [x] Overiview section: Total and Billable amounts - letter spacing -0,2
- [x] Pie chart list: font size to 15pt; letter spacing -0,2; client name color #5E5B5B

### Reports
toggl - ios - reports 1x

Zeplin: https://zpl.io/bll9RdG

### LINKED ISSUE #1599: Reports: pie chart list doesn't display client names
In addition to a project name we should also display a client name (if any) in the pie chart list view.

simonrozsival:
Closes #1599, closes #1598, ref #1616

The client name is loaded only from the local database. If the client wasn't synced, it won't be shown. If we want to always load all clients, we will have to implement the reports API for searching clients by IDs.

Squash whenever


willsb:
No me gusta. We are using the weird api search function thingy for the projects. Why implement a half baked version of it for the clients? 🤔

amulware:
Since clients cannot be archived, only deleted, which is advertised as an irreversible action on the web, I think this is fully sufficient.

simonrozsival:
@amulware but will we always get all clients when we don't specify any `since` parameter? Won't we get only a list of "recent" clients instead?

amulware:
No, only TEs behave funny that way. Project/client/tag lists are complete (assuming the sync doesn't fail), minus archived projects.

amulware:
`since` makes sure that we don't repull the entire list every time but only those that have changed since then.

Toggl.Foundation/Reports/ReportsProvider.cs
```diff
@@ -56,16 +58,24 @@ private IObservable<ProjectSummaryReport> summaryReportFromResponse(IProjectsSum
...
         {
             var percentage = totalSeconds == 0 ? 0 : (summary.TrackedSeconds / (float)totalSeconds) * 100;
             var billableSeconds = summary.BillableSeconds ?? 0;
             var project = projectsInReport.FirstOrDefault(p => p.Id == summary.ProjectId);
+            var client = project != null && project.ClientId.HasValue
+                ? clientsRepository.GetAll()
+                    .SelectMany(clients => clients)
+                    .Where(c => c.Id == project.ClientId.Value)
+                    .FirstOrDefaultAsync()
+                    .Wait()
```
arturs105: I'm not thrilled about the `Wait()`. Isn't there a way to do this without it?

simonrozsival: I will try to get rid of it.

Toggl.Foundation/Reports/ChartSegment.cs
```diff
@@ -15,19 +15,25 @@ public struct ChartSegment
 
         public string ProjectName { get; }
 
+        public string ClientName { get; }
+
+        public bool HasClient => ClientName != null && ClientName.Length > 0;
```
arturs105: `string.IsNullOrEmpty()`?

Toggl.Foundation/Reports/ReportsProvider.cs
```diff
@@ -50,22 +52,37 @@ private IObservable<ProjectSummaryReport> summaryReportFromResponse(IProjectsSum
...
+            => project != null && project.ClientId.HasValue
+                ? clientsRepository.GetAll()
+                    .SelectMany(clients => clients)
+                    .Where(c => c.Id == project.ClientId.Value)
+                    .FirstOrDefaultAsync()
+                : Observable.Return<IClient>(null);
+
+        private ChartSegment chartFromSummary(IProjectSummary summary, 
+                                                     IProject project,
+                                                     IClient client,
```
arturs105: What is this alignment?

Issue #1623, opened 2018-03-02
Reports: Calendar


pe0ny:
- [ ] By default show "This week".
- [ ] Close the calendar once date range has been selected (after the second tap or selecting a range from shortcuts).
- [ ] User can also select one day from the calendar and close the calendar manually - then we'll display data of the selected day only.
- [ ] User can move back in the calendar for one year, not more.
- [ ] When a user selects one of the shortcuts "this week", "This month" etc in the middle of the week/month/year, then show a full week. For example: On Wednesday it will show range from Monday to Sunday OR when the user has set in settings that weeks start with Sunday, then from Sunday to Saturday.

**Interactions:**
- A user can tap on the nav bar "This month" to show/hide the calendar.
- Calendar expands the nav bar and content (summary section, bar chart etc..) reveal the calendar.
- When the user starts scrolling down (swipes up), then content covers the calendar and in the end closes it.
- A user can swipe left and right to switch between months.
- To pick a date range: first tap - start date (background behind the date turns blue), second tap - end date (background between the dates turns blue).
- A user can scroll left-right to browse the shortcuts.

**Shortcuts:** Today, Yesterday, This week, Last week, This month, Last month, This year

### Calendar

![reports - calendar](https://user-images.githubusercontent.com/5064445/36898211-887f89ee-1e22-11e8-8cdf-676ec1034617.png)

Zeplin: https://zpl.io/bzYRl3A

PR #1804, opened 2018-03-26
Calendar scroll behaviour

willsb:
Ref #1623 

This adds only the scrolling behaviour. The fragment itself will come in a different PR. Please do a functional test before approving.

I’m using a custom LinearLayout because this is the only way of intercepting the scrollview’s touch events. Making the calendar part of the scrollview would be much harder.


kornelijepetak:
When I open the reports, this screen appears (in this state):

image

Note that the scroll is a bit off, summary card is not fully visible. When I try to drag down, this happens:

image

The dark background goes down as I pull.

kornelijepetak:
I should read descriptions more carefully. Anyways, I have noticed three _issues_:

1. When reports is first opened (but also every time it's opened), the summary card is partly visible, as if somebody already scrolled a bit down. When you drag on the summary/piechart area, the calendar view is expanded instead of scrolling. On another dragging, the   calendar rolls back up. Afterwards, scrolling works as normal until reports screen is reopened.

2. When the reports screen is first opened, and you click the `This Week` the calendar will not be opened and the scrolling view will snap to what would be `scroll to top`, with summary card visible correctly (similar to what happens in 1.) After that, clicking on the `this week` will toggle the calendar view's expansion correctly.

3. When expanding calendar view, the view below is moved down along with it. When collapsing it, the view below immediately (no animation) snaps to the top of the screen, while the collapse of calendar view is animated.

willsb:
The view still loads and shows the last item. In order to get the ball rolling, I say we fix that in a different PR. Please check the other points, for those should have been addressed.

PR #1816, opened 2018-03-28
Fix invalid dialog buttons' texts

kornelijepetak:
Somebody was naughty.
Or drunk.

Squash whenever.


kornelijepetak:
Just to note, iOS's `DialogService` implementation is fine.

PR #1822, opened 2018-03-29
Make sure all separators are correct

### LINKED ISSUE #1615: Make sure all separator lines are #E8E8E8
Make sure all separator lines' colour is #E8E8E8
Places where we use separator lines:
- [x] Timer page (Log)
- [x] Start view (Autocomplete items)
- [x] Edit view
- [x] Projects view
- [x] Tags view
- [x] Clients view
- [x] Workspaces
- [x] New project view
- [x] Reports (List view)
- [x] Settings

arturs105:
Closes #1615 

All of the separators already had the correct color, i just renamed the separator views in xib files. Also fixed the double top separator in the settings view.

However there is one place, where separators might be a bit different - in the `SelectTagsViewController`. There we are using separators provided by `UITableView`, but everywhere else we are using our own. To me it looks like the default separators are a bit lighter/thinner than ours, but i'd like another pair of eyes to confirm this, **so please*check (🇨🇿) if those separators look different to you.** If they do, i will switch the default ones for our separators.

Squash whenever


simonrozsival:
Those separators are definitely different. Please change them.

PR #1819, opened 2018-03-28
Reports calendar

willsb:
Ref #1623

This adds the calendar itself.

I’m not quite happy with the scrolling behaviour right now. It’s a bit clumsy. After using it once or twice you kinda get over it, but it definitely can be improve. Since we are still adding yet another scroll view (for the shortcuts) I say we ignore it for now until we settle with all the conflicting scrolls and fix it for good.

Squash whenever.

PR #1823, opened 2018-03-29
Add selection splash of color to calendar cells

willsb:
Ref #1623

~Blocked by #1819~

Squash whenever


Toggl.Giskard/Views/CalendarDayView.cs
```diff
@@ -0,0 +1,126 @@
...
+        {
+            var width = Width;
+            var height = Height - verticalPadding * 2;
+
+            if (IsSelected)
+            {
+                var roundRect = new RectF(0, verticalPadding, width, height + verticalPadding);
+                canvas.DrawRoundRect(roundRect, cornerRadius, cornerRadius, selectedPaint);
+
+                var squareRectLeft = RoundLeft ? cornerRadius : 0;
```
kornelijepetak: Oh, dear... It took me two minutes to realize what you are doing here. And then I thought, seriously droid developers?! You couldn't have done the corner radius for all corners separately!?

Not sure if there is any real need for that, but it would be cool to have an extension method that takes 4 radii and draws appropriately, albeit by drawing two shapes.

Issue #1764, opened 2018-03-20
Time/date/duration popup


pe0ny:
- [ ] Open this popup when user taps
a) Clock icon or Duration in the Start view
b) Start/stop/duration in the Edit view
- [ ] User can switch between three options: start time/date, stop time/date and duration
- [ ] Swipe gestures are used for navigating between tabs. Also, user can tap the titles to navigate between tabs.
- [ ] Tabs show currently selected time or duration, eg 12:30 PM or 0:30:15. The selected tab shows instead name of the tab/content, eg "Start", "End" or "Duration".
- [ ] User can also edit the date in the same view by tapping on the date in the blue area. Clock content will be replaced with Date picker.

### Time/date/duration popup

![edit - time](https://user-images.githubusercontent.com/5064445/37646380-3404bdf8-2c32-11e8-9b8d-490b32bbee3a.png)

Zeplin: https://zpl.io/aBOz3Mq

### Date

![edit - date](https://user-images.githubusercontent.com/5064445/37646648-dc3eba96-2c32-11e8-94f8-f55136663f81.png)

Zeplin: https://zpl.io/aRN4k5z

willsb:
Has this been superseded by #2040?
If so, shouldn't this be closed?

kornelijepetak:
Not superseded, but split. Indeed, it should.

PR #1826, opened 2018-03-30
Remove date field

### LINKED ISSUE #1542: Edit view UI updates
- [x] Move description to the top on the toolbar like we have it in the start view.
- [x] When a user is editing the description, the action changes from "Save" to "Done".
- Selecting "Done" will bring them back into "Edit everything" mode where description field is not active (user is not currently editing the description).
- Selecting "Save" in "Edit everything" mode saves the changes and closes the edit view.
- [x] Tapping 'X' will discard the changes and closes the edit view. 'X' behaves the same way when the description field is not active and when it's active (when a user is currently typing the description). *Check confirmation #1543*
- [x] Remove the delete icon from the toolbar and move it at the bottom of the edit view list.
- [x] Tapping "Delete this entry" opens confirmation #1269 and tapping "Delete" once more, deletes this entry.
- [x] Show "Duration" label in the duration field.

### Edit view
edit

Zeplin: https://zpl.io/bPrLJQD

kornelijepetak:
Fixes the following points from #1542:

- [x] Remove date field
- [x] Show shortened date in the start/end time fields and use profile settings format. *For example: When user has set in the settings MM/DD/YYYY, then show only MM/DD*

Closes #1542 (since those are the last points)


kornelijepetak:
I have added a WIP label because the revamping of the start/stop/duration has started, so if this is merged, there is no way to edit some of the fields.

willsb:
One hint: I'd use a `Format` Value combiner in the bindings, but I'll let you decide if you wanna change that or not

kornelijepetak:
I have put this PR `on-hold`.

⚠️Date field must not be removed until #1764 is implemented, otherwise there will be no way to set the date of a TE.

willsb:
To the owner: Please take care not to reintroduce `Resource.designer.cs` when merging this.

willsb:
#1764 is very much implemented, can we move on with this now? :)

kornelijepetak:
@willsb This is no longer stale. I have merged develop and modified it as needed. Please rereview and we're done.

Toggl.Foundation.MvvmCross/ViewModels/EditTimeEntryViewModel.cs
```diff
@@ -119,6 +119,8 @@ public TimeSpan DisplayedDuration
         [DependsOn(nameof(StopTime))]
         public bool IsTimeEntryRunning => !StopTime.HasValue;
 
+        public DateTimeOffset StopTimeOrCurrent => StopTime ?? default(DateTimeOffset);
```
willsb: This is weird, why not use the time service, given the name of this property?

kornelijepetak: oh, I just forgot to change this. I was trying something out and somehow it slipped :/

PR #1835, opened 2018-04-03
Fix google api tests

amulware:
Backend is returning the more specific 403, instead of 400 now, so that was changed. Also, there was a test which had a type and wasn't enabled, so that's also fixed.

Api tests ran at https://www.bitrise.io/build/cd58006acbadfbf4 after the second commit, where they still fail for the re-enabled test (fixed in commit 3), and task tests, because of a failing subscription endpoint, which is up to backend to fix.


amulware:
🦑

Issue #1743, opened 2018-03-16
Start view UI


pe0ny:
We decided to put #1143 on hold until we decide what's the best solution for the bottom nav bar and see how iOS solution is going to work out. We finally made up our minds and decided to go with iOS-way.

- [x] Remove FAB and use 'Save' label at the top bar instead.
- [x] Use X instead of the back arrow.
- [x] Use bottom nav bar for:
- Duration
- Clock icon which opens similar date/time popup that we use in iOS *(Issue for that coming soon)*
- Projects
- Tags
- Billable flag
- [x] Make tapping area for icons 40x56dp.
- [x] And for the duration tapping area's height 56dp and give it 8pt padding from left and right.

### Start view
![new time entry](https://user-images.githubusercontent.com/5064445/37523099-ed02d58e-28f3-11e8-8bbc-fb108ccc363a.png)

Zeplin: https://zpl.io/bzy1R3z

### Tapping areas
<img width="361" alt="screen shot 2018-03-19 at 10 50 37" src="https://user-images.githubusercontent.com/5064445/37585940-7a22b760-2b63-11e8-9ac4-78d3d5f0e974.png">

PR #1839, opened 2018-04-03
Token colors

willsb:
Ref #1743 

This makes the tokens visible when displayed in the blue background.

I need @kornelijepetak to confirm if this breaks any of his usages of the tokens. If it does, please list what’s wrong so I can fix it.

Otherwise, squash whenever


kornelijepetak:
Indeed it does break something. ⚠️  `EditActivity.axml` uses the `TextViewTagListTargetBinding` (MvxBind of `Tags`).

Since the tags displayed are used on the white background, setting color to white indeed makes the tags invisible.

Somehow, the color should be parameterized...

kornelijepetak:
@willsb But I believe the best way is for my use case to inherit from `TokenSpan` again and set a different color. Something like `EditViewTagsTokenSpan`. If you want to take that route, please say so and create an issue that edit view's tags are invisible  (and assign me to it). I will then approve this PR cause otherwise it's alright.

PR #1848, opened 2018-04-04
Fix the :bug: - once and for all

### LINKED ISSUE #1523: Various GroupedCollectionTableViewSource crashes

### LINKED ISSUE #1784: App crashes during animations after pull sync
- [ ] scenario 1
    - create two TEs: X and Y on some date (doesn't matter if you create them on the webapp or in the mobile app)
    - sync the mobile app
    - change the date of both those TEs to a different date in the web app
    - sync the mobile app
    - :boom:
- [ ] scenario 2
    - create two TEs: U and V on 27. 3. (assuming you already don't have no other TEs on that date, otherwise pick some other suitable date, as in the previous case it doesn't matter if it is in the webapp or    the mobile app)
    - sync the mobile app
    - change the date of U to 28. 3. in the web app (also must be emtpy before this step)
    - change the date of V to 29. 3. in the web app (also must be empty before this step)
    - sync the mobile app
    - :boom:
- [ ] scenario 3
    - create two TEs: M and N on two different dates where no other TEs are yet
    - sync the mobile app
    - delete M and N in the web app
    - sync the mobile app
    - :boom:

simonrozsival:
Closes #1523, closes #1784

The main idea behind this PR is that the `GroupedCollectionTableViewSource` handles its own internal state (`ItemsSource`) and makes sure that the displayed sections and rows in the `UITableView` always match the state of the `ItemsSource`. This means that it is not possible to bind anything directly to the `ItemsSource` (as we normally do), but we must rather bind the collection from the VM to the `ObservableCollection` property.

Whenever the `ObservableCollection` changes, the `CollectionChanged` or `OnChildCollectionChanged` events are invoked. We know exactly how the collection was changed (add/delete/replace/...) and we can update both the internal collection and the `UITableView` state - it is important to do this in between `TableView.BeginUpdate()` and `TableView.EndUpdate()` so that we don't get anymore because of the number of rows in the `ItemsSource` and the `UITableView` is different (internal inconsistency).

This PR doesn't try to merge updates into batches, because that would make a lot more changes and it isn't necessary for this fix. We should still do that in the future - it will give use smoother animations. It can certainly wait for 1.5 or 1.6 though. Let's verify this fix and :shipit: ASAP!

Squash whenever

### DIFF
--- a/Toggl.Daneel/ViewSources/GroupedCollectionTableViewSource.cs
+++ b/Toggl.Daneel/ViewSources/GroupedCollectionTableViewSource.cs
@@ -145,31 +145,32 @@ namespace Toggl.Daneel.ViewSources
 
         protected void OnCollectionChanged(object sender, NotifyCollectionChangedEventArgs args)
         {
-            InvokeOnMainThread(() =>
-            {
-                if (!UseAnimations)
-                {
-                    cloneCollection();
-                    ReloadTableData();
-                    return;
-                }
-
-                animateSectionChangesIfPossible(args);
-            });
+            tryAnimateOnMainThread(() => animateSectionChangesIfPossible(args));
         }
 
         protected void OnChildCollectionChanged(object sender, ChildCollectionChangedEventArgs args)
+        {
+            tryAnimateOnMainThread(() => animateRowChangesIfPossible(args));
+        }
+
+        private void tryAnimateOnMainThread(Action animate)
         {
             InvokeOnMainThread(() =>
             {
                 if (!UseAnimations)
                 {
-                    cloneCollection();
-                    ReloadTableData();
+                    reloadTable();
                     return;
                 }
 
-                animateRowChangesIfPossible(args);
+                try
+                {
+                    animate();
+                }
+                catch
+                {
+                    reloadTable();
+                }
             });
         }
 
--- a/Toggl.Daneel/ViewSources/GroupedCollectionTableViewSource.cs
+++ b/Toggl.Daneel/ViewSources/GroupedCollectionTableViewSource.cs
@@ -24,7 +24,7 @@ namespace Toggl.Daneel.ViewSources
 
         private NestableObservableCollection<TCollection, TItem> observableCollection;
 
-        private List<TCollection> displayedGroupedItems = new List<TCollection>();
+        private readonly List<TCollection> displayedGroupedItems = new List<TCollection>();
 
         protected IReadOnlyCollection<TCollection> GroupedItems { get; }
 
--- a/Toggl.Daneel/ViewSources/GroupedCollectionTableViewSource.cs
+++ b/Toggl.Daneel/ViewSources/GroupedCollectionTableViewSource.cs
@@ -182,38 +182,33 @@ namespace Toggl.Daneel.ViewSources
                 switch (args.Action)
                 {
                     case NotifyCollectionChangedAction.Add:
-                        var indexToAdd = NSIndexSet.FromIndex(args.NewStartingIndex);
-
-                        var addedSection = new TCollection();
-                        addedSection.AddRange(observableCollection[args.NewStartingIndex]);
-
+                        var addedSection = CloneCollection(observableCollection[args.NewStartingIndex]);
                         displayedGroupedItems.Insert(args.NewStartingIndex, addedSection);
+
+                        var indexToAdd = NSIndexSet.FromIndex(args.NewStartingIndex);
                         TableView.InsertSections(indexToAdd, UITableViewRowAnimation.Automatic);
                         break;
 
                     case NotifyCollectionChangedAction.Remove:
-                        var indexToRemove = NSIndexSet.FromIndex(args.OldStartingIndex);
                         displayedGroupedItems.RemoveAt(args.OldStartingIndex);
+
+                        var indexToRemove = NSIndexSet.FromIndex(args.OldStartingIndex);
                         TableView.DeleteSections(indexToRemove, UITableViewRowAnimation.Automatic);
                         break;
 
                     case NotifyCollectionChangedAction.Move when args.NewItems.Count == 1 && args.OldItems.Count == 1:
+                        var movedSection = CloneCollection(observableCollection[args.NewStartingIndex]);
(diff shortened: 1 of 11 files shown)


Toggl.Daneel/ViewSources/MainTableViewSource.cs
```diff
@@ -20,10 +21,16 @@ public IMvxAsyncCommand<TimeEntryViewModel> ContinueTimeEntryCommand
...
+        {
+            get => timeEntriesLogViewSource.ObservableCollection;
+            set => timeEntriesLogViewSource.ObservableCollection = value;
+        }
+
         public override IEnumerable ItemsSource
         {
             get => timeEntriesLogViewSource.ItemsSource;
-            set => timeEntriesLogViewSource.ItemsSource = value;
+            set { throw new InvalidOperationException($"You must bind to the {nameof(ObservableCollection)} and not the {nameof(ItemsSource)}"); }
```
willsb: This makes me happy

Toggl.Daneel/ViewSources/GroupedCollectionTableViewSource.cs
```diff
@@ -163,93 +177,125 @@ private void animateSectionChangesIfPossible(NotifyCollectionChangedEventArgs ar
...
                         break;
                 }
+
+                TableView.EndUpdates();
+            }
+        }
+
+        private void cloneCollection()
+        {
+            internalCollection = new List<TCollection>();
```
amulware: Consider initialising with capacity.

Toggl.Daneel/ViewSources/MainTableViewSource.cs
amulware: Why? (curious)

Toggl.Daneel/ViewSources/GroupedCollectionTableViewSource.cs
```diff
@@ -22,18 +22,49 @@ public abstract class GroupedCollectionTableViewSource<TCollection, TItem> : Mvx
...
 
-        protected IEnumerable<MvxObservableCollection<TItem>> GroupedItems
-            => ItemsSource as IEnumerable<MvxObservableCollection<TItem>>;
+        private NestableObservableCollection<TCollection, TItem> observableCollection;
 
-        protected NestableObservableCollection<TCollection, TItem> AnimatableCollection
-            => ItemsSource as NestableObservableCollection<TCollection, TItem>;
+        private readonly IList<TCollection> internalCollection;
+
+        protected IList<TCollection> GroupedItems => internalCollection;
```
amulware: Is this actually exposed to anybody else? Or used at all?

simonrozsival: Yes, it is used 3 times in the 3 derived classes.

amulware: If it's only used for reading, I'd prefer exposing only a cached `internalCollection.AsReadonly()` (could be assigned in the constructor), or is there a reason not to?

Toggl.Daneel/ViewSources/GroupedCollectionTableViewSource.cs
```diff
@@ -22,18 +22,49 @@ public abstract class GroupedCollectionTableViewSource<TCollection, TItem> : Mvx
...
+            }
+        }
 
         protected GroupedCollectionTableViewSource(UITableView tableView, string cellIdentifier, string headerCellIdentifier)
             : base(tableView)
         {
             this.cellIdentifier = cellIdentifier;
             this.headerCellIdentifier = headerCellIdentifier;
 
+            internalCollection = new List<TCollection>();
```
amulware: Initialise field directly?

simonrozsival:
Inserts/deletions aren't `O(n^2)` - they are `O(m * (n + m))`, where `n` represents the number of items in the list and `m` the number of inserted/deleted rows. For small `m`s this converges to `n` and for large `m`s this converges to `m^2`.

We could theoretically achieve `O(n + m)` with a one-pass "merging" of the old data and the changes, but I am not sure if that would be a good idea.

We might use some different data structure - e.g. a heap to store the ordered items of the collection, then the time complexity would be `O(m * ln (n + m))`.

This might be a nice optimization in some of the next releases (1.6?). I think this might be good enough for now.

When I think about it, a much easier way of dealing with the performance might be to set a "limit"  of changes for which we will still animate it (e.g., 20) and for more than that we simply reload the table without any animation.

Toggl.Daneel/ViewSources/GroupedCollectionTableViewSource.cs
```diff
@@ -22,11 +22,40 @@ public abstract class GroupedCollectionTableViewSource<TCollection, TItem> : Mvx
         private readonly object animationLock = new object();
         private readonly List<IDisposable> disposables = new List<IDisposable>();
 
-        protected IEnumerable<MvxObservableCollection<TItem>> GroupedItems
-            => ItemsSource as IEnumerable<MvxObservableCollection<TItem>>;
+        private NestableObservableCollection<TCollection, TItem> observableCollection;
 
-        protected NestableObservableCollection<TCollection, TItem> AnimatableCollection
-            => ItemsSource as NestableObservableCollection<TCollection, TItem>;
+        private List<TCollection> displayedGroupedItems = new List<TCollection>();
```
amulware: This can still be `readonly` right?

amulware:
While we are geeking out about big-O, if we sort a list of updates, and then merge them into the list, it should be `O(n + m * log m)`, right? Anyways. For the future.

simonrozsival:
You are right, I assumed the list of changes was sorted - which it probably wouldn’t be.

Toggl.Daneel/ViewSources/MainTableViewSource.cs
willsb: Because that was a very big concern of mine: people binding to the wrong property

simonrozsival: This is still unfortunately only a silent error - MvvmCross catches it and puts it into the application output. But I hope that it is still better than nothing.

PR #1879, opened 2018-04-06
Add AppStore link for iOS

willsb:
Deprecation now works. Yay.

Squash whenever


Toggl.Daneel/Services/BrowserService.cs
```diff
@@ -9,7 +9,7 @@ public sealed class BrowserService : IBrowserService
     {
         public void OpenStore()
         {
-            OpenUrl("");
+            OpenUrl("https://itunes.apple.com/us/app/toggl/id1291898086?mt=8");
```
amulware: Should the `mt` parameter be included? It does not seem necessary. Also, should it always link to the US store?

simonrozsival: Will the `Media Type` of the app change? I don't think so, I think the app has to know how the AppStore links work. I would keep it like this.

Are we able to detect that we should take user to a different store than the US store? Won't Apple redirect him/her, if necessary?

kornelijepetak: As far as I know if `mt=8` should be included.
Without it, it may work only on iphones, but fail on ipads.
Also it has something to do with forwarding from iTunes to the app itself if it's included.
I think we should leave it as is.
About the `/us/` part, I think @amulware should have a final say.

PR #1880, opened 2018-04-06
Configure Giskard PlayStore builds

amulware:
Do NOT review per commit, unless you like to suffer.

🦑


amulware:
Test build on latest commit: https://www.bitrise.io/build/e91eda287c6d8ab3

Toggl.Giskard/Toggl.Giskard.csproj
```diff
@@ -36,10 +36,10 @@
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
     <AndroidUseSharedRuntime>false</AndroidUseSharedRuntime>
-    <AndroidCreatePackagePerAbi>false</AndroidCreatePackagePerAbi>
```
willsb: This should not have been removed

Toggl.Giskard/Toggl.Giskard.csproj
```diff
@@ -50,18 +50,18 @@
     <AndroidSigningKeyPass>{KEYSTORE_ALIAS_PASSWORD}</AndroidSigningKeyPass>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release.PlayStore|AnyCPU' ">
-    <Optimize>false</Optimize>
```
willsb: Why was the Optimize flag removed?

Toggl.Giskard/Toggl.Giskard.csproj
amulware: Why not? what do we gain from a bunch of different files, except more effort to upload them (and dangers of forgetting one of them)?

Toggl.Giskard/Toggl.Giskard.csproj
amulware: Strictly speaking, because it was different than the adhoc build. But do we need it? Do we not want to optimize the build? (Note that the value was 'false' before.)

Toggl.Giskard/Toggl.Giskard.csproj
willsb: Smaller, more optimized binaries.

willsb: And this flag was set to false on the Release config; It should be exactly like that for AdHoc

Toggl.Giskard/Toggl.Giskard.csproj
willsb: I don't know what this defaults to, so I have no idea if removing the false sets it to true. Hence why I ask

Toggl.Giskard/Toggl.Giskard.csproj
amulware: I think Google might split the APK for us with their new (re)signing thing. I'd prefer trying to see if that works for now, looking at the download size in the store. We can change this in the future, if we think it'll be better.

Toggl.Giskard/Toggl.Giskard.csproj
amulware: I've never seen it set to true specifically, so I assume that's the default.

PR #1886, opened 2018-04-09
Avoid an exception when no rows are displayed

simonrozsival:
Fixes https://appcenter.ms/orgs/Toggl/apps/Toggl-iOS/crashes/groups/35a4d22e64d58802312ffea98f1319c243096667?appVersion=1.4.3

How to reproduce the bug:
- go to the main view
- scroll so much that none of the rows is visible
- 💥 

This mainly affected iPhone SE, where you don't have to scroll so much.

Squash whenever


amulware:
This needs a functional test, but I can confirm that the code very much looks like it will fix about 20% of crashes in the current version.

PR #1925, opened 2018-04-18
Fix picker issue when editing end time

### LINKED ISSUE #1887: January 1st/December 31st displayed when editing time entry end time
>When editing a time entry, if you click on the stop time it brings up the circular diagram allowing you to edit the start and stop times by dragging them around the circle. If you again click on the stop time to edit it manually, it brings up the date and time editing interface, but it quickly sets the date to January 1 and sets the time to the start time, rather than leaving everything at the already existing stop time.

case/675120

When replicating the issue, I didn’t see the date being set to January 1st, but to December 31st, but in principle the date change was there. 

When testing, the issue seemed to be present only when attempting to edit the end time before having clicked on start time at all. When I had clicked on start time first, hitting the end time seemed to work properly as well.

juanlaube:
Closes #1887 

This fixes the glitch for sure, but I'm not 100% sure we're on the correct path here, since the problem happens only when tapping the end time and I would expect the same issue with start time too.

### DIFF
--- a/Toggl.Daneel/Binding/DatePickerDateTimeOffsetTargetBinding.cs
+++ b/Toggl.Daneel/Binding/DatePickerDateTimeOffsetTargetBinding.cs
@@ -20,12 +20,6 @@ namespace Toggl.Daneel.Binding
 
         protected override void SetValue(DateTimeOffset value)
         {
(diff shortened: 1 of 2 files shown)


amulware:
When **squashing** this PR, please make sure to include an appropriate emoji as per our guidelines.

Issue #1777, opened 2018-03-21
Track project and tag addition in TE creation view


amulware:
There are three different ways to open the project/tag list:
- typing @/#
- using the button right above the keyboard
- using the button that shows below the text field, that has to share space with the auto completion

Let's have two events, with a property that represents the three options above each.

I think it's fine to send when the user opens the project/tag list, we don't have to worry about whether they select stuff for now.

Odrakir:
Should we track this for both TE creation and TE editing? Should we differentiate those?

pe0ny:
We already have an issue for the edit view as well #1778 and yep, we should differentiate those.

amulware:
Yeah, please check the referenced issues. There is some overlap, and we should probably do this in some smart way. :D

Odrakir:
But you do want different events for edit/create and not the same event with different parameters, right?

amulware:
yeah, we want at least two different kinds of events. Maybe even more. I haven't thought much about what makes the most sense.

simonrozsival:
We would have to have specific commands for tapping the button and typing the symbols - then from these commands we would call some common logic. It would make the code a bit uglier, but it would do the job...

Odrakir:
Ok, this is almost done, but still, every way of showing project/tag suggestions end up adding "@" or "#" to the text field, so in order to not track text field input when indeed it was a button tap I end up having to set a property in the ViewModel only to track this.

It works but it's ugly and, in my opinion, it has too much logic for something like this. Is it good enough if every method tracks text change but also the buttons track each event separately?

amulware:
So four events, instead of two then?

One each when the character is added to the text, and one each that records just the non-keyboard button presses, which is a subset of the first thing?

If it makes that much of a difference, sure, this will do.

PR #1932, opened 2018-04-20
Track project and tag selector opening from edit view

### LINKED ISSUE #1778: Track user opening project/tag selection in edit view
Two events. Fired when the user opens the list, we don't care if they actually change the data at this point.

Include one property for both: `EntryIsRunning` that is true when the time entry is running.

Odrakir:
Closes #1778 

It creates custom events and methods in the analytics service. It might make sense to merge them with the ones needed for #1777, we'll see.


Toggl.Foundation.Tests/MvvmCross/ViewModels/EditTimeEntryViewModelTests.cs
```diff
@@ -55,30 +58,32 @@ protected void ConfigureEditedTimeEntry(DateTimeOffset now, bool isRunning = fal
             }
 
             protected override EditTimeEntryViewModel CreateViewModel()
-                => new EditTimeEntryViewModel(TimeService, DataSource, InteractorFactory, NavigationService, OnboardingStorage, DialogService);
+            => new EditTimeEntryViewModel(TimeService, DataSource, InteractorFactory, NavigationService, OnboardingStorage, DialogService, AnalyticsService);
```
amulware: The indentation was removed.

PR #1938, opened 2018-04-20
Fix updating the log when a TE is created

simonrozsival:
Closes #1937 

Squash whenever

PR #1939, opened 2018-04-20
Fix randomly failing unit tests

### LINKED ISSUE #1409: Fix flaky tests
Here is a list of tests which are flaky and fail just "sometimes" and often require the Bitrise test to be rebuilt:

- [ ] Toggl.Foundation.Tests.MvvmCross.ViewModels.CalendarQuickSelectShortcuts
.CalendarLastWeekQuickSelectShortcutTests.WhenBeginningOfWeekIsSunday.TheGetDateRangeReturnsExpectedDateRange

    > Error Message: MvvmCross.Platform.Exceptions.MvxIoCResolveException : Failed to resolve type MvvmCross.Core.Platform.IMvxSettings

- [ ] Toggl.Foundation.Tests.MvvmCross.ViewModels.EditTimeEntryViewModelTests
+TheSelectTagsCommand.SetsTheReturnedTags

    > FsCheck.Xunit.PropertyFailedException : 
    > [xUnit.net 00:00:22.7492850]       Falsifiable, after 4 tests (0 shrinks) (StdGen (919480337,296408108)):
    > [xUnit.net 00:00:22.7493900]       Original:
    > [xUnit.net 00:00:22.7494820]       NonEmptyArray [|NonNegativeInt 1; NonNegativeInt 0|]
    > [xUnit.net 00:00:22.7495690]       
    > [xUnit.net 00:00:22.7496590]       ---- Expected collection to contain 2 item(s), but found 4.

There are quite a few tests which cause problem, I will keep extending this list. Feel free to add other tests as well so we can eventually fix them all!

### LINKED ISSUE #1640: Fix flaky test for MainViewModel
This test occasionally fails:
```
Failed   Toggl.Foundation.Tests.MvvmCross.ViewModels.MainViewModelTests+TheShouldShowWelcomeBackProperty.ReturnsFalseWhenThereAreSomeTimeEntries
Error Message:
 Expected False, but found True.
Stack Trace:
   at FluentAssertions.Execution.XUnit2TestFramework.Throw(String message)
   at FluentAssertions.Execution.AssertionScope.FailWith(String message, Object[] args)
   at FluentAssertions.Primitives.BooleanAssertions.BeFalse(String because, Object[] becauseArgs)
   at Toggl.Foundation.Tests.MvvmCross.ViewModels.MainViewModelTests.TheShouldShowWelcomeBackProperty.<ReturnsFalseWhenThereAreSomeTimeEntries>d__2.MoveNext() in /Users/vagrant/git/Toggl.Foundation.Tests/MvvmCross/ViewModels/MainViewModelTests.cs:line 639
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
```

simonrozsival:
Closes #1640 

Well... we had a problem in our tests. The problem was, that we ran tests in parallel. That was not actually the problem itself. The problem was, that our setup for MvvmCross tests wasn't meant to be run in parallel as we used a shared static instance of `MvxIoCProvider`. This meant that in some cases certain VMs shared an instance of e.g., the data source, and mocked the return values for certain method calls which meant a lot of race conditions and failing tests as the result.

Fortunately, we used the shared IoC provider to create VMs directly in our tests just twice - when creating `MainViewModel`, which contains `TimeEntriesLogViewModel` and `SuggestionsViewModel`, the latter two were constructed using `Mvx.IocConstruct<TimeEntriesLogViewModel>()` which conveniently injects the dependencies for us. I personally don't think that it is necessary to do that - we can manually pass the deps to these classes, most of them are already dependencies of the `MainViewModel` anyway.

These changes make the tests run deterministically and without race conditions. Creating the VMs in the `MainViewModel` by hand isn't the nicest solution for this subproblem. On the other hand, I think we should consider merging these three VMs into a single one as there is no need for sub views now that the main view was redesigned.

Squash whenever


simonrozsival:
I tired rebuilding the tests a few times and even though it seemed that the random fails are now gone, I had a few cases of a test failing:
```
Failed   Toggl.Foundation.Tests.MvvmCross.ViewModels.TimeEntriesLogViewModelTests+WhenReceivingAnEventFromTheTimeEntryCreatedObservable.SetsTheIsWelcomePropertyToFalse
Error Message:
 Expected False, but found True.
Stack Trace:
   at FluentAssertions.Execution.XUnit2TestFramework.Throw(String message)
   at FluentAssertions.Execution.AssertionScope.FailWith(String message, Object[] args)
   at FluentAssertions.Primitives.BooleanAssertions.BeFalse(String because, Object[] becauseArgs)
   at Toggl.Foundation.Tests.MvvmCross.ViewModels.TimeEntriesLogViewModelTests.WhenReceivingAnEventFromTheTimeEntryCreatedObservable.<SetsTheIsWelcomePropertyToFalse>d__2.MoveNext() in /Users/vagrant/git/Toggl.Foundation.Tests/MvvmCross/ViewModels/TimeEntriesLogViewModelTests.cs:line 230
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
```

This means that further investigation is necessary. I am convinced that these changes are necessary, but they might not be enough.

simonrozsival:
From looking at the tests and the implementation, it seems that there might be a problem, the test which failed (and the other tests in the `WhenReceivingAnEventFromTheTimeEntryCreatedObservable` class also fail from time to time and are simillar) looks like this:

```
[Fact, LogIfTooSlow]
public async ThreadingTask SetsTheIsWelcomePropertyToFalse()
{
    var observable = Observable.Return(true);
    OnboardingStorage.IsNewUser.Returns(observable);
    await ViewModel.Initialize();

    TimeEntryCreatedSubject.OnNext(NewTimeEntry.With((long)TimeSpan.FromHours(1).TotalSeconds));
    await ThreadingTask.Delay(200);

    ViewModel.IsWelcome.Should().BeFalse();
}
```

The VM buffers the changes in the data source in 100ms intervals, which might lead to the fact that `200ms` delay isn't enough to propagate the change and the value of the property isn't well set during assertion when tests run in parallel. When only this single test is run, it "never" fails. It seems to fail only if a "lot of tests" run at the same time.

Buffering will be removed by #1938. Still the delays should be removed and this test should work in a different way.

simonrozsival:
I decided that it is necessary to remove Buffering in this PR (buffering and the delays were introduced in this PR: https://github.com/toggl/mobileapp/pull/1765) so that we get rid of all of the failing tests.

After the last commit the tests don't fail anymore when I run them in a loop many times. I am quite confident in the VM tests now.

Toggl.Foundation.MvvmCross/ViewModels/TimeEntriesLogViewModel.cs
```diff
@@ -155,49 +154,39 @@ private void onTimeEntryUpdated((long Id, IDatabaseTimeEntry Entity) tuple)
...
+        private void safeInsertTimeEntry(IDatabaseTimeEntry timeEntry)
         {
-            if (!timeEntries.Any()) return;
-
             IsWelcome = false;
 
-            var groupsToInsert = timeEntries.GroupBy(te => te.Start.LocalDateTime.Date);
-
-            foreach (var group in groupsToInsert)
+            var indexDate = timeEntry.Start.LocalDateTime.Date;
```
amulware: Given the previous line 164, is this a possible NullReferenceException? Worth checking for at the top in either case? (until C#8 that is)

Toggl.Foundation.MvvmCross/ViewModels/TimeEntriesLogViewModel.cs
```diff
@@ -92,8 +92,7 @@ public sealed class TimeEntriesLogViewModel : MvxViewModel
             var createDisposable =
                 dataSource.TimeEntries.TimeEntryCreated
                     .Where(isNotRunning)
-                    .Buffer(timeEntryCreationBufferDuration)
```
amulware: Are we sure that the removal of the buffer won't reintroduce bugs that it was trying to fix in the past? Like the one where the start/stop button animated several times too fast so that it gets stuck in a funny state?

simonrozsival: The buffering had zero effect. The way _the bug_ was fixed doesn't really care about this part of code - rows are still processed one by one.

Also about the animation bug - there is throttling in Main VM. Buffering, again, has nothing to do with it.

amulware: Ah yes, I was confusing two things. :)

Toggl.Foundation.MvvmCross/ViewModels/TimeEntriesLogViewModel.cs
simonrozsival: Line 164 was there simply because `Buffer` would call this method EVERY 100 ms, no matter if there were any buffered changes or if the buffer was empty. If there was a `timeEntry == null` at this point, it would be an irrecoverable error I think. What should be the behavior if it was `null`?!

amulware: I guess it would just return. But if you say it really shouldn't happen, then I'm good.

PR #1940, opened 2018-04-21
Make deadlock tests SUPER fast

simonrozsival:
This PR makes deadlock tests SUPER fast.

- there were 7 property tests
- property tests run 100x each
- each of the tests awaits a delay of 1 ms 1-255x - the mean value is 128
- that made the deadlock tests 7x100x128 ms = 90s

Running these tests is not so important a few reasonably long tests are more than enough.

Squash whenever


simonrozsival:
The scary thing about this PR is that Bitrise runtime is just `9.2 min`... If I compare it to the average of  `17.2 min` on Friday, it makes me a bit ashamed for writing these tests so inefficiently in the first place (although they test a very serious thing!). My estimate of `1.5 min` is nowhere near the `8 min` this actually saves us on _every_ build.

Toggl.Foundation.Tests/Sync/DeadlockTests.cs
```diff
@@ -47,21 +47,21 @@ protected void Reset()
...
+            protected StateResult PrepareTransitions(StateResult entryPoint, int n)
             {
                 var lastResult = entryPoint;
                 for (byte i = 0; i < n; i++)
                 {
                     var nextResult = new StateResult();
                     Func<IObservable<ITransition>> transition = () => Observable.Create<ITransition>(async observer =>
                     {
-                        await Task.Delay(1);
+                        await Task.Delay(1).ConfigureAwait(false);
```
amulware: Why this change? Isn't this potentially changing the assumptions of the tests drastically?

simonrozsival: TBH the main reason for this change was that Bitrise got stuck on this line without this change (it would block and never continued). Maybe there would be a different way to get past this, but I don't think it is really an issue to use it like this.

I don't think it is change the assumptions of the tests. Why do you think it does?

amulware: I don't know, it's just... different. ;) But yes, I have noticed bitrise having issues with certain threading configurations for some reason. All good then.

PR #1945, opened 2018-04-23
Make calendar shortcuts in reports appear to be scrollable

### LINKED ISSUE #1705: Reports calendar shortcuts are not obviously scrolling
I haven't checked other phones, but on an iPhone 8 I can hardly tell that I can scroll the shortcuts under the calendar, given how the last one is cut off.

screen shot 2018-03-13 at 09 34 02

In fact I at first thought it was a smudge on my screen. 🙊

Maybe there is a way to make the shortcuts a bit wider so that the last one is always cut off in the text, not this close to the border.

Just an idea for @pe0ny

abhisheksengar:
Closes #1705


abhisheksengar:
@simonrozsival Thanks for the catch. I've pushed a fix.

PR #1946, opened 2018-04-23
Add a list of endpoints

### LINKED ISSUE #1921: Create maintainable list of endpoints used
I want to provide backend/other teams with a full list of endpoints we use (even those hardcoded in the integration tests), with small descriptions each on how they are used.

This will help them let us know about changes and in general be good for long term communication and planning and all that.

Ideally, we could extract (most?) of this from our code, or we'll write it by hand if that is easier. In either case, I want it to be always accessible, always up to date and version controlled.

Ideas welcome.

arturs105:
Closes #1921 

Squash whenever.


endpoints.md
```diff
@@ -0,0 +1,66 @@
+# Toggl endpoints
+These are endpoints that are prefixed with `https://mobile.toggl.com/api/v9/` (`mobile.toggl.space` domain is used when building the app in debug mode)
+
+* GET `me`
```
simonrozsival: I would consider using `GET` instead of GET.

endpoints.md
```diff
@@ -0,0 +1,66 @@
...
+  * `app_name` might be set to `toggl_mobile`, when using google login
+* POST `signup`
+  * `app_name` might be set to `toggl_mobile`, when using google login
+* PUT `me`
+* POST `me/lost_passwords`
+
+<br>
```
simonrozsival: I would prefer using the second level heading to separate the sections - `## Workspaces`, `## Tags`, ... what do you think?

amulware: Multi level headleads are good, since they can be automatically indexed.

endpoints.md
arturs105: Do you mean:
* `GET me/clients`
* `GET` `me/clients`

simonrozsival: The first one.

arturs105:
I don't think grouping by scope is necessary, because adding the headlines for current groups (by entity) makes the whole thing so much better.

arturs105:
I'm not sure about adding info on how we call the endpoints - isn't that more or less the same as the documentation of the endpoint?

amulware:
I would like to give backend/ops more information, so that they can verify that we're using the api correctly. But this is something we can still improve on in the future. I'll check with them as well, how useful this would be, and how much info they want, etc.

endpoints.md
```diff
@@ -0,0 +1,65 @@
...
+## Other
+
+* `GET status`
+* `GET me/features`
+
+# Reports endpoints
+These endpoints are prefixed with `https://mobile.toggl.com/reports/api/v3/` (`mobile.toggl.space` domain is used when building the app in debug mode)
+
+* `POST workspace/{workspace_id}/projects/summary`
+* `POST workspace{workspace_id}/search/projects`
```
simonrozsival: There is a missing `/`

endpoints.md
```diff
@@ -0,0 +1,65 @@
+# Toggl endpoints
+These are endpoints that are prefixed with `https://mobile.toggl.com/api/v9/` (`mobile.toggl.space` domain is used when building the app in debug mode)
```
amulware: Please add 'in debug mode [or for testing]', and a period at the end (same for the reports urls).

Issue #1947, opened 2018-04-23
Enable more StyleCop rules


amulware:
We haven't updated this in a long time. Let's see if there is more we can enable (that makes sense for us). Also, let's maybe double check if it's actually currently working in all projects.

kornelijepetak:
@amulware, what did you have in mind?
I can't think of anything that is really that important to introduce... 🤔

amulware:
Is having StyleCop at all useful then? I don't think we have that many rules enabled.

I was hoping we could enable some of the rules we already follow as per our guidelines.

And maybe there are other things as well that would make sense.

kornelijepetak:
I suggest waiting until I get more information from other devs when they return from 🌴

amulware:
If you need more time to get information, or if you want to do this after Mvexit like discussed in the PR, feel free to move the issue to the 2.1 release projects (both platforms), but do coordinate with the release managers.

kornelijepetak:
Yes, I do want to move this issue out because it does not make sense to repair stuff until MVX is out of the app, in `develop`.

heytherewill:
With #5460 merged, this can be safely moved back into the refactor drawing board. Let's latter discuss whether we should enable more rules or not before we close this for good :)

kornelijepetak:
@heytherewill I'm confused about this comment of yours.
Did not #5460 apply all proposed changes?
If not, why does it not specify which ones it did in particular?
Or did I understand your comment wrongly?

heytherewill:
@kornelijepetak It did, but that doesn't mean we don't wanna enable more rules in the future; hence why I'm not closing it

PR #1949, opened 2018-04-23
Deserialize errors from JSON and use message key

### LINKED ISSUE #1922: Ensure API errors can be either strings or JSON
To ensure forward compatibility, we should accept both strings (legacy) and JSON (the way forward) for all endpoints.

The JSON looks like:

```
{
    "message" : "I'm a user readable error message"
}
```

In the future, once all clients support the JSON format, backend can deprecate the string version and start sending additional data in the JSON errors.

juanlaube:
Closes #1922


Toggl.Ultrawave.Tests/Exceptions/ApiErrorResponsesTests.cs
```diff
@@ -159,13 +159,30 @@ public void SerializesMultipleHeaderKeyWithZeroOneOrMultipleValues()
 
                 serialized.Should().Be(expectedSerialization);
             }
+
+            [Fact, LogIfTooSlow]
+            public void SerializesErrorsAsJsonInResponse()
+            {
+                string message = "Couldn't find workspace with id blah blah blah....";
```
simonrozsival: You can use simply `var`, typing `string` is so many extra characters..! 😃

Toggl.Ultrawave/Exceptions/ApiException.cs
```diff
@@ -13,11 +15,18 @@ public class ApiException : Exception
...
-        internal ApiException(IRequest request, IResponse response, string message)
+        internal ApiException(IRequest request, IResponse response, string defaultMessage)
         {
             Request = request;
             Response = response;
-            this.message = message;
+
+            if (response.ContentType.Equals("application/json")) {
+                var serializer = new JsonSerializer();
+                var error = serializer.Deserialize<ResponseError>(response.RawData);
```
simonrozsival: This might throw - we should add a try-catch block here. Or will it always be caught somewhere?

Toggl.Ultrawave/Exceptions/ApiException.cs
```diff
@@ -13,11 +15,18 @@ public class ApiException : Exception
...
         private readonly string message;
 
-        internal ApiException(IRequest request, IResponse response, string message)
+        internal ApiException(IRequest request, IResponse response, string defaultMessage)
         {
             Request = request;
             Response = response;
-            this.message = message;
+
+            if (response.ContentType.Equals("application/json")) {
```
simonrozsival: It would be nice to have a small extension method `bool IsJson()` for the `IResponse` type.

Toggl.Ultrawave/Exceptions/ApiException.cs
juanlaube: I'm now catching and setting Message as the defaultMessage.

Toggl.Ultrawave/Exceptions/ApiException.cs
juanlaube: I've added IsJson as a property to IResponse.

Toggl.Ultrawave/Exceptions/ApiException.cs
```diff
@@ -20,12 +20,16 @@ internal ApiException(IRequest request, IResponse response, string defaultMessag
...
-            if (response.ContentType.Equals("application/json")) {
-                var serializer = new JsonSerializer();
-                var error = serializer.Deserialize<ResponseError>(response.RawData);
-                this.message = error.Message;
+            if (response.IsJson) {
+                try {
+                    var serializer = new JsonSerializer();
+                    var error = serializer.Deserialize<ResponseError>(response.RawData);
+                    this.message = error.Message;
+                } catch {
```
simonrozsival: Let's catch just `DeserializationException<ResponseError>`, everything else should propagate up the stack.

Toggl.Ultrawave/Exceptions/ApiException.cs
```diff
@@ -20,12 +20,16 @@ internal ApiException(IRequest request, IResponse response, string defaultMessag
             Request = request;
             Response = response;
 
-            if (response.ContentType.Equals("application/json")) {
-                var serializer = new JsonSerializer();
-                var error = serializer.Deserialize<ResponseError>(response.RawData);
-                this.message = error.Message;
+            if (response.IsJson) {
```
simonrozsival: One thing I didn't notice eariler - please put the opening bracket on a separate line. We don't speak Egyptian in the mobile team (although maybe if we could somehow overthrow our management... PM me, it is not safe to talk about that here..!)

The same applies to the rest of the try-catch block.

Toggl.Ultrawave.Tests/Exceptions/ApiErrorResponsesTests.cs
```diff
@@ -159,13 +159,30 @@ public void SerializesMultipleHeaderKeyWithZeroOneOrMultipleValues()
...
+            }
         }
 
         private static Request createRequest(HttpMethod method)
             => new Request("{\"a\":123}", new Uri("https://integration.tests"), new[] { new HttpHeader("X", "Y") }, method);
 
-        private static Response createErrorResponse(HttpStatusCode code, string rawData = "")
-            => new Response(rawData, false, "application/json", new List<KeyValuePair<string, IEnumerable<string>>>(), code);
+        private static Response createErrorResponse(HttpStatusCode code, string contentType = "plain/text", string rawData = "")
+        => new Response(rawData, false, contentType, new List<KeyValuePair<string, IEnumerable<string>>>(), code);
```
amulware: Removed indentation.

Toggl.Ultrawave/Exceptions/ApiException.cs
```diff
@@ -13,11 +15,22 @@ public class ApiException : Exception
...
-        internal ApiException(IRequest request, IResponse response, string message)
+        internal ApiException(IRequest request, IResponse response, string defaultMessage)
         {
             Request = request;
             Response = response;
-            this.message = message;
+
+            if (response.IsJson) {
+                try {
+                    var serializer = new JsonSerializer();
```
simonrozsival: Also one thing that I think of right now - the JSON will contain more information in the future (a code of the error and maybe even something else, who knows right now) - maybe it would make sense to put this code into a separate method and maybe even use the try-out pattern like this: `this.message = tryParseJson(response.RawData, out var message) ? message : defaultMessage;`

simonrozsival: Or maybe even better - the function would return `null` if it failed to deserialize it - then we could do just `message = getMessageFromJson(response) ?? defaultMessage;`

Toggl.Ultrawave/Exceptions/ApiException.cs
```diff
@@ -13,11 +15,22 @@ public class ApiException : Exception
...
+                    var error = serializer.Deserialize<ResponseError>(response.RawData);
+                    this.message = error.Message;
+                } catch {
+                    this.message = defaultMessage;
+                }
+            } else {
+                this.message = defaultMessage;
```
simonrozsival: If the type of the response isn't JSON, we should consider the whole body as the message => this should store `response.RawData` as the message here. This will be used as a fallback in the period (a few months?) when we release this to the user in an update, but backend still won't return JSON errors.

simonrozsival: https://github.com/toggl/mobileapp/pull/1938

Toggl.Ultrawave/Models/ResponseError.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System;
+using Newtonsoft.Json;
+
+namespace Toggl.Ultrawave.Models
+{
+    internal sealed class ResponseError
+    {
+        public string Message { get; set; }
+
+        [JsonConstructor]
```
amulware: There is no need for a json constructor and a publicly settable message property. Any reason to have both?

Toggl.Ultrawave/Exceptions/ApiException.cs
kornelijepetak: The latter suggestion is much better. However, `null` represents a lack of value, not an error value. But I definitely like `null` much more than `out`s. Another option would be to use tuple `(bool hasError, T message)`. Maybe we should use the `Maybe<T>` pattern. Pun intended. 😄

Toggl.Ultrawave/Exceptions/ApiException.cs
simonrozsival: The fallback in case of non-JSON response should be the whole `response.RawData` - the way it works right now. We will need this for a transition period when we release this code to the users, but backend still won't send us the JSONs (we need "all" users to update the app first before we can switch the behaviors).

Toggl.Ultrawave/Network/Response.cs
```diff
@@ -12,6 +12,8 @@ internal sealed class Response : IResponse
         public HttpStatusCode StatusCode { get; }
         public IEnumerable<KeyValuePair<string, IEnumerable<string>>> Headers { get; }
 
+        public bool IsJson => ContentType.Equals("application/json");
```
amulware: Any reason to use the Equals method over `==`?

Toggl.Ultrawave/Exceptions/ApiException.cs
simonrozsival: Well we have the mysterious `Either<TLeft, TRight>` type in the `Multivac` project, but nobody really knows (or remembers) how it works and how it should be used. Maybe somebody could give us a short intro to the `Multivac` helpers `New` and `Either`... (@amulware)

Toggl.Ultrawave/Exceptions/ApiException.cs
amulware: That exception is generic? Huh.

Toggl.Ultrawave/Exceptions/ApiException.cs
```diff
@@ -13,11 +15,22 @@ public class ApiException : Exception
...
         {
             Request = request;
             Response = response;
-            this.message = message;
+
+            if (response.IsJson) {
+                try {
+                    var serializer = new JsonSerializer();
+                    var error = serializer.Deserialize<ResponseError>(response.RawData);
+                    this.message = error.Message;
```
amulware: This assignment is wrong. The `message` field is a developer readable string (see usage for logging/debugging below). The correct property to use is `LocalizedApiErrorMessage` which contains the error string we can show to the user.

Toggl.Ultrawave/Exceptions/ApiException.cs
amulware: Very willing to do that, poke me on Slack. They are not terribly difficult.

Toggl.Ultrawave/Exceptions/ApiException.cs
amulware: No. Please see my comment below.

simonrozsival: I don't understand how your comment explains this... currently we use the whole body of the response as if it were the `LocalizedApiErrorMessage`. Please explain in more detail.

Toggl.Ultrawave/Exceptions/ApiException.cs
simonrozsival: Yep.

amulware:
On another note, it might be nice to have some more testing for this, to make sure exceptions are handled and thrown correctly when using this in an actual api client.

Toggl.Ultrawave/Exceptions/ApiException.cs
juanlaube: I'm all in for a revolution 😄 
On a more serious question, why don't we use a linter for things like this?

amulware: Either because StyleCop (our linter) doesn't support it, or because we haven't actually been doing much with StyleCop since the initial setup: https://github.com/toggl/mobileapp/issues/1947

Toggl.Ultrawave/Exceptions/ApiException.cs
juanlaube: I went with `LocalizedApiErrorMessage = getLocalizedMessageFromResponse(response) ?? defaultMessage;`

Toggl.Ultrawave.Tests/Exceptions/ApiErrorResponsesTests.cs
juanlaube: Could this be some configuration in VS? I don't remember removing this 🤔

Toggl.Ultrawave/Network/Response.cs
juanlaube: No reason at all

Toggl.Ultrawave/Exceptions/ApiException.cs
juanlaube: You're right, I left message as it was, now I'm assigning the `LocalizedApiErrorMessage`.

Toggl.Ultrawave.Tests/Exceptions/ApiErrorResponsesTests.cs
```diff
@@ -159,13 +159,53 @@ public void SerializesMultipleHeaderKeyWithZeroOneOrMultipleValues()
...
+            [Fact, LogIfTooSlow]
+            public void DeserializesLocalizedErrorMessageAsJsonInResponse()
+            {
+                var message = "Couldn't find workspace with id blah blah blah...";
+                var body = $"{{\"message\": \"{message}\"}}"; 
+                var endpoint = new Uri("https://www.some.url");
+                var method = new HttpMethod("GET");
+                var request = new Request("", endpoint, new HttpHeader[0], method);
+                var response = new Response(body, false, "application/json", new List<KeyValuePair<string, IEnumerable<string>>>(), HttpStatusCode.NotFound);
+                var exception = new NotFoundException(request, response);
```
amulware: Why not test `ApiException` directly in these tests? I think that's better than testing a single specific case.

amulware:
Otherwise looking good I think. :)

Toggl.Ultrawave.Tests/Exceptions/ApiErrorResponsesTests.cs
juanlaube: Done

juanlaube:
I've found a problem with the JSON deserialization, for some reason the property wasn't being mapped, the fix I've found was to add a [JsonProperty("message")] annotation.

amulware:
The mapping problem is very strange. Maybe it's related to the default configuration of the json serialiser.

amulware:
Also note that the last commit significantly changes behaviour (to be more in line with how this worked before the PR). But that's fine. Probably better even.

juanlaube:
I know, this is why CI was failing, some tests in Foundation where using the message differently.
Some tests are failing now for unrelated reasons? 🤔

Toggl.Ultrawave/Network/Interface/IResponse.cs
```diff
@@ -14,5 +14,9 @@ internal interface IResponse
         HttpStatusCode StatusCode { get; }
 
         IEnumerable<KeyValuePair<string, IEnumerable<string>>> Headers { get; }
+
+        bool IsJson { get; }
+
+        bool IsText { get; }
```
simonrozsival: Since it is not used anymore, we don't need it and it can be 🔥

Toggl.Ultrawave/Exceptions/ApiException.cs
```diff
@@ -35,5 +37,23 @@ internal static string SerializeHeaders(IEnumerable<KeyValuePair<string, IEnumer
...
+            {
+                try
+                {
+                    var serializer = new JsonSerializer();
+                    var error = serializer.Deserialize<ResponseError>(response.RawData);
+                    return error.Message;
+                }
+                catch (DeserializationException<ResponseError>)
+                {
+                    return null;
```
simonrozsival: Maybe we should return the `""` directly from here and get rid of the `?? "";` on line 24.

Toggl.Ultrawave/Network/Interface/IResponse.cs
juanlaube: 🔥ed

Toggl.Ultrawave/Models/ResponseError.cs
```diff
@@ -1,7 +1,10 @@
-﻿namespace Toggl.Ultrawave.Models
+﻿using Newtonsoft.Json;
+
+namespace Toggl.Ultrawave.Models
 {
     internal sealed class ResponseError
     {
-        public string Message { get; }
+        [JsonProperty("message")]
```
simonrozsival: This is definitely not needed. 🔥

Toggl.Ultrawave/Models/ResponseError.cs
```diff
@@ -0,0 +1,10 @@
+﻿using Newtonsoft.Json;
+
+namespace Toggl.Ultrawave.Models
+{
+    internal sealed class ResponseError
+    {
+        [JsonProperty("message")]
```
simonrozsival: This is definitely not needed.

juanlaube: 🔥ed

PR #1951, opened 2018-04-24
:construction: Refactor StartTimeEntryViewModel initialization

Odrakir:
This is not directly addressing any issue, but it takes care of a potentially ugly bug.

The thing is that `Prepare(StartTimeEntryParameters parameter)` was being called twice in `StartTimeEntryViewModel`, which caused the observable sequence to execute twice. This is due to the use of `Init`, which is needed for 3D Touch to work.

My solution is to use the method `Prepare` (without params) along with `Init`. `Init` is called only for 3D Touch, and `Prepare()` is called always. According to the [docs](https://www.mvvmcross.com/documentation/fundamentals/navigation) `Prepare()` is where parameter agnostic stuff should be initialized.

Please review and test thoroughly, I have no idea what I'm doing.

### DIFF
--- a/Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
+++ b/Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
@@ -222,6 +222,22 @@ namespace Toggl.Foundation.MvvmCross.ViewModels
             Prepare(startTimeEntryParameters);
         }
 
+        public override void Prepare()
+        {
+            var queryByTypeObservable = queryByTypeSubject
+                .AsObservable()
+                .SelectMany(type => dataSource.AutocompleteProvider.Query(new QueryInfo("", type)));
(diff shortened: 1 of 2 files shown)


Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
```diff
@@ -214,29 +214,19 @@ public bool ShouldShowNoProjectsInfoMessage
         }
 
         public void Init()
+        {
+            //Prepare(startTimeEntryParameters);
```
simonrozsival: 🔥 !

arturs105:
Overall looks good and everything seems to be working, however a lot of tests are failing, so you have to update those as well. I suspect the view model isn't properly initialized in the tests, so adding `Prepare` and `Initialize` calls here and there might solve the issue.

PR #1952, opened 2018-04-24
Refactor models

### LINKED ISSUE #1890: Refactor models
We want to:
- list all differences between models
- see which of these can be unified to make handling them generically easier
- consider moving more stuff to the IDatabaseSyncable interface (or make (an)other ones?)
- do that

simonrozsival:
Closes #1890 

This PR:
- unifies the data types of common properties (e.g, the `At` property)
- adds several interfaces which contain certain traits of the models
- updates the base pull state to show the use case of the updated models
- breaks the functionality of the T4 templates - these should be fixed once this pr is merged

Squash whenever


simonrozsival:
Please note that the generated files were edited manually and the T4 templates will be changed in a different PR later.

Toggl.Multivac/Models/ISingleEntity.cs
```diff
@@ -0,0 +1,6 @@
+namespace Toggl.Multivac.Models
+{
+    public interface ISingleEntity : IIdentifiable
```
amulware: What's this for?

Toggl.Multivac/Models/ISyncable.cs
```diff
@@ -0,0 +1,9 @@
+using System;
+
+namespace Toggl.Multivac.Models
+{
+    public interface ISyncable
```
amulware: I'm not sure I like the name, it's talking about stuff that happens in foundation which other projects don't have to know about.

Toggl.PrimeRadiant.Realm/Database.cs
```diff
@@ -61,13 +62,25 @@ private Realms.Realm getRealmInstance()
...
                     }
+
+                    if (oldSchemaVersion < 3)
+                    {
+                        var newTags = migration.NewRealm.All<RealmTag>();
+                        var oldTags = migration.OldRealm.All("RealmTag");
+                        for (var i = 0; i < newTags.Count(); i++)
+                        {
+                            var oldTag = oldTags.ElementAt(i);
+                            var newTag = newTags.ElementAt(i);
```
amulware: the order of elements is guaranteed to stay the same?

Toggl.PrimeRadiant.Realm/Models/RealmPreferences.cs
```diff
@@ -1,11 +1,15 @@
...
 using Toggl.Multivac;
 using Toggl.PrimeRadiant.Models;
+using System;
 
 namespace Toggl.PrimeRadiant.Realm
 {
     internal partial class RealmPreferences : RealmObject, IDatabasePreferences
     {
+        [Ignored]
+        public DateTimeOffset At => DateTimeOffset.Now;
```
amulware: Yeah, we should really get rid of these 'now's and use some fixed default date.

amulware: Not to mention we shouldn't have a model property that returns different values at different times!

Toggl.Foundation/Models/FoundationModels.cs
```diff
@@ -267,6 +269,10 @@ internal partial class WorkspaceFeature : IDatabaseWorkspaceFeature
 
     internal partial class WorkspaceFeatureCollection : IDatabaseWorkspaceFeatureCollection
     {
+        public long Id => WorkspaceId;
```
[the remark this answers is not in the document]
simonrozsival: We already use this as an "id" of the collection - https://github.com/toggl/mobileapp/blob/develop/Toggl.Foundation/Sync/States/Pull/PersistWorkspaceFeaturesState.cs#L18

Toggl.Foundation/Models/FoundationModels.cs
```diff
@@ -267,6 +269,10 @@ internal partial class WorkspaceFeature : IDatabaseWorkspaceFeature
...
+
+        public DateTimeOffset At => DateTimeOffset.Now;
```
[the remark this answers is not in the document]
simonrozsival: Do you have some favourite date which should be used instead?

I will try to get rid of this property altogether.

Toggl.Foundation/Sync/States/Pull/BasePersistState.cs
```diff
@@ -64,14 +79,11 @@ private IObservable<ITransition> processError(Exception exception)
         private bool shouldRethrow(Exception e)
             => e is ApiException == false || e is ApiDeprecatedException || e is ClientDeprecatedException || e is UnauthorizedException;
 
-        protected abstract long GetId(TDatabaseInterface entity);
+        private DateTimeOffset? lastUpdated(IEnumerable<ISyncable> entities)
+            => entities.Select(p => p?.At).Where(d => d.HasValue).DefaultIfEmpty(DateTimeOffset.Now).Max();
```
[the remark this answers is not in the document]
simonrozsival: This code doesn't make any sense when I look at it now. We removed nullable `At`s, so the only possibility for it to be empty is if the entities list is empty, but then it should not be updated at all. I will get rid of this.

Toggl.Multivac/Models/ISingleEntity.cs
simonrozsival: Maybe it won't be needed at all, but I wanted to make it explicit that this ID which is inherited from the `IIdentifiable` is in fact "fake". I haven't come up with a better way of expressing it so far.

Toggl.Multivac/Models/ISyncable.cs
simonrozsival: I am sure I don't like the name. Suggestions are welcome!

Toggl.PrimeRadiant.Realm/Database.cs
simonrozsival: This was taken directly from Realm docs. I think it only makes sense to be in the same order - otherwise these migrations won't be "`O(n)`" if you had to search for the old version for every entity.

Toggl.PrimeRadiant.Realm/Models/RealmPreferences.cs
simonrozsival: You are right. We should pick a fixed date. I will try to get rid of this property altogether - the same thing as with the workspace feature collection.

Toggl.PrimeRadiant.Realm/Database.cs
amulware: One can index them. /shrug but sure.

Toggl.PrimeRadiant.Realm/Models/RealmPreferences.cs
amulware: Sounds good.

simonrozsival:
Just to be clear and have it written somewhere: I want to rewrite the syncing mechanism & the states => the what is now `BasePersistState` will be changed into a different generic (sealed) "state" and will be probably modified. This means that right now I don't really bother with the entity-specific unit tests (e.g., persist workspace feature collection state + persist preferences state skip some tests right now) and I don't want to spend much time on making the old test code work the same way it used to do right now.

amulware:
Sounds good, I didn't bother checking the test code in detail either :D

Toggl.Multivac/Models/ISyncable.cs
simonrozsival: `ILastChange`, `IChangeable`, `IModificationTrackable` ... naming is hard :D

simonrozsival:
This PR is now ready for review, @amulware

Toggl.Foundation/Sync/States/Pull/BasePersistState.cs
```diff
@@ -26,35 +30,48 @@ internal abstract class BasePersistState<TInterface, TDatabaseInterface>
...
-                .Select(sinceParameters => new FetchObservables(fetch, sinceParameters))
+                .Select(listOfDatabaseEntities)
+                .SelectMany(batchUpdate)
+                .Select(updateFetchObservable(fetch))
                 .Select(FinishedPersisting.Transition)
                 .Catch((Exception exception) => processError(exception));
-        }
+
+        private IList<TDatabaseInterface> listOfDatabaseEntities(IEnumerable<TInterface> entities)
+            => entities?.Where(entity => entity != null).Select(convertToDatabaseEntity).ToList()
```
amulware: Why would entities ever be null here?

Toggl.Multivac/Models/ISingleEntity.cs
amulware: I'd rather remove it until we need it. Especially since it's only used in one place.

Toggl.PrimeRadiant.Realm/Database.cs
```diff
@@ -61,13 +62,25 @@ private Realms.Realm getRealmInstance()
...
+                            newTag.ServerDeletedAt = oldTag.DeletedAt;
```
amulware: The old property is nullable. Are we guaranteed that it always has a value?

Toggl.Ultrawave/Models/Project.cs
```diff
@@ -18,6 +19,14 @@ internal sealed partial class Project : IProject
 
         public bool Active { get; set; }
 
+        [JsonProperty("at")]
+        public DateTimeOffset? DirtyAt
+        {
+            get => At;
+            set => At = value ?? DateTimeOffset.UtcNow;
```
amulware: Similar as above.

amulware: Ok, and several times below. Not putting comments on all of them.

Toggl.Foundation/Models/Client.cs
```diff
@@ -19,7 +20,15 @@ internal sealed class Builder
 
             public long? WorkspaceId { get; private set; }
 
-            public DateTimeOffset? At { get; private set; }
+            [JsonProperty("at")]
+            public DateTimeOffset? DirtyAt
+            {
+                get => At;
+                set => At = value ?? DateTimeOffset.UtcNow;
```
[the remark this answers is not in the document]
simonrozsival: Why wouldn't it be `default(DTO?) == null`?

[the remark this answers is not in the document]
amulware: I mean the At property which will be used by Foundation.

Toggl.PrimeRadiant.Realm/Database.cs
amulware: Ok, the new one is nullable too. I was confusing two things. Modified question: are we migrating any `At` values from nullable to non-nullable and are we sure THOSE always have values?

simonrozsival:
This PR will be replaced by several smaller PRs.

PR #1955, opened 2018-04-24
Integrate Firebase Analytics for Giskard

juanlaube:


amulware:
Is it sending events also for debug/adhoc builds? I believe for iOS we only send events with some compiler flag that gets added only for release builds. We should probably do the same here too.

simonrozsival:
@juanlaube does the build work for you locally? (using Cake)

amulware:
The code looks ok at a glance. Let's get that build fixed and we'll be good to go 🙌

PR #1964, opened 2018-04-26
📐Update constraints

### LINKED ISSUE #1485: Update hit boxes
It can be a bit difficult to tap some elements like duration boxes, confirm buttons, etc.

We should check that the tappable areas for all elements are as big as they should be (in many cases bigger than the visual element, especially for small things).

sofokli1:
Closes #1485

Most changes are to the back/close and and save/create button sizes. Also I increased some textFields heights and some other small buttons that were very hard to click without a mouse.


amulware:
🙌 Please don't forget about emojis when squashing. ;)

simonrozsival:
Oh boy so many changed xib files 😱

PR #1965, opened 2018-04-26
Fix 3D shortcuts when app is closed

arturs105:
All of the shortcuts are now working even if the app is closed, however there is a glitch, where the current entry card does not show for a few seconds when using "Continue most recent entry". I'm gonna create a separate issue for that.

Edit: here's the issue: #1966


amulware:
If a potential reviewer agrees, this _could_ be merged into the release branch.

Toggl.Daneel/Startup/AppDelegate.cs
```diff
@@ -74,8 +80,31 @@ public override void DidEnterBackground(UIApplication application)
...
+                    break;
+
+                case ShortcutType.Reports:
+                    navigationService.Navigate<ReportsViewModel>();
+                    break;
+
+                case ShortcutType.StartTimeEntry:
+                    navigationService.Navigate<StartTimeEntryViewModel>();
+                    break;
+            }
```
simonrozsival: The default case is "just open the app with the root controller"?

Toggl.Daneel/Startup/AppDelegate.cs
```diff
@@ -74,8 +80,31 @@ public override void DidEnterBackground(UIApplication application)
         public override void PerformActionForShortcutItem(UIApplication application, UIApplicationShortcutItem shortcutItem, UIOperationHandler completionHandler)
         {
             analyticsService.TrackAppShortcut(shortcutItem.LocalizedTitle);
-            var url = shortcutItem.UserInfo[nameof(ApplicationShortcut.Url)].ToString();
-            navigationService.Navigate(url);
+
+            var shortcutType = (ShortcutType)(int)(NSNumber)shortcutItem.UserInfo[nameof(ApplicationShortcut.Type)];
```
simonrozsival: Could this be simplified somehow? 😃

arturs105:
I don't think we can remove url navigation, since it may still be used in Android - William implemented opening the app with urls and we should do the same for iOS with x-callback-url.

Toggl.Daneel/Startup/AppDelegate.cs
arturs105: Yes. iOS opens the app in the main screen by default - it would happen even if we didn't implement `PerformActionForShortcutItem`.

Toggl.Daneel/Startup/AppDelegate.cs
arturs105: I wasn't able to find a better way to do this. I tried `(ShortcutType)(object)theThing` - it at least compiled, but then threw a runtime exception 😞

simonrozsival:
Let's merge this into `neptune` as Paul said.

arturs105:
Moved to #1975

PR #1973, opened 2018-04-26
Track showing project or tag suggestions when creating a time entry

### LINKED ISSUE #1777: Track project and tag addition in TE creation view

Odrakir:
Closes #1777 

As described in the comments to the issue this tracks every different way of showing the project or tags suggestions in the creation view.

It tracks two events for tag and projects, each of those with a parameter indicating the source.

The parameters are as follows:
When typed -> TextField
When the button over the keyboard is tapped -> ButtonOverKeyboard, TextField
When the button on the table view is tapped -> TableCellButton, TextField

Squash whenever.


Toggl.Foundation.Tests/MvvmCross/ViewModels/StartTimeEntryViewModelTests.cs
```diff
@@ -1182,6 +1223,26 @@ public void SetsTheIsDirtyFlag()
 
                     ViewModel.IsDirty.Should().BeTrue();
                 }
+
+                [Fact, LogIfTooSlow]
+                public void TracksProjectSelectionWhenProjectSymbolSelected()
+                {
+                    QuerySymbolSuggestion projectSuggestion = QuerySymbolSuggestion.Suggestions
+                                                                                   .Where((s) => s.Symbol == QuerySymbols.ProjectsString)
+                                                                                   .First();
```
arturs105: I think formatting here should be improved. I suggest this:
```
QuerySymbolSuggestion projectSuggestion = QuerySymbolSuggestion
    .Suggestions
    .Where((s) => s.Symbol == QuerySymbols.ProjectsString)
    .First();
```

The same applies to the next test as well.

Toggl.Foundation.Tests/MvvmCross/ViewModels/StartTimeEntryViewModelTests.cs
```diff
@@ -1684,4 +1745,4 @@ public sealed class TheShouldShowNoTagsInfoMessage : StartTimeEntryViewModelTest
             }
         }
     }
-}
+}
```
arturs105: Let's keep the newline at the end

Toggl.Foundation/Analytics/ProjectTagSuggestionSource.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
```
arturs105: None of these `using` statements are necessary.

Toggl.Foundation.Tests/MvvmCross/ViewModels/StartTimeEntryViewModelTests.cs
Odrakir: Is this in the style guide? Because I don't agree. As I read it `QuerySymbolSuggestion.Suggestions` is the object on which the following operators (one for each line) are going to act.

If it's in the style guide I'll change it, but in my opinion it's better this way.

Toggl.Foundation.Tests/MvvmCross/ViewModels/StartTimeEntryViewModelTests.cs
Odrakir: Ok, but why?

arturs105: Consistency

Toggl.Foundation.Tests/MvvmCross/ViewModels/StartTimeEntryViewModelTests.cs
amulware: I don't care where the lines are split exactly, but four spaces of indentation, please.

Toggl.Foundation.Tests/MvvmCross/ViewModels/StartTimeEntryViewModelTests.cs
Odrakir: Yeah, of course, we don't want someone having a hard time because of a missing new line at the end of the line... :trollface:

Toggl.Foundation.Tests/MvvmCross/ViewModels/StartTimeEntryViewModelTests.cs
arturs105: Okay, we can leave the `.Suggestions` on the previous line (that wasn't the problem), but the next lines shouldn't be aligned below the `.Suggestions`, but instead use an additional level of indentation (yes, the guidelines say so). Something like this would be fine:
```
QuerySymbolSuggestion projectSuggestion = QuerySymbolSuggestion.Suggestions
    .Where((s) => s.Symbol == QuerySymbols.ProjectsString)
    .First();
```

Odrakir: Yes, that makes sense.

PR #1982, opened 2018-04-27
Fix "continue most recent entry" shortcut glitch

### LINKED ISSUE #1966: Fix visual glitch when using "Continue last entry" app shortcut
The shortcut itself is working and starts tracking time, however for the first few seconds the current entry card is not shown and the green start button is disabled. The card appears after about 5 seconds.

Steps to reproduce:
1. Close the app completely (swipe away from recents)
2. 3D touch on the app
3. Press "Continue last entry"
4. 💥 

This is how it looks for about 5 seconds:
screen shot 2018-04-26 at 12 09 55

arturs105:
Closes #1966 

Squash whenever.


simonrozsival:
Wow. How does this fix the problem? Do you have some more info/reasoning? I am just interested because it is not very obvious. I will test this in a minute.

arturs105:
I assume something with race conditions - this removed line set the card to invisible and at the same time `TimeEntriesDataSource.CurrentlyRunningTimeEntry` would emit an entry and the card would be set to visible and since those would happen at roughly the same time, the card would be set to visible and invisible at the same time, so 💥

simonrozsival:
And isn't it possible that without this line it could happen that the card could sometimes be hidden, even though it should be visible?

arturs105:
I tested out these scenarios:
* There is a running entry:
  * Continue app from recent apps
  * Clear recents and then start the app
* There isn't a running entry:
  * Continue app from recent apps
  * Clear recents and then start the app

The app behaved as expected in all cases. Though we should probably pay close attention to this, when testing the RC.

Also: i could add a test that checks, whether `TimeEntriesDataSource.CurrentlyRunningTimeEntry` observable emits a value upon subscription.

PR #1985, opened 2018-04-27
Fix app shortcut bug

### LINKED ISSUE #1984: The AppDelegate class holds reference to the old data source after logout
Steps to reproduce (on the `neptune` branch):
- log in
- make sure there is at least one TE in the log
- stop any running time entry
- log out
- log in
- use the force-touch shortcut to continue most recent time entry
- the current runnint TE card is in a broken state

The fix should be easy - just don't store the reference to the interactor factory in the `AppDelegate` class and resolve it just before continuing the TE.

arturs105:
Closes #1984 

Squash whenever.


Toggl.Daneel/Startup/AppDelegate.cs
```diff
@@ -86,6 +84,8 @@ public override void PerformActionForShortcutItem(UIApplication application, UIA
             switch (shortcutType)
             {
                 case ShortcutType.ContinueLastTimeEntry:
+                    var interactorFactory = Mvx.Resolve<IInteractorFactory>();
+                    if (interactorFactory == null) return;
                     IDisposable subscription = null;
                     subscription = interactorFactory
                         .ContinueMostRecentTimeEntry()
```
simonrozsival: Wouldn't it be enough to just do `interactorFactory?.ContinueNostRecentTimeEntry()`?

PR #1992, opened 2018-04-30
Remove all of the syncing code

simonrozsival:
This is the first step of the refactoring:
- removing all of the old internal implementation, so that future changes to the syncing algorithm are more straightforward; these changes will include:
    - completely rewritten `StateMachine` and all of the states
    - there won't be any `StateMachineOrchestrator` (it will not be necessary)
    - all of the states will be simplified and will be 100% generic
    - models will be updated to allow ☝️ 
    - the repositories will be updated to simplify working with the entities (UI safe models and repository wrappers - no need to explicitly cast from realm to foundation models in the states)
- this PR does not change the interface of the syncing algorithm, so the app should still work (but it won't sync at all... obviously)


Toggl.Foundation.Tests/Sync/SyncStateQueueTests.cs
```diff
@@ -1,9 +1,9 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
-using FluentAssertions;
 using FsCheck;
 using FsCheck.Xunit;
+using Toggl.Foundation.Tests.Extensions;
```
amulware: Huh?

Toggl.Foundation.Tests/Extensions/ListOfSyncStateExtensions.cs
```diff
@@ -0,0 +1,34 @@
...
+using System.Collections.Generic;
+using FluentAssertions;
+using Toggl.Foundation.Sync;
+using Toggl.Multivac;
+
+namespace Toggl.Foundation.Tests.Extensions
+{
+    internal static class ListOfSyncStateExtensions
+    {
+        internal static void ShouldBeSameEventsAs(this List<SyncState> actualEvents,
```
amulware: Why is this here?

Toggl.Foundation.Tests/Sync/SyncStateQueueTests.cs
simonrozsival: I had to move an extension method from the orchestrator tests to a separate file.

Toggl.Foundation.Tests/Extensions/ListOfSyncStateExtensions.cs
simonrozsival: ... this extension method.

PR #1993, opened 2018-04-30
Syncing refactoring

### LINKED ISSUE #1309: Use generic composition instead of inheritance for sync states
Currently, we use inheritance to implement all the entity-specific sync states from their abstract generic base classes.

I think the code and especially the tests could be much simpler if instead we have only a single generic sync state and inject the specific behaviour using interfaces and small objects for the different entities.

Consider this partly a suggestion, I didn't actually make the changes to see if works better. It's more of a hunch really. Other opinions welcome.

simonrozsival:
Closes #1309 

The PR refactoring removes a lot of entity-specific syncing code. There is now single state for persisting an entity and one push state per operation (create, update, delete on server, delete locally) and marking entity as unsyncable instead of having a specific class per each entity - a lot of code and especially test code could be 🔥 🎉 

A lot of changes had to be made to make this possible - the models had to be standardized (interfaces `IIdentifiable`, `IDeletable`, `ILastChangedDatable`), the API clients had to be standardized (interfaces `IPullingApiClient`, `IPullingSingleApiClient`, `IPullingChangedApiClient`, `ICreatingApiClient`, `IUpdatingApiClient`, `IDeletingApiClient`). Next, the database repositories had to be wrapped to ensure that we don't access the raw Realm objects and cause exceptions - this meant adding the `IThreadsafeModel` and all the "threadsafe" entity models (e.g., `IThreadsafeTag`) into Foundation. We now have generic data sources (`IDataSource`, `IObservableDataSource`, `ISingletonDataSoruce`) which can be safely and easily used for conflict resolution.

What this PR doesn't change: the sync manager/state machine/state machine orchestrator and related mechanisms.

This PR is very large - many smaller PRs were merged into this branch and so review commit by commit (these sub-PRs were already reviewed by someone). There are a few 💚 commits which were made directly into this branch - pay special attention to these as they hadn't been reviewed before.


simonrozsival:
The refactoring is now complete. It is now waiting for a release branch for 1.6 to be created so `develop` is "safe" to be merged into. Also @willsb wanted to have a look at PR #2133 and maybe request some further changes.

amulware:
I checked the conflicts, it's all in using statements, so shouldn't be a big deal. Maybe it'll need another 💚 commit, if so, I'd like to check that before this gets 🦑 :)

PR #1994, opened 2018-04-30
Rename the `IBaseModel` interface

simonrozsival:
The old name did not describe what it actually represents.

Squash whenever


simonrozsival:
Ref  #1890

PR #1996, opened 2018-04-30
Add IDeletable interface

simonrozsival:
This PR standardizes the `DateTimeOffset? ServerDeletedAt { get; }` property for the models, which can be deleted on the server and later removed from the local storage on pull-sync.

This change requires a simple database migration of the tag entity because of a change in the name of the `DeletedAt` property to `ServerDeletedAt`.

Squash whenever


simonrozsival:
Ref  #1890

PR #1999, opened 2018-04-30
`SinceParameterStorage` refactoring

simonrozsival:
This PR introduces a different implementation of the `SinceParameterStorage` than we used previously. The differences are:
- previously the storage contained a single object - this object had a column for each since date
    - during pull sync, we had to pass this object between states and had to carefully update its values
- now the storage contains several objects - at most one for each supported type
    - it might happen that we don't have any value for a given type - in this case it will either return the default value (for `Get<T>()` or create a new record (for `Set<T>(since)`)
    - there is a fixed mapping from types to IDs - this should not be changed or it would lead to potentially incorrect behaviors
    - the pull states can now work independently accessing the entity type they are responsible for

We could possibly create a migration which would store the values from the old storage to the new, but I don't think it is necessary - the first sync will fetch all and then it will create the records automatically, so the next fetch will use the since params as usual.

Squash whenever


Toggl.PrimeRadiant.Realm/SinceParameterStorage.cs
```diff
@@ -1,56 +1,74 @@
...
-        public SinceParameterStorage(Func<Realms.Realm> getRealmInstance)
-        {
-            Ensure.Argument.IsNotNull(getRealmInstance, nameof(getRealmInstance));
-
-            this.getRealmInstance = getRealmInstance;
-        }
+        private static readonly Dictionary<Type, long> typesToIdsMapping
+            = new Dictionary<Type, long>
+            {
+                [typeof(IDatabaseClient)] = 0,
```
amulware: This syntax. I'm not convinced I like it. :D

Toggl.PrimeRadiant.Realm/SinceParameterStorage.cs
```diff
@@ -1,56 +1,74 @@
...
-            var realm = getRealmInstance();
-            using (var transaction = realm.BeginWrite())
+            try
             {
-                parameters = getOrCreateRealmObject(realm);
-                mutateParameters?.Invoke(parameters);
-                transaction.Commit();
+                realmAdapter.Update(id, record);
+            }
+            catch (InvalidOperationException)
```
amulware: Isn't it nicer to try-fetch and update/create in a transaction compared to handling exceptions?

Toggl.PrimeRadiant.Realm/SinceParameterStorage.cs
```diff
@@ -1,56 +1,74 @@
...
+        private readonly IRealmAdapter<IDatabaseSinceParameter> realmAdapter;
 
-        public SinceParameterStorage(Func<Realms.Realm> getRealmInstance)
-        {
-            Ensure.Argument.IsNotNull(getRealmInstance, nameof(getRealmInstance));
-
-            this.getRealmInstance = getRealmInstance;
-        }
+        private static readonly Dictionary<Type, long> typesToIdsMapping
+            = new Dictionary<Type, long>
```
amulware: Any reason to use `long` over `int`?

Toggl.PrimeRadiant/Models/ISinceParameters.cs
```diff
@@ -1,14 +1,10 @@
 ﻿﻿using System;
+ using Toggl.Multivac.Models;
 
 namespace Toggl.PrimeRadiant.Models
 {
-    public interface ISinceParameters
+    public interface IDatabaseSinceParameter : IIdentifiable
```
amulware: Filename?

Toggl.PrimeRadiant.Tests/SinceParametersStorageTests.cs
```diff
@@ -0,0 +1,187 @@
...
+            public void ThrowsForUnsupportedTypes(Type type)
+            {
+                Action get = () => Get(type);
+
+                get.ShouldThrow<ArgumentException>();
+            }
+
+            [Theory]
+            [MemberData(nameof(SupportedTypes), MemberType = typeof(SinceParametersStorageTests))]
+            public void DoesNotThrowForUnsupportedTypes(Type type)
```
amulware: Type: `supported`

Toggl.PrimeRadiant.Tests/SinceParametersStorageTests.cs
```diff
@@ -0,0 +1,187 @@
...
+            public void ThrowsForUnsupportedTypes(Type type)
+            {
+                Action get = () => Set(type, someDate);
+
+                get.ShouldThrow<ArgumentException>();
+            }
+
+            [Theory]
+            [MemberData(nameof(SupportedTypes), MemberType = typeof(SinceParametersStorageTests))]
+            public void DoesNotThrowForUnsupportedTypes(Type type)
```
amulware: typo

Toggl.PrimeRadiant.Tests/SinceParametersStorageTests.cs
```diff
@@ -0,0 +1,187 @@
...
+                Set(type, someDate.AddDays(1));
+
+                var allRecords = TestAdapter.GetAll().ToList();
+                allRecords.Should().HaveCount(1);
+                allRecords[0].Since.Should().Be(someDate.AddDays(1));
+            }
+
+            [Theory]
+            [MemberData(nameof(SupportedTypes), MemberType = typeof(SinceParametersStorageTests))]
+            public void UpdatesARecordToNullWhenThereWasNoRecordForThisTypeBefore(Type type)
```
amulware: typo

Toggl.PrimeRadiant.Tests/SinceParametersStorageTests.cs
```diff
@@ -0,0 +1,187 @@
...
+            {
+                Set(type, someDate);
+
+                Set(type, null);
+
+                var allRecords = TestAdapter.GetAll().ToList();
+                allRecords.Should().HaveCount(1);
```
amulware: Do we really care about the count? as long as Get(type) returns null?

Toggl.PrimeRadiant.Realm/SinceParameterStorage.cs
simonrozsival: You should get used to it :D Or do you mean using `typeof(X)` as keys?

amulware: I mean `[key] = value` as opposed to `{key, value}`.

simonrozsival: Yes, the type of the `id` in Realm is `long`.

Toggl.PrimeRadiant.Realm/SinceParameterStorage.cs
simonrozsival: I would like to do it in a transaction, but it is not possible with `RealmAdapter`. If we go back to using `Realm` directly, we won't be able to test the behavior of this class. A solution would be to add a method `UpdateOrCreate(long id, T entity)` to the `RealmAdapter`, but maybe it is not a good idea, I'm not sure right now.


A safer variant maybe would to swap the `Update`/`Create` calls - try to create and if it fails (== already exists) try updating. Then if there is a race between two threads, one or both of them will fail when calling `Create` and will call `Update` which will now succeed. Or maybe just adding a lock?

amulware: I don't like the idea of switching these since updating will be by far the more common case. Do we have any sort of indication of how much of a performance drain the catch is? I figure it's at the order of millisecond tops, and it happens in a background thread, so it's probably fine?

Toggl.PrimeRadiant.Realm/SinceParameterStorage.cs
simonrozsival: OK.

Toggl.PrimeRadiant.Realm/SinceParameterStorage.cs
simonrozsival: Yes, the type of the id in Realm is long.

Toggl.PrimeRadiant.Realm/SinceParameterStorage.cs
kornelijepetak: @amulware there's nothing to like 😄 It's about consistency.
The usual way to access dictionary is: `dict[key] = something`.
The same syntax is used in initializer. Like @simonrozsival said, get used to it!

Toggl.PrimeRadiant.Realm/SinceParameterStorage.cs
amulware: We _could_ use batch update this this as well. That would be safe, but potentially overkill?

Toggl.PrimeRadiant.Realm/SinceParameterStorage.cs
amulware: I disagree. `dict.Add(key, value)` is very valid, and can be safer if you don't like accidental key replacement.

kornelijepetak: Indeed, but it's not `dict.Add({key, value})`.

Toggl.PrimeRadiant.Realm/SinceParameterStorage.cs
simonrozsival: Let's not use `BatchUpdate` here please.

I agree that updating will be the more frequent case (creating only once, for every other sync just update, we never delete).

I don't think that the catch has any performance implications for us, but unless we measure it, we won't know for sure.

The only thing I would be "worried" about is a scenario when there is no record, `Set<T>(date)` is called twice in parallel, both threads fail with the `Update` (there is no record in the db...) so they both try to call `Create` - one succeeds, the other throws.

Toggl.PrimeRadiant.Realm/SinceParameterStorage.cs
amulware: You're right. What we really needed wasn't the silly syntax above, but support for value tuples:
```
new Dictionary<int, string>{
(0, "none"),
(1, "some")
}
```
Oh wait, we can simply add this as an extension method. :trollface:

simonrozsival: I like the syntax with `[key] = value`. It will stay anyway.

Toggl.PrimeRadiant.Realm/SinceParameterStorage.cs
amulware: Given that this should never happen (who else is touching this repository?), I'd say go for a simple lock then, just in case?

simonrozsival: It would never happen in this case, yes. But since this is just a half-generic implementation of a key-value storage, maybe we will later decide to refactor this and use it for a different purpose and that might see some heavier usage. It is hard to predict the future. I will add the lock (which is another try-catch block, so let's not worry about performance then :D)

PR #2002, opened 2018-05-02
Abstraction of API clients methods

simonrozsival:
This PR standardizes the way we interact with the API clients during sync - there are 5 specific cases:
- create an entity - `ICreatingApiClient`
- update an entity - `IUpdatingApiClient`
- delete an entity - `IDeletingApiClient`
- get all available entities - `IPullingApiClient`
- get all available entities changed since some date - `IPullingChangedApiClient`
- get single entity - `IPullingSingleApiClient`

Integration tests: https://www.bitrise.io/build/19b4156e1d2277f4


amulware:
One more review welcome, just in case.

amulware:
I have no issues with the code itself. But:

1. What about the names
```
ICreatingApiClient
IUpdatingApiClient
IDeletingApiClient
IPullingApiClient
IPullingChangedApiClient
```
OR, given the namespace of these interfaces:
```
ICreating
IUpdating
IDeleting
IPulling
IPullingChanged
```

2. I'm really not happy with adding the fake endpoints.

The reason they are added is so that foundation code can treat users and time entris the same.

But they fundamentally are not the same, and I'm not convinced they should be handled by the exact same code. I'm sort of worried that this will get us into trouble in the future, or be confusing, or... It feels off to me.

I would very much like other opinions on this matter.

@simonrozsival 
> I would argue that we should have a common interface for all entities which will be pulled

This is just a statement of preference. What is the actual argument? Why do you think this is better?

Toggl.Ultrawave.Tests.Integration/UserApiTests.cs
```diff
@@ -94,6 +94,40 @@ protected override IObservable<IUser> CallEndpointWith(ITogglApi togglApi)
             }
         }
 
+        public sealed class TheGetAllMethod : AuthenticatedEndpointBaseTests<List<IUser>>
+        {
+            protected override IObservable<List<IUser>> CallEndpointWith(ITogglApi togglApi)
+                => togglApi.User.GetAll();
+
+            [Fact, LogTestInfo]
+            public async Task AlwaysReturnsSingleEntityForAValidRequest()
```
amulware: If we keep these endpoints, ideally I'd like to have a test where another user is in the workspace of the user that we fetch, to make sure still only a single user is returned.

Though to be fair, it occurs to me that the endpoint does not even return a JSON array, and would probably horribly fail if it tried returning multiple things, right?

In either case, we have no workspace management endpoints, so this is too much work to do.

Toggl.Ultrawave/ApiClients/Interfaces/IClientsApi.cs
```diff
@@ -1,13 +1,11 @@
-﻿using System;
-using System.Collections.Generic;
-using Toggl.Multivac.Models;
+﻿using Toggl.Multivac.Models;
 
 namespace Toggl.Ultrawave.ApiClients
 {
     public interface IClientsApi
+        : IPullingApiClient<IClient>,
+            IPullingChangedApiClient<IClient>,
```
willsb: I'm twelve and what's this indentation?

Toggl.Foundation/Sync/PullingApiClientAdapter.cs
```diff
@@ -0,0 +1,23 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Reactive.Linq;
+using Toggl.Multivac;
+using Toggl.Ultrawave.ApiClients;
+
+namespace Toggl.Foundation.Sync
+{
+    public class PullingApiClientAdapter<T> : IPullingApiClient<T>
```
willsb: Why is this not sealed?

Toggl.Ultrawave/ApiClients/Interfaces/IClientsApi.cs
simonrozsival: That's the way R# does it. I think it is in line with the guidelines - ident when splitting a line of code - but I can align the interfaces and see how it looks.

Toggl.Foundation/Sync/PullingApiClientAdapter.cs
simonrozsival: 🙈

Issue #1891, opened 2018-04-14
Rewrite model generators


amulware:
They are not very readable. Let's try and do less string manipulation, and do all of it at the beginning of the file and then generate the code, possibly even using functions to encapsulate and name blocks of code.

Note that R# has a plugin that enables intellisense and pretty colours in tt files, which is nice.

arturs105:
Leaving this here for now (might be a separate issue): the `FoundationModels.tt` generator is broken due to `ICountry` model. It tries to generate `class Country : IDatabaseCountry`, but we have no `IDatabaseCountry`, which leads to a compile time error.

amulware:
There are no issues with reports models which are also not persisted, why is that?

arturs105:
Maybe it's been a while, since we regenerated the file?

amulware:
Or maybe they sit in a different folder in Multivac? :D

simonrozsival:
The problem with `Country` will be resolved when #1993 is merged. I would wait with any changes to the templates until this PR is merged - there were already significant changes to that file.

Odrakir:
We completely removed the generators (at least for now)

PR #2005, opened 2018-05-02
Fix T4 model generators

simonrozsival:
Ref #1891 

This PR fixes the T4 templates to output the same code as before (only the order of `At` and `ServerDeletedAt` now changed - this has no effect on the semantics ofc).

Squash whenever


Toggl.Multivac/Models/ISingleEntity.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Toggl.Multivac.Models
+{
+    public interface ISingleEntity : IIdentifiable
```
amulware: I'm still not liking this, and it doesn't seem to be necessary? Did I miss something? I'm not convinced that our API should care about the difference between single and multiple entities.

amulware: We could detect if the `Id` property has a setter in the generator? Not sure if that is much better..

simonrozsival: You wouldn't get this information from the interface file - you would have to open also the class implementation from Ultrawave and look for the setter there. I would say that is not a good and robust way of doing it.

One option would be to remove the `IIdentifiable` from the model after all and deal with this problem in the database layer (we would add the `IIdentifiable` interface explicitly to the `IDatabase*` interface and implement the `Id` with a getter only. I am still not sure if that is better. Definitely not nicer.

simonrozsival: Do you want me to investigate a different solution than the one that is in the code right now? Or is my argumentation enough?

amulware: I feel that the IDatabase solution is nicer, especially since that represents the semantics much more correctly. But I'm ok with merging this for now and we can look at it again down the line I think?

Toggl.Multivac/Models/ISingleEntity.cs
```diff
@@ -0,0 +1,6 @@
+﻿namespace Toggl.Multivac.Models
+{
+    public interface ISingleEntity
```
amulware: It still feels wrong to me to have this interface in shared code. The only thing it seems to do is generate fake ids in the database models, under the assumption that the database model will implement `IIdentifiable`? It seems inverted and redundant. What would we have to do to just get rid of this?

Toggl.Multivac/Models/ISingleEntity.cs
willsb: Stepping in a little latter, but would it be the case to have some different file describing our entities for our generators; e.g.: a json that describes with ids should be readonly, when applicable?

PR #2007, opened 2018-05-02
Sign up endpoint

### LINKED ISSUE #2004: Update signup endpoint
There are two new fields in POST data for signup endpoint:
- `country_id` - this field should contain the `integer` value of the id field from countries endpoint,
- `tos_accepted` - this field should contain whether the user accepted the TOS (`boolean`). In the future, this will make the endpoint fail if 'false' is passed, or if it is missing. However, I don't want to hardcode 'true', since it should be up to whoever writes the UI to consciously confirm that the user read the TOS.

Example JSON:
```
{
	"email": "emails@ex.com",
	"password": "pass",
	"workspace": {
		"initial_pricing_plan": 0,
		"name": "workspace 1"
	},
	"tos_accepted": true,
	"country_id": 156
}
```

Since both fields will be mandatory in the future, let me know if there are ideas for whether we should make them mandatory in the API somehow. I suppose, we also do not have other fields mandatory and leave it up to the API user right now, so maybe it's not necessary to make this more complicated.

In either case, I want integration tests that confirm that currently these fields _work_ but are _not mandatory_. Those tests will fail once the endpoints are made mandatory, which will be good for us to know, and we'll update the tests then.

juanlaube:
Closes #2004


Odrakir:
Looks fine to me, but I rather let someone with more experience check.

Toggl.Foundation/Login/LoginManager.cs
```diff
@@ -83,7 +83,7 @@ public IObservable<ITogglDataSource> SignUp(Email email, Password password)
 
             return database
                     .Clear()
-                    .SelectMany(_ => apiFactory.CreateApiWith(Credentials.None).User.SignUp(email, password))
+                    .SelectMany(_ => apiFactory.CreateApiWith(Credentials.None).User.SignUp(email, password, termsAccepted, countryId))
```
simonrozsival: This line starts being too long - could you maybe split this into two `Select`s or create a separate function for it? Or just splitting into several lines with some nice indenting would be enough 👍

Toggl.Foundation.MvvmCross/ViewModels/LoginViewModel.cs
```diff
@@ -62,6 +62,10 @@ public string Title
 
         public Password Password { get; set; } = Password.Empty;
 
+        public bool TermsAccepted { get; set; } = false;
+
+        public int CountryId { get; set; } = 0;
```
simonrozsival: Is 0 a valid country ID? What would happen if the user doesn't change anything in the app and hits sign up with this being 0?

simonrozsival: Maybe this could be nullable and it could be used to check if user selected some country or not.

Toggl.Ultrawave/ApiClients/UserApi.cs
```diff
@@ -44,7 +44,7 @@ public IObservable<string> ResetPassword(Email email)
                 .Select(instructions => instructions.Trim('"'));
         }
 
-        public IObservable<IUser> SignUp(Email email, Password password)
+        public IObservable<IUser> SignUp(Email email, Password password, bool termsAccepted, int countryid)
```
simonrozsival: Consider splitting this into several lines.

amulware:
No full review, but please also add integration tests for the country parameter seeing what values it allows, and of course, a test to validate that we not sending it right now does not fail yet.

Toggl.Ultrawave/ApiClients/UserApi.cs
```diff
@@ -44,7 +44,11 @@ public IObservable<string> ResetPassword(Email email)
                 .Select(instructions => instructions.Trim('"'));
         }
 
-        public IObservable<IUser> SignUp(Email email, Password password)
+        public IObservable<IUser> SignUp(Email email,
+            Password password,
+            bool termsAccepted,
+            int countryId
```
arturs105: I don't really like the way the arguments are split - the email argument should be on the next line be together with his argument friends.

Toggl.Foundation.MvvmCross/ViewModels/LoginViewModel.cs
simonrozsival: I don't think we will necessarily have the default valid ID - what if the user doesn't allow us to access his/her location? Then it will stay 0, right? Or can we always have reliable information about the user's country on both iOS and Android?

I like having a test which fails when backend changes the implementation, but I think we should implement our code in the way how it should be so we don't have to change it when backend changes it (which might also break things for users who don't update), we should just remove/change the integration test.

juanlaube: I confirmed with Heidi and we won't ask location permission, but we will use an endpoint that returns user location (based on IP address I guess). Anyways, that's async could take time and/or fail, so I made CountryId nullable. There are also plenty of integration tests that should fail when backend makes breaking changes.

Toggl.Ultrawave/ApiClients/UserApi.cs
juanlaube: Me neither, but I saw this style in other files ¯\_(ツ)_/¯.
I've moved the first argument to the next line.

Toggl.Foundation.MvvmCross/ViewModels/LoginViewModel.cs
```diff
@@ -241,6 +246,17 @@ private void validatePassword()
             RaisePropertyChanged(nameof(InfoText));
         }
 
+        private void validateCountryId()
+        {
+            IsErrorText = true;
```
simonrozsival: I don't understand this - this will always be true??

juanlaube: This changes the info text color, only in Android.

PR #2010, opened 2018-05-02
DateTime popup - UI Interaction

kornelijepetak:
Refs #1764 

⚠️ This should be reviewed commit by commit.

This PR deals only with the UI interactions (you will see some ignorable stuff in comments). Those are there only to provide functionality until full behaviour is implemented.

For review simplicity, I will keep adding commits to this branch until the behavior is done.

TODO: 

* [x] Duration shortcuts
* [x] Selecting tabs should reset each tab to starting position (i.e. hour selection)
* [x] Sometimes unknown behaviour crashes the app (need to check this) 
* [x] Tab's headers should update with changes/time
* [x] Height is not calculated correctly from the content so OK/Cancel buttons are not visible :/
* [x] Split start/stop when editing
* [x] Finish editing dates
* [x] Running entry's stop button
* [x] Open appropriate tab depending on where it's clicked
* [x] Hook commands and events
* [x] Duration editing


Toggl.Giskard/Resources/layout/SelectStopDateTimeEditorPage.axml
```diff
@@ -0,0 +1,171 @@
...
+    </RelativeLayout>
+    <View
+        android:layout_marginRight="13dp"
+        android:layout_marginLeft="13dp"
+        android:layout_marginTop="112dp"
+        android:background="@color/popupBlueBackground"
+        android:layout_width="match_parent"
+        android:layout_height="36dp" />
+    <TextView
+        local:MvxBind="Click StopTimeEntryCommand"
```
juanlaube: I think binding also needs `Visibility(!IsTimeEntryStopped)`

Toggl.Giskard/Resources/layout/SelectStopDateTimeEditorPage.axml
```diff
@@ -0,0 +1,171 @@
...
+            android:layout_height="wrap_content" />
+    </RelativeLayout>
+    <View
+        android:layout_marginRight="13dp"
+        android:layout_marginLeft="13dp"
+        android:layout_marginTop="112dp"
+        android:background="@color/popupBlueBackground"
+        android:layout_width="match_parent"
+        android:layout_height="36dp" />
+    <TextView
```
juanlaube: Shouldn't this be a button? This works, but doesn't have the ripple effect 🤔

juanlaube:
The stop button is always visible.
There's a lot of white space in the end page when TE is running.

Other than that, everything looks 👌 

screenshot_1525450953
screenshot_1525450986

Toggl.Giskard/Resources/layout/SelectStopDateTimeEditorPage.axml
daividssilverio: At the moment, you can keep clicking on the "Stop Timer" and it will keep updating the duration.

Toggl.Giskard/Fragments/SelectTimeFragment.cs
```diff
@@ -0,0 +1,206 @@
...
+        public override View OnCreateView(LayoutInflater inflater, ViewGroup container, Bundle savedInstanceState)
+        {
+            base.OnCreateView(inflater, container, savedInstanceState);
+            var view = this.BindingInflate(Resource.Layout.SelectTimeFragment, null);
+
+            controlButtons = view.FindViewById<LinearLayout>(Resource.Id.SelectTimeFragmentControlButtons);
+            pager = view.FindViewById<ViewPager>(Resource.Id.SelectTimeFragmentPager);
+            tabLayout = view.FindViewById<TabLayout>(Resource.Id.SelectTimeTabView);
+
+            IMvxBindingContextOwner bindingContext = (IMvxBindingContextOwner)this;
```
daividssilverio: It looks like you are not using this variable.

Toggl.Giskard/Adapters/SelectTimePagerAdapter.cs
```diff
@@ -0,0 +1,64 @@
...
+    public sealed class SelectTimePagerAdapter : PagerAdapter
+    {
+        private int[] resources =
+        {
+            Resource.Layout.SelectStartDateTimeEditorPage,
+            Resource.Layout.SelectStopDateTimeEditorPage,
+            Resource.Layout.SelectTimeDurationPage
+        };
+
+        private readonly Context context;
```
daividssilverio: This context is not being used.

Toggl.Giskard/Fragments/SelectTimeFragment.cs
```diff
@@ -0,0 +1,206 @@
...
+            onDurationChangedDisposable =
+                ViewModel.WeakSubscribe<PropertyChangedEventArgs>(nameof(ViewModel.Duration), onDurationChanged);
+
+            onStartTimeChangedDisposable =
+                ViewModel.WeakSubscribe<PropertyChangedEventArgs>(nameof(ViewModel.StartTime), onStartTimeChanged);
+
+            onStopTimeChangedDisposable =
+                ViewModel.WeakSubscribe<PropertyChangedEventArgs>(nameof(ViewModel.StopTime), onStopTimeChanged);
+
+            pager.SetCurrentItem(ViewModel.StartingTabIndex, false);
```
daividssilverio: What do you think about setting pager.OffscreenPageLimit to 2 so it won't re-instantiate the pages as the user moves through pages?

Toggl.Giskard/Fragments/SelectTimeFragment.cs
kornelijepetak: Refactoring debris.

Toggl.Giskard/Adapters/SelectTimePagerAdapter.cs
kornelijepetak: Deleted.

Toggl.Giskard/Fragments/SelectTimeFragment.cs
kornelijepetak: Good idea. Done.

juanlaube:
For some reason when I open the duration page, it's empty. However, if I go to the start or end pages and then back to duration, it renders correctly.

screenshot_1525711916

Toggl.Giskard/Adapters/SelectTimePagerAdapter.cs
```diff
@@ -0,0 +1,62 @@
...
+        private int[] resources =
+        {
+            Resource.Layout.SelectStartDateTimeEditorPage,
+            Resource.Layout.SelectStopDateTimeEditorPage,
+            Resource.Layout.SelectTimeDurationPage
+        };
+
+        private readonly IMvxAndroidBindingContext bindingContext;
+
+        public SelectTimePagerAdapter(Context context, IMvxBindingContext bindingContext)
```
daividssilverio: The context could be removed from the constructor too.

Toggl.Giskard/Resources/layout/SelectStopDateTimeEditorPage.axml
```diff
@@ -0,0 +1,173 @@
...
+            android:textSize="16sp"
+            android:layout_alignParentRight="true"
+            android:fontFamily="sans-serif-medium"
+            android:layout_alignParentBottom="true"
+            android:layout_marginBottom="20dp"
+            android:layout_marginRight="16dp"
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content" />
+    </RelativeLayout>
+    <FrameLayout
```
daividssilverio: If your intention using a FrameLayout with a Button inside was to keep both the ripple effect and the background colour, I would suggest creating and using a custom shape drawable having both the ripple and the background colour you desired.

Toggl.Giskard/Resources/layout/SelectStopDateTimeEditorPage.axml
```diff
@@ -0,0 +1,173 @@
...
+        local:MvxBind="Visibility Visibility(!IsTimeEntryStopped)"
+        android:background="@color/popupBlueBackground"
+        android:layout_marginRight="13dp"
```
daividssilverio: We should avoid using odd margins/sizes and prefer multiples of 4 or 2 to avoid pixel density problems. Android will round up ldpi and hdpi, causing layouts to look a little off for keen eyes in some cases.

kornelijepetak: Yes, it makes sense.
I was following a Zeplin design. If you think this is important (and it seems to be the case), maybe it would be good to inform @pe0ny and @lizaPlotnikova to make sure designs adhere to this.

pe0ny: In principle I'm ok with that but i would like to see the end result as well :)

kornelijepetak: In general I don't think difference of 1 would make that much of a visual difference. 😉

kornelijepetak:
I have moved the no-content bug to the issue #2051.
Since there's a workaround and to avoid stale issues, I'd like this re-reviewed and merged.

Toggl.Giskard/Resources/layout/SelectStopDateTimeEditorPage.axml
daividssilverio: I'm fine with refactoring later. One additional FrameLayout doesn't add to much, but is nice to keep things as flat as possible if we can do it.

PR #2014, opened 2018-05-03
Fix crash when selecting No project

### LINKED ISSUE #2013: App crashing for existing user when Starting a new time entry and selecting "No Project"
The app crashes when I start a new time entry from the start button,
Select a project and select "No project" from any workspace regardless of default or other.
The app crashes.

This behaviour does not change if I've selected a description or tag.

Repeated this several times on several clean installs with several accounts.

Accounts had multiple and single workspaces and regardless always crashed when I selected no project.

simonrozsival:
Closes #2013 

It is interesting that this wasn't a problem before.. 🤔 

Squash whenever


amulware:
Can we add a test for this?

simonrozsival:
You are right, there should be a test. I will add it ASAP.

simonrozsival:
@amulware There is now a test

amulware:
What I don't understand about the code is: It's a nullable id where both null and 0 indicate no project. Should this not be fixed so that it actually gets passed null? And 0 should not be an allowed value?

simonrozsival:
Project suggestions use `0` as "No project" ID since the beginning. I don't want to refactor that in this PR. Please create an issue to deal with this.

I am not sure why it worked until now (or did it?) and why it doesn't work now - I didn't notice any changes in the `StartTimeEntryViewModel` recently (maybe something related to Android?).

PR #2015, opened 2018-05-03
Improve api exception

### LINKED ISSUE #1983: Possible NullReferenceException in JSON api error handling
We recently changed our api error handling to support JSON errors for future prosperity. #1949 

I noticed that while we catch deserialisation exceptions in case we get malformed JSON, we do not catch a possible null-ref that can happen if the JSON is well-formed but does not include a `message` property:

https://github.com/toggl/mobileapp/commit/8dcd9ed51dd4a027b32a887efd89f822d1c863ee#diff-5c8eefd9b6948b1fc0ef82621e6e001dR49

We should probably catch this, or just `return error?.Message ?? "";` to cover all cases.

amulware:
Closes #1983

I encourage review by commit.


amulware:
Looks like my new tests caught a bug. 🤔

amulware:
At first, I was happy that I caught a bug. Now I'm not so sure anymore. Maybe we shouldn't be ignoring bad error JSON? I'm not really sure what the correct way is of handling malformed error information.

Other's thoughts are welcome.

Toggl.Ultrawave/Exceptions/ApiException.cs
```diff
@@ -36,26 +34,25 @@ public override string ToString()
                 + $"(Message: {message})";
 
         internal static string SerializeHeaders(IEnumerable<KeyValuePair<string, IEnumerable<string>>> headers)
-            => String.Join(", ", headers.Select(pair => $"'{pair.Key}': [{String.Join(", ", pair.Value.Select(v => $"'{v}'").ToArray())}]").ToArray());
+            => Join(", ", headers.Select(pair => $"'{pair.Key}': [{Join(", ", pair.Value.Select(v => $"'{v}'").ToArray())}]").ToArray());
```
kornelijepetak: Please name the inner Join and pull it out of the expression into a named variable. It is utterly unreadable.
`())}]")`?
C# should not look like that :)

Toggl.Ultrawave/Exceptions/ApiException.cs
```diff
@@ -48,10 +48,10 @@ private static string getLocalizedMessageFromResponse(IResponse response)
                 var error = JsonConvert.DeserializeObject<ResponseError>(response.RawData);
                 return error?.Message ?? badJsonLocalisedError;
             }
-            catch (JsonSerializationException)
-            {
-                return badJsonLocalisedError;
-            }
+            catch (JsonSerializationException) { }
```
kornelijepetak: btw. wasn't here from the getgo, why do we swallow exceptions here?

Toggl.Ultrawave/Exceptions/ApiException.cs
```diff
@@ -34,26 +36,22 @@ public override string ToString()
                 + $"(Message: {message})";
 
         internal static string SerializeHeaders(IEnumerable<KeyValuePair<string, IEnumerable<string>>> headers)
-            => String.Join(", ", headers.Select(pair => $"'{pair.Key}': [{String.Join(", ", pair.Value.Select(v => $"'{v}'").ToArray())}]").ToArray());
-
-        public override string Message => ToString();
+            => Join(", ", headers.Select(pair => $"'{pair.Key}': [{Join(", ", pair.Value.Select(v => $"'{v}'").ToArray())}]").ToArray());
```
kornelijepetak: Please name the inner Join and pull it out of the expression into a named variable. It is utterly unreadable.

Interpolated strings are great to have, but we should avoid that much work inside it. 
`())}]")`? C# should not look like that :)

Toggl.Ultrawave/Exceptions/ApiException.cs
```diff
@@ -34,26 +36,22 @@ public override string ToString()
...
-                }
-                catch (DeserializationException<ResponseError>)
-                {
-                    return "";
-                }
+                var error = JsonConvert.DeserializeObject<ResponseError>(response.RawData);
+                return error?.Message ?? badJsonLocalisedError;
             }
-            return response.RawData;
+            catch (JsonSerializationException) { }
```
kornelijepetak: I was not here from the get-go... why do we swallow these exceptions here?
So the whole hell doesn't break loose when backend changes the API all of a sudden?

amulware:
Suffice it to say, refactoring that formatting string led to me replacing the whole thing and just use proper (instead of pretend) JSON instead. It's a bit out of the scope of the issue/pr, so let me know if you want me to split this up.

Otherwise the discussion with regards to the `catch`ing remains, and I am so far none the wiser.

Toggl.Ultrawave/Exceptions/ApiException.cs
```diff
@@ -46,8 +45,7 @@ private static string getLocalizedMessageFromResponse(IResponse response)
 
             try
             {
-                var serializer = new JsonSerializer();
-                var error = serializer.Deserialize<ResponseError>(response.RawData);
+                var error = JsonConvert.DeserializeObject<ResponseError>(response.RawData);
```
willsb: Should we hard-depend on a library like that? I see tons of reasons for ditching Newtonsoft.Json. Using our own serializer should be the correct way

Toggl.Ultrawave/Exceptions/ApiException.cs
```diff
@@ -1,59 +1,56 @@
...
-                {
-                    var serializer = new JsonSerializer();
-                    var error = serializer.Deserialize<ResponseError>(response.RawData);
-                    return error.Message;
-                }
-                catch (DeserializationException<ResponseError>)
-                {
-                    return "";
-                }
+                var error = JsonConvert.DeserializeObject<ResponseError>(response.RawData);
```
willsb: I thought we had agreed on not taking any hard dependencies. I can see a few reason on why we'd want to ditch Newtonsoft.Json (e.g.: for using a faster, allocation free library). I think the previous way, using our own serializer abstraction, was better and also better aligned with the philosophy of our app.

amulware: That is fair enough. I'll re-add the abstraction asap.

amulware: While changing this back, thanks to the new tests I notice that our JsonSerializer is very leaky when it comes to exceptions. So the only way of fixing this properly might be editing that to catch exceptions from Json.NET.

amulware:
🦑

PR #2044, opened 2018-05-07
About page

### LINKED ISSUE #1450: About page
Part of #1331

- [x] Privacy Policy: https://toggl.com/legal/privacy
- [x] Terms of Service: https://toggl.com/legal/terms
- [x] Licenses #1172

### About page
toggl - ios - about 1x

Zeplin: https://zpl.io/aN4Z9Ad

dshr:
Closes #1450

Added the two extra faux-TableViewCell's and pointed them to the WebViews the same way we do in the login flow. Also DRYed things up a bit by moving the urls to Resources.resx 🙌


Toggl.Foundation/Resources.resx
```diff
@@ -156,9 +156,15 @@ time entries will appear here.</value>
     <data name="PrivacyPolicy">
         <value>Privacy Policy</value>
     </data>
+    <data name="PrivacyPolicyUrl">
```
kornelijepetak: I have mixed feelings about this. While I generally like extracting constants, I find `resources.resx` a bag of mutually unrelated reusable strings, and I'm not sure it's a good place for these URLs. Having them as `const strings` only at the place where they are used seems reasonable to me, as they are not really needed anywhere else.

But I would like a comment from @willsb or @simonrozsival before forcing changes or approving.

Toggl.Foundation.MvvmCross/ViewModels/Settings/AboutViewModel.cs
```diff
@@ -2,14 +2,17 @@
...
 
 namespace Toggl.Foundation.MvvmCross.ViewModels.Settings
 {
     [Preserve(AllMembers = true)]
     public sealed class AboutViewModel : MvxViewModel
     {
         private readonly IMvxNavigationService navigationService;
-
+      
+        public IMvxAsyncCommand PrivacyPolicyCommand { get; }
```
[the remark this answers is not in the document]
willsb: While I agree that this is not @dshr's fault, I still think he should make this change, for great justice

Toggl.Foundation/Resources.resx
dshr: Wouldn't we use different URLs for different languages too?

willsb: Touchè 😄 
I'm okay with that, but if you are using the Resources approach, make sure to change the url in other locations where it's being used (e.g.: the SignUpViewModel).

dshr: Already did that ;)

PR #2046, opened 2018-05-08
Fix visibility of the current first day of the week in the settings

### LINKED ISSUE #1906: Settings: first day of the week selection doesn't show selected option
The first day of the week selection works in principle but when opening the popup, th UI doesn't show me which one of the options is currently selected.

daividssilverio:
closes #1906 

🦑 whenever


Toggl.Foundation.MvvmCross/ViewModels/Selectable/SelectableBeginningOfWeekViewModel.cs
```diff
@@ -4,7 +4,7 @@
 
 namespace Toggl.Foundation.MvvmCross.ViewModels
 {
-    [Preserve]
+    [Preserve(AllMembers = true)]
```
kornelijepetak: How on earth did you detect this was a problem?

PR #2057, opened 2018-05-08
Track login and sign up errors

### LINKED ISSUE #1662: Track login/signup failures
I want to know how often login and signup fail, and for what reason.

Reasons I can think of right now:
- invalid email/password (login only)
- offline
- server error?
- missing api token in response (see #1652)

There are probably more reasons, feel free to add to the list. It should be easy to filter on app center to see what reasons correspond with signups and what reasons correspond with logins. Maybe two different events make sense, dunno.

juanlaube:
Closes #1662


Toggl.Foundation.MvvmCross/ViewModels/LoginViewModel.cs
```diff
@@ -420,6 +420,42 @@ private void onError(Exception exception)
                     InfoText = getGenericError();
                     break;
             }
+
+            trackError(exception);
```
simonrozsival: Let's split this in two methods - `trackLoginError` and `trackSignUpError` - move those `when IsSignup` to a if here. It could make it easier to understand what is going on and maintain - now the switch is huge and the cases are shuffled somehow.

Toggl.Foundation.MvvmCross/ViewModels/LoginViewModel.cs
```diff
@@ -420,6 +420,56 @@ private void onError(Exception exception)
...
+            {
+                trackSignUpError(exception);
+            }
+            else
+            {
+                trackLoginError(exception);
+            }
+        }
+
+        private void trackLoginError(Exception exception)
```
simonrozsival: Maybe this could just return the reason `LoginErrorSource` and you could have just one `analyticsService.TrackLoginErrorEvent` call. Same for sign-up.

PR #2060, opened 2018-05-08
Auto droid build number

### LINKED ISSUE #1882: Automate Giskard build numbers
Daneel already uses commit count as build number. Let's do the same for Giskard! Automation!

willsb:
Closes #1882 

I took the opportunity to already replace some the app name and the package identifier, just like in the iOS pipeline. In a future issue we can also replace the app icon, but that will be a little harder because of the adaptive icons.

Squash whenever


daividssilverio:
It seems ok for the Android part. I tried to run on the commands with the AdHoc and Playstore targets, but it ended up with an error, but I guess I'm missing some local configuration.
I guess it would become more bullet-proof if you use regular expressions to replace your strings.

amulware:
> replace some other things

Please list them. :P

willsb:
@daividssilverio: 
No regex pls

@amulware Updated

build.cake
```diff
@@ -266,6 +266,69 @@ private TemporaryFileTransformation GetIosInfoConfigurationTransformation()
     };
 }
 
+private TemporaryFileTransformation GetAndroidManifestTransformation()
+{
+	//Toggl.Giskard/Startup/SplashScreen.cs
```
kornelijepetak: Some ➡️ here... and also, a comment?

build.cake
```diff
@@ -266,6 +266,69 @@ private TemporaryFileTransformation GetIosInfoConfigurationTransformation()
...
+    }
+
+    var filePath = GetFiles(path).Single();
+    var file = TransformTextFile(filePath).ToString();
+
+    return new TemporaryFileTransformation
+    { 
+        Path = path, 
+        Original = file,
+        Temporary = file.Replace("987654321", commitCount)
```
kornelijepetak: Maybe name this constant and put it close to other consts here. Would make more sense to know why this specific number. (I know, it's not difficult to understand, but still.)

build.cake
willsb: :eyes:

build.cake
willsb: Agreed

Issue #2040, opened 2018-05-07
Time-date-duration popup fixes/improvements


pe0ny:
Remaining features to implement:

* [x] Preventing selection in controls (min max dates etc)
* [x] Consolidate header colors between tab header and controls
* [x] Apply user's settings for formatting dates and time
* [x] Integrate popup with the start view as well
* [x] Prevent one second initial delay

kornelijepetak:
Refs #2010

pe0ny:
**Start/stop content is empty when opening Duration tab first**
1. Open the edit view and tap on the Duration field -> Popup opens
2. Tap either Start or Stop tap -> Content is missing
After tapping Start/Stop again, the content appears.

kornelijepetak:
This has already been reported [here](https://github.com/toggl/mobileapp/pull/2010#issuecomment-386738784).

pe0ny:
**"Stop timer" button doesn't disappear**

1. Open the edit view when entry is running
2. Tap "Set" -> Popup opens
3. Tap "Stop timer" button -> Time entry is stopped but the button is still there though it shouldn't be.

Note: The button is visible even for already stopped entries.

pe0ny:
☝️ And that one too. Nvm me then.

kornelijepetak:
☝️ That one has already been fixed 😄

PR #2062, opened 2018-05-09
Prevent one second initial delay

kornelijepetak:
Fixes:

* [x] Prevent one second initial delay

of #2040.

Now, the tab headers should show immediately, not only after a time service provides a current time.

🦑 at your earliest convenience.


Toggl.Giskard/Fragments/SelectTimeFragment.cs
```diff
@@ -96,11 +96,28 @@ public override View OnCreateView(LayoutInflater inflater, ViewGroup container,
             onStopTimeChangedDisposable =
                 ViewModel.WeakSubscribe<PropertyChangedEventArgs>(nameof(ViewModel.StopTime), onStopTimeChanged);
 
+            initializeTabHeaders(ViewModel.StartingTabIndex);
```
willsb: Why is this needed even after the `StartWith`?

kornelijepetak: Because the `Initialize` method on the VM is called before `OnCreateView` here, so the weak subscription will not fire until there's a change in time after the `OnCreateView` has run. But before that, initial values have to be set up correctly.

Toggl.Foundation.MvvmCross/ViewModels/SelectTimeViewModel.cs
```diff
@@ -251,6 +252,7 @@ public override void Prepare(SelectTimeParameters parameter)
             await base.Initialize();
 
             timeService.CurrentDateTimeObservable
```
willsb: You should hold a reference to this observable

kornelijepetak:
This PR must be postponed to after the #2085 is merged as the core things changed. This way, most of the requested changes will simply not be valid and this issue will be able to be merged.

PR #2066, opened 2018-05-09
Resurect syncing

simonrozsival:
The most important part of this PR is reverting the tripple 🔥 commit. It will be better to merge all the changes into develop soon - before starting to rewrite the whole state machine. The rest of the commits just make the code work - fixing some renaming and dropped classes.

Review this PR commit-by-commit.

A note: there is a weird property `bool canUpdateSinceDates()` in `BasePersistState` which works well, but is not nice - in the very near future I would like to have the state responsible for both fetching, persisting, and updating the since date (if the entity is `ILastChangedDatable`) - which would be split into two states which could be composed together.

Squash whenever


Toggl.Foundation/Sync/States/Pull/BasePersistState.cs
```diff
@@ -64,14 +61,24 @@ private IObservable<ITransition> processError(Exception exception)
         private bool shouldRethrow(Exception e)
             => e is ApiException == false || e is ApiDeprecatedException || e is ClientDeprecatedException || e is UnauthorizedException;
 
+        private void maybeUpdateSinceDates(IEnumerable<TDatabaseInterface> entities)
+        {
+            var datableAndSyncableEntities = entities
+                .OfType<IDatabaseSyncable>()
+                .OfType<ILastChangedDatable>()
+                .ToList();
```
amulware: Maybe not a major issue, but I'd probably prefer accumulating the max date without creating a copy of the collection(use nullable or something for the condition above). That being said, why doesn't the method interface work with an IList (or at least ICollection)? Wouldn't it be easier to pass those around, instead of enumerables?

Toggl.Foundation/Sync/States/Pull/BasePersistState.cs
```diff
@@ -80,5 +82,13 @@ private void maybeUpdateSinceDates(IEnumerable<TDatabaseInterface> entities)
...
 
         protected abstract TDatabaseInterface ConvertToDatabaseEntity(TInterface entity);
+
+        private bool canUpdateSinceDates()
+            => typeof(TDatabaseInterface) is IDatabaseClient
+                || typeof(TDatabaseInterface) is IDatabaseProject
+                || typeof(TDatabaseInterface) is IDatabaseTag
+                || typeof(TDatabaseInterface) is IDatabaseTask
+                || typeof(TDatabaseInterface) is IDatabaseTimeEntry
+                || typeof(TDatabaseInterface) is IDatabaseWorkspace;
```
amulware: Why not check if the type implements `ILastChangedDatable` instead?

Toggl.Foundation/Sync/States/Pull/BasePersistState.cs
```diff
@@ -63,15 +63,17 @@ private bool shouldRethrow(Exception e)
 
         private void maybeUpdateSinceDates(IEnumerable<TDatabaseInterface> entities)
         {
-            var datableAndSyncableEntities = entities
-                .OfType<IDatabaseSyncable>()
-                .OfType<ILastChangedDatable>()
-                .ToList();
-
-            if (datableAndSyncableEntities.Count > 0)
+            if (canUpdateSinceDates())
```
amulware: Why do this?

Toggl.Foundation/Sync/States/Pull/BasePersistState.cs
simonrozsival: That is not the same thing - we don’t store since dates for all entities which implement that interface (e.g., user).

Toggl.Foundation/Sync/States/Pull/BasePersistState.cs
simonrozsival: Because we don’t store the since date for all of the entity types - just for some of them.

Toggl.Foundation/Sync/States/Pull/BasePersistState.cs
simonrozsival: Ok, I can polish this bit of the code 👍

Toggl.Foundation/Sync/States/Pull/BasePersistState.cs
amulware: Understood. I'm not a big fan of having that information in multiple places (it's somewhere in the repository/database as well, right?). Maybe we can abstract that somewhere?

PR #2072, opened 2018-05-09
Threadsafe models and refactored data sources

simonrozsival:
This PR abstracts a lot of the code in the data sources. A data source is now a wrapper around a database which makes it easier and safer to manipulate with the DB data from the UI.
- it handles the conversion between the database (Realm) and the Foundation models and makes it invisible for the business logic (this will be useful in the syncing states and it will prevent any future crashes we had due to accessing Realm objects from wrong threads)
- it abstracts the way we can observe the changes in the data over time (as we do with time entries and preferences) through the `ObservableDataSource` interface/class
- it hides the conflict/rivals resolution from the business logic (e.g., the syncing states)

_One thing worth noting:_ The syncing states now expect to be given an `IRepository` - none of the data sources implement this interface anymore so I changed the parameters passed to the syncing states in the `TogglStateManagerFactory` - we now pass the database repositories instead of the data sources to the time entries' states and the preferences' states.

This will be definitely changed in the next PR - the syncing states shouldn't touch the database directly, they should work with `IDataSource<T>` everywhere.

Squash whenever


simonrozsival:
Two more things for the reviewers:
- we used a `Delete` method in the TEs' source to soft-delete TEs from the UI. This was now weird and confusing - because we need `Delete` to in fact hard delete the TE from the repository. I added a `SoftDelete` method to the DS an interactor so nobody from the devs confuses the soft deletion with hard deletion
- our old interactors still use the `IDatabase*` interfaces and I didn't go into refactoring this. And probably not only interactors. This PR was already big, so I didn't change any occurrence of the DB interfaces to the threadsafe one if it wasn't necessary to make the app work

Toggl.Foundation.Tests/MvvmCross/ViewModels/TimeEntriesLogViewModelTests.cs
```diff
@@ -390,18 +391,19 @@ private TimeEntryViewModel createTimeEntryViewModel()
 
         public sealed class TheDeleteCommand : TimeEntriesLogViewModelTest
         {
-            [Property]
-            public void DeletesTheTimeEntry(long id)
+            [Fact]
```
willsb: What's the reasoning for this?

Toggl.Foundation/DataSources/DataSource.cs
```diff
@@ -0,0 +1,66 @@
...
+
+namespace Toggl.Foundation.DataSources
+{
+    public abstract class DataSource<T, U> : IDataSource<T, U>
+        where U : IDatabaseSyncable
+        where T : IThreadsafeModel, IIdentifiable, U
+    {
+        protected readonly IRepository<U> Repository;
+
+        protected virtual IRivalsResolver<U> RivalsResolver { get; } = null;
```
willsb: Why must this be here? 🤔

Toggl.Foundation/DataSources/Interfaces/IObservableDataSource.cs
```diff
@@ -0,0 +1,18 @@
...
+namespace Toggl.Foundation.DataSources.Interfaces
+{
+    public interface IObservableDataSource<TThreadsafe, TDatabase>
+        : IDataSource<TThreadsafe, TDatabase>
+        where TDatabase : IDatabaseSyncable
+        where TThreadsafe : IThreadsafeModel, TDatabase
+    {
+        IObservable<TThreadsafe> Created { get; }
+        
+        IObservable<(long Id, TThreadsafe Entity)> Updated { get; }
```
willsb: I don't like using Tuples in interfaces. This should be a class/struct of its own. Maybe even an `Either`, given that the fields are not always needed

Toggl.Foundation/DataSources/ObservableDataSource.cs
```diff
@@ -0,0 +1,76 @@
...
+        
+        protected ObservableDataSource(IRepository<TDatabase> repository)
+            : base(repository)
+        {
+            Created = CreatedSubject.AsObservable();
+            Updated = UpdatedSubject.AsObservable();
+            Deleted = DeletedSubject.AsObservable();
+        }
+
+        public override IObservable<TThreadsafe> Create(TThreadsafe entity)
```
willsb: Maybe seal these guys?

Toggl.Foundation/Toggl.Foundation.csproj
```diff
@@ -58,4 +58,7 @@
   <ItemGroup>
     <Folder Include="Analytics\" />
   </ItemGroup>
+  <ItemGroup>
+    <Service Include="{508349b6-6b84-4df5-91f0-309beebad82d}" />
```
willsb: 👀

Toggl.Foundation/DataSources/DataSource.cs
simonrozsival: For batch update.

Toggl.Foundation/DataSources/ObservableDataSource.cs
simonrozsival: Hmm... I am not sure if we should do that in an abstract class. I think this should stay unsealed for future extensions.

Toggl.Foundation/Toggl.Foundation.csproj
simonrozsival: 🙈

simonrozsival:
I don't think the `SoftDelete` method could be an extension method - I need to access the repository directly here. It might be refactored somehow, but I don't think we have to do that right now.

I agree that we should go through the data sources and remove any specific code to interactors and end up with just generic data sources.

simonrozsival:
Right now we can't simply change the interfaces from public to internal - we need the data sources in Foundation.MvvmCross. I don't think we should seal them and deal with all the problems related to it in this PR. Please create an issue and we can deal with it later.

PR #2078, opened 2018-05-09
Get all non deleted time entries interactor

simonrozsival:
Squash whenever


Toggl.Foundation/Interactors/TimeEntry/ContinueMostRecentTimeEntryInteractor.cs
```diff
@@ -14,32 +14,34 @@ namespace Toggl.Foundation.Interactors
 {
     public class ContinueMostRecentTimeEntryInteractor : IInteractor<IObservable<IDatabaseTimeEntry>>
     {
+        private readonly IInteractor<IObservable<IEnumerable<IThreadsafeTimeEntry>>> getNonDeleted;
```
willsb: Here I would inject the data source instead.

simonrozsival: Hmm, why so?

simonrozsival: It seemed nice to me to use an interactor from within an interactor. Resembles composition a bit.

willsb: Because then I would have the full understanding of how the interactor works directly in the interactor, without having to look somewhere else. This cross-dependency of interactors might bite us in the rear if it goes out of hand, so I'd rather avoid it.

This will also simplify the dependencies of this interactor, so win-win

PR #2084, opened 2018-05-09
Ask users to rate us in the app

### LINKED ISSUE #1972: Ask users to rate us in the app
## A/B Testing
We would like to ABCD test when is the best time to ask rating from the users.

~a) Show this pop to users who have at least 10 time entries in total over the course of 5 distinct days & trigger the popup when user is on the Timer page & stops the time entry by tapping the red Stop button.~
~b) Show this pop to users who have at least 10 time entries in total over the course of 5 distinct days & trigger the popup when user is on the Timer page & the TE is currently running.~
~c) Show this pop to users who have at least 20 time entries in total & have used the app at least 14 days & trigger the popup when user is on the Timer page & stops the time entry by tapping the red Stop button.~
~d) Show this pop to users who have at least 20 time entries in total & have used the app at least 14 days & trigger the popup when user is on the Timer page & the TE is currently running.~

~👆 This part is WIP, so feel free to suggest ideas. Main thing we should keep in mind is that we should target users who are actually a bit more familiar with the app and use it. We don't want to annoy first-second day users who are very much in the discovering mode.~

### 📣 Update

We want to test when is the best time to ask feedback/rating from the users and for that we came up with **3 different triggers**:
1) Show the popup when user is on the Timer page & stops the time entry by tapping the red Stop button.
2) Show the popup when user is on the Timer page & continues any TE (either using a swipe to continue or tapping the continue button on a log or the suggestion item).
3) Show the popup when user is on the Timer page & started a new time entry (comes from the Start view and did confirm the changes there).

In addition to these ☝️ We want to test different time periods how long user has been using the app:
a) after 5 days
b) after 10 days
☝️ These will be managed in Firebase and we can change those if needed.

### 💡Other things to keep in mind
- [ ] Don’t block the UI. Meaning that user can continue his flow (starting a new entry, go to reports etc) if he wants to.
- [ ] When user hasn’t interacted with the card in a 5 min, hide it and don't show it anymore.
       - When the app goes into background and user comes back before 5 mis has passed, he/she can still see the popup. After 5 min, the popup will be hidden.
      - When user closes the app completely, the popup will be hidden despite the time passed.

### ❓Do we ever wanna show it again?
In the future we might wanna show it again and for that let's store locally: Which one of the 7 options user ended up with and when.

**7 options:**
- No iteraction
- Yes
- No
- Rating: yes
- Rating: no
- Feedback: yes
- Feedback: no

## Flow
Enjoying -> Yes -> Would you mind rating on the App Store? -> Yes -> App Store popup i.e. sytem rating prompt https://developer.apple.com/ios/human-interface-guidelines/system-capabilities/ratings-and-reviews#system-rating-and-review-prompts
Enjoying -> Yes -> Would you mind rating on the App Store? -> No -> Close the popup
Enjoying -> No -> Would you mind giving us some feedback? -> Yes -> Open feedback form
Enjoying -> No -> Would you mind giving us some feedback? -> No -> Close the popup

## UI notes
- [ ] Display the Rating card below the suggestions (if there are any) or just at the top of the page below the nav bar if there are no suggestions.
- [ ] Padding from the Suggestions/Nav bar: 24pt.
- [ ] Padding from the Log: 56pt for the first popup (step1) - when second/third one appears, the log content remains to the same position.

### Step1: with and without suggestions
toggl - ios - timer - rate - step1 1x toggl - ios - timer - rate - step1 no suggestions 1x

Zeplin: https://zpl.io/VDOWvzr & https://zpl.io/brrL7E5

### Step2: v1 When user said "Yes" in the step1
toggl - ios - timer - rate - step2 rating 1x

Zeplin: https://zpl.io/VkmeNnM

### Step2: v2 When user said "No" in the step1
toggl - ios - timer - rate - step2 feedback 1x

Zeplin: https://zpl.io/VxkwJ1m

### LINKED ISSUE #1986: Ask users to rate us in the app: track events
Track events when we ask users to rate us in the app #1972
- [ ] Track how many users see the popup (step1)
- [ ] Track how many users tap "Yes" in step1
- [ ] Track how many users tap "Not really" in step1
- [ ] Track how many users tap agree to rate us in step2 (rating)
- [ ] Track how many users tap disagree "No, thanks" to rate us in step2 (rating)
- [ ] Track how many users give the rating (sytem rating prompt)
- [ ] Track how many users tap agree to give us feedback in step2 (feedback)
- [ ] Track how many users tap disagree "No, thanks" to give us feedback in step2 (feedback)

dshr:
Closes #1972
Closes #1986


arturs105:
This PR is now ready for review. If you feel brave, please test out that you receive the remote config values from firebase (must be done on an iPhone). Message me for instructions on how to do that.

Toggl.Daneel/Services/RatingService.cs
```diff
@@ -0,0 +1,13 @@
+﻿using StoreKit;
+using Toggl.Foundation.Services;
+
+namespace Toggl.Daneel.Services
+{
+    public sealed class RatingService : IRatingService
+    {
+        public void AskForRating()
+        {
+            SKStoreReviewController.RequestReview();
```
juanlaube: ⚠️ Oh my, you're gonna love this @arturs105... ⚠️ 
This method to request ratings and reviews is limited to three times every 365 days [see docs](https://developer.apple.com/documentation/storekit/skstorereviewcontroller/requesting_app_store_reviews). We need some logic to decide when to show the default alert from Apple or some custom view to redirect the users to the App Store.

arturs105: I knew that the amount of times we can ask for review is limited, though there is no way to tell if the rating view was actually shown or how many times have we shown it. But it shouldn't really matter, since we're showing our rating view only once.

Toggl.Foundation.Tests/MvvmCross/Services/FeedbackServiceTests.cs
```diff
@@ -0,0 +1,126 @@
...
+                        subject,
+                        Arg.Any<string>())
+                   .Wait();
+            }
+
+            [Fact, LogIfTooSlow]
+            public async Task SendsAnEmailWithAppVersionPhoneModelAndOsVersion()
+            {
+                platformConstants.PhoneModel.Returns("iPhone Y");
+                platformConstants.OperatingSystem.Returns("iOS 4.2.0");
```
juanlaube: I see what you did here 😂

arturs105: Is this a change request? 😄

Toggl.Foundation/Experiments/RatingViewExperiment.cs
```diff
@@ -0,0 +1,83 @@
...
+        private readonly ITimeService timeService;
+        private readonly ITogglDataSource dataSource;
+        private readonly IOnboardingStorage onboardingStorage;
+        private readonly IRemoteConfigService remoteConfigService;
+
+        public IObservable<bool> RatingViewShouldBeVisible
+            => remoteConfigService
+                .RatingViewConfiguration
+                .SelectMany(criterionMatched)
+                .Select(tuple => tuple.criterionMatched && dayCountPassed(tuple.configuration));
```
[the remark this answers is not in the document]
arturs105: Could be useful in the future, though we're currently showing the rating view once. Pinging @pe0ny and @amulware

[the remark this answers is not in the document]
amulware: We'll be storing the last time, and over a couple of months or whatever limit we'll have, there will definitely be updates to the app. So I don't think this is necessary.

Toggl.Foundation.MvvmCross/ViewModels/MainViewModel.cs
```diff
@@ -224,6 +249,34 @@ public void Init(string action)
                     await stopTimeEntry();
                     break;
             }
+
+            ratingViewExperimentDisposable = ratingViewExperiment
+                .RatingViewShouldBeVisible
+                .Subscribe(presentRatingViewIfNeeded, disposeRatingViewExperiment);
```
Odrakir: I don't understand why you are doing this. Why do you need to dispose the `ratingViewExperiment`? Why not just use a disposebag as usual?

Toggl.Foundation/DataSources/TimeEntriesDataSource.cs
```diff
@@ -56,8 +68,13 @@ internal sealed class TimeEntriesDataSource
                     .Merge(Created)
                     .SelectMany(_ => GetAll(te => te.IsDeleted == false))
                     .Select(timeEntries => !timeEntries.Any());
-            
+
             RivalsResolver = new TimeEntryRivalsResolver(timeService);
+
+            TimeEntryStarted = timeEntryStartedSubject.AsObservable();
```
Odrakir: Isn't moving this up to the declaration (with => ) clearer?

Toggl.Foundation.MvvmCross/ViewModels/MainViewModel.cs
```diff
@@ -224,6 +249,34 @@ public void Init(string action)
...
+        }
+
+        private void presentRatingViewIfNeeded(bool shouldBevisible)
+        {
+            if (shouldBevisible)
```
daividssilverio: You could do an early return here.

Toggl.Foundation.MvvmCross/ViewModels/RatingViewModel.cs
```diff
@@ -0,0 +1,135 @@
...
+        }
+
+        private void registerImpression(bool isPositive)
+        {
+            GotImpression = true;
+            impressionIsPositive = isPositive;
+            analyticsService.UserFinishedRatingViewFirstStep.Track(isPositive);
+            if (isPositive)
+            {
+                CtaTitle = Resources.RatingViewPositiveCTATitle;
```
daividssilverio: Sorry, but what is a CTA?

Toggl.Foundation.MvvmCross/ViewModels/RatingViewModel.cs
```diff
@@ -0,0 +1,135 @@
...
+                CtaTitle = Resources.RatingViewNegativeCTATitle;
+                CtaDescription = Resources.RatingViewNegativeCTADescription;
+                CtaButtonTitle = Resources.RatingViewNegativeCTAButtonTitle;
+                onboardingStorage.SetRatingViewOutcome(RatingViewOutcome.NegativeImpression, timeService.CurrentDateTime);
+            }
+        }
+
+        private async Task performMainAction()
+        {
+            if (impressionIsPositive == null)
```
daividssilverio: This should be a in a single line.

Toggl.Foundation.MvvmCross/ViewModels/RatingViewModel.cs
juanlaube: Call to action. Maybe the name is not good 🤔

Toggl.Foundation.MvvmCross/ViewModels/RatingViewModel.cs
```diff
@@ -0,0 +1,135 @@
...
+            }
+        }
+
+        private void dismiss()
+        {
+            navigationService.ChangePresentation(
+                new ToggleRatingViewVisibilityHint(forceHide: true)
+            );
+
+            if (impressionIsPositive.HasValue)
```
daividssilverio: Early return over nested ifs.
```
if (!impressionIsPositive.HasValue) return;
```

Toggl.Daneel/ViewControllers/MainViewController.cs
```diff
@@ -313,6 +337,35 @@ public override void ViewDidLayoutSubviews()
...
+            ratingView.ConstrainInView(ratingViewContainer);
+        }
+
+        public void HideRatingView()
+        {
+            ratingView.RemoveFromSuperview();
+            ratingView.Dispose();
+            ratingView = null;
+
+            //We have to scroll a little to update the header size.
```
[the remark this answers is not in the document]
arturs105: I'd also prefer this approach, however, I just tried this out and it's not working, even with calling `SetNeedsLayout()` and `LayoutIfNeeded()` for the header view.

Toggl.Foundation.MvvmCross/ViewModels/RatingViewModel.cs
daividssilverio: I thought it was something "customer t???? action" sort of thing. Or some rating specific term.

Toggl.Foundation/DataSources/TimeEntriesDataSource.cs
arturs105: It may be clearer to read, but then it would work a bit different - every time that somebody would use those properties, the `subject.AsObservable()` code would be executed, however currently the `AsObservable` gets called once per observable and the result is stored in a variable and that variable is returned when using the properties.

Unless C# has some smart optimisation that reduces both approaches to the same IL code.

Toggl.Foundation.MvvmCross/ViewModels/MainViewModel.cs
```diff
@@ -260,8 +260,8 @@ private void presentRatingViewIfNeeded(bool shouldBevisible)
         {
             if (shouldBevisible)
             {
-                var previousRatingOutcome = onboardingStorage.RatingViewOutcome();
-                if (previousRatingOutcome != null) return;
+                //var previousRatingOutcome = onboardingStorage.RatingViewOutcome();
```
daividssilverio: why the comments?

daividssilverio: :face_palm:

Toggl.Daneel/Services/RemoteConfigService.cs
```diff
@@ -0,0 +1,38 @@
...
+namespace Toggl.Daneel.Services
+{
+    public sealed class RemoteConfigService : IRemoteConfigService
+    {
+        public IObservable<RatingViewConfiguration> RatingViewConfiguration
+            => Observable.Create<RatingViewConfiguration>( observer =>
+            {
+                observer.OnNext(new Multivac.RatingViewConfiguration(0, "stop"));
+                observer.OnCompleted();
+            return Disposable.Empty;
```
daividssilverio: 🤔 This makes the code bellow unreachable, right? Was it intentional?
Also the indentation is wrong.

Toggl.Daneel/Services/RemoteConfigService.cs
```diff
@@ -0,0 +1,38 @@
...
+using Toggl.Multivac;
+
+namespace Toggl.Daneel.Services
+{
+    public sealed class RemoteConfigService : IRemoteConfigService
+    {
+        public IObservable<RatingViewConfiguration> RatingViewConfiguration
+            => Observable.Create<RatingViewConfiguration>( observer =>
+            {
+                observer.OnNext(new Multivac.RatingViewConfiguration(0, "stop"));
```
daividssilverio: If you have a enum for this "stop" here, why are you not using it? `RatingViewCriterion.Stop`

Toggl.Giskard/Services/RemoteConfigService.cs
```diff
@@ -0,0 +1,13 @@
...
+using System.Reactive.Linq;
+using Toggl.Foundation.Services;
+using Toggl.Multivac;
+
+namespace Toggl.Giskard.Services
+{
+    public class RemoteConfigService : IRemoteConfigService
+    {
+        public IObservable<RatingViewConfiguration> RatingViewConfiguration { get; }
+            = Observable.Return(new RatingViewConfiguration(int.MaxValue, "none"));
```
daividssilverio: Same here, why not using the enum directly?

Toggl.Daneel/Services/RemoteConfigService.cs
arturs105: This constructor isn't meant to be used like this. Actual usage is a few lines below - i just pass whatever i got from firebase, but asking for the enum in the constructor might be a better idea.

Toggl.Foundation/DataSources/TimeEntriesDataSource.cs
Odrakir: Of course, I keep forgetting this is not Swift. How about using `get { ... }`.

I'm proposing this because these observables are only exposing the subjects without doing anything else, so it feels cleaner to assign them in the definition. But I'm ok with this approach too.

PR #2085, opened 2018-05-10
Use user date/time format settings & co.

kornelijepetak:
⚠️ Easier to review commit by commit, but feel free to be brave otherwise.

Fixes

* [x] Apply user's settings for formatting dates and time

of #2040 

Also, fixes:

* [x] Running time entry now shows correct dialog size when `Stop` tab is shown
* [x] `Stop` copy is used instead of `Set` to mimic iOS version
* [x] Correct icons are used on the popup

PR #2086, opened 2018-05-10
Fix no-content bug in popup

### LINKED ISSUE #2051: Content is missing in DateTime Popup
For some reason when pages are first opened, they are empty. However, once a tab has changed (once or sometimes a few times is needed), the views show correctly.

This to some extent depends on the tab which was opened first.

screenshot_1525711916

kornelijepetak:
For simplicity of the review, I have based this PR in `popup-user-datetime-format` which needs to be merged ASAP. It will be rebased onto develop when #2085 is merged.

Otherwise,
Closes #2051


Toggl.Giskard/Fragments/SelectTimeFragment.cs
```diff
@@ -68,21 +68,25 @@ public override View OnCreateView(LayoutInflater inflater, ViewGroup container,
...
             onModeChangedDisposable =
                 ViewModel.WeakSubscribe<PropertyChangedEventArgs>(nameof(ViewModel.IsCalendarView), onIsCalendarViewChanged);
 
-            var startPageView = this.BindingInflate(Resource.Layout.SelectDateTimeStartTimeTabHeader, null);
-            tabLayout.GetTabAt(StartTimeTab).SetCustomView(startPageView);
+            tabLayout.Post(() =>
+            {
+                tabLayout.SetupWithViewPager(pager, true);
+
+                var startPageView = this.BindingInflate(Resource.Layout.SelectDateTimeStartTimeTabHeader, null);
```
willsb: Why not inflate these outside the post? Can you check if there's any perceivable different doing it the way I mention?

kornelijepetak: Partially. `SetupWithViewPager` needs to be Posted, and TabLayout will not yield any items (in `GetTabAt`) before it's been setup with the ViewPager, so at least the `SetCustomView` calls must be in Post. But for the process of inflation, you may be right, I don't think it'd makes a difference. I will try and change to make sure.

PR #2100, opened 2018-05-10
Genreic syncing states: Pull

simonrozsival:
Ref #1309 

This PR needs some more changes in the datasources:
- not all entities are IDatabaseSyncable, but I need some common ancestor to work with them easily, so I added `IDatabaseModel`
- the threadsafe interfaces are now partial - the generated interfaces can be extended manually (a use-case: `IThreadsafeWorkspaceFeatureCollection`)
- it is necessary to access a fetch observable in a similar way we do for since parameters - by the type - but the implementation is ugly :-/ suggestions for improvement are welcome!
- I selected tags as an entity for implementing the test for the generic state. I wanted to use the `TestModel` but there was a problem because of the `FetchObservables` which require using one of the real entities. We can have a discussion about this if you want.

Squash whenever


Toggl.Foundation/Sync/States/Pull/FetchObservables.cs
```diff
@@ -1,27 +1,30 @@
...
 using System.Collections.Generic;
+using System.Reactive.Linq;
 using Toggl.Multivac.Models;
 using Toggl.PrimeRadiant.Models;
 
 namespace Toggl.Foundation.Sync.States
 {
     internal sealed class FetchObservables
-    {
+    {
```
amulware: 👁

Toggl.Foundation/Sync/States/Pull/FetchObservables.cs
```diff
@@ -45,5 +47,37 @@ IObservable<IPreferences> preferences
             Tasks = tasks;
             Preferences = preferences;
         }
+
+        public IObservable<List<T>> GetByType<T>()
+        {
+            if (typeof(T) == typeof(IWorkspace))
+                return (IObservable<List<T>>)Workspaces;
```
amulware: I don't have a better idea right now, but as you say, it's a bit gross. :/

amulware: Only other thing I have would be sticking the Observables into a `Dictionary<Type, Observable>` but not sure that's much better.

Toggl.Foundation/Sync/States/Pull/PersistState.cs
```diff
@@ -2,56 +2,56 @@
...
         public IObservable<ITransition> Start(FetchObservables fetch)
-        {
-            return FetchObservable(fetch)
+            => fetch.GetByType<TInterface>()
                 .SingleAsync()
-                .Select(entities => entities ?? new List<TInterface>())
-                .Select(entities => entities.Select(ConvertToDatabaseEntity).ToList())
-                .SelectMany(databaseEntities =>
-                    repository.BatchUpdate(databaseEntities.Select(entity => (GetId(entity), entity)), conflictResolver.Resolve, rivalsResolver)
+                .Select(entities => entities?.ToList() ?? new List<TInterface>())
```
amulware: Why ToList()?

simonrozsival: To get rid of many R# warnings (multiple enumerations).

Toggl.Foundation/Sync/States/Pull/FetchObservables.cs
simonrozsival: I thought about that. I think both of the variants are equally sad. The solution would be not to have one object holding all of the observables, but I don't think we should do such complex change right now.

Toggl.Foundation/Sync/States/Pull/FetchObservables.cs
simonrozsival: Noooo! 🙈

Toggl.Foundation/Sync/States/Pull/PersistState.cs
amulware: Isn't it already a list?

simonrozsival: I will check what was the problem and if I can get rid of this.

Toggl.Foundation/Sync/States/Pull/FetchObservables.cs
amulware: agreed

Toggl.Foundation/Sync/States/Pull/FetchObservables.cs
```diff
@@ -45,5 +47,37 @@ IObservable<IPreferences> preferences
             Tasks = tasks;
             Preferences = preferences;
         }
+
+        public IObservable<List<T>> GetByType<T>()
```
amulware: I indeed don't like the use of tags in the model. We could fix that by injecting this method.

Toggl.Foundation.Tests/Sync/States/Pull/PersistStateTests.cs
```diff
@@ -16,323 +20,260 @@
...
+            var observables = createObservables(entities);
 
-            public virtual void EmitsTransitionToPersistFinished()
-            {
-                var observables = CreateObservables();
+            var transition = (Transition<FetchObservables>)state.Start(observables).SingleAsync().Wait();
 
-                var transition = state.Start(observables).SingleAsync().Wait();
+            transition.Result.Should().Be(state.FinishedPersisting);
+            assertBatchUpdateWasCalled(new List<ITag>());
```
amulware: It batch updates, even if there is nothing to do?

Toggl.Foundation.Tests/Sync/States/Pull/PersistStateTests.cs
```diff
@@ -16,323 +20,260 @@
...
-                state.Start(observables).SingleAsync().Wait();
+            transition.Result.Should().Be(state.Failed);
+            reason.Should().BeAssignableTo<ServerErrorException>();
+        }
 
-                sinceParameterRepository.Received().Set<TDatabaseInterface>(Arg.Is<DateTimeOffset?>(at));
-            }
+        [Theory, LogIfTooSlow]
+        [MemberData(nameof(ApiExceptions.ExceptionsWhichCauseRethrow), MemberType = typeof(ApiExceptions))]
+        public void ThrowsWhenCertainExceptionsAreCaught(Exception exception)
```
amulware: Bad test name I think? What are 'certain' exceptions?

Toggl.Foundation/Sync/States/Pull/FetchObservables.cs
simonrozsival: You mean using tags in the tests?

Toggl.Foundation.Tests/Sync/States/Pull/PersistStateTests.cs
simonrozsival: I will try to find a better name - I just used what was there before, I will have to think for a while what this test really does.

Toggl.Foundation.Tests/Sync/States/Pull/PersistStateTests.cs
simonrozsival: 🤔

simonrozsival: It does call the batch update, but as the list is empty, it doesn't do anything. Do you think we should skip batch update for empty lists? I don't think that is necessary...

amulware: It's not a big deal maybe. Does it open a realm transaction though? That's something I'd like to avoid if the list is empty. But if we're not doing so yet, that should probably be in realm, not here.

Toggl.Foundation.Tests/Sync/States/Pull/PersistStateTests.cs
```diff
@@ -180,7 +180,7 @@ public void ReturnsServerErrorTransitionWhenHttpFailsWithServerErrorException(Se
 
         [Theory, LogIfTooSlow]
         [MemberData(nameof(ApiExceptions.ExceptionsWhichCauseRethrow), MemberType = typeof(ApiExceptions))]
-        public void ThrowsWhenCertainExceptionsAreCaught(Exception exception)
+        public void ThrowsWhenExceptionsWhichShouldCauseRetrhrowAreCaught(Exception exception)
```
amulware: Typo 'trhow'

PR #2111, opened 2018-05-11
Various bugs and fixes in Start/Stop/Duration popup

### LINKED ISSUE #2040: Time-date-duration popup fixes/improvements

kornelijepetak:
Closes #2040 

by completing the following point:

* [x] Prevent selection in controls (min max dates etc)

🦑 whenever.


Toggl.Foundation.MvvmCross/ViewModels/SelectTimeViewModel.cs
```diff
@@ -24,11 +25,18 @@ public sealed class SelectTimeViewModel
...
         private IDisposable timeServiceDisposable;
 
+        private bool isViewModelPrepared;
+
         private readonly TimeSpanToDurationValueConverter durationConverter = new TimeSpanToDurationValueConverter();
 
         private readonly DateTimeOffsetTimeFormatValueCombiner timeFormatValueCombiner;
         private readonly DateTimeOffsetDateFormatValueCombiner dateFormatValueCombiner;
 
+        public DateTimeOffset MinStartTime { get; set; }
```
willsb: We usually put newlines between our properties

Toggl.Foundation.MvvmCross/ViewModels/SelectTimeViewModel.cs
```diff
@@ -24,11 +25,18 @@ public sealed class SelectTimeViewModel
         private readonly ITimeService timeService;
         private IDisposable timeServiceDisposable;
 
+        private bool isViewModelPrepared;
+
         private readonly TimeSpanToDurationValueConverter durationConverter = new TimeSpanToDurationValueConverter();
 
         private readonly DateTimeOffsetTimeFormatValueCombiner timeFormatValueCombiner;
```
willsb: How is this still here? 👀

kornelijepetak: Still?! This is a new thing. Basically, the calculation of `Min`s and `Max` must not happen before all the props: `StartTime` and `StopTime`. This guard bool prevents updating mins and maxs from being invalid as the initial properties are being set up.

Toggl.Foundation.MvvmCross/ViewModels/SelectTimeViewModel.cs
kornelijepetak: No grouping of similar properties? Hm... too bad. Will fix.

daividssilverio:
As it is, you can't select both your StartTime and StopTime, like one day to the past, or one day to the future. You have to change your stop time one day to the future, press ok, then go back and edit your start time to make this.

Also, you can actually select your start time for the 1st day of the month, then your stop time to the last day of the month, press ok, then go back and make a negative time range.

Saving the time entry won't work, but it does allow you see it on the EditTimeEntry screen.

Toggl.Foundation.MvvmCross/ViewModels/SelectTimeViewModel.cs
kornelijepetak: Please try to be more precise in review comments. 😄 
Otherwise, I have removed the combiners. But now I see that I have forgotten to remove the usings. I will push that change as soon as I'm done with what I'm doing. Feel free to review the rest as though there are no unused usings (_pun intended_).

kornelijepetak:
I believe either the datepicker or the logic is wrong on this one, so I have set it as `wip`. I need to look into it.

daividssilverio:
Currently, it is possible to select the earliest possible start date and latest possible stop date; then arrange the time picker to go past the 999 hours limit duration.

kornelijepetak:
@daividssilverio 999 hours constraint is fixed by #2188

Toggl.Foundation.MvvmCross/ViewModels/SelectTimeViewModel.cs
```diff
@@ -286,9 +302,68 @@ private void togglClockCalendarMode()
...
+            if (!isViewModelPrepared)
+                return;
+
+            if (StopTime.HasValue && StopTime < StartTime)
+            {
+                StartTime = StopTime.Value;
+                StartTimeAfterStopTime.Raise(this);
+            }
+
+            // Because of the bug in datepicker, MaxStopTime must be set before MinStopTime
```
willsb: Leaky abstractions, yay

Toggl.Foundation.MvvmCross/ViewModels/SelectTimeViewModel.cs
```diff
@@ -286,9 +302,68 @@ private void togglClockCalendarMode()
...
+            if (StopTime.HasValue)
+            {
+                if (StopTime < StartTime)
+                {
+                    StopTime = StartTime;
+                    StopTimeBeforeStartTime.Raise(this);
+                }
+
+                // Because of the bug in datepicker, MaxStopTime must be set before MinStopTime
+                MaxStartTime = StopTime.Value;
```
willsb: Why not enforce this by the means of a method that sets both in the correct order?

Toggl.Foundation.MvvmCross/ViewModels/SelectTimeViewModel.cs
```diff
@@ -36,13 +41,19 @@ public sealed class SelectTimeViewModel
...
         public TimeFormat TimeFormat { get; set; }
 
+        private bool stopTimeEntryRequested;
+
         [DependsOn(nameof(CurrentDateTime))]
         public DateTimeOffset? StopTime { get; set; }
 
         public DateTimeOffset StopTimeOrCurrent => StopTime ?? CurrentDateTime;
 
+        public event EventHandler StopTimeBeforeStartTime;
```
willsb: This event is only used as an observable, so please turn it into an observable

Toggl.Giskard/Bindings/DatePickerMaxInclusiveDateTargetBinding.cs
```diff
@@ -0,0 +1,36 @@
+﻿using System;
+using MvvmCross.Binding;
+using MvvmCross.Binding.Droid.Target;
+using Android.Widget;
+
+namespace Toggl.Giskard.Bindings
+{
+    public sealed class DatePickerMaxInclusiveDateTargetBinding : MvxAndroidTargetBinding<DatePicker, DateTimeOffset>
```
willsb: Why do you need both inclusive and exclusive?

daividssilverio:
The date picker UI will be improved in #2194

Toggl.Foundation.MvvmCross/ViewModels/SelectTimeViewModel.cs
kornelijepetak: I agree :( Do you think it'd be an overkill to move this into the datepicker itself? 
By means of introducing `DatePickerBoundaries` type and an atomic property in the datepicker that would be set through the binding? it would require a new class, but fewer targetbindings and better abstraction...

Toggl.Foundation.MvvmCross/ViewModels/SelectTimeViewModel.cs
kornelijepetak: This was already done in #2188

Toggl.Foundation.MvvmCross/ViewModels/SelectTimeViewModel.cs
```diff
@@ -4,14 +4,17 @@
...
 using PropertyChanged;
-using Toggl.Foundation.MvvmCross.Combiners;
-using Toggl.Foundation.MvvmCross.Converters;
+using Toggl.Foundation.Helper;
 using Toggl.Foundation.MvvmCross.Parameters;
 using Toggl.Multivac;
 
 namespace Toggl.Foundation.MvvmCross.ViewModels
 {
+    using System.Reactive.Subjects;
```
willsb: Are we doing inner usings? Because I'd rather, like, NOT

Toggl.Foundation.MvvmCross/ViewModels/SelectTimeViewModel.cs
```diff
@@ -36,13 +49,19 @@ public sealed class SelectTimeViewModel
...
         public TimeFormat TimeFormat { get; set; }
 
+        private bool stopTimeEntryRequested;
+
         [DependsOn(nameof(CurrentDateTime))]
         public DateTimeOffset? StopTime { get; set; }
 
         public DateTimeOffset StopTimeOrCurrent => StopTime ?? CurrentDateTime;
 
+        public IObservable<TemporalInconsistency> TemporalInconsistencyDetected
```
willsb: `{ get ; }` to make this a prop. No need to reevaluate that every time

Toggl.Foundation.MvvmCross/ViewModels/SelectTimeViewModel.cs
kornelijepetak: I'd like to use `using static SelectTimeViewModel.TemporalInconsistency` inside the namespace to avoid having to fully qualify the type. 
However, the subjects one was autogenerated by `add reference refactoring` so it probably just realized there's some usings inside so it added it there... not sure though. Do you want me to move both out?

willsb: You can move only one of them, yes. Should we to agree on doing that for every file for consistency, maybe? pinging @amulware

amulware: Everything to the top for now for consistency. If you want to propose a change to that (currently implicit) rule, then let's make that solution wide by FIRST submitting a PR to the style guide that we can debate before merging and then fixing up all files (or maybe just do it one by one as we touch them, and not all in one huge pr).

kornelijepetak:
Closes #2194

PR #2112, opened 2018-05-11
Fix usage of disposable subscriptions

kornelijepetak:
Disposable subscriptions should be disposed when the views that use the VMs are destroyed.


willsb:
Isn't there a `Dispose` method to use instead of `ViewDestroy`? If there is, please use it. I'd rather see it there.

kornelijepetak:
@willsb I have added the same thing for other view models so to fix them all.

The `MvxViewModel` does not implement a disposable pattern, so there is no `Dispose` method to rely on.

Care for rereview? Or should I just merge it away?

willsb:
Merge away!

Issue #2110, opened 2018-05-11
Write tests for SelectTimeViewModel


kornelijepetak:
The popup's view model needs tests.

heytherewill:
🐗

kornelijepetak:
The popup's viewmodel needs 🔥, not tests.

Issue #2096, opened 2018-05-10
Times in the edit view and in the popup don't match


pe0ny:
I’m testing the popup again. Edit view shows me 13:40 as a start time and popup shows 10:40. Timezones?

Also when I open the popup from the start view (doesn't happen when i open the popup from the Edit view):

![screenshot_20180510-140019](https://user-images.githubusercontent.com/5064445/39866938-4f9a7462-545b-11e8-860d-5e7f4a876a88.png)

kornelijepetak:
When both #2122 and #2123 are merged, this issue is officially done.

kornelijepetak:
Indeed. 

I'd be great to be able to do dependencies like this :

`Closes after merge 2122 2123`

PR #2122, opened 2018-05-11
Fixes for start time entry duration handling

### LINKED ISSUE #2061: Start TE View does not respect stopped time entries
Steps to reproduce:

* Start a new TE
* Click on a clock icon (to change the `Start and stop times`)
* Click on the `Stop` button as to stop the time entry.
* _The stop time shows correctly on the dialog_
* Click `Save`
* _User is returned to the start time entry_
* 🐛 _The duration is still going on at the top of the screen.
* 🗒 Note: The time entry **is** stopped, however. If you save the TE, it will be in the stopped mode, not still running.

Basically, only the UI should be changed to show that the time entry was stopped. Instead of duration, maybe display something like `06/14 16:22 - 16:28`. @pe0ny please advise 😄

### LINKED ISSUE #2115: Manual mode: duration is running on the popup
1) Turn on manual mode
2) Tap '+' to add a new entry
3) Tap on the duration -> Popup opens and you can see the timer is running
☝️ I think it doesn't affect the end result because when I save the entry, it save it with 30min duration, so it's only UI error?

willsb:
Closes #2061 
Closes #2115 
Ref #2096 and #2110 

Squash whenever


Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
```diff
@@ -127,6 +133,8 @@ public TimeSpan DisplayedTime
                 }
 
                 displayedTime = value;
+
+                RaisePropertyChanged();
```
kornelijepetak: Is this missing the `nameof(DisplayedTime)` or is this raising all of them... I've never seen this called parameterless. 😕

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
```diff
@@ -458,6 +461,28 @@ private void clearTagsIfNeeded(long currentWorkspaceId, long newWorkspaceId)
...
 
+        private void OnDurationChanged()
+        {
+            if (Duration == null)
+            {
+                setUpTimeSubscriptionIfNeeded();
+                return;
+            }
+
+            DisplayedTime = Duration.Value;
```
kornelijepetak: This property `DisplayedTime` contains a timespan, I'd rather have this called `DisplayedDuration`. It really doesn't make sense that timespan is called `*Time`.

kornelijepetak:
Also, note the failing test in bitrise. 😕

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
willsb: You've never seen that because of Fody, that does the heavy lift for us.

This works because of [CallerMemberNameAttribute](
https://msdn.microsoft.com/en-us/library/system.runtime.compilerservices.callermembernameattribute(v=vs.110).aspx)

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
willsb: Do you think this should be in this PR? Because the name was like that before

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
kornelijepetak: So, we should use parameters only when it's needed to raise notifications for other members then, right?

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
kornelijepetak: Make sense. I will add it to my `to refactor` list. And then when enough of those accumulate, I'll do a simple refactor PR.

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
```diff
@@ -477,7 +477,7 @@ private void OnDurationChanged()
 
         private void setUpTimeSubscriptionIfNeeded()
         {
-            if (Duration == null || elapsedTimeDisposable != null) return;
+            if (Duration != null || elapsedTimeDisposable != null) return;
```
kornelijepetak: khm... I was looking at this expression and something looked wrong, but then I went around and explained it to myself some other way (wrongly obviously). 

Too tired I guess :)

PR #2124, opened 2018-05-11
Fix increase duration buttons behavior

### LINKED ISSUE #2120: Duration shortcuts override the initial duration
Repro steps:

- Open Edit view
- Click duration
- Press any of the shortcuts

Expected: 
Duration becomes X + Y where X is the current duration and Y the value of the button you clicked

Actual:
Duration becomes Y where Y is the value of the button you clicked

juanlaube:
Closes #2120 

🦑 whenever


willsb:
Ref #2110

juanlaube:
And that's why I don't like the MvxBind in the xml :(

PR #2126, opened 2018-05-11
Adjust datetime fragment dialog window gravity

### LINKED ISSUE #2125: Date-time-duration popup: Fix position from the top
Currently, the datetime popup is centered on screen. Could we please fix it so that the margin above it is 40dp? It feels a bit less jumpy this way:

top of dialog fixed

daividssilverio:
Closes #2125 

🦑 whenever


kornelijepetak:
I have temporarily put this on hold, until we hear from our designers. I am bothered by putting so much stuff at the top where it is not accessible by a thumb, forcing you to use the app with two hands.

Toggl.Giskard/Fragments/SelectTimeFragment.cs
```diff
@@ -87,9 +87,19 @@ public override View OnCreateView(LayoutInflater inflater, ViewGroup container,
...
+
             return view;
         }
 
+        private void setupDialogWindowPosition()
+        {
+            var window = Dialog.Window;
+            var layoutParams = window.Attributes;
+            layoutParams.Gravity = GravityFlags.Top;
+            window.Attributes = layoutParams;
```
kornelijepetak: Shouldn't any such kind of layouting be done in `Post`?

daividssilverio: Not necessarily, you do have to be in the Ui Thread though. As far as I'm aware, the lifecycle events from Activities and Fragments are always run on the Ui Thread.

kornelijepetak: That's what I thought at first before having to put that `SetupWithViewPager` thing in the `Post`.
But, that's probably a different kind of issue. Setting the layouts should work like this.

PR #2129, opened 2018-05-14
Refactor IncreaseDuration command

kornelijepetak:
This is a quick PR making `increase duration` command much simpler.

Also, improves some tests.


Toggl.Foundation.Tests/MvvmCross/ViewModels/SelectTimeViewModelTests.cs
```diff
@@ -1,5 +1,7 @@
 ﻿using System;
 using FluentAssertions;
+using FsCheck;
```
willsb: I see you tried to add a property and failed :eyes:

Next time consider adding Start and End to avoid ArgumentOutOfRangeExceptions. This is a nice use case for property testing

PR #2132, opened 2018-05-14
Add padding to main recycler view when running entry card is visible

### LINKED ISSUE #2114: Can't see the last time entries when the timer is running
When you scroll at the bottom of the Timer page, you can't see the last entries when the timer is running. When the timer isn't running, it looks good.

- When the timer is running the last entry should have 104dp padding from the bottom. See Zeplin: https://zpl.io/25gqy0J

juanlaube:
Closes #2114 

🦑 whenever

### DIFF
--- a/Toggl.Giskard/Activities/MainActivity.cs
+++ b/Toggl.Giskard/Activities/MainActivity.cs
@@ -29,6 +29,7 @@ namespace Toggl.Giskard.Activities
 
         private IDisposable disposable;
         private View runningEntryCardFrame;
+        private RecyclerView mainRecyclerView;
         private FloatingActionButton playButton;
         private FloatingActionButton stopButton;
         private CoordinatorLayout coordinatorLayout;
@@ -49,6 +50,7 @@ namespace Toggl.Giskard.Activities
             runningEntryCardFrame = FindViewById(Resource.Id.MainRunningTimeEntryFrame);
             runningEntryCardFrame.Visibility = ViewStates.Invisible;
 
+            mainRecyclerView = FindViewById<RecyclerView>(Resource.Id.MainRecyclerView);
             playButton = FindViewById<FloatingActionButton>(Resource.Id.MainPlayButton);
             stopButton = FindViewById<FloatingActionButton>(Resource.Id.MainStopButton);
             coordinatorLayout = FindViewById<CoordinatorLayout>(Resource.Id.MainCoordinatorLayout);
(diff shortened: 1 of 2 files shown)


Toggl.Giskard/Activities/MainActivity.cs
```diff
@@ -115,6 +117,9 @@ void onRetryTapped(View view)
                     .OnAnimationEnd(_ => playButton.Show())
                     .Start();
             }
+
+            int bottomPadding = visible ? 104.DpToPixels(this) : 0;
```
daividssilverio: Maybe you could name this 104 dp value. What is this size?

daividssilverio: Also it looks like this padding is too big.

juanlaube:
@willsb @daividssilverio take a look again, please. I don't think this is final, probably there's a better way to do this, maybe with some mvx magic, but it's a better solution now 😄

kornelijepetak:
You can do this much simpler and I believe @willsb was going in that direction:

1. Create another target binding `ViewHeightTargetBinding` similarly to an existing `ViewMarginTargetBinding`. Go to `Startup/TargetFactories.tt` and regenerate it.
2. Make another value converter in `BoolToConstantOverrides` that will switch from 70 to 124 (or whatever constants you see fit.
3. There is a `MainLogFooter` which currently has `70dp` height.
Use that converter in `MainLogFooter` with 
`local:MvxBind="Height YourConverter(IsRunning)"`
4. Profit

daividssilverio:
It works nicely now! Your implementation is something I would do myself if I were working with just Android/Kotlin. But as @kornelijepetak said, you can burn a lot of stuff using Mvx!

juanlaube:
doesn't count if nobody notices 😄

juanlaube:
Waiting for @willsb approval... 😄

Toggl.Foundation.MvvmCross/ViewModels/MainViewModel.cs
```diff
@@ -235,6 +237,8 @@ private void setRunningEntry(IDatabaseTimeEntry timeEntry)
             StopTimeEntryCommand.RaiseCanExecuteChanged();
             StartTimeEntryCommand.RaiseCanExecuteChanged();
             EditTimeEntryCommand.RaiseCanExecuteChanged();
+
+            RaisePropertyChanged(nameof(IsTimeEntryRunning));
```
willsb: Fody should weave this already.

kornelijepetak: Without this line, IIRC, didn't work for me. Please when you remove it, do a functional again to be sure it's working.

PR #2133, opened 2018-05-14
Generic syncing states: Push

### LINKED ISSUE #1309: Use generic composition instead of inheritance for sync states

simonrozsival:
Closes #1309 

This PR changes the way we implement the push-sync states. There are now no entity-specific states, everything is generic 🎉 

Some changes to the data sources were necessary to make everything work well.

There are quite a few changed files - mainly because a lot of files was 🔥.

I had to add a common interface to the API models - to add an extension method to check if the entity changed or not. We can discuss this change if the reviewer disagrees with it.

Please test that pushing works the way it should (focus mainly on the time entries, user data, and preferences).

I know you will be mad at me when you review this - it is A LOT of changes. I don't think it is possible to split it into several PRs though :/ Feel free to ask me anything and if it should be helpful, I can have a call with you explaining what and why was changed.

Squash whenever


Toggl.Foundation/Sync/States/Push/DeleteLocalEntityState.cs
```diff
@@ -0,0 +1,36 @@
...
+            this.dataSource = dataSource;
+        }
+
+        public IObservable<ITransition> Start(TThreadsafeModel entity)
+            => delete(entity)
+                .Select(_ => Deleted.Transition())
+                .Catch((Exception e) => Observable.Return(DeletingFailed.Transition()));
+
+        private IObservable<Unit> delete(TThreadsafeModel entity)
+            => entity == null
```
kornelijepetak: Interestingly enough, the compiler is unable to infer here that `TThreadSafeModel` must be a `class` so it warns about possible comparison of value types with `null`. Strange stuff. Constraints are hard 😛

Toggl.Foundation/Sync/States/Push/BasePushEntityState.cs
```diff
@@ -0,0 +1,49 @@
...
+
+        protected Func<Exception, IObservable<ITransition>> Fail(TThreadsafeModel entity)
+            => exception => shouldRethrow(exception)
+                ? Observable.Throw<ITransition>(exception)
+                : Observable.Return(failTransition(entity, exception));
+
+        private bool shouldRethrow(Exception e)
+            => e is ApiDeprecatedException || e is ClientDeprecatedException || e is UnauthorizedException || e is OfflineException;
+
+        private ITransition failTransition(TThreadsafeModel entity, Exception e)
```
kornelijepetak: Maybe it's just me, but I really dislike nested LINQ queries and nested ternary conditionals.
I would find this more readable (albeit less terse):

```
private ITransition failTransition2(TThreadsafeModel entity, Exception e)
{
    var result = (e, entity);

    if (e is ServerErrorException)
        return ServerError.Transition(result);

    if (e is ClientErrorException)
        return ClientError.Transition(result);

    return UnknownError.Transition(result);
}
```

Toggl.Foundation/Sync/States/Push/DeleteLocalEntityState.cs
kornelijepetak: Interestingly enough, the compiler is unable to infer here that `TThreadSafeModel` must be a `class` so it warns about possible comparison of value types with `null`. Strange stuff. Constraints are hard 😛

Toggl.Multivac/Extensions/ApiModelExtensions.cs
```diff
@@ -0,0 +1,28 @@
...
+               && originalApiModel is ILastChangedDatable originalLastChangeDatable
+                ? originalLastChangeDatable.At > lastChangeDatable.At
+                : !propertiesMatch(apiModel, originalApiModel);
+
+        private static bool propertiesMatch(IApiModel first, IApiModel second)
+        {
+            if (first.GetType() != second.GetType())
+                throw new ArgumentException($"Cannot check if the properties are the same for objects of different types ({first.GetType()}, {second.GetType()})");
+
+            foreach (var property in first.GetType().GetProperties())
```
kornelijepetak: You can replace the `foreach` and the `return` by:

```
return first.GetType()
    .GetProperties()
    .All(p => p.GetValue(first, null).Equals(p.GetValue(second, null)));
```

Toggl.Foundation/Sync/States/Push/BasePushEntityState.cs
simonrozsival: Please make an issue and we can have a discussion there if you think this should be changed - I don't want to change the internal implementation of the states in this PR.

Toggl.Foundation/Sync/States/Push/DeleteLocalEntityState.cs
simonrozsival: The compiler is probably right - `TDatabaseModel` can be an interface, `TDatabaseModel` can be a struct implementing that interface. I will add the constraint here.

Toggl.Foundation/TogglSyncManagerFactory.cs
```diff
@@ -156,7 +156,7 @@ public static class TogglSyncManager
 
         private static IStateResult configurePushTransitionsForTags(
```
amulware: There is some duplication with these methods now. Maybe that can be made generic and shared?

Toggl.Multivac/Models/IApiModel.cs
```diff
@@ -0,0 +1,6 @@
+﻿namespace Toggl.Multivac.Models
+{
+    public interface IApiModel
+    {   
+    }
+}
```
amulware: 👀

Toggl.Multivac/Extensions/ApiModelExtensions.cs
```diff
@@ -0,0 +1,28 @@
+﻿using System;
+using Toggl.Multivac.Models;
+
+namespace Toggl.Multivac.Extensions
+{
+    public static class ApiModelExtensions
+    {
+        public static bool HasChanged(this IApiModel apiModel, IApiModel originalApiModel)
```
amulware: This I'm sort of ok with, though it's not ideal.

Toggl.Multivac/Extensions/ApiModelExtensions.cs
```diff
@@ -0,0 +1,28 @@
...
+                : !propertiesMatch(apiModel, originalApiModel);
+
+        private static bool propertiesMatch(IApiModel first, IApiModel second)
+        {
+            if (first.GetType() != second.GetType())
+                throw new ArgumentException($"Cannot check if the properties are the same for objects of different types ({first.GetType()}, {second.GetType()})");
+
+            foreach (var property in first.GetType().GetProperties())
+            {
+                if (!property.GetValue(first, null).Equals(property.GetValue(second, null)))
```
amulware: But this...

Toggl.Multivac/Extensions/ApiModelExtensions.cs
```diff
@@ -0,0 +1,28 @@
...
+using Toggl.Multivac.Models;
+
+namespace Toggl.Multivac.Extensions
+{
+    public static class ApiModelExtensions
+    {
+        public static bool HasChanged(this IApiModel apiModel, IApiModel originalApiModel)
+            => apiModel is ILastChangedDatable lastChangeDatable
+               && originalApiModel is ILastChangedDatable originalLastChangeDatable
+                ? originalLastChangeDatable.At > lastChangeDatable.At
```
amulware: Is this comparison not inverted?

Toggl.Multivac/Extensions/ApiModelExtensions.cs
amulware: There need to be tests though.

Toggl.Multivac/Extensions/ApiModelExtensions.cs
```diff
@@ -0,0 +1,28 @@
...
+                : !propertiesMatch(apiModel, originalApiModel);
```
amulware: If the properties are the same, what do we care if it gets overridden or not

Toggl.Foundation/Sync/States/Push/Interfaces/IPushEntityState.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+using Toggl.Foundation.Models.Interfaces;
+
+namespace Toggl.Foundation.Sync.States.Push.Interfaces
+{
+    public interface IPushEntityState<TThreadsafeModel>
```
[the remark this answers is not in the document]
simonrozsival: Exactly.

Toggl.Foundation/Sync/States/Push/PushState.cs
```diff
@@ -0,0 +1,46 @@
...
+                        ? (ITransition)PushEntity.Transition(entity)
+                        : NothingToPush.Transition());
+
+        private IObservable<TThreadsafeModel> getOldestUnsynced()
+            => dataSource
+                .GetAll(entity => entity.SyncStatus == SyncStatus.SyncNeeded)
+                .SingleAsync()
+                .Select(entities => entities
+                    .OrderBy(entity => entity.At)
+                    .FirstOrDefault());
```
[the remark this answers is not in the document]
simonrozsival: Yeah, sorting the whole list is unnecessary. We should create an issue.

Toggl.Foundation/Sync/States/Push/PushSingleState.cs
```diff
@@ -0,0 +1,38 @@
...
+using System.Reactive.Linq;
+using Toggl.Foundation.DataSources.Interfaces;
+using Toggl.Foundation.Models.Interfaces;
+using Toggl.Foundation.Sync.States.Push.Interfaces;
+using Toggl.Multivac;
+using Toggl.PrimeRadiant;
+
+namespace Toggl.Foundation.Sync.States.Push
+{
+    internal sealed class PushSingleState<TDatabaseModel, TThreadsafeModel> : IPushState<TThreadsafeModel>
```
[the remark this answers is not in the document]
simonrozsival: 🤔

Toggl.Foundation/Sync/States/Push/UnsyncableEntityState.cs
```diff
@@ -0,0 +1,56 @@
...
+                failedPush.Reason == null
+                    ? nameof(failedPush.Reason)
+                    : nameof(failedPush.Entity)));
+
+        private IObservable<ITransition> failBecauseOfUnexpectedError(Exception reason)
+            => Observable.Throw<Transition<TThreadsafeModel>>(reason);
+
+        private IObservable<ITransition> markAsUnsyncable(TThreadsafeModel entity, string reason)
+            => dataSource
+                .UpdateWithConflictResolution(entity, createUnsyncableFrom(entity, reason), new OverwriteIfLocalDidNotChange<TDatabaseModel, TDatabaseModel>(entity))
```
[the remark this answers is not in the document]
simonrozsival: 😆

Toggl.Foundation/Sync/States/Push/UnsyncableEntityState.cs
```diff
@@ -0,0 +1,56 @@
...
+            Ensure.Argument.IsNotNull(createUnsyncableFrom, nameof(createUnsyncableFrom));
+            
+            this.dataSource = dataSource;
+            this.createUnsyncableFrom = createUnsyncableFrom;
+        }
+
+        public IObservable<ITransition> Start((Exception Reason, TThreadsafeModel Entity) failedPush)
+            => failedPush.Reason == null || failedPush.Entity == null
+                ? failBecauseOfNullArguments(failedPush)
+                : failedPush.Reason is ApiException apiException
```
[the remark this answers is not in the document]
simonrozsival: I didn't change the logic - it is the original implementation that has been there from the first release. It should be entered only with client exceptions, yes. I am not sure about this implementation detail - maybe we should throw if it is not a client error exception?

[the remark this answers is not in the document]
simonrozsival: Still it would be good to have a separate issue and PR to change this.

Toggl.Foundation/TogglSyncManagerFactory.cs
simonrozsival: 🤔 .. it might be possible to do, I am not really sure if it is a good idea to do - we have good control of the process for every type of entity. I will look into it and if it is simple, I will change it.

Toggl.Multivac/Extensions/ApiModelExtensions.cs
simonrozsival: Do you see any potential problems with this?

I will add tests.

Toggl.Multivac/Extensions/ApiModelExtensions.cs
simonrozsival: I already changed this with help from @kornelijepetak to use LINQ and make it much nicer.

Toggl.Multivac/Extensions/ApiModelExtensions.cs
simonrozsival: If the properties are the same -> we can simply override it. If they aren't the same -> the entity changed (e.g., preferences) and we don't want to override the local changes - that is why we care. Or did you mean something else by this comment?

Toggl.Multivac/Models/IApiModel.cs
simonrozsival: Another empty interface, I know... but I wasn't able to create the extension method without this interface. Also I think it is good if the models in one layer/project have some common ancestor without any exceptions.

Toggl.Foundation.Tests/Toggl.Foundation.Tests.csproj
```diff
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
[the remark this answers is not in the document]
simonrozsival: 🙈

Toggl.Foundation.Tests/Sync/States/Push/PushStateTests.cs
```diff
@@ -0,0 +1,84 @@
...
+            dataSource.GetAll(Arg.Any<Func<IDatabaseTestModel, bool>>())
+                .Returns(Observable.Throw<IEnumerable<IThreadsafeTestModel>>(new Exception()));
+
+            Action callingStart = () => state.Start().SingleAsync().Wait();
+
+            callingStart.ShouldThrow<Exception>();
+        }
+
+        [Fact, LogIfTooSlow]
+        public void ReturnsPushEntityTransitionWithTheOldestEntity()
```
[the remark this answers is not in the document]
simonrozsival: I am not sure how you mean it. We always set the at dates.

Toggl.Multivac/Extensions/ApiModelExtensions.cs
amulware: I don't like the shared interface just for this one method, just to then inside check if it's a more specific type. Not to mention that this is only used in a single place as far as I can tell?

Toggl.Multivac/Extensions/ApiModelExtensions.cs
amulware: What I don't like is the reflection.

Toggl.Multivac/Extensions/ApiModelExtensions.cs
amulware: I think I'm not fully understanding how this is used.

Is the following correct?

```
original = realm.Get()
updatedFromBackend = push(original) // might take a long time
realm.transaction {
  current = realm.Get()
  if original == current // this is where we need this method
    realm.Overwrite(updatedFromBackend)
}
```

Toggl.Multivac/Extensions/ApiModelExtensions.cs
kornelijepetak: Yeah... Me neither. Can't see how to do a generic one without reflection here... T4?

amulware: The extension method is only called from a single place, where we generically know the type of the model. Why not make the extension method generic and check the type? Or just use pattern matching to check the type, and then... have generated/hand written comparison methods for the types that don't have at dates? -.-

Completely other proposal: Can we have all IDatabaseSyncable models implement the `At` property?

simonrozsival: You are right that it is now used only from one place - originally it was used in several places.

We could hand write the comparison. Sure.

Putting `At` into all syncable entities - well, no - `IDatabasePreferences`.

Toggl.Multivac/Extensions/ApiModelExtensions.cs
simonrozsival: That is correct.

Toggl.Foundation.Tests/Toggl.Foundation.Tests.csproj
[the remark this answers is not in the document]
simonrozsival: I guess I removed BOM. I won't bother adding it back, sorry.

Toggl.Foundation/Sync/ConflictResolution/IConflictResolver.cs
```diff
@@ -2,7 +2,7 @@
 
 namespace Toggl.Foundation.Sync.ConflictResolution
 {
-    interface IConflictResolver<T>
+    public interface IConflictResolver<T>
```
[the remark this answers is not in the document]
kornelijepetak: But why?

Toggl.Foundation/Sync/States/Push/DeleteLocalEntityState.cs
kornelijepetak: Yes, you are right. The dangerous world of structs that implement interfaces.

kornelijepetak: Have you forgotten to add a constraint here? Or did I get something wrong?

kornelijepetak:
Should I take it you have decided not to change the name to `CollectionDataSource`?

simonrozsival:
Yes, the renaming I used was like this:
- `IDataSource` => `IBaseDataSource`
- `ISingleObjectDataSource` => `ISingletonDataSource`
- `IManyObjectsDataSource` => `IDataSource`

Toggl.Foundation/Sync/ConflictResolution/OverwriteIfLocalDidNotChange.cs
```diff
@@ -1,23 +1,33 @@
...
-            => localEntity.HasChanged(currentLocal)
+            => hasChanged(currentLocal, localEntity)
                 ? ConflictResolutionMode.Ignore
                 : ConflictResolutionMode.Update;
+
+        private bool hasChanged(TDatabase currentLocal, TDatabase originalLocal)
+            => currentLocal is ILastChangedDatable currentLastChangedDatable &&
+               originalLocal is ILastChangedDatable originalLastChangedDatable
+                ? currentLastChangedDatable.At != originalLastChangedDatable.At
+                : !noPropertyChanged(currentLocal, originalLocal);
```
amulware: Double negation, please change method name to `allPropertiesTheSame` or similar.

Toggl.Foundation.Tests/Sync/States/Push/PushStateTests.cs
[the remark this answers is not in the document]
amulware: I mean, if we change, for example, a time entry locally, do we change the `at` date to 'now'? every time, i.e. it changes in the database with every change the user makes?

amulware:
The changes are looking alright to me, but I haven't checked f1d6260edde2bc92087afde0efeb4f3a9c34e1e4 in detail yet.

Toggl.Foundation.Tests/Sync/States/Push/PushStateTests.cs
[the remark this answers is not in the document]
simonrozsival: Yes, that is exactly what happens.

Toggl.Foundation/Sync/ConflictResolution/IConflictResolver.cs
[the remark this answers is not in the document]
simonrozsival: I made a few changes to keep this internal.

simonrozsival:
About the last changes:
- the data source doesn't expose conflict and rivals resolvers anymore
- I added the extension method back - it is used from two places, so it is useful - it is now based on database models, which actually makes sense
- the conflict resolver can now be internal anymore

Toggl.Foundation/DataSources/SingletonDataSource.cs
```diff
@@ -53,35 +51,24 @@ public override IObservable<TThreadsafe> Create(TThreadsafe entity)
...
+                    return new UpdateResult<TThreadsafe>(u.OriginalId, u.Entity);
 
-                case UpdateResult<TThreadsafe> updated:
-                    currentSubject.OnNext(updated.Entity);
-                    break;
+                case IgnoreResult<TThreadsafe> i:
+                    return new IgnoreResult<TThreadsafe>(i.Id);
             }
+
+            throw new InvalidOperationException($"Cannot handle conflict resolution result {result.GetType().FullName} in the singleton data source.");
```
amulware: Creation is not supported? How do we create the very first object?

Toggl.Foundation/DataSources/ObservableDataSource.cs
```diff
@@ -54,18 +53,14 @@ public override IObservable<Unit> Delete(long id)
...
-            => base.UpdateWithConflictResolution(original, entity, conflictResolver, rivalsResolver)
+        public override IObservable<IConflictResolutionResult<TThreadsafe>> OverwriteIfOriginalDidNotChange(
+            TThreadsafe original, TThreadsafe entity)
+            => base.OverwriteIfOriginalDidNotChange(original, entity)
                 .Do(handleBatchUpdateResult);
 
         public override IObservable<IEnumerable<IConflictResolutionResult<TThreadsafe>>> BatchUpdate(IEnumerable<TThreadsafe> entities)
             => base.BatchUpdate(entities)
                 .Do(updatedEntities => updatedEntities
-                    .Where(result => !(result is IgnoreResult<TThreadsafe>))
```
amulware: Why is this removed?

simonrozsival: It will be ignored - the switch doesn't have a case for it.

Toggl.Foundation/DataSources/SingletonDataSource.cs
simonrozsival: 🤔

Toggl.Foundation/Extensions/DatabaseModelExtensions.cs
```diff
@@ -0,0 +1,20 @@
...
+    public static class DatabaseModelExtensions
+    {
+        public static bool DiffersFrom<TDatabase>(this TDatabase first, TDatabase second)
+            where TDatabase : IDatabaseModel
+            => first is ILastChangedDatable currentLastChangedDatable &&
+               second is ILastChangedDatable originalLastChangedDatable
+                ? currentLastChangedDatable.At != originalLastChangedDatable.At
+                : anyPropertyChanged(first, second);
+
+        private static bool anyPropertyChanged<TDatabase>(TDatabase second, TDatabase originalLocal)
```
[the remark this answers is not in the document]
simonrozsival: 🙈

Toggl.Foundation/Extensions/DatabaseModelExtensions.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System.Linq;
+using Toggl.Multivac.Models;
+using Toggl.PrimeRadiant.Models;
+
+namespace Toggl.Foundation.Extensions
+{
+    public static class DatabaseModelExtensions
+    {
+        public static bool DiffersFrom<TDatabase>(this TDatabase first, TDatabase second)
```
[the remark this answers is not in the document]
simonrozsival: I will rename it to `T`.

Toggl.Foundation/DataSources/SingletonDataSource.cs
simonrozsival: It must support creation, otherwise it would 💥 I will put it back.

amulware:
f1d6260edde2bc92087afde0efeb4f3a9c34e1e4 hurts my brain a bit, but I think it's alright.

Issue #1334, opened 2018-01-25
UI to see projects/clients/tags that failed to sync


amulware:
- [ ] We should try to show meaningful error messages, either those from backend or custom ones
- [ ] The user should be given a hint on how to resolve the issue (for example getting access back to a workspace)
- [ ] Users should be able to RETRY, DELETE(, RENAME?)
- In the future, if we allow projects/tags/clients to be edited, we may want to include the option to 'reset' the change by downloading the entity from backend

Overall, I think this screen should require the user to do a lot manually, but also give them the power to 'fix' their own syncing issues.

There are two different cases that should maybe be handled differently, in terms of what errors/hints to show:
- Actual sync of an entity failed
  - use backend error messages
- We find out that the user lost access to a workspace where they have unsynced entities
  - use custom error message
  - give hint about getting access back to workspace

amulware:
On a related note, time entries affected by any of the cases above should not even try to sync, since that sync is bound to fail. However, if the user manages to fix an issue, the affected time entries should again sync.

Maybe it is worth linking to this screen from the error message shown in such TEs.

amulware:
For now we can go ahead on this without design and hide the screen behind an `#if DEBUG` so we can already start using it for testing. The button for it can be next to the toggl logo or something in the navigation bar on the main view (only visible in debug builds!)

The screen should show all unsynced/unsyncable entities in a list, with a gray 'needs syncing' or red 'sync error' icon, and the sync error string next to that.

The screen does not have to update live but only when opening, and there needs to be a back button leading back to the main view.

amulware:
Entities to show: TEs, clients, projects, tags, user, user preferences (not sure about the latter two, but in principle something could go wrong there)

amulware:
I suggest @Odrakir or @sofokli1 handle this one. :)

amulware:
Also, for the first version of this, we don't need the screen to be interactive yet. Just as long as we can look at the data.

Odrakir:
Any hints on how to get from the DB those entities that failed to sync?

amulware:
Certainly, the relevant models should all have a `SyncState` which you can filter on.

I propose maybe using an interactor for this since that's the hip new thing we're doing now.

PR #2138, opened 2018-05-15
Sync failure interactors

Odrakir:
This is just the first part for #1334 

Adds interactors to get projects, tags and clients that failed to sync, and another one to get the total number of those. It also has all the unit tests.

According to comments on the issue I might need to add more interactors for other entities, but I think that's better done in another PR, once everything is working.


Toggl.Foundation/Interactors/Sync/GetNumberOfItemsThatFailedToSyncInteractor.cs
```diff
@@ -0,0 +1,42 @@
...
+                tags,
+                (projectsCount, clientsCount, tagsCount) => projectsCount + clientsCount + tagsCount);
+        }
+
+        private IEnumerable<IDatabaseSyncable> failedToSync(IEnumerable<IDatabaseSyncable> items)
+        {
+            return items.Where(p => p.SyncStatus == SyncStatus.SyncFailed);
+        }
+
+
```
simonrozsival: 🔥

Toggl.Foundation/Interactors/InteractorFactory.Project.cs
```diff
@@ -1,4 +1,6 @@
 ﻿using System;
+using System.Collections.Generic;
+using Toggl.PrimeRadiant.Models;
```
simonrozsival: Are these usings necessary when you didn't edit the rest of the file in this PR?

Toggl.Foundation/Interactors/Sync/GetItemsThatFailedToSyncInteractor.cs
```diff
@@ -0,0 +1,30 @@
...
+            this.repository = repository;
+        }
+
+        public IObservable<IEnumerable<T>> Execute()
+            => repository.GetAll()
+                .Select(failedToSync);
+
+        private IEnumerable<T> failedToSync(IEnumerable<T> items)
+        {
+            return items.Where(p => p.SyncStatus == SyncStatus.SyncFailed);
```
simonrozsival: This could be expression bodied function.

Toggl.Foundation/Interactors/Sync/GetItemsThatFailedToSyncInteractor.cs
```diff
@@ -0,0 +1,30 @@
...
+        public GetItemsThatFailedToSyncInteractor(IRepository<T> repository)
+        {
+            Ensure.Argument.IsNotNull(repository, nameof(repository));
+
+            this.repository = repository;
+        }
+
+        public IObservable<IEnumerable<T>> Execute()
+            => repository.GetAll()
+                .Select(failedToSync);
```
simonrozsival: This is actually quite dangerous - the repository returns the Realm objects, which can't be accessed from different threads than the one from which they were fetched. You must convert the Realm object to a Foundation model using e.g. `TimeEntry.From` static function.

Maybe this is fine now though - the new data sources which will be hopefully merged soon have the same `GetAll` method, but they already do the conversion and return a threadsafe model. So I would say let's keep it like this and change this to use data sources instead of repositories when it is ready.

Toggl.Foundation/Interactors/Sync/GetItemsThatFailedToSyncInteractor.cs
```diff
@@ -0,0 +1,30 @@
...
+
+        public GetItemsThatFailedToSyncInteractor(IRepository<T> repository)
+        {
+            Ensure.Argument.IsNotNull(repository, nameof(repository));
+
+            this.repository = repository;
+        }
+
+        public IObservable<IEnumerable<T>> Execute()
+            => repository.GetAll()
```
simonrozsival: You can put the predicate directly to the `GetAll(entity => ...)` and avoid having an extra call to `Where`.

Toggl.Foundation/Interactors/Sync/GetItemsThatFailedToSyncInteractor.cs
simonrozsival: But this separate method is probably not needed and the predicate can be moved to the `GetAll` fn.

Toggl.Foundation/Interactors/Sync/GetItemsThatFailedToSyncInteractor.cs
Odrakir: Oh my... This needs a much deeper explanation I'm afraid.

Toggl.Foundation/Interactors/Sync/GetNumberOfItemsThatFailedToSyncInteractor.cs
```diff
@@ -0,0 +1,39 @@
...
+        public GetNumberOfItemsThatFailedToSyncInteractor(ITogglDatabase database)
+        {
+            Ensure.Argument.IsNotNull(database, nameof(database));
+
+            this.database = database;
+        }
+
+        public IObservable<int> Execute()
+        {
+            var projects = database.Projects.GetAll().Select(failedToSync).Select((a) => a.Count());
```
kornelijepetak: I see you are in the process of changing stuff, but just a few things (hence not a request for changes, but a plain comment):

1) You can simplify things (avoid double selects) like this
```
public IObservable<int> Execute()
{
    var projects = database.Projects.GetAll().Select(failedToSync);
    var clients = database.Clients.GetAll().Select(failedToSync);
    var tags = database.Tags.GetAll().Select(failedToSync);

    return Observable.CombineLatest(
        projects,
        clients,
        tags,
        (projectsCount, clientsCount, tagsCount) => projectsCount + clientsCount + tagsCount);
}

private int failedToSync(IEnumerable<IDatabaseSyncable> items)
{
    return items.Count(p => p.SyncStatus == SyncStatus.SyncFailed);
}
```

2) There is no need to put parentheses around a single lambda parameter:
`.Select(a => a.Count())` will do.

3) For the name of the lambda parameter in LINQ queries, there's an undocumented convention to use first letter of what that actually is, for example `c` for `collection`, or `p` for `project`.

Toggl.Foundation.Tests/Interactors/Sync/GetClientsThatFailedToSyncInteractorTests.cs
```diff
@@ -0,0 +1,44 @@
...
+            {
+                MockClient[] clients = {
+                    new MockClient { Id = 0, SyncStatus = SyncStatus.SyncFailed },
+                    new MockClient { Id = 1, SyncStatus = SyncStatus.InSync },
+                    new MockClient { Id = 2, SyncStatus = SyncStatus.SyncFailed },
+                    new MockClient { Id = 3, SyncStatus = SyncStatus.SyncNeeded },
+                    new MockClient { Id = 4, SyncStatus = SyncStatus.InSync },
+                    new MockClient { Id = 5, SyncStatus = SyncStatus.SyncFailed }
+                };
+
```
kornelijepetak: I think we are using one empty line between segments of AAA in tests, and no empty lines in a segment.

Toggl.Foundation/Interactors/Sync/GetNumberOfItemsThatFailedToSyncInteractor.cs
```diff
@@ -0,0 +1,41 @@
...
+                projects,
+                clients,
+                tags,
+                (projectsCount, clientsCount, tagsCount) => projectsCount + clientsCount + tagsCount);
+        }
+
+        private int countFailedToSync(IEnumerable<IDatabaseSyncable> items)
+        {
+            return items
+                .Where(p => p.SyncStatus == SyncStatus.SyncFailed)
```
kornelijepetak: `Where(predicate).Count()` can be replaced by `Count(predicate)`.

It reads: `Count how many satisfy a condition`

Odrakir: The functional programmer in my likes the `.Count()` much better ;)

PR #2139, opened 2018-05-15
Forgot password

### LINKED ISSUE #2023: New Forgot your password UI
- New Forgot your password UI is accessible from the Log in page.
- User can use "Back" button on the nav bar to navigate back to the Log in page.
- If user entered already the email in the Log in form, use the same email to auto fill Forgot your password form. If user didn't enter anything, keep the field empty.
- Auto focus the email field when user opens this view.
- When resetting the password succeeded, make sure the keyboard is closed and show Done card at the bottom of the screen. (Card is positioned 40pt from the bottom.) Display the card for 4 seconds and then auto navigate back to the Log in page.
- Email field is always positioned 64pt from the nav bar on smaller screens (SE), and 136pt on bigger screens.
- Button is always positioned 32pt from the bottom.

### Forgot your password view
toggl - ios - forgot your password 1x

Zeplin: https://zpl.io/VDlYABe

### Forgot your password view: error message
toggl - ios - forgot your password error 1x

Zeplin: https://zpl.io/2pojE9o

### Forgot your password view: done
Copy: "Link sent. Please, check your email to reset the password."

toggl - ios - forgot your password done 1x

Zeplin: https://zpl.io/a3voRGA

### Log in: password is reset
toggl - ios - forgot your password ok 1x

Zeplin: https://zpl.io/adAwgO7

arturs105:
Closes #2023 

Squash whenever.


Toggl.Foundation.MvvmCross/ViewModels/ForgotPasswordViewModel.cs
```diff
@@ -0,0 +1,123 @@
...
+                    ErrorMessage = apiException.LocalizedApiErrorMessage;
+                    break;
+
+                default:
+                    ErrorMessage = Resources.PasswordResetGeneralError;
+                    break;
+            }
+        }
+
+        private void close() => returnEmail();
```
simonrozsival: 🔥

Toggl.Foundation.MvvmCross/ViewModels/ForgotPasswordViewModel.cs
```diff
@@ -0,0 +1,123 @@
...
+            Email = parameter.Email;
+        }
+
+        private void reset()
+        {
+            ErrorMessage = "";
+            IsLoading = true;
+            PasswordResetSuccessful = false;
+
+            loginManager
```
simonrozsival: This might be a nice `async` command - this could be awaited in a try-catch block and avoid using `Do` for side effect and  you wouldn't have to hold the disposable - which might potentially be disposed during the HTTP request and the result would never be observed.

Toggl.Foundation.MvvmCross/ViewModels/ForgotPasswordViewModel.cs
```diff
@@ -0,0 +1,123 @@
...
+        private readonly IAnalyticsService analyticsService;
+        private readonly IMvxNavigationService navigationService;
+
+        private readonly TimeSpan delayAfterPassordReset = TimeSpan.FromSeconds(4);
+
+        public Email Email { get; set; } = Email.Empty;
+
+        public string ErrorMessage { get; private set; }
+
+        public bool HasError => !string.IsNullOrEmpty(ErrorMessage);
```
simonrozsival: Shouldn't this depend on `ErrorMessage`?

Toggl.Foundation.MvvmCross/ViewModels/ForgotPasswordViewModel.cs
simonrozsival: Also I think that we then wouldn't need the `ITimeService.RunAfterDelay` thing. I am not sure if it is a good idea to have that method around or not 🤔

arturs105: I don't really see, what's wrong with the current command.

simonrozsival: Just keeping the reference to the disposable should do the job. I thought this could be simplified a bit with no need to add another method to the time service but I leave that up to you - it should be just fine.

simonrozsival:
Let's keep it like this for now.

PR #2140, opened 2018-05-15
Fix invalid tag colors

### LINKED ISSUE #2039: Tag text is not visible in Edit View
Probably just a color issue because the width of the tag is OK. So the data is probably fine. Maybe some of the changes in previous PRs introduced this problem.

image

kornelijepetak:
Closes #2039 

Ok, the issue describes two different problems. First one (in which text is not visible) was solved by something else in other PRs, basically, a wrong color in XML or something similar. (Maybe it is the same as the second problem, but then it was by luck).

However, the second one, where the tag background is black (or any other color than it should be) is related to using the same drawable in combination with `DrawableColorTargetBinding`. For some reason, I believe that the same instance of `GradientDrawable` is being modified in reports and changed to black.

This is a potential problems for all places where `RoundedRectangle` is used, because changing the color thru the target binding will change the color for all uses of `RoundedRectangle`. I have tested this by adding another View with `background` set to `RoundedRectangle` drawable on Edit View. When I go to reports and go back to edit view, the view is black instead of blue.

🦑 whenever.


daividssilverio:
It works nicely now.
But I wonder how could this be happening. I always thought that drawables instances were not shared across views. The EditTimeEntryTagCell doesn't even have any bindings for the drawable.

kornelijepetak:
It doesn't have to have it, @daividssilverio. It used the drawable that gets changed by a target binding on another place. 
I also thought they are inflated whenever needed from scratch, but now I am not so sure. I am waiting for comments from @willsb .
Indeed something unexpected...

willsb:
They are shared, yes. You should always call `Mutate()` when changing drawables. This also works, though, so I'm approving it as is 😄

kornelijepetak:
I will post this here just for the reference: 
https://developer.android.com/reference/android/graphics/drawable/Drawable.html#mutate()
@daividssilverio

PR #2141, opened 2018-05-15
Fix sign up in Android

### LINKED ISSUE #2089: Cannot create new account.
On the sign up password page it displays "please select your country" and there is no country entry field, and it won't let you leave the page or continue through to create an account, there is no country selection field. 
screenshot_20180510-172605

juanlaube:
Closes #2089 

🦑 whenever


Toggl.Foundation.MvvmCross/ViewModels/LoginViewModel.cs
```diff
@@ -346,16 +329,13 @@ private void googleLogin()
...
 
-                loginDisposable =
-                    loginManager
-                        .SignUp(Email, Password, TermsAccepted, CountryId.Value)
-                        .Do(_ => analyticsService.TrackSignUpEvent(AuthenticationMethod.EmailAndPassword))
-                        .Subscribe(onDataSource, onError, onCompleted);
-            }
+            loginDisposable =
+                loginManager
+                    .SignUp(Email, Password, true, null)
```
kornelijepetak: I would have this hard coded as well, but please do it thru constants. Much easier to change things later on.

PR #2145, opened 2018-05-15
Sync failures interactor refactor

Odrakir:
Part of #1334 

After finally merging #2138 I realized it can be done in a way that is more useful for the views. So this is it.

I'm not sure at all on how I concat the Observable arrays in `GetItemsThatFailedToSyncInteractor`, please take a look at that.

Also, for the `SyncFailureItemTests` I had to create the namespace `Toggl.Foundation.Tests.SyncModel` because if I did `Toggl.Foundation.Tests.Models` I got errors in some other files confused by the Models namespace.


Toggl.Foundation/Models/SyncFailureItem.cs
```diff
@@ -0,0 +1,41 @@
...
+
+namespace Toggl.Foundation.Models
+{
+    public class SyncFailureItem
+    {
+        public SyncStatus SyncStatus { get; }
+
+        public string SyncErrorMessage { get; }
+
+        public string Type { get; }
```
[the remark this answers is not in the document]
kornelijepetak: If this was a ViewModel I would disagree, but I am also against losing any data until the UI. I am good with `Type Type { get; }` here.

Toggl.Foundation/Interactors/Sync/GetItemsThatFailedToSyncInteractor.cs
```diff
@@ -1,29 +1,52 @@
...
+                .Select(convertToSyncFailures);
+
+            var tags = database.Tags
                 .GetAll(p => p.SyncStatus == SyncStatus.SyncFailed)
-                .Select(items => items.Select(convert));
+                .Select(convertToSyncFailures);
+
+            return Observable
+                .Concat(clients, projects, tags)
+                .Aggregate(Enumerable.Empty<SyncFailureItem>(), concat);
```
kornelijepetak: Isn't `Concat` doing all the work for you? Why add Aggregate?

Toggl.Foundation/Models/SyncFailureItem.cs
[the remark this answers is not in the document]
Odrakir: Yes, you are right, this was intended to be a ViewModel at the beginning, but then I changed my mind and forgot to change this. @simonrozsival about storing the type itself... You mean there's a type type (or class, or whatever) in C# ? What's that?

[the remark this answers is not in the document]
kornelijepetak: Yes, class `Type` holds all the metadata about a type.
Every object has `GetType()` method that will return an instance of `Type`. From it you can get to all definitions of type's members (reflection).

Toggl.Foundation/Interactors/Sync/GetItemsThatFailedToSyncInteractor.cs
Odrakir: Concat takes the 3 Observable<Enumerable> and concats all the events into a single Observable. In this case the Observable will generate 3 events with each of the Enumerables. I want just one event with all of them. 

Am I making sense? http://reactivex.io/documentation/operators/concat.html

kornelijepetak: Indeed the behaviour is different. 

I think it's ok if you did it with lambda, as it's not that big:
`.Aggregate(Enumerable.Empty<SyncFailureItem>(), (acc, next) => acc.Concat(next));`

Odrakir: If this was Swift I would have an extension on IEnumerable with a concat method that concats two IEnumerables and returns the result... 🤔  But ok, let's put that inline.

Toggl.Foundation/Interactors/Sync/GetItemsThatFailedToSyncInteractor.cs
```diff
@@ -1,29 +1,47 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Reactive.Linq;
+using System.Collections.ObjectModel;
```
simonrozsival: 🤔 I think you should remove this.

Toggl.Foundation/Interactors/Sync/GetItemsThatFailedToSyncInteractor.cs
```diff
@@ -1,29 +1,47 @@
...
+                .Select(convertToSyncFailures);
+
+            var projects = database.Projects
+                .GetAll(p => p.SyncStatus == SyncStatus.SyncFailed)
+                .Select(convertToSyncFailures);
+
+            var tags = database.Tags
                 .GetAll(p => p.SyncStatus == SyncStatus.SyncFailed)
-                .Select(items => items.Select(convert));
+                .Select(convertToSyncFailures);
```
simonrozsival: If you did `SelectMany` (flat map) here, then the `Aggregate` later wouldn't be necessary (instead you would call `ToList` to merge all the events into a single enumerable).

Toggl.Foundation/Interactors/Sync/GetItemsThatFailedToSyncInteractor.cs
```diff
@@ -1,29 +1,47 @@
...
-            this.convert = convert;
+            this.database = database;
         }
 
-        public IObservable<IEnumerable<T>> Execute()
-            => repository
+        public IObservable<IEnumerable<SyncFailureItem>> Execute()
+        {
+            var clients = database.Clients
+                .GetAll(p => p.SyncStatus == SyncStatus.SyncFailed)
```
simonrozsival: This is a bit repetitive - you could abstract this into a function which takes a repository and returns an observable, but since this is quite short and it probably won't be touched frequently, I am fine with keeping it like this.

Toggl.Foundation/Interactors/Sync/GetItemsThatFailedToSyncInteractor.cs
Odrakir: ⁉️ That's a crazy overload. Something similar to what used to be in Swift with Optionals and Arrays (and got finally replaced with `compactMap`). I'll use it because it makes the code much simpler, but I'm not sure I like that...

Toggl.Foundation/Interactors/Sync/GetItemsThatFailedToSyncInteractor.cs
Odrakir: That will imply generic methods, right? I'll try.

Issue #2079, opened 2018-05-09
Reports: clicking on current date range on the toolbar won't open the calendar


daividssilverio:
I tested on my Nexus 5. Sometimes it just won't work, sometimes I have to scroll down to see the calendar. If I do get to see the calendar, clicking on the current date range does close it.

Detail: I don't have any time entry this week on the account I'm using for testing.

It works on the emulator.

PR #2146, opened 2018-05-15
Fix reports initial animation glitch

### LINKED ISSUE #2081: Reports: Scroll on the charts is not working properly
The content is moving up and I'm not able to scroll down anymore. It gets worse as I try to click on the current date range from the toolbar.

It doesn't matter if it has time entries or not. But the problem only happens on the Nexus 5 without entries.

daividssilverio:
Closes #2081 #2079 

This fixes actually both problems, the animation wasn't being called on my device (had to force it with RequestLayout and Invalidate).
The other problem is that, also depending on the device, the TopMargin of the CalendarContainer wont't match it's initial height, causing the animation to misbehave ever increasingly the CalendarContainer's TopMargin.

🦑 if possible.


Toggl.Giskard/Views/ReportsLinearLayout.cs
```diff
@@ -192,11 +192,15 @@ private void animateCalendar(bool forceHide, bool isFling)
         {
             var marginParams = CalendarContainer.LayoutParameters as MarginLayoutParams;
             var isHidding = forceHide || marginParams.TopMargin != negativeContainerHeight;
-            var newMargin = isHidding ? negativeContainerHeight : 0;
 
+            if (isHidding && marginParams.TopMargin < negativeContainerHeight) return;
```
kornelijepetak: Return in the same line only for very short expressions (and no return value). As per guidelines.

PR #2164, opened 2018-05-16
Sync failures viewmodel

Odrakir:
Part of #1334 

This only adds the view model that exposes a simple list for the view controller in my next PR to display the items in a table view.

My doubts are in the way I await the observable from the interactor. By doing that I don't need to dispose anything?


Toggl.Foundation.MvvmCross/ViewModels/SyncFailuresViewModel.cs
```diff
@@ -0,0 +1,36 @@
...
+            this.interactorFactory = interactorFactory;
+        }
+
+        public override async Task Initialize()
+        {
+            await base.Initialize();
+
+            SyncFailures = await interactorFactory
+                .GetItemsThatFailedToSync().Execute()
+                .Select(i => i.ToList())
```
simonrozsival: This is a bit weird - maybe you can change directly the interactor's return type to list.

Toggl.Foundation.Tests/MvvmCross/ViewModels/SyncFailuresViewModelTests.cs
```diff
@@ -0,0 +1,72 @@
...
+using Toggl.PrimeRadiant.Models;
+using Xunit;
+
+namespace Toggl.Foundation.Tests.MvvmCross.ViewModels
+{
+    public sealed class SyncFailuresViewModelTests
+    {
+        public abstract class SyncFailuresViewModelTest : BaseViewModelTests<SyncFailuresViewModel>
+        {
+
```
simonrozsival: 🔥

Toggl.Foundation.Tests/MvvmCross/ViewModels/SyncFailuresViewModelTests.cs
```diff
@@ -0,0 +1,72 @@
...
+        }
+
+        public sealed class TheSyncFailuresProperty : SyncFailuresViewModelTest
+        {
+            [Fact, LogIfTooSlow]
+            public void ExecutesGetItemsThatFailedInteractor()
+            {
+                var mockedInteractor = Substitute.For<IInteractor<IObservable<IEnumerable<SyncFailureItem>>>>();
+                InteractorFactory.GetItemsThatFailedToSync().Returns(mockedInteractor);
+                ViewModel.Initialize();
```
simonrozsival: Separate this with blank lines to achieve AAA (Arrange Act Assert)

Toggl.Foundation.MvvmCross/ViewModels/SyncFailuresViewModel.cs
simonrozsival: Or maybe - do you need a list here? Isn't `IEnumerable` enough for the table?

Odrakir: I don't know, I'm guessing an `IEnumberable` is not enough as it doesn't have random access to its elements, right? I could do the conversion later though...

PR #2174, opened 2018-05-17
Onboarding: replace the last screen with login page

### LINKED ISSUE #2165: New login screen should replace the last onboarding screen
> #1962 New Log in UI replaces the currently in use last onboarding step (blue screen with illustrations and login/signup buttons).

Currently, we still have the last onboarding step with blue background where user can choose between Login or Signup. This should be replaced with the new Login screen, so after finishing the onboarding, user is directed to the Login page.

simonrozsival:
Closes #2165 

I also refactored the code in the `OnboardingViewModel`. I hope the reviewer won't mind.

Don't forget to test in the simulator or on a device when reviewing.

Squash whenever


Toggl.Foundation.MvvmCross/ViewModels/OnboardingViewModel.cs
```diff
@@ -20,22 +19,19 @@ public sealed class OnboardingViewModel : MvxViewModel
...
         private static readonly string[] pageNames =
         {
             nameof(TrackPage),
             nameof(MostUsedPage),
-            nameof(ReportsPage),
-            nameof(LoginPage)
+            nameof(ReportsPage)
         };
 
         private static readonly (MvxColor BackgroundColor, MvxColor BorderColor)[] PageInfo =
```
sofokli1: Is there a reason why there this array and the above instead of one array with a struct/tuple to hold the nameof, bacgroundcolor and bordercolor? I am not asking for changes just wondering.

Toggl.Foundation.MvvmCross/ViewModels/OnboardingViewModel.cs
```diff
@@ -44,16 +40,20 @@ public sealed class OnboardingViewModel : MvxViewModel
 
         private readonly bool[] pagesVisited = new bool[PageInfo.Length];
 
+        private bool visitedAllPages => pagesVisited.All(visited => visited);
+
         private int currentPage;
```
sofokli1: I can see from the code in the vc that `currentPage` is initialised there, but while reading this file that wan't very clear at first. Just speaking out loud 😄, no changes needed.

Toggl.Foundation.MvvmCross/ViewModels/OnboardingViewModel.cs
simonrozsival: We could do that in some future PR 👍

Toggl.Foundation.MvvmCross/ViewModels/OnboardingViewModel.cs
simonrozsival: You are right. It was previously explicitly set in the `Prepare` method. I now rely on the fact that it is 0 by default and only set the first page as visited in the `Initialize` method. It would be nicer to set it explicitly.

simonrozsival:
This is now ready to be reviewed.

PR #2176, opened 2018-05-17
Sync failures views

### LINKED ISSUE #1334: UI to see projects/clients/tags that failed to sync

Odrakir:
Closes #1334 

This probably still misses some stuff, but as it's going to work only for develop builds I say we start with this and work on it as needed.


Toggl.Foundation.MvvmCross/ViewModels/MainViewModel.cs
```diff
@@ -193,10 +198,16 @@ public void Init(string action)
                     RaisePropertyChanged(nameof(TimeEntriesCount));
                 });
 
+            var getNumberOfSyncFailuresDisposable = interactorFactory
+                .GetItemsThatFailedToSync().Execute()
+                .Select(i => i.Count())
+                .Subscribe(n => { NumberOfSyncFailures = n; });
```
simonrozsival: Please remove the curly braces, they are not needed.

Toggl.Foundation.MvvmCross/ViewModels/MainViewModel.cs
```diff
@@ -193,10 +198,16 @@ public void Init(string action)
                     RaisePropertyChanged(nameof(TimeEntriesCount));
                 });
 
+            var getNumberOfSyncFailuresDisposable = interactorFactory
+                .GetItemsThatFailedToSync().Execute()
```
simonrozsival: We put every method call on separate line in our codebase - so please break the line before the `.Execute()`.

Toggl.Foundation.Tests/MvvmCross/ViewModels/MainViewModelTests.cs
```diff
@@ -4,6 +4,7 @@
 using System.Linq;
 using System.Reactive.Linq;
 using System.Reactive.Subjects;
+using System.Runtime.CompilerServices;
```
simonrozsival: 👀

Toggl.Daneel/Views/SyncFailures/SyncFailureCell.cs
```diff
@@ -0,0 +1,43 @@
...
+        }
+
+        public void Update(SyncFailureItem model)
+        {
+            typeLabel.Text = model.Type.ToString() + ": ";
+            nameLabel.Text = model.Name;
+            syncStatusLabel.Text = model.SyncStatus.ToString();
+            errorMessageLabel.Text = model.SyncErrorMessage;
+            switch (model.SyncStatus) {
+                case PrimeRadiant.SyncStatus.SyncFailed:
```
kornelijepetak: Use the `using` in that file:
```
namespace Toggl.Daneel.Views.SyncFailures
{
    using static Toggl.PrimeRadiant.SyncStatus;  // add this line
```

then just ditch the prefix and just do 

`case SyncNeeded:` and 
`case SyncFailed: `

Toggl.Foundation.Tests/MvvmCross/ViewModels/MainViewModelTests.cs
```diff
@@ -514,6 +525,29 @@ public sealed class TheCurrentTimeEntryClientProperty : CurrentTimeEntryproperty
...
         }
 
+        public sealed class TheNumberOfSyncFailuresProperty : MainViewModelTest
+        {
+            [Fact, LogIfTooSlow]
+            public async ThreadingTask ReturnsTheCountOfInteractorResult()
+            {
+                var syncables = new IDatabaseSyncable[]
+                {
+                    new MockTag { Name = "Tag", SyncStatus = SyncStatus.SyncFailed, LastSyncErrorMessage = "Error1" },
```
kornelijepetak: Another candidate for `using static SyncStatus`

Toggl.Daneel/ViewSources/ListTableViewSource.cs
```diff
@@ -0,0 +1,34 @@
...
+using Foundation;
+using UIKit;
+
+namespace Toggl.Daneel.ViewSources
+{
+    public class ListTableViewSource<ModelType, CellType> : UITableViewSource where CellType : UITableViewCell
+    {
+        private readonly IList<ModelType> items;
+        private readonly string cellIdentifier;
+        private readonly Func<UITableViewCell, ModelType, CellType> configureCell;
```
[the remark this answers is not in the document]
Odrakir: Every time I see things like "handler" or "manager" in code I see someone having a hard time naming things. This is `configureCell` because that's exactly what this "method" (call it delegate or whatever) does. I could go as far as to call it `cellConfigurator`, but I find that using a noun makes the call feel less natural. I rather do `configureCell(whatever)` that `cellConfigurator(whatever)`

Toggl.Daneel/ViewSources/ListTableViewSource.cs
```diff
@@ -0,0 +1,34 @@
...
+        }
+
+        public override UITableViewCell GetCell(UITableView tableView, NSIndexPath indexPath)
+        {
+            var cell = tableView.DequeueReusableCell(cellIdentifier, indexPath);
+            var model = items[indexPath.Row];
+            return configureCell(cell, model);
+        }
+
+        public override nint RowsInSection(UITableView tableview, nint section)
```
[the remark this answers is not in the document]
Odrakir: WTF is EBM?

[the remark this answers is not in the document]
kornelijepetak: Expression-bodied member.
I gave an exact solution to my proposal.

Toggl.Daneel/ViewSources/ListTableViewSource.cs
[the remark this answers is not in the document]
kornelijepetak: I agree with your reasoning. My proposal was to name the Type `CellConfigurationHandler`, but the parameter itself could be called `configureCell` to make the call more natural.
However, even if named with handler, every delegate can also be invoked like this:
`var cell = configureCellHandler.Invoke(cell, model);`

But like I said, it's not necessary to change this. This is a lose-lose scenario.

Toggl.Foundation.Tests/MvvmCross/ViewModels/MainViewModelTests.cs
Odrakir: I'm going to let this one be, just because we are already using `PrimeRadiant` and (though a little more verbose) I think it's a bit clearer this way.

Toggl.Foundation.Tests/MvvmCross/ViewModels/MainViewModelTests.cs
```diff
@@ -157,7 +159,7 @@ await NavigationService.Received()
 
                 var expected = isInManualMode ^ flipManualModeChecks
                     ? TimeSpan.FromMinutes(DefaultTimeEntryDurationForManualModeInMinutes)
-                    : (TimeSpan?)null;
+                    : (TimeSpan?) null;
```
simonrozsival: 👀

Toggl.Foundation.Tests/MvvmCross/ViewModels/MainViewModelTests.cs
```diff
@@ -532,9 +538,14 @@ public sealed class TheNumberOfSyncFailuresProperty : MainViewModelTest
             {
                 var syncables = new IDatabaseSyncable[]
                 {
-                    new MockTag { Name = "Tag", SyncStatus = SyncStatus.SyncFailed, LastSyncErrorMessage = "Error1" },
-                    new MockTag { Name = "Tag2", SyncStatus = SyncStatus.SyncFailed, LastSyncErrorMessage = "Error1" },
-                    new MockProject { Name = "Project", SyncStatus = SyncStatus.SyncFailed, LastSyncErrorMessage = "Error2" }
+                    new MockTag {Name = "Tag", SyncStatus = SyncStatus.SyncFailed, LastSyncErrorMessage = "Error1"},
+                    new MockTag {Name = "Tag2", SyncStatus = SyncStatus.SyncFailed, LastSyncErrorMessage = "Error1"},
```
simonrozsival: 👀

Toggl.Daneel/Views/SyncFailures/SyncFailureCell.cs
```diff
@@ -5,6 +5,8 @@
 
 namespace Toggl.Daneel.Views.SyncFailures
 {
+    using static Toggl.PrimeRadiant.SyncStatus;
```
kornelijepetak: Unfortunately, @amulware did not allow me to use static usings inside the namespace in my PR(yet), so I will assume the same principle applies to this as well. Just pull it before the namespace declaration, next to other usings.

PR #2177, opened 2018-05-17
Delete TE metric

### LINKED ISSUE #1917: Delete TE metric
An event that is recorded when a time entry is deleted. Simple as that.

simonrozsival:
Closes #1917 

Squash whenever


Toggl.Foundation.MvvmCross/ViewModels/TimeEntriesLogViewModel.cs
```diff
@@ -232,10 +237,10 @@ private Task edit(TimeEntryViewModel timeEntryViewModel)
...
         private async Task delete(TimeEntryViewModel timeEntryViewModel)
         {
-            await dataSource
-                .TimeEntries
-                .Delete(timeEntryViewModel.Id)
-                .Do(_ => dataSource.SyncManager.PushSync());
+            await dataSource.TimeEntries.Delete(timeEntryViewModel.Id);
+
+            analyticsService.TrackDeletingTimeEntry();
+            dataSource.SyncManager.PushSync();
```
Odrakir: Why the change from Rx code to await? Is this the direction we want to go?  Also... As far as I understand with the previous syntax you only did `dataSource.SyncManager.PushSync()` when the deletion was successful, now you do it regardles.

simonrozsival: The `Do` thing seemed really unnecessary - I would need to add a second `Do` here or use a block `{ ... }` to put the two calls to both the analytics service and the sync manager. The `Do` simply made no sense, when the semantics are the same and the syntax is cleaner without so much LINQ chaining.

Odrakir: I'm sorry, I don't understand. What does await do if there's an error? Are you ok with doing the PushSync and sending the analytics event even if the deletion fails?

simonrozsival: If it fails, then an exception will be thrown. This is just a change in syntax, not a change in semantics.

Odrakir:
Yes, it's ok, I just wanted to understand it. The whole await <-> rx thing is new to me. (and I still like plain rx more)

simonrozsival:
@Odrakir please review again

PR #2179, opened 2018-05-17
Fix date picker bug

### LINKED ISSUE #2127: Start time on running time entry and editing existing time entry is displayed as UTC
When editing the Start time, by clicking on the Start time in the wheel menu, of a running or stopped time entry.

The time on the wheel is displayed as UTC.

image

arturs105:
Closes #2127

This is a workaround for the issue, because i couldn't find out why it was working for the end time, but not for the start time. More in-depth explanation can be found in the issue.

Squash whenever.


Toggl.Foundation.MvvmCross/ViewModels/EditDurationViewModel.cs
```diff
@@ -184,6 +187,7 @@ private void editStartTime()
             }
             else
             {
+                AboutToEditStartTime.Raise(this);
```
simonrozsival: We should test this.

simonrozsival:
What is the status of this PR? We had an internal discussion, but I am not sure what was the final decision - will you rewrite this so it uses observables instead of events? Or will it stay this way?

arturs105:
Yeah, i've already rewritten the thing locally, i just need to test it out before pushing.

simonrozsival:
OK 👌

Toggl.Daneel/ViewControllers/EditDurationViewController.cs
```diff
@@ -200,6 +204,13 @@ public override void ViewDidLayoutSubviews()
             adjustHeight();
         }
 
+        protected override void Dispose(bool disposing)
```
willsb: Please dispose your stuff here

PR #2183, opened 2018-05-17
Revert tag deserialization change

simonrozsival:
This is mostly reverting #2163. The only new change is the default value for `At` - it is now set to an obviously fake value (no real data in Toggl can have `At` in the year 2000).

Squash whenever


Toggl.Ultrawave/Models/Tag.cs
```diff
@@ -1,10 +1,13 @@
 ﻿using System;
 using Toggl.Multivac.Models;
+using Newtonsoft.Json;
 
 namespace Toggl.Ultrawave.Models
 {
     internal sealed partial class Tag : ITag
     {
+        private readonly DateTimeOffset defaultAt = new DateTimeOffset(2000, 01, 02, 03, 04, 05, TimeSpan.Zero);
```
amulware: Static please.

PR #2187, opened 2018-05-18
Add vibrations on popup temporal inconsistencies

kornelijepetak:
This PR adds a vibration feedback when user fails to respect the inherent properties of the space-time continuum and tries to move the start time after stop time or stop time before the start time.

This depends on #2111 so it should be merged first.

Time


Toggl.Giskard/Helper/VibratorExtensions.cs
```diff
@@ -0,0 +1,27 @@
...
+    internal static class VibratorExtensions
+    {
+        public static void ActivateVibration(this Vibrator vibrator, int duration)
+        {
+            if (!vibrator.HasVibrator)
+                return;
+
+            try
+            {
+                if (Build.VERSION.SdkInt >= BuildVersionCodes.O)
```
willsb: Add this to our Api checking helpers

PR #2188, opened 2018-05-18
Prevent the duration from exceeding 999 hours

kornelijepetak:
Implements the 999 hours constraint on duration in popup.

🚧 Also refactors the way events are generated for simplicity/readability

Branched from #2111, will rebase on `develop` after #2111 is merged.

### DIFF
--- a/Toggl.Foundation.MvvmCross/ViewModels/SelectTimeViewModel.cs
+++ b/Toggl.Foundation.MvvmCross/ViewModels/SelectTimeViewModel.cs
@@ -20,6 +20,13 @@ namespace Toggl.Foundation.MvvmCross.ViewModels
         public const int StopTimeTab = 1;
         public const int DurationTab = 2;
 
+        public enum TemporalInconsistency
+        {
+            StartTimeAfterStopTime,
+            StopTimeBeforeStartTime,
+            DurationTooLong
+        }
+
         private readonly IMvxNavigationService navigationService;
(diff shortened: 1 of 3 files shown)


Toggl.Foundation.MvvmCross/ViewModels/SelectTimeViewModel.cs
```diff
@@ -51,8 +58,7 @@ public sealed class SelectTimeViewModel
 
         public DateTimeOffset StopTimeOrCurrent => StopTime ?? CurrentDateTime;
 
-        public event EventHandler StopTimeBeforeStartTime;
-        public event EventHandler StartTimeAfterStopTime;
+        public event EventHandler<TemporalInconsistency> TemporalInconsistencyDetected;
```
juanlaube: Why is this an event handler? Feels weird to me triggering events from a view model, I'd expect that on the view side, but not here. Shouldn't this be an Observable maybe?

kornelijepetak: Events in C# are first class observer pattern.
I don't think events need to be used only in views.
However, I agree that using RX here would be better and more consistent so I will change this.

kornelijepetak: Done.

PR #2193, opened 2018-05-18
Replace interactorFactory with interactor

sofokli1:
Doesn't close any issue

Replace interactorFactory with interactor in selectCountry

🦑 whenever


Toggl.Foundation.Tests/MvvmCross/ViewModels/SelectCountryViewModelTests.cs
```diff
@@ -40,14 +40,13 @@ protected override SelectCountryViewModel CreateViewModel()
         public sealed class TheConstructor : SelectCountryViewModelTest
         {
             [Theory, LogIfTooSlow]
-            [ClassData(typeof(TwoParameterConstructorTestData))]
-            public void ThrowsIfAnyOfTheArgumentsIsNull(bool useInteractorFactory, bool useNavigationService)
+            [ClassData(typeof(OneParameterConstructorTestData))]
```
simonrozsival: This attribute is not needed - the `useNavigationService` will always be `false`. You can just pass `null` directly to the constructor.

Issue #2169, opened 2018-05-17
Sign up UI: show an error when the country is missing


pe0ny:
If the user still taps the button, and the country is not selected, it should show a little exclamation next to it.

### Missing country
![toggl - ios - sign up missing country 1x](https://user-images.githubusercontent.com/5064445/40156926-dbf629da-59a4-11e8-8470-8a715ea63e90.png)

Zeplin: https://zpl.io/anm5x1J

Issue #1960, opened 2018-04-25
Sign up flow changes: New Sign up UI


pe0ny:
- New Sign up UI is accessible from the Log in page.
- New Sign up UI contains all needed fields and inputs: Email and Password field, Country selector, Google Sign up, and option to switch back to Log in page.
- By default, all the fields are not focused.
- Error messages are shown below the Password field (if the message is longer then in two rows).
- When the form can't be submitted, disable the main "Sign up for free" button. Text opacity = 48%
- Tapping Country dropdown at the top of the screen opens Country of Residence view #1610
- Form won't be submitted after tapping "Sign up for free", instead ToS & PP popup will be opened #1611 after which we can post it to BE.
- When Email or Password field is focused or when user has already typed something in, then the placeholder moves up.
- Bottom card is always positioned at the bottom of the screen and directs user back to the Log in page. The whole card should be tappable.
- Everything else is positioned at the top of the screen:
   - On iphone SE sized screens content starts 24pt from the nav bar (country dropdown)
   - On bigger screens content starts 72pt from the nav bar

### Sign up page
![toggl - ios - sign up 1x](https://user-images.githubusercontent.com/5064445/39251132-3a652544-48ab-11e8-926d-1076ffabcfdd.png)

Zeplin: https://zpl.io/2vl6LKr

### Active field & Error message
![toggl - ios - sign up keyboard error message 1x](https://user-images.githubusercontent.com/5064445/39252111-98c85622-48ad-11e8-9b25-02e13255b350.png)

Zeplin: https://zpl.io/VqxY8Wl

amulware:
Note that the country is non-optional, and we have to make sure we always post a valid one when it comes to the signup endpoint. This means that in order to go through: fetching countries has to work AND (fetching preselected country has to work OR user has to select country manually)

amulware:
Note further that tapping the google signup button will first show the TOS/PP popup, and only direct to google upon agreement by the user. Further, we also send the currently selected country with the signup request when signup up via Google.

pe0ny:
Sign up spec: https://docs.google.com/document/d/1mWz4tw8nA6k7aQrxK0r52dvetjazWjbgmzFc-Zbr_Sc/edit?usp=sharing

amulware:
See changes here: https://github.com/toggl/mobileapp/issues/1610#issuecomment-387321024

PR #2195, opened 2018-05-18
Add location api

arturs105:
Ref #2169
Ref #1960

I was kinda split between using `string`s for country name and code vs `ICountry`, but i went with strings to keep it closer to the json that backend returns.

Squash whenever.


Toggl.Ultrawave/Models/Location.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System;
+using Newtonsoft.Json;
+using Toggl.Multivac.Models;
+
+namespace Toggl.Ultrawave.Models
+{
+    internal sealed partial class Location : ILocation
+    {
+        [JsonProperty("city")]
```
simonrozsival: This shouldn't be needed.

Toggl.Ultrawave/Models/Location.cs
```diff
@@ -0,0 +1,21 @@
...
+        public string City { get; set; }
+
+        [JsonProperty("state")]
```
simonrozsival: This shouldn't be needed.

Toggl.Ultrawave/Models/Location.cs
```diff
@@ -0,0 +1,21 @@
...
+        public string State { get; set; }
+
+        [JsonProperty("country_name")]
+        public string Country { get; set; }
```
simonrozsival: This property should be called `CountryName` so the property is not needed.

amulware:
`docs/endpoints.md` is lonely :troll:

PR #2186, opened 2018-05-17
Add "this week" default selection to the Reports Calendar VM

daividssilverio:
Closes #1941

I tested on Android (emulator/ actual device) and on the iOS emulator.
It would be nice to test this on actual iOS device as well.

It works nicely on Android, but the calendar is looking weird on iOS:

PR #2201, opened 2018-05-18
Fix reports calendar day cell masking layer

### LINKED ISSUE #2199: Reports calendar: "this week" first day of default selection highlighted area glitch
This appeared when I was trying the reports calendar after changing Foundation to setup the "first week" default selection.

ref #2186

daividssilverio:
Closes #2199

ref #2186


simonrozsival:
It seems the problem isn't there anymore, but I don't understand how this code fixed it - could you add some more info please?

PR #2205, opened 2018-05-21
Fix me/location API

### LINKED ISSUE #2204: LocationApiTests.ReturnsNonEmptyLocation includes wrong assertion
The test fails for IP addresses not specific enough to return a city/state. For example, I get:

`{"city":"","state":"","country_name":"Ireland","country_code":"IE"}`

So city and state should be optional.

We can still check that the properties are not null, just to make sure the endpoint continues specifying them, even if empty.

simonrozsival:
Closes #2204 

We only need the information about the country, we don't really care about the city/state.

We also never push the country model to the server, so serialization is now "disabled".

Integration tests: https://www.bitrise.io/build/bb5224d019c2ec37

Squash whenever


simonrozsival:
@amulware 🤔 I think we could remove that in the future and squash this now.

willsb:
I'd leave the country name in for debugging purposes. It's not like we're losing anything by doing so.

amulware:
Go ahead. /shrug

PR #2219, opened 2018-05-22
Do not display tooltips before the previous steps are finished

### LINKED ISSUE #2029: Step 2 and Step 4 displaying simultaneously
Fresh install, new signup.
Proceeded through onboarding to step 2, didn’t click on entry to edit and just created more entries.
Step 2 remained on screen and step 4 was also displayed. 
Not sure if this is functioning as intended.
41cf54b1-44ed-4f54-bacc-1fa11bff6ce6
351c9a4c-19dc-4591-8a25-ef79db9c5300

simonrozsival:
Closes #2029

Squash whenever


amulware:
You say 'previous steps' - does this enforce an order to the onboarding tooltips? Or does it just make sure that no two of them can be active at once?

simonrozsival:
This just makes sure that they can't be shown at once - it doesn't handle the case in which the "previous steps" were hidden and they haven't been displayed yet - it will still be possible to display the "next" state - it would be hidden if the "previous" step was displayed though.

Maybe I can change the naming of the variables so it is not confusing.

PR #2220, opened 2018-05-22
Avatar image

### LINKED ISSUE #1629: Settings UI - Avatar
I have no idea what we need to do with this one. 
Get it from backend or what?
Whatever we need to do, this is the issues for that.

Ref #1268

kornelijepetak:
Closes #1629 

This fetches an avatar image from the server.

Note, however that this is `WIP` and that I need some architectural advice on best practices in following areas:

* Should I store the image as a file asset in the internal storage so that I only fetch it if `image_url` has changed.
* Where to store the information about the last known url so that I can detect changes in image

Also any other comment or idea is much appreciated.

🦑 whenever.


kornelijepetak:
This is not a priority for beta, so @willsb and @daividssilverio please take a look and give advice but don't waste much time on this at this point.

Toggl.Giskard/Activities/SettingsActivity.cs
```diff
@@ -26,7 +36,29 @@ protected override void OnCreate(Bundle bundle)
...
+        {
+            if (string.IsNullOrEmpty(ViewModel.ImageUrl))
+                return;
+
+            avatarBitmap = await ImageUtils.GetImageFromUrl(ViewModel.ImageUrl);
+
+            if (avatarBitmap == null)
+                return;
+
+            RunOnUiThread(() =>
```
willsb: Since you are not using `ConfigureAwait(false)`, this is not needed. `setupAvatar` is being called `OnCreate`, which always runs on the UIThread.

Toggl.Giskard/Activities/SettingsActivity.cs
```diff
@@ -26,7 +36,29 @@ protected override void OnCreate(Bundle bundle)
 
             OverridePendingTransition(Resource.Animation.abc_fade_in, Resource.Animation.abc_fade_out);
 
+            avatarView = FindViewById<ImageView>(Resource.Id.SettingsViewAvatarImage);
+            avatarContainer = FindViewById<LinearLayout>(Resource.Id.SettingsViewAvatarImageContainer);
+
             setupToolbar();
+
+            Task task = setupAvatar();
```
willsb: I see you are not awaiting this. Can I suggest observables, since Android is not giving you any awaitable points?

Toggl.Giskard/Resources/layout/SettingsActivity.axml
```diff
@@ -10,11 +10,27 @@
         android:layout_height="match_parent"
         android:layout_marginTop="?attr/actionBarSize">
         <LinearLayout
+            android:animateLayoutChanges="true"
             android:paddingTop="24dp"
             android:orientation="vertical"
             android:background="@color/lightGray"
             android:layout_width="match_parent"
             android:layout_height="match_parent">
+            <LinearLayout
```
willsb: An explanation of why you included this container and not just the imageview would be nice

Toggl.Giskard/Activities/SettingsActivity.cs
kornelijepetak: Yeah, an observable would make sense here. Thanks for suggestions...

willsb:
@kornelijepetak This is not much work and will reduce our PR count by one. Can you please submit the fixes so we can close this guy? :)

kornelijepetak:
@willsb the LinearLayout container has to stay as it seems to be the only way to horizontally center an image (as far as I could find out by trial and error and also by [this](https://stackoverflow.com/a/36366912/93770) and some other sources).

Now, let's merge this PR. But please if you have a minute, give some suggestions (or maybe open another issue with that information) about how to deal with the concerns I mentioned in the PR description. Mostly, let us not fetch the image every time settings screen is open. If you think this should be done before merging, I will return the WIP label and fix it tomorrow. But I'd really like some advice on that.

PR #2222, opened 2018-05-22
Add location picking to signup

### LINKED ISSUE #2169: Sign up UI: show an error when the country is missing

arturs105:
Closes #2169

Do a functional test as well.

Squash whenever.


Toggl.Foundation.MvvmCross/ViewModels/SignupViewModel.cs
```diff
@@ -91,6 +108,25 @@ public override void Prepare(CredentialsParameter parameter)
...
         }
 
+        public override async Task Initialize()
+        {
+            await base.Initialize();
+
+            allCountries = await new GetAllCountriesInteractor().Execute();
+
+            var api = apiFactory.CreateApiWith(Credentials.None);
+            new GetCurrentLocationInteractor(api)
```
simonrozsival: Awaiting this would be much nicer than the subscription IMO.

Toggl.Foundation.MvvmCross/ViewModels/SignupViewModel.cs
```diff
@@ -166,9 +202,27 @@ private void togglePasswordVisibility()
...
+            SelectCountryParameter selectedCountryParameter;
+            if (countryId == null)
+            {
+                selectedCountryParameter = await navigationService.Navigate<SelectCountryViewModel, SelectCountryParameter>();
+            }
+            else
+            {
+                selectedCountryParameter = await navigationService
+                    .Navigate<SelectCountryViewModel, SelectCountryParameter, SelectCountryParameter>(
+                        SelectCountryParameter.With(countryId.Value));
```
simonrozsival: Do we need this if? It would make sense to make the `SelectCountryParameter` accept `long?` => you could directly pass `countryId`.

Toggl.Foundation.MvvmCross/ViewModels/SignupViewModel.cs
arturs105: That would require a `try catch` and i still have some ptsd from mixing those with async code.

simonrozsival: Ok. But then let's keep the disposable and also dispose it.

Toggl.Foundation.MvvmCross/ViewModels/SignupViewModel.cs
```diff
@@ -44,7 +54,11 @@ public sealed class SignupViewModel : MvxViewModel<CredentialsParameter>
 
         public bool IsPasswordMasked { get; private set; } = true;
 
-        public bool SignupEnabled => Email.IsValid && Password.IsValid && !IsLoading;
+        public bool SignupEnabled
+            => Email.IsValid
+            && Password.IsValid
+            && !IsLoading
+            && countryId != null;
```
amulware: I'd prefer `.HasValue`

Toggl.Daneel/ViewControllers/SignupViewController.cs
```diff
@@ -80,6 +83,10 @@ public override void ViewDidLoad()
                       .For(v => v.BindFirstResponder())
                       .To(vm => vm.IsShowPasswordButtonVisible)
                       .Mode(MvxBindingMode.OneWayToSource);
+
+            bindingSet.Bind(CountryNotSelectedImageView)
+                      .For(v => v.BindAnimatedVisibility())
+                      .To(vm => vm.FailedToGetCountry);
```
amulware: If I understand correctly, it shows this button if the location endpoint fails (to return a country) for some reason.

The issue specifies that it should appear if the user taps the sign up button and no country is selected, not before (I don't think we want to scare people with the red exclamation point before it's necessary).

I suppose that also makes disabling the signup button if no country is set impossible, and a check has to be put in the `signup` method?

In the majority of cases this doesn't matter, but there are two cases where it might:

1. location endpoint fails -> showing red right away might scare the user (they might have picked a country without the hint just fine)
2. the location endpoint takes a long time to respond (crappy mobile internet, phone is offline but not recognizing it, so we're stuck in a long timeout) -> user cannot do anything if they don't notice the country selection at the top (since the signup button is disabled)

amulware: If you think the way it's implemented is better, please argue though. (Also educate me if I misunderstood - I haven't written any UI code in... some time.)

Toggl.Foundation.MvvmCross/ViewModels/SignupViewModel.cs
```diff
@@ -29,8 +35,12 @@ public sealed class SignupViewModel : MvxViewModel<CredentialsParameter>
 
         private IDisposable signupDisposable;
         private bool termsOfServiceAccepted;
+        private List<ICountry> allCountries;
+        private long? countryId;
 
-        public ICountry Country { get; private set; } = new Country("Estonia", "EE", 1);
+        public string CountryButtonTitle { get; private set; } = Resources.SelectCountry;
+
+        public bool FailedToGetCountry { get; private set; } = false;
```
amulware: The name seems leaky to me. What about `IsCountryPickerHintVisible`? (or maybe something similar but shorter, can't think of anything right now)

Toggl.Foundation.MvvmCross/Toggl.Foundation.MvvmCross.csproj
```diff
@@ -25,4 +25,7 @@
   <ItemGroup>
     <Folder Include="ViewModels\Settings\" />
   </ItemGroup>
+  <ItemGroup>
+    <Compile Remove="Parameters\SelectCountryParameter.cs" />
```
simonrozsival: 👀

Toggl.Foundation.MvvmCross/ViewModels/SignupViewModel.cs
```diff
@@ -166,9 +206,19 @@ private void togglePasswordVisibility()
...
-        }
+            var selectedCountryId = await navigationService
+                .Navigate<SelectCountryViewModel, long?, long?>(countryId);
+
+            if (selectedCountryId == null) return;
+
+            var selectedCountry = allCountries
+                .Single(country => country.Id == selectedCountryId.Value);
+            
+            FailedToGetCountry = false;
```
amulware: This is an odd line given the current naming (see above).

Toggl.Foundation.MvvmCross/ViewModels/SignupViewModel.cs
```diff
@@ -91,6 +108,29 @@ public override void Prepare(CredentialsParameter parameter)
...
+                .Select(location => allCountries.Single(country => country.CountryCode == location.CountryCode))
+                .Subscribe(
+                    setCountry,
+                    _ => FailedToGetCountry = true,
+                    () => getCountrySubscription.Dispose());
+        }
+
+        private void setCountry(ICountry country)
+        {
+            countryId = country.Id;
```
simonrozsival: I don't think this is the correct behavior - if the user selected the country himself, it shouldn't be overwritten.

Toggl.Foundation.MvvmCross/ViewModels/SignupViewModel.cs
```diff
@@ -103,7 +143,7 @@ public override void Prepare(CredentialsParameter parameter)
 
             signupDisposable =
                 loginManager
-                    .SignUp(Email, Password, true, (int)Country.Id)
+                    .SignUp(Email, Password, true, (int)countryId)
```
amulware: Here I'd prefer `.Value` instead of the cast to make more explicit what's going on (it was not obvious to me that `countryId` was nullable at first).

Toggl.Foundation.Tests/Interactors/Location/GetCurrentLocationInteractorTests.cs
```diff
@@ -0,0 +1,26 @@
...
+    {
+        public sealed class TheExecuteMethod
+        {
+            [Fact, LogIfTooSlow]
+            public async Task ReturnsTheLocationApiResponse()
+            {
+                var api = Substitute.For<ITogglApi>();
+                var interactor = new GetCurrentLocationInteractor(api);
+
+                await interactor.Execute();
```
simonrozsival: You should test the interface of the interactor, not its internal implementation - set the return value of the API and assert the returned value by the interactor - check that they match - instead of checking if the API was called (what if that value is not returned, what if completely different value is returned, but the API call is still done?).

Toggl.Daneel/ViewControllers/SignupViewController.cs
arturs105: Yeah, i misunderstood the issue, will fix.

arturs105: Wait, i reread the issue and now i'm confused - the title says "show an error when the country is missing". Calling @pe0ny for clarification.

pe0ny: I'm not sure what is the question?
Description says: 
> If the user still taps the button, and the country is not selected, it should show a little exclamation next to it.

arturs105: Does "The button" mean country button or signup button?

kornelijepetak: The signup button. Meaning you are trying to signup without setting something up. This exclamation would show what needs to be setup in order to proceed.

Toggl.Foundation.MvvmCross/ViewModels/SignupViewModel.cs
kornelijepetak: I also wonder why initialize the property to a value it already has by default?

Toggl.Foundation.MvvmCross/ViewModels/SignupViewModel.cs
kornelijepetak: even if it was `long?` please in general, try and make things clearer by doing `(int)countryId.Value` to both make it clear that it's nullable and that you are casting to another primitive type.

Toggl.Foundation.MvvmCross/ViewModels/SignupViewModel.cs
kornelijepetak: I'd even go that far as to suggest naming this variable in the positive way.
`IsCountryCorrectlySet` or something and invert it in the binding...

Toggl.Daneel/ViewControllers/SignupViewController.cs
arturs105: I think i get it now. If the location api fails to get the country, we still allow the user to press the signup button and we show the error after he presses it.

But now i have an argument to keep things as they are currently - the error indicator is shown right after the api fails and we don't even allow the user to press the signup button. The argument is consistency - I think this keeps the country input consistent with the email and password fields, because we don't allow the user to press the button with invalid (empty) email or password.

Now what if the user doesn't notice that the country is missing and tries to push the grayed out signup button, but nothing happens? In that case we can add a shake animation to the invalid field. The same shake animation, when you enter a wrong password while logging into your mac.

[This isn't the greatest feature in the world, this is just a suggestion](https://tonedeaf.com.au/wp-content/uploads/2018/05/tenaciousd.jpg), so i'm fine with changing to what the issue says.

arturs105: Yep, that was my suggestion. And if we're doing this to the country button, it makes sense to add the same thing to email and password fields as well.

Toggl.Foundation.MvvmCross/ViewModels/SignupViewModel.cs
amulware: Personally, I appreciate the clarity.

Toggl.Foundation.MvvmCross/ViewModels/SignupViewModel.cs
```diff
@@ -29,8 +35,12 @@ public sealed class SignupViewModel : MvxViewModel<CredentialsParameter>
 
         private IDisposable signupDisposable;
         private bool termsOfServiceAccepted;
+        private List<ICountry> allCountries;
+        private long? countryId;
 
-        public ICountry Country { get; private set; } = new Country("Estonia", "EE", 1);
+        public string CountryButtonTitle { get; private set; } = Resources.SelectCountry;
+
+        public bool CountryErrorVisible { get; private set; } = false;
```
kornelijepetak: Can I please ask you to turn this into `IsCountryErrorVisible` for consistency?

Toggl.Foundation.MvvmCross/ViewModels/SignupViewModel.cs
```diff
@@ -91,6 +107,34 @@ public override void Prepare(CredentialsParameter parameter)
...
+            try
+            {
+                await new GetCurrentLocationInteractor(api)
+                    .Execute()
+                    .Select(location => allCountries.Single(country => country.CountryCode == location.CountryCode))
+                    .Do(setCountry);
+            }
+            catch
+            {
+                IsCountryErrorVisible = true;
```
amulware: This should only be called if the country is not set.

Toggl.Foundation.MvvmCross/ViewModels/SignupViewModel.cs
```diff
@@ -114,27 +115,31 @@ public override void Prepare(CredentialsParameter parameter)
...
-                    .Do(setCountry);
-            }
-            catch
-            {
-                IsCountryErrorVisible = true;
-            }
+            getCountrySubscription = new GetCurrentLocationInteractor(api)
+                .Execute()
+                .Select(location => allCountries.Single(country => country.CountryCode == location.CountryCode))
+                .Do(setCountryIfNeeded)
```
amulware: No double dipping! 🔥 :P

Toggl.Foundation.MvvmCross/ViewModels/SignupViewModel.cs
```diff
@@ -166,9 +215,22 @@ private void togglePasswordVisibility()
...
-            //The dialog is not implemented yet
-            await Task.CompletedTask;
-        }
+            getCountrySubscription?.Dispose();
+            getCountrySubscription = null;
+
+            var selectedCountryId = await navigationService
+                .Navigate<SelectCountryViewModel, long?, long?>(countryId);
+
+            if (selectedCountryId == null) return;
```
amulware: In this case, I'd like to also
```
setCountryErrorIfNeeded(null);
```
so that the user realises they really have to do this.

arturs105: Do you mean something like this?
```
if (selectedCountryId == null)
{
    setCountryErrorIfNeeded(null);
    return;
}
```

amulware: Yeah, that's what I mean - assuming it makes sense to you?

arturs105: Yeah, i think it does.

PR #2223, opened 2018-05-22
Filter deleted time entries and archived projects from suggestions

simonrozsival:
Ref #2208 

The only problem now is that we don't sync archived projects... which will be fixed "soon". Then these projects won't be suggested anymore.

Squash whenever


Toggl.Foundation/Suggestions/MostUsedTimeEntrySuggestionProvider.cs
```diff
@@ -31,18 +30,21 @@ public sealed class MostUsedTimeEntrySuggestionProvider : ISuggestionProvider
...
+                       .GetAll(isSuitableForSuggestion)
                        .SelectMany(mostUsedTimeEntry)
                        .Take(maxNumberOfSuggestions);
 
-        private bool isWithinThresholdAndHasDescription(ITimeEntry timeEntry)
-        {
-            if (string.IsNullOrEmpty(timeEntry.Description))
-                return false;
+        private bool isSuitableForSuggestion(IDatabaseTimeEntry timeEntry)
+            => string.IsNullOrEmpty(timeEntry.Description) == false
```
juanlaube: I'd use `!string.IsNullOrEmpty(timeEntry.Description)` vs `string.IsNullOrEmpty(timeEntry.Description) == false`

PR #2226, opened 2018-05-22
Add retryWhen operator

Odrakir:
The implementation is [not mine](https://github.com/dotnet/reactive/issues/449), but I've read it, tested it and makes sense and seems to work.

Ideally, somebody would merge [this PR](https://github.com/dotnet/reactive/pull/486) with the "official" implementation, but until then, this is what we have.

This will come handy for #1796, the suggested implementation for the retry with backoff strategy operator could be something like this:

```c#
public static IObservable<T> RetryWithBackOffStrategy<T>(
    this IObservable<T> source,
    int maxRetries,
    Func<int, TimeSpan> backOffStrategy,
    IScheduler scheduler)
{
    return source.RetryWhen(errors =>
        errors.SelectMany((error, index) =>
            {
                if (index > maxRetries) {
                    throw error;
                }

                return Observable.Interval(backOffStrategy(index), scheduler);
           }
       )
   );
}
```

I leave to @daividssilverio the decision whether to create another extension method or use it as it is.


Toggl.Giskard/Toggl.Giskard.csproj
```diff
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
```
willsb: Not needed

Odrakir: Not needed means it's going to break if I leave it? I'm asking because something (I'm guessing Rider) keeps adding that.

willsb: And VS will remove it, so 🤷‍♂️

Odrakir: I hate everything

PR #2227, opened 2018-05-22
Group small pie chart segments

### LINKED ISSUE #1800: Group smaller projects into a single segment on reports pie chart
Right now, we are using the `Segments` property of the `ReportsViewModel` both for plotting the Pie Chart and for showing the Legend. What we should do instead is having a property like `LegendItems` that has all the segments and perform a transformation on the segments that will be used in the pie chart to group them according to the rules stated in #1622.

I'm opening this as a separate issue because it will affect iOS as well, so we're better off doing this after the basic reports functionality is stable on Android.

simonrozsival:
Closes #1800 

This PR groups chart segments which have < 10% into a single segment called "Other".

I have one question: what if the user has a project called "Other" - wouldn't it be confusing then? :)

Please test Android if I didn't break it.

Squash whenever


Toggl.Foundation.MvvmCross/ViewModels/ReportsViewModel.cs
```diff
@@ -49,26 +50,39 @@ public sealed class ReportsViewModel : MvxViewModel<long>
...
-                    otherProjects.Sum(project => project.Percentage),
-                    otherProjects.Sum(project => (float)project.TrackedTime.TotalSeconds),
-                    otherProjects.Sum(project => project.BillableSeconds),
-                    Color.Reports.OtherProjectsSegmentBackground.ToHexString(),
-                    DurationFormat);
-
-                return segments
-                    .Where(segment => segment.Percentage >= minimumPieChartSegmentPercentage)
-                    .Append(otherSegment);
+                if (groupedSegments == null)
```
sofokli1: why not?
```csharp
if (groupedSegments != null)
{
    return groupedSegments;
}
...
```
so you don't have all that logic indented for nothing

sofokli1: not a request but a suggestion. Maybe split this computed var into 2 functions, one public for the `get` part and one private for the `set` part.

Issue #2022, opened 2018-05-03
New Log in UI


pe0ny:
- [x] New Log in UI replaces the currently in use last onboarding step (blue screen with illustrations and login/signup buttons).
- [x] New Log in UI contains all needed fields and inputs: Email and Password field, Password manager, Forgot password, Google Log in, and option to switch to Sign up page.
- [x] By default, all the fields are not focused.
- [ ] The error message should show below the associated field (if we know which field is wrong). Focus should be kept in that field still
_e.g. Not a valid email address error, should appear below the email field_
- [ ] If the error is not associated with a particular field, then we should remove focus from all fields and display the error under the Password field
_e.g. We couldn't log you in error should appear below Password field as in the mockup_
- [x] When the form can't be submitted, disable the main "Log in" button. Text opacity = 48%
- [ ] Use default input field styling with label and error messages – adjust colors to match app colors.
   *e.g. When Email field is focused or when user has already typed something in, then the placeholder moves up.*
- [x] Bottom card is always positioned at the bottom of the screen and directs user back to the Sign up page. The whole card should be tappable.
- [ ] When the keyboard is closed padding from top of screen is 88dp
- [ ] When the keyboard is open padding from top of screen is 48dp
*The goal is to keep "Forgot your password link" above the keyboard whenever possible, but still have a top/bottom balanced layout on the Login screen when the keyboard is closed*

### Log in page
![log in](https://user-images.githubusercontent.com/5064445/39583552-0706bbdc-4ef9-11e8-89ce-5386becca232.png)

Zeplin: https://zpl.io/25yo4Y0

### Error message
![log in - step2 error2](https://user-images.githubusercontent.com/5064445/39583802-96fbcb56-4ef9-11e8-8d0a-122f461a41f0.png)

Zeplin: https://zpl.io/a8EAPY6

### Loading state
![log in - step2 password entered](https://user-images.githubusercontent.com/5064445/39583845-af14d76e-4ef9-11e8-87d1-f6aa0cad0f2a.png)

Zeplin: https://zpl.io/an6PYWl

arturs105:
Please work on this only after #2018 has been merged, since it contains `Foundation` code for the new login view.

arturs105:
[...]

PR #2228, opened 2018-05-22
New login for droid

willsb:
This implements the new login screen. Ref #2022 

Things missing:

- The error reporting needs to be redone, for it doesn’t behave like specs not even on iOS
- Changing padding when the keyboard changes doesn’t sound like a good idea, given how keyboards are weird and behave poorly. It looks good as is, so I’ll discuss with designers to see what gives.
- _When Email field is focused or when user has already typed something in, then the placeholder moves up_: Since this is not the default behaviour of the floating tooltip, I’ll discuss about whether this should be included or not.

Squash as soon as approved.


Toggl.Foundation.MvvmCross/App.cs
```diff
@@ -9,7 +9,8 @@
 
 namespace Toggl.Foundation.MvvmCross
 {
-    public sealed class App : MvxApplication
+    public sealed class App<TFirstViewModelWhenNotLoggedIn> : MvxApplication
```
kornelijepetak: I think this is definitely the longest generic type I've ever seen.
Please do not change it. I like descriptive names. I hate all the `T`s and `U`s.

Toggl.Foundation.MvvmCross/ViewModels/NewLoginViewModel.cs
```diff
@@ -1,214 +0,0 @@
-﻿using System;
```
kornelijepetak: So long, dear old new LoginViewModel.

Toggl.Giskard/Resources/layout/OnboardingActivity.axml
```diff
@@ -1,76 +1,7 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<RelativeLayout
-    xmlns:tools="http://schemas.android.com/tools"
+<?xml version="1.0" encoding="utf-8"?>
+<RelativeLayout xmlns:tools="http://schemas.android.com/tools"
     xmlns:local="http://schemas.android.com/apk/res-auto"
     xmlns:android="http://schemas.android.com/apk/res/android"
     android:background="@drawable/OnboardingGradient"
     android:layout_width="match_parent"
-    android:layout_height="match_parent">
```
kornelijepetak: `And good riddance...` 😛

willsb:
A followup issue has been created for the logout problem

PR #2230, opened 2018-05-22
Fix navigation spacings

### LINKED ISSUE #2054: Navbar paddings are off
- Navbar's height should be 48pt, atm it's bigger
- Paddings between icons are off

Please, check zeplin for correct paddings: https://zpl.io/ag9Dd1R

sofokli1:
Closes #2054

The changes may look trivial but it took me a long time to understand that the navigation bar height was being faked. If I had the 3d view hierarchy from xcode I would have made this PR a long time ago.

Also made the navigation buttons a bit thinner so functional testing with fingers (no simulator) is needed. Also please test on iOS 10.

I removed the negative fixedSpace because it is deprecated in iOS 11.

🦑  whenever


sofokli1:
@pe0ny if the `after` screenshots looks acceptable give a thumbs up

Before             |  After
:-------------------------:|:-------------------------:
simulator screen shot - iphone 8 - 2018-05-22 at 21 32 39 | simulator screen shot - iphone 8 - 2018-05-22 at 21 31 53

PR #2235, opened 2018-05-23
Fix debug configuration

Odrakir:
First I deleted some configurations that were incorrect according to Rider, then I removed all testing targets from the "Debug | iPhoneSimulator" configuration, which makes it work for Rider.

I've tested in VS as well and it still works.

Some more cleanup in the configurations should be done, I don't think we need all those, do we? But that's for another PR and for someone who knows about it.


Odrakir:
Also, @willsb What about @amulware concerns of this removing the App Store configuration?

amulware:
Yeah, I can't let this go though. I'm not saying that what we have right now makes much more sense (why is daneel building for PLAYstore and api test builds? why are there duplicate entries?). But this seem to remove daneel building for app store builds as far as I can tell.

The best idea that I have is to go over all this manually and fixing it up that way, hoping that doesn't break anything. Or maybe the editor inside VS can actually do this properly, but that's how the currently weird state came to be at all.

I'm very confused by this. I'll have a look at the configs on develop myself...

amulware:
Ok, a bit problem with the configurations is that VS insists that it needs to add Release.PlayStore|IPhone(Simulator), etc. (even for unit tests and stuff) which just clutters things up to no end.

And I think it keeps re-adding these every time you open the build configs and make changes. <insert facepalm emoji that github has a lot of trouble with because of silly gendered emoji>

Odrakir:
I would suggest doing this in Rider and not touching them in VS. Ideally don't touch anything in VS.

Odrakir:
Probably a lot of this is caused by VS, but here are my comments:
- We don't need to differentiate Simulator/Device builds, that's another step.
- We don't need Testflight (should be exactly the same as AppStore)
- We might need Debug.Daneel and Debug.Giskard though
- The same for Release and AdHoc builds

So, in my opinion, the list should be:
- ApiTests
- Debug.Daneel
- Debug.Giskard
- Release.Daneel
- Release.Giskard
- AdHoc.Daneel
- AdHoc.Giskard
- UITests.Daneel
- UITests.Giskard
- UnitTests

amulware:
I think creating that list (possibly from scratch) is worth a try. However we also need to update a lot of stuff in cake and bitrise, so it's a bit of a bigger thing.

Odrakir:
I'm closing this to clean up all configurations in another PR.

PR #2237, opened 2018-05-23
Integrate with Facebook

### LINKED ISSUE #2049: Integrate with Facebook
We want minimal integration only, just so that we can measure the effectiveness of Facebook ads.

There is a lot of useful information here: #1923

Please coordinate the integration with @djalex15 and make sure that all secret keys etc. are treated appropriately.

Please also test the integration in debug mode, just to make sure that whatever we are sending gets through correctly.

Odrakir:
Closes #2049 

Still missing the part where we configure everything in Facebook's developers' panel, so I'm adding the WIP label until that is done.


Odrakir:
New id generated and set in bitrise (just in case, although it's probably public)

PR #2246, opened 2018-05-23
Fix login button glitch

### LINKED ISSUE #2137: Login button glitches sometimes
This happens, when `LoginCommand` executes too fast and the loading indicator remains visible.

Steps to reproduce:
1. Replace `NewLoginViewModel.login()` with
```
Observable
    .Throw<ITogglDataSource>(new Exception())
    .Subscribe(onDataSource, onError, onCompleted);
```
2. 💥 

screen shot 2018-05-15 at 11 02 29

simonrozsival:
Closes #2137 

The problem was that fade-out animation was called at the time when the fade-in animation hadn't had time to increase Alpha to 0.5 or more. This mean that it wouldn't be hidden and it would be displayed over the text. The condition wasn't necessary at all, removing it doesn't break anything.

I think that as a side-effect this removed a glitch from the reports (fade in/out of the pie chart and the header view).

Squash whenever


Odrakir:
I'm trying to test this but I can't reproduce the initial bug. Also I do see a second fade in when entering reports, but I see it regardless of this change. 🤔

simonrozsival:
@Odrakir in `LoginViewModel` use this:
```
        private void login()
        {
            IsLoading = true;
            ErrorMessage = "";

            loginDisposable =
                Observable.Throw<ITogglDataSource>(new Exception())
                    .Subscribe(onDataSource, onError, onCompleted);
        }
```

Then it should be easy to reproduce.

Odrakir:
Yep, I was missing the first two lines from Arturs issue. Still reports fade in twice, should you take a look or can we leave that for another issue?

simonrozsival:
I will look at the reports thing and if it is unrelated, I will create a new issue, otherwise I will fix it within this issue.

simonrozsival:
The issue with the reports is an unrelated issue.

PR #2248, opened 2018-05-23
New sign up

juanlaube:
Closes #2025 

I will fake the country selection and terms acceptance in another branch to test things, but it's mostly done.


Toggl.Giskard/Converters/BoolToConstantOverrides.cs
```diff
@@ -81,4 +81,10 @@ public sealed class MainFooterHeightValueConverter : BoolToConstantValueConverte
     {
         public MainFooterHeightValueConverter() : base (104, 70) { }
     }
+
+    public sealed class ButtonEnabledTextColorValueConverter : BoolToConstantValueConverter<Color>
```
kornelijepetak: I believe this name is too generic. Maybe some other buttons will have to be modified in other way. Maybe just add a prefix that is related to the usage, i.e. `SignupButtonTextColor` and then you don't even have to have button. Most of converters in this file are rather specific. 
But I don't know. Maybe I am ok with what you did... Think again and decide for yourself.

juanlaube: I'm not a fan of this either, I agree SignupButtonTextColor is better, but I was planning to use the same on the login activity after this is merged. Suggestions?

kornelijepetak: Oh, if this is kind of reusable thing, go away and merge it like this.

PR #2251, opened 2018-05-23
Fix reports being loaded twice on initialization

### LINKED ISSUE #2250: Reports are initially loaded twice
As we discussed in #2246, reports appear twice in a row when user navigates into this view. It looks weird.

simonrozsival:
Closes #2250 

The first initialization was in the `ReportsViewModel` and the second one in the `ReportsCalendarViewModel`. The first one isn't really necessary - the second one also preselects the current week in the calendar (the first one also selects the current week, but it doesn't change the calendar).

One small change which might not be obvious is necessary - if the `IsLoading` property wouldn't be set to `true`, then  "Nothing here" would be shown for a brief moment, which would be weird. The actual loading will start after the view is displayed (the calendar VM is initialized in the `ViewAppeared` method) - if you know how it would be possible to start initializing this earlier, please tell me!

Squash whenever


Toggl.Foundation.Tests/MvvmCross/ViewModels/ReportsViewModelTests.cs
```diff
@@ -33,6 +33,24 @@ protected override ReportsViewModel CreateViewModel()
                 DataSource.ReportsProvider.Returns(ReportsProvider);
                 return new ReportsViewModel(DataSource, TimeService, NavigationService);
             }
+
+            protected async Task Initialize()
+            {
+                var block = new AutoResetEvent(false);
```
kornelijepetak: From the .NET docs:

> This type implements the `IDisposable` interface. When you have finished using the type, you should dispose of it either directly or indirectly. To dispose of the type directly, call its `Dispose` method in a try/catch block. To dispose of it indirectly, use a language construct such as `using`.

amulware:
The code is a bit complicated but if you guys all agree that it's good, and it tests well, please merge it into `uranus`.

amulware:
If Kornelije approves and the tests pass, feel free to 🦑 into my branch.

PR #2257, opened 2018-05-24
Update No project color

### LINKED ISSUE #2249: The background color of `No Project` and `Other` is the same in the reports chart
- the two segments blend together when they are next to each other
- it is even more obvious when “Other” is < 10%
    - the title is not displayed (this is now the only segment which might not have the title displayed)
    - it seems like “No Project” is a bit bigger than it should be

Same thing probably could happen for other types of projects. We should somehow separate them.

arturs105:
Closes #2249

Updated the no project color in:
* reports
* project selection view

Squash whenever.


Toggl.Foundation/Autocomplete/Suggestions/ProjectSuggestion.cs
```diff
@@ -39,7 +39,7 @@ private ProjectSuggestion(long workspaceId, string workspaceName)
         {
             ProjectId = NoProjectId;
             ClientName = "";
-            ProjectColor = "#A3A3A3";
+            ProjectColor = "#B5BCC0";
```
simonrozsival: Use the helper here.

amulware:
Feel free to 🦑 into uranus if Simon approves and the builds pass.

PR #2267, opened 2018-05-24
Foundation.MvvmCross: IDatabase to IThreadSafe

simonrozsival:
Review commit by commit.

### DIFF
--- a/Toggl.Foundation/Models/FoundationModels.cs
+++ b/Toggl.Foundation/Models/FoundationModels.cs
@@ -16,7 +16,9 @@ namespace Toggl.Foundation.Models
 
         public string Name { get; }
 
-        public IDatabaseWorkspace Workspace { get; }
+        public IDatabaseWorkspace Workspace => ThreadSafeWorkspace;
+
+        public IThreadSafeWorkspace ThreadSafeWorkspace { get; }
 
         public bool IsDeleted { get; }
 
@@ -76,11 +78,17 @@ namespace Toggl.Foundation.Models
 
         public int? ActualHours { get; }
 
-        public IDatabaseClient Client { get; }
+        public IDatabaseClient Client => ThreadSafeClient;
 
-        public IDatabaseWorkspace Workspace { get; }
+        public IThreadSafeClient ThreadSafeClient { get; }
 
-        public IEnumerable<IDatabaseTask> Tasks { get; }
+        public IDatabaseWorkspace Workspace => ThreadSafeWorkspace;
+
+        public IThreadSafeWorkspace ThreadSafeWorkspace { get; }
+
+        public IEnumerable<IDatabaseTask> Tasks => ThreadSafeTasks;
+
+        public IEnumerable<IThreadSafeTask> ThreadSafeTasks { get; }
 
         public bool IsDeleted { get; }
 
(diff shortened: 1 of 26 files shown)


Toggl.Foundation/Models/FoundationModels.cs
```diff
@@ -76,11 +78,17 @@ internal partial class Project : IThreadSafeProject
 
         public int? ActualHours { get; }
 
-        public IDatabaseClient Client { get; }
+        public IDatabaseClient Client => ThreadSafeClient;
```
sofokli1: Why not just change the type from IDatabaseClient to IThreadSafeClient?

simonrozsival: Because the `IThreadSafeProject` interface extends the `IDatabaseProject` - this class must then implement the `IDatabaseProject` interface as well. It isn't very elegant, I know :/

amulware: Just checking: you can't use the 'new' keyword or something to 'override' the property?

simonrozsival: I am not sure how that would work 🤔
- I have to implement `IDatabaseClient Client { get; }` which is defined in the `IDatabaseProject` interface
- `IThreadSafeProject` extends `IDatabaseClient` - I can ofcourse add the `new IThreadSafeClient Client { get; }` property here
- The compiler still requires the implementation of `IDatabaseProject.Client` in the `Toggl.Foundation.Models.Project` class

IMO this is how it should behave (how could you then cast this class as `IDatabaseProject`?), but maybe I am using it wrong and it can be done in some other way so it is possible to hide it somehow and avoid these properties.

amulware: I'm not sure if this is possible here, but can you do something like this in the model?
```
class ThreadSafeProject : IThreadSafeProject
{
    IDatabaseClient IDatabaseProject.Client => Client; // for compatibility with the database interface
    public IThreadSafeProject { get; } // actual data here
}
```

simonrozsival:
@willsb please have another look at this PR - I changed the implementation based on internal discussion so that the properties of the thread-safe interfaces hide the properties of the database interfaces.

Odrakir:
I don't have a problem with this PR per se, but as it builds upon how things are done I'll just put my two cents: I think due to the heavy use of interface inheritance (not interfaces themselves, I'm fine with that), concerns cross layers when they shouldn't. For example `Project` in Foundation is an `IThreadSafeProject`, which is already leaking some implementation details I don't really care about, but what is worse, it's also an `IDatabaseProject`. Why would I want to know anything about the DB from Foundation? I should know nothing at all if possible. If I do need some DB info (to handle syncing) that should be mapped to business logic properties.

In my opinion, every layer should have its own models, and interfaces shouldn't be shared between layers. Not counting whatever you need to actually pass the models between the layers.

Normally you should start with the business logic layer defining there models and relationships as the business understands and uses them. This layer is the center of the app and only its models should cross layers to UI and DB (maybe with interfaces if needed for the assemblies thing). Then in the other layers, you map to and from those models.

This is my understanding on how a layered app should work, of course, I could be understanding everything wrong due to my lack of experience in C# / Xamarin, but allow me to share my thoughts.

Finally, I want to insist on the fact that code generators make the code much harder to understand (all those models in the same file).

simonrozsival:
@Odrakir I agree with what you say. Maybe it would be a good idea to decouple `IThreadSafe*` from the database interfaces. There will be some issues with some generic constraints, but I think that can be dealt with. I will try it and if it is doable without any crazy refactoring, I will create a PR.

simonrozsival:
Please review this PR again - @Odrakir, @sofokli1, @amulware, @willsb

PR #2269, opened 2018-05-24
Rxfactor: A Proposal for our bindings and more

willsb:
:warning: DISCLAIMER :warning: 
This is, by no means, final. This is meant to spark a discussion to decide what we’ll do next. Please comment and try finding problems with this implementation. Disregard bugs and the CI not working. This is not a PR that’s merge-ready, but rather a proposal.

The discussion about changing the way we do bindings might make sense. This is a proposal of how we could do things. The idea is that:

- We now depend on IObservable instead of INotifyPropertyChanged
- We use `I*View` interfaces to ensure that our VCs/Activities/Fragments/Views always have the appropriate bindings and that the needed methods are always overriden 
- On Droid we’ll have statically typed bindings, something much more future proof than the way we are doing things right now.
- These changes will cause the rest of the app to progressively become more reactive. This means we’ll end up with a overall more functional and predictable approach, contrasting the current imperative one we have atm.

Please find cons, discuss what you like and dislike and let’s get this ball rolling!

~NOTE: Most of that code from the extension *could* be auto-generated, but I’m not so sure about that.~


simonrozsival:
I must say I like this concept.

- it will make animated UI changes much easier
- those who haven't had the opportunity to work with value combiners will be spared of having this "opportunity"
- the view interface will make it harder to break the other platform (Android)
- I like some of the new extensions

I am not really sure how I feel about the view interface extension methods 🤔 it will probably be OK. But we shouldn't generate them - generating stuff is a pain to maintain/extend - just look at what became of our model generators. I am quite sure that matching view properties/methods to the view model properties/methods will be quite complex and we would have to have some naming conventions or some other ugly things to make it work and it will still break half of the time...

I like the idea of changing just the bindings (first?). I like the way Mvx works with DI and the life cycle methods make sense and we don't have to invent this ourselves - it's the bindings which are sometimes a pain (adding custom bindings, so many converters, value combiners...).

arturs105:
At first i wasn't sure about `CreateBindings` extensions, but i think i get the point of it now with the `ISettingsView` interface.

Though i think we could hide the `ObserveOn(SynchronizationContext.Current)` in some kind of extension method, like `TogglSubscribe`, which would combine `ObserveOn`, `Subscribe` and `DisposedBy` calls. Would lead to much less code repetition.

I'm in for this change - i like the idea of making droid less breakable by accident, but i agree with @simonrozsival about the generated stuff being a pain to maintain.

Odrakir:
I totally agree with Simon. I like this approach much better, less magical. I don't like the idea of using generators if they can be avoided, they just add another thing that can fail and that everyone has to learn.

Correct me if I'm wrong, but the whole idea of the I*View is to have that code in Foundation instead of each platform, right?

To me it seems a little strange to do it that way, having to have all those properties for those events (taps) and the I*View having references to both, the view and the VM... I don't know about that. Ideally, Rx allows you to have dependencies pointing only one way.

That said I'm not sure about the whole I*View extension thing. It might be that I'm not getting something related to Android XML... In that case, my comments might make no sense. Maybe if there was another example of all those bindings working for Android I could have a better opinion.

But initially, I like better the idea of the VM exposing the inputs/outputs needed for the View to work and not knowing anything about its existence. Also, I like how [Kickstarter](https://github.com/kickstarter/ios-oss), for example, separates inputs and outputs in two interfaces.

willsb:
🔥 I updated the description: We're not using generator on this one 😄 

@arturs105 I like this extension method idea 👍 

@Odrakir There's no droid knowledge required what-so-ever, so your comments do make sense.

I think we should have a simple interface that gives allows us to simply implement a couple of methods and be good to go. When working on a new platform, this allows us to predict all possible states of the VM that trigger UI. It's very foolproof. This allows the Observables to be internal in the ViewModel, making abuse much harder. I'm simply putting the extension method in foundation because it's merely boilerplate; there's no reason to write it twice.

I, however, agree that the I*View is not in a state I like well enough. I also do not like the fact that the tapping observables are exposed.  How do you suggest we do that? Because adding subscribe code twice is not something that makes sense, IMO. The current way makes the view very foolproof.

Odrakir:
I want to think about it a bit more, there seems to be some stuff that would easily be managed by a framework that links the UI with Rx. Is ReactiveUI supposed to do that? Are we going to use it?

daividssilverio:
I like this approach! It feels a lot more familiar compared to my previous Android projects.

If we had something like the `RxAndroid` for our view bindings, it would be really easy to attach our VM to our views.

These I*Views feels a lot like the idea of Contracts with Views and Presenters (only we are using VMs here). I like it, and we could add simple tests for them, just to check if the view is updated when the VM is changed (If we want to).  And like you said, it makes our views really foolproof.  

I like this CreateBindings extension, it's even less code to implement on both platforms <3. Nice stuff.

About Mvx (or C#), I don't know much about it, but I enjoyed the navigation service and how you could do stuff like: `someVar = await SomethingAsync` that can involve changing Activities, opening Dialogs etc, are we still gonna have that feature?

amulware:
There is a certain kind of elegance to it. But I'm really not happy with the I*View interface.

It very much prescribes the implementation of the view (by strongly implying what buttons should exist, and that there is a concept of 'tapping' them).

Further, I think
> we should have a simple interface that gives allows us to simply implement a couple of methods and be good to go

really does not apply. The interface has a lot of members, and it is now pretty annoying to implement a partial view and not support all the features (which has to be possible trivially). If we add a new feature on one platform, we now have to edit additional files and also implement it as a dummy on the other platform, just to get it to compile. Not ideal.

Unless we find a better solution here, I would prefer to have the code duplicated in each view. At least that is specific and clear. The current approach might be a bit too clever and has the downsides mentioned above.

Odrakir:
I see the code in `SettingsActivity.cs` and that's what I would like for the iOS UI too. Hopefully, it could get a bit more concise with a UI-Rx binding framework, but that's it. The View knows about the ViewModel (an interface to be precise), but the view model doesn't know anything about the view, not even an interface.

willsb:
@amulware I understand your concern, but it's not enough to justify code repetition. 

Most (almost all) of the existing ViewModels have a View counterpart in both platforms with not differences between then. On the views where such different exists, it's either temporary (for means of testing a feature's popularity) or something that will never get implemented on the other platform. This is easy to solve: 

- if we plan on implementing a feature in an existing I*View interface and it won't be implemented on the other platform anytime soon, we simply use a different interface (which can be either merged or deleted, depending on how successful the feature is)
- If we don't plan on implementing it altogether on the other platform, no interface is needed and the bindings go straight into the platform specific code.

> At least that is specific and clear
It might be clear, but not specific. The interface tells the view which methods should be overriden exactly for the UI to perform as expected, and that's very foolproof, even for reviewers.

willsb:
@daividssilverio Yes, the navigation service will remain untouched. We are only dealing with 2WDB.

juanlaube:
I really like the approach! 😃 

- I prefer to multiple libraries doing small things (binding, navigation, dependency injection, etc) instead of a big framework doing everything. This is only a small part, but it's in the right way.
- I like the explicit bindings in `.cs` files instead of stringly typed APIs in `.axml`. It's safer to break things and get an error in compile time.

About the `I*View` interfaces, the only problem I could see with this and the `CreateBindings` is platform specific UIs (e.g: think the wheel in iOS vs the clock thingy in Android), but that could be solved by having a common `defaultBindings` plus specific platform bindings.

kornelijepetak:
I have mixed feelings about this.

Similar to @amulware 's observation, I am somewhat afraid of the differences between platforms (design-wise)... It can be done with proper interface segregation, so it's not a dealbreaker, I just believe it could get a bit ugly... But this problem exists even now so... 🤷‍♂️ 

The other thing I did notice is the sheer amount of boilerplate code, which if not generated, will be tedious. and is also needed in multiple files. Regarding android - while using string is never the best option, switching 

`Text Convert(Label)`

into

```
ViewModel
    .Label
    .Subscribe(label => textField.Text = label)
    .DisposedBy(disposeBag);
```
is kind of, meh... It's still kind of declarative, but much more verbose. Which I do not like. 

However, since it's much more testable and compile-time-proof, I would definitely consider it a good idea. I am just thinking if something can be reduced...

There's also another problem I see... There will sometimes be a need for multiple `CreateBindings` methods for the same VM... I've had a scenario with viewpager in which the pages are not inflated immediately, but with a delay (next UI pass, with `Post`)... and `FindViewById` needs to be called at different places, so only parts of UI should be bound at start and another part at some later time. This is probably rather rare, but maybe something to consider.

All in all, I welcome the proposal, I just have a feeling some parts of it could be a bit more thought out, but I feel bad saying that as I can't think of a better way at this moment. 🤷‍♂️

Odrakir:
@kornelijepetak I kind of agree with you that the whole subscribe, dispose thing is verbose, but what you lose in conciseness you win in flexibility.

And then there's also the possibility of creating UI extensions so you would have, for example, a `RxText` property on Label, which would be an Observer so you could do:

```c#
ViewModel
   .Label
   .Subscribe(Label.RxTest)
   .DisposeBy(disposeBag);
```

Which could be converted into a method call like `bind(observable, observer)`

amulware:
## On interfaces

I think one of the common things we (you) do is adding new features by creating a couple more bindings. However, especially since we want to split up issues and work into doing foundation first and only then do the UI more in the future, do we have to:

1. add a new interface every time? And then merge them later and forget half of the time?
2. Or will we always have to add dummy observers/observables in the views, on both platforms?

This is really what bothers me, and I don't see a good way around that.

## On boilerplate

This I am in principle fine with. However, it _IS_ another hurdle for new devs to learning how things work. But at least they can find the extensions or whatever we use easily in our own code, and they'll be short and make sense. _Maybe_ that's ok. As long as things are really really well named.

## Lastly

I worry again how this is going to affect our work day to day. Will it really be easier to work like this? You guys know better than me. Please think about your recent UI work, and if this approach would have helped, or if it would have made it more work.

willsb:
Closing this in favour of a proper implementation in another PR, with the concrete ideas from everyone.

PR #2270, opened 2018-05-24
Fix feedback navigation bar glitch

sofokli1:
Doesn't close any issue

Makes the feedback ui show navigation bar by removing the UIAppearance setBacgroundImage before showing ui and sets it back when dismissing the ui.


sofokli1:
@arturs105 did you try to login with that setup?

arturs105:
No, i had not. But now i tried it out and there was a small UI glitch when switching between login and signup, but that can be solved by adding `NavigationController.NavigationBarHidden = true` to both view controller.

sofokli1:
I get a navigation bar in the blue onboarding screens. Don't you get that?

arturs105:
Oh, yeah. That can be solved with the same `NavigationController.NavigationBarHidden = true`

sofokli1:
So i am setting the navigation appearance colors to the same color as main view background color but the navigation looks a bit different like in this image 
screen shot 2018-05-25 at 13 19 38

So i made a extension that will make a small 1x1 image with the desired color and use that as background image for the bar. now I get the wright color.

Toggl.Daneel/Services/MailService.cs
```diff
@@ -3,6 +3,8 @@
 using Toggl.Foundation;
 using Toggl.Foundation.Services;
 using Toggl.Multivac;
+using UIKit;
+using Toggl.Foundation.MvvmCross.Helper;
```
arturs105: I don't thing these are needed.

Toggl.Daneel/Startup/AppDelegate.cs
```diff
@@ -131,10 +132,10 @@ private void setupNavigationBar()
 
             //Title and background
             UINavigationBar.Appearance.ShadowImage = new UIImage();
-            UINavigationBar.Appearance.BarTintColor = UIColor.Clear;
-            UINavigationBar.Appearance.BackgroundColor = UIColor.Clear;
+            UINavigationBar.Appearance.BarTintColor = Color.NavigationBar.BackgroundColor.ToNativeColor();
+            UINavigationBar.Appearance.BackgroundColor = Color.NavigationBar.BackgroundColor.ToNativeColor();
             UINavigationBar.Appearance.TintColor = Color.NavigationBar.BackButton.ToNativeColor();
-            UINavigationBar.Appearance.SetBackgroundImage(new UIImage(), UIBarMetrics.Default);
+            UINavigationBar.Appearance.SetBackgroundImage(ImageExtension.ImageWithColor(Color.NavigationBar.BackgroundColor.ToNativeColor()), UIBarMetrics.Default);
```
arturs105: I think you should export `Color.NavigationBar.BackgroundColor.ToNativeColor()` as a varibale.

sofokli1:
@arturs105 So the issue was that by using a solid color image the bar lost its translucency thus ending in a strange behaviour in iOS 10 only. (also with `develop` implementation and setting the navigation bar `translucent = false` gives the same result). It needed `ExtendedLayoutIncludesOpaqueBars = true` on all VCs in the navigation stack. Take a look at the fix and let me know.

Issue #2281, opened 2018-05-25
Make sync status bar less intrusive


simonrozsival:
I think we should change the way we display the sync status bar - it is intrusive. We push every change right when the user does the change - when you delete something or when you start/continue/stop a time entry, which is good, but then the whole interface starts jumping up and down (or rather down and up) and I really don’t like it - not to mention that we broke it somehow and the animation looks weird sometimes (I’m not sure what is wrong with it..).

Maybe the solution would be to show the syncing bar only when user forces the sync with the pull-to-sync gesture and when there is a problem (offline or failure) - that wouldn’t even be hard to implement at all.

amulware:
I agree with the proposed change. I would also like it show the bar if syncing takes longer than 2-3 seconds however, to make the user aware that something is going on.

Alternatively (or additionally?) I would like the syncing icon arrows ♻️on the unsynced time entries to spin around the exclamation point _if and only if_ syncing is in progress. I think that would be a nice indicator, especially if there are multiple, and the user can see the icons spinning and then going away one by one.

pe0ny:
I agree that the jumping is really weird and I would propose another solution - new design that doesn't make the whole interface move up-down.

During the customer meetings, I heard more than once that they like to see when the syncing happens because this helps them to make sure their data is synced. Makes sense imo.

amulware:
Ths solution is probably simple: just have the bar that pops up automatically be over other stuff, not displacing it. (I still want my spinning icons though :D)

amulware:
alternatively, for something more subtle, consider the Toggl logo in the top animating where the ring rotates, and the little tick ' turns into an arrow head > and when syncing is done it turns back into the Toggl logo. Just an idea.

amulware:
This should be a full project.

PR #2282, opened 2018-05-25
Pull to refresh refactoring - do not show it all the time

simonrozsival:
Ref #2281 

The syncing bar is now not show when we sync stuff without user's explicit interaction (after opening the app after > 5 minutes, when user starts/stops/continues a TE, ...) and everything goes well - the bar will be always shown when user is offline or if the sync fails.

This PR also refactors the delegate a bit...

This PR isn't meant to be merged into develop just yet.


simonrozsival:
This is not important at the moment, I will close it but keep the branch for now.

Issue #1332, opened 2018-01-25
Always have two months of time entries fetched


amulware:
We want to fetch these as soon as the user logs in. We still have to figure out what endpoints to use. Ideally, we can fetch the TEs by start date, not by AT date.

This issue might be relevant: https://github.com/toggl/backend/issues/6110

Note that if the user has not synced for too long so that we would reset the time entry since date, we should again fetch the full two months from scratch.

amulware:
The above issue is indeed relevant. The endpoint created there is exactly what we'll use. Note that we should still always do the since call/regular call without since that data will be cached and faster.

We can display that data and persist it right away and then call the new endpoint to get the last two months of data. **This only has to happen if we called the regular endpoint without since date.**

amulware:
This'll require the new endpoint to be implemented, ideally in its own PR with good integration tests.

simonrozsival:
I am quite confused - we have three cases:
- get "all" (last 10 days of activity - compared by `at`, not by start) - "regular call without since"
- get all since some date
- get last two months

Which case should be used when? Could you summarize it somehow?

Right now the logic is:
```
if (since.HasValue && since.Value is less than two months old):
    get all since
else:
    get all without since (last 10 days)
```

simonrozsival:
ping @amulware just in case you missed my question, ignore if it is already in your queue

amulware:
_in principle_ we should do this:
```
if (since.HasValue && since.Value is less than two months old):
    get all since
else:
    get all in last 8 weeks
```

However, the 10-day endpoint is potentially a lot faster, since that data should be cached or something, so we should do:

```
if (since.HasValue && since.Value is less than two months old):
    get all since
else:
    get all without since (last 10 days)
    get all in last 8 weeks in parallel/after
```

Now the question is: Does this really get us any speedup anywhere? Maybe we should just remove the 10-day thing?

Certainly on login right now, we wait before showing UI until everything is pulled.

Ideally, we would maybe show UI after pulling 10 days, and then load 8 weeks in the background while the user can already show the app. Though I'm not sure how tricky that is. Nor do I know how much time it will save, since we're not measuring right now (can you try just curling locally to see if there is a significant difference in time? especially on the first request (repeated ones might be cached)).

In the case of already being logged in and the user coming back after two+ months, do we care much how long the initial sync takes? Ideally, we show data as fast as possible of course...

It comes down to the time difference between the endpoints. To we have any idea? Can we test it?

[...]

PR #2284, opened 2018-05-25
Fetch last two months time entries

simonrozsival:
Ref #1332 

Integration tests: https://www.bitrise.io/build/d75c893bbc97e9e8 - ✅ 

Squash if the integration tests pass


Toggl.Ultrawave.Tests.Integration/TimeEntriesApiTests.cs
```diff
@@ -74,6 +74,70 @@ protected override IObservable<ITimeEntry> PostModelToApi(ITogglApi api, ITimeEn
...
+                var timeEntry = createTimeEntry(user, start.AddDays(-1));
+                await api.TimeEntries.Create(timeEntry);
+
+                var list = await api.TimeEntries.GetAll(start, end);
+
+                list.Should().BeEmpty();
+            }
+
+            [Fact]
+            public async Task DoesNotReturnTimeEntriesWhichEndAfterTheEndParameter()
```
amulware: The endpoint should be filtering by start time only. End time really isn't a thing in Toggl business logic, it is merely implied by UI for UX reasons.

simonrozsival: The endpoint requires the `end_date` to be specified, so I think it should be explicitly defined in the method and that it should be also tested. Do you think that we should assume that the `end_date` will always be `today`?

amulware: No, I mean that the api does not look at end times of time entries. This endpoint should return TEs with a start time between the two given parameters. But yes, there should also be a test that checks that the default is 'today'.

simonrozsival: Now I get it. But I think that the end date should be specified by the syncing state, not by the API client.

amulware: Yes, except it should be possible not to specify an end date at all, and let backend figure it out.

amulware: It seems to me like that would simplify things from our side.

simonrozsival: I am not sure if I understand you correctly again - backend requires the end date, it will return 400 if the end date is omitted (error msg: `start_date and end_date are both required`)

amulware: Then I misunderstood the endpoint. All good.

Toggl.Ultrawave.Tests.Integration/TimeEntriesApiTests.cs
```diff
@@ -74,6 +74,58 @@ protected override IObservable<ITimeEntry> PostModelToApi(ITogglApi api, ITimeEn
...
+            private static readonly DateTimeOffset end = start.AddDays(60);
+
+            protected override IObservable<List<ITimeEntry>> CallEndpointWith(ITogglApi togglApi)
+                => togglApi.TimeEntries.GetAll(start, end);
+
+            [Fact]
+            public async Task DoesNotReturnTimeEntriesWhichStartBeforeTheStartParameter()
+            {
+                var (api, user) = await SetupTestUser();
+                 var timeEntry = createTimeEntry(user, start.AddDays(-1));
```
sofokli1: fix indentation

Toggl.Ultrawave.Tests.Integration/TimeEntriesApiTests.cs
```diff
@@ -74,6 +74,58 @@ protected override IObservable<ITimeEntry> PostModelToApi(ITogglApi api, ITimeEn
...
+                await api.TimeEntries.Create(timeEntryA);
+                await api.TimeEntries.Create(timeEntryB);
+                await api.TimeEntries.Create(timeEntryC);
+
+                var list = await api.TimeEntries.GetAll(start, end);
+
+                list.Should().HaveCount(1);
+                var item = list[0];
+                item.Start.ToUnixTimeSeconds().Should().BeGreaterOrEqualTo(start.ToUnixTimeSeconds());
+                item.Start.ToUnixTimeSeconds().Should().BeLessThan(end.ToUnixTimeSeconds());
```
sofokli1: No request here just a conversation:
i guess we do not have a endTime for time entries, right? just start and duration?
if we have and time shouldn't the end time also be in the range?
if we don't have end time shouldn't entries that have `start` is outside the range and `start + duration` inside the range be also part of the results?

sofokli1: if we just do not get it from the server that is fine, not much we can do about it

amulware: You are correct, end times aren't really a thing business logic wise. TEs are generally categorized by their start time (for example, if I have a 10 hours entry starting at 23:00 on Monday, Monday's report has 10 hours, Tuesday's has 0). Everything else follows the same logic, and while sometimes people complain about it, it is at least consistent and relatively easy to handle. So it's very much on purpose to keep things simple.

PR #2287, opened 2018-05-25
Delete older time entries state

### LINKED ISSUE #1333: Prune time entries older than two months
By start date. We have to decide if we can do this on every app wake, or only on app wake after a certain amount of time, or only on app start. Maybe doing it on every app wake isn't too bad, if the query is either very fast, or we just do it in the background anyways.

I think this should always be rounded to midnight (i.e. setting hour/minute/second to 0) as the cutoff date, to not have time entries disappear one by one during the day. As such, maybe we can just prune on app activation once per day?

One thing to take into account is what will happen if the app is scrolled all the way to the end of the time entry list as the entries get removed. We have to make sure the user can't continue a time entry just in the moment it gets deleted, causing a crash, or something similar.

willsb:
Closes #1333 

~Still WIP, but I’d like @amulware and @simonrozsival to take a look at it and confirm that I’m doing it right®~

Squash whenever


Toggl.Foundation/DataSources/Interfaces/IBaseDataSource.cs
```diff
@@ -17,5 +18,7 @@ public interface IBaseDataSource<TThreadsafe, TDatabase>
         IObservable<IConflictResolutionResult<TThreadsafe>> OverwriteIfOriginalDidNotChange(TThreadsafe original, TThreadsafe entity);
 
         IObservable<IEnumerable<IConflictResolutionResult<TThreadsafe>>> BatchUpdate(IEnumerable<TThreadsafe> entities);
+
+        IObservable<IEnumerable<IConflictResolutionResult<TThreadsafe>>> DeleteAll(IEnumerable<TThreadsafe> entities);
```
simonrozsival: I think this shouldn't be in `IBaseDataSource`, but in `IDataSource` - now it is inherited also by `ISingletonDataSource` which probably shouldn't have this method.

Toggl.Foundation/DataSources/BaseDataSource.cs
```diff
@@ -47,6 +47,10 @@ public virtual IObservable<IEnumerable<IConflictResolutionResult<T>>> BatchUpdat
                     RivalsResolver)
                 .ToThreadSafeResult(Convert);
 
+        public virtual IObservable<IEnumerable<IConflictResolutionResult<T>>> DeleteAll(IEnumerable<T> entities)
+            => Repository.BatchUpdate(convertEntitiesForBatchUpdate(entities), alwaysDelete)
```
amulware: Will this work if a TE is deleted between fetching the list, and calling this? I.e. will this correctly ignore entities given that do not exist in the repository?

amulware: In either case, this needs tests.

Toggl.Foundation.Tests/Sync/States/Pull/DeleteOldEntriesStateTests.cs
```diff
@@ -0,0 +1,86 @@
...
+                .DeleteAll(Arg.Is<IEnumerable<IThreadSafeTimeEntry>>(tes => tes.Count() == 5));
+        }
+
+        private void configureDataSourceReturn(IEnumerable<IDatabaseTimeEntry> timeEntries)
+        {
+            dataSource
+                .GetAll(Arg.Any<Func<IDatabaseTimeEntry, bool>>())
+                .Returns(callInfo =>
+                {
+                    var filteredClients = timeEntries.Where(callInfo.Arg<Func<IDatabaseTimeEntry, bool>>());
```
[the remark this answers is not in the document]
willsb: 👀

Toggl.Foundation/DataSources/BaseDataSource.cs
willsb: What do you specifically want to see tested, @amulware ?

willsb: My thoughts exactly...
What do?

simonrozsival: Put the resolver in a separate class, test this class. We can't test `RealmAdapter`...

willsb: In another class? But this is just a Func, would you create an entire class just for that?

amulware: This class doesn't refer to anything realm though? Can't we just inject a mocked repository and make sure that the passed resolved behaves correctly?

Toggl.Foundation.Tests/DataSources/BaseDataSourceTests.cs
```diff
@@ -24,4 +34,61 @@ private void Setup()
...
+            dataSource = new TimeEntriesDataSource(repository, timeService);
+        }
+
+
+        [Fact]
+        public async Task TheDeleteAllMethodIgnoresTheConflictIfTheOldEntityIsNull()
+        {
+            var entities = Enumerable.Range(0, 10).Select(i => new MockTimeEntry { Id = i });
+
+            repository.BatchUpdate(
```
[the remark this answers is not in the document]
kornelijepetak: Am I too old already?

Toggl.Foundation.Tests/Sync/States/Pull/DeleteOldEntriesStateTests.cs
```diff
@@ -0,0 +1,87 @@
...
+        {
+            var timeEntries = getTimeEntries();
+            configureDataSourceReturn(timeEntries);
+
+            await state.Start().SingleAsync();
+
+            await dataSource.Received()
+                .DeleteAll(Arg.Is<IEnumerable<IThreadSafeTimeEntry>>(tes => tes.Count() == 10));
+        }
+
```
kornelijepetak: :fire:`\n`

Toggl.Foundation.Tests/DataSources/BaseDataSourceTests.cs
```diff
@@ -24,4 +34,61 @@ private void Setup()
...
+        private readonly IRepository<IDatabaseTimeEntry> repository
+            = Substitute.For<IRepository<IDatabaseTimeEntry>>();
+
+        private readonly DataSource<IThreadSafeTimeEntry, IDatabaseTimeEntry> dataSource;
+
+        public DataSourceTests()
+        {
+            dataSource = new TimeEntriesDataSource(repository, timeService);
+        }
+
```
kornelijepetak: 🔥 `\n`

Toggl.Foundation.Tests/DataSources/BaseDataSourceTests.cs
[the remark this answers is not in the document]
amulware: local functions are a thing, in case Will wants to use those

amulware:
I don't have time for an in-detail review, but since the other guys approved and you addressed my issue, please go ahead.

PR #2293, opened 2018-05-25
Save and restore state in start TE activity

### LINKED ISSUE #2172: App discards any started time entry details that haven't been confirmed
If the app is minimized after starting a new time entry, after a description and project are assigned but before the entry has been saved then after a certain amount of time when the app is brought back up again the started entry is discarded and the app appears in the same state before that entry was started.

juanlaube:
Closes #2172 

I'm not convinced about this for multiple reasons:

- I wasn't able to properly test a tombstone scenario, even with a low ram emulator.
- I tested this by enabling `Don't keep activities` in the developer options and I had to add things in the Activity to make things work.


Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
```diff
@@ -286,6 +294,43 @@ public override void Prepare(StartTimeEntryParameters parameter)
...
+                TextFieldInfo = textFieldInfo.WithWorkspace(workspaceId);
+            }
+            catch (JsonSerializationException) { }
+
+            DateTimeOffset startTime;
+            DateTimeOffset.TryParse(state.Data[StateBundleParams.StartTime.ToString()], out startTime);
+            StartTime = startTime;
+
+            bool isBillable;
+            bool.TryParse(state.Data[StateBundleParams.IsBillable.ToString()], out isBillable);
```
kornelijepetak: `out var isBillable`

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
```diff
@@ -286,6 +294,43 @@ public override void Prepare(StartTimeEntryParameters parameter)
...
+                long workspaceId;
+                long.TryParse(state.Data[StateBundleParams.WorkspaceId.ToString()], out workspaceId);
+
+                var textFieldInfo = JsonConvert.DeserializeObject<TextFieldInfo>(state.Data[StateBundleParams.TextFieldInfo.ToString()]);
+                TextFieldInfo = textFieldInfo.WithWorkspace(workspaceId);
+            }
+            catch (JsonSerializationException) { }
+
+            DateTimeOffset startTime;
+            DateTimeOffset.TryParse(state.Data[StateBundleParams.StartTime.ToString()], out startTime);
```
kornelijepetak: `out var startTime`

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
```diff
@@ -286,6 +294,43 @@ public override void Prepare(StartTimeEntryParameters parameter)
             hasAnyProjects = (await dataSource.Projects.GetAll()).Any();
         }
 
+        protected override void SaveStateToBundle(IMvxBundle bundle)
+        {
+            try {
+                var textFieldInfo = JsonConvert.SerializeObject(TextFieldInfo);
+                bundle.Data[StateBundleParams.TextFieldInfo.ToString()] = textFieldInfo;
+            } 
+            catch (JsonSerializationException) { }
```
kornelijepetak: If this fails, the exception is swallowed. That means the state you save to the bundle will not be complete and that reload from bundle won't be complete either, leading to inconsistent state of the app.

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
```diff
@@ -286,6 +294,43 @@ public override void Prepare(StartTimeEntryParameters parameter)
...
+            base.SaveStateToBundle(bundle);
+        }
+
+        protected override void ReloadFromBundle(IMvxBundle state)
+        {
+            base.ReloadFromBundle(state);
+
+            try {
+                long workspaceId;
+                long.TryParse(state.Data[StateBundleParams.WorkspaceId.ToString()], out workspaceId);
```
kornelijepetak: since C# 6, you can do `TryParse(..., out var workspaceId)`. No need to declare a variable before.

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
```diff
@@ -286,6 +294,43 @@ public override void Prepare(StartTimeEntryParameters parameter)
...
+
+                var textFieldInfo = JsonConvert.DeserializeObject<TextFieldInfo>(state.Data[StateBundleParams.TextFieldInfo.ToString()]);
+                TextFieldInfo = textFieldInfo.WithWorkspace(workspaceId);
+            }
+            catch (JsonSerializationException) { }
```
kornelijepetak: The same here. I am not sure whether we should allow this to just go unnoticed. 
When I was a kid, I was afraid of heights and darkness. Now, I am afraid of inconsistent states.

juanlaube:
And I'm also requesting @willsb review.

Toggl.Foundation.MvvmCross/Extensions/BundleExtensions.cs
```diff
@@ -0,0 +1,118 @@
...
+                                bundle.Data.TryGetValue($"{nameof(TextFieldInfo)}.{nameof(TextFieldInfo.TaskName)}", out var taskName))
+                            {
+                                textFieldInfo = textFieldInfo.WithProjectAndTaskInfo(workspaceId, projectId, projectName, projectColor, taskId, taskName);
+                            }
+                            else
+                            {
+                                textFieldInfo = textFieldInfo.WithProjectInfo(workspaceId, projectId, projectName, projectColor);
+                            }
+                        }
+                    } else
```
kornelijepetak: Else on new line

Toggl.Foundation/Autocomplete/TextFieldInfo.cs
```diff
@@ -99,6 +99,16 @@ public TextFieldInfo AddTag(TagSuggestion tagSuggestion)
                 Text, CursorPosition, WorkspaceId, ProjectId, ProjectName, ProjectColor, TaskId, TaskName, tags);
         }
 
+        public TextFieldInfo AddTags(TagSuggestion[] tagSuggestions)
+        {
+            var result = this;
+            foreach (var tagSuggestion in tagSuggestions)
```
kornelijepetak: Since `.AddTag` creates a whole new `TextFieldInfo` object, I find this very unperformant. Similar to concatenating a lot of small strings with `+=`, there will be a lot array recreations and member copying...

I suggest creating another `TextFieldInfo` method: `AddTags` that would check for the existance of multiple tags and add them all in one batch and not recreate a `TFI` object in every step.

I like immutability, but this is really necessary IMO.

Toggl.Foundation.MvvmCross/Extensions/BundleExtensions.cs
```diff
@@ -0,0 +1,118 @@
...
+                    }
+                }
+
+                if (hasTags)
+                {
+                    if (bundle.Data.TryGetValue($"{nameof(TextFieldInfo)}.{nameof(TextFieldInfo.Tags)}", out var tagsIdsString))
+                    {
+                        var tagIds = tagsIdsString.Split(';').Select(str => long.Parse(str)).ToArray();
+                        viewModel.TagIdsToReload = tagIds;
+                    } else {
```
kornelijepetak: do an early return instead.

Toggl.Foundation.MvvmCross/Extensions/BundleExtensions.cs
```diff
@@ -0,0 +1,118 @@
...
+                            {
+                                textFieldInfo = textFieldInfo.WithProjectAndTaskInfo(workspaceId, projectId, projectName, projectColor, taskId, taskName);
+                            }
+                            else
+                            {
+                                textFieldInfo = textFieldInfo.WithProjectInfo(workspaceId, projectId, projectName, projectColor);
+                            }
+                        }
+                    } else
+                    {
```
kornelijepetak: early return here to avoid that much branch nesting...

juanlaube:
@simonrozsival I've removed the JSON serialization code I added.
I was going that way first but @willsb didn't want to depend on json.net for this because it was supposed to be easy to serialize manually.

Narrator: it wasn't.

Toggl.Foundation/Serialization/IJsonSerializer.cs
```diff
@@ -2,7 +2,7 @@
 
 namespace Toggl.Foundation.Serialization
 {
-    internal interface IJsonSerializer
+    public interface IJsonSerializer
```
simonrozsival: This could be reverted as well, right?

Toggl.Foundation/Serialization/JsonSerializer.cs
```diff
@@ -2,7 +2,7 @@
 
 namespace Toggl.Foundation.Serialization
 {
-    internal sealed class JsonSerializer : IJsonSerializer
+    public sealed class JsonSerializer : IJsonSerializer
```
simonrozsival: This could be reverted as well, right?

juanlaube: Right

Toggl.Foundation/Autocomplete/TextFieldInfo.cs
```diff
@@ -101,12 +101,12 @@ public TextFieldInfo AddTag(TagSuggestion tagSuggestion)
 
         public TextFieldInfo AddTags(TagSuggestion[] tagSuggestions)
         {
-            var result = this;
-            foreach (var tagSuggestion in tagSuggestions)
-            {
-                result = result.AddTag(tagSuggestion);
-            }
-            return result;
+            var currentTagIds = Tags.Select(tag => tag.TagId);
```
kornelijepetak: I would do 

`var currentTagIds = new HashSet<long>(Tags.Select(tag => tag.TagId));`

here, to change the following `.Where` from `O(M x N)` into a `O(N)`

Toggl.Foundation.MvvmCross/Extensions/BundleExtensions.cs
```diff
@@ -89,10 +89,15 @@ public static void ReloadPropertiesInto(this IMvxBundle bundle, StartTimeEntryVi
                             }
                             else
                             {
-                                textFieldInfo = textFieldInfo.WithProjectInfo(workspaceId, projectId, projectName, projectColor);
+                                return;
```
kornelijepetak: What I meant with early return is to revert the condition and return immediately so there wouldn't be an `else` at all.

Toggl.Foundation.MvvmCross/Extensions/BundleExtensions.cs
```diff
@@ -89,10 +89,15 @@ public static void ReloadPropertiesInto(this IMvxBundle bundle, StartTimeEntryVi
...
                             }
                         }
-                    } else
+                        else
+                        {
+                            textFieldInfo = textFieldInfo.WithProjectInfo(workspaceId, projectId, projectName, projectColor);
+                        }
+                    }
+                    else
```
kornelijepetak: Same here.

Toggl.Foundation.MvvmCross/Extensions/BundleExtensions.cs
```diff
@@ -104,7 +109,9 @@ public static void ReloadPropertiesInto(this IMvxBundle bundle, StartTimeEntryVi
                     {
                         var tagIds = tagsIdsString.Split(';').Select(str => long.Parse(str)).ToArray();
                         viewModel.TagIdsToReload = tagIds;
-                    } else {
+                    }
+                    else
```
kornelijepetak: And here.

Toggl.Foundation.MvvmCross/Extensions/BundleExtensions.cs
```diff
@@ -0,0 +1,117 @@
...
+                }
+
+                if (hasTags)
+                {
+                    if (!(bundle.Data.TryGetValue($"{nameof(TextFieldInfo)}.{nameof(TextFieldInfo.Tags)}", out var tagsIdsString)))
+                    {
+                        return;
+                    }
+                    var tagIds = tagsIdsString.Split(';').Select(str => long.Parse(str)).ToArray();
+                    viewModel.TagIdsToReload = tagIds;
```
daividssilverio: I'm a little uneasy about this extension. Isn't modifying the VM from an extension of the MvxBundle a little bit of an inappropriate intimacy? It feels like the VM (or one extension for the VM) should be responsible for these methods.

Especially the ReloadPropertiesInto part, which is actually modifying stuff on the VM.

Toggl.Foundation.MvvmCross/Extensions/BundleExtensions.cs
```diff
@@ -0,0 +1,117 @@
...
+                    }
+                    else
+                    {
+                        textFieldInfo = textFieldInfo.WithProjectInfo(workspaceId, projectId, projectName, projectColor);
+                    }
+                }
+
+                if (hasTags)
+                {
+                    if (!(bundle.Data.TryGetValue($"{nameof(TextFieldInfo)}.{nameof(TextFieldInfo.Tags)}", out var tagsIdsString)))
```
daividssilverio: there is one extra pair of unnecessary parenthesis here. after the `!`.

Toggl.Foundation.MvvmCross/Extensions/BundleExtensions.cs
simonrozsival: If this method returned some DTO which would then be used by the VM itself to update its state it would be nicer - agreed. The properties which were exposed could remain hidden.

juanlaube:
@kornelijepetak I didn't have the time to review #2410 yet, but I plan to do it today :D

Toggl.Foundation.MvvmCross/Extensions/BundleExtensions.cs
juanlaube: I like this idea, the extension doesn't need to know the internals of the view model

juanlaube:
Ping @willsb and @daividssilverio 😄

daividssilverio:
Other than comments, maybe we could something later about where those extensions are placed. Maybe a partial class would be nice? It feels like this something really specific for the `StartTimeEntryViewModel`.
This is a place where some "magic" would be cool. Haha.

(I'm not requesting changes for this now, just leaving my opinion).

PR #2294, opened 2018-05-25
Fix time entry card visibility bug

### LINKED ISSUE #2097: Bottom card freezes after deleting a time entry from the edit view
Steps to reproduce:
1) Start a new entry and tap "Save"
2) Open the edit view of the running entry
3) Tap "Delete this entry" and confirm

The card at the bottom of the screen remains there but is not responsive anymore. Other parts of the app work fine though. Restart helps.

### LINKED ISSUE #2232: UI doesn't updated when new TE is started while wifi/data is turned off
When wifi or mobile data is turned off, I have managed twice to get the app into the state when after starting a new TE and tapping "Save", UI doesn't update.
1) Start a new TE & tap "Save"
2) App directs you to the Timer page but the bottom card doesn't appear and Start button is visible. I can't start a new entry because actually TE is running but I can't see it and Start button is disabled (nothing happens when I tap it).

Restarting the app doesn't help. Only when I connect to the wifi, it syncs and I can see the running entry I started before.

It doesn't happen every time though.

willsb:
Closes #2097 
Closes #2232 

The bug was simple: When the update to the current TE happened too fast (because sync failed too quickly or because it was a simple local change), the hint would be dispatched before the MainActivity was the main activity. 

Replacing with an observable makes things work as expected, but it also makes it more readable and overall simpler to understand. 

A step in the direction of #2269, if I must say.


Toggl.Daneel/ViewControllers/MainViewController.cs
```diff
@@ -106,6 +109,9 @@ public override void ViewDidLoad()
 
             source.Initialize();
 
+            cardDisposable = ViewModel.TimeEntryCardVisibility
```
kornelijepetak: Why is this not disposed anywhere? I like `IDisposable`s to be disposed of.

Toggl.Giskard/Activities/MainActivity.cs
```diff
@@ -53,8 +55,12 @@ protected override void OnCreate(Bundle bundle)
...
             coordinatorLayout = FindViewById<CoordinatorLayout>(Resource.Id.MainCoordinatorLayout);
 
-            disposable =
-                ViewModel.WeakSubscribe<PropertyChangedEventArgs>(nameof(ViewModel.SyncingProgress), onSyncChanged);
+            var disposeBag = new CompositeDisposable();
+
+            disposeBag.Add(ViewModel.TimeEntryCardVisibility.Subscribe(onTimeEntryCardVisibilityChanged));
+            disposeBag.Add(ViewModel.WeakSubscribe<PropertyChangedEventArgs>(nameof(ViewModel.SyncingProgress), onSyncChanged));
+
+            disposable = disposeBag;
```
kornelijepetak: This `disposable` is not disposed as well... is it?

PR #2295, opened 2018-05-26
Add TextInputLayout with support to helper text

### LINKED ISSUE #2278: Forgot your password UI: Activation indicator should be blue
It should be blue like in the log in and sign up views but atm it's grey when the input is focused.

Zeplin: https://zpl.io/an1NzRr

daividssilverio:
It also closes #2278 

This view helps to support helper texts with TextInputLayout, as described here: https://material.io/design/components/text-fields.html#anatomy


Toggl.Giskard/Views/TextInputLayoutWithHelperText.cs
```diff
@@ -0,0 +1,231 @@
...
+
+        public TextInputLayoutWithHelperText(Context context, IAttributeSet attrs, int defStyleAttr) : base(context, attrs, defStyleAttr)
+        {
+        }
+
+
+        public override void AddView(View child, int index, ViewGroup.LayoutParams @params)
+        {
+            base.AddView(child, index, @params);
+            if (child != null && child.GetType() == typeof(EditText))
```
willsb: Early return instead of nesting. I think that's the problem of doing things in Kotlin first 😄 

```csharp
if (child?.GetType() != type(EditText))
    return;
```

Toggl.Giskard/Views/TextInputLayoutWithHelperText.cs
```diff
@@ -0,0 +1,231 @@
...
+            {
+                if (!TextUtils.IsEmpty(helperText))
```
willsb:
```csharp
if (TextUtils.IsEmpty(helperText))
    return;
```

willsb: Also, why is this not an extension method or why are you not using `string.IsNullOrWhiteSpace`?

Toggl.Giskard/Views/TextInputLayoutWithHelperText.cs
```diff
@@ -0,0 +1,231 @@
...
+        }
+
+        public bool HelperTextEnabled
+        {
+            set
+            {
+                if (value == helperTextEnabled) return;
+                if (value && ErrorEnabled)
+                {
+                    ErrorEnabled = false;
```
willsb: `ErrorEnabled = !value && ErrorEnabled`?

Toggl.Giskard/Views/TextInputLayoutWithHelperText.cs
```diff
@@ -0,0 +1,231 @@
...
+                }
+
+                if (value)
+                {
+                    helperTextTextView = new AppCompatTextView(Context);
```
willsb: Why not cache this?

Toggl.Giskard/Views/TextInputLayoutWithHelperText.cs
```diff
@@ -0,0 +1,231 @@
...
+                    var useDefaultColor = false;
+                    try
+                    {
+                        TextViewCompat.SetTextAppearance(helperTextTextView, textHelperTextAppearance);
+                        if (MarshmallowApis.AreAvailable && helperTextTextView.TextColors.DefaultColor == Color.Magenta)
```
willsb: Why are you checking for Magenta? That's really cryptic

Toggl.Giskard/Views/TextInputLayoutWithHelperText.cs
```diff
@@ -0,0 +1,231 @@
...
+                        Color defaultTextColor = new Color(ContextCompat.GetColor(Context, Resource.Color.defaultText));
+                        helperTextTextView.SetTextColor(defaultTextColor);
+                    }
+
+                    if (helperTextColor != null)
+                    {
+                        helperTextTextView.SetTextColor(helperTextColor);
+                    }
+
+                    if (helperTextSize > 0)
```
willsb: `string.IsNullOrEmpty` would convey your intention much better here

Toggl.Giskard/Views/TextInputLayoutWithHelperText.cs
daividssilverio: I was using the old Android java stuff we used even before kotlin 🙈

willsb: Then go for `string.IsNullOrWhitespace` :)

Toggl.Giskard/Views/TextInputLayoutWithHelperText.cs
daividssilverio: I was following the implementation for the setErrorEnabled on the original TextInputLayout. It seems that unresolved theme for the TextAppearance attrs will result in Magenta instead of throwing an exception.

Toggl.Giskard/Views/TextInputLayoutWithHelperText.cs
daividssilverio: I'll rename this property. This is the textSize attribute set from xml for the helperText font size, not the current text length.

daividssilverio:
To answer your comment. Basically, we could use the TextView, but we wouldn't have the activation colour for errors and helper texts as easily. The activation colour for this widget is either based on themes or the colour of the text on the error label.

Since we are already using the TextInputLayout for the floating label (hint) + password toggle, it makes sense to also use its error indicator as well, since it can be easily styled (as long as you give it enough room) and is nicely animated as well.

The only thing missing is the helper text indicator, which will be probably added to the design library in the future, since it is described in the guidelines.

Unfortunately, the error assistive text is either always enabled (results on the TextInputLayout taking extra space, which is ok if we have the helper text as well) or added/removed at runtime, which might get a little jumpy, since we would also have to animate the TextView

willsb:
I see. Fair enough, that makes sense.

Toggl.Giskard/Views/TextInputLayoutWithHelperText.cs
kornelijepetak: Oooh, the Kotlin reference is painful :)

Toggl.Giskard/Views/TextInputLayoutWithHelperText.cs
daividssilverio: Done. I've used the backing errorEnabled though.

Toggl.Giskard/Views/TextInputLayoutWithHelperText.cs
daividssilverio: No reason. Cached!

Toggl.Giskard/Views/TextInputLayoutWithHelperText.cs
daividssilverio: here o/

daividssilverio: @kornelijepetak

PR #2302, opened 2018-05-28
Fix singleton data store

simonrozsival:
There were 3 problems:

- after pulling the data from the server, the value was not updated (`BatchUpdate` was not overriden)
- the code was ugly
- there were no tests

I tried to address these problems and make the situation better.

Squash whenever


Toggl.Foundation/DataSources/SingletonDataSource.cs
```diff
@@ -58,6 +59,10 @@ public override IObservable<TThreadsafe> Overwrite(TThreadsafe original, TThread
             TThreadsafe original, TThreadsafe entity)
             => base.OverwriteIfOriginalDidNotChange(original, entity).Do(handleConflictResolutionResult);
 
+        public override IObservable<IEnumerable<IConflictResolutionResult<TThreadsafe>>> BatchUpdate(
+            IEnumerable<TThreadsafe> entities)
+            => base.BatchUpdate(entities).Do(results => results.ForEach(handleConflictResolutionResult));
```
willsb: Why would this ever be needed? 🤔

willsb: Should we maybe consider moving this method to the `IDataSource` interface?

simonrozsival: E.g., pulling preferences from the server. We need it in the base data source, unfortunately 😢

willsb: The name here is really misleading, though. Couldn't we call the repository directly from the SingletonDataSource to avoid this weird naming, at least in the DataSource layer?

simonrozsival: I agree with you that it is ugly and I don't like it either. Changing this will require additional changes - like creating a specific state for persisting items which are stored using the singleton DS. But it might be worth it. I will create a separate PR to do that though,.

willsb:
Putting it `on-hold` so others know this doesn't need reviewing for the time being

kornelijepetak:
I see this has moved to `on-hold` status. I just wanted to say this indeed did solve the issue I had with preferences being updated. I have merged this branch into mine to test this and it worked. So, @simonrozsival I will just reference this in my PR and when you get to the final version, let me know to rereview it... For now, I will just revert the changes until you are satisfied with the solution.

Issue #2308, opened 2018-05-28
Editing running time entry time should now allow a start time in the future


amulware:
Right now I can select a start time in the future, and even a date, without any issues.

PR #2309, opened 2018-05-28
Fix UTC and duration problems in start/stop/duration popup

### LINKED ISSUE #2272: Manual mode: editing the end time doesn't respect the duration
In **manual mode**, the following occurs: 
1. Tap on the '+' button to add an entry
2. Tap on '0:30:00' to edit the duration
3. Tap on '+ 30m' a couple of times (get to 2:00:00, for example) 
4. Tap on the (updated) end time tab and change the end time (you can see at the top bar that duration changes as well)
6. Tap on the 'Duration' tab 
7. Tap '+30m'

-----> As a result, the 30 minutes that should be added to the previously edited Stop time are added to the 2:00:00 that I ended up with while previously tinkering with the duration, resulting in 2:30:00.

The 30 minutes to be added to the edited end time I chose in the end time tab, so these all work together.

If I *first edit the end time* with the nice :clock1: (middle tab) and _then_ tinker with the duration, I can successfully add whatever minutes I wanted to add to the new, edited end time.

### LINKED ISSUE #2273: Date-time-duration popup: Start time switches to UTC
1. Start a new TE
2. Go through Start and End time tabs -> All looks good (both times are in my TZ)
3. Go to Duration and tap on '+ 30m' a couple of times (get to 2:00:00, for example) -> You can see that Start time changed
4. Go to Start time -> It's switched to UTC again

It shows Start time it UTC only in the picker, on the tabs it's OK.

kornelijepetak:
Closes #2273 
Closes #2272 

Basically, now the DatePicker and TimePicker will handle stuff in local time (which they should). Before, some of the conversion was being done in VM which is not a good place for that IMO.

Please do a functional...

🦑 Squash whenever.


willsb:
Correct me if I'm wrong, @kornelijepetak, but this also fixes #2308, right?

Toggl.Giskard/Views/TogglDroidDatePicker.cs
```diff
@@ -0,0 +1,132 @@
...
+using Android.Widget;
+using MvvmCross.Platform.Droid.Platform;
+using Toggl.Giskard.Extensions;
+using Toggl.Multivac;
+
+namespace Toggl.Giskard.Views
+{
+    [Register("toggl.giskard.views.togglDroidDatePicker")]
+    public class TogglDroidDatePicker
+        : DatePicker
```
willsb: No need to inherit from Mvx?

kornelijepetak:
@willsb indeed it does. I have added it to the description.

Toggl.Giskard/Views/TogglDroidDatePicker.cs
```diff
@@ -0,0 +1,132 @@
...
+        , DatePicker.IOnDateChangedListener
+    {
+        private bool isInitialized;
+
+        private readonly long DefaultMinimum =
```
willsb: Why not `defaultMinimum`?

Toggl.Giskard/Views/TogglDroidDatePicker.cs
```diff
@@ -0,0 +1,132 @@
...
+            new DateTimeOffset(1900, 1, 1, 0, 0, 0, TimeSpan.Zero)
+            .ToUnixTimeMilliseconds();
+
+        private readonly long DefaultMaximum =
```
willsb: Ditto

Toggl.Giskard/Views/TogglDroidDatePicker.cs
```diff
@@ -0,0 +1,132 @@
...
+                 * and the dates are different, which is bad logic.
+                 * https://stackoverflow.com/a/19722636/93770
+                 * 
+                 * Also, because of the bug in DatePicker widget, make 
+                 * sure this order is not reversed. MaxDate must be
+                 * set before MinDate.
+                 */
+
+                MaxDate = DefaultMaximum;
+                MaxDate = ((DateTimeOffset)boundaries.Maximum.Date).ToUnixTimeMilliseconds();
```
willsb: Why the cast?

Toggl.Giskard/Views/TogglDroidDatePicker.cs
```diff
@@ -0,0 +1,132 @@
...
+
+                MinDate = DefaultMinimum;
+                MinDate = ((DateTimeOffset)boundaries.Minimum.Date).ToUnixTimeMilliseconds();
+
+                BoundariesChanged?.Invoke(this, null);
```
willsb: This is for Mvx purposes, corred?

Toggl.Giskard/Views/TogglDroidDatePicker.cs
kornelijepetak: I think not. This class was inspired by the Mvx though... However, they are using `DateTime` and `TimeSpan` in Date and Time pickers, respectively. And I wanted to do pure `DateTimeOffset` solution.

Toggl.Giskard/Views/TogglDroidDatePicker.cs
kornelijepetak: Because the `.Date` is a `DateTime`, and not a `DateTimeOffset`, so it has no unix milliseconds.

Toggl.Giskard/Views/TogglDroidDatePicker.cs
kornelijepetak: yes, indeed. Like I said, it's inspired by the code they use.

PR #2324, opened 2018-05-29
🐛 Fix autocorrect replacing suggesting with something else

### LINKED ISSUE #2158: Autocorrect causing errors when using existing description
replicated using the same word as the client "Kettlebells"
Created time entry with description "Kettlebells"
Stopped entry
Started new time entry
Selected "Kettlebells" description autocorrect changed the first few letters of the description to the letters it interpreted that I was attempting to enter, one time it was "Let tlebells" and another was "Jet tlebells"

Case/691857

sofokli1:
Closes #2158

Autocorrect was replacing the selected suggestion as if the user typed something and pressed space. I just disabled autocorrect for the text field in IB.

Ps: If you are trying to reproduce in develop and testing the branch.
1. Enable autocorrect for your device
2. Start a time entry with description `Kettlebells` and stop it
3. Start an other time entry with description `Let maples` and stop it
4. Try a new entry with `Kettle` and select the suggestion

Since it is based on autocorrect it is on a hit and miss bases.

Squash whenever


sofokli1:
I am open to suggestions if this solution should be applied to other textfields. if so, let me know.

amulware:
I'm pretty sure we would like to have auto correction at least when entering description though. :/

Also: please branch from and pr against `hotfix-1.6.1`.

sofokli1:
@amulware I tried an other approach but couldn't get a good result out of it. So the idea was to `resign as first responder` set the textfield as `autocorrect = false` change the `value` and set `autocorrect = default` and `set as first responder`. I tried to do this but all the changes happen for all key types not just when selecting from the list of suggestions (maybe i am not applying the changes in the right place). But the main side effect is the bar on the top of the keyboard jumps constantly due to `first responder` changes. I found somewhere that in order for `autocorrect` flag changes to take effect the field should not be first responder. Would you like to dig dipper into this or just make a branch with the simple autocorrection off by default (same as this PR)?

amulware:
Ok, sure. Let's disable the auto correct for now to get rid of the bug. In the long term it would be great if we can find a solution that makes it work properly though. 🤔

Issue #1888, opened 2018-04-14
Update to MvvmCross 6.0


amulware:
It's pretty cool.

amulware:
Let's make sure we test this THOROUGHLY before merging.

willsb:
MvvmCross 6.0.1 is not yet unstable enough. We should wait for newer versions and check again.

PR #2326, opened 2018-05-29
Clean up app initialization logic

willsb:
While migrating code for #1888, I noticed how ugly our foundation init logic was. This PR is a proposal to make it nicer, but there are no functional changes.

Squash whenever.


Toggl.Foundation/TogglFoundation.cs
```diff
@@ -0,0 +1,225 @@
...
+            public Builder WithMailService(IMailService mailService)
+            {
+                MailService = mailService;
+                return this;
+            }
+
+            public Builder WithTimeService(ITimeService timeService)
+            {
+                TimeService = timeService;
+                BackgroundService = BackgroundService ?? new BackgroundService(timeService);
```
kornelijepetak: I have mixed feelings about this. It is not evident from the `WithTimeService` that it will under the hood potentially make changes to `BackgroundService` and also that the order in which builder methods are called is relevant to the side effects of this method. 🤷‍♂️ 

However, I am aware that this is called basically only once (+ tests) so it may not be such an issue... But still I wanted to voice my sentiment.

simonrozsival: I would prefer this to be explicit - this seems like a tricky side effect which could make the person who reads the code of the `Setup` classes wonder where the `BackgroundService` comes from. The same applies to the `ApiEnvironment` => `ApiFactory`.

willsb: Guys, there is a method to make it explicit! You can still specify your own BackgroundService, if you want to. And the order doesn't matter, @kornelijepetak, check the code again and you'll see: If you specify your own BackgroundService, it will be the one used when building, regardless.

kornelijepetak: True. Like I said, I don't mind this that much as it will be called only once and won't change much. I just didn't understand why not simply expect the BackgroundService to be initialized on its own and not rely on a default being provided in case nothing is specified.

I will change my review into approval. It seems you have decided 😄

simonrozsival: I would prefer if all of the services were defined explicitly and it was obvious from the `Setup` class which services were registered and which weren't - there were several bugs where we registered services the wrong way (e.g., #1948. #1927), so I think it should be always easy to find where and when a service is registered and not make it a side effect of something which seems unrelated.

Issue #2329, opened 2018-05-30
Make small-slice grouping in reports smarter


amulware:
Follow-up to #2327

As it stands, grouping projects that are <10% can have several weird outcomes:

- a single project is less than 10% -> might as well show it, or at least the colour
- several almost 10% projects are grouped, for a very large gray area

In general, I think we need a better solution.

Maybe we should always show project colours for all of them, but just disable the text under a certain percentage to not make things overlap?

Or maybe we should show a fixed number of projects ungrouped?

Or maybe we should lower the 10% to 5% or whatever to make it a bit nicer?

Ideas welcome.

amulware:
We decided to do something else instead. Issue coming soon.

PR #2330, opened 2018-05-30
Fix reports grouping

### LINKED ISSUE #2327: Ungroup projects in reports legend
Follow up to #1800

We only want projects grouped in the pie chart, not in the legend.

In the legend (list below the pie chart) we want all the projects of the report shown, no matter how small their percentage.

simonrozsival:
Closes #2327 
Ref #2329 

Squash whenever


Toggl.Foundation.MvvmCross/ViewModels/ReportsViewModel.cs
```diff
@@ -219,7 +224,7 @@ private void onPreferencesChanged(IDatabasePreferences preferences)
         private IReadOnlyList<ChartSegment> groupSegments()
         {
             var otherProjects = segments.Where(segment => segment.Percentage < minimumPieChartSegmentPercentage).ToList();
-            if (otherProjects.Count == 0)
+            if (otherProjects.Count <= 1 || otherProjects.Count == segments.Count)
```
amulware: If all are <10%, we show them all? We really need to figure out a better rule than the 10% limit...

PR #2335, opened 2018-05-30
Fix autocorrect replacing suggesting with something else

### LINKED ISSUE #2158: Autocorrect causing errors when using existing description

sofokli1:
Closes #2158

Autocorrect was replacing the selected suggestion as if the user typed something and pressed space. I just disabled autocorrect for the text field in IB.

Ps: If you are trying to reproduce in develop and testing the branch.
1. Enable autocorrect for your device
2. Start a time entry with description `Kettlebells` and stop it
3. Start an other time entry with description `Let maples` and stop it
4. Try a new entry with `Kettle` and select the suggestion

Since it is based on autocorrect it is on a hit and miss bases.

I am open to suggestions if this solution should be applied to other textfields. if so, let me know.

Squash whenever


amulware:
Yes, please.

PR #2339, opened 2018-05-30
Fix TE description extraction method

### LINKED ISSUE #2274: Start view: Project gets deleted after moving the cursor to another place and start typing description
_This happens both in the Start view in timer and in manual mode._

1. Start a new TE
2. Tap '# Search tags' and select a tag or two
3. Tap '@ Search projects' and select a project
4. Move the cursor at the end of the line and try to add a description

-----> First, the cursor seems to move to the beginning of the input field; when trying to type again the second time, the project gets deleted.

Or another case:

1. Start a new TE
2. Tap '# Search tags' and select a tag or two
3. Tap '@ Search projects' and select a project
4. Move the cursor between the project and tags

-----> When I start typing something inbetween the project and tag, for example, the behavior gets interesting: First, the cursor seems to move to the middle of the tag; when typing the 2nd letter, the project gets deleted.

daividssilverio:
Closes #2274

What does it do:

While the user is trying to input the TE description, he/she might have something like this:

"My task `projectY` `tag1` `tag2`"

And then, the user might decide to move his cursor to the middle of the projects and tags tokens and type something, resulting in something like this: 

* "My task `projectY`X`tag1` `tag2`"
* "My task `projectY` `tag1`X`tag2`" or
* "My task `projectY` `tag1` `tag2`X"

The implementation of the `GetDescription` in this PR transforms any of the examples above into "My task X".

The old implementation of the `GetDescription` method is considering only text typed before any Token, and in the examples above, the resulting description would be "My task ", which would cause problems while updating the `TextFieldInfo` in the `BaseTextFieldInfoTargetBinding` and `EditTextTextFieldInfoTargetBinding` triggering feedback loops, that sometimes result in some of the Tokens being removed, the cursor appearing in the middle of a Token, or worse, crashing the App.

Please do a functional test as well.

🦑 if you can.


Toggl.Giskard/Autocomplete/TokenExtensions.cs
```diff
@@ -17,10 +16,27 @@ public static class TokenExtensions
...
             var spannable = self.TextFormatted as SpannableStringBuilder;
-            var nextTransition = spannable.NextSpanTransition(0, spannable.Length(), Class.FromType(typeof(CharacterStyle)));
+            if (spannable == null) return self.Text;
 
-            var isNextTransitionNormalText = !spannable.GetSpans(0, nextTransition, Class.FromType(typeof(TokenSpan))).Any();
-            return isNextTransitionNormalText ? self.TextFormatted.SubSequence(0, nextTransition) : "";
+            var transitions = spannable.GetSpans(0, spannable.Length(), Class.FromType(typeof(TokenSpan)));
+            if (transitions.Length == 0) return spannable.ToString();
+
+            var description = string.Empty;
```
kornelijepetak: Please use StringBuilder.
Concatenating strings with += creates a lot of intermediate unused strings.

Toggl.Giskard/Autocomplete/TokenExtensions.cs
```diff
@@ -17,10 +17,27 @@ public static class TokenExtensions
...
+            if (spannable == null) return self.Text;
 
-            var isNextTransitionNormalText = !spannable.GetSpans(0, nextTransition, Class.FromType(typeof(TokenSpan))).Any();
-            return isNextTransitionNormalText ? self.TextFormatted.SubSequence(0, nextTransition) : "";
+            var transitions = spannable.GetSpans(0, spannable.Length(), Class.FromType(typeof(TokenSpan)));
+            if (transitions.Length == 0) return spannable.ToString();
+
+            var description = new StringBuilder();
+            var textParts = transitions.AsEnumerable()
+                .Select(i => (spanIndexStart: spannable.GetSpanStart(i), spanIndexEnd: spannable.GetSpanEnd(i))).ToList();
```
willsb: Why `.ToList` here and again at the end?

I'd rather you manually add the missing items afterwards, without the need to call ToList() twice.

Toggl.Giskard/Autocomplete/TokenExtensions.cs
```diff
@@ -17,12 +20,27 @@ public static class TokenExtensions
         public static string GetDescription(this EditText self)
         {
             var spannable = self.TextFormatted as SpannableStringBuilder;
-            var nextTransition = spannable.NextSpanTransition(0, spannable.Length(), Class.FromType(typeof(CharacterStyle)));
+            if (spannable == null) return self.Text;
```
kornelijepetak: Please, only void-returns in the same line.
If it was up to me, I'd ban all returns/continues/breaks in the same line. But unfortunately, it's not up to me.

Toggl.Giskard/Autocomplete/TokenExtensions.cs
```diff
@@ -17,12 +20,27 @@ public static class TokenExtensions
...
 
-            var isNextTransitionNormalText = !spannable.GetSpans(0, nextTransition, Class.FromType(typeof(TokenSpan))).Any();
-            return isNextTransitionNormalText ? self.TextFormatted.SubSequence(0, nextTransition) : "";
+            var tokenSpans = spannable.GetSpans(0, spannable.Length(), Class.FromType(typeof(TokenSpan)));
+            if (tokenSpans.Length == 0) return spannable.ToString();
+
+            var descriptionStringBuilderBeforeTokenSpans = new StringBuilder(spannable.SubSequence(0, spannable.GetSpanStart(tokenSpans.First())));
```
kornelijepetak: including `StringBuilder` in name sounds a bit 🇭🇺 to me.

daividssilverio: Is 🇮🇹 good or bad?

PR #2342, opened 2018-05-30
Reactive Dialogs

willsb:
As part of the big RxFactor, we need to have more Reactive services. This changes our dialogs to use Observables. While simple at heart, this change allows us to chain dialog calls in another observables if needed while not preventing existing code to consume it as a Task.

Squash whenever


Toggl.Giskard/Services/DialogService.cs
```diff
@@ -11,40 +13,54 @@ namespace Toggl.Giskard.Services
...
+                            observer.OnNext(false);
+                            observer.OnCompleted();
+                        });
+                    }
 
-            return tcs.Task;
+                    var dialog = builder.Create();
+                    dialog.CancelEvent += (s, e) =>
+                    {
+                        observer.OnNext(true);
```
simonrozsival: This should be `false`, right?

daividssilverio: Yes, it should.

willsb: :eyes:

PR #2346, opened 2018-05-31
Use ImmutableStack in DurationFieldInfo

Odrakir:
This is very simple and totally dispensable. But it was something that took me a minute to understand when checking the code, so I decided to change it.

Also, we might have more use for immutable collections in the future.


Odrakir:
Also... This file is in the root of Foundation, shouldn't it be somewhere else?

Issue #1335, opened 2018-01-25
Sync archived projects


amulware:
When pulling time entries with unknown project ids, we should create 'temporary'/archived/disabled projects inside realm, that have correct ids, but otherwise no data. That will ensure we don't lose references to projects. We may want to show '(inaccessible project)' in the UI for these.

Then, after pulling those TEs, if we are missing projects, we should fetch those one by one from backend with the single-project endpoint, and if we have access to them persist them in realm, overriding the temporary ones.

From that point on the time entries will be rendered correctly. But we have to make sure not to include those projects in auto completion, etc.

If for some reason we really cannot get the project (I'm not sure if this is possible, might be worth checking with backend), we can continue to use the temporary object and show '(inaccessible project)'.

We have to also figure out how to show this in the edit view when the user tries to edit the project of a time entry with an archived/unknown project.

amulware:
If a time entry shows with an archived projects, it's probably best to render the project name as `My Project (archived)` to be really clear about it, and to better 'explain' to the user why they cannot use the continue button on it for example.

amulware:
Two thoughts:

## Failing Sync
In case sync gets interrupted after storing the dummy projects, it should still try and fetch them next time, so it has to grab the projects to try and fetch from the DB. If a project is entirely inaccessible, this might mean trying to pull it with every pull sync, but not sure what we can do about that. Giving it a flag that we already tried syncing means that if we get access back in the future, we wouldn't notice, which is also bad.

## Realm weirdness
With the way Realm.Get works inside transactions, we might have to persist the fake projects before we insert TEs into the list. I think using two transactions for this should be (strictly logically speaking) fine. I can't think of a way this could fail.

I.e. do something like:

- get list of all project ids of the pulled TEs
- batch update projects to ignore existing projects, and create non-existing ones (as fake/ghost/whatever)
- now persist all TEs, and project ids will resolve fine

simonrozsival:
It should be easy to pull projects, pull time entries, look for the projects that the pulled TEs refer to, look if these projects are either among the pulled projects or in the DB projects, and create ghost entities for those, which don't exist.

In fact, when we persist TEs, the projects are already persisted, so we need to look just into the DB. The easiest way to hammer this into the pull sync flow would be to create a specific state for time entries, which will first make sure that we have all the projects in the DB and then use the regular persist state implementation to do its job.

simonrozsival:
[...]

PR #2351, opened 2018-05-31
Ghosts

simonrozsival:
Ref #1335 

This PR adds a syncing state which creates a 👻 projects (we don't get them during normal pull sync of projects because they are e.g. archived) which will be fetched using the reports API (will be part of a follow-up PR).

Questions:
- is it OK to show the "(inaccessible project)" until we fetch the project?
- any naming suggestions for the `RefetchingNeeded` sync status?
- I reused the "no project" color (dark gray) - is that OK?


Toggl.Foundation/Sync/States/Pull/CreateGhostProjectsState.cs
```diff
@@ -0,0 +1,60 @@
...
+        {
+            Ensure.Argument.IsNotNull(dataSource, nameof(dataSource));
+
+            this.dataSource = dataSource;
+        }
+
+        public IObservable<ITransition> Start(IFetchObservables fetch)
+            => fetch.GetList<ITimeEntry>()
+                .SingleAsync()
+                .SelectMany(timeEntries => timeEntries)
```
willsb: We really need `CommonFunctions.Identity` 🤷‍♂️

Toggl.Foundation/Sync/States/Pull/CreateGhostProjectsState.cs
```diff
@@ -0,0 +1,60 @@
...
+                .Distinct(timeEntry => timeEntry.ProjectId)
+                .SelectMany(createGhostForTimeEntryIfProjectIsNotInDatabase)
+                .ToList()
+                .Select(_ => FinishedPersisting.Transition(fetch));
+
+        private IObservable<Unit> createGhostForTimeEntryIfProjectIsNotInDatabase(ITimeEntry timeEntry)
+            => timeEntry.ProjectId.HasValue
+                ? dataSource.GetById(timeEntry.ProjectId.Value)
```
willsb: We need an extension like `NotNullable` to avoid this funkyness here:

```
public IObservable<TValue> NotNullabel<TOptional, TValue>(this IObservable<TOptional> observable)
    where TOptional : Nullable<TValue>
    => observable.Where(x => x != null).Select(x => x.Value);
```

Toggl.Foundation/Sync/States/Pull/CreateGhostProjectsState.cs
```diff
@@ -0,0 +1,60 @@
...
+                    .Catch((EntityNotFoundException exception) => createGhostProjectFor(timeEntry))
+                    .Select(_ => Unit.Default)
```
willsb: A `SelectUnit` extension would also be nice, actually. I'm gonna create a PR with all of our Reactive extensions, if you don't mind.

Toggl.Foundation/Sync/States/Pull/CreateGhostProjectsState.cs
```diff
@@ -0,0 +1,60 @@
...
+                : Observable.Return(Unit.Default);
+
+        private IObservable<IThreadSafeProject> createGhostProjectFor(ITimeEntry timeEntry)
+        {
+            if (!timeEntry.ProjectId.HasValue)
```
willsb: Since you are only using the projectId, I don't even see the need of passing the entire timeEntry down the chain

Toggl.Foundation/Sync/States/Pull/CreateGhostProjectsState.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        public CreateGhostProjectsState(IProjectsSource dataSource)
+        {
+            Ensure.Argument.IsNotNull(dataSource, nameof(dataSource));
+
+            this.dataSource = dataSource;
+        }
+
+        public IObservable<ITransition> Start(IFetchObservables fetch)
+            => fetch.GetList<ITimeEntry>()
```
willsb: What timeEntries does `fetch` return? Everything in the DB? Because I see that you are not applying any filters other than distinct on the projectId, so maybe that's it?

Toggl.PrimeRadiant/SyncStatus.cs
```diff
@@ -1,10 +1,10 @@
...
 {
     public enum SyncStatus
     {
-        InSync,
-        SyncNeeded,
-        SyncFailed
+        InSync = 0,
+        SyncNeeded = 1,
+        SyncFailed = 2,
+        RefetchingNeeded = 3
```
willsb: Can you clarify why a new status is needed?

Toggl.Foundation/TogglSyncManagerFactory.cs
```diff
@@ -135,7 +137,8 @@ public static class TogglSyncManager
             transitions.ConfigureTransition(persistTags.FinishedPersisting, persistClients.Start);
             transitions.ConfigureTransition(persistClients.FinishedPersisting, persistProjects.Start);
             transitions.ConfigureTransition(persistProjects.FinishedPersisting, persistTasks.Start);
-            transitions.ConfigureTransition(persistTasks.FinishedPersisting, persistTimeEntries.Start);
+            transitions.ConfigureTransition(persistTasks.FinishedPersisting, createGhostProjects.Start);
```
willsb: Why run this every time? Wouldn't it be better to only run it when needed? Or is it the case where there's no clear way to know if it's needed so we just do it all the time for safety®? Can't recall right now

Toggl.Foundation/Sync/States/Pull/CreateGhostProjectsState.cs
simonrozsival: I also use the workspace ID. I might pass both of the IDs as params, why not.

Toggl.Foundation/TogglSyncManagerFactory.cs
simonrozsival: It will mostly just "skip" this step because it only refetches the projects once per day.

simonrozsival: We must do this before every time we fetch time entries, because there might be new references to projects which don't exist in our DB, and persisting TEs would crash.

Toggl.PrimeRadiant/SyncStatus.cs
simonrozsival: None of the previous sync states is suitable for this task:
- in sync is obviously not good
- sync needed would be pushed to the server
- sync failed would mix with locally created projects which can't be pushed

Toggl.Foundation/Sync/States/Pull/CreateGhostProjectsState.cs
simonrozsival: This is the list of TEs fetched from the API (sometimes all TEs for the prev two months, sometimes a few TEs since some date).

willsb: How does this ensure that all the time entries in our database have a valid project, however?

simonrozsival:
It occured to me that we need to also modify the `PreferNewer` conflict resolver - if the local TE sync status is `RefetchNeeded` then we should always override it - e.g., when a project is re-activated.

Toggl.Foundation/Sync/States/Pull/CreateGhostProjectsState.cs
```diff
@@ -0,0 +1,59 @@
...
+                .SingleAsync()
```
amulware: Are we sure this won't cause the TE fetch to run twice, or for the actual TE persist state not to get any data?

Toggl.Foundation/Sync/States/Pull/CreateGhostProjectsState.cs
```diff
@@ -0,0 +1,59 @@
...
+                .SelectMany(CommonFunctions.Identity)
+                .Distinct(timeEntry => timeEntry.ProjectId)
+                .SelectMany(createGhostForTimeEntryIfProjectIsNotInDatabase)
+                .ToList()
+                .Select(FinishedPersisting.Transition(fetch));
+
+        private IObservable<Unit> createGhostForTimeEntryIfProjectIsNotInDatabase(ITimeEntry timeEntry)
+            => timeEntry.ProjectId.HasValue
+                ? dataSource.GetById(timeEntry.ProjectId.Value)
+                    .Catch((EntityNotFoundException exception) => createGhostProjectFor(timeEntry.WorkspaceId, timeEntry.ProjectId.Value))
```
amulware: No way to batch update these? I.e. create a list of ghost projects, batch update them so it always prefers the existing one in the DB and only creates if there is none?

amulware: I ask because that seems a lot more efficient than throwing lots of exceptions.

Toggl.Foundation/Sync/States/Pull/CreateGhostProjectsState.cs
```diff
@@ -0,0 +1,59 @@
...
+                    .SelectUnit()
+                : Observable.Return(Unit.Default);
+
+        private IObservable<IThreadSafeProject> createGhostProjectFor(long workspaceId, long projectId)
+        {
+            var ghost = Project.Builder.Create(projectId)
+                .SetName(Resources.InaccessibleProject)
+                .SetWorkspaceId(workspaceId)
+                .SetColor(Color.NoProject)
+                .SetActive(false)
```
amulware: This is basically the same as the archived flag. That's not future proof for other ghost entities in the future though, is it? Time entries, workspaces, tags, clients... How will those be encoded? Maybe with another separate sync-state? Or the same sync state as these projects?

simonrozsival: This PR deals just with projects - I don't know how it would work for other entities. We would definitely have to do some entity specific adjustments. The ghost projects must be inactive, because otherwise they would be shown in autocomplete suggestions, which does not make any sense.

I would say we should think of generalization of this state once we start the implementation of ghosts for other entity types.

Toggl.Foundation/Sync/ConflictResolution/PreferNewer.cs
```diff
@@ -33,7 +33,8 @@ public ConflictResolutionMode Resolve(T localEntity, T serverEntity)
             if (localEntity == null)
                 return Create;
 
-            if (localEntity.SyncStatus == SyncStatus.InSync)
+            if (localEntity.SyncStatus == SyncStatus.InSync
+                || localEntity.SyncStatus == SyncStatus.RefetchingNeeded)
```
amulware: Maybe this is good to have in either case, but isn't the condition below already making sure we update?

Toggl.Foundation/Sync/States/Pull/CreateGhostProjectsState.cs
```diff
@@ -0,0 +1,59 @@
...
+                .SetAt(default(DateTimeOffset))
```
amulware: Is `At` nullable? Otherwise this line doesn't do anything, does it?

Toggl.Foundation/Sync/States/Pull/CreateGhostProjectsState.cs
simonrozsival: The fetch observables are replayed (https://github.com/toggl/mobileapp/blob/develop/Toggl.Foundation/Sync/States/Pull/FetchAllSinceState.cs#L38) - which means that it is OK.

The bad thing is that if this state was used with "incorrect" fetch observables, it would probably break (the persist state wouldn't get any data). Maybe we could change the `FetchObservables` class/interface to return `IConnectableObservable<T>` instead of just `IObservable<T>`? I know that still doesn't imply that the observable will be replayed for every subscription, but I'm not sure if there is anything more specific we could do.

amulware: I'm ok assuming that the fetch observables are configured correctly. There are tests for it.

Toggl.Foundation/Sync/States/Pull/CreateGhostProjectsState.cs
simonrozsival: In principle it is possible to do it. It will require allowing passing custom conflict resolution mode to the `BatchUpdate` method of the data source (which I tried to hide as much as possible), or adding a different method to the data source - maybe just to the projects data source.

We might want to do this in a separate PR if we decide to do it that way.

Another option which wouldn't include exceptions is to not use `GetById` which throws, but directly `GetAll(proj => proj.Id == id).Any()` - that should be enough for fast checking of existence of an entity.

Toggl.Foundation/Sync/States/Pull/CreateGhostProjectsState.cs
simonrozsival: `At` is nullable and validation throws when it does not have a value.

Toggl.Foundation/Sync/States/Pull/CreateGhostProjectsState.cs
amulware: The GetAll runs in memory though, doesn't it? So it'll fetch the entire project list for each project? That seems worse. If you can guarantee that the GetAll lambda runs in Realm, then I'd be ok with that solution. Otherwise, I rather keep the current, or pull the whole list into memory once and perform the lookups on that?

Toggl.Foundation/Sync/ConflictResolution/PreferNewer.cs
simonrozsival: Without this the entity might be easily ignored (see #2355). We can discuss the usage of the `At` attribute (it makes sense to me to use it to mark when was the last time we tried fetching it so we don't end up in an infinite loop for a project which can't be obtained from the server) - if we didn't use it and update it, then it you would be right and this condition would be unnecessary.

Toggl.Foundation/Sync/States/Pull/CreateGhostProjectsState.cs
simonrozsival: It should run in Realm. I will check it.

simonrozsival: I can confirm that the predicate is passed directly to the repository (`Adapter.GetAll().Where(predicate)` - `GetAll()` returns `IQueryable<T>` and the predicate is a supported one according to https://realm.io/docs/dotnet/latest/api/linqsupport.html#predicate-operations)

Toggl.Foundation/Sync/States/Pull/CreateGhostProjectsState.cs
simonrozsival: The ghost entity is marked with the sync status `RefetchingNeeded` - maybe not a good name. Maybe we can call it just `Ghost` or something. We can then easily select/filter out these entities.

The `Activity` (this means exactly archived project) is here "just" for suggestions. Maybe we should update project suggestions to exclude also ghosts? Then we won't need to have "archived" 👻 ...

amulware: Let's leave it like it is for now. Incremental steps.

amulware:
My approval indicates that the things questioned in the original comment are fine for now. They are easily changed (though not retroactively... maybe we should handle it differently, but as long as we don't release, this is fine).

Issue #1486, opened 2018-02-12
Project names not updating on Timer page


ApocalypticFish:
If a projects name is changed the app will refresh this on the most recent entry.

(Here I changed "Work" to "Work0")
![image](https://user-images.githubusercontent.com/18132080/36089787-dd4db1c4-0fdd-11e8-8b05-b37aada62d73.png)

(as you can see when you scroll further down that the other entries haven't updated even if I have forced sync)
![image](https://user-images.githubusercontent.com/18132080/36089806-f1bfa964-0fdd-11e8-92bb-22f118856d67.png)

(But when you click into the entry it displays the correct project)
![image](https://user-images.githubusercontent.com/18132080/36089818-fdb574b0-0fdd-11e8-82a6-40ea868afc7a.png)

amulware:
I take it you updated the project name online, and had the app sync for it to notice the change in the first place?

ApocalypticFish:
Yeah, updated the project on web.
Closed and reopened the app and it refreshes the data, however.

amulware:
Yeah, this is just a UI glitch where it doesn't bother updating the entries. I'm actually surprised it updates the last entry. What happens if you edit an older entry with that project (change description or whatever), then rename the project. I predict it would update the project name only on the last edited entry, not actually the newest one.

ApocalypticFish:
I can just open the project and confirm changes without actually changing anything and it updates on the timer page. it does this for each entry as it's opened and confirmed until they all reflect the change

amulware:
That's not what I meant. I mean that:
- you report the project name change only automatically happens for the last entry
- I think it happens for the entry that was EDITED last

This is due to the way we pull data. We will always re-pull the TE changed last (technical: since we use it's `since` date - it's a bit silly but the only way we can be sure we don't miss changes).

In either case, the bug is clear enough.

I'm not sure what the best way of fixing it is. Ideally, we could update time entries with a project, if said project changes. Alternatively, we could also just reload the whole log after completing a pull sync? If there are no graphical glitches, the latter might be the easiest and work just fine.

simonrozsival:
I think our best strategy to avoid this kind of bugs would be to redraw the whole log after we finish pull-sync (without the user noticing - e.g., preserving user's scroll position). It would be quite difficult to find individual entries which changed and should be re-rendered, at least I can't think of a reasonable way to do that right now.

amulware:
[...]

PR #2355, opened 2018-05-31
Try fetch inaccessible projects

simonrozsival:
Ref #1335 

This PR introduces a sync state which tries to find projects which are 👻 and fetches them using reports API.


simonrozsival:
There is one problem with this PR which I am not really sure how to address - when a project is re-fetched, the UI doesn't update (the TE still shows its project as "(inaccessible project)".

This is a problem with projects/clients/tasks in general - if we pull changes to these entities (e.g., change of the name), the TEs which contain these entities are not updated in the UI (the log and the running TE).

Maybe we just need to redraw everything after pull sync finishes? I will look for an issue and create it if there is nothing related to this problem.

amulware:
Either redraw everything, or just get all time entries of the updated project (are clients rendered in the log?) or tag(!) and push them to the ui as updated?

Please make sure there is an issue so we can fix it for all cases. :)

simonrozsival:
@amulware projects, tasks, clients are, tags are not

Ref https://github.com/toggl/mobileapp/issues/1486

Toggl.Foundation/Sync/States/Pull/TryFetchInaccessibleProjectsState.cs
```diff
@@ -0,0 +1,82 @@
...
+                    projects.GroupBy(project => project.WorkspaceId).OrderBy(group => group.Key))
+                .FirstOrDefaultAsync();
+
+        private IObservable<Unit> refetch(IGrouping<long, IThreadSafeProject> projectsToFetch)
+            =>
+                projectsApi.Search(
+                    workspaceId: projectsToFetch.Key,
+                    projectIds: projectsToFetch.Select(p => p.Id).ToArray())
+                .SingleAsync()
+                .SelectMany(foundProjects => foundProjects)
```
willsb: Identity

Toggl.Foundation/Sync/States/Pull/TryFetchInaccessibleProjectsState.cs
```diff
@@ -0,0 +1,82 @@
...
+                .SelectMany(projects =>
+                    projects.GroupBy(project => project.WorkspaceId).OrderBy(group => group.Key))
+                .FirstOrDefaultAsync();
+
+        private IObservable<Unit> refetch(IGrouping<long, IThreadSafeProject> projectsToFetch)
+            =>
+                projectsApi.Search(
+                    workspaceId: projectsToFetch.Key,
+                    projectIds: projectsToFetch.Select(p => p.Id).ToArray())
+                .SingleAsync()
```
willsb: Why is the call to `Single` even needed?

Toggl.Foundation/Sync/States/Pull/TryFetchInaccessibleProjectsState.cs
```diff
@@ -0,0 +1,82 @@
...
+                .SelectMany(foundProjects => foundProjects)
+                .Select(Project.Clean)
+                .SelectMany(dataSource.Update)
+                .ToList()
+                .SelectMany(updatedProjects =>
+                    projectsToFetch.Where(project =>
+                        !updatedProjects.Select(p => p.Id).Contains(project.Id)))
+                .SelectMany(updateAtValue)
```
willsb: Can you clarify why you need to update the `At` of projects that couldn't be fetched?

willsb: Also, please extract the predicate into a function; it took me a while to understand what was going on here

willsb: Since you are capturing a variable there, something like `.SelectMany(projectsThatWereNotUpdatedByTheApi)` or better

simonrozsival: Updating `At` -> do not try to fetch this project for the next 24 hours.

I will extract the predicate to a method.

Toggl.Foundation/Sync/States/Pull/TryFetchInaccessibleProjectsState.cs
simonrozsival: 👍 

We might add an alias for `SelectMany(Identity)`... `FlattenEnumerable()` or something. But not in this PR...

Toggl.Foundation/Sync/States/Pull/TryFetchInaccessibleProjectsState.cs
simonrozsival: It is not needed. 🔥

PR #2360, opened 2018-05-31
Load time pickers on DateTime Popup with fixed 24hrs mode

### LINKED ISSUE #2336: TimePicker does not show AM/PM or inner circle hours
When Start/Stop/Duration popup opens, the 24 hour mode is being set up from user preferences.
However, no matter what I tried, the outcome is always one of the following:

1. AM/PM mode: The AM/PM buttons are not visible (first screenshot) and can't be clicked.

2. 24 Hours mode: The inner circle of hours is not visible. The hours can be selected (second screenshot), but the labels are not visible. (If `SetIs24HourView` is called within `OnFinishInflate`, the inner circle shows, but this results in case 1. if it was the actual mode).

Basically it seems that change is not good.

What I've tried (to no avail):
- setting up the timer so that this mode is changed every second, the same problem happens
- set up to other mode and then set to the desired mode right after
- RequestLayout, Invalidation - nope.

image

image

daividssilverio:
Closes #2336

It would be better if tested on actual devices.

🦑 whenever


Toggl.Giskard/Fragments/SelectTimeFragment.cs
```diff
@@ -92,6 +90,9 @@ public override View OnCreateView(LayoutInflater inflater, ViewGroup container,
             onTemporalInconsistencyDisposable = ViewModel.TemporalInconsistencyDetected
                                                          .Subscribe(onTemporalInconsistency);
 
+            onStartTimeChangedDisposable = ViewModel.WeakSubscribe<PropertyChangedEventArgs>(nameof(ViewModel.StopTime), onStopTimeChanged);
```
juanlaube: This works, but semantics are wrong, should be `onStopTimeChangedDisposable = ...`.

Toggl.Giskard/Fragments/SelectTimeFragment.cs
```diff
@@ -92,6 +90,9 @@ public override View OnCreateView(LayoutInflater inflater, ViewGroup container,
...
+            onStopTimeChangedDisposable = ViewModel.WeakSubscribe<PropertyChangedEventArgs>(nameof(ViewModel.StartTime), onStartTimeChanged);
```
juanlaube: Idem

Toggl.Giskard/Fragments/SelectTimeFragment.cs
```diff
@@ -215,6 +237,8 @@ protected override void Dispose(bool disposing)
 
             if (disposing == false) return;
 
+            onStartTimeChangedDisposable?.Dispose();
```
kornelijepetak: While you are at it, can you please use a dispose bag here, to remove some boilerplate?

Toggl.Giskard/Views/TogglDroidTimePicker.cs
```diff
@@ -23,9 +23,10 @@ public class TogglDroidTimePicker
     {
         private bool isInitialized;
 
-        public TogglDroidTimePicker(Context context)
+        public TogglDroidTimePicker(Context context, bool is24HoursMode)
             : base(context)
         {
+           Update24HourMode(is24HoursMode);
```
kornelijepetak: Since this function is no longer used from outside, you can just inline the code here. There will be no additional changes to the mode the way we are currently doing things.

PR #2362, opened 2018-06-01
Ghosts tracking

simonrozsival:
Ref #1335 

Squash whenever


simonrozsival:
I updated the base branch to `develop` (this included some cherry-picking and force pushing to this branch).

Toggl.Foundation/Sync/States/Pull/CreateGhostProjectsState.cs
```diff
@@ -34,16 +41,18 @@ public IObservable<ITransition> Start(IFetchObservables fetch)
                 .Distinct(timeEntry => timeEntry.ProjectId)
                 .SelectMany(createGhostForTimeEntryIfProjectIsNotInDatabase)
                 .ToList()
+                .Select(list => list.Count(ghostWasCreated => ghostWasCreated))
+                .Do(analyticsService.TrackCreatingProjectGhosts)
```
juanlaube: Is the idea to also track an event when no ghost projects are created?
If that's the idea, maybe a test can save some questions in the future 😄

simonrozsival: This should track `0` when either the fetched list of time entries is empty or when no ghost is created (most of the times) - I can add a test case for that 👍

juanlaube: If that's the case I don't see the point of adding another tests. :shipit:

Toggl.Foundation/Sync/States/Pull/CreateGhostProjectsState.cs
```diff
@@ -34,16 +41,18 @@ public IObservable<ITransition> Start(IFetchObservables fetch)
                 .Distinct(timeEntry => timeEntry.ProjectId)
                 .SelectMany(createGhostForTimeEntryIfProjectIsNotInDatabase)
                 .ToList()
+                .Select(list => list.Count(ghostWasCreated => ghostWasCreated))
```
willsb: `Identity`

simonrozsival: I am not sure in this case - the name of the boolean clearly says what is counted, so you don't have to think about this too much. Although I can rename `list` into something that makes more sense 🤔 give me a sec..

PR #2364, opened 2018-06-01
Ghost busters

simonrozsival:
Ref #1335 

This PR adds a sync state which removes ghosts which are not really needed anymore - they are (by design) inactive (archived) and if they are not referenced by any time entry, then we don't need to keep them, because they won't be used anymore (until we pull another TE which references it)..

Squash whenever


Toggl.Foundation/Sync/States/Pull/DeleteNonReferencedProjectGhostsState.cs
```diff
@@ -0,0 +1,42 @@
...
+            => projectsDataSource.GetAll(project => project.SyncStatus == SyncStatus.RefetchingNeeded)
+                .SelectMany(CommonFunctions.Identity)
+                .SelectMany(notReferencedByAnyTimeEntryOrNull)
+                .Where(project => project != null)
+                .ToList()
+                .SelectMany(projectsDataSource.DeleteAll)
+                .Select(FinishedDeleting.Transition());
+
+        private IObservable<IThreadSafeProject> notReferencedByAnyTimeEntryOrNull(IThreadSafeProject project)
+            => timeEntriesDataSource.GetAll(timeEntry => timeEntry.ProjectId == project.Id)
```
amulware: Similar comment as the discussion in #2351. This might be REALLY slow if not run in Realm.

amulware: Though to be honest, it might be slow in Realm too, hard to tell. Maybe we shouldn't worry too much, and instead of optimising, just measure and then we can decide more objectively later.

simonrozsival: Maybe this whole state is unnecessary - I am not sure 🤷‍♂️ - maybe we need a third loop which is not connected to pull or push and is only for "rarely needed stuff".

amulware: Maybe we do. We could stick the nightly TE removal in there.. or have a daily loop and a weekly one, and.. :D

simonrozsival: I tested if this condition works without 💥 the app. It does work correctly. It does not crash. I can even put a breakpoint there. Which means that it does not run in the database, but in memory. I am not really sure what to do about it right now, but we shouldn't rush this PR and try poking it on Monday. Maybe @willsb will have some ideas how to test/debug this potential issue.

amulware:
Are we still waiting for a comment by @willsb ?

PR #2365, opened 2018-06-01
Remove model generators

Odrakir:
In order to refactor the models (#1681), we are going to delete the code generators first. (hopefully forever 🤞 )


amulware:
I want all the requested, and at least one more our of { @kornelijepetak, @arturs105, @juanlaube } (my estimate of who has worked with this stuff the most, but other's are welcome to comment, obviously).

If there are concerns about this, speak up!

juanlaube:
I don't like the `.tt` templates, so I'm more than ok with this. That being said, I'm wondering if we don't introduce a bug in the future because we forget to add/remove things in the constructors 🤔

amulware:
Updated original message to mention #1891.

Odrakir:
That's for another PR, @kornelijepetak. I'm on it.

willsb:
I'm pro removing the generators, but this PR *does not* close #1891. It only burns files without proposing what's the solution.

Odrakir:
Are you sure you are ok with this 😀

As I said, improvements come in another PR. I told @amulware this would happen, but we agreed on doing it in separate PRs to simplify reviews.

willsb:
Again, if the solution comes in another PR, how does this close the issue?

Either change the description or propose the changes you are going to do here.

Odrakir:
Oh! Doh...

Sorry, a lot on my plate lately.

PR #2369, opened 2018-06-01
Add firebase performance analytics

### LINKED ISSUE #2299: Integrate with Firebase performance metrics
Investigated in: https://github.com/toggl/mobileapp/issues/1332

Looks like it's easy enough, and would provide us with interesting data right from the start.

Let's do so.

arturs105:
Closes #2299

If i understood the documentation correctly, all i had to do was ~follow the~ add the nuget package and call `Firebase.Core.App.Configure()` (which we already do).

Squash whenever.


simonrozsival:
Can we somehow test that we are getting some information (e.g., startup times) to the firebase console on this branch? An ad-hoc build maybe?

arturs105:
I will try to build a version with analytics enabled and see if anything appears in Firebase.

arturs105:
I just tested out the app with analytics enabled and even created a custom trace, but the documentation states that those will appear in about 12 hours, so tomorrow we'll see if this is working.

Toggl.Daneel/Toggl.Daneel.csproj
```diff
@@ -29,7 +29,7 @@
     <MtouchLink>Full</MtouchLink>
     <MtouchArch>x86_64</MtouchArch>
     <MtouchHttpClientHandler>NSUrlSessionHandler</MtouchHttpClientHandler>
-    <MtouchExtraArgs>--linkskip Toggl.Daneel</MtouchExtraArgs>
+    <MtouchExtraArgs>--linkskip Toggl.Daneel --registrar:static</MtouchExtraArgs>
```
willsb: y tho? This kind of registrar already is the default for release builds; and shouldn't be needed for simulator builds

arturs105: Otherwise the build failed with 7 errors about some duplicate symbol. Other issues/PRs mention the same attribute and the same 7 errors.

arturs105:
The performance data is now available in Firebase, so this is definitely working.

arturs105:
See #2403

Issue #2361, opened 2018-06-01
Google sign up not working


ApocalypticFish:
On the sign up page the option under the standard email sign up shows "Sign in with Google".
In the iOS app this shows "Sign up with Google"

Not sure if it's related to this but whenever I try with my Gmail address with no Toggl account attached to sign up it displays "Sorry, we're having trouble signing you up."

juanlaube:
@pe0ny the design copy is "Sign in with Google", that's a little weird for me too, should we match iOS?

amulware:
@juanlaube Yes, let's match iOS. This sounds like an oversight.

I also changed the title and tags, since there is more than just a copy issue.

lizaPlotnikova:
Copy's def. an oversight – sorry guys, now updated in designs

PR #2370, opened 2018-06-01
Fix sign up with Google button title

juanlaube:
Ref #2361 

🦑 whenever


juanlaube:
Google's own strings are localized, we don't want things like "Iniciar sesión con Google" 😄

Issue #2102, opened 2018-05-10
Fix duplicate running time entries in database


amulware:
It happens. It crashes the app. Uh oh.

Here are several stack traces that I _think_ are related.

https://appcenter.ms/orgs/Toggl/apps/Toggl-iOS/crashes/groups/6047277ea69dfd9257b04797321b6c9f68abb357?appVersion=1.5
```
System.InvalidOperationException: Sequence contains more than one matching element

RealmResultsVisitor.VisitMethodCall (System.Linq.Expressions.MethodCallExpression node)
MethodCallExpression.Accept (System.Linq.Expressions.ExpressionVisitor visitor)
ExpressionVisitor.Visit (System.Linq.Expressions.Expression node)
RealmResultsProvider.Execute (System.Linq.Expressions.Expression expression)
RealmResultsProvider.Execute[T] (System.Linq.Expressions.Expression expression)
Queryable.SingleOrDefault[TSource] (System.Linq.IQueryable`1[T] source, System.Linq.Expressions.Expression`1[TDelegate] predicate)
RealmAdapter`2[TRealmEntity,TModel].resolvePotentialRivals (Realms.Realm realm, TRealmEntity entity, Toggl.PrimeRadiant.IRivalsResolver`1[TModel] resolver, System.Collections.Generic.List`1[T] results)
RealmAdapter`2[TRealmEntity,TModel].BatchUpdate (System.Collections.Generic.IEnumerable`1[T] batch, System.Func`3[T1,T2,TResult] conflictResolution, Toggl.PrimeRadiant.IRivalsResolver`1[TModel] rivalsResolver)
Repository`1+<>c__DisplayClass2_0[TModel].<BatchUpdate>b__0 ()
BaseStorage`1+<>c__DisplayClass8_0`1[TModel,T].<CreateObservable>b__0 (System.IObserver`1[T] observer)
```

https://appcenter.ms/orgs/Toggl/apps/Toggl-iOS/crashes/groups/0c5756bf4d4062b89886e54e0aeb2476033088ee?appVersion=1.5
```
System.InvalidOperationException: Sequence contains more than one element

Enumerable.SingleOrDefault[TSource] (System.Collections.Generic.IEnumerable`1[T] source)
TimeEntriesDataSource+<>c.<Stop>b__29_1 (System.Collections.Generic.IEnumerable`1[T] timeEntries)
Select`2+_[TSource,TResult].OnNext (TSource value)
ExceptionDispatchInfo.Throw ()
[...]

PR #2378, opened 2018-06-04
Simplify rivals resolution

simonrozsival:
Ref #2102 

~On-hold: #2365 must be merged first.~

We had a problem with copying tags - it was "fixed" in #1729 - I believe this PR would fix the same problem, but in a better way without the need to compare equivalencies to update the object only if it will change.

I think this might in certain situation lead to a situation when _neither_ of the two rivals is overridden and both are running after the rivals resolution. I am not really sure how it would happen, but I would like to eliminate that possibility anyway, when it simplifies the code.


amulware:
Itsn't this basically reverting the other PR?

simonrozsival:
Just the `RealmAdapter` part.

PR #2380, opened 2018-06-04
Remove keyboard icon from timepicker

### LINKED ISSUE #2263: Hide manual editing on time widget
Pertains to API 26+:
Timepicker on Oreo actually has a button in lower left part that shows the keyboard and another way to enter the time. Since in general, android timepicker is bad and we are hacking it to add calendar icon on top of it, I think it'd be wise to remove the possibility to do this manually with keyboard because, the `Set time` label then overlaps our calendar icon.

@pe0ny please advise.

I hope there's a way to remove it.

image

image

kornelijepetak:
Closes #2263 

On Oreo, the keyboard icon was introduced to allow manual entry of time. We don't want it and this PR removes it.


kornelijepetak:
DroidDatePicker has no such a thing... At least not in the mode we are using it (Calendar view).

PR #2384, opened 2018-06-04
Remove duplicated height convertion to dips

### LINKED ISSUE #2376: Custom colour picker UI: too much padding at the bottom
Custom colour picker popup has too much padding at the bottom.

screenshot_20180603-140236

Check Zeplin for correct UI: https://zpl.io/VkYkd9W

Because of this #2373 I can't test regular colour picker, so please make sure it looks ok as well: https://zpl.io/bLBRnlm

daividssilverio:
Closes #2376

`Dialog.Window.SetDefaultDialogLayout` already does the conversion, so there is no need to convert the height to dips anymore.

🦑 whenever


Toggl.Giskard/Fragments/SelectColorFragment.cs
```diff
@@ -16,14 +16,14 @@ namespace Toggl.Giskard.Fragments
     [MvxDialogFragmentPresentation(AddToBackStack = true)]
     public sealed class SelectColorFragment : MvxDialogFragment<SelectColorViewModel>
     {
-        private int customColorEnabledHeight;
-        private int customColorDisabledHeight;
+        private readonly int customColorEnabledHeight = 425;
```
kornelijepetak: `const` is better in this case.

PR #2385, opened 2018-06-04
Fix workspace feature enabling logic

### LINKED ISSUE #2373: New project view: Custom colours should be available only for paid plans
Currently, custom colours are also available for Free WS, though they shouldn't be.

Steps to reproduce:
1. Sign up (by default you're on free plan)
2. Create a new project and select colour -> Custom colour picker is displayed and allows the user to select a custom colour.

daividssilverio:
Closes #2373

🦑 whenever


simonrozsival:
I think it was a bad thing to introduce this method 🙈 - as William suggested, this could be a nice extension method for `IWorkspaceFeatureCollection` and this PR might be a good opportunity to change it.

PR #2387, opened 2018-06-05
Fix Edit View running TE stop/current time

### LINKED ISSUE #2374: Date-time-duration popup: Content under the "Stop" tab is a bit off
When editing running TE, content under the "Stop" tab is a bit off.

1. Open date-time-duration popup when TE is running
2. Open "Stop" tab
3. You can see that AM / PM font sizes and positions are a bit off

### LINKED ISSUE #2386: Time popup shows AM/PM even when in 24 hour mode
`20:17 PM`? Who dat!?

image

kornelijepetak:
Closes #2374 
Closes #2386 

Since AM/PM labels are not part of the time label in popup, the AM/PM is stripped from the label.

🦑 whenever.


Toggl.Foundation.MvvmCross/Converters/StripAmPmValueConverter.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+using System.Globalization;
+using MvvmCross.Platform.Converters;
+using Toggl.Multivac;
+using static System.Math;
```
daividssilverio: 🔥

Toggl.Foundation.MvvmCross/Converters/StripAmPmValueConverter.cs
```diff
@@ -0,0 +1,15 @@
...
+
+namespace Toggl.Foundation.MvvmCross.Converters
+{
+    [Preserve(AllMembers = true)]
+    public sealed class StripAmPmValueConverter : MvxValueConverter<string, string>
+    {
+        protected override string Convert(string value, Type targetType, object parameter, CultureInfo culture) 
+            => value.Replace("AM", "").Replace("PM", "").Trim();
```
daividssilverio: I'm guessing it would be easier to tweak with the Android UI if you create a binding to, instead of setting the emptying the AM/PM labels to empty strings or not, just set the visibility of the TextViews to Visible or Invisible, because it would still take the same space, making the Time label stay still on both 24hrs and 12hrs modes.

daividssilverio:
Right now, it looks like this when you change the time format, sync and go back to the edit popup:
It looks alright after you change tabs, or close and open the popup again.

Shouldn't the time label (12:46) be center aligned, like the header on the start time tab?

Toggl.Foundation.MvvmCross/Converters/StripAmPmValueConverter.cs
kornelijepetak: I have no idea what exactly is your point. Could you please describe a bit better?
What is the problem with this converter?

kornelijepetak:
@daividssilverio in regards to centered time, according to designs, no. It is right aligned. However, if @lizaPlotnikova or @pe0ny think it should be centered to be consistent with the stopped time entry's time picker, it's not difficult to change.

kornelijepetak:
@daividssilverio please give reproduction steps for the overlapping labels bug (shown in your screenshots). How did you run the sync while the popup is open? I am unable to reproduce the problem you've described.

Toggl.Foundation.MvvmCross/Converters/StripAmPmValueConverter.cs
daividssilverio: Sorry I wasn't clear. The only a problem with this adapter is that since the TextViews where they were used have the width set to `wrap_content`, they will take variable space in the layout, pushing the TE time to the left when 24hrs mode is disabled.

But, if the layout is to be aligned to the left, there's no problem here.

Toggl.Giskard/Resources/layout/SelectStopDateTimeEditorPage.axml
```diff
@@ -115,39 +116,45 @@
...
-            android:textSize="16sp"
-            android:layout_alignParentRight="true"
-            android:fontFamily="sans-serif-medium"
-            android:layout_alignParentBottom="true"
-            android:layout_marginBottom="20dp"
+        <LinearLayout
+            android:id="@+id/SelectDateTimeAmPmLabels"
+            local:MvxBind="Visibility Visibility(!TimeFormat.IsTwentyFourHoursFormat)"
+            android:orientation="vertical"
+            android:gravity="right|bottom"
```
daividssilverio: Since the width of the LinearLayout is set to `wrap_content`, you don't need the `right` gravity here, `bottom` is enough.

Toggl.Giskard/Resources/layout/SelectStopDateTimeEditorPage.axml
```diff
@@ -115,39 +116,45 @@
...
+                android:layout_width="wrap_content"
+                android:layout_height="wrap_content" />
+            <TextView
+                android:id="@+id/SelectDateTimeRunningTimeEntryPM"
+                local:MvxBind="Visibility Visibility(!IsRunningTimeEntryAM)"
+                android:text="PM"
+                android:textColor="@android:color/white"
+                android:textSize="16sp"
+                android:fontFamily="sans-serif-medium"
+                android:layout_marginBottom="20dp"
```
daividssilverio: It looks like we need these 20dp for the SelectDateTimeRunningTimeEntryAM TextView as well. Or add to the container itself.

daividssilverio:
It looks like this on AM.
screen shot 2018-06-07 at 3 58 11 pm

Toggl.Giskard/Resources/layout/SelectStopDateTimeEditorPage.axml
kornelijepetak: Now both labels are visible and their transparency is varied.

PR #2391, opened 2018-06-05
Fix popup height when time entry is stopped.

### LINKED ISSUE #2375: Date-time-duration popup: after stopping an entry, picker is still a bit hidden
1. Open date-time-duration popup when TE is running
2. Open "Stop" tab and tap "Stop timer" -> You can see that the picker is still a bit hidden. After changing tabs, picker becomes available.

kornelijepetak:
Closes #2375 

🦑  whenever.


Toggl.Giskard/Fragments/SelectTimeFragment.cs
```diff
@@ -82,7 +82,12 @@ public override View OnCreateView(LayoutInflater inflater, ViewGroup container,
             pager.Adapter = new SelectTimePagerAdapter(BindingContext);
             tabLayout.AddOnTabSelectedListener(this);
 
-            subscribeAndAddToDisposableBag(nameof(ViewModel.IsCalendarView), onIsCalendarViewChanged);
+            var isCalendarViewDisposable = ViewModel.IsCalendarViewObservable
```
daividssilverio: Why not just use `.DisposedBy(disposableBag)`?

PR #2392, opened 2018-06-05
Skip rows which are out of range in the reports table view source

### LINKED ISSUE #2105: Index out of range exception in reports table view source
One of the more common crashes in 1.5.

https://appcenter.ms/orgs/Toggl/apps/Toggl-iOS/crashes/groups/6bebe5c5d134975ac7e16920ecc6ad6bd65ea3ba?appVersion=1.5
```
System.ArgumentOutOfRangeException: Index was out of range. Must be non-negative and less than the size of the collection. Parameter name: index
ThrowHelper.ThrowArgumentOutOfRangeException (System.ExceptionArgument argument, System.ExceptionResource resource)
ThrowHelper.ThrowArgumentOutOfRangeException ()
List`1[T].get_Item (System.Int32 index)
IList.get_Item (System.Int32 index) in <0f7865fc4fbb4fddaab6f86c915ebae0#6388cdb77cc7e11608c230aa83e65725>:0
MvxEnumerableExtensions.ElementAt (System.Collections.IEnumerable items, System.Int32 position)
MvxTableViewSource.GetItemAt (Foundation.NSIndexPath indexPath)
ReportsTableViewSource.GetCell (UIKit.UITableView tableView, Foundation.NSIndexPath indexPath)
```

simonrozsival:
Closes #2105 

I suppose the problem occurs because of a race condition - the list of project summaries is changed while the legend is being rendered and the number of project summaries decreases.

This PR makes sure the app does not crash when it wants to render rows which aren't available.

Squash whenever


simonrozsival:
You are right. I am not really sure why I thought that it would work - and also why it worked when I tested it before - I did something wrong... I will try again tomorrow :)

PR #2396, opened 2018-06-05
Refactor client model

Odrakir:
Part of #1681 

Just for `Client` for now.

This is just the next logical step after removing the generators. I basically put it all in one file, this one is small so I didn't feel the need to split it, we can see about that with more complex models though.

I also didn't see the need of using the `with...new` solution here, basically because it wouldn't be used.

I wish I could remove the dependency from thread safe models to database models, I was looking at it for a while, but that's far too complex for my current knowledge of the code base.


Toggl.Foundation/Models/Client.cs
```diff
@@ -1,99 +1,73 @@
...
 using Toggl.PrimeRadiant;
+using Toggl.PrimeRadiant.Models;
 
 namespace Toggl.Foundation.Models
 {
-    internal partial class Client
+    internal class Client : IThreadSafeClient
     {
-        internal sealed class Builder
+        public long Id { get; }
```
juanlaube: I prefer this style, but in other PRs was suggested to have an empty line between fields/properties.

Toggl.Foundation/Models/Client.cs
```diff
@@ -1,99 +1,73 @@
...
-            public bool IsDeleted { get; private set; }
+        private Client(IClient entity, SyncStatus syncStatus, string lastSyncErrorMessage, bool isDeleted = false)
+            : this(entity.Id, entity.WorkspaceId, entity.Name, entity.At, syncStatus, lastSyncErrorMessage, isDeleted, entity.ServerDeletedAt)
+        { }
 
-            private Builder(long id)
-            {
-                Id = id;
-            }
+        private Exception parameterException(string name)
```
juanlaube: This can be moved to an extension.
Also, I'm not sure `InvalidOperationException` is the best usage here, I think it should be `ArgumentException` 🤔

Toggl.Foundation/Models/Client.cs
```diff
@@ -1,99 +1,73 @@
...
-            }
+        public Client(long id, long workspaceId, string name, DateTimeOffset at, SyncStatus syncStatus, string lastSyncErrorMessage = "", bool isDeleted = false, DateTimeOffset? serverDeletedAt = null)
+        {
 
-            public Builder SetName(string name)
-            {
-                Name = name;
-                return this;
-            }
+            if (string.IsNullOrEmpty(Name))
```
juanlaube: We already have `Ensure.Argument.IsNotNull` in Multivac, maybe another extension called `Ensure.Argument.IsNotNullOrEmpty`?

Toggl.Foundation/Models/Client.cs
```diff
@@ -1,99 +1,73 @@
...
+                throw parameterException("name");
 
-            public Builder SetAt(DateTimeOffset at)
-            {
-                At = at;
-                return this;
-            }
+            if (WorkspaceId == 0)
```
juanlaube: Same idea.

Toggl.Foundation/Models/Client.cs
```diff
@@ -1,99 +1,73 @@
...
+                throw parameterException("workspace id");
 
-            public Builder SetServerDeletedAt(DateTimeOffset? serverDeleteAt)
-            {
-                ServerDeletedAt = serverDeleteAt;
-                return this;
-            }
+            if (At == null)
```
juanlaube: Idem.

Toggl.Foundation/Models/Client.cs
simonrozsival: Here you are validating the object's property `Name` which wasn't assigned yet, so this would always fail - the same goes for the `WorkspaceId` and `At`.

Toggl.Foundation/Models/Client.cs
Odrakir: I kept a new line for most of my time working with this PR, but in the end, I find it much more clear this way. I'll wait to see what the others say about it.

amulware: I honestly hate having new lines between all fields/properties. Such a waste of space. 🙊

Toggl.Foundation/Models/Client.cs
```diff
@@ -1,99 +1,73 @@
...
+        public Client(long id, long workspaceId, string name, DateTimeOffset at, SyncStatus syncStatus, string lastSyncErrorMessage = "", bool isDeleted = false, DateTimeOffset? serverDeletedAt = null)
+        {
 
-            public Builder SetName(string name)
-            {
-                Name = name;
-                return this;
-            }
+            if (string.IsNullOrEmpty(Name))
+                throw parameterException("name");
```
amulware: Use `nameof()`

Toggl.Foundation/Models/Client.cs
```diff
@@ -1,99 +1,73 @@
...
-            public string Name { get; private set; }
-
-            public SyncStatus SyncStatus { get; private set; }
-
-            public long? WorkspaceId { get; private set; }
-
-            public DateTimeOffset? At { get; private set; }
-
-            public DateTimeOffset? ServerDeletedAt { get; private set; }
+            Workspace = entity.Workspace == null ? null : Models.Workspace.From(entity.Workspace);
```
amulware: This constructor does work, which is not clear from the signature. I think I much rather have a static method with a name for that.

Odrakir: I'm not sure I understand.

amulware: this constructor doesn't just create a client from the given parameters, it also creates a new workspace (for the thread-safety thing). I think that's too much work to do in a constructor, which by design doesn't have a name explaining what it does.

Odrakir: I see... But you just mean to move that line to a static method, right?

What I don't understand is if a `Client` can (should) exist without the `Wokspace` property... Does it make sense? Shouldn't we verify as much as we can from the type point of view?

amulware: Clients cannot exist without workspace. We should probably check that, yes. And I mean not only that line of code, but the entire constructor should not exist like this. Instead, it should be a static method that calls a 'dumb' constructor.

Odrakir:
This is not it, but I need to discuss this with you, so...

In theory (that's what I've been told), a `Client` should have all its properties (including `Workspace`). In practice, right now, `Clients` created via `Clean`, `Dirty` and `Unsyncable` don't set the `Workspace` property. And neither do the ones created with the Builder (which I changed to a normal constructor), and also don't set other properties (see `ClientDataSource`)

Odrakir:
BTW, tests pass, but the model is unsafe in my opinion.

simonrozsival:
The usage of these methods in tests is a bit hacky - it is an easy way of converting the different interfaces between each other. The mocks used in the tests didn't specify some of the properties simply because we don't need them in that test.

I would say we should keep the validation. I am not really sure if it would make any sense right now to not have the validation. If we used hacks in the tests and it is the problem now, we should change those tests.

Odrakir:
If it was only the tests I would have already suggested that, but I see those static methods are used in real code... I have no idea what they are for, though...

simonrozsival:
OK. These methods are used just to take the entity and create a new one which has the same properties but the `SyncState` is changed - there is one static method for every state (Clean - `InSync`, Dirty - `NeedsSycning`, Unsyncable - `SyncingFailed`). If the implementation of these static methods is incorrect, we might want to rewrite them to use the new constructor.

I need to look into the code to remind myself how exactly it works and if I can see what the problem is and what the proper solution could be.

Issue #1336, opened 2018-01-25
Fetch all workspaces on every pull-sync and detect missing/added ones


amulware:
There is really no issue with pulling all workspaces (almost always 1-2, very rarely more than 5) of a user on every pull sync.

When persisting the full list we should compare it to the list of local workspaces, and where they differ resolve conflicts accordingly, IN THIS ORDER:

If we lose access (don't confuse this with the existence of a local WS that was not synced yet!):
- [ ] remove all synced TEs from the workspace
- [ ] remove all synced projects/clients/tags that are not referred to by any TE (clients are referred through the projects and their check is maybe a special case)
- [ ] #1337 mark all remaining (i.e. unsynced) entities especially, so that they cannot show up in auto-completion, suggestions. TEs like that should not be able to be continued either

If we gain access to a new WS:
- [ ] add that workspace to the DB
- [ ] make sure that if we had access to this WS before, and there is unsynced data, all that can sync again (we may need to undo whatever we did to them in #1377), and that data is merged correctly when pulling
- [ ] then reset all ?since dates to null
- [ ] go to the top of the pull sync (which will automatically fetch all data for those workspaces, and also for other workspaces, but that's something to optimise later if we need to) - this prevents the stuff from below from running until after the full pull

If the user has no workspaces left:
- [ ] check this by looking if the database has any non-ghost (#1337) workspaces, it does not matter if they are synced or not
- [ ] throw an exception and halt syncing if this happens - the exception will cause a prompt making the user create or retry sync (and the retry will cause another exception if nothing has changed; if it completes however without throwing that exception, the prompt can be closed)

Further:
- [ ] on app start/activation(every X minutes? - depending on the cost of this operation) all 'invalid' entities that are no longer referred to by any time entries can be removed (alternatively, when TEs are edited/deleted, but that might be more complicated?)

See also: #1334

simonrozsival:
Do I understand it correctly that we are gonna keep the workspaces to which the user loses access (which have some unsynced data)  in a somewhat similar state to "ghost projects" - it will be a "ghost workspace".

I am not sure how will we mark the unsynced time entries/projects/clients which belong to this workspace - will we make them "unsynced ghosts"? or will we add another property to the database which will that they are inaccessible (something that I "solved" for ghost projects by setting `Active` to `false`).

Gaining access to a workspace back means that we fetch a workspace which is locally marked as inaccessible - now I guess that we should first update the workspace data in the DB and only then start making the related entities accessible, right?

[...]

PR #2398, opened 2018-06-05
Detect if no workspaces exist

### LINKED ISSUE #2331: Detect if no workspaces exist after (workspace) pull
I'm not sure if this should cause sync to be interrupted at that point, but it definitely should cause a special event to be fired off that will cause #1342/#1343 to show.

Reading those issues, yes, we probably want to interrupt sync (since the user can retry syncing again on the popup).

sofokli1:
Closes #2331

🦑 whenever


Toggl.Foundation/Sync/States/Pull/NoWorkspaceExceptionsCatchingPersistState.cs
```diff
@@ -0,0 +1,40 @@
...
+using System.Linq;
+using System.Reactive.Linq;
+using Toggl.Foundation.Exceptions;
+using Toggl.Foundation.Sync.States.Pull;
+using Toggl.Multivac;
+using Toggl.Multivac.Models;
+
+namespace Toggl.Foundation.Sync.States
+{
+    internal sealed class NoWorkspaceExceptionsCatchingPersistState : IPersistState
```
simonrozsival: I am a bit confused by the name of this class - it doesn't catch the exception, it throws it.

Toggl.Foundation/Sync/States/Pull/NoWorkspaceExceptionsCatchingPersistState.cs
```diff
@@ -0,0 +1,40 @@
...
+    {
+        private readonly IPersistState internalState;
+
+        public StateResult<IFetchObservables> FinishedPersisting { get; } = new StateResult<IFetchObservables>();
+
+        public StateResult<Exception> Failed { get; } = new StateResult<Exception>();
```
simonrozsival: I don't think this transition is needed - we don't want to catch database exceptions.

Toggl.Foundation/Sync/States/Pull/NoWorkspaceExceptionsThrowingPersistState.cs
```diff
@@ -0,0 +1,34 @@
...
+
+        public IObservable<ITransition> Start(IFetchObservables fetch)
+            => fetch.GetList<IWorkspace>()
+                .SelectMany(workspaces => workspaces.Any()
+                    ? handlePresenceOfWorkspaces(fetch)
+                    : Observable.Throw<ITransition>(new NoWorkspaceException()));
+
+        private IObservable<ITransition> handlePresenceOfWorkspaces(IFetchObservables fetch)
+            => internalState.Start(fetch)
+                .Select(_ => FinishedPersisting.Transition(fetch));
```
simonrozsival: If you add `using Toggl.Multivac.Extensions;` you can use our new extension method to do `Select(FinishedPrsisting.Transition(fetch));` to make this even nicer.

Toggl.Foundation/DataSources/TogglDataSource.cs
```diff
@@ -132,8 +132,9 @@ private IObservable<bool> hasUnsyncedData<TModel>(IRepository<TModel> repository
 
         private void onSyncError(Exception exception)
         {
-            if (apiErrorHandlingService.TryHandleDeprecationError(exception)
-                || apiErrorHandlingService.TryHandleUnauthorizedError(exception))
+            if (errorHandlingService.TryHandleDeprecationError(exception)
+                || errorHandlingService.TryHandleUnauthorizedError(exception)
+                || errorHandlingService.TryHandleNoWorkspaceError(exception))
             {
                 stopSyncingOnSignal();
```
simonrozsival: We must not forget to call `DataSource.StartSyncing` when the user creates the workspace - because otherwise the app will stop syncing automatically until the next restart of the app - not a concern for this PR though.

Toggl.Foundation/Sync/States/Pull/NoWorkspaceExceptionsThrowingPersistState.cs
```diff
@@ -0,0 +1,35 @@
...
+        public NoWorkspaceExceptionsThrowingPersistState(IPersistState internalState)
+        {
+            Ensure.Argument.IsNotNull(internalState, nameof(internalState));
+
+            this.internalState = internalState;
+        }
+
+        public IObservable<ITransition> Start(IFetchObservables fetch)
+            => fetch.GetList<IWorkspace>()
+                .SelectMany(workspaces => workspaces.Any()
```
amulware: I'm not convinced we want to use the return of the remote workspaces to be the trigger here. There could be a local unsynced workspace, which would allow the user to track time just fine, even if backend does not return any workspaces.

I think this should rather check the database for the existence of valid workspaces. Am I missing something @simonrozsival?

amulware:
Apart from my comment above, this is probably fine. I just want to make sure we get this right and don't work ourselves into a corner.

Toggl.Foundation/Sync/States/Pull/NoWorkspaceExceptionsThrowingPersistState.cs
simonrozsival: I think we will implement/extend this logic when we implement #1336 - should it be part of this PR as well?

sofokli1:
@amulware Are you still thinking about this? if not, give us a ✅

amulware:
I'm honestly not sure if the current implementation will add anything. The only way to make it work is if the upcoming 'no workspace' views will force a PUSH sync before trying to pull again. If they just pull again, it'll fail at exactly the same spot, and the user will be stuck.

However, I'm not seeing the push happening easily, especially since the no-workspace situation would cause the sync machine to freeze, and a newly created one always starts a pull sync, does it not? (again pinging @simonrozsival who worked on a lot of that and is currently thinking about the upcoming logic as well)

amulware:
My questions are not answered yet, but I'm ok merging this for now to get it out of the way. I'm very sure we'll have to change this again down the line to make things work properly.

PR #2399, opened 2018-06-05
Reports Calendar: Jump calendar page to selected view when selecting shortcut

### LINKED ISSUE #1943: Reports calendar: jump calendar into selected view
> When I've navigated to another month in the past, and then click "This week", I think it would be helpful if the selection could be brought into view (i.e. jump calendar to include "this week"... maybe the rule is "Jump to the month that contains the first date of the selection")

It makes a lot of sense to me. 👆Let's make it so, that the calendar jumps into the month that contains the first date of the selection.

daividssilverio:
Closes #1943

squash whenever


Toggl.Giskard/Fragments/ReportsCalendarFragment.cs
```diff
@@ -26,7 +28,8 @@ public sealed class ReportsCalendarFragment : MvxFragment<ReportsCalendarViewMod
         private int rowHeight;
         private ViewPager pager;
         private int currentRowCount;
-        private IDisposable disposable;
+        private IDisposable onRowsInCurrentMonthChangedDisposable;
```
willsb: Use a CompositeDisposable instead. Also, dispose it properly while at it :)

Issue #1915, opened 2018-04-18
Wheel metrics


amulware:
I'd like to get some data on how our users are editing times and durations.

For that I'd like to add an event `ClosedTimeView` (or similar) with the following properties:

- changed start date with barrel
- changed start time with barrel
- changed end date with barrel
- changed end time with barrel
- changed start time using wheel
- changed end time using wheel
- changed start+end using wheel (dragged the filled area)
- changed duration using numpad
- entry was running when opening view
- stopped running entry
- navigation origin (start/edit view)
- result (save/cancel)

I figure we can have a little container with a lot of bools for this that initialises to all false when the view is opened, and when the user changes anything the respective values get set to true. In the end, this is turned into a properties dictionary and pushed as an event. Nothing more complicated (like comparing data on open and data on closed) is needed I think. This should give us a good idea of how the view is used.

simonrozsival:
What about splitting this into several events?
- `ClosedTimeViewBarrel`
    - changed start date with barrel
    - changed start time with barrel
    - changed end date with barrel
    - changed end time with barrel
- `ClosedTimeViewWheel`
    - changed start time using wheel
    - changed end time using wheel
    - changed start+end using wheel (dragged the filled area)
- `ClosedTimeViewNumpad`
    - changed duration using numpad
- `ClosedTimeView`
    - entry was running when opening view
    - stopped running entry
    - navigation origin (start/edit view)
    - result (save/cancel)

amulware:
When exactly are each of these fired?

simonrozsival:
When the view is closed if the barrel/wheel/numpad was used at all.

Two benefits:
- we would see which components are (not) used - some events might not be tracked at all (👀 _numpad_)
- on Android there is no wheel/components - easier code sharing

amulware:
I see, interesting. Though, we can see all the same stuff if it's a single even too, right? Or... you mean this helps us group related events a bit easier?

On the other hand it makes it harder to find out how users change start times specifically for example.

simonrozsival:
The first reason - I am lazy and too many params is more work that more events (if we merge #2401).

Second I think splitting this will be easier to analyze in AppCenter.

You want to track how users combine these components? If they use the wheel for the star time and the barrel for the end time? 🤔

amulware:
I was more thinking: If I want to know if users use wheel or barrel to edit start time, I have to check two events with this proposal.

[...]

PR #2401, opened 2018-06-06
Refactor analytics service

### LINKED ISSUE #1532: Track OfflineException
Let's track how often the sync man detects being offline, not as an exception, but as a special event. We might decide we don't care about this in the future, but lets at least see how often this happens and if the data is of any use.

simonrozsival:
This PR simplifies the way how to add and track events using the analytics service - and add a nice extension method for tracking events in observables (instead of using the ugly `Do(_ => analyticsService.Track*(...))`).

In the future as we add more events, we might have to add more `IAnalyticsEvent<T...>` interfaces and the classes, but that is easy and reusable (except in #1915 we will have to be a bit more creative, because the number of params is too big to define all the generics for an event with an abnormal number of params (threre are mostly 0 or 1).

Most of the changes in this PR are updated tests - I recommend going commit by commit to see what was added and changed.

As an example of what it now takes to add an event this PR also closes #1532 (commit 
0a58248)

Squash whenever


Toggl.Foundation/Analytics/BaseAnalyticsService.cs
```diff
@@ -1,195 +1,79 @@
...
+            CurrentPage = new AnalyticsEvent<Type>(this, nameof(CurrentPage), "CurrentPage");
+            TimeEntryStarted = new AnalyticsEvent<TimeEntryStartOrigin>(this, nameof(TimeEntryStarted), "Origin");
+            DeleteTimeEntry = new AnalyticsEvent(this, nameof(DeleteTimeEntry));
+            ApplicationShortcut = new AnalyticsEvent<string>(this, nameof(ApplicationShortcut), "ApplicationShortcutType");
+            EditEntrySelectProject = new AnalyticsEvent(this, nameof(EditEntrySelectProject));
+            EditEntrySelectTag = new AnalyticsEvent(this, nameof(EditEntrySelectTag));
+            StartEntrySelectProject = new AnalyticsEvent<ProjectTagSuggestionSource>(this, nameof(StartEntrySelectProject), "Source");
+            StartEntrySelectTag = new AnalyticsEvent<ProjectTagSuggestionSource>(this, nameof(StartEntrySelectTag), "Source");
+            ReportsSuccess = new AnalyticsEvent<ReportsSource, int, int, double>(this, nameof(ReportsSuccess), "Source", "TotalDays", "ProjectsNotSynced", "LoadingTime");
+            ReportsFailure = new AnalyticsEvent<ReportsSource, int, double>(this, nameof(ReportsFailure), "Source", "TotalDays", "LoadingTime");
```
amulware: I think the setup here, with strings and all that is more prone to error than the previous approach. Is there anything that can be done about that?

Toggl.Foundation/Analytics/IAnalyticsService.cs
```diff
@@ -1,42 +1,54 @@
 ﻿using System;
+using System.Collections.Generic;
 
 namespace Toggl.Foundation.Analytics
 {
     public interface IAnalyticsService
```
amulware: I feel a bit weird about still having this interface, now that it only has properties, no methods. Nobody would ever implement this again in the current state, so the base class seems like it is enough. We really only have this for testing now, right? (Just like before I suppose.)

simonrozsival: Yes, we have this (just) for testing.

simonrozsival: We could of course create a TestingAnalyticsService based on the base class - that would be easy and might be enough.

I am not sure how exactly will we work with DI if we remove the interface - but it should work with the abstract base class the same way.

simonrozsival:
Please note that we can easily remove the latest commit if you don't like it - I am quite sure that @amulware won't like having this much reflection in our own code.

The good thing about this is that it runs this code when the app is started - if this code fails, the app will crash during startup => we would never release it to end users (unless we don't even try running it before we ship it...).

Toggl.Foundation/Analytics/IAnalyticsService.cs
simonrozsival: I am not sure if removing the interface in this PR wouldn't be too much - it is already quite big. We can create a follow-up issue to remove it and update all of the tests.

simonrozsival:
> are there other benefits to this approach compared to the one before?

I would say that the benefit is a lot less code in the `BaseAnalyticsService` class (either with or without `dc02c5f`) which makes adding more and more events simpler while keeping it type-safe.

> I think the setup here, with strings and all that is more prone to error than the previous approach. Is there anything that can be done about that?

I am not exactly sure how you mean it - the strings are just the names of the parameters which will be used for creating the dictionaries of parameters. We had each of these strings in a constant which were weird. If we want to do something about this, we might add an enum with different parameter names - that is easy to convert into a string + is somewhat type safe.

Toggl.Foundation/Analytics/IAnalyticsService.cs
amulware: I'm not sure what our prefered policy is for that case. Whether we rather duplicate in the real code with an unnecessary interface or duplicate in the tests with a mock implementation.

One key problem: Without the interface, we can't create substitutes for tests that need the service to run, but where we don't actually test the service either. Maybe that's a downside we don't like.

Toggl.Foundation/Analytics/AnalyticsEventAttributeInitializer.cs
```diff
@@ -0,0 +1,60 @@
...
+            var eventPropertiesInfo = GetType().GetProperties().Where(prop => Attribute.IsDefined(prop, typeof(AnalyticsEventAttribute)));
+            foreach (var eventPropertyInfo in eventPropertiesInfo)
+            {
+                var attribute = eventPropertyInfo.GetCustomAttribute<AnalyticsEventAttribute>();
+                var type = eventPropertyInfo.PropertyType;
+                var name = eventPropertyInfo.Name;
+
+                if (!type.IsGenericType)
+                {
+                    var property = new AnalyticsEvent(analyticsService, name) as object;
```
amulware: Is the cast needed?

Toggl.Foundation/Analytics/AnalyticsEventAttributeInitializer.cs
```diff
@@ -0,0 +1,60 @@
...
+        };
+
+        private Type getGenericType(Type[] genericTypeArguments)
+        {
+            if (genericTypes.TryGetValue(genericTypeArguments.Length, out var type))
+            {
+                return type.MakeGenericType(genericTypeArguments);
+            }
+
+            throw new ArgumentOutOfRangeException($"There is not analytics even type for {genericTypeArguments.Length} parameters.");
```
amulware: Typos

amulware:
I'm not sure what I think about the reflection. Would like some other opinions.

Toggl.Foundation/Analytics/AnalyticsEventAttributeInitializer.cs
```diff
@@ -0,0 +1,60 @@
...
+using System.Reflection;
+using Toggl.Multivac;
+using System.Collections.Generic;
+
+namespace Toggl.Foundation.Analytics
+{
+    [Preserve(AllMembers = true)]
+    public abstract class AnalyticsEventAttributeInitializer
+    {
+        protected void InitializeAttributedProperties(IAnalyticsService analyticsService)
```
juanlaube: I have to admit this was weird to read first, but it's clear what it does now. In other languages reflection is supposed to be expensive (I assume it's the same in C#), and I didn't see this technique used in other places (but I assume Mvx does this too). So the question is, is this something we want to do?

juanlaube: Having said that, this will happen only once when the `BaseAnalyticsService` is initialized, so from a performance perspective, maybe it's not a big deal...

simonrozsival: I tested it and it takes < 1 ms on average. I wouldn't be concerned with the performance in this case. But using reflection is indeed slow - let's not use `Activator` to instantiate interactors :trollface:

Toggl.Foundation.Tests/AnalyticsService/AnalyticsEventAttributeInitializerTests.cs
```diff
@@ -0,0 +1,99 @@
...
+
+            instance.EventWithoutParameters.Should().BeOfType<AnalyticsEvent>();
+            instance.EventWithOneParameter.Should().BeOfType<AnalyticsEvent<int>>();
+            instance.EventWithTwoParameters.Should().BeOfType<AnalyticsEvent<string, DateTimeOffset>>();
+            instance.EventWithThreeParameters.Should().BeOfType<AnalyticsEvent<long, bool, object>>();
+            instance.EventWithFourParameters.Should().BeOfType<AnalyticsEvent<long, bool, object, Exception>>();
+        }
+
+        [Fact]
+        public void TracksTheEventWithCorrectNameAndNamedParameters()
```
amulware: I'd like one of these for each of the events, just in case, since they are not tested otherwise.

Toggl.Foundation/Analytics/IAnalyticsService.cs
simonrozsival: I added an issue to maybe deal with this later: https://github.com/toggl/mobileapp/issues/2416

Toggl.Foundation.Tests/AnalyticsService/AnalyticsEventAttributeInitializerTests.cs
```diff
@@ -0,0 +1,127 @@
...
+            public IAnalyticsEvent<long, bool, object> EventWithThreeParameters { get; protected set; }
+
+            [AnalyticsEvent("NameOfParameterI", "NameOfParameterII", "NameOfParameterIII", "NameOfParameterIV")]
+            public IAnalyticsEvent<long, bool, object, Exception> EventWithFourParameters { get; protected set; }
+        }
+
+        private class TestClassWithZeroParameterAttributeWhichThrows : BaseTestClassWithAttributes
+        {
+            [AnalyticsEvent("NameOfParameter1")]
+            public IAnalyticsEvent EventWithFourParameters { get; protected set; }
```
amulware: Naming here, and below?

PR #2406, opened 2018-06-07
Extract DurationFormat value converter into a method

willsb:
Follow up #2269

Since ValueConverters will no longer be needed, we’ll instead have functions that perform the transformations we want. This is a sample with the only value converter used in the settings view. 

Since this converter is used elsewhere I couldn’t delete the Mvx dependent file. The idea is that we’ll delete those converters, one by one, until we no longer depend on any of them.

Squash whenever


Toggl.Foundation.MvvmCross/Transformations/DurationFormatToString.cs
```diff
@@ -0,0 +1,28 @@
...
+                    return Resources.DurationFormatDecimal;
+
+                default:
+                    throw new ArgumentException(
+                        $"Duration format must be either: {nameof(DurationFormat.Classic)}, {nameof(DurationFormat.Improved)} or {nameof(DurationFormat.Decimal)}"
+                    );
+            }
+        }
+    }
+}
```
simonrozsival: 👀

Toggl.Foundation.MvvmCross/Transformations/DurationFormatToString.cs
```diff
@@ -0,0 +1,28 @@
+﻿using System;
+using Toggl.Multivac;
+
+namespace Toggl.Foundation.MvvmCross.Transformations
+{
+    public static class DurationFormatToString
+    {
+        public static string Convert(DurationFormat value)
```
simonrozsival: Would it be a good thing if this was an extension method? I am not sure, just 🤔 out loud - maybe something to consider/discuss.

willsb: I was thinking more along the lines of exposing the raw thin in the view model:

`MyUserObservable.Select(user => user.DurationFormat);`

And then allow for easy convertion in the View:

```csharp
ViewModel.DurationObservable
    .Select(DurationFormatToString.Convert)
    .AsDriver()
    .Subscribe(Label.TextObserver())
    .DisposedBy(disposeBag)
```

That's why I went with static over extension and I believe that has the biggest reusability and conciseness.

kornelijepetak: I like the regular static version better purely because of the conciseness of `Select` as shown in the example. It will not be as nice when we start introducing combiners the same way but still I'd go with static. I was considering the implication of converters being instances, instead of static methods, but I can't remember any case where converter's state would be changed at any point in time, so this is much better. And is really easy to read and reason about.

PR #2407, opened 2018-06-07
Add common extensions to core projects

willsb:
Follow up #2269

This adds a couple of methods that are reusable and will be extensively needed now that we are ditching Mvx bindings. 

- An `AsDriver` extension that does not require a default parameter
- `Subscribe` and `VoidSubscribe`* method to reduce boilerplate when subscribing to `Unit` observables
- The `Invert` common function (very useful since now we won’t have the `InvertedVisibility` value converter)

Squash whenever

—

* A different name is needed because C# has a hard time telling `Func<Task>` and `Action` apart.


Toggl.Foundation.MvvmCross/Extensions/ObservableExtensions.cs
```diff
@@ -0,0 +1,19 @@
...
+{
+    public static class ObservableExtensions
+    {
+        public static IObservable<T> AsDriver<T>(this IObservable<T> observable)
+            => observable.AsDriver(default(T));
+
+        public static IDisposable VoidSubscribe(this IObservable<Unit> observable, Action onNext)
+            => observable.Subscribe((Unit _) => onNext());
+
+        public static IDisposable Subscribe(this IObservable<Unit> observable, Func<Task> onNext)
```
kornelijepetak: Any overloads that handle onError?

Toggl.Multivac/Extensions/FunctionalExtensions.cs
```diff
@@ -7,6 +7,8 @@ namespace Toggl.Multivac.Extensions
     public static class CommonFunctions
     {
         public static T Identity<T>(T x) => x;
+
+        public static bool Invert(bool x) => !x;
```
kornelijepetak: I wanted to add one the other day. 😄  I would call it `Negate` (since the `!` operator is called the negation operator) but `Invert` is by an order of magnitude better name.

Toggl.Foundation.MvvmCross/Extensions/ObservableExtensions.cs
```diff
@@ -0,0 +1,19 @@
...
+
+namespace Toggl.Foundation.MvvmCross.Extensions
+{
+    public static class ObservableExtensions
+    {
+        public static IObservable<T> AsDriver<T>(this IObservable<T> observable)
+            => observable.AsDriver(default(T));
+
+        public static IDisposable VoidSubscribe(this IObservable<Unit> observable, Action onNext)
+            => observable.Subscribe((Unit _) => onNext());
```
kornelijepetak: For some reason `VoidSubscribe` totally feels 🇭🇺 ...
Wouldn't simply `Subscribe` work? I mean, there are no other overloads that takes `Action` as a parameter...

willsb: Read the pr description, please

Toggl.Foundation.MvvmCross/Extensions/ObservableExtensions.cs
willsb: We can add those when needed. This PR merely adds the barebones needed for the settings view for you to reason about

Toggl.Foundation.MvvmCross/Extensions/ObservableExtensions.cs
Odrakir: Can you describe what the alternative without `VoidSubscribe` would be? Isn't there an overload of `Subscribe` that takes an `Action`? (I read the PR, I just don't understand this)

Toggl.Foundation.MvvmCross/Extensions/ObservableExtensions.cs
Odrakir: Same as with `VoidSubscribe`...

Toggl.Multivac/Extensions/FunctionalExtensions.cs
juanlaube: You should call this `Toggl` :D

Toggl.Foundation.MvvmCross/Extensions/ObservableExtensions.cs
juanlaube: The existing overload takes `Action<Unit>`, usages are `.Subscribe(_ => ...)`, usages of this would be `.VoidSubscribe(() => ...)` which takes more characters, but I think improves readability in general 👍

willsb: It doesn't really need more character, @juanlaube, since it allows for one to use void method, so comparably:

`.VoidSubscribe(onSomething)`
vs
`.Subscribe(_ => onSomething())`

juanlaube: You're right, that's much better. But even if it required more typing, it's more readable, so that's good 😄

Odrakir: Oh, it's basically to omit the parameter?

juanlaube: That's the idea, yes

PR #2409, opened 2018-06-07
Add initial Android binding helpers

willsb:
Follow up #2269

This is a tad more complicated than the iOS PR because the android version of the settings view has a RecyclerView in it.

What I’m doing here is: 

- Creating an Activity that is able to [bypass the call to `BindingInflate` in SetContentView](https://github.com/MvvmCross/MvvmCross/blob/develop/MvvmCross.Android.Support/V7.AppCompat/MvxAppCompatActivity.cs#L68-L73). This is needed because we will no longer be using the binding inflater, which is slower.
- Creating a basic ViewHolder that has all we need for easy extensibility
- Creating the ViewHolder that will be used by the settings view
- Creating a BaseAdapter that can be used as a base for all of our future adapters, be they simple or complex
- Creating a simple adapter that can be reused for all collections that don’t require more than 1 kind of layout (that’s most of our collections, so win!)
- Adding the usual methods that will be used in the future for bindings


amulware:
Consider the naming suggestions from https://github.com/toggl/mobileapp/pull/2408 here also.

Toggl.Giskard/Activities/NonBindingAppCompatActivity.cs
```diff
@@ -0,0 +1,91 @@
...
+using MvvmCross.Binding.Droid.BindingContext;
+using MvvmCross.Core.ViewModels;
+using MvvmCross.Core.Views;
+using MvvmCross.Droid.Support.V7.AppCompat.EventSource;
+using MvvmCross.Droid.Views;
+using MvvmCross.Binding.Droid.Views;
+
+namespace Toggl.Giskard.Activities
+{
+    public abstract class NonBindingAppCompatActivity<TViewModel> : MvxEventSourceAppCompatActivity, IMvxAndroidView
```
kornelijepetak: I am in a way troubled by the choice of the name `NonBinding`, because there will be bindings, just not by the mvx `BindingInflate`. So, in a way I would like the name to contain something that would describe that the binding is not automatic... The issue I have with this is that when you inherit from `NonBindingAppCompatActivity` it almost feels like the name wants to tell you that there are no bindings... I dunno... If it weren't for your description (and link), I'd be much more confused about this cause the name itself feels misleading. 
In the end, if we are ditching the bindings thru MVX, I feel that the name should not even mention binding. Why mention something that isn't a default and is also not used. In that sense we could also name it `DoesNotNameCatsByFoodAppCompatActivity` cause, really, it doesn't. 😄

Toggl.Giskard/Extensions/UIExtensions.cs
```diff
@@ -0,0 +1,34 @@
...
+        public static Action<bool> BindIsVisible(this View view)
+            => isVisible => view.Visibility = isVisible.ToVisibility();
+
+        public static Action<string> BindText(this TextView textView)
+            => text => textView.Text = text;
+
+        public static Action<string> BindText(this EditText textEdit)
+            => text => textEdit.Text = text;
+
+        public static Action<bool> BindIsOn(this Switch @switch)
```
daividssilverio: We could make it generic by using the CompoundButton instead. It would cover CheckBoxes and RadioButtons as well.

daividssilverio:
I like the new extensions and base classes.
It makes the RecyclerViews less magical, simpler to understand and debug, which is good in my perspective.
Let's keep this going! I wanna see the UiExtensions in action o/

Toggl.Giskard/Activities/NonBindingAppCompatActivity.cs
daividssilverio: I thought that too, but then I associated with NonMxvBoundActivity somehow.

Toggl.Giskard/Activities/NonBindingAppCompatActivity.cs
```diff
@@ -0,0 +1,91 @@
...
+        where TViewModel : class, IMvxViewModel 
+    {
+
```
kornelijepetak: One empty line too much

Toggl.Giskard/Activities/NonBindingAppCompatActivity.cs
kornelijepetak: In general it's not that big of an issue... just something that felt unnatural. @willsb if you'd still want this to remain as you've put it, please dismiss my review.

Also, did you consider renaming `ObservableExtensions` to something else?

willsb: What name do you suggest?

kornelijepetak:
I approved this on the contingency that we think of a better name in the future, once we phase out Mvx way of doing things... 😄  Naming is hard, so I will not pretend that I have a better name at the moment. 😄

PR #2410, opened 2018-06-07
Some kind of experimental refactoring with expression trees

kornelijepetak:
Just a thought experiment.
I created a PR just to be able to compare with the original branch.

Do not even consider merging 😛


juanlaube:
This is super interesting, why we shouldn't even consider merging? I get this could be performance expensive, but is it _that_ bad? 🤔

daividssilverio:
I find this idea interesting as well. It would be nice not to worry about bundle key names.

kornelijepetak:
Since this uses Lambda.Compile, it will probably (aka most definitely) not work on iOS.

PR #2415, opened 2018-06-07
Fix feedback intent

daividssilverio:
Closes nothing.

It makes the feedback action launch email apps only (supposedly, only apps that can handle the uri "mailto:").

It also fixes this crash when trying to start a new activity outside of an Activity context.


amulware:
Just for clarification, is this the crash fixed here? https://appcenter.ms/orgs/Toggl/apps/Toggl-Android/crashes/groups/3d1d032ad99decc81a53537a1d13fd5e213e2953?appVersion=0.4&period=last7Days

willsb:
Precisely

PR #2417, opened 2018-06-08
Fix text field title glitch on login screen

### LINKED ISSUE #2303: Text field title glitch on login screen
If I focus the email field and then go and grab login data from the password manager, the title animates back as if the field was empty and unfocussed, causing it to overlay the email. Let's fix that

img_1406

sofokli1:
Closes #2303

The issue was that the textfield if first responder was still like that even when the passManager button was pressed and the keyboard was not visible anymore. So resigningFirstResponder and setting the text manually from the vm creates a strange state where the one is sending a goDown and the other is sending a goUp. The solution, resign both textfield asFirstResponder when passManager button pressed.


arturs105:
Yes tested on an actual iPhone with 1Password and didn't see the glitch.

sofokli1:
@Odrakir Please try again because it works fine on my device as well as arturs. If after retesting you still see the issues, some feedback is appreciated (device, os version, how are you reproducing the issue)

Odrakir:
I think I know what's going on. I'm pressing the keychain password button in the keyboard, not the custom one, so the error still happens because the new code is never executed.

Maybe try with `viewWillDisappear`?

sofokli1:
@Odrakir can you make a video/gif so that i can better see the issue on ios 12. i have tested it with 1password and ios autofill but i do not see the issue anymore (on ios 11)

sofokli1:
TLDR; fixed in `BecomeFirstResponder` and counterpart.

Before finding out why this was not working properly I tried to convert the animation to `CABasicAnimation` because at first i thought that there where happening multiple animation at the same time thus the second one was not executing. With `CABasicAnimation` there is the option to cancel animations and do a new one. The code grew too big and I didn't like it. Reverted to the old animation code and started creating a logic to keep track of the last animation requested and if the animation are in progress using the `completionBlock` of the `CATransaction` to keep track of everything. Needless to say that even though the logic was sound the result was the same as if without it. I started looking in to the `BecomeFirstResponder` override which was always returning true even though the `base` (`super` for swift lovers) was returning false. I added an `If` to check the `base` result and then call the animations, also return the `base` from the function and lo and behold it works as it should have been in the first place and now it should continue to work for many iOS versions in the future. And they lived happily ever after.

Odrakir:
I restored iOS 11 and it works there. So this is good to go, just be aware, we should check it in iOS 12 when it ships.

PR #2419, opened 2018-06-08
Refactor preferences model

Odrakir:
More or less the same deal as with `Client`.

We loose validation for now. And also, in this case I'm forced to set a default value in the constructor for `SyncStatus`. Does that look ok?


Toggl.Foundation/DataSources/PreferencesDataSource.cs
```diff
@@ -29,10 +29,12 @@ protected override ConflictResolutionMode ResolveConflicts(IDatabasePreferences
...
-            => Preferences.Builder
-                .FromExisting(existing)
-                .SetFrom(dto)
-                .SetSyncStatus(SyncStatus.SyncNeeded)
-                .Build();
+            => new Preferences(
+                dto.TimeOfDayFormat ?? existing.TimeOfDayFormat,
+                dto.DateFormat ?? existing.DateFormat,
+                dto.DurationFormat ?? existing.DurationFormat,
+                dto.CollapseTimeEntries ?? existing.CollapseTimeEntries,
```
amulware: This will fail for values where `null` is a valid value, and it will require us to change this everytime the preferences constructor updates, even if it's not necessary.

I much rather use the With() method pattern proposed in #1711

kornelijepetak: I agree with both of you. In this case it'd be ok as all items are non-nullable, but I am afraid that this mental flow will break when, as Paul already mentioned, the `null` indeed is a value we want to set. I am afraid that `With` is a much preferred way of doing things. What I would not like to do one way here and other way somewhere else if possible. Also, using `New<T>` helps with those as well.

Odrakir:
I think the only way to not loose validation is having creation DTOs. Otherwise, we need to have these models without some of its parameters when instantiating them for DB creation. We could have 2 different ways of constructing them (which is kind of what was going on until now), but I don't believe that's a good solution.

Toggl.Foundation/Models/Preferences.cs
```diff
@@ -1,129 +1,60 @@
...
-            {
-                if (dto.DateFormat.HasValue)
-                    DateFormat = dto.DateFormat.Value;
+        private Preferences(IPreferences entity, SyncStatus syncStatus, string lastSyncErrorMessage, bool isDeleted = false)
+            : this(entity.TimeOfDayFormat, entity.DateFormat, entity.DurationFormat, entity.CollapseTimeEntries, syncStatus, lastSyncErrorMessage, isDeleted)
+        { }
 
-                if (dto.DurationFormat.HasValue)
-                    DurationFormat = dto.DurationFormat.Value;
+        // Public initializers
```
arturs105: Those are called constructors in C#

amulware: They are also called comments: :fire:

amulware:
On validation: Are we ever creating an incomplete preference object like with other entities though? Since it's a singleton, I think we can keep the validation by doing it in the constructor.

Toggl.Multivac/New.cs
```diff
@@ -1,4 +1,6 @@
-﻿namespace Toggl.Multivac
+﻿using System;
+
+namespace Toggl.Multivac
```
amulware: :fire:

Toggl.Foundation/Models/Preferences.cs
```diff
@@ -1,129 +1,63 @@
...
-                if (DateFormat.Localized == null)
-                    throw new InvalidOperationException($"You must set a valid value to the {nameof(DateFormat)} property before building preferences.");
-
-                if (TimeOfDayFormat.Localized == null)
-                    throw new InvalidOperationException($"You must set a valid value to the {nameof(TimeOfDayFormat)} property before building preferences.");
-            }
+        public Preferences(TimeFormat timeOfDayFormat, DateFormat dateFormat, DurationFormat durationFormat, bool collapseTimeEntries, SyncStatus syncStatus = default(SyncStatus), string lastSyncErrorMessage = "", bool isDeleted = false)
+        {
+            Ensure.Argument.IsNotNull(dateFormat.Localized, nameof(dateFormat));
+            Ensure.Argument.IsNotNull(timeOfDayFormat.Localized, nameof(timeOfDayFormat));
```
amulware: These are structs though and can never be null, right? I might be missing something.

amulware: Yap, I missed the string property access.

Toggl.Foundation/Models/Extensions/ModelExtensions.cs
```diff
@@ -0,0 +1,29 @@
+﻿using Toggl.Foundation.Models.Interfaces;
+using Toggl.Multivac;
+using Toggl.PrimeRadiant;
+
+namespace Toggl.Foundation.Models
+{
+    internal static class ModelExtensions
+    {
+        public static IThreadSafePreferences With(
+            this IThreadSafePreferences original,
```
amulware: Should this be IDatabasePreferences? Or will that operation never be necessary?

Odrakir: In my opinion, `IThreadSafe*` and `IDatabase*` should be different types, not inheriting from each other (even if they have some common parent). But that's for another discussion, in the meantime, yes I would leave it like this and see...

PR #2421, opened 2018-06-08
Track exceptions using AppCenter Crashes

### LINKED ISSUE #2322: Send handled exceptions to AppCenter properly
AppCenter allows for sending exceptions manually now. We should use that and replace the `HandledException` event.

simonrozsival:
Closes #2322 

I decided to add an attribute `IsAnonymizedAttribute` which can be used to mark an exception as safe to be uploaded to a third party. We ask for the existence of the attribute only rarely when the exception is actually caught and processed, so the use of reflection here is OK.

Squash whenever


Toggl.Ultrawave/Exceptions/ApiException.cs
```diff
@@ -25,9 +25,14 @@ internal ApiException(IRequest request, IResponse response, string defaultMessag
             LocalizedApiErrorMessage = getLocalizedMessageFromResponse(response);
         }
 
+        #if DEBUG
         public override string ToString()
             => $"{GetType().Name} ({message}) for request {Request.HttpMethod} {Request.Endpoint} "
                + $"with response {serialisedResponse}";
+        #else
+        public override string ToString()
+            => $"{GetType().Name} ({message}) for request {Request.HttpMethod} {Request.Endpoint}";
```
amulware: Random idea: how about we obfuscate number strings in the endpoint? That would make this entirely anonymous I think.

amulware: i.e. just replace them by "0" or something.

amulware: Or maybe another number we are really sure couldn't happen by accident. We might want to know if we accidentally send 0, so maybe replace it with a non-number-string like "REDACTED" :D

amulware: ideally only for patterns like `/\d+/` though maybe.. I think that covers all numbers we want to replace? (and doesn't replace the api version or date parameters..)

Toggl.Foundation/Analytics/BaseAnalyticsService.cs
```diff
@@ -78,8 +79,23 @@ protected BaseAnalyticsService()
         [AnalyticsEvent("NumberOfCreatedGhosts")]
         public IAnalyticsEvent<int> ProjectGhostsCreated { get; protected set; }
 
+        [AnalyticsEvent("ExceptionType", "ExceptionMessage")]
```
kornelijepetak: Do we need to prefix `ExceptionType` and `ExceptionMessage`? I think simply `Type` and `Message` would be fine. The event type already tells you that it's about exceptions...

Toggl.Foundation/Extensions/ExceptionExtensions.cs
```diff
@@ -0,0 +1,14 @@
...
+
+namespace Toggl.Foundation.Extensions
+{
+    public static class ExceptionExtensions
+    {
+        public static bool CanContainSensitiveInformation(this Exception exception)
+            => !isWhiteListed(exception);
+
+        private static bool isWhiteListed(Exception exception)
+            => exception is ApiException;
```
Odrakir: So, `ApiException` are the only ones that won't get anonimized, right? I understood it the other way around in the issue... 🤔

Toggl.Foundation/Analytics/BaseAnalyticsService.cs
simonrozsival: This is for backwards compatibility.

simonrozsival:
I like the idea of adding an interface which would say if an exception is anonymized or not.

simonrozsival:
Although I think we shouldn't use an interface in this case, but an attribute - I just need to do some research on what are the performance implications, because that is something I am not really sure about (`is` might be just as slow as checking if a class or its ancestors have an attribute, but I am not sure).

Toggl.Foundation/Analytics/BaseAnalyticsService.cs
simonrozsival: Just to make it clear: we already used these names so I don't want to change them so we can keep tracking exceptions the same we did until now.

simonrozsival:
Anonymization of IDs in endpoints was added. Another benefit might be that the exceptions could be grouped by the messages better.

Toggl.Foundation/Analytics/BaseAnalyticsService.cs
```diff
@@ -78,8 +79,23 @@ protected BaseAnalyticsService()
...
+
+        public void Track(Exception exception)
+        {
+            if (exception.IsAnonymized())
+            {
+                TrackException(exception);
+            }
+            else
+            {
+                HandledException.Track(exception.GetType().FullName, exception.Message);
```
juanlaube: Couldn't `exception.Message` also contain sensitive information? This all goes to Firebase as events... 🤔

Toggl.Ultrawave/Exceptions/ApiException.cs
```diff
@@ -26,8 +29,18 @@ internal ApiException(IRequest request, IResponse response, string defaultMessag
...
+            => anonymizedMessage;
+        #endif
+
+        private string detailedMessage
             => $"{GetType().Name} ({message}) for request {Request.HttpMethod} {Request.Endpoint} "
-               + $"with response {serialisedResponse}";
+                + $"with response {serialisedResponse}";
+
+        private string anonymizedMessage
+            => $"{GetType().Name} ({message}) for request {Request.HttpMethod} {Request.Endpoint.Anonymize()}: {LocalizedApiErrorMessage}";
```
juanlaube: Do we know for sure the endpoint is the only thing that contains sensitive information?

Toggl.Ultrawave.Tests/Extensions/UriExtensionsTests.cs
```diff
@@ -0,0 +1,65 @@
...
+        {
+            var uri = endpoints.Workspaces.GetById(id.Get).Url;
+
+            var anonymizedUri = uri.Anonymize();
+
+            anonymizedUri.ToString().Should().Be("https://mobile.toggl.space/api/v9/workspaces/{id}");
+        }
+
+        [Property]
+        public void RemovesIdFromTheUriWitaATrailingSlash(PositiveInt id)
```
daividssilverio: typo?: `RemovesIdFromTheUriWitaATrailingSlash` -> `RemovesIdFromTheUriWithATrailingSlash`

Toggl.Ultrawave/Extensions/UriExtensions.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+using System.Text.RegularExpressions;
+
+namespace Toggl.Ultrawave.Extensions
+{
+    public static class UriExtensions
+    {
+        private static readonly Regex patternForId = new Regex(@"(?<=/)\d+(?=[/]?)", RegexOptions.Compiled | RegexOptions.CultureInvariant);
```
amulware: Can't this be simplified to `/(\d+)`? (you can definitely skip the `[/]?` group, that doesn't add anything.

Toggl.Ultrawave/Exceptions/ApiException.cs
amulware: Yes. the type name and http method are obvious. The message is set in the constructors of all the implementations of this class and are fixed strings that are mostly used as fallback message. The localised error message has been confirmed to be safe by backend.

Toggl.Foundation/Analytics/BaseAnalyticsService.cs
simonrozsival: It could. But it worked like this until now and there are no sensitive data in the event log.

Do we want to do something about this in this PR, @amulware?

simonrozsival:
@amulware hmm

1. I don't know. I am not aware of any deserialization problems we have right now (doesn't mean there aren't any). We would have to ask backend to send us some weird data. Or we could disable the default `At`s for tags (just for testing)
2. Not an issue - it happens so rarely that the impact is non-existent

Toggl.Foundation/Analytics/BaseAnalyticsService.cs
amulware: No, we can revisit it if we have to.

amulware:
1. I'm mostly concerned with the #if DEBUG switch making testing this tricky. But maybe it's ok the way it is.

2. Fair enough.

PR #2422, opened 2018-06-08
Add metrics to Edit View

### LINKED ISSUE #1916: Edit view metrics
I'd like to record the following actions:

- user taps description
- user taps project/task
- user taps tags
- user taps start
- user taps end
- user taps duration
- user taps start date
- user taps billable

I'm not sure right now if these should all be different events or maybe they should be a single `EditViewAction` event with an `Action` property that has one of the above. The latter is probably a bit more future proof and easier to handle on app center/whatever

Alternatively, we could have an event for when the edit view is closed with similar properties as for the time view event: #1915

kornelijepetak:
Closes #1916 

Besides, this issue, this PR also (necessarily) cleans up the commands regarding the Start, Stop and Duration tapping.

Previously on iOS a single command was being called for all three fields, which now had to be changed to allow for metrics hook.

Also, as the views differed over time between iOS and Android, some of the commands were left unused. Now, everything should be correctly hooked up with no unused commands.

ℹ️ Also, it should be noted that iOS has StartDate widget that can be tapped. This does not exist on Android, so this event will not show on android builds. Whether this has a potential to mess up our numbers I have no idea.


Toggl.Foundation.MvvmCross/ViewModels/EditTimeEntryViewModel.cs
```diff
@@ -342,19 +345,32 @@ private Task close()
...
+                case "Duration":
+                    return EditViewTapSource.Duration;
+                default:
+                    throw new ArgumentException("Binding parameter is incorrect.");
+            }
         }
 
         private async Task selectTime(string bindingParameter)
         {
+            var tapSource = getTapSourceFromBindingParameter(bindingParameter);
```
simonrozsival: This smells like platform specific code - can't we pass the `EditViewTapSource` directly from Giskard somehow? Maybe a value converter?

Toggl.Foundation.MvvmCross/ViewModels/EditTimeEntryViewModel.cs
```diff
@@ -387,30 +403,16 @@ private void stopTimeEntry()
             await SelectTimeCommand.ExecuteAsync(bindingParameter);
         }
 
-        private async Task selectEndTime()
+        private async Task selectStopTime()
         {
-            if (IsTimeEntryRunning)
-            {
-                stopTimeEntry();
```
simonrozsival: Wait - where did this code go? I can't see how `selectStartStopDuration` is equivalent to this code. Won't this break something? Like stopping a running time entry?

Toggl.Foundation.MvvmCross/ViewModels/EditTimeEntryViewModel.cs
```diff
@@ -387,30 +403,16 @@ private void stopTimeEntry()
...
-                return;
-            }
-
-            var earliestAllowedTime = StartTime;
-            var latestAllowedTime = StartTime.Add(MaxTimeEntryDuration);
-
-            var parameters = DateTimePickerParameters.WithDates(
-                DateTimePickerMode.Time,
-                StopTime.Value,
-                earliestAllowedTime,
```
simonrozsival: Also why is this irrelevant - the `selectStartStopDuration` does not specify the limits.

Toggl.Foundation.MvvmCross/ViewModels/EditTimeEntryViewModel.cs
kornelijepetak: I agree, but this method was introduced prior to this PR so I suggest you ignore this in this PR, on the contingency that I will do another PR where I will make this better, by exposing this thing as a foundation enum.

Toggl.Foundation.MvvmCross/ViewModels/EditTimeEntryViewModel.cs
kornelijepetak: If you take a look at this at develop, you will see that it is never called. I did not however, look into whether this code was by mistake left here and not moved to some place else during some of the previous PRs that changed the calls to the commands.

Toggl.Foundation.MvvmCross/ViewModels/EditTimeEntryViewModel.cs
kornelijepetak: Issue has been created #2429

Toggl.Foundation.MvvmCross/ViewModels/EditTimeEntryViewModel.cs
simonrozsival: OK. This is indeed weird. It seemed pretty scary, but you are right that we don't use that anywhere (anymore? did we ever use it?).

Toggl.Foundation.MvvmCross/ViewModels/EditTimeEntryViewModel.cs
```diff
@@ -184,6 +184,8 @@ public TimeSpan DisplayedDuration
 
         public IMvxCommand ToggleBillableCommand { get; }
 
+        public IMvxCommand StartEditingDescriptionCommand { get; }
```
simonrozsival: Why not reuse the `FocusDurationCommand`?

simonrozsival: Why not reuse the FocusDurationCommand?

Toggl.Daneel/ViewControllers/EditTimeEntryViewController.cs
```diff
@@ -80,8 +80,8 @@ public override void ViewDidLoad()
                 Color.EditTimeEntry.ClientText.ToNativeColor(),
                 false
             );
-            var stopRunningTimeEntryAndEditDurationForStoppedConverter = new BoolToConstantValueConverter<IMvxCommand>(
-                ViewModel.StopCommand, ViewModel.EditDurationCommand);
+            var stopRunningTimeEntryAndSelectStopTimeForStoppedConverter = new BoolToConstantValueConverter<IMvxCommand>(
```
Odrakir: I know this wasn't don't in this PR, but as I'm finding out now, I might as well comment:

I don't like this, we are executing commands inside a conversion method. Which to my understanding is like doing side effects inside a `map`, not a great idea. Also, in my opinion, the end time button should map to just one command, and do the necessary magic in the viewmodel and not here.

Toggl.Foundation.MvvmCross/ViewModels/EditTimeEntryViewModel.cs
```diff
@@ -387,30 +403,16 @@ private void stopTimeEntry()
             await SelectTimeCommand.ExecuteAsync(bindingParameter);
```
Odrakir: Also, not from this PR, but... oh well. Calling a command from another command? Isn't that confusing?

simonrozsival: A command might have additional logic which will check whether it could be executed or not - doing this might skip executing the private function which could be called directly otherwise. We might remove this call if it is unnecessary in this VM.

Toggl.Daneel/ViewControllers/EditTimeEntryViewController.cs
simonrozsival: I am not sure, but I think this is iOS specific logic - @kornelijepetak might prove me wrong. If it is shared logic then it could be in fact handled in the VM.

Odrakir: Even if it's iOS specific, shouldn't there be a better way to do it?

kornelijepetak: 1. It is iOS specific.
2. This is not a subject of this PR (but if you are bothered by this, I am ok with creating an issue)
3. I think the there are better ways to do it, and I would much rather do it by branching in VM than in VC.

kornelijepetak: But this is also something not required for this PR.

@Odrakir I am also a fan of partial classes so that we can separate different concerns into separate files (re: order of vars)... But this is only helpful if we have good tooling, which we kind of don't.

kornelijepetak: I will add this to my refactoring PR which will come soon, so please ignore the order of variables for this PR.

Toggl.Foundation.MvvmCross/ViewModels/EditTimeEntryViewModel.cs
kornelijepetak: What do you mean, the name? Cause this VM does not have such a command.

simonrozsival: 🙈 you are right. I looked into a wrong file.

I don't like this command - it doesn't have any effect on the UI. This seems like this should be a side effect of the `IsEditingDescription` setter. On iOS it should be easy to change the `TextViewWithPlaceholder` to expose `bool isFocused` and bind it to `IsEditingDescription`.

arturs105:
I think this may conflict with #2430, since it deals with the same commands. You should probably talk to @khoi to decrease the pain of merge conflicts.

Toggl.Foundation.MvvmCross/ViewModels/EditTimeEntryViewModel.cs
kornelijepetak: I am afraid this has to be done this way. There are other cases that the description gets focus which were not user taps. For example, user can tap the description (TRACK!) and then edit it. Then he can tap on the project field. The project selection gets opened. Then he closes it and bang, all of a sudden, the field gets this field focused again (which should not be tracked, but IS). If I'm holding it wrong, please advise.

Toggl.Foundation/Analytics/BaseAnalyticsService.cs
```diff
@@ -75,6 +75,9 @@ protected BaseAnalyticsService()
         [AnalyticsEvent]
         public IAnalyticsEvent OfflineModeDetected { get; protected set; }
 
+        [AnalyticsEvent("TapSource")]
+        public IAnalyticsEvent<EditViewTapSource> TrackEditViewTap { get; set; }
```
amulware: Are you sure about the naming of this? Why not just `EditViewTap(ped)`?

In general, we should be careful how we name our events, to ensure consistency. Changing them in the future might be annoying when extracting statistics and should be avoided.

kornelijepetak: `EditViewTapped` seems good. Thanks for the suggestion. The name remained from the previous way of doing things `Track*`.

Toggl.Foundation.MvvmCross/ViewModels/EditTimeEntryViewModel.cs
```diff
@@ -342,19 +344,47 @@ private Task close()
...
                 .Navigate<SelectDateTimeViewModel, DateTimePickerParameters, DateTimeOffset>(parameters)
                 .ConfigureAwait(false);
+
+            if (IsTimeEntryRunning == false)
+            {
+                StopTime = StartTime + duration;
+            }
+        }
+
+        private async Task selectDuration()
```
Odrakir: Maybe move this one up with `selectStopTime` and `selectStartDate`? (the commands too)

PR #2423, opened 2018-06-08
Fix Google login

### LINKED ISSUE #2361: Google sign up not working

juanlaube:
Closes #2361


Toggl.Giskard/Services/GoogleService.cs
```diff
@@ -21,6 +23,7 @@ public sealed class GoogleService : MvxAndroidTask, IGoogleService
 
         private bool isLoggingIn;
         private Subject<string> subject = new Subject<string>();
+        private string scope = $"oauth2:{Scopes.Profile}";
```
willsb: `readonly`

daividssilverio:
The Google Login is nice and working.

The Google Sign Up have one UI issue: The sign up with Google button goes down, touching the bicycle card in the bottom where there’s some error showing.

Other issues:
1 - We can’t pick accounts after we’ve already selected one.
2 - After trying to sign up with google with an email that is already registered, it will fail, but the error is not clear, I guess it should be: “You already have an account, try Loggin in” or just login for the user.
3 - If you try signing up again with google after the first error, it will show the loading indicator, but nothing happens. It loads forever, no Api call/response.

daividssilverio:
ref #2432 #2433 #2434

PR #2424, opened 2018-06-08
Add ISyncState interface

amulware:
## What's this?

This adds two interfaces, `ISyncState` and `ISyncState<T>` with respective `Start` methods and makes all sync states implement the appropriate one.

## Why do we need this?

The three reasons to have this, from least to most important, are:

- It makes the SyncManagerFactory setup code slightly more concise
- This will be needed in a follow-up PR to generate sync graph diagrams
- It enforces convention that we were already following, making it easier to write syncing code correctly

This PR itself does not change or add to the functionality of the syncing algorithm in any way.

## Review hints

This modifies a lot of files since all sync states need to implement the new interface, but most of the changes are trivial.

Still, to make it possibly easier to review, I split this up into several commits, so you can take a look at specific parts.

All commits represent working states of Foundation, but 👷 8fd3354 breaks the tests, before they are fixed in 💚 70a969d.


Toggl.Foundation/Sync/ISyncState.cs
```diff
@@ -0,0 +1,14 @@
...
+namespace Toggl.Foundation.Sync
+{
+    public interface ISyncState
+    {
+        IObservable<ITransition> Start();
+    }
+
+    public interface ISyncState<in T>
+    {
+        IObservable<ITransition> Start(T _);
```
amulware: I can see the naming of this parameter being controversial. The reason I named it this is because if I called it `parameter(s)`, someone writing a new state might be tempted to just take that name, after having the method automatically generated by the IDE. Instead, I hope this name encourages coming up with a proper name.

arturs105: What about `IObservable<ITransition> Start(T insertYourOwnParameterName);`? 😄
Although this is meant as a joke, i think this would make sure that no one uses the default name.

willsb: I agree with arturs on this one. A name that literally says "rename this" is the way.

Example: https://github.com/xamarin/xamarin-macios/blob/c0202e0e2be22bc1e9ff08e26473b9ae102842d7/src/CoreLocation/CLLocationManagerDelegate.cs#L47

PR #2425, opened 2018-06-09
Buffed te creation event without type safety

daividssilverio:
I tried to make the Non type safe version of the IAnalyticsEvent, but didn't like the end result when using the Analytics service.

It is quite error prone, you have to look at the `BaseAnalyticsService` to match the parameters.

But since I made it already, what do you think about this?

Don't bother about failing tests, I didn't fix all usages of the old TimeEntryStarted event.

Don't merge this.


simonrozsival:
I don't think we should use this - at least it seems that for events with too many params we should use specific event types as I already wrote in #2426

daividssilverio:
I'm closing this as we are not going this way.

PR #2426, opened 2018-06-09
Improve Start te analytics event to send more data

### LINKED ISSUE #1919: Improve TE creation event
We have an event for creating time entries that specifies the source of creation.

In addition I'd like to include the following properties:

- TE has non-empty description
- TE has project
- TE has task
- number of tags on TE
- TE is billable
- TE is running

daividssilverio:
Closes #1919 

I tried creating a Non type safe version, but didn't like the result because it felt weird and the IDE wasn't really helpful when using it, and it felt really error prone and not dev friendly.

I've created a wip for that, if you want to compare: #2425


daividssilverio:
Wouldn't it be ok to just have a lot of `IAnalyticsEvent<T1, T2, T3, T4, T5, T6...` overloads?
Maybe we could just hava small data class for Events and handle their conversion to key-pairs on the Track method, it would work too.

kornelijepetak:
Am I seeing something incorrectly? This does not is the new `Track()` way of doing this... Maybe it needs develop merged into it?

daividssilverio:
@kornelijepetak, This branch is updated with develop. This PR does use the old tracking way. There was one small discussion whether to use multiple events or the old tracking way in cases like this here #1915. 

I wasn't sure which way to go, it seemed bad to have the event divided into multiple smaller events because it would take more work to synthesize the tracked data, so I've tried creating a not type-safe version of the new tracking way here: #2425 for cases like this with many parameters to track. But it seemed bad to use and quite error prone, then I've used the old way in this PR. Would you rather add the `IAnalyticsEvent<T1, T2, T3, T4, T5, T6, T7>` interface to track this event with all its parameters using the new `Track(...)` way?

simonrozsival:
We had an internal discussion about this with @kornelijepetak and @amulware and we found a possible solution with specific event classes. Kornelije should create a PR tomorrow so let's wait until then and see what the solution will look like. I think we should close this PR as I believe we shouldn't do things _the old way_ (million string consts in the analytics class...).

Issue #1436, opened 2018-02-07
Polish numpad duration editing


amulware:
- [ ] Do not allow user to select text or change the location of the cursor (i.e. disable tapping while editing the duration completely)
- [ ] If possible, disable letters on numpad
- [ ] Trailing 0s should be gray, and the colon should be gray for durations <1h (see designs)

khoi:
>If possible, disable letters on numpad
This is not supported by the official Apple's keyboards.

khoi:
@amulware Please collaborate more on 
>Trailing 0s should be gray, and the colon should be gray for durations <1h (see designs)

Did you mean `leading zeros`? Since the numbers are filled from right to left.

>duration < 1h

I also need clarification for this, what happens if the user enters 99?

amulware:
Yap, leading zeros makes a lot more sense. If the user enters 99, let's still keep the colon gray. It's more about the number of digits I suppose.

PR #2428, opened 2018-06-11
show keyboard when edit view appears

### LINKED ISSUE #1434: Wheel view: open the numpad automatically
When a user taps the duration field in the edit view which opens the wheel view, open numpad automatically.
Do not open the numpad when a user opens the wheel view by tapping start/end times.

khoi:
Closes #1434, maybe related to #1436

Making the `DurationInput` first responder of the View Controller.  

I was wondering whether we should not replace the current duration with `00:00` since now the keyboard opens automatically. (See the attached GIF)

Thoughts would be appreciated.

### DIFF
--- a/Toggl.Daneel/ViewControllers/EditDurationViewController.cs
+++ b/Toggl.Daneel/ViewControllers/EditDurationViewController.cs
@@ -218,6 +218,8 @@ namespace Toggl.Daneel.ViewControllers
             base.ViewDidAppear(animated);
 
             viewDidAppear = true;
+
+            DurationInput.BecomeFirstResponder();
         }
 
         protected override void KeyboardWillShow(object sender, UIKeyboardEventArgs e)


kornelijepetak:
Summoning @pe0ny for the `00:00` question.

pe0ny:
Like @kornelijepetak already mentioned in Slack, the main reason for resetting it to `00:00` is because most of the time users want to add a completely different duration. Also, editing duration with the keyboard controls could be difficult when having existing duration there first.
I agree with the part that I would be nice to show existing duration somewhere. I'll add this to my list and see what we can figure out UX-wise. Thanks for pointing it out though :)

amulware:
@pe0ny one thing we could do: Show the real duration instead of 00:00 as empty text, but in gray, implying that the user can type over it? If they start typing it would override the numbers. If they hit < it could make them real and delete the last of it.

Or maybe this is much too complicated /shrug

PR #2430, opened 2018-06-11
Show numpad only when user clicks on duration for edit view

### LINKED ISSUE #1434: Wheel view: open the numpad automatically

khoi:
Closes #1434 

In `EditTimeEntryViewModel`, I'm exposing both `EditDurationCommandAndShowNumpad` and `EditDurationCommand` as 2 separated commands 

Instead of having one 
```csharp
EditDurationCommandAndShowNumpad = new MvxAsyncCommand<bool>(editDurationAndShowNumpad);
```

My reasoning for that is the View Controller shouldn't know about the boolean logic of mapping when to show the numPad or not. 

This solution is a little bit verbose, do you think I should go with a single Command instead?

**Update**

After the refactor from #2422, the above is no longer valid. Now selecting duration has its own separated command. This ticket will take leverage of that.


Toggl.Foundation.MvvmCross/ViewModels/EditTimeEntryViewModel.cs
```diff
@@ -170,6 +170,8 @@ public TimeSpan DisplayedDuration
 
         public IMvxAsyncCommand EditDurationCommand { get; }
 
+        public IMvxAsyncCommand EditDurationCommandAndShowNumpad { get; }
```
juanlaube: I think `EditDurationAndShowNumpadCommand` reads better. I'm not sure if this is a convention (probably it is), but in general, all commands are named `DoStuffCommand`.

khoi:
I've updated the tests and also the naming 💃

Toggl.Foundation.MvvmCross/Parameters/EditDurationParameters.cs
```diff
@@ -0,0 +1,21 @@
...
+{
+    public sealed class EditDurationParameters
+    {
+        public enum EditDurationInitialFocus
+        {
+            Duration,
+            None
+        }
+
+        public DurationParameter DurationParam { get; }
```
simonrozsival: I think we don't need the "old" `DurationParameter` class anymore - in this case I think you should just change the old class instead of adding a new one which uses it internally. Less code is always better than more code! :)

Toggl.Foundation.MvvmCross/ViewModels/EditTimeEntryViewModel.cs
```diff
@@ -170,6 +170,8 @@ public TimeSpan DisplayedDuration
 
         public IMvxAsyncCommand EditDurationCommand { get; }
 
+        public IMvxAsyncCommand EditDurationAndShowNumpadCommand { get; }
```
simonrozsival: I would keep just the `EditDurationCommand` but I would make it of type `IMvxAsyncCommand<EditDurationInitialFocus>` - the focus type would be specified explicitly in the binding.

Toggl.Foundation.MvvmCross/ViewModels/EditTimeEntryViewModel.cs
```diff
@@ -467,11 +471,22 @@ private void stopTimeEntry()
...
+
+        private async Task editDurationAndShowNumpad()
+        {
+            editDuration(true);
+        }
+
+        private async Task editDuration(bool showNumpad)
         {
             var duration = StopTime.HasValue ? Duration : (TimeSpan?)null;
             var currentDuration = DurationParameter.WithStartAndDuration(StartTime, duration);
```
simonrozsival: We should pass the initial focus parameter to this factory method (after merging the `EditDurationParameters` and `DurationParameter` classes) and rename it just to `With` - then use named params if you like them, but let's not keep the param names in the method name anymore.

arturs105:
I think the refactored commands from #2422 could come in handy for this PR. Anyways, coordinate with @kornelijepetak so you get less merge conflicts.

Toggl.Foundation.MvvmCross/Parameters/EditDurationParameters.cs
khoi: Correct me if I'm wrong but `DurationParameter` is purely a bag to store date related information. And it's used not only in this `EditDurationViewModel` but also others.

I didn't want to add this `InitialFocus` thingy to `DurationParameter` since it doesn't really belong there.

So this `EditDurationParameters` is like a decoration for `DurationParameter` adding `InitialFocus` property which is specifically for `EditDurationViewModel`

khoi:
The refactor from @kornelijepetak certainly help with this. 🙇

Toggl.Foundation.MvvmCross/Parameters/EditDurationParameters.cs
```diff
@@ -0,0 +1,21 @@
+﻿namespace Toggl.Foundation.MvvmCross.Parameters
+{
+    public sealed class EditDurationParameters
+    {
+        public enum EditDurationInitialFocus
```
Odrakir: Could this be called just `InitialFocus`?

khoi: Certainly, consider it's already namespaced under `EditDurationParameters`

kornelijepetak: I was just writing that in my review. 😄

Toggl.Foundation.MvvmCross/Parameters/EditDurationParameters.cs
kornelijepetak: Why didn't you just inherit from it then and just add the `InitialFocus`? Since this new parameters class is indeed in an `is-a` relationship with the `DurationParameters`.

khoi: I have to disagree. This is not an `is-a` relationship but rather `has-a`. `EditDurationParameters` own a `DurationParameters` but is NOT a `DurationParameters`

kornelijepetak: Well... depends on how you look at it... But I could agree in a way.

khoi:
@kornelijepetak Since the user can come to Edit Duration screen by 3 methods
- Hitting start time, stop time, duration

I intended for `InitialFocus` to be an enum that eventually may contains `StartTime` and `StopTime`.

This also explains the reason why I use the switch statement instead of a simple `if`.

kornelijepetak:
@khoi I see...

Will we ever focus other controls automatically when start or stop are tapped? 
@pe0ny @lizaPlotnikova, please advise.

amulware:
I may not have the final word on this, but I don't think we will ever focus start/stop when opening the time view directly. I think using the wheel should be the default.

khoi:
Let's wait for @pe0ny and then I'll make necessary adjustment

pe0ny:
@khoi I agree with Paul. Start/stop shouldn't be focused.

khoi:
@Odrakir Can i have your approval on this?

PR #2438, opened 2018-06-12
Shake the email, password fields if invalid

### LINKED ISSUE #2234: Log in: Use shake animation to indicate missing/invalid field
When the user taps "Log in" while one of the two (email, password) fields is missing/invalid, use the shake animation (like when you enter a wrong password while logging into your mac) to draw attention to that field.

The Log in button should still look disabled but it should trigger the shake animation once user taps on it.

khoi:
Close #2234


Toggl.Daneel/Views/ViewShaker.cs
```diff
@@ -0,0 +1,30 @@
...
+    public sealed class ViewShaker
+    {
+        private readonly UIView view;
+
+        public ViewShaker(UIView view)
+        {
+            this.view = view;
+        }
+
+        public void Shake(double duration = 0.05, int repeatCount = 5)
```
simonrozsival: I think this should be an extension method for the `UIView` class - the `Shake` method is called just once on each instance of `ViewShaker`, right?

Toggl.Daneel/Views/ViewShaker.cs
```diff
@@ -0,0 +1,30 @@
...
+        {
+            var animation = CABasicAnimation.FromKeyPath("position");
+            animation.Duration = duration;
+            animation.RepeatCount = repeatCount;
+            animation.AutoReverses = true;
+            animation.From =
+                NSValue.FromCGPoint(new CGPoint(view.Center.X - 4.0, view.Center.Y));
```
simonrozsival: Please use constants instead of magic numbers.

Toggl.Daneel/ViewControllers/Login.storyboard
```diff
@@ -40,6 +40,10 @@
                                 <fontDescription key="fontDescription" type="system" pointSize="14"/>
                                 <textInputTraits key="textInputTraits" keyboardType="emailAddress" returnKeyType="next" textContentType="email"/>
                             </textField>
+                            <button opaque="NO" contentMode="scaleToFill" contentHorizontalAlignment="center" contentVerticalAlignment="center" buttonType="roundedRect" lineBreakMode="middleTruncation" translatesAutoresizingMaskIntoConstraints="NO" id="Sak-LC-gZ0">
+                                <rect key="frame" x="16" y="297.5" width="343" height="40"/>
+                                <state key="normal" title="Button"/>
```
simonrozsival: Why do we need this btn? Is it because we can't recognize gestures on disabled buttons?

khoi: yes, detecting tap on a disabled button is tricky. And also adding 1 more event to the Login button increases the complexity of the ViewModel.

Using this invisible button, we can keep ViewModel as is.

simonrozsival: OK. No problem then. I wanted to ask you to remove the title, but I see you did that already :D good job reading my thoughts!

PR #2440, opened 2018-06-12
Shake the invalid fields for signup

### LINKED ISSUE #2233: Sign up: Use shake animation to indicate missing/invalid field
When the user taps "Sign up" while one of the three (email, password, country) fields is missing/invalid, use the shake animation (like when you enter a wrong password while logging into your mac) to draw attention to that field.

The Sign up button should look disabled but it should trigger the shake animation once user taps on it.

Note: In the country field, the red exclamation mark should be displayed anyways when the API fails. Also, the error message below the password field (if any).

khoi:
Close #2233

### DIFF
--- a/Toggl.Foundation.Tests/MvvmCross/ViewModels/SignupViewModelTests.cs
+++ b/Toggl.Foundation.Tests/MvvmCross/ViewModels/SignupViewModelTests.cs
@@ -188,7 +188,7 @@ namespace Toggl.Foundation.Tests.MvvmCross.ViewModels
             }
 
             [Fact, LogIfTooSlow]
-            public async Task IsCountryValidShouldBeFaulseWhenNetworkFailed()
+            public async Task IsCountryValidShouldBeFalseWhenNetworkFailed()
             {
                 Api.Location.Get().Returns(Observable.Throw<ILocation>(new Exception()));
 
@@ -198,7 +198,7 @@ namespace Toggl.Foundation.Tests.MvvmCross.ViewModels
             }
 
             [Fact, LogIfTooSlow]
-            public async Task IsCountryValidShouldBeTrueWhenApiSucceed()
+            public async Task IsCountryValidShouldBeTrueWhenApiSucceeds()
             {
                 await ViewModel.Initialize();
 
--- a/Toggl.Foundation.Tests/MvvmCross/ViewModels/SignupViewModelTests.cs
+++ b/Toggl.Foundation.Tests/MvvmCross/ViewModels/SignupViewModelTests.cs
@@ -186,6 +186,20 @@ namespace Toggl.Foundation.Tests.MvvmCross.ViewModels
 
                 ViewModel.IsCountryErrorVisible.Should().BeFalse();
             }
+
+            [Fact, LogIfTooSlow]
+            public async Task IsCountryValidShouldBeFaulseWhenNetworkFailed()
+            {
+                Api.Location.Get().Returns(Observable.Throw<ILocation>(new Exception()));
+                await ViewModel.Initialize();
+                ViewModel.IsCountryValid.Should().BeFalse();
+            }
+
+            public async Task IsCountryValidShouldBeTrueWhenApiSucceed()
+            {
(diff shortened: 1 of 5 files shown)


Toggl.Foundation.Tests/MvvmCross/ViewModels/SignupViewModelTests.cs
```diff
@@ -186,6 +186,20 @@ await NavigationService
...
+
+            [Fact, LogIfTooSlow]
+            public async Task IsCountryValidShouldBeFaulseWhenNetworkFailed()
+            {
+                Api.Location.Get().Returns(Observable.Throw<ILocation>(new Exception()));
+                await ViewModel.Initialize();
+                ViewModel.IsCountryValid.Should().BeFalse();
+            }
+
+            public async Task IsCountryValidShouldBeTrueWhenApiSucceed()
```
simonrozsival: Attribute is missing here.

Toggl.Foundation.Tests/MvvmCross/ViewModels/SignupViewModelTests.cs
```diff
@@ -186,6 +186,20 @@ await NavigationService
 
                 ViewModel.IsCountryErrorVisible.Should().BeFalse();
             }
+
+            [Fact, LogIfTooSlow]
+            public async Task IsCountryValidShouldBeFaulseWhenNetworkFailed()
+            {
+                Api.Location.Get().Returns(Observable.Throw<ILocation>(new Exception()));
+                await ViewModel.Initialize();
```
simonrozsival: Also please add blank lines - AAA - http://wiki.c2.com/?ArrangeActAssert

Odrakir:
The issue itself is a little weird, being the button disabled it shouldn't do anything at all.

But if we do want to implement it... Isn't' there a cleaner way that doesn't imply adding a new invisible button? I might be nitpicking, but this is that kind of thing that two months from now is hard to understand why it was there.

This is just a comment, not a change request.

khoi:
That's the tricky part, a disabled button shouldn't do anything at all. That's also why adding an action to a disabled button is kinda tricky.

@simonrozsival suggested that we could fake the `disable` state by changing the title color. But I think that's actually more complicated than this implementation.

I understand your point. That's why I try to be very clear regarding what the button does by named it `ShakeTriggerButton`

khoi:
@simonrozsival your requested GIF

arturs105:
I think that it looks weird that only the text shakes, we should shake the exclamation mark and the caret as well.
jun-12-2018 13-29-48

khoi:
new shaking behavior

Toggl.Foundation.Tests/MvvmCross/ViewModels/SignupViewModelTests.cs
```diff
@@ -186,6 +186,24 @@ await NavigationService
 
                 ViewModel.IsCountryErrorVisible.Should().BeFalse();
             }
+
+            [Fact, LogIfTooSlow]
+            public async Task IsCountryValidShouldBeFaulseWhenNetworkFailed()
```
kornelijepetak: `False` instead of `Faulse`

Toggl.Foundation.Tests/MvvmCross/ViewModels/SignupViewModelTests.cs
```diff
@@ -186,6 +186,24 @@ await NavigationService
...
+            {
+                Api.Location.Get().Returns(Observable.Throw<ILocation>(new Exception()));
+
+                await ViewModel.Initialize();
+
+                ViewModel.IsCountryValid.Should().BeFalse();
+            }
+
+            [Fact, LogIfTooSlow]
+            public async Task IsCountryValidShouldBeTrueWhenApiSucceed()
```
kornelijepetak: Succeed`s`

khoi: I can't spell 🤦🏻‍♂️

Issue #1918, opened 2018-04-18
Start view metrics


amulware:
Things I care about:

- user tapped project button
- user tapped tag button
- user tapped billable button
- user tapped calendar button
- user tapped time view button
- user tapped duration field

- user opened project selection by typing @
- user opened tag selection by typing #
- user tapped the project button in the empty state
- user tapped the tag button in the project state

- user auto completed from a suggestion

There is some overlap in #1777

Similar to #1916 I'm not sure what the best form for this event/these events is.

kornelijepetak:
@amulware can you please explain what this means:
* [ ] user tapped the project button in the empty state
* [ ] user tapped the tag button in the project state

In addition, since these are currently rather difficult to separate the way we are doing the query and also in light of upcoming changes to the TextFieldInfo, I will move these to another issue that should be fixed when that is done:
* [ ] user opened project selection by typing @
* [ ] user opened tag selection by typing #

amulware:
The two points in question are referring to these buttons:

![4bb5ae97-9b43-4959-aa94-50df8513d0e1](https://user-images.githubusercontent.com/3987892/41335978-b07ffe0e-6ee2-11e8-967a-fe92788544ab.JPG)

amulware:
I believe that we are already tracking the three different ways to open project/tag selection though? I remember @Odrakir working on that 🤔

kornelijepetak:
This was done by #2448 and the rest is in #2447.

Clean up is required in the form of #2486

Issue #2429, opened 2018-06-11
Refactor EditTimeEntryViewModel to remove the Giskard dependency


kornelijepetak:
Currently, the binding parameter is being used in one method in the above mentioned VM.

The idea is to change this into a foundation-based enum and use the enum. 

Giskard should call this command (if possible) in such a way that VM gets only an enum value, and not a string. Dunno if that's possible with MVX binding though...

PR #2448, opened 2018-06-12
Startview metrics

kornelijepetak:
Closes most of #1918 (two of the points are not clear to me)...

This PR does does this: 

* [x] user tapped project button
* [x] user tapped tag button
* [x] user tapped billable button
* [x] user tapped calendar button
* [x] user tapped time view button
* [x] user tapped duration field
* [x] user auto completed from a suggestion


Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
```diff
@@ -558,11 +576,30 @@ private void toggleTaskSuggestions(ProjectSuggestion projectSuggestion)
 
         private void toggleBillable()
         {
+            analyticsService.StartViewTapped.Track(StartViewTapSource.Billable);
```
sofokli1: Are we tracking enable and disable of the toggle as if those are the same thing?

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
```diff
@@ -558,11 +576,30 @@ private void toggleTaskSuggestions(ProjectSuggestion projectSuggestion)
...
+                    return StartViewTapSource.Duration;
+                default:
+                    throw new ArgumentException("Binding parameter is incorrect.");
+            }
+        }
+
         private async Task selectTime(string bindingString)
         {
+            var tapSource = getTapSourceFromBindingParameter(bindingString);
+            analyticsService.StartViewTapped.Track(tapSource);
```
daividssilverio: I didn't find tests related to the `StartViewTapped.Track` originated from here. Does it make sense to add them to the `TheSelectTimeCommand` test class?

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
```diff
@@ -558,11 +576,30 @@ private void toggleTaskSuggestions(ProjectSuggestion projectSuggestion)
...
+            switch (bindingParameter)
+            {
+                case "StartTime":
+                    return StartViewTapSource.StartTime;
+                case "StartDate":
+                    return StartViewTapSource.StartDate;
+                case "Duration":
+                    return StartViewTapSource.Duration;
+                default:
+                    throw new ArgumentException("Binding parameter is incorrect.");
```
sofokli1: If we add new bindingParameters in the future why should this throw an exception?
Would it make sense to rename this func and call `analyticsService.StartViewTapped.Track(tapSource)` in here and not throwing as default but just return.
Just an idea not a request.

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
sofokli1: What I am thinking is that if we add a feature to set the default state of the toggle to always `on` or `off`, than it is not clear what this event means. Are they disabling the toggle because the default is `on` or the other way around?

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
kornelijepetak: This really should not allow any other way of reaching this screen. That is why it throws. However, this will be improved by #2429 .

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
kornelijepetak: I do not know what was the original idea. @amulware what do we actually want to track. Only taps, or the values as well?

Toggl.Foundation.Tests/MvvmCross/ViewModels/StartTimeEntryViewModelTests.cs
```diff
@@ -1756,6 +1760,40 @@ private Task callCommandCorrectly(int? hoursToAddToStopTime = null)
...
+            }
+
+            [Fact, LogIfTooSlow]
+            public async Task TracksStartDateTap()
+            {
+                ViewModel.Prepare();
+
+                await ViewModel.SelectTimeCommand.ExecuteAsync("StartDate");
+
+                ViewModel.TextFieldInfo = ViewModel.TextFieldInfo.WithTextAndCursor("abcde @fgh", 10);
```
daividssilverio: :fire:

Toggl.Foundation.Tests/MvvmCross/ViewModels/StartTimeEntryViewModelTests.cs
```diff
@@ -1756,6 +1760,40 @@ private Task callCommandCorrectly(int? hoursToAddToStopTime = null)
...
+            }
+
+            [Fact, LogIfTooSlow]
+            public async Task TracksDurationTap()
+            {
+                ViewModel.Prepare();
+
+                await ViewModel.SelectTimeCommand.ExecuteAsync("Duration");
+
+                ViewModel.TextFieldInfo = ViewModel.TextFieldInfo.WithTextAndCursor("abcde @fgh", 10);
```
daividssilverio: 🔥

kornelijepetak:
Some tests are misbehaving, I'll have to look into that. Until then, I've put this `on-hold`.

PR #2449, opened 2018-06-12
Inaccessible project display

### LINKED ISSUE #2366: Polish inaccessible project display
From https://github.com/toggl/mobileapp/pull/2351

- is it OK to show the "(inaccessible project)" until we fetch the project?
- I reused the "no project" color (dark gray) - is that OK?

More details later/upon request.

juanlaube:
Closes #2366 

🦑 whenever


Toggl.Foundation/Extensions/ThreadSafeExtensions.cs
```diff
@@ -0,0 +1,31 @@
...
+        public static string DisplayName(this IThreadSafeProject project)
+        {
+            var name = project.Name ?? "";
+
+            if (name == Resources.InaccessibleProject)
+            {
+                return name;
+            }
+
+            return project.Active ? name : $"{name} (archived)";
```
sofokli1: Why not put `(archived)` in Resources?

Toggl.Foundation/Extensions/ThreadSafeExtensions.cs
```diff
@@ -0,0 +1,31 @@
...
+{
+    public static class ThreadSafeExtensions
+    {
+        private const string archivedColor = "#cecece";
+
+        public static string DisplayName(this IThreadSafeProject project)
+        {
+            var name = project.Name ?? "";
+
+            if (name == Resources.InaccessibleProject)
```
amulware: I'm not liking this check, here and below. Is there no safer way to do this based on the project syncstate maybe? In fact, I'd say that if the project is a ghost, we shouldn't even consider it's name and just return the resource string maybe?

juanlaube: I don't like this either and was expecting this to appear here :) I will see if there's anything else that gives us a hint of ghost projects, but so far I couldn't find anything

amulware: Check the PRs on the subject, they use a different sync state if I'm not mistaken.

Toggl.Foundation/Extensions/ThreadSafeExtensions.cs
```diff
@@ -0,0 +1,33 @@
+﻿using Toggl.Foundation.Models.Interfaces;
+
+namespace Toggl.Foundation.Extensions
+{
+    public static class ThreadSafeExtensions
+    {
+        private const string archivedColor = "#cecece";
```
amulware: We don't actually use this colour for archived projects, so I think it's better to name the constant `inaccessibleProjectColor` or similar.

Toggl.Foundation/Extensions/ThreadSafeExtensions.cs
```diff
@@ -0,0 +1,33 @@
...
+
+        public static string DisplayName(this IThreadSafeProject project)
+        {
+            var name = project.Name ?? "";
+
+            switch (project.SyncStatus)
+            {
+                case PrimeRadiant.SyncStatus.RefetchingNeeded:
+                    return name;
```
amulware: I think we should return a constant string from resources here and not store the name in the database (we can maybe remove it from there later). Unless there is some point in having it in the database, but I don't see any.

PR #2452, opened 2018-06-13
Order the projects in select views

### LINKED ISSUE #1472: Grouped projects/tags/clients from the default workspace should be displayed first
... and all the other workspaces under it in some order (currently arbitrary, maybe it could be alphabetically by the name of the workspace?).

This is a suggestion for organizing the groups of tags/projects/clients in the selection views - I find it very inconvenient that I have to scroll down to look for the projects I use the most, which are obviously (?) in the default workspace.

Please consider this, @pe0ny and @amulware.

khoi:
Close #1472

I'm sorting the project so that default workspace always appear first in the list. The rest is ordered alphabetically.


Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
```diff
@@ -697,8 +701,11 @@ private IEnumerable<AutocompleteSuggestion> filterSuggestions(IEnumerable<Autoco
             IEnumerable<AutocompleteSuggestion> suggestions)
         {
             var firstSuggestion = suggestions.FirstOrDefault();
+            interactorFactory.GetDefaultWorkspace().Execute();
```
kornelijepetak: 👀 Where does the result of this go?

khoi: no where, I used this file as my playground to test out the code. Let me discard it

PR #2453, opened 2018-06-13
Use DTO to create new Client

Odrakir:
This PR is mainly so we can discuss.

The primary goal is to have a `Client` in Foundation, which has validation of its mandatory fields. This is not possible at the moment because we use that object to send to the repository the necessary data for creation as well.

Ideally, in my opinion, we should send a different interface to the repository, but that change is very complicated for me at the moment.

So here comes another idea. Make a `ClientDTO` that implements `IThreadSafeClient` and doesn't have validation (we could add some for mandatory fields for creation).

Tests currently fail because somehow they are creating `IDatabaseClient` instances without `Name`.

Please discuss if this makes sense.

### DIFF
--- a/Toggl.Foundation/DTOs/ClientDTO.cs
+++ b/Toggl.Foundation/DTOs/ClientDTO.cs
@@ -1,0 +1,31 @@
+﻿using System;
+using Toggl.Foundation.Models.Interfaces;
+using Toggl.PrimeRadiant;
+using Toggl.PrimeRadiant.Models;
+
+namespace Toggl.Foundation.DTOs
+{
+    public class ClientDTO : IThreadSafeClient
+    {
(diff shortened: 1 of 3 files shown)


amulware:
I agree with the overall idea, however: Since we're not using references when creating entities, the creation DTO should not have the references (to a workspace in this case), only ids. They should (in this case) only implement IClient and IDatabaseSyncable, so that they can be persisted (maybe this requires interfaces to be shuffled around a bit).

The DB then resolves the references and returns a model that has references as well.

In general, this is not a priority right now though. I don't mind discussion and/or creation of an issue but let's not spend too much time on this.

Odrakir:
There'll be another proposal to fix this for all models.

PR #2455, opened 2018-06-13
Refactor project model

Odrakir:
Part of #1711 

This is not it, discussion needed.

Removing Builders implies having a constructor with all parameters (for the `With` implementation). Some other languages have those constructors by default (for structs), but we could, maybe, use code generators for this.

And there's also the ugly constructor, that only takes some params, that's for creation. I still think we should have some other model for that.

I still have problems with validation and tests. I've commented it out because otherwise test fail.

### DIFF
--- a/Toggl.Foundation/Models/Project.cs
+++ b/Toggl.Foundation/Models/Project.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Net.Http.Headers;
 using Toggl.Foundation.Models.Interfaces;
 using Toggl.Multivac;
 using Toggl.Multivac.Models;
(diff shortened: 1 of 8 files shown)


Toggl.Foundation/Models/Project.cs
```diff
@@ -1,160 +1,156 @@
...
-            public bool IsDeleted { get; private set; }
+        public static Project Unsyncable(IProject entity, string errorMessage)
+            => new Project(entity, SyncStatus.SyncFailed, errorMessage);
 
-            public long? ClientId { get; private set; }
+        public static Project Dirty(IProject entity)
+            => new Project(entity, SyncStatus.SyncNeeded, null);
 
-            public bool Active { get; private set; } = true;
+        public class ProjectBuilder
```
amulware: Is this being used anywhere?

Toggl.Foundation/Sync/States/Pull/TryFetchInaccessibleProjectsState.cs
```diff
@@ -73,9 +73,8 @@ private IObservable<Unit> refetch(IGrouping<long, IThreadSafeProject> projectsTo
 
         private IObservable<IThreadSafeProject> updateAtValue(IThreadSafeProject project)
         {
-            var updatedProject = Project.Builder.From(project)
-                .SetAt(timeService.CurrentDateTime)
-                .Build();
+            var updatedProject = Project.From(project)
+                .With(timeService.CurrentDateTime);
```
amulware: It doesn't matter much right now, but I think for these With calls it's nice to name the parameters specifically. That'll be more forwards compatible.

amulware:
By 'looking fine' I mean: For now. We'll revisit things in the future about the creation constructor issue.

simonrozsival:
I don't have any problem with this - it looks nice, it removes a lot of code.

The larger problem is that we shouldn't have any of the builders. The only way it should be possible to create a Foundation model should be from a database model (just making it thread safe).

The builder cannot populate any of the relationships between the entities - it has just the IDs + sync status information, which is enough to put the data into the database. And the only reason why we need these builders is to create an object in Foundation so we can stick it into the data source which sticks it into the database repository and from there we get a real database object with all the relationships which we then convert into a real foundation/threadsafe model which we can then use.

The only real solution to this problem is to create DTOs which would be just `struct ClientDto : IClient, IDatabaseSyncable` and use these as input arguments for all the `Create`/`Update`/`BatchUpdate`/`Overwrite` methods of our data sources.

Creating constructors/factories for these DTOs should be trivial.

Odrakir:
I totally agree with @simonrozsival , but what's the plan then? Should we try to merge this PR and do those other changes in the future? Or is better to just let it be until we actually do those changes.

With my current PR adding validation is not easy (@amulware), because it breaks the tests. I can take a look at that, but everything I can do will probably be a patch until we can do the correct thing.

Odrakir:
There's another, better, solution coming.

PR #2457, opened 2018-06-13
Add sync diagram drawing tool

amulware:
## :wat:

This PR adds a ~~unit test to Foundation~~ _[a new console app project]_ that creates a file `sync-graph.gv` containing a serialised representation of our sync machine state tree graph that can be turned into an image by [Graphviz](http://www.graphviz.org/).

~~The file is generated every time the tests are run but do not fret.~~ It is very fast.

## How does it even work?

Simple:
- I added an `ITransitionConfigurator` interface so that I can inject a custom state machine configurator to record all transitions in the graph (into the actual production state machine factory!)
- Thanks to the `ISyncState` refactor in #2424 I have a list of all sync states as well after configuration
- I generate nodes from all states
- I use reflection on the `ISyncState` instances to find all `IStateResult(<T>)` properties which I turn into outgoing arcs and connect based on the transitions configured
- I find all dead ends and entry points and add 'fake' nodes and arcs for them
- Lastly, I serialise all nodes and arcs into the correct format for graphviz

Simple.

I also extract the names of state classes and state result properties and the types of transitions and use them as labels on the nodes in the graph to help us follow data more easily.

I also use different colours for entry points, dead ends and `InvalidTransitionState`s as visual aid.

## Remaining issues

First of all, the code is very complicated and not at all easy to understand. This could probably be improved ~~a lot~~ [a little (now that it's in a separate project and uses more smaller files)], but I'm not sure if that is worth it, ~~given that it is a test and won't be changed terribly often~~.

~~Secondly, the current code does not deal well with the new decorating sync-states we started to use, since there are now a lot of 'hidden' transitions. I'm not sure if it can be adapted to deal with these easily, or if we have to consider dropping the decorating states.~~ [We decided not to use decorators anymore, and there is a PR to remove them.]

Also, the graph is not laid out very well by graphviz without some further help from us. Again, I'm not sure if this is worth putting more effort in though.

## Yeah, but... pretty images?

Ok. This PR only generates a long text file with nodes and arcs. But for your convenience, I stuck that file into [Graphviz Online](https://dreampuf.github.io/GraphvizOnline/) and took a screenshot of part of the generated file.

screen shot 2018-06-13 at 17 07 11

This part of the tree represents pushing of preferences in the current implementation. I can already see a few things about this I might want changed, but the key point is that I would have never been able to find them without generating this visial aid. So.. 🎉 

## As a project

This is now a separate project instead of a test. It is built every time the unit tests are, but not run automatically. It can also be built manually with the `BuildSyncDiagramGenerator` cake target (also builds the dependencies, but nothing else).

To run it, there is the `GenerateSyncDiagram` cake target. This does nothing except for running it.

The output directory can be configured with the `SYNC_GRAPH_OUT_DIRECTORY` environment variable.

When this gets merged, I will update bitrise to run the generating cake flow after the unit tests and deploy the resulting DOT file so we can download it easily, for every single commit down the line. The separate cake flow shouldn't be an issue, it takes only a few seconds to run. The generator itself runs in well less than a second on my machine.

## Other changes

I did not make any meaningful changes to the DOT file generation when moving it to the new project, but I did spread the code across more files and add some debug printing, because that's nice.


Toggl.Foundation.Tests/Sync/SyncDiagramGenerator.cs
```diff
@@ -0,0 +1,308 @@
...
+                    break;
+                case Node.NodeType.InvalidTransitionState:
+                    attributes.Add(("color", "red"));
+                    break;
+            }
+
+            return attributes;
+        }
+
+        private void idNodes(List<Node> nodes)
```
kornelijepetak: Oh, dear. Is `id` a verb now? 😛 
At least it's not `idify`.

Toggl.Foundation.Tests/Sync/SyncDiagramGenerator.cs
```diff
@@ -0,0 +1,308 @@
...
+                    var edge = new Edge
+                    {
+                        From = node,
+                        To = stateNodes[state.State],
+                        Label = ""
+                    };
+                    edges.Add(edge);
+                }
+            }
+
```
kornelijepetak: Enough space to dance a Viennese Waltz here.

Toggl.Foundation.Tests/Sync/SyncDiagramGenerator.cs
amulware: According to https://www.merriam-webster.com/dictionary/id it is.

simonrozsival:
I like this very much. It would be very nice if we could somehow group states e.g., by entity in push sync and then display these using graphviz clusters - it would make the layout nicer.

_Thinking out loud:_ There must be a way of not-breaking-stuff-with-the-decorations - I will give this a thought today and hopefully come up with something (could it be solved by the clusters? 🤔)
_A question:_ would it be possible to install graphviz to Bitrise (so that it does not have to be installed on every build) and generate the SVG on nightly builds and post it to the slack channel? That would be so cool.

amulware:
We can't modify what comes installed on Bitrise machines (unless we ask them to change it), but in principle, we can just download graphviz for nightly runs no problem. I'll put this on my list for sometime down the line. I'm not really convinced we'll get much value out of it unless we can diff the graphs as well.

Toggl.Foundation.Tests/Sync/SyncDiagramGenerator.cs
```diff
@@ -0,0 +1,306 @@
...
+using Toggl.Foundation.Sync.States;
+using Toggl.PrimeRadiant;
+using Toggl.Ultrawave;
+using Xunit;
+
+namespace Toggl.Foundation.Tests.Sync
+{
+    public sealed class SyncDiagramGenerator
+    {
+        class Configurator : ITransitionConfigurator
```
willsb: Implicit accessibility modifiers are a no for me

Toggl.Foundation.Tests/Sync/SyncDiagramGenerator.cs
```diff
@@ -0,0 +1,306 @@
...
+        {
+            public List<object> AllDistinctStatesInOrder { get; } = new List<object>();
```
willsb: If this is distinct, y u no HashSet?

Toggl.Foundation.Tests/Sync/SyncDiagramGenerator.cs
```diff
@@ -0,0 +1,306 @@
...
+            private void addToListIfNew(object state)
+            {
+                if (AllDistinctStatesInOrder.Contains(state))
+                    return;
+
+                AllDistinctStatesInOrder.Add(state);
+            }
+        }
+
+        class Node
```
willsb: internal/public class pl0x

Toggl.Foundation.Tests/Sync/SyncDiagramGenerator.cs
```diff
@@ -0,0 +1,306 @@
...
+                DeadEnd = 2,
+                InvalidTransitionState = 3,
+            }
+
+            public string Id { get; set; }
+            public string Label { get; set; }
+            public NodeType Type { get; set; }
+        }
+
+        class Edge
```
willsb: y u hate access modifiers??

Toggl.Foundation.Tests/Toggl.Foundation.Tests.csproj
```diff
@@ -2,6 +2,7 @@
   <PropertyGroup>
     <TargetFramework>netcoreapp2.0</TargetFramework>
     <IsPackable>false</IsPackable>
+    <LangVersion>7.1</LangVersion>
```
willsb: Such bleeding edge, wow

Toggl.Foundation.Tests/Sync/SyncDiagramGenerator.cs
amulware: Because the order matters, as the name clearly implies. I could use a HashSet in addition, but given the numbers we are dealing with, and the circumstances in which we run this, I think that's a waste of code.

Toggl.Foundation.Tests/Sync/SyncDiagramGenerator.cs
amulware: Nested classes, don't judge my implementation details. You shouldn't be even looking at those. That's how that works, right?

amulware:
I'll think about splitting this out into a project. I agree that having another folder with all the tools projects would be awesome. But I also really like that this here gets run on every test run to make sure that nobody changes syncing code without making sure this here still works. I don't see a simple solution the ensure the latter if this is turned into a separate project right now.

amulware:
@simonrozsival @juanlaube @kornelijepetak 

What do you think about creating a separate project for this?
I'd probably end up sticking the different helper classes into additional files and clean it up just a tiny bit more, but wouldn't change the code itself too much, unless otherwise requested.

I'm also sort of waiting for @simonrozsival's word on decorated states and how we should best handle them.

simonrozsival:
I agree with creating a separate project for this.

My suggestion for the decorated states: create an `interface IComposedState<T> { T InternalState { get; } }` which would be used by the transitions analyzer to recursively find the bottom-most state and use this one to find the node it belongs to.

amulware:
Uuuhhh... a generic interface makes things annoying. But I'll see what I can do.

simonrozsival:
Having a separate project would allow as to run this just during the nightly build and post the link to the generated graph to the slack channel. That could be nice and it won't slow down our daily workflows.

amulware:
It doesn't take much time to run though, and I see no reason to not run it with every build. Don't merge sync machine changes that break the graph drawing.

simonrozsival:
@amulware maybe it doesn't have to be generic, I am just not sure right now what is the easiest solution. We can discuss this later in Slack if you want to.

amulware:
Note: We discussed internally that there are other problems with the decorating states, since they expose the internal state and thus aren't clean decorators, and it's just too easy to set them up wrong. We will remove them in favour of small transition-chained states.

PR #2460, opened 2018-06-14
Refactor complex event metrics

kornelijepetak:
If it was up to me, I would change all events into types. I like types. Types represent categories of things. Events are a thing. Methods are work, not things. 😛 

Anyways, an event class has to implement `ITrackableEvent`, and then

```cs
new RandomEvent(some, params, here)
    .TrackWith(analyticsService);
```
or
```cs
someObservable
    .Track(new RandomEvent(some, params, here), analyticsService)
    .Subscribe(...);
```
or
```cs
analyticsService.Track(new RandomEvent(some, params, here));
```

#### What's going on?
Basically, the first time tracking for an event type is requested, the lambda that gets stuff from the object into a dictionary gets compiled and cached. 

Then, the next time the same event is being tracked, the lambda from the cache is run.

If for some reason, compilation of lambda fails, the system will fall back to getting the properties from the object by simple reflection's `GetValue`.

#### Names and values
The default name for an event is its name, and the default name for a parameter is its name as well. However, `[DisplayName("OverriddenName")]` can be used on both event class or any of its parameters to output an event with any other name.

Values are calculated by calling `ToString` on every `(public|instance)` property of the class.

#### Performance
Basically, the first call (lambda creation and caching) to an event takes around `30ms` on my device (in debug mode)... Every subsequent call takes a `few microseconds` (it's just a normal call at that point). The first call is not that good (it's not bad either), but amortized over several calls it gets rather good.

#### iOS
iOS does not JIT. That means that `Reflection.Emit` is 💥 on iOS. However, I am not sure how much of Emit does `Expression.Lambda().Compile()` use, if any. This needs to be tested on iOS. As far as I could read, there are some fallbacks within mono that will do the job, but somebody who has an actual iOS device should try running this in a debug mode and see if it breaks. (Ping me if you want to pair-test this).

### DIFF
--- a/Toggl.Foundation.Tests/AnalyticsService/TrackableEventsTests.cs
+++ b/Toggl.Foundation.Tests/AnalyticsService/TrackableEventsTests.cs
@@ -1,0 +1,87 @@
+﻿using System;
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.Diagnostics;
+using FluentAssertions;
+using NSubstitute;
+using Toggl.Foundation.Analytics;
+using Xunit;
+
+namespace Toggl.Foundation.Tests.AnalyticsService
+{
+    public sealed class TrackableEventsTests
+    {
+        private const string OverriddenEventName = "Overridden";
+        private const string OverriddenParameterName = "CreationOrigin";
+
+        private class RandomEvent : ITrackableEvent
+        {
+            public int Ergalactic { get; } = 199;
+            public string Theory { get; } = "Quantum entanglement";
+            public bool George { get; } = true;
+            public DateTimeOffset Time { get; } = DateTimeOffset.Parse("2016-12-08 04:00");
+        }
+
+        [DisplayName(OverriddenEventName)]
+        private class StartTimeEntryEvent : ITrackableEvent
+        {
+            [DisplayName(OverriddenParameterName)]
+            public TimeEntryStartOrigin Origin { get; } = TimeEntryStartOrigin.Timer;
+
+            public bool IsDescriptionNotEmpty { get; } = true;
+            public bool HasProject { get; } = true;
+            public bool HasTask { get; } = false;
+            public int TagCount { get; } = 42;
+            public bool IsRunning { get; } = false;
+            public bool IsBillable { get; } = false;
+        }
+
+        public IAnalyticsService AnalyticsService { get; } = Substitute.For<IAnalyticsService>();
+
+        private RandomEvent createRandomEvent() => new RandomEvent();
+        private StartTimeEntryEvent createStartTimeEntryEvent() => new StartTimeEntryEvent();
+
+        public TrackableEventsTests()
+        {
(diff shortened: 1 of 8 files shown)


kornelijepetak:
Maybe useful for #2426 / #1919

juanlaube:
I'm not sure I like all the runtime magic :/
Considering iOS does AOT, and it's also an option in Android (we're not using it right now), I don't see this working correctly.
I'm in favor of (manually?) assembling this cache in compilation time.

kornelijepetak:
@Odrakir Dictionaries are not cached. The methods that produce the dictionary from the event instances are, simply to avoid multiple compilations.

kornelijepetak:
OK. So, about this... I am very sad 😢 

It will not work. The iOS thing, as feared, doesn't work. Even worse, it doesn't even throw on attempts at compile, but just swallows it and _probably_ turns into a **no-op** method... So it will not even hit the fallback way of doing things...

I am about to give up on this, but before I close the PR, I wanna talk about our options...

1) Simply keep on using the methods as now, but extend it to more and more parameters, as they come
2) Switch to types and use the `ITrackableEvent` as defined in this PR, but add to it a `Dictionary<string, string> ToEventParameters()` and a `string Name { get; }` so that this can be called from analytics... It's not that there'll be like 2000 of those...
3) Maybe even use T4 to generate `ToEventParameters` boilerplate as partial methods if we decide that we still want to be lazy.
4) ________________________________ (fill the blank)

In all honesty, I would most likely chose 2) if it was up to me, but since it's up to all of us, please speak...

Odrakir:
4. Use `ITrackableEvent` with a `toDictionary` method and let every event handle it? (we might be able to use some generic introspection) 

I'm sorry your PR is not working @kornelijepetak, you deserve a poem:
Roses are red
Violets are blue
Your PR is broken.

That's it, I'm not good at poetry.

kornelijepetak:
@Odrakir your 4. is the same as my 2.
Did I explain it poorly?

Odrakir:
No, sorry, I read poorly.

daividssilverio:
I'm sad it doesn't work on iOS :c
I'd go with option 2 as well. Seems easy to understand.

Odrakir:
@kornelijepetak Are you going to refactor this PR into option number 2?

kornelijepetak:
I would, but I need a consensus from the rest of the team.

juanlaube:
I like option 2, it's not like we have thousands of event types...

amulware:
I'm not convinced by any of the options (and neither by the current implementation).

The currently favoured option 2 still leaves dictionary manipulations which are easy places to mess up across a lot of different files.

Maybe we'll move in that direction, at least for events with many parameters or with a complex DTO as parameter, or a similar one in the future, but let's stick to what we have right now for the time being instead of holding back development on other issues.

kornelijepetak:
Superseded by #2485

PR #2464, opened 2018-06-14
Add means to bind lists on iOS

willsb:
This is similar to the adapter structure we have for Android. It will allow us to bind simple lists without any special classes and will serve as base for more complex cases.

Squash whenever


Odrakir:
You probably can convince me easily, but for now, I don't see the point of all this. You constrain your cells to that inherit from that type, be configured that way and with that specific identifier. And in the end I think you don't accomplish much more than what is done in `ListTableViewSource` with much less code...

Toggl.Daneel/ViewSources/SimpleTableViewSource.cs
```diff
@@ -0,0 +1,25 @@
...
+            TableView.RegisterClassForCellReuse(typeof(TCell), nameof(TCell));
+        }
+
+        protected SimpleTableViewSource(IntPtr handle)
+            : base(handle)
+        {
+        }
+
+        protected override BaseTableCell<TItem> GetOrCreateCellFor(UITableView tableView, NSIndexPath indexPath, object item)
+            => TableView.DequeueReusableCell(nameof(TCell)) as BaseTableCell<TItem>;
```
kornelijepetak: Same here.

Toggl.Daneel/ViewSources/SimpleTableViewSource.cs
```diff
@@ -0,0 +1,25 @@
...
+
+namespace Toggl.Daneel.ViewSources
+{
+    public class SimpleTableViewSource<TItem, TCell> : BaseTableViewSource<TItem>
+        where TCell : BaseTableCell<TItem>
+    {
+        protected SimpleTableViewSource(UITableView tableView)
+            : base(tableView)
+        {
+            TableView.RegisterClassForCellReuse(typeof(TCell), nameof(TCell));
```
kornelijepetak: Maybe I understood incorrectly what you are trying to achieve, but `nameof(TCell)` will always return the string `TCell`, regardless of the actual generic type used.
If you wanted to do it by type name, I am afraid you have to use `typeof(TCell).Name`.

Odrakir:
As discussed in Slack, I'm happy with something like `ListTableViewSource`, I don't think we need the inheritance and indirection at this point. Maybe it's useful further down the road if we add more types of data sources, but for now, I would keep it simple.

That said,  I see the use of inheritance for this kind of patterns is very common in the code base, so if more people agree on this PR, I'm ok with it being merged.

Toggl.Daneel/ViewSources/SimpleTableViewSource.cs
willsb: We don't need the actual cell name. All we need is a string. I could use "Care bear hugs", but that's not as professional

Odrakir: Even so I wouldn't leave it like it is. It will potentially lead to bugs if we ever want to evolve this into a tv source with two types of cells.

willsb: If you wanna have multiple cells, you extend from base. Simple is for single cells. Maybe naming could improve? 🤔

arturs105: I agree with Kornelije on this one. Although using `"TCell"` as the identifier would probably work just fine, having the actual type name as the identifier is better.

kornelijepetak: In that case, I would much rather have `Care bear hugs` that looks as if it's not important, than something that is related, unneeded and looks wrong and like a bug.

Toggl.Daneel/ViewSources/ListTableViewSource.cs
```diff
@@ -1,31 +1,39 @@
...
+        private readonly IImmutableList<TModel> items;
 
-        public ListTableViewSource(IEnumerable<TModel> items, string cellIdentifier, Func<UITableViewCell, TModel, TCell> configureCell)
+        public Func<TModel, Task> OnItemTapped { get; set; }
+
+        public ListTableViewSource(IImmutableList<TModel> items, string cellIdentifier)
         {
-            this.items = items.ToList();
+            this.items = items;
             this.cellIdentifier = cellIdentifier;
```
Odrakir: Does it make sense to move the identifier to a static property on the cell?

willsb: Sadly there's no way to constraint it :/

Odrakir: I'm not sure what you mean... 🤔 

Having a property for the cell would allow us to also use it when registering it.

PR #2466, opened 2018-06-14
Remove Giskard-like dependency from VM by RXFactorization

### LINKED ISSUE #2429: Refactor EditTimeEntryViewModel to remove the Giskard dependency

kornelijepetak:
Closes #2429.


Toggl.Giskard/Activities/StartTimeEntryActivity.ViewBinds.cs
```diff
@@ -0,0 +1,23 @@
...
+    public partial class StartTimeEntryActivity
+    {
+        private TextView durationLabel;
+
+        private void initializeViews()
+        {
+            durationLabel = FindViewById<TextView>(StartTimeEntryDurationText);
+        }
+
+        private void setupBindings()
```
daividssilverio: 🤔 I feel like those bindings should be in the other part of the Activity.
Although the name of the view indicates that the bindings should be placed here, I was under the impression that this `*.ViewBinds.cs` would basically contain only `FindViewById` stuff.

I'm not against this approach, but it would be nice to keep the consistency. @willsb @juanlaube what do you think?

daividssilverio:
Oh yeah, basically by ` Giskard-like dependency`, you mean all those parameters that were just strings?

Toggl.Giskard/Activities/StartTimeEntryActivity.ViewBinds.cs
willsb: I'm with you, @daividssilverio. This is just to remove boilerplate from the Activity, not to contain any logic

Toggl.Giskard/Activities/EditTimeEntryActivity.ViewBinds.cs
```diff
@@ -0,0 +1,41 @@
...
+        private View durationArea;
+
+        private void initializeViews()
+        {
+            startTimeArea = FindViewById(EditTimeLeftPart);
+            stopTimeArea = FindViewById(EditTimeRightPart);
+            durationArea = FindViewById(EditDuration);
+        }
+
+        private void setupBindings()
```
daividssilverio: I didn't see this before, but it follows the same idea from `StartTimeEntryActivity.ViewBinds.cs`, this file should only contain the Activity related boilerplate, and no other logic.

Toggl.Giskard/Activities/EditTimeEntryActivity.cs
```diff
@@ -1,25 +1,22 @@
...
-using TextView = Android.Widget.TextView;
 
 namespace Toggl.Giskard.Activities
 {
     [MvxActivityPresentation]
     [Activity(Theme = "@style/AppTheme")]
-    public sealed class EditTimeEntryActivity : MvxAppCompatActivity<EditTimeEntryViewModel>
+    public sealed partial class EditTimeEntryActivity : MvxAppCompatActivity<EditTimeEntryViewModel>, IReactiveBindingHolder
     {
+        public CompositeDisposable DisposeBag { get; private set; } = new CompositeDisposable();
```
daividssilverio: Correct me if I'm wrong, but we still need to explicitly dispose this DisposeBag overriding `Dispose` if we are not using the new `ReactiveActivity`, right?

Toggl.Giskard/Activities/StartTimeEntryActivity.cs
```diff
@@ -9,15 +9,19 @@
...
 namespace Toggl.Giskard.Activities
 {
     [MvxActivityPresentation]
     [Activity(Theme = "@style/AppTheme",
               ScreenOrientation = ScreenOrientation.Portrait,
               ConfigurationChanges = ConfigChanges.Orientation | ConfigChanges.ScreenSize)]
-    public sealed class StartTimeEntryActivity : MvxAppCompatActivity<StartTimeEntryViewModel>
+    public sealed partial class StartTimeEntryActivity : MvxAppCompatActivity<StartTimeEntryViewModel>, IReactiveBindingHolder
     {
+        public CompositeDisposable DisposeBag { get; private set; } = new CompositeDisposable();
```
daividssilverio: Same here, we should explicitly dispose it.

PR #2467, opened 2018-06-14
:sparkles: Add NoWorkspaceViewModel

juanlaube:
Closes #2334 

Still WIP because some tests are pending, like testing the `IsLoading` property. I'd like to get feedback though.

This is kind of blocked also because the logic to detect when there are no workspaces doesn't work correctly, backend returns a 403 on `GET /me/workspaces/` when there are no workspaces.

### DIFF
--- a/Toggl.Foundation.MvvmCross/ViewModels/NoWorksaceViewModel.cs
+++ b/Toggl.Foundation.MvvmCross/ViewModels/NoWorksaceViewModel.cs
@@ -1,0 +1,98 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reactive;
+using System.Reactive.Disposables;
+using System.Reactive.Linq;
+using System.Reactive.Subjects;
+using System.Threading.Tasks;
+using MvvmCross.Core.Navigation;
+using MvvmCross.Core.ViewModels;
+using Toggl.Foundation.DataSources;
+using Toggl.Foundation.Models.Interfaces;
+using Toggl.Multivac;
+using Toggl.Multivac.Extensions;
+using Toggl.Multivac.Models;
+
+namespace Toggl.Foundation.MvvmCross.ViewModels
+{
+    public class NoWorksaceViewModel : MvxViewModel
+    {
+        private IMvxNavigationService navigationService;
+        private ITogglDataSource dataSource;
+
+        private CompositeDisposable disposeBag = new CompositeDisposable();
+        private Subject<Unit> syncSubject = new Subject<Unit>();
+        private Subject<Unit> getWorkspacesSubject = new Subject<Unit>();
+
+        public bool IsLoading { get; private set; }
+
+        public IMvxCommand TryAgainCommand { get; }
+
+        public IMvxAsyncCommand CreateWorkspaceCommand { get; }
+
+        public NoWorksaceViewModel(
+            IMvxNavigationService navigationService, 
+            ITogglDataSource dataSource
+        )
+        {
+            Ensure.Argument.IsNotNull(navigationService, nameof(navigationService));
+            Ensure.Argument.IsNotNull(dataSource, nameof(dataSource));
+
+            this.navigationService = navigationService;
+            this.dataSource = dataSource;
+
+            disposeBag.Add(
+                getWorkspacesSubject
+                    .AsObservable()
+                    .Do(_ => IsLoading = true)
+                    .SelectMany(_ => dataSource.Workspaces.GetAll())
+                    .Subscribe(onWorkspaces)
+            );
+
+            disposeBag.Add(
+                syncSubject
+                    .AsObservable()
+                    .SelectMany(dataSource.SyncManager.ForceFullSync())
+                    .SelectUnit()
+                    .Subscribe(getWorkspacesSubject.OnNext)
+            );
+
+            TryAgainCommand = new MvxCommand(tryAgain);
+            CreateWorkspaceCommand = new MvxAsyncCommand(createWorkspace);
+        }
+
+        private void tryAgain()
+            => syncSubject.OnNext(Unit.Default);
+
+        private async Task createWorkspace()
+        {
+            IsLoading = true;
+            var workspaces = await dataSource.Workspaces.GetAll();
+            if (workspaces.Any())
+            {
+                IsLoading = false;
+                close();
+                return;
+            }
+
+            var user = await dataSource.User.Current;
+            var workspace = await dataSource.Workspaces.Create($"{user.Fullname}'s Workspace");
+            syncSubject.OnNext(Unit.Default);
+        }
+
+        private void onWorkspaces(IEnumerable<IThreadSafeWorkspace> workspaces)
+        {
+            IsLoading = false;
+            if (workspaces.Any())
+            {
+                close();
+            }
+        }
+
+        private void close()
+        {
+            navigationService.Close(this);
+        }
+    }
+}
--- a/Toggl.Foundation.MvvmCross/Services/ErrorHandlingService.cs
+++ b/Toggl.Foundation.MvvmCross/Services/ErrorHandlingService.cs
@@ -63,7 +63,7 @@ namespace Toggl.Foundation.MvvmCross.Services
         {
             if (error is NoWorkspaceException)
             {
-                // Show no workspace UI here. Also don't forget to call `DataSource.StartSyncing` when the user creates the workspace
+                navigationService.Navigate<NoWorksaceViewModel>();
                 return true;
             }
 
--- a/Toggl.Foundation.Tests/MvvmCross/ViewModels/NoWorkspaceViewModelTests.cs
+++ b/Toggl.Foundation.Tests/MvvmCross/ViewModels/NoWorkspaceViewModelTests.cs
@@ -1,0 +1,78 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reactive.Linq;
+using System.Threading.Tasks;
+using FluentAssertions;
+using NSubstitute;
+using Toggl.Foundation.Models.Interfaces;
+using Toggl.Foundation.MvvmCross.ViewModels;
+using Toggl.Foundation.Tests.Generators;
+using Xunit;
+
+namespace Toggl.Foundation.Tests.MvvmCross.ViewModels
+{
+    public sealed class NoWorkspaceViewModelTests
+    {
+        public abstract class NoWorkspaceViewModelTest : BaseViewModelTests<NoWorksaceViewModel>
+        {
+            protected override NoWorksaceViewModel CreateViewModel()
+                => new NoWorksaceViewModel(NavigationService, DataSource);
+        }
+
+        public sealed class TheConstructor : NoWorkspaceViewModelTest
+        {
+            [Theory, LogIfTooSlow]
+            [ClassData(typeof(TwoParameterConstructorTestData))]
+            public void ThrowsIfAnyOfTheArgumentsIsNull(bool useNavigationService, bool useDataSource)
+            {
+                var navigationService = useNavigationService ? NavigationService : null;
+                var dataSource = useDataSource ? DataSource : null;
+
+                Action tryingToConstructWithEmptyParameters =
+                    () => new NoWorksaceViewModel(navigationService, dataSource);
+
+                tryingToConstructWithEmptyParameters.Should().Throw<ArgumentNullException>();
+            }
+        }
+
+        public sealed class TheTryAgainCommand : NoWorkspaceViewModelTest
+        {
+            [Fact, LogIfTooSlow]
+            public async Task ClosesWhenAnotherWorkspaceIsFetched()
+            {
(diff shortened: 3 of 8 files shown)


Toggl.Foundation.MvvmCross/ViewModels/NoWorksaceViewModel.cs
```diff
@@ -0,0 +1,98 @@
...
+            ITogglDataSource dataSource
+        )
+        {
+            Ensure.Argument.IsNotNull(navigationService, nameof(navigationService));
+            Ensure.Argument.IsNotNull(dataSource, nameof(dataSource));
+
+            this.navigationService = navigationService;
+            this.dataSource = dataSource;
+
+            disposeBag.Add(
```
daividssilverio: What about using `.disposedBy(disposeBag)` instead? less lines, less indentation. Looks nice.

Toggl.Foundation.MvvmCross/ViewModels/NoWorksaceViewModel.cs
```diff
@@ -0,0 +1,98 @@
...
+                getWorkspacesSubject
+                    .AsObservable()
+                    .Do(_ => IsLoading = true)
+                    .SelectMany(_ => dataSource.Workspaces.GetAll())
+                    .Subscribe(onWorkspaces)
+            );
+
+            disposeBag.Add(
```
daividssilverio: Here too.

Toggl.Foundation.MvvmCross/ViewModels/NoWorksaceViewModel.cs
```diff
@@ -0,0 +1,98 @@
...
+        )
+        {
+            Ensure.Argument.IsNotNull(navigationService, nameof(navigationService));
+            Ensure.Argument.IsNotNull(dataSource, nameof(dataSource));
+
+            this.navigationService = navigationService;
+            this.dataSource = dataSource;
+
+            disposeBag.Add(
+                getWorkspacesSubject
```
simonrozsival: I don't think we need this subject - all this logic can be simply chained to the other observable, right?

Toggl.Foundation.MvvmCross/ViewModels/NoWorksaceViewModel.cs
```diff
@@ -0,0 +1,98 @@
...
+                    .AsObservable()
+                    .Do(_ => IsLoading = true)
+                    .SelectMany(_ => dataSource.Workspaces.GetAll())
+                    .Subscribe(onWorkspaces)
+            );
+
+            disposeBag.Add(
+                syncSubject
```
simonrozsival: Also I don't think this should be a subject - we can use the `dataSource.SyncManager.ProgressObservable.Where(progress => progress == SyncProgress.Synced)` observable and not worry about any imperative `OnNext` in this VM.

Toggl.Foundation.MvvmCross/ViewModels/NoWorksaceViewModel.cs
```diff
@@ -0,0 +1,98 @@
...
+            if (workspaces.Any())
+            {
+                IsLoading = false;
+                close();
+                return;
+            }
+
+            var user = await dataSource.User.Current;
+            var workspace = await dataSource.Workspaces.Create($"{user.Fullname}'s Workspace");
+            syncSubject.OnNext(Unit.Default);
```
simonrozsival: If you change the observables, here you can just force full sync.

Toggl.Foundation.MvvmCross/ViewModels/NoWorksaceViewModel.cs
```diff
@@ -0,0 +1,98 @@
...
+                    .SelectUnit()
+                    .Subscribe(getWorkspacesSubject.OnNext)
+            );
+
+            TryAgainCommand = new MvxCommand(tryAgain);
+            CreateWorkspaceCommand = new MvxAsyncCommand(createWorkspace);
+        }
+
+        private void tryAgain()
+            => syncSubject.OnNext(Unit.Default);
```
simonrozsival: If you change the observables, here you can just force full sync.

Toggl.Foundation.MvvmCross/ViewModels/NoWorksaceViewModel.cs
```diff
@@ -0,0 +1,98 @@
...
+using MvvmCross.Core.ViewModels;
+using Toggl.Foundation.DataSources;
+using Toggl.Foundation.Models.Interfaces;
+using Toggl.Multivac;
+using Toggl.Multivac.Extensions;
+using Toggl.Multivac.Models;
+
+namespace Toggl.Foundation.MvvmCross.ViewModels
+{
+    public class NoWorksaceViewModel : MvxViewModel
```
amulware: `Worksace` typo

arturs105: Y u no `seal`?

Toggl.Foundation.MvvmCross/ViewModels/NoWorkspaceViewModel.cs
```diff
@@ -0,0 +1,74 @@
...
+    public sealed class NoWorkspaceViewModel : MvxViewModel
+    {
+        private IMvxNavigationService navigationService;
+        private ITogglDataSource dataSource;
+
+        public bool IsLoading { get; private set; }
+
+        public IMvxAsyncCommand TryAgainCommand { get; }
+
+        public IMvxAsyncCommand CreateWorkspaceCommand { get; }
```
daividssilverio: Before reading the related issues, I thought this would start a new popup, window, such that the user could fill the workspace name and whatever else he would need.
This is not the case, this creates a default/fallback with the user name set to it. It would be nice if you could give a better name for this command. `CreateDefaultWorkspace`, `CreateFallbackWorkspace`, `CreateWorkspaceWithUserName`, something like that.

Toggl.Foundation.MvvmCross/ViewModels/NoWorkspaceViewModel.cs
```diff
@@ -0,0 +1,79 @@
...
+        }
+
+        private async Task tryAgain()
+        {
+            IsLoading = true;
+
+            var workspaces = await dataSource
+                .SyncManager
+                .ForceFullSync()
+                .SelectMany(dataSource.Workspaces.GetAll());
```
simonrozsival: You still need to wait for the sync to completely finish - add `Where(state => state == SyncState.Sleep)`

Toggl.Foundation.MvvmCross/ViewModels/NoWorkspaceViewModel.cs
```diff
@@ -0,0 +1,79 @@
...
+        private async Task createWorkspace()
+        {
+            IsLoading = true;
+
+            var workspaces = await dataSource.Workspaces.GetAll();
+            if (!workspaces.Any())
+            {
+                var user = await dataSource.User.Current;
+                await dataSource.Workspaces.Create($"{user.Fullname}'s Workspace");
+            }
```
simonrozsival: You forgot to push the created workspace here - await PushSync here + call `dataSource.StartSyncing` to re-enable background syncing. Maybe it would be enough to just await `StartSyncing`.

juanlaube: @amulware can clarify here, but from the comments in the issue, we won't push immediately, instead mark the workspace as needs syncing and will be pushed later.

Toggl.Foundation.MvvmCross/ViewModels/NoWorkspaceViewModel.cs
juanlaube: I will look for a better name, notice that `default workspace != only one workspace` though.

Toggl.Foundation.MvvmCross/ViewModels/NoWorkspaceViewModel.cs
amulware: I see no point in waiting for the push. The user now has a workspace and can start using the app. Sure, we have to start syncing up again and trigger a full sync, but that's all that should be needed, and the actual sync can happen while the user is back in the app.

I'll be writing some more detailed docs on how this whole thing should work later this week btw, just for reference.

arturs105:
What if you made this view model rx-binding-friendly?  🤔  Wouldn't have to refactor this later.

simonrozsival:
BTW is there some set of steps how to get to a state when the user has no workspace but his/her account is not closed?

juanlaube:
Yes, I will try to write the steps today, however that won't navigate to this viewmodel because the endpoint returns a 403 and the exception is handling before the no workspaces validation.

simonrozsival:
We could change the detection algorithm and it would catch the 403 - it woudl be very easy, but if backend will eventually change this to a different behavior, I guess we shouldn't do that.

simonrozsival:
@juanlaube Did you make any progress on this issue in the last few days?

simonrozsival:
What is the state of this PR now?

simonrozsival:
You don't have to wait for my approval.

Toggl.Foundation.MvvmCross/ViewModels/NoWorkspaceViewModel.cs
```diff
@@ -0,0 +1,72 @@
...
+                close();
+            }
+        }
+
+        public async Task CreateWorkspaceWithDefaultName()
+        {
+            isLoading.OnNext(true);
+
+            var workspaces = await dataSource.Workspaces.GetAll();
+            if (!workspaces.Any())
```
amulware: Why this check at all?

Toggl.Foundation.MvvmCross/ViewModels/NoWorkspaceViewModel.cs
```diff
@@ -0,0 +1,72 @@
...
+            {
+                var user = await dataSource.User.Current;
+                await dataSource.Workspaces.Create($"{user.Fullname}'s Workspace");
+            }
+
+            isLoading.OnNext(false);
+            close();
+        }
```
amulware: Don't we need to recreate the sync manager, since it was frozen? And trigger a full-sync, unless it does so automatically (it might, not sure).

Toggl.Foundation.MvvmCross/ViewModels/NoWorkspaceViewModel.cs
```diff
@@ -0,0 +1,72 @@
...
+            this.dataSource = dataSource;
+        }
+
+        public async Task TryAgain()
+        {
+            isLoading.OnNext(true);
+
+            var workspaces = await dataSource
+                .SyncManager
+                .ForceFullSync()
```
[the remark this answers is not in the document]
juanlaube: Ok, this was confusing but the sync manager won't freeze with the no-workspace exception.
It will pass the exception to the progress observable and return before calling `.Freeze()`.
The progress observer in the data source will pass the exceptions to the error handling service.

Toggl.Foundation.MvvmCross/ViewModels/NoWorkspaceViewModel.cs
juanlaube: This is not needed anymore, I don't remember why this was here in the first place 🙂

Toggl.Foundation.MvvmCross/ViewModels/NoWorkspaceViewModel.cs
juanlaube: Given the sync manager won't freeze due to the no-workspace exception, we don't need this.

Toggl.Foundation.MvvmCross/ViewModels/NoWorkspaceViewModel.cs
```diff
@@ -0,0 +1,72 @@
...
+                .Where(state => state == SyncState.Sleep)
+                .SelectMany(dataSource.Workspaces.GetAll());
```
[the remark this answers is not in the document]
juanlaube: Right now, it will throw another exception and we need to consider that to avoid navigating multiple times to this view model. I think we can improve that in another pull request.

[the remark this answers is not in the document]
amulware: What happens when it navigates to the same view model again? Will it have multiple overlapping popups, or break otherwise?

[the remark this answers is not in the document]
juanlaube: With what we have right now, we need to control that before navigating. If this is already presented, we cannot navigate again.
I think we should freeze the sync manager on the no-workspace exception and recreate it on the try again command, but that should be for another PR.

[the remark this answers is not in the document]
juanlaube: I will wait for you to see this in case there are more questions, otherwise, I merge at dawn 😄

PR #2469, opened 2018-06-15
Make report pie chart grouping smart

### LINKED ISSUE #2439: Make report pie chart grouping smart
~~Turns out our users aren't liking the grouping, even for the pie chart. We're already not showing labels for small slices, so there isn't a terrible amount of visual noise, and we might as well go back to showing all project colours.~~

~~You win some, you lose some.~~

Surprise! We'll keep grouping, after all. However, we'll make it a lot smarter to keep it from grouping too many small projects into a huge 'other' slice.

The algorithm we will use for this, as developed and used by frontend, is as follows.

* For all possible displayed segments:
 * If a segment would be greater than 5% on the final pie chart it definitely gets its own segment
 * If a segment would be less than 1% on the final pie chart, it definitely gets put into the Other category
 * For the other segments, they are put into Other, smallest first, until we run out of them or Other becomes 5% *or over* on the final pie chart
* If Other is too small (but not empty) make it at least 1% on the final pie chart
* If, actually, there was only 1 item in Other, change the name & color (except how much space it takes up on the final chart! I.e. it remains at least 1%) of Other to be that original item.

Unless users have a lot of <1% projects, this will result in an other-segment of at most 10%, which is good.

Please remember to stuck the 1% and 5% thresholds into constants so we can easily change them should the need arise.

sofokli1:
Closes #2439

Since it is hard to test in practice, suggestions for more unit test are welcome.

🦑 whenever


Toggl.Foundation.MvvmCross/ViewModels/ReportsViewModel.cs
```diff
@@ -254,26 +254,85 @@ private void onPreferencesChanged(IThreadSafePreferences preferences)
...
         {
-            var otherProjects = segments.Where(segment => segment.Percentage < minimumPieChartSegmentPercentage).ToList();
-            if (otherProjects.Count <= 1 || otherProjects.Count == segments.Count)
+            var aboveStandAloneThreasholdSegments = segments
+                .Where(segment => segment.Percentage >= minimumPieChartSegmentPercentageToBeOnItsOwn)
+                .ToList();
+            
+            var otherProjectsCandidates = segments
+                .Where(segment => segment.Percentage < minimumPieChartSegmentPercentageToBeOnItsOwn)
+                .ToList();
```
amulware: You could do the above two at the same time using the `GroupBy` operator. I think that might be nicer to read and make obviously sure that all segments are included in one of the two lists.

Toggl.Foundation.MvvmCross/ViewModels/ReportsViewModel.cs
```diff
@@ -254,26 +254,85 @@ private void onPreferencesChanged(IThreadSafePreferences preferences)
...
+
+            var finalOtherProjects = otherProjectsCandidates.Where(segment => segment.Percentage < maximumPieChartSegmentPercentageToEndUpInOther).ToList();
+
+            otherProjectsCandidates = otherProjectsCandidates.Except(finalOtherProjects)
```
amulware: Maybe introduce another variable like `remainingOtherProjectCandidates`?

Toggl.Foundation.MvvmCross/ViewModels/ReportsViewModel.cs
```diff
@@ -254,26 +254,85 @@ private void onPreferencesChanged(IThreadSafePreferences preferences)
...
+
+            return aboveStandAloneThreasholdSegments
+                .Append(lastSegment)
                 .ToList()
                 .AsReadOnly();
         }
+
+        private float percentageOf(List<ChartSegment> list)
+        {
+            if (list.Any())
```
amulware: The Any check is not needed I think. `Sum` should return 0 for empty lists.

Toggl.Foundation.MvvmCross/ViewModels/ReportsViewModel.cs
```diff
@@ -254,26 +254,85 @@ private void onPreferencesChanged(IThreadSafePreferences preferences)
...
+
+            var finalOtherProjects = otherProjectsCandidates.Where(segment => segment.Percentage < maximumPieChartSegmentPercentageToEndUpInOther).ToList();
+
+            otherProjectsCandidates = otherProjectsCandidates.Except(finalOtherProjects)
+                .OrderBy(segment => segment.Percentage)
+                .ToList();
+
+            foreach (var segment in otherProjectsCandidates)
+            {
+                if (percentageOf(finalOtherProjects) + segment.Percentage <= 5)
```
amulware: inverting the condition could reduce nesting, and remove the need for `else`.

amulware: Also please introduce a named constant for the '5'.

amulware: Aaaalso, I think according to the specs, we perform the check _after_ adding the segment, so that the Other segment can become larger than 5%. With this implementation, that would not happen I think.

Toggl.Foundation.MvvmCross/ViewModels/ReportsViewModel.cs
```diff
@@ -254,26 +254,85 @@ private void onPreferencesChanged(IThreadSafePreferences preferences)
...
 
+        private const float minimumPieChartSegmentPercentageToBeOnItsOwn = 5f;
+        private const float maximumPieChartSegmentPercentageToEndUpInOther = 1f;
+        private const float minimumOtherChartSegmentDisplayPercentage = 1f;
+
         private IReadOnlyList<ChartSegment> groupSegments()
         {
-            var otherProjects = segments.Where(segment => segment.Percentage < minimumPieChartSegmentPercentage).ToList();
-            if (otherProjects.Count <= 1 || otherProjects.Count == segments.Count)
+            var aboveStandAloneThreasholdSegments = segments
```
amulware: Typo: `threshold`

Toggl.Foundation.MvvmCross/ViewModels/ReportsViewModel.cs
```diff
@@ -254,26 +254,85 @@ private void onPreferencesChanged(IThreadSafePreferences preferences)
...
+                }
+            }
+
+            if (!finalOtherProjects.Any())
+            {
                 return segments;
+            }
+
+            var leftOutOfOther = otherProjectsCandidates.Except(finalOtherProjects).ToList();
+            aboveStandAloneThreasholdSegments.AddRange(leftOutOfOther);
```
amulware: Could maybe use another variable as well.

amulware: And then combined with the line below maybe.

sofokli1:
@juanlaube there are cases where `other` should be forced to be 1% (even thought it could be 0.5%) so not sure how to achieve 100% before and after.

> If Other is too small (but not empty) make it at least 1% on the final pie chart

juanlaube:
That makes sense, my only concern is we left something bigger outside, but if we have enough test cases it's not a problem.

Toggl.Foundation.MvvmCross/ViewModels/ReportsViewModel.cs
```diff
@@ -254,26 +254,84 @@ private void onPreferencesChanged(IThreadSafePreferences preferences)
...
+                var singleSmallSegment = finalOtherProjects.First();
+                lastSegment = new ChartSegment(
+                    singleSmallSegment.ProjectName,
+                    string.Empty,
+                    singleSmallSegment.Percentage >= minimumOtherChartSegmentDisplayPercentage ? singleSmallSegment.Percentage : minimumOtherChartSegmentDisplayPercentage,
+                    finalOtherProjects.Sum(segment => (float)segment.TrackedTime.TotalSeconds),
+                    finalOtherProjects.Sum(segment => segment.BillableSeconds),
+                    singleSmallSegment.Color,
+                    DurationFormat);
+            } else {
```
kornelijepetak: `else` and braces each on new line.

Toggl.Foundation.MvvmCross/ViewModels/ReportsViewModel.cs
```diff
@@ -254,26 +254,84 @@ private void onPreferencesChanged(IThreadSafePreferences preferences)
...
-                .Append(otherSegment)
+            return onTheirOwnSegments
+                .Append(lastSegment)
                 .ToList()
                 .AsReadOnly();
         }
+
+        private float percentageOf(List<ChartSegment> list)
+        {
+            return list.Select(segment => segment.Percentage).Sum();
```
kornelijepetak: Just to spite @Odrakir , please turn this into an expression bodied member.
Also, the `Sum`  LINQ operator has a selector in itself, so you can shorten this to:

```
private float percentageOf(List<ChartSegment> list)
    => list.Sum(segment => segment.Percentage);
```

Toggl.Foundation.MvvmCross/ViewModels/ReportsViewModel.cs
```diff
@@ -261,12 +261,16 @@ private void onPreferencesChanged(IThreadSafePreferences preferences)
 
         private IReadOnlyList<ChartSegment> groupSegments()
         {
-            var aboveStandAloneThresholdSegments = segments
-                .Where(segment => segment.Percentage >= minimumPieChartSegmentPercentageToBeOnItsOwn)
+            var groupedData = segments.GroupBy(segment => segment.Percentage >= minimumPieChartSegmentPercentageToBeOnItsOwn);
+
+            var aboveStandAloneThresholdSegments = groupedData
```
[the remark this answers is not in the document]
sofokli1: if i use `CommonFunctions.Identity` i have a problem in `calendarViewModel.SelectedDateRangeObservable.Subscribe(` with this error `mobileapp/Toggl.Foundation.MvvmCross/ViewModels/ReportsViewModel.cs(63,63): Error CS0121: The call is ambiguous between the following methods or properties: 'System.ObservableExtensions.Subscribe<T>(System.IObservable<T>, System.Action<T>)' and 'Toggl.Multivac.Extensions.ReactiveExtensions.Subscribe<T>(System.IObservable<T>, System.Action<System.Exception>)' (CS0121) (Toggl.Foundation.MvvmCross)`

Not sure how to handle this. Any suggestion is welcome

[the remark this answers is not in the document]
kornelijepetak: Oh, yes. The problem is that in order for `CommonFunctions` to work, you had to include the `Toggl.Multivac.Extensions` namespace which in turn made another `Subscribe` method visible, which is what caused the ambiguity in your case, as there are now multiple that satisfy the callsite.
The easiest solution is to add:

`using CommonFunctions = Toggl.Multivac.Extensions.CommonFunctions;` to the usings which will effectively pull the name CommonFunction from the correct namespace without making other members of the namespace visible.

Toggl.Foundation.MvvmCross/ViewModels/ReportsViewModel.cs
```diff
@@ -254,26 +254,84 @@ private void onPreferencesChanged(IThreadSafePreferences preferences)
...
+                .Where(group => group.Key == true)
+                .SelectMany(g => g.ToList())
+                .ToList();
+
+            var otherProjectsCandidates = groupedData
+                .Where(group => group.Key == false)
+                .SelectMany(g => g.ToList())
+                .ToList();
+
+            var finalOtherProjects = otherProjectsCandidates.Where(segment => segment.Percentage < maximumPieChartSegmentPercentageToEndUpInOther).ToList();
```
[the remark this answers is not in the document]
kornelijepetak: What I meant is:

```c#
= otherProjectCandidates
    .Where(...)
    .ToList();
```

and also down there 

```C#
otherProjectsCandidates
    .Except(finalOtherProjects)
    .OrderBy(segment => segment.Percentage)
    .ToList();
```

PR #2479, opened 2018-06-15
Track when the user doesn't have a default workspace

### LINKED ISSUE #2458: Track user not having a default workspace
We will soon make users select a default workspace if they don't have one.

However, there might be race conditions or other weird things going on, so I still want to keep the current interactor that returns an arbitrary workspace if no default is set.

In addition, I want it to track whenever that case happens so we know if it actually does.

juanlaube:
Closes #2458


simonrozsival:
I must say that I am not sure if putting this in the interactor would be beneficial in any way, tracking during sync makes more sense to me.

amulware:
Why?

amulware:
Tracking it in the interactor would certainly be a lot less code.

juanlaube:
@amulware @simonrozsival I thought about using the interactor first, but that's the problem, it's just an interactor we can use it everywhere thus tracking the event multiple times. That's why I think it makes more sense to do this during the sync process.

Let me know if you prefer to keep it this way or move the tracking to the interactor and I will add tests.

willsb:
I'm with Simon and Juan, for the reasons Juan mentioned there. Also, I'm just commenting here, but code looks good and can be merged after tests are included

amulware:
Very well, though note that sync also happens arbitrarily often /shrug

PR #2481, opened 2018-06-18
Allow simultaneous recognizers

### LINKED ISSUE #2480: Reactive UIExtension for UIView taps not working properly
~~The code for Reactive UIExtensions needs tests still (#2356), but we are already using it.~~ It needs tests, there's no issue for that yet.

There's a problem with the `UIView.Tapped` extension method. It disables any prior gestures added to that view, probably because of `ShouldRecognizeSimultaneously` not being implemented.

Also, I think this Observable should be shared, so new subscribers don't create a new gesture recognizer every time.

Odrakir:
Closes #2480 

Accessing this extension method on a `UIView` that already had a gesture recognizer, would result in the first one being disabled. This fixes it.

I think we should also be sharing this Observable, but it's not possible with this approach, as it will generate a new one every time the method is called. Maybe we should not have this extension at all, and just add an extension on `UITapGestureRecognizer`?


simonrozsival:
I am not sure about what you write in the last paragraph - maybe just make an issue from it and we can discuss it there.

simonrozsival:
Or maybe don't create another issue, but change this PR so that it doesn't close the issue yet.

Odrakir:
I will wait for @willsb approval as well, as he wrote that code.

The issue is that it makes sense to share the tap `Observable`, because it's basically the same behavior for every subscriber. But right not it's not possible (there are no extension properties or anything like it, as far as I know).

One way, maybe, to avoid this is to not have this extension at all, and instead do one on `UIGestureRecognizer`. So we would need to create the recognizer manually, but not the target-action.

I can create a new issue to discuss it if necessary, but let's wait for what @willsb has to say.

Toggl.Daneel/Extensions/UIExtensions.cs
```diff
@@ -18,6 +18,7 @@ public static IObservable<Unit> Tapped(this UIView view)
             => Observable.Create<Unit>(observer =>
             {
                 var gestureRecognizer = new UITapGestureRecognizer(() => observer.OnNext(Unit.Default));
+                gestureRecognizer.ShouldRecognizeSimultaneously = (recognizer, otherRecognizer) => true;
```
kornelijepetak: Thank you for not using `_` and `__` here. 😄

Odrakir: Oh!. Great idea! 😄

PR #2485, opened 2018-06-18
Complex event metrics

kornelijepetak:
This PR supersedes #2460 (but this solution was discussed there so some of the comments there are still valid).

The last attempt at complex event metrics.

Complex events can be declared as separate types with all the properties needed for metrics. (`StartTimeEntryEvent.cs` is just an example that would be deleted before merge.)

There was a concern by @amulware that this way we would be able to forget to add something to dictionary. Therefore I have included a test that checks all instance of `ITrackableEvent` whether there is the same number of properties (sans the name) as there are keys in the returned dictionary.

There is only one (potential) issue with this, which is not a big deal IMO, but because of what the test is doing (creating an instance of each of those types). The constraint is that with this test the trackable event classes must not contain properties of other reference types (except string), or if they do, the `ToDictionary` would need to  treat them as thought they will be `null`, because in tests, they indeed will be. However, since we track mostly primitives, I don't think this is a problem.

Anyways, usage:

```cs
new RandomEvent(some, params, here)
    .TrackWith(analyticsService);
```
or
```cs
someObservable
    .Track(new RandomEvent(some, params, here), analyticsService)
    .Subscribe(...);
```
or
```cs
analyticsService.Track(new RandomEvent(some, params, here));
```


Odrakir:
I just have one question. Can the event be a struct? Isn't it more efficient? Does it make sense?. The docs say: "Unless you need reference-type semantics, a small class may be more efficiently handled by the system if you declare it as a struct instead."

amulware:
The key point being 'small', which these events probably won't be. Also, using structs and interfaces could lead to boxing which will end up worse than using a class in the first place - though this could probably be mitigated by using proper generics. In general, I would not worry about this here. Allocation is cheap in any case, and we're not allocating a lot of these. If we started allocating in a loop, I'd be more concerned.

Toggl.Foundation.Tests/AnalyticsService/ComplexEventsTrackingTests.cs
```diff
@@ -0,0 +1,61 @@
...
+using FluentAssertions;
+using Toggl.Foundation.Analytics;
+using Xunit;
+using static System.Reflection.BindingFlags;
+
+namespace Toggl.Foundation.Tests.AnalyticsService
+{
+    public sealed class ComplexEventsTrackingTests
+    {
+        public void MakesSureAllEventsCorrectlySerializeToDictionary()
```
amulware: Is this a `Fact`?

Also naming. This class/method won't read nicely in our logs.

Toggl.Foundation.Tests/AnalyticsService/ComplexEventsTrackingTests.cs
```diff
@@ -0,0 +1,61 @@
...
+        {
+            var trackableEventType = typeof(ITrackableEvent);
+
+            trackableEventType.Assembly
+                .GetTypes()
+                .Where(type => trackableEventType.IsAssignableFrom(type))
```
amulware: method group?

Toggl.Foundation/Analytics/ComplexEvents/StartTimeEntryEvent.cs
```diff
@@ -0,0 +1,32 @@
...
+
+namespace Toggl.Foundation.Analytics.ComplexEvents
+{
+    /// <summary>
+    /// This is just an example event.
+    /// It will be removed and replaced with complex events that we need.
+    /// </summary>
+    public class StartTimeEntryEvent : ITrackableEvent
+    {
+        public string EventName { get; } = "StartTimeEvent";
```
amulware: I'd recommend `=>` to make the object smaller.

Toggl.Foundation/Analytics/ComplexEvents/StartTimeEntryEvent.cs
```diff
@@ -0,0 +1,32 @@
...
+        public string Description { get; set; }
+        public int ProjectId { get; set; }
+        public DateTimeOffset StartTime { get; set; }
+        public string Tags { get; set; }
+
+        public Dictionary<string, string> ToDictionary()
+        {
+            return new Dictionary<string, string>
+            {
+                [nameof(ProjectId)] = ProjectId.ToString(),
```
amulware: Will we be careful enough to not 'accidentally' rename these properties and end up with inconsistent data?

Toggl.Foundation/Analytics/ComplexEvents/StartTimeEntryEvent.cs
```diff
@@ -0,0 +1,32 @@
...
+    /// It will be removed and replaced with complex events that we need.
+    /// </summary>
+    public class StartTimeEntryEvent : ITrackableEvent
+    {
+        public string EventName { get; } = "StartTimeEvent";
+
+        public string Description { get; set; }
+        public int ProjectId { get; set; }
+        public DateTimeOffset StartTime { get; set; }
+        public string Tags { get; set; }
```
amulware: Mutable event objects? What prevents us from forgetting to assign all values? What if we add a new property, do we have to make sure to search the solution for all usages? I much rather have a constructor for these.

Toggl.Foundation.Tests/AnalyticsService/ComplexEventsTrackingTests.cs
kornelijepetak: Oh, this was a copy/paste error from windows where I was trying that in LinqPad outside of testing environment.

Toggl.Foundation.Tests/AnalyticsService/ComplexEventsTrackingTests.cs
kornelijepetak: Yes please. 😄

Toggl.Foundation/Analytics/ComplexEvents/StartTimeEntryEvent.cs
kornelijepetak: What do you mean? I mean, @amulware just look at the code. You can accidentally do so many wrong things. But when was the last time you accidentally did something like that in class this simple? I am not saying it's impossible, I'm just saying that it's kind of unlikely.

kornelijepetak: Also, the tests checks whether every property (except EventName) is in the dictionary.

Toggl.Foundation/Analytics/ComplexEvents/StartTimeEntryEvent.cs
kornelijepetak: I agree. These should have constructors and readonly props.

kornelijepetak:
@amulware here are my reasons for types vs methods.

1. An event is a thing (object), not something to do (method)
2. `IAnalyticsEvent` (current solution) is a surrogate class that is supposed to cover all kinds of events an has the analytics service reflectiomagically injected on startup
3. Longer startup because of 2.
4. Having to add a new generic type `IAnalyticsEvent<T1, T2, ... TN>` when adding a new event with an unseen parameter feels kind of wrong.
5. Potential to subtype events (if/when such a need arises)

Toggl.Foundation/Analytics/ITrackableEvent.cs
```diff
@@ -0,0 +1,16 @@
...
+    public interface ITrackableEvent
+    {
+        string EventName { get; }
+        Dictionary<string, string> ToDictionary();
+    }
+
+    public static class TrackableEventExtensions 
+    {
+        public static void TrackWith(this ITrackableEvent trackableEvent, IAnalyticsService analyticsService)
+            => analyticsService.Track(trackableEvent);
```
simonrozsival: We don't need this extension method - it just swaps the parameters.

simonrozsival: But maybe - we could remove the `void Track(ITrackableEvent trackableEvent);` method from the analytics service and use this extensiom method to call the `void Track(string eventName, Dictionary<string, string> parameters = null);` method of the `IAnalyticsService` instead?

kornelijepetak: I agree. We need only one way of doing things. I included both just to see which API has a better feel to it.

amulware:
Is the idea then to use this for all events, or only for those with more parameters?

kornelijepetak:
That is up to all the devs. I would be fine with both outcomes.

But for now, I didn't mean to destroy the @simonrozsival's effort. The only issue that bothers me though with that solution is the startup time which is already long... but maybe that's not even that much. Maybe he did some benchmarking so he would know better.

simonrozsival:
I measured it and the impact of my changes to the startup time is somewhere between 0.1 to 10 ms :) (it depends on the direction of the wind and the angle of the Sun) so I wouldn't worry about that.

Odrakir:
I do have a problem with current approach. For #1915 I need an event with 12 properties.

Odrakir:
Oh, sorry. I just read we mean to use both ways.

Toggl.Foundation/Analytics/ComplexEvents/StartTimeEntryEvent.cs
amulware: But it does not check if we changed the property name and now appcenter/firebase have weird data split between two different properties.

Toggl.Foundation/Analytics/ComplexEvents/StartTimeEntryEvent.cs
```diff
@@ -0,0 +1,40 @@
...
+using System.Linq;
+using System.Collections.Generic;
+using Toggl.Multivac.Models;
+
+namespace Toggl.Foundation.Analytics.ComplexEvents
+{
+    /// <summary>
+    /// This is just an example event.
+    /// It will be removed and replaced with complex events that we need.
+    /// </summary>
```
amulware: No comments in the code please.

amulware:
I don't have time to review in detail right now. If other's are alright with this, then go ahead without me. Apart from the comment just now, it seems alright at a glance.

Toggl.Foundation/Analytics/ComplexEvents/StartTimeEntryEvent.cs
kornelijepetak: 🤣 Come on, read it.
I will remove it on approval. This was just to show how it would look like. This is not a real event.

PR #2488, opened 2018-06-18
Push workspaces

### LINKED ISSUE #2333: Add workspace creation sync

simonrozsival:
Closes #2333 

I am not really sure about this "serialization" of the workspace. It works well now - it works exactly as it should - but I am not sure if it is future proof enough.

We have feature-dependent serialization kind of implemented, but we don't use it, and it is not even specified how exactly the serialized workspace for the POST endpoint should look like (e.g., the rounding prop is not allowed for free workspaces).

The thing is... we have 12 properties of the workspace entity, but we only use one - the name. Why implement a complex serialization, when we don't use it? Will we ever use some of it? Should we keep these data until we actually start using it (probably yes)?

Please let me know in the comments if you think this is a good thing.


amulware:
This doesn't feel quite right, but I get why you're doing it, and maybe it's ok for now?

Toggl.Multivac.Tests/EnsureTests.cs
```diff
@@ -143,7 +142,7 @@ public void ThrowsWhenTheArgumentIsNotADefinedEnumValue()
             public void DoesNotThrowWhenTheArgumentIsNotNull()
             {
                 Action whenTheCalledArgumentIsNotADefinedEnumValue =
-                    () => Ensure.Argument.IsADefinedEnumValue<SyncStatus>(SyncStatus.SyncNeeded, "argument");
+                    () => Ensure.Argument.IsADefinedEnumValue(BeginningOfWeek.Monday, "argument");
```
amulware: Not related to the PR.

Toggl.Multivac.Tests/Toggl.Multivac.Tests.csproj
```diff
@@ -25,6 +25,5 @@
   </ItemGroup>
   <ItemGroup>
     <ProjectReference Include="..\Toggl.Multivac\Toggl.Multivac.csproj" />
-    <ProjectReference Include="..\Toggl.PrimeRadiant\Toggl.PrimeRadiant.csproj" />
```
amulware: Not related to the PR.

simonrozsival:
I removed the unrelated changes.

PR #2489, opened 2018-06-18
Rxfactor the Licenses view

willsb:
Thanks to @Odrakir, this is now just downright simpler; No need to use observables when we don’t need them

Squash whenever


Toggl.Daneel/ViewSources/LicensesTableViewSource.cs
```diff
@@ -1,54 +1,51 @@
...
         private const string headerIdentifier = nameof(LicensesHeaderViewCell);
 
-        public LicensesTableViewSource(UITableView tableView) : base(tableView)
+        private readonly IImmutableList<License> items;
+
+        public LicensesTableViewSource(UITableView tableView, IImmutableList<License> items)
         {
+            this.items = items;
+
+            tableView.EstimatedRowHeight = 396;
```
Odrakir: I think these three lines shouldn't be in the TableViewSource. They belong to the UI (VC)

Toggl.Foundation.MvvmCross/ViewModels/Settings/LicensesViewModel.cs
```diff
@@ -1,34 +1,29 @@
...
-
-        public override Task Initialize()
-        {
-            Licenses = licenseProvider
-                .GetAppLicenses()
-                .Select(keyValuePair => new License(keyValuePair.Key, keyValuePair.Value))
-                .ToList();
-
-            return base.Initialize();
+            Licenses =
```
Odrakir: I don't like this formatting... Is it just me?

Toggl.Daneel/ViewSources/LicensesTableViewSource.cs
```diff
@@ -1,54 +1,51 @@
...
+            tableView.SectionHeaderHeight = 44;
+            tableView.RowHeight = UITableView.AutomaticDimension;
+
             tableView.RegisterNibForCellReuse(LicensesViewCell.Nib, cellIdentifier);
```
Odrakir: This should probably also go out. If you create the table view cell in interface builder you won't need these lines (I'm not sure if they would actually cause an error)

Toggl.Daneel/ViewControllers/Settings/LicensesViewController.cs
```diff
@@ -1,37 +1,19 @@
...
+﻿using MvvmCross.iOS.Views;
 using Toggl.Daneel.ViewSources;
 using Toggl.Foundation;
 using Toggl.Foundation.MvvmCross.ViewModels;
-using UIKit;
 
 namespace Toggl.Daneel.ViewControllers
 {
-    public sealed partial class LicensesViewController : MvxViewController<LicensesViewModel>
+    public sealed class LicensesViewController : MvxTableViewController<LicensesViewModel>
```
Odrakir: Does it make sense to modify `ListTableViewSource` to work with sections? You would only need to have an `IImmutableList<IImmutableList<TModel>>` instead of the current one... or maybe with another object to also add section titles. I know this Licenses view is an special case (just one row per section), but still, it could be done.

Alternatively we could have a `SectionedListTableViewSource`.

Toggl.Daneel/ViewSources/LicensesTableViewSource.cs
Odrakir: And with those two, you could remove the reference to the table view from the constructor.

Toggl.Daneel/ViewControllers/Settings/LicensesViewController.cs
willsb: We can do that, but that's something for another PR

PR #2490, opened 2018-06-19
Wheel metrics

### LINKED ISSUE #1915: Wheel metrics

Odrakir:
Closes #1915 

The way analytics events are sent from the view controller might be a little weird right now. This is because this VC is not using the new bindings yet, but when we move in that direction the events and the actions should be together in a single stream.

For now, they are completely separated from the rest of the code.


Toggl.Daneel/Views/EditDuration/WheelForegroundView.cs
```diff
@@ -87,6 +91,10 @@ public sealed class WheelForegroundView : BaseWheelView
 
         public event EventHandler EndTimeChanged;
 
+        private readonly  Subject<EditTimeSource> timeEditedSubject = new Subject<EditTimeSource>();
+        public IObservable<EditTimeSource> timeEdited
```
simonrozsival: Capital `T`

Toggl.Daneel/ViewControllers/EditDurationViewController.cs
```diff
@@ -196,6 +204,38 @@ public override void ViewDidLoad()
...
+            Observable.Merge(
+                    dateComponentChanged,
+                    timeComponentChanged,
+                    WheelView.timeEdited,
+                    durationInputChanged
+                )
+                .Distinct()
+                .Subscribe(ViewModel.TimeEditedWithSource)
+                .DisposedBy(disposeBag);
+
```
simonrozsival: 👀

Toggl.Foundation/Analytics/EditDurationEvent.cs
```diff
@@ -0,0 +1,126 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace Toggl.Foundation.Analytics
+{
+    public class EditDurationEvent : ITrackableEvent
+    {
+        public enum NavigationOrigin
+        {
+            start,
```
simonrozsival: Use `PascalCase` for all enum values.

Toggl.Daneel/Views/EditDuration/WheelForegroundView.cs
Odrakir: I will never get used to that.

Toggl.Daneel/Views/EditDuration/WheelForegroundView.cs
```diff
@@ -211,12 +219,15 @@ public override void TouchesMoved(NSSet touches, UIEvent evt)
             switch (updateType)
             {
                 case WheelUpdateType.EditStartTime:
+                    timeEditedSubject.OnNext(EditTimeSource.WheelStartTime);
```
simonrozsival: This emits every time the user interacts with the wheel this might be potentially tens of times per second. Even though you use `Distinct()` and every event is logged at most once, wouldn't it be better to move this into `TouchesBegan`? Or is this concern groundless?

Toggl.Foundation/Analytics/EditDurationEvent.cs
```diff
@@ -0,0 +1,126 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace Toggl.Foundation.Analytics
+{
+    public class EditDurationEvent : ITrackableEvent
```
kornelijepetak: @Odrakir why don't you create a private dictionary and fill it/overwrite values as they come either through the constructor or `With`/`Update` methods?
That way you don't even need private fields. or all that copying around.
Since the dictionary will remain private, you don't lose immutability either (from outside).
And then simply do;
`public Dictionary<string, string> ToDictionary() => internalData.ToDictionary();`

Toggl.Daneel/ViewControllers/EditDurationViewController.cs
```diff
@@ -211,6 +250,8 @@ protected override void Dispose(bool disposing)
 
             if (!disposing) return;
 
+            disposeBag.Dispose();
```
kornelijepetak: I know that the chance for someone to mess up the bag is borderline 0, but just for the sake of `Dispose should never throw`, protect it with a `?`.

Toggl.Daneel/Views/EditDuration/WheelForegroundView.cs
```diff
@@ -87,6 +91,10 @@ public sealed class WheelForegroundView : BaseWheelView
 
         public event EventHandler EndTimeChanged;
 
+        private readonly  Subject<EditTimeSource> timeEditedSubject = new Subject<EditTimeSource>();
+        public IObservable<EditTimeSource> TimeEdited
```
kornelijepetak: Give those members some space, they are not lovers.

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
```diff
@@ -637,7 +637,7 @@ private void toggleBillable()
             var currentDuration = DurationParameter.WithStartAndDuration(StartTime, Duration);
 
             var selectedDuration = await navigationService
-                .Navigate<EditDurationViewModel, EditDurationParameters, DurationParameter>(new EditDurationParameters(currentDuration))
+                .Navigate<EditDurationViewModel, EditDurationParameters, DurationParameter>(new EditDurationParameters(currentDuration, true))
```
daividssilverio: It would be easier to read if this `true` as passed as a named parameter:
`new EditDurationParameters(currentDuration, isStartingNewEntry: true)`

Toggl.Daneel/Views/EditDuration/WheelForegroundView.cs
Odrakir: Actually, we need to talk about this. Those two are grouped because they belong together, but I think it's something we should do with all properties as well. Maybe at first public properties together, then a new line, and then private properties. And if there are too many of them use some other kind of grouping.

But this one empty line per property nonesense is a waste of pixels!

Toggl.Foundation/Analytics/EditDurationEvent.cs
Odrakir: I think that would complicate things in terms of the parameter types. In the Dictionary they are always `string`, but I'm sending them to the constructor as `bool` or `Result`, so it would imply converting back and ford in `With`.

Toggl.Daneel/ViewControllers/EditDurationViewController.cs
Odrakir: I don't understand what you mean here.

daividssilverio: The disposeBag is not final, someone might nullify it by mistake and this would throw. The ? would prevent that.

Toggl.Daneel/Views/EditDuration/WheelForegroundView.cs
kornelijepetak: Ok, about this... I also like grouping (sometimes even domain-separated `partial` classes) but unless we make some conventions about how to group stuff, I don't think everyone should do whatever...
I would very much like you (or us, if you are willing) to propose a set of rules so we can discuss.

Toggl.Foundation/Analytics/EditDurationEvent.cs
kornelijepetak: I do not think you are completely right here, but it's not that what you are doing is wrong, so I will not force the changes here. The point I was making is that you don't need to recreate the objects all the time in order to achieve the same level of immutability.

Toggl.Daneel/Views/EditDuration/WheelForegroundView.cs
Odrakir: My idea is to group first by access (public/private) and then, if needed, semantically, which is kind of subjective.

Issue #2075, opened 2018-05-09
Custom feedback screen internals


amulware:
Backend will make us a custom feedback endpoint so that we don't have to rely on email availability.

That endpoint will also allow us to send more custom data without presenting it to the user or giving them the option to change it.

Here is a list of things that I can think of that I would like to send:
```
- phone model
- platform and os version
- app name and version
- number of workspaces available to the user
- number of time entries in our database (total, unsynced, unsyncable)
- time of last attempted sync
- time of last successful (full?) sync
- time of sending this feedback thing
- manual mode setting
- time of login
```

Maybe there is more, but this is my list right now. The idea here is that this data should allow our support people to help debug things faster without having to go look at user data or having to ask questions we can give them the answers to already.

amulware:
Note the backend issue for the format of the data expected. It's basically key-value tuples which should serialise trivially from a Dictionary<string, string> (maybe cast to Enuemerable<KeyValuePair<string, string>>).

amulware:
This is probably blocked by https://github.com/toggl/mobileapp/issues/2350

simonrozsival:
We are waiting for the API to be implemented and deployed.

PR #2491, opened 2018-06-19
Store last time usage (login and sync)

simonrozsival:
Needed for #2075 

This adds more capabilities to the `SettingsStorage` - we need to remember some times when the last time the user used certain features of the app (login and full sync attempt/success).

It didn't make sense to put it into the existing interfaces, so I added a new one.

Squash whenever


Toggl.PrimeRadiant/Settings/LastTimeStorage.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System;
+
+namespace Toggl.PrimeRadiant.Settings
+{
+    public interface ILastTimeUsageStorage
```
arturs105: The file name should have the same name as the interface (missing an `I`).

simonrozsival:
I agree that we should discuss and revisit how to organize all these stored things. Please create an issue :)

Toggl.Giskard/Services/SharedPreferencesStorage.cs
```diff
@@ -20,6 +20,17 @@ public bool GetBool(string key)
...
 
+        public DateTimeOffset GetDateTimeOffset(string key)
+        {
+            var serialized = GetString(key);
+            if (DateTimeOffset.TryParse(serialized, out var parsed))
+            {
+                return parsed;
+            }
+
+            return default(DateTimeOffset);
```
arturs105: I'm thinking that returning a nullable `DateTimeOffset` would better indicate, that the operation might fail. When returning `default(DateTimeOffset)`, people might forget to check that an actual date time was returned and just use the default date time as an actual stored date time.

ppl might use default(date) as a valid date

PR #2493, opened 2018-06-19
Daneel 1.7 - Neptune

amulware:
```
- Fixed rare cases where running time entry would not display properly

- Changed the app to always keep two months of time entries locally
- Made some major technical changes under the hood to more easily support upcoming features

- Made several other UI improvements and fixes
```


amulware:
Made an empty commit to increase the build number, because appstore-connect is broken.

PR #2495, opened 2018-06-19
Fix time entries create method in data source

simonrozsival:
@willsb noticed a problem that the time entries' log is not updated when you continue a time entry while another time entry is already running and the device is offline - the stopped time entry is not shown in the log.

The problem was that the data source used the `UpdateWithConflictResolution` method and ignored the `update` result (the stopped TE) and only processed the `create` result - there was an assumption that for every entity in the arguments list of batch update a single result would be returned - but the rivals resolver can add one additional result for the stopped TE. Restarting the app fixed the UI state.

Now the `BatchUpdate` method is used to be sure it returns multiple results. I will add a follow-up issue which will change the signature of the `UpdateWithConflictResolution` method to return a list of results instead of a single result.

Squash whenever


willsb:
Will you rebase this into the release branch?

simonrozsival:
I will rebase it once there is a hotfix branch.

amulware:
The changelog and changeset confuse me. Did something go wrong with a rebase here?

simonrozsival:
You are right, I am not sure what happened... everything this PR does is in the last two commits. I will fix it.

simonrozsival:
Now the PR should be easy to review. Sorry for the weirdness guys.

PR #2497, opened 2018-06-19
Add metrics for checking status of entities push sync

### LINKED ISSUE #1323: Track successful/failed sync of entities

kornelijepetak:
Closes #1323 

(unless #1323 deals with errors in Pull as well).
In any case this PR adds tracking for entity pushes.

There are several types of events (included because of the way the events are shown in FB/AC).

Each attempt at entity sync will result either in a `*SyncError` event or an `EntitySynced`.
The reason why there is an event for every entity is to be able to easily find out the distribution of causes. However, there is only one way for an entity to succeed (backend accepts the request), in case of which, we just need the information about the entity.

And the last, but not the least, is the `EntitySyncStatus` which has an `entity | method-status` pair, eg. `Client | Update:Success` or `TimeEntry | Delete:Failure` which will give us better understanding across entity as one dimension and success/failure as the other, without caring much about the reason of failure.

There is some redundancy of course, but since we can't do better queries in FB/AC, I am afraid these are needed. But if @amulware thinks it's an overkill, just say so and we'll modify this.

~Also, I have yet to make tests for this, so... no 🦑 until that happens.~
Can be 🦑 whenever.


kornelijepetak:
Also, I will fix the tests when I get confirmation from @simonrozsival that this is a valid case.

Toggl.Foundation/Sync/StateMachineOrchestrator.cs
```diff
@@ -59,7 +59,10 @@ private void onStateEvent(StateMachineEvent @event)
                 completeCurrentSync(new Success(State));
 
             if (@event is StateMachineError error)
+            {
                 completeCurrentSync(new Error(error.Exception));
+
```
simonrozsival: 👀

Toggl.Foundation/Sync/States/Push/CreateEntityState.cs
```diff
@@ -35,9 +38,10 @@ internal sealed class CreateEntityState<TModel, TThreadsafeModel>
         public override IObservable<ITransition> Start(TThreadsafeModel entity)
             => create(entity)
                 .SelectMany(Overwrite(entity))
+                .Do(TrackEntitySyncSuccess)
```
simonrozsival: I would remove the protected method and call this directly - `.Track(AnalyticsService.EntitySynced, PushSyncOperation.Create, typeof(T).Name)` (where `PushSyncOperation` is a new enum with the 4 operations).

Toggl.Foundation/Sync/States/Push/BasePushEntityState.Tracking.cs
```diff
@@ -0,0 +1,48 @@
...
+        private const string success = "Success";
+        private const string failure = "Failure";
+
+        private void trackEntitySyncFailure(Exception exception, T entity)
+        {
+            var methodName = getMethodName();
+            var entityName = entity.GetType().Name;
+
+            var entitySyncFailureParameter = $"{methodName} : {exception.Message}";
+            entity.ToAnalyticsEvent(AnalyticsService)?.Track(entitySyncFailureParameter);
```
simonrozsival: We should not allow nulls. There should be a test for each entity type that this works.

Toggl.Foundation/Sync/States/Push/CreateEntityState.cs
kornelijepetak: `typeof(T)` won't work. `typeof(T)` is `TThreadSafeModel`. I need to use `entity.GetType().Name`.

Toggl.Foundation/Analytics/SyncAnalyticsExtensions.cs
```diff
@@ -0,0 +1,37 @@
...
+                case IThreadSafeClient _: 
+                    return analyticsService.ClientsSyncError;
+                case IThreadSafeProject _: 
+                    return analyticsService.ProjectsSyncError;
+                case IThreadSafeTask _: 
+                    return analyticsService.TasksSyncError;
+                case IThreadSafeTimeEntry _: 
+                    return analyticsService.TimeEntrySyncError;
+                default:
+                    return null;
```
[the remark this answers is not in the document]
kornelijepetak: @simonrozsival I realized this has to stay like this, because of the tests. Test implements an interface and a new foundation-unrelated model `TestModel` that would throw, effectively failing lots of tests. Because the `Track` RX operator is now in the pipeline, it would throw. Please comment if you have better ideas... 🤔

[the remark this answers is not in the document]
simonrozsival: I don't think that is a good reason to remove this switch. If this is not good for testing, then maybe think about a way how to inject this method instead of being a fixed implementation of the extension method so that it can be mocked in the tests?

Or keep this method but use all of the supported types in the tests using a `Theory` - the same thing that we do for `ApiExceptions`.

Toggl.Foundation/Sync/States/Push/CreateEntityState.cs
```diff
@@ -39,8 +39,8 @@ internal sealed class CreateEntityState<TModel, TThreadsafeModel>
         public override IObservable<ITransition> Start(TThreadsafeModel entity)
             => create(entity)
                 .SelectMany(Overwrite(entity))
-                .Track(AnalyticsService.EntitySynced, Create, typeof(TThreadsafeModel).ToString())
-                .Track(AnalyticsService.EntitySyncStatus, typeof(TThreadsafeModel).ToString(), $"{Create}:{success}")
+                .Track(AnalyticsService.EntitySynced, Create, entity.GetSafeTypeName())
+                .Track(AnalyticsService.EntitySyncStatus, entity.GetSafeTypeName(), $"{Create}:{Resources.Success}")
```
simonrozsival: What was wrong with `typeof(TThreadsafeModel).ToString()`? If you don't want the full name, then `typeof(TThreadSafe).Name` will do the job and we won't need another extension method.

Toggl.Foundation.Tests/Sync/States/Push/TestModel.cs
```diff
@@ -17,12 +17,22 @@ public interface IThreadSafeTestModel : IThreadSafeModel, IDatabaseTestModel
     {
     }
 
+    public interface IThreadSafeWorkspaceTestModel : IThreadSafeTestModel, IThreadSafeWorkspace { }
```
simonrozsival: Why do we need these test interfaces?

kornelijepetak:
@simonrozsival FYI, this is ready for rereview...

Just a few notes.
For properly testing whether the correct event was being chosen in case of `ITestThreadSafeModel`s, I had to introduce a new surrogate event into analytics service and since this should not pollute the foundation, I had to inherit from `IAnalyticsService` and use that in base tests. This interface has one event more for testing purposes.

Also, since the actual type in tests is unknown at compile time, I've had to use `.GetType()` on the entities, which means that I had to do `IsAssignableFrom` instead of `== tyepeof(ISomethingSomething)`.

I suggest you don't have any more new ideas for improving this 😛

kornelijepetak:
@amulware agreed to merge this without his review.

PR #2498, opened 2018-06-19
Refactor TextFieldInfo

willsb:
:construction: :rotating_light: WIP WIP WIP WIP WIP :rotating_light: :construction: 

Do not:

- Review the entire thing
- Expect it to be bug free
- Mind common bugs that should be caught by the tests. Those will be fixed when tests are redone

Do:
- Test the iOS version and report bugs
- Review the TextFieldInfo and find problems in it
- Give ideas of possible points of improvement
- Not approve/merge this PR


juanlaube:
I will install this on my phone with iOS 11 and start testing without taking a look at a single line of code.

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
```diff
@@ -541,23 +441,15 @@ private void toggleProjectSuggestions()
             analyticsService.StartEntrySelectProject.Track(ProjectTagSuggestionSource.ButtonOverKeyboard);
             OnboardingStorage.ProjectOrTagWasAdded();
 
-            if (TextFieldInfo.ProjectId != null)
+            if (textFieldInfo.HasProject)
```
daividssilverio: Shouldn't the projects suggestions be off when the textFieldInfo already has a project set?

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
```diff
@@ -541,23 +441,15 @@ private void toggleProjectSuggestions()
...
-        }
-
-        private void appendSymbol(string symbol)
-        {
-            var cursor = TextFieldInfo.DescriptionCursorPosition;
-            var shouldAddWhitespace = cursor > 0 && Char.IsWhiteSpace(TextFieldInfo.Text[cursor - 1]) == false;
-            var textToInsert = shouldAddWhitespace ? $" {symbol}" : symbol;
-            var newText = TextFieldInfo.Text.Insert(cursor, textToInsert);
-            TextFieldInfo = TextFieldInfo.WithTextAndCursor(newText, cursor + textToInsert.Length);
+            textFieldInfo = textFieldInfo.AddQuerySymbol(QuerySymbols.ProjectsString);
```
daividssilverio: Shouldn't the `IsSuggestingProjects` be set to true at this point? or that subject will update the ui which will update `IsSuggestingProjects ` from `onParsedQuery`?

willsb: I tried not changing logic outside the TextFieldInfo, this PR is already too complex. When Rxfactoring this View we can fix this

Toggl.Daneel/Autocomplete/AutocompleteExtensions.cs
```diff
@@ -0,0 +1,236 @@
...
+                    {
+                        yield return new TextSpan(subText);
+                    }
+                }
+
+                start = end;
+            }
+        }
+
+        public static (NSAttributedString, int) AsAttributedText(this TextFieldInfo self)
```
daividssilverio: 🤔 Maybe a struct/class would be easier to understand when reading just from the signature of this extension? Or maybe just renaming it to `AsAttributedTextWithCursorPosition` or similar.

Toggl.Daneel/ViewControllers/StartTimeEntryViewController.cs
```diff
@@ -184,6 +187,30 @@ public override void ViewDidLoad()
...
+                .ObserveOn(SynchronizationContext.Current)
+                .Subscribe(async info => await ViewModel.OnTextFieldInfoFromView(info))
+                .DisposedBy(DisposeBag);
+        }
+
+        private void onTextFieldInfo(TextFieldInfo textFieldInfo)
+        {
+            var (attributedText, cursorPosition) = textFieldInfo.AsAttributedText();
+
+            DescriptionTextView.AttributedText = attributedText;
```
sofokli1: If you add `DescriptionTextView.InputDelegate = new AutocompleteTextViewInputDelegate();` above this line (also coping the implantation if `AutocompleteTextViewInputDelegate` from my old PR it should still fix the issue that I fixed there.

daividssilverio:
Bugs when interacting with the bottom toolbar on iOS:
1.
```
(Project/Tags) suggestions won't show when the description is empty when tapping the icons from the bottom toolbar.

If you type something after tapping the project icon, it will add the description with `@{letter you typed}` 
and the suggestions appear.
```

2. Projects are overwritten when you select the tag icon from the toolbar
```
Steps to reproduce
- Type your project @existing_project
- Tap one project from the suggestions
- Tap the tag icon if you selected a project before
- You will end up with: "@ #Whatever you have typed before selecting a project from the suggestions"
```

3. The ☝️ applies when selecting a tag from suggestions then the project icon from the toolbar.

4. Sometimes I'm getting crashes when I try adding projects when there is already other project or tags around. Can't reproduce really well.

Toggl.Foundation.Tests/Autocomplete/QueryInfoTests.cs
```diff
@@ -51,7 +53,8 @@ public void ExtractsTheProjectNameWhileTyping(NonEmptyString nonEmptyString)
                     if (text.Length < 2 || text.Contains("#") || text.Contains("@"))
                         return;
 
-                    var textFieldInfo = TextFieldInfo.Empty(1).WithTextAndCursor(text, text.Length);
+                    var textFieldInfo = TextFieldInfo.Empty(1)
```
daividssilverio: The name of the test doesn't make much sense to me. What is this testing? 
Based on the asserts, it looks like it checks that when typing some random texts without `@` or `#`, the suggestions should be of the `AutocompleteSuggestionType.TimeEntries` type and the parsed text should match the typed text.

Toggl.Foundation.Tests/Autocomplete/TextFieldInfoTests.cs
```diff
@@ -15,179 +12,106 @@ public sealed class TextFieldInfoTests
...
 
             [Fact, LogIfTooSlow]
-            public void RemovesTheTaskIdAndTaskName()
+            public void ReplacesTheExistingProjectSpanIfItExists()
             {
-                const long newProjectId = 11;
-                const string newProjectName = "Some other project";
-                const string newProjectColor = "Some other project";
+                var oldTextFieldInfo = CreateDefaultTextFieldInfo();
+                var expected = oldTextFieldInfo.Spans.First();
```
daividssilverio: This `expected` wasn't used for anything. Does `oldTextFieldInfo.Spans.First()` produce any side effects?

Toggl.Foundation/Autocomplete/TextFieldInfo.cs
```diff
@@ -1,152 +1,208 @@
...
-            && Tags == other.Tags;
+            spanBuilder.AddRange(spansAfterReplacement);
+
+            return new TextFieldInfo(workspaceId, spanBuilder.ToImmutable());
+        };
+
+        internal char[] ValidQuerySymbols()
+            => HasProject ? QuerySymbols.ProjectSelected : QuerySymbols.All;
+
+        private string calculateDescription()
```
simonrozsival: This function seems complicated at first but it is not - it is in fact very simple.

I would do something like:

```csharp
private string calculateDescription()
{
    ISpan previous = null;
    return Spans
        .Aggregate(
            new StringBuilder(),
            (builder, span) =>
            {
                builder.Append(chooseSeparator(previous, span));
                builder.Append(toString(span));
                previous = span;
                return builder;
            })
        .ToString()
        .Trim();
}

private string toString(ISpan span)
    => span is TextSpan text ? text.Text.Trim() : string.Empty;

private string chooseSeparator(ISpan first, ISpan second)
    => first is TextSpan && second is TextSpan ? string.Empty : " ";
```

and those two helper methods could be even part of the `ISpan` interface and implemented in the spans themselves or they could be extension methods.

Toggl.Daneel/ViewControllers/StartTimeEntryViewController.cs
```diff
@@ -184,6 +187,30 @@ public override void ViewDidLoad()
...
+            // Reactive
+            ViewModel.TextFieldInfoObservable
+                .AsDriver()
+                .Subscribe(onTextFieldInfo)
+                .DisposedBy(DisposeBag);
+
+            DescriptionTextView.AttributedText()
+                .CombineLatest(DescriptionTextView.CursorPosition(), (text, _) => text)
+                .SubscribeOn(ThreadPoolScheduler.Instance)
+                .Select(text => text.AsImmutableSpans((int)DescriptionTextView.SelectedRange.Location))
```
simonrozsival: I am not sure I understand how this works and if I understand that then I don't understand why it is a good idea to do it that way. So let's see:

- when the user types something or moves the cursor, we take that attributed text...
- ... and convert it into spans by analyzing the attributes of the text
- then we do some operations with this list (like we show suggestions)
- when the user selects some suggestion, we change the list, convert it to the attributed text and replace the old attributed text with this one

I like the fact that we don't update the text after every keystroke, but I am not a fan of converting the text by analyzing the attributes. It feels weird. But we don't need to update any previous state, which is great. So maybe it is just smart and not obvious at first.

Toggl.Daneel/ViewControllers/StartTimeEntryViewController.cs
```diff
@@ -184,6 +187,30 @@ public override void ViewDidLoad()
...
+
+            // Reactive
+            ViewModel.TextFieldInfoObservable
+                .AsDriver()
+                .Subscribe(onTextFieldInfo)
+                .DisposedBy(DisposeBag);
+
+            DescriptionTextView.AttributedText()
+                .CombineLatest(DescriptionTextView.CursorPosition(), (text, _) => text)
+                .SubscribeOn(ThreadPoolScheduler.Instance)
```
simonrozsival: Wouldn't this be a good place where we would put some debouncing?

Toggl.Foundation.MvvmCross/Autocomplete/AutocompleteExtensions.cs
```diff
@@ -0,0 +1,88 @@
...
+                    timeEntrySuggestion.ProjectName,
+                    timeEntrySuggestion.ProjectColor,
+                    timeEntrySuggestion.TaskId,
+                    timeEntrySuggestion.TaskName
+                );
+
+                builder.Add(projectSpan);
+            }
+
+            return textFieldInfo.ReplaceSpans(builder.ToImmutable());
```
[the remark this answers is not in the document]
willsb: Possible bug 👀

Toggl.Foundation.MvvmCross/Autocomplete/AutocompleteExtensions.cs
```diff
@@ -0,0 +1,88 @@
...
+            this TextFieldInfo textFieldInfo,
+            QuerySymbolSuggestion querySymbolSuggestion)
+        {
+            var result = textFieldInfo.AddQuerySymbol(querySymbolSuggestion.Symbol);
+            return result;
+        }
+
+        public static TextFieldInfo HandleTimeEntrySuggestion(
+            this TextFieldInfo textFieldInfo,
+            TimeEntrySuggestion timeEntrySuggestion)
```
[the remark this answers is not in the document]
willsb: This is done in the ViewModel. And, as a matter of fact, since I'm renaming the methods to `From*`, I'm removing the only method that has an AnalyticsService reference and putting it back in the ViewModel

Toggl.Foundation.MvvmCross/Autocomplete/AutocompleteExtensions.cs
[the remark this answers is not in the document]
willsb: Ref Copying tags: ping @amulware and @pe0ny . This might be interesting, though I'd open an issue instead

Toggl.Foundation/Autocomplete/TextFieldInfo.cs
```diff
@@ -1,152 +1,208 @@
...
-            => new TextFieldInfo(Text, CursorPosition, WorkspaceId, null, "", "", null, "", Tags);
+            var padding = needsSpace ? " " : "";
+            var textToAppend = $"{padding}{querySymbol}";
+            return addSpanToRelevantPosition(WorkspaceId, new TextSpan(textToAppend));
+        }
 
+        public TextFieldInfo AddTag(long tagId, string tagName)
+        {
+            var spanToAdd = new TagSpan(tagId, tagName);
+            var spansToReuse = Spans.Where(span => span is TagSpan tagSpan ? tagSpan.TagId != tagId : true);
```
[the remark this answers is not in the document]
willsb: `||` and `!=` is quite hard to read. Maybe extracting a function helps?

Toggl.Foundation.MvvmCross/Autocomplete/AutocompleteExtensions.cs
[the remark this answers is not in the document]
amulware: copying tags to a new workspace? no. we don't do that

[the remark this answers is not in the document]
willsb: Add the tags from the existing time entry into the one being started when you select a StartTimeEntrySuggestion. There's no copying into new workspace involved, whatsoever

Toggl.Foundation/Autocomplete/TextFieldInfo.cs
```diff
@@ -1,152 +1,208 @@
...
-        public TextFieldInfo AddTag(TagSuggestion tagSuggestion)
-        {  
-            if (Tags.Any(t => t.TagId == tagSuggestion.TagId))
-               return this;
+        internal QueryTextSpan GetSpanWithCurrentTextCursor()
+            => Spans.OfType<QueryTextSpan>().SingleOrDefault();
+
+        private TextFieldInfo removeQuery(QueryTextSpan currentQuerySpan, char querySymbol)
+        {
+            var indexOfProjectQuerySymbolInSpan = currentQuerySpan.Text.IndexOf(querySymbol);
```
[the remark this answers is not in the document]
willsb: Which bits do you see being moved?

Toggl.Foundation/Autocomplete/TextFieldInfo.cs
willsb: Great suggestion :)

Toggl.Daneel/ViewControllers/StartTimeEntryViewController.cs
willsb: > but I am not a fan of converting the text by analyzing the attributes

How else would you store state? Observing text changes and saving a copy of that state internally (either in the VC or in some dedicated View) is a recipe to disaster.

Toggl.Foundation/Autocomplete/TextFieldInfo.cs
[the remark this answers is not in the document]
simonrozsival: lines 101 - 104 are not trivial and they could be moved to the `QyeryTextSpan` class

Toggl.Foundation.Tests/Autocomplete/QueryInfoTests.cs
willsb: The naming was like that before, so in order to keep this PR from getting bigger, how about we refactor those on another PR?

Toggl.Foundation.Tests/Autocomplete/TextFieldInfoTests.cs
willsb: Nope

Toggl.Daneel/ViewControllers/StartTimeEntryViewController.cs
```diff
@@ -193,6 +194,7 @@ public override void ViewDidLoad()
 
             DescriptionTextView.AttributedText()
                 .CombineLatest(DescriptionTextView.CursorPosition(), (text, _) => text)
+                .Where(_ => !isUpdatingDescriptionField)
```
simonrozsival: Does this work? If I type "Ab" and this condition prevents the second update after typing "b" from being processed, what will be the outcome? We don't update the list with the suggestions? It might be possible that in some scenario user will type " #" and we won't start suggesting tags?

willsb: This will only ever be false when the description field is updated by any means other than typing (e.g.: pressing the projects button on the bottom bar). This check is only in place to prevent a feedback loop when we set the cursor position.

simonrozsival: OK

Toggl.Foundation/Autocomplete/TextFieldInfo.cs
```diff
@@ -1,152 +1,208 @@
...
+        public bool HasProject
+        {
+            get
+            {
+                if (hasProject.HasValue)
+                    return hasProject.Value;
 
-        public int DescriptionCursorPosition
-            => Math.Min(CursorPosition, Text.Length);
+                projectSpan = Spans.OfType<ProjectSpan>().SingleOrDefault();
```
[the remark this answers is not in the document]
daividssilverio: this was fixed 👍

Toggl.Daneel/Autocomplete/AutocompleteExtensions.cs
```diff
@@ -0,0 +1,236 @@
...
+                MaximumLineHeight = lineHeight
+            };
+
+            tagAttributes = new UIStringAttributes
+            {
+                Font = tokenFont,
+                ForegroundColor = Color.StartTimeEntry.TokenText.ToNativeColor()
+            };
+
+            textVerticalOffset = (lineHeight / 2) - (tokenFont.CapHeight / 2) - 3;
```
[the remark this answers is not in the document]
willsb: I'm changing this in another PR, there's already an issue for this

PR #2499, opened 2018-06-19
Start time entry event with more data

### LINKED ISSUE #1919: Improve TE creation event

daividssilverio:
Closes #1919 

~Still have to add more tests for the parameters of the event.~

I would like opinions on the new `IAnalyticsService.cs`. I've added another Track to the interface to be able to use `AnalyticsService.Track(iTrackableEventInstance)`.
The `BaseAnalyticsService` already had a `Track(ITrackableEvent trackableEvent)`, so I didn't have to do change much.


Toggl.Foundation.Tests/AnalyticsService/StartTimeEntryEventTests.cs
```diff
@@ -0,0 +1,154 @@
...
+using Toggl.Foundation.Analytics;
+using Toggl.Multivac.Extensions;
+using Toggl.Multivac.Models;
+using Xunit;
+
+namespace Toggl.Foundation.Tests.AnalyticsService
+{
+    public sealed class StartTimeEntryEventTests
+    {
+
```
simonrozsival: 👀

Toggl.Foundation/Interactors/TimeEntry/CreateTimeEntryInteractor.cs
```diff
@@ -66,7 +66,7 @@ public IObservable<IThreadSafeTimeEntry> Execute()
                 .Select(userFromPrototype)
                 .SelectMany(dataSource.TimeEntries.Create)
                 .Do(_ => dataSource.SyncManager.PushSync())
-                .Track(analyticsService.TimeEntryStarted, origin);
+                .Do(timeEntry => analyticsService.Track(StartTimeEntryEvent.With(origin, timeEntry)));
```
simonrozsival: What if we added another `Track` extension method which would allow doing this in a nice clean way? Something so the usage could be `.Track(timeEntry => StartTimeEntryEvent.With(origin, timeEntry), analyticsService)` - the implementation would be `static IObservable<T> Track<T>(IObservable<T> observable, Func<T, ITrackableEvent> eventFactory, IAnalyticsService analyticsService) => observable.Do(value => analyticsService.Track(eventFactory(value));`

I don't like using `Do` too much. This will save a few parentheses and make it read nicer. Maybe this is overengineering though... feel free to dismiss this.

Even nicer (and more overengineered) would be to have the `With` function return a `Func<IThreadSafeTimeEntry, StartTimeEntryEvent>` - then it could be called like `.Track(StartTimeEntryEvent.With(origin), anayticsService)`. Maybe don't even consider this, too much work for too little gain.

Toggl.Foundation/Analytics/IAnalyticsService.cs
```diff
@@ -58,5 +56,7 @@ public interface IAnalyticsService
         void Track(string eventName, Dictionary<string, string> parameters = null);
 
         void Track(Exception exception);
+
+        void Track(ITrackableEvent trackableEvent);
```
simonrozsival: Kornelije wanted to add the similar method, but I convinced him to keep just the `TrackWith` extension method, to make this interface cleaner. Maybe it just makes sense to have this method directly in the interface and it will keep returning. If we're doing this, then please also remove the extension method, because it only swaps the arguments now... I will be OK with a few seemingly unrelated changes like this in this PR (I believe it is only used in one place now, so it shouldn't be too much).

Toggl.Foundation.Tests/AnalyticsService/StartTimeEntryEventTests.cs
```diff
@@ -0,0 +1,174 @@
...
+                => Enum.GetValues(typeof(TimeEntryStartOrigin)).Cast<TimeEntryStartOrigin>()
+                    .Select(origin => new object[] { origin });
+        }
+
+        public sealed class TheToDictionaryMethod : StartTimeEntryEventBaseTest
+        {
+            [Fact, LogIfTooSlow]
+            public void MapsTheSevenEventParameterKeys()
+            {
+                StartTimeEntryEvent startTimeEntryEvent = new StartTimeEntryEvent(TimeEntryStartOrigin.Manual, true, true, true, 0, true, true);
```
kornelijepetak: I see that the first two methods of all tests are all the same, which could be made in a more reusable way. But since these are tests, maybe AAA is better expressed in this way. Up to you if you want to change. I am approving it as it is.

daividssilverio: 🤔 I like to write AAAs whenever possible, I've added a private method to create the time entry events.  On the other test class, unfortunately, the "act part" repeats for every test :/ but I guess it really should.
```
StartTimeEntryEvent startTimeEntryEvent = StartTimeEntryEvent.With(default(TimeEntryStartOrigin), TimeEntry);
```

Toggl.Foundation.Tests/AnalyticsService/StartTimeEntryEventTests.cs
```diff
@@ -48,7 +48,7 @@ public void MapsTheSevenEventParameterKeys()
             [MemberData(nameof(TimeEntryStartOriginValues))]
             public void ConvertsTheOriginToAString(TimeEntryStartOrigin origin)
             {
-                StartTimeEntryEvent startTimeEntryEvent = new StartTimeEntryEvent(origin, true, true, true, 0, true, true);
+                StartTimeEntryEvent startTimeEntryEvent = createTimeEntryEvent(origin: origin);
```
kornelijepetak: since the `origin` parameter is the first one, no need to name it, unless for some reason you did that deliberately, although I can't see how `origin: origin` would give more info than simply `origin`. 😄

daividssilverio: I did that on purpose, but you are right, it doesn't add much. Just removed it!

Odrakir:
While I completely agree with the changes to `IAnalayticsService`, please, next time try to make that kind of changes in separate PRs. Smaller PRs are easier to review and also it's easier to follow changes on code you are using.

I had another PR using `IAnalyticsService` and it suddenly started failing, took me a bit to understand why.

PR #2504, opened 2018-06-20
Fix update with conflict resolution

### LINKED ISSUE #2496: Change the signature of the `UpdateWithConflictResolutionMethod`
The method assumes that the internally used `BatchUpdate` method will return just one result for the single input, which might not be the case. Rivals resolver might add another entity into the list... The signature should be changed to return a list of conflict resolution results.

simonrozsival:
Closes #2496 

This PR is a follow-up of #2495.

The `UpdateWithConflictResolution` method now does not ignore any results from rivals resolution and returns a list of results instead of a single result.

Squash whenever


simonrozsival:
PR #2495 was merged to the release branch - this must still wait until it is merged into `develop`.

Toggl.Foundation/DataSources/SingletonDataSource.cs
```diff
@@ -56,14 +56,17 @@ public override IObservable<TThreadsafe> Update(TThreadsafe entity)
...
+            => base.OverwriteIfOriginalDidNotChange(original, entity)
+                .Do(results => results.Do(handleConflictResolutionResult));
 
         public virtual IObservable<IConflictResolutionResult<TThreadsafe>> UpdateWithConflictResolution(
             TThreadsafe entity)
             => Repository.UpdateWithConflictResolution(entity.Id, entity, ResolveConflicts, RivalsResolver)
-                .Select(result => result.ToThreadSafeResult(Convert))
+                .ToThreadSafeResult(Convert)
+                .SelectMany(CommonFunctions.Identity)
+                .FirstAsync()
```
willsb: Not sure if this _could_ or _should_ be `SingleAsync`

simonrozsival: Yeah this could be single - there will always be exactly 1 item in the storage at this point.

PR #2519, opened 2018-06-21
Fix setups missing WithLastTimeUsageStorage call

daividssilverio:
Closes nothing, fixes the current develop build.

squash whenever.


Toggl.Daneel/Startup/Setup.cs
```diff
@@ -64,6 +64,7 @@ protected override IMvxNavigationService InitializeNavigationService(IMvxViewMod
 
             Mvx.RegisterSingleton<IMvxNavigationService>(navigationService);
             return navigationService;
+
```
kornelijepetak: 👀

PR #2443, opened 2018-06-12
Feedback API

simonrozsival:
Closes #2350 

~On-hold: https://github.com/toggl/toggl_api/pull/828 must be finished and deployed to staging before we can start testing this +~ it must be deployed into production before we can merge this into `develop` (so it can be part of the next release for any platform).

WIP: we still don't know how exactly will the API work and if this will work as is or if it will need any further changes + also whether the integration tests are sufficient or if we need any more (or less).

Integration tests: https://app.bitrise.io/build/315c7c2024950136

PR #2526, opened 2018-06-22
Add feedback interactor

### LINKED ISSUE #2075: Custom feedback screen internals

simonrozsival:
Closes #2075 

On-hold: the API (#2443) must be finished & merged first.


daividssilverio:
#2443 is merged. This PR is ready to be reviewed (again?).

PR #2527, opened 2018-06-22
Add long press to start time entry button

### LINKED ISSUE #2130: Shortcut for manual entry
- [x] When Timer mode is active (manual mode is off), allow long press on a start button to start a new entry in Manual mode.
- [x] When Manual mode is active (timer mode is off), allow long press on a start button to start a new entry in Timer mode.

arturs105:
Closes #2130 

Squash whenever.


Toggl.Daneel/Binding/ViewLongPressCommandTargetBinding.cs
```diff
@@ -0,0 +1,29 @@
...
+        {
+        }
+
+        protected override void SetValue(IMvxCommand value)
+        {
+            if (gestureRecognizer != null)
+                Target.RemoveGestureRecognizer(gestureRecognizer);
+
+            gestureRecognizer = new UILongPressGestureRecognizer(() => value.Execute());
+            Target.AddGestureRecognizer(gestureRecognizer);
```
willsb: Please remove it on dispose

simonrozsival:
This does what the issue says, but I think that this way this feature is confusing and it is missing something....

a) it is not easy to discover this feature - it will be only for power users
b) if a user does not know about this feature and holds the button a tiny bit longer than usual, now the app behaves in a different way than he/she is used to and it could confuse them (they probably wouldn't notice until hitting the "done" button)
c) the only way to distinguish if it "worked or not" is to look at the timer at the top of the creation view - if it is at `0:00:00` and counting or if it is 30 min. There is no visual or other feedback - it would be nice if the start button would grow if you pressed it for a longer period of time and there would be different haptics feedbacks for the normal start or the alternative (long press) start

☝️ this is of course addressed to @pe0ny

arturs105:
@simonrozsival is right and i also think that we should definitely add an onboarding tooltip for long pressing the button. Or maybe add this info below the manual mode switch in the settings, but that's up to @pe0ny.

Odrakir:
I think rather than a tooltip which can clutter the UI, it's more about making the change more obvious. The screens on which you enter time should have something that makes it clear are different.

Another option is:
- On long press the floating button shows two sub-buttons (two options kind of like the categories on Superday used to show up)
- User then can drag to select which one he wants (manual or auto)
- If he stays in the same button, the default option (current mode) is selected.

PR #2530, opened 2018-06-22
Logout to allow account selection

### LINKED ISSUE #2432: Google login change account
After signing in and logging out, when trying to sign in again, user can't pick a different account.

### LINKED ISSUE #2434: Google login doesn't work after an error
If you try signing in with Google and it fails, then try to sign in again with google, it will show the loading indicator, but nothing happens.

juanlaube:
Closes #2432 
Closes #2434 

Needs functional testing, let me know if you need help setting up everything for this.


daividssilverio:
I can test with the emulator today and with my phone after the dog food testing :)

Toggl.Giskard/Services/GoogleService.cs
```diff
@@ -22,35 +25,59 @@ public sealed class GoogleService : MvxAndroidTask, IGoogleService
...
+        public GoogleService()
+        {
+            var signInOptions = new GoogleSignInOptions.Builder(GoogleSignInOptions.DefaultSignIn)
                     .RequestIdToken("{TOGGL_DROID_GOOGLE_SERVICES_CLIENT_ID}")
                     .RequestEmail()
                     .Build();
 
-                var googleApiClient = new GoogleApiClient.Builder(activity)
-                    .EnableAutoManage(activity, onError)
+            googleApiClient = new GoogleApiClient.Builder(Application.Context)
```
daividssilverio: 🤔 Do you have a reason to change from the old way using the activity and `EnableAutoManage` for the `GoogleApiClient`? That was the "easy" way of handling google apis.
If you are not using that, please consider the other constructor with two extra parameters: `IConnectionCallbacks` and `IOnConnectionFailedListener` and all the boilerplate involved.

https://developers.google.com/android/reference/com/google/android/gms/common/api/GoogleApiClient.Builder.html#enableAutoManage(android.support.v4.app.FragmentActivity,%20com.google.android.gms.common.api.GoogleApiClient.OnConnectionFailedListener) for reference.

juanlaube: Good catch, I didn't take this into consideration. I wanted to remove the coupling from the activity because it can change, forcing us to rebuild the GoogleApiClient instance and it loses it's state.

daividssilverio:
I'll take the opportunity to check on an actual device now.

PR #2533, opened 2018-06-23
Let the tags on the Edit TE view be clickable to open the tags selection view

### LINKED ISSUE #2520: Clicking the tag itself in EditView does not open tag selector
Clicking on the empty space on that row works though (the left hand side icon as well).

I am strongly against releasing with this bug still in effect.

@amulware if you think it deserves a `critical` label, please include it.

daividssilverio:
Closes #2520

~This is a workaround 🙈 that detects when a click is done inside the tags recycler view and propagates it to its parent! It does not give a ripple when you hold the tags.~

~It will die when we rxfactor the EditTimeEntryActivity and EditTimeEntryViewModel.
The right thing to do here would be adding the item clicks on the tags inside the recycler view as well. Which actually wouldn't be really good as well... the ripple would happen in the individual tag, but the action would be the same as clicking the parent of the recycler view.~

~The recycler view is consuming the touches to handle scrolling and that's why the parent won't receive touch events as well.~

Simple fix. It adds the click callback.

🦑 whenever.


daividssilverio:
@willsb I could change the bindings and introduce a new Adapter to pass to command for the VHs.

PR #2535, opened 2018-06-25
Fix duration field ellipsis

### LINKED ISSUE #2512: Wheel view: seconds are missing
App Store version VS TF 1.7

image from ios 1 image from ios

simonrozsival:
Closes #2512 

~I need help: So far, I "only" simplified the logic with a _simple_ regex so that it is now simpler to reason about how the attributed string is created.~

~The problem is, that it does not fix the issue so far. Why though? I don't understand the behavior and I would like to ask somebody from the more experienced iOS guys to help me with this.~

~It seems to me that setting the foreground color somehow changes the font and it forgets that it is monospaced. Maybe the solution is to somehow convert the `UIFont` into a monospaced font so that it is not ignored. Any ideas?~

For some reason the rendering of the monospaced font does not work well when the color of the font changes in the middle of the text:

monospaced_font

The fix suggested by @sofokli1 works - by adjusting the size of the frame, the text is not truncated. The view is aligned to the center and so it sometimes jumps a bit. Most users won't probably even notice it. But I do notice it and I don't like it. I also don't like spending any more time on this issue right now. I will create a new issue which we can tackle later and use this fix for the 1.7.1.

Squash whenever


sofokli1:
I think the issues is with the internal logic of the UITextField subclass that calculates the size based on the text. I think it is not taking in to account the `MonospacedNumbers`. I tried the following and seems to work.
```csharp
        public override void LayoutSubviews()
        {
            base.LayoutSubviews();

            updateFrame();
        }

        private void updateFrame()
        {
            var attributes = new UIStringAttributes()
            {
                ForegroundColor = UIColor.Red,
                Font = this.Font
            };

            var size = AttributedText.Size;

            Console.WriteLine(size);

            var newFrame = Frame;
            newFrame.Size = size;
            Frame = newFrame;
        }
```

simonrozsival:
This is an improvement, but still the fact that the frame size changes means that the font is not monospaced. The whole thing jumps around a bit now - it is aligning to the center. It is most obvious when looking at the icon to the left of the duration.

It would be better to have this in the hotfix rather than the ellipsis, but I think we should still try to fix this properly tomorrow.

Toggl.Daneel/Views/EditDuration/DurationFieldTextFormatter.cs
```diff
@@ -1,44 +1,21 @@
...
 {
-    public sealed class DurationFieldTextFormatter
+    public static class DurationFieldTextFormatter
     {
-        private const char splitDelimiter = ':';
-        private static UIColor placeHolderColor = Color.Common.PlaceholderText.ToNativeColor();
+        private static readonly UIColor placeHolderColor = Color.Common.PlaceholderText.ToNativeColor();
 
-        public static NSAttributedString AttributedStringFor(string durationText)
+        public static NSAttributedString AttributedStringFor(string durationText, UIFont font)
```
juanlaube: The text color is the placeholder color in some cases depending if you use the wheel or type the duration.
I'd expect UIKit to set a default foreground color but that doesn't seem to happen, so setting the font is not enough.

Something like this solves the problem though:

```
var result = new NSMutableAttributedString(durationText, attributes: new NSMutableDictionary<NSString, NSObject>()
{
    { UIStringAttributeKey.Font, font.GetMonospacedDigitFont() },
    { UIStringAttributeKey.ForegroundColor, UIColor.Black }
});
```

juanlaube: Notice it's a `NSMutableDictionary` because the compiler complains about things in the initialization, but could be just a `NSDictionary` if there's a better way.

simonrozsival:
This is the problem I was talking about:

duration_formatting_glitch

I am not sure how you mean the problem with the color - can you send some screenshots or give a set of steps to reproduce it please?

simonrozsival:
I guess we will keep the jumping thing there, still better than the ellipsis... I will create an issue to fix the jumping and maybe we will be more successful next time.

juanlaube:
Cannot be unseen now 👀

PR #2536, opened 2018-06-25
Fix no start button after starting the app

### LINKED ISSUE #2482: Start time entry button does not appear
Repro steps:

- Fresh install from App Store
- Login (using PW manager, not that it matters)
- Turned connection off after login succeeds

The start time entry button did not appear. Restarting made it work and I could not reproduce since. FWIW, I think we should investigate the relationship between our observable lifecycle vs the internet connection dropping suddenly.

<sub>This is a bug and not a glitch, since it effectivelly makes it impossible for people to create new TEs without resetting, which is gross and gives the impression that the app does not work</sub>

simonrozsival:
Closes #2482 

I was not able to reproduce the problem in the issue, but the code which updates the UI is definitely not really nice.

I am not very happy to use the queue - it shouldn't be necessary to perform all of the changes, but at this point I think we don't have to optimize this. If we see some weird chain of animations, then we should start fixing this.

Even though it might seem as a lot of changes from the diff, I didn't change the animations at all - I just made them awaitable.

Although everything should be perfectly OK, I would like some Android dev to confirm that I didn't break anything.


Toggl.Daneel/ViewControllers/MainViewController.cs
```diff
@@ -346,45 +333,92 @@ private void prepareViews()
...
-                        () => StartTimeEntryButton.Transform = CGAffineTransform.MakeScale(1f, 1f)
-                    );
-                });
+                    await showTimeEntryCard();
+                }
+                else
+                {
+                    await hideTimeEntryCard();
+                }
+            } while (true);
```
sofokli1: not sure about the `while (true)` also not a big fan of using `locks`. but i leave it up to you.

Toggl.Daneel/ViewControllers/MainViewController.cs
```diff
@@ -500,15 +533,15 @@ private void prepareSwipeGesturesOnboarding(IOnboardingStorage storage, IObserva
             swipeRightAnimationDisposable = swipeRightStep.ManageSwipeActionAnimationOf(firstTimeEntry, Direction.Right);
 
             swipeToContinueWasUsedDisposable = Observable.FromEventPattern(source, nameof(MainTableViewSource.SwipeToContinueWasUsed))
-                .Subscribe(_ =>
+                .Subscribe((EventPattern<object> _) =>
```
sofokli1: why change this?

simonrozsival: Because of conflicting extension methods. I will create a separate PR which will clean this up a bit.

Toggl.Daneel/ViewControllers/MainViewController.cs
simonrozsival: the reason for the `while(true)` is that the termination condition check is inside of a critical section and I don't want to lock the whole loop.

I don't like locks either, but I can't think of any lock-free solution which would guarantee thread safety.

simonrozsival:
@willsb suggested a different, simpler, solution. Let's close this without merging.

PR #2552, opened 2018-06-26
Move OnboardingSteps from Daneel to Foundation

juanlaube:
Requires functional testing in iOS 10 and 11 (so far I tested with iOS 11 and 12 in my devices).

`SwipeRightOnboardingStep` depended on UIKit to check the iOS version but I don't see any reason for it, I removed this dependency.


simonrozsival:
I am now a bit hesitant - should this be in `Foundation` or `Foundation.MvvmCross`? I almost thing it should be in MvvmCross even though it does not contain any UI logic, but the naming suggests that it should be used in some views and Foundation has no clue that there is something like views.

So either we rename the namespaces and classes, or move it to the `Mvvm` project. What do you think?

juanlaube:
Good point, I like this more in Foundation.MvvmCross even if this doesn't have any UI logic.

PR #2555, opened 2018-06-27
Fix invalid cast

### LINKED ISSUE #2544: InvalidCastException in AppDelegate.OpenUrl
This currently is responsible for almost half of the crashes in 1.7. This needs to be fixed for 1.7.1

```
System.InvalidCastException: Specified cast is not valid.
AppDelegate.OpenUrl (UIKit.UIApplication app, Foundation.NSUrl url, Foundation.NSDictionary options)
```
https://appcenter.ms/orgs/Toggl/apps/Toggl-iOS/crashes/groups/99a0be410390a3fd713b9531d812dd976005830d?appVersion=1.7

simonrozsival:
Fixes #2544 introduced in #2503 

Squash whenever


Toggl.Daneel/Startup/AppDelegate.cs
```diff
@@ -69,7 +70,8 @@ public override bool OpenUrl(UIApplication app, NSUrl url, NSDictionary options)
             var openUrlOptions = new UIApplicationOpenUrlOptions(options);
             var googleResponse = Google.SignIn.SignIn.SharedInstance.HandleUrl(url, openUrlOptions.SourceApplication, openUrlOptions.Annotation);
 
-            var facebookResponse = Facebook.CoreKit.ApplicationDelegate.SharedInstance.OpenUrl(app, url, (NSDictionary<NSString, NSObject>)options);
+            var facebookOptions = new NSDictionary<NSString, NSObject>(options.Keys.Select(key => (NSString)key).ToArray(), options.Values);
```
juanlaube: Maybe this can be done in another pull request later in develop, but we should early return if `googleResponse` is true. We don't want to give the login information that comes in the URL to Facebook 👀

Issue #1650, opened 2018-03-05
[onboarding] phase 4


pe0ny:
**This phase showcases continue and delete gestures.**
The animation entices users to see what's behind the TE, and invites them to swipe.

- [ ] Show this once user has at least 5 saved time entries.
- [ ] Users don't have to commit to the swipe (they can back out of a swipe if they don't "flick it"), so they don't have to carry out that action.
- [ ] Hide the animation once the user interacts with the time entry in question or taps the tooltip to dismiss.
- [ ]  At first show the continue hint and then the delete one.

### Animation example:
![peek](https://user-images.githubusercontent.com/5064445/36981376-b2d30638-2095-11e8-826f-49eeaf958404.gif)

### Swipe to continue
![quickstart phase 4 1v2](https://user-images.githubusercontent.com/5064445/42081281-b09c02a4-7b8d-11e8-89d8-3c36e0090ffb.png)

Zeplin: https://zpl.io/bAG5keP

### Swipe to delete
![quickstart phase 4 2v2](https://user-images.githubusercontent.com/5064445/42081286-b2b3aa2e-7b8d-11e8-8947-95ab6621a222.png)

Zeplin: https://zpl.io/aRN0glK

lizaPlotnikova:
Hey, I cleaned up the animation a little bit, so it doesn't seem so spastic.

![bounce to reveal gesturel](https://user-images.githubusercontent.com/24964497/37115002-ec1686b2-2217-11e8-8838-d66b7178d2b8.gif)

Here's the code from Framer: 

`animationA = new Animation sketch.entry_1,
	x: 50
	options: 
		time:.5
		curve: Bezier(.1,.2,.1,0)
	 
animationB = new Animation sketch.entry_1,
	x: 0
	options:
		time:.1
		curve: Bezier(.1,.1,.1,0)

animationC = new Animation sketch.entry_1,
	x:3.5
	options:
		time:.13
		curve: Bezier(.0,0.32,.7,1)

animationD = new Animation sketch.entry_1,
	x: 0
	options:
		time:.08
		curve: Bezier(.38,0.1,.7,1)
		
#Alternate between the four animations 
animationA.start()
animationA.on Events.AnimationEnd, animationB.start
animationB.on Events.AnimationEnd, animationC.start
animationC.on Events.AnimationEnd, animationD.start
animationD.on Events.AnimationEnd, ->
	Utils.delay .7, ->
		animationA.start()`

[...]

Issue #1649, opened 2018-03-05
[onboarding] phase 3


pe0ny:
**This phase will be shown to all users who have at least one saved time entry.**

- [ ] Show this after user has created (and stopped) at least one time entry.
- [ ] Show these hints for the first 25 times ever since user has created at least one time entry.
- [ ] Whenever the user types the @ or the # to autocomplete, we remove 5 "bonus" views. That way if they're using the feature they don't see the hints for as long.
*For example, if user has used @ or # to autocomplete 5 times, this means, we won't display the hints anymore.*

### Phase 3
![quickstart phase 3](https://user-images.githubusercontent.com/5064445/36980751-eb712e72-2093-11e8-918c-cfc022d8a5d5.png)

Zeplin: https://zpl.io/aM4qq77

lizaPlotnikova:
After much deliberation, an alternative design was created for this step;
 
This step will only be shown one time.

<img width="277" alt="screen shot 2018-04-02 at 9 25 44 am" src="https://user-images.githubusercontent.com/24964497/38197764-e19b7342-3657-11e8-955a-0afbbc765d37.png">

Zeplin: https://zpl.io/adAjA3K

juanlaube:
This requires a lot of reusable code from the phase 1 merged. The implementation should be really easy once we have that.

Issue #1647, opened 2018-03-05
[onboarding] phase 2


pe0ny:
**Show this phase to the new users after they have signed up for Toggl.**

**Step 1:** Tooltip prompting to edit
- [ ] Show once the user stops the first TE in app.
- [ ] Hide if user taps the tooltip to dismiss it OR opens the edit view OR navigates to any other view (reports, new TE, settings, etc.)

Note: If the user pulls more TEs from the server. E.g. Starts a new time entry on mobile, goes offline, creates bunch of new entries in webapp and pulls new data, we're still showing the tooltip.

**Step 2:** Project field tooltip in TE edit view
- [ ] Show when user opens the edit view for the first time AND TE doesn't have a project assigned
- [ ] Hide when user taps the tooltip to dismiss it OR when user taps the project field *(user doesn't have to assign the project to the TE, but we should hide the tooltip once they tap on the project field)* 
- **Clarification**: Tooltip should show **only the first time** user opens the Edit View for a TE that **does not have** a project assigned. If user closes the Edit View without either tapping the tooltip to dismiss it or tapping the project field, the tooltip **will not show** on subsequent openings of Edit View for that TE (or any other).

### Step1
![quickstart phase 2 1](https://user-images.githubusercontent.com/5064445/42080893-b4c44e6e-7b8c-11e8-8c47-3ae80876e209.png)

Zeplin: https://zpl.io/b6lMM6j

### Step2
![quickstart phase 2 2](https://user-images.githubusercontent.com/5064445/42080904-b8ac3a00-7b8c-11e8-9eac-168a09e6357c.png)

Zeplin: https://zpl.io/V1Zqq6w

pe0ny:
Note: Since the tooltip is above the whole table, not an inner part of it, we decided to hide/remove it when the user starts scrolling and TE leaves the view due to scrolling. This only happens when the user pulls more TEs from the server.

daividssilverio:
@pe0ny about step 2: should it show only for the **first time ever** the user enters the edit view when there's no project assigned to the TE?
If the user tries to edit a TE which doesn't have project set yet for the first time ever, sees the the tooltip, closes the screen and goes back to the same screen, should the user see the tooltip again?

IMO it's ok to see the tooltip until he/she uses the select project feature.

daividssilverio:
I was thinking about showing it until the user meets the criteria to hide the tooltip:

[...]

Issue #1646, opened 2018-03-05
[onboarding] phase 1


pe0ny:
**Show this phase to the new users after they have signed up for Toggl.**

**Step 1:** Tooltip in the timer page pointing a user to start the timer
- [x] Show when the user first signs up and sees the timer page
- [x] Hide if the user has tapped the tooltip to dismiss it OR if user tapped the start button.

**Step 2:** New TE view, once user taps "Start" for the first time
- [ ] Show when the user triggers the Start view for the first time.
- [ ] Show suggestions that are describing the essence of the Description field. Those suggestions are tappable.
- [ ] ~Save button at the bottom right is greyed out. When the save button (checkmark) is grey, it is not disabled, meaning in principle user can still tap on it to save the entry. We just want to emphasize the importance of the save button.~
~It will be grey only the first time when a user opens the start view. Other times there will be a green circle with a white checkmark.~

~**Step 3:** Checkmark button highlighted~
- [ ] ~Show after step 2 when user has typed or added at least one character to the field.~
- [ ] ~Highlight the save button.~

**Step 3:** Second tooltip in log view pointing to stop the time entry
- [ ]  Show after step 3: the first time when a user has a running TE in the app
- [ ] Hide if user taps the tooltip to dismiss it OR if user taps the stop button.

### Step1
![quickstart phase 1 1](https://user-images.githubusercontent.com/5064445/42080537-b1b3b7f6-7b8b-11e8-8cf5-b911fff7750a.png)

Zeplin: https://zpl.io/bAGqq6n

### Step2
![quickstart phase 1 2](https://user-images.githubusercontent.com/5064445/42080539-b620d6c0-7b8b-11e8-919b-528238d94b44.png)

Zeplin: https://zpl.io/bzYWW97

### Step3
![quickstart phase 1 4](https://user-images.githubusercontent.com/5064445/42080542-b997b6de-7b8b-11e8-82f3-a9baacb7e67e.png)

Zeplin: https://zpl.io/Vkr55ZG

daividssilverio:
@pe0ny , the layouts on Zeplin don't have that green button; We have that "Save" button on the toolbar only. Should I just apply the same idea of greying out the button text and making it white again when the user types something?

lizaPlotnikova:
[...]

PR #2556, opened 2018-06-27
Add tooltip layouts for Droid

daividssilverio:
Closes nothing.
Related to #1646 #1647 #1649 and #1650 

I added the layouts for the tooltips to be used with the PopupWindow.

squash please


daividssilverio:
@juanlaube , it should work now. It looks like the srcCompat used for vector drawables doesn't work with PascalCase. Seriously.

PR #2562, opened 2018-06-28
🚧 Use 'InputDelegate`

### LINKED ISSUE #2338: Figure out auto correction for all text fields
#2335 disabled auto correction in the start view, because it's a bit derpy. We should investigate and figure out if we can fix it properly, and then apply that fix to all text fields.

sofokli1:
closes #2338

I enabled the textView autocomplete and with the changes made it should work correctly. The solution is a bit obscure but is the best I could find. I set the `InputDelegate` of the textView to an empty implementation so the internal autocomplete logic does not get notified of the changes that we do internally to the textview, this helps with the issue where autocomplete is conflicting with the selection from the list below thus ending in setting the test from the selection and getting the autocompletion as well after the text as shown below
jun-27-2018 18-18-55

After the fix the autocorrect still works as seen below:
jun-28-2018 12-57-29

and the selection from the list as well
jun-28-2018 12-56-55

I set the `InputDelegate` every time the text changes because if set in `awakeFromNib` for some reason it becomes null after some characters are typed. So this needs some functional testing.

PS: to test this it helps to go into Settings > General > Keyboard > Text Replacement > Add new (like phrase: `Kettlebells` and shortcut: `Kett`)

🦑 whenever

### DIFF
--- a/Toggl.Daneel/ViewControllers/StartTimeEntryViewController.xib
+++ b/Toggl.Daneel/ViewControllers/StartTimeEntryViewController.xib
@@ -95,7 +95,7 @@
                         <constraint firstAttribute="height" constant="68" id="u0i-Qv-oYj"/>
                     </constraints>
                     <fontDescription key="fontDescription" type="system" pointSize="17"/>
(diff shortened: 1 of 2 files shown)


sofokli1:
waiting for #2498

PR #2565, opened 2018-06-28
Fix problems with the start button and the time entry card animations

### LINKED ISSUE #2482: Start time entry button does not appear

simonrozsival:
Closes #2482 

There were two problems:

1. the way we dealt with the observed changes of the visibility before we could animate anything was weird (and probably the cause of the issue in the first place)
2. if one animation started before the first animation ended, the UI would "explode" and look really weird

Please check if the way I am using cancellation token is OK... I don't have much experience with using them.

Running the app in the simulator is mandatory when reviewing.

Before:
fun_animation_show_hide

After:
boring_animation_show_hide

Squash whenever


Toggl.Foundation.MvvmCross/ViewModels/MainViewModel.cs
```diff
@@ -173,12 +175,21 @@ public void Init(string action)
...
+                    .Select(timeEntry => timeEntry != null);
+
+            var connectableTimeEntryCardVisibility
+                = IsTimeEntryRunningObservable
+                    .DistinctUntilChanged()
+                    .Replay(1);
+
+            connectableTimeEntryCardVisibility.Connect();
+
+            TimeEntryCardVisibility = connectableTimeEntryCardVisibility;
```
willsb: Why having two separate observables? I think `IsTimeEntryRunningObservable` should also be `DistinctUntilChanged`

Toggl.Foundation.MvvmCross/ViewModels/MainViewModel.cs
```diff
@@ -61,6 +61,8 @@ public sealed class MainViewModel : MvxViewModel
 
         public int NumberOfSyncFailures { get; private set; }
 
+        public IObservable<bool> IsTimeEntryRunningObservable { get; private set; }
```
willsb: Consider dropping the `Observable` suffix

Toggl.Daneel/ViewControllers/MainViewController.cs
```diff
@@ -57,15 +59,15 @@ public partial class MainViewController : MvxViewController<MainViewModel>
...
+        private CancellationTokenSource cardAnimationCancellation;
 
         private DismissableOnboardingStep tapToEditStep;
         private DismissableOnboardingStep swipeLeftStep;
         private DismissableOnboardingStep swipeRightStep;
 
         private UIGestureRecognizer swipeLeftGestureRecognizer;
 
-        private IDisposable cardDisposable;
+        private CompositeDisposable disposeBag = new CompositeDisposable();
```
willsb: Consider inheriting from ReactiveViewController, since we'll Rxfactor this eventually and this will allow us to have the dispose logic in a single point

simonrozsival: I don't think I want to make so many changes in this PR. Let's keep this for Rxfactoring.

Issue #2524, opened 2018-06-22
Performance issues that need improvement


pe0ny:
- Starting the app takes up to 9 seconds (3 seconds white screen, 6 seconds splash screen).
- Navigating from the Timer page to Reports takes ~3-5 seconds.

amulware:
These here take a second or two as well:
- opening start view
- opening edit view
- opening time popup

simonrozsival:
I noticed that closing the start view is annoyingly slow as well.

amulware:
I observe that all operations mentioned above involve poking the database - coincidence? (or is it just that every meaningful operation pokes the db..)

simonrozsival:
Do we somehow always run Realm queries on the UI thread? Seems unlikely, but if it was the case...

amulware:
I've seen nothing happening here in a while. Unassigning and prioritising.

PR #2570, opened 2018-06-28
Enable AOT for release builds

willsb:
Ref #2524 

This addresses the startup time issue. Please test the app as a whole, using the adhoc build from this branch in Bitrise and compare startup times.

Squash whenever.


ApocalypticFish:
Downloaded the Adhoc development build. It takes 3.5 seconds to open still after testing several times.

PR #2573, opened 2018-06-29
Fix putting two running time entries into the database

simonrozsival:
The problem was with tapping the continue button when the device is online then...

- the TEs will be pushed - created
- we will send a POST request
- we will stop the time entry locally (because we started another time entry in the meantime)
- we will override the local data with the response from the server - _no questions asked_
- this will turn the already stopped TE into a running one
- we have two running time entries
- next time we try to access the database (either in rivals resolution during `BatchUpdate`, or right after the start when we look up the current running time entry)

~WIP: This still needs proper testing + it should be merged into the second Neptune hotfix, which doesn't have a branch yet.~

I am not very happy with the fact that `RealmAdapter` would throw if it the `ChangeId` method is called with something that does not implement `IModifiableId` and there is no way to prevent it at compile time right now. I have to think about this and fix this on Monday somehow. Suggestions are welcome.

As a follow-up of this PR I would like to create a PR which would rename `IBaseStorage` to `IBaseRepository` and `ISingleObjectStorage` to `ISingletonRepository` to match the naming of the data sources.


Toggl.PrimeRadiant.Realm/Models/RealmConstructors.cs
```diff
@@ -262,9 +287,10 @@ public void SetPropertiesFrom(IDatabaseTimeEntry entity, Realms.Realm realm)
             Start = entity.Start;
             Duration = entity.Duration;
             Description = entity.Description;
+            RealmTags.Clear();
 
             var tags = entity.TagIds?.Select(id =>
-                realm.All<RealmTag>().Single(x => x.Id == id || x.OriginalId == id)) ?? new RealmTag[0];
+                               realm.All<RealmTag>().Single(x => x.Id == id || x.OriginalId == id)) ?? new RealmTag[0];
```
juanlaube: 👀

Toggl.Foundation/Sync/States/Push/UpdateEntityState.cs
```diff
@@ -39,23 +38,17 @@ internal sealed class UpdateEntityState<TModel, TThreadsafeModel>
...
-        private Func<TModel, IObservable<IConflictResolutionResult<TThreadsafeModel>>> tryOverwrite(TModel entity)
-            => updatedEntity => dataSource.OverwriteIfOriginalDidNotChange(
-            convertToThreadsafeModel(entity), convertToThreadsafeModel(updatedEntity));
+        private Func<TThreadsafeModel, IObservable<ITransition>> tryOverwrite(TThreadsafeModel originalEntity)
+            => updatedEntity
+                => dataSource.OverwriteIfOriginalDidNotChange(originalEntity, updatedEntity)
+                    .Select(result =>
+                        result is IgnoreResult<TThreadsafeModel>
+                            ? EntityChanged.Transition(originalEntity)
+                            : Finished.Transition(extractFrom(result)));
```
amulware: Why all these changes? Is there a change of behaviour?

Toggl.Foundation/Sync/States/Push/CreateEntityState.cs
```diff
@@ -3,44 +3,77 @@
...
+                ? Observable.Throw<TModel>(new ArgumentNullException(nameof(entity)))
+                : api.Create(entity);
+
+        private Func<TThreadsafeModel, IObservable<ITransition>> tryOverwrite(TThreadsafeModel originalEntity)
+            => serverEntity
+                => dataSource.OverwriteIfOriginalDidNotChange(originalEntity, serverEntity)
+                    .SelectMany(result =>
+                        result is IgnoreResult<TThreadsafeModel>
+                            ? updateId(originalEntity.Id, serverEntity.Id)
+                            : Finished.Transition(extractFrom(result)).Apply(Observable.Return));
```
amulware: This critical code looks good to me.

Toggl.PrimeRadiant.Realm/Models/IModifiableId.cs
```diff
@@ -1,9 +1,13 @@
-﻿namespace Toggl.PrimeRadiant.Realm.Models
+﻿using Toggl.Multivac.Models;
+
+namespace Toggl.PrimeRadiant.Realm.Models
 {
-    interface IModifiableId
+    interface IModifiableId : IIdentifiable
```
amulware: Why the inheritance?

Toggl.Foundation/Sync/States/Push/UpdateEntityState.cs
simonrozsival: The code is not shared with the `CreateEntityState` class anymore.

Toggl.Foundation/Sync/States/Push/CreateEntityState.cs
```diff
@@ -1,52 +1,86 @@
...
         private readonly ICreatingApiClient<TModel> api;
 
+        private readonly IDataSource<TThreadsafeModel, TDatabaseModel> dataSource;
+
         private readonly Func<TModel, TThreadsafeModel> convertToThreadsafeModel;
 
-        public StateResult<TThreadsafeModel> CreatingFinished { get; } = new StateResult<TThreadsafeModel>();
+        public StateResult<TThreadsafeModel> EntityChanged { get; } = new StateResult<TThreadsafeModel>();
+
+        public StateResult<TThreadsafeModel> Finished { get; } = new StateResult<TThreadsafeModel>();
```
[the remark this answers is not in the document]
simonrozsival: Theoretically, we don't. It would work even without that special case. Now that we have the graph visualization I am more confident in having more results which can be connected properly.

Toggl.PrimeRadiant.Realm/Models/IModifiableId.cs
simonrozsival: There was a conflict because of the `long Id` property.

amulware:
@simonrozsival I'm alright with this as it is. Did you still want to change it?

PR #2577, opened 2018-06-29
Android onboarding phase 1 step 2

daividssilverio:
Part of #1646 - Step 2.

This wraps around the current autocompleteProvider to provide suggestions for onBoarding.

WIP because it doesn't have to logic to enable/disable the use of the wrapper for the autocomplete provider.

This affects iOS, as the Phase 1 - Step 2 part was not implemented there.


daividssilverio:
@pe0ny and I discussed about Phase-1 Step-2, and since it’s not done on iOS already and it would hide `@ Search Projects` and `# Search Tags` suggestions on onboarding; and it get can confusing to see mixed results when the user already has some TEs and starts typing; we should drop it.
Closing the PR because the old step 2 became obsolete.

PR #2579, opened 2018-07-01
Rxfactor login view

arturs105:
1. This needs a review from both: iOS and Android devs.
2. Since i'm on vacation for the next few weeks, a volunteer is needed, that would address all change requests.

Squash whenever.


willsb:
Let me channel my inner Jennifer Lawrence to say that I volunteer
image)

Toggl.Foundation.MvvmCross/ViewModels/LoginViewModel.cs
```diff
@@ -95,28 +88,66 @@ public bool IsShowPasswordButtonVisible
...
+                    (passwordIsLongEnough, valueFromSubject) => passwordIsLongEnough && valueFromSubject);
+
+            HasError = ErrorMessage.Select(error => !string.IsNullOrEmpty(error));
+            LoginEnabled = Email
+                .CombineLatest(
+                    Password,
+                    IsLoading,
+                    (email, password, isLoading) => email.IsValid && password.IsValid && !isLoading)
+                .DistinctUntilChanged();
+            IsPasswordManagerAvailable = Observable.Create((IObserver<bool> observer) =>
```
simonrozsival: I don't understand why to do it this way. Am I missing something? 🤔

willsb: I also don't get it and I'll burn this

Toggl.Daneel/Views/LoginTextField.cs
```diff
@@ -28,6 +28,7 @@ public sealed class LoginTextField : UITextField
         private readonly CATextLayer placeholderLayer = new CATextLayer();
 
         private bool placeholderDrawn;
+        private bool placeholderIsUp;
```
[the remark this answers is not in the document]
willsb: I think it reads better when the `is` is in the middle:

```
private void movePlaceholderUp()
{
    if (placeholderIsUp) return;
    // ...
}
```

simonrozsival:
I noticed that when you enter the view when some of the items already have a value, the animation is kind of weird:

animated_placeholder

simonrozsival:
Other than what I posted in my previous comment, I think this is OK and merge ready. I will be on vacation so don't wait for my approval, but please do something about the animations.

Toggl.Foundation.MvvmCross/ViewModels/LoginViewModel.cs
```diff
@@ -86,148 +89,192 @@ public bool IsShowPasswordButtonVisible
...
+                    .Track(analyticsService.Login, AuthenticationMethod.EmailAndPassword)
+                    .Subscribe(onDataSource, onError, onCompleted);
         }
 
-        private void OnIsLoadingChanged()
+        public async Task StartPasswordManager()
         {
-            LoginCommand.RaiseCanExecuteChanged();
-        }
+            if (!passwordManagerService.IsAvailable) return;
```
juanlaube: `passwordManagerService.IsAvailable` always throws `NotImplementedException` in Giskard.
Either catch the exception (ugly) or just return false in the `IOnePasswordService` implementation (nicer).

Toggl.Foundation.MvvmCross/ViewModels/LoginViewModel.cs
```diff
@@ -86,148 +89,192 @@ public bool IsShowPasswordButtonVisible
...
+                return;
             }
-        }
 
-        private void onCompleted()
-        {
-            loginDisposable?.Dispose();
-            loginDisposable = null;
-        }
+            if (isLoadingSubject.Value == true) return;
```
daividssilverio: here you've used `isLoadingSubject.Value == true`, and on line 187, you've just used `isLoadingSubject.Value`. Which one do you prefer? 
It would be nice to keep the consistency if possible.

daividssilverio:
It's super weird... but that bug with the cursor getting back to beginning of the text as you type that @juanlaube reported seems to be caused by the way the bindings are being done. It looks like the the EditText Selection doesn't get enough time to update its selection position (I guess this shouldn't be happening).

I'll investigate it a little further, but something like this works (it needs type checking and boundaries check for the selection as well):
```
public static Action<string> BindText(this TextView textView)
            => text =>
            {
                var oldSelectionStart = textView.SelectionStart;
                textView.Text = text;
                if (oldSelection >= 0)
                {
                    ((EditText)textView).SetSelection(oldSelection, oldSelectionEnd);
                }
            };
```

PR #2596, opened 2018-07-03
Android onboarding phase 1 step 3

### LINKED ISSUE #1646: [onboarding] phase 1

daividssilverio:
Closes #1646

This is actually the last step of Phase 1, after some discussions with @pe0ny and because of some differences between android & iOS layouts on the Edit TE, the old steps 2 and 3 became obsolete.

🦑 whenever


Toggl.Giskard/Activities/MainActivity.cs
```diff
@@ -38,6 +38,7 @@ public sealed class MainActivity : MvxAppCompatActivity<MainViewModel>
         private FloatingActionButton stopButton;
         private CoordinatorLayout coordinatorLayout;
         private PopupWindow playButtonTooltipPopupWindow;
+        private PopupWindow stopButtonTooltipPopupWindow;
```
juanlaube: Maybe this should be in a `.ViewBinds.cs` file.

daividssilverio: 🤔 Kind of disagree, `.ViewBinds.cs` right now should be for layout views, stuff you get references to using FindViewById.
We could make it so, if the majority likes the idea. I don't like it because we are actually creating the view manually.

kornelijepetak:
Please rename this PR to Step 3

Toggl.Giskard/Extensions/PopupWindowWithDefaultGravityExtenstions.cs
```diff
@@ -25,5 +25,21 @@ public static class PopupWindowWithDefaultGravityExtenstions
             var verticalOffset = -(contentWindow.MeasuredHeight + Math.Abs(contentWindow.MeasuredHeight - anchor.Height) / 2);
             return new PopupOffsets(horizontalOffset, verticalOffset);
         }
+
+        public static PopupOffsets TopRightOffsetsTo(this PopupWindow popupWindow,
+            View anchor,
+            int dpExtraRightMargin = 0,
+            int dpExtraBottomMargin = 0)
```
kornelijepetak: `TopRightOffsetsTo` and then `dpExtraRightMargin` and `dpExtraBottomMargin`? Is it top, or bottom? I think the naming in this extension class is confusing. Could you consider a more comprehensible names?

daividssilverio: Hm.. what I meant with that this extension will provide OffSets for the popupWindow so that its BottomRight will be placed at the TopRight of the anchor, and you can extra bottom margin and extra right margins.

Would `TopRightOf(anchor, dpExtraRightMargin = 8, dpExtraBotttomMargin = 8)` be more meaningful to you?

daividssilverio: Or, BottomRightToTopRightOf(anchor)

kornelijepetak: A bit better. I'd just go with `From` instead of `of`. `TopRightFrom(anchor, ...)`

Toggl.Giskard/Activities/MainActivity.cs
juanlaube: Yeah, we agreed not to do this 👍

Toggl.Giskard/Extensions/PopupWindowWithDefaultGravityExtenstions.cs
daividssilverio: Ok!

PR #2602, opened 2018-07-04
Clean up sync graph labels

amulware:
This does two things: 
- it removes the suffix 'State' from sync states that have it (which are most)
- it splits sync state labels into two lines, where the type parameters are in the second line

So,
```
PersistListState<IWorkspace, IDatabaseWorkspace, IThreadSafeWorkspace>
```
becomes
```
PersistList
<IWorkspace, IDatabaseWorkspace, IThreadSafeWorkspace>
```

Similarly, for transitions, it splits off the transition parameter into a second line.

The resulting image is a little bit more compact this way, which is nice.

Before:
screen shot 2018-07-04 at 14 23 11

After:
screen shot 2018-07-04 at 14 24 16

EDIT:

Also made the nodes be rounded boxes instead of ovals, which makes them take up less space horizontally. Thanks @simonrozsival for that idea, and @kornelijepetak for splitting lines.


Toggl.Tools/SyncDiagramGenerator/DotFileWriter.cs
```diff
@@ -23,6 +28,39 @@ public void WriteToFile(string outPath, List<Node> nodes, List<Edge> edges)
...
+        private string cleanLabel(string label)
+        {
+            var typeSeparatorIndex = label.IndexOf('<');
+
+            var (name, types) = typeSeparatorIndex == -1
+                ? (label, null)
+                : (label.Substring(0, typeSeparatorIndex), label.Substring(typeSeparatorIndex));
+
+            var cleanName = name.EndsWith("state", StringComparison.OrdinalIgnoreCase)
+                ? name.Substring(0, name.Length - 5)
```
kornelijepetak: sometimes it's good to be explicit: `name.Length - "state".Length` instead of magic numbers.

Toggl.Tools/SyncDiagramGenerator/DotFileWriter.cs
```diff
@@ -23,6 +28,39 @@ public void WriteToFile(string outPath, List<Node> nodes, List<Edge> edges)
             File.WriteAllText(outPath, fileContent);
         }
 
+        private void cleanUpLabels(List<Edge> edges)
```
kornelijepetak: If `Edge` and `Node` would both implement `interface { Label }`, these two would be just one generic function 😄

amulware: Ain't nobody got time for generics.

Toggl.Tools/SyncDiagramGenerator/DotFileWriter.cs
amulware: Reusing constants? Please.

PR #2609, opened 2018-07-05
Add Linq

### LINKED ISSUE #2608: AppStore build fails
This is the error:
```
Startup/AppDelegate.cs(68,30): error CS1061: 'NSObject[]' does not contain a definition for 'Select' and no extension method 'Select' accepting a first argument of type 'NSObject[]' could be found (are you missing a using directive or an assembly reference?) [/Users/vagrant/git/Toggl.Daneel/Toggl.Daneel.csproj]
```

Looks like a missing using statement/namespace qualifier or something that we didn't catch because of `#if SOMETHING`

Odrakir:
Closes #2608 

Rider suggest to remove that using statement because it can't detect it being used between preprocessor tags. I added a comment just to prevent future deletions.


amulware:
Isn't there a special comment you can add that will make Rider not complain? ALT+Enter should offer to add that comment for you

Odrakir:
Not in a project wide way, as far as I know. I can do it for my own config only.

PR #2616, opened 2018-07-06
🚧 Add override to fix pop glitch

### LINKED ISSUE #1981: Copy, Paste, Select All fields in Password only display momentarily
The copy, paste and select all options only display for a fraction of a second on the password field so it's quite difficult to select all and delete or paste a password there.

Not sure if this is localised or device specific or even something we can change.

sofokli1:
Possible fix to close #1981

I personally can't reproduce the issues, but this looks like a possible fix in the case of subclassed UITextFields. It goes without saying ( but i am saying it anyways), if you can't reproduce the issues you can't review this PR.

If it works, 🦑 whenever. If not I will look deeper into this

### DIFF
--- a/Toggl.Daneel/Views/LoginTextField.cs
+++ b/Toggl.Daneel/Views/LoginTextField.cs
@@ -119,6 +119,9 @@ namespace Toggl.Daneel.Views
             return resignFirstResponder;
         }
 
+        public override bool CanBecomeFirstResponder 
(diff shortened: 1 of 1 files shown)


amulware:
It works fine in this branch, but just to make sure I checked the latest release and it worked fine there too. Very confusing, but let's close it for now?

Issue #1337, opened 2018-01-25
Support keeping data that the user has strictly speaking no access to


amulware:
We need this for the following cases:

- #1335 archived projects, before they have been fetched (the user really has access, but we still don't want to show them in any lists, we just have them to show in the log)
- #2593 losing access to a workspace, but retaining entities that unsynced TEs refer to
  - TEs still show in the log, but cannot be continued, and won't be synced
  - other entities (ws, client, project, task, tag) that are referred to be time entries are kept alive, but may not show up in any completions or suggestions

We need to think about whether these cases are actually handled the same or differently (probably the latter).

We have to make very sure that such entities are never exposed anywhere, except for where they have to be to render other data correctly.

- [ ] Never show such data in auto-completions or other lists
- [ ] Do not allow to continue a TE referring to such entities
- [ ] Suggestions should never include such entities

Further, on app start (or at another point), any such invalid entities that are no longer referred to by anything else, should be removed (also see #1336)

amulware:
It is probably best to start this with a proposal that takes all the linked issues into account, and review that before implementation to make sure we are not missing anything important.

amulware:
Two obvious options are:
- Another sync state, but then we might lose information on what objects were changed locally and what now, which might matter when access to a workspace is regained
- A separate boolean property, which requires a database migration (trivial if initialised to false)

There may be other options, and I want to see a discussion of all options and a recommendation of which is the best and why.

amulware:
Assigned to @juanlaube after internal discussion the other day.

juanlaube:
## A possible approach to this problem

When we pull sync happens, we cannot delete synced data in one step because data a project for example, is persisted before a time entry that could reference to it and is synced later.
Introducing a new `candidate_for_deletion` flag in all persisted data should help us mark all entities that should be deleted but we don't know yet if they can actually be deleted.


### Changes to be made in pull sync

* Fetch all since as usual
* Do not delete inaccessible data, instead mark as `candidate_for_deletion`
* When everything is persisted, we process deletions in reverse order of persisting
* Data that is `candidate_for_deletion` but referenced by time entries, stays in the database


### Changes to be made in push sync

* Just ignore things marked as `candidate_for_deletion`? 🤔 


### Regain access

When user regains access, we just set the `candidate_for_deletion` flag to false. Gaining access to a workspace should pull all other data again.


### Visible data

We need to filter everything that is `candidate_for_deletion`, except in the main log. Interactors FTW!


[...]

PR #2621, opened 2018-07-06
Check for no-workspace in database

### LINKED ISSUE #2592: Change no-workspace detection to check database instead
Right now we check if the workspace endpoint returns any workspaces, but that won't be a good check in the future when we'll allow the situation where the user has a local unsynced workspace, and no remote ones.

The check should instead check the local database, and see if there are any workspaces there that are non-ghost, as will be defined in #1337

I believe the current handling of the situation which causes the sync machine to freeze and a special exception to be thrown does not have to change.

sofokli1:
Closes #2592

Checks database for workspaces after pulling and persisting workspaces from backend.

🦑 whenever


Toggl.Foundation/Sync/States/Pull/NoWorkspaceDetectingState.cs
```diff
@@ -2,16 +2,24 @@
...
+        private readonly ITogglDataSource dataSource;
+
+        public NoWorkspaceDetectingState(ITogglDataSource dataSource)
+        {
+            this.dataSource = dataSource;
+        }
+
         public IObservable<ITransition> Start(IFetchObservables fetch)
-            => fetch.GetList<IWorkspace>()
+            => dataSource.Workspaces.GetAll()
```
juanlaube: Should this select only workspaces that are in sync or in sync needed? 🤔 Ping @amulware

amulware: This really depends on #1337. Right now those are the only two states that workspaces can be in, so I'd say this is fine as long as we remember to change it later according to what we do with #1337.

PR #2642, opened 2018-07-09
Fix editable token area bug

### LINKED ISSUE #2618: Start view: Description won't be saved when added after the project
1. Start new time entry
2. Add a project
3. *Cursor is now showed after the project token.* Start typing -> You can see that description is added inside the token.
4. Tap "Save"

Result: Description wasn't saved. Only the project.

willsb:
Closes #2618 

This is a droid bug and, as every fix that tries to solve a platform problem, the solution is not the prettiest. 

While investigating it was found out that the problem didn’t occur on projects ending with a number (e.g.: `Test Project 1`). Appending a `0` at the end of every project name solved the problem.  Since appending a 0 is not a valid solution, I’m appending a non-breaking space at the end (normal spaces don’t work, for whichever reason).

Squash whenever


Toggl.Giskard/Autocomplete/AutocompleteExtensions.cs
```diff
@@ -119,26 +120,28 @@ private static void AppendNormalText(this SpannableStringBuilder mutableString,
...
 
-        private static void AppendProjectText(this SpannableStringBuilder mutableString, ProjectSpan projectSpan)
+        private static void AppendProjectText(this SpannableStringBuilder spannable, ProjectSpan projectSpan)
         {
-            var start = mutableString.Length();
-            mutableString.Append(projectSpan.ProjectName);
-            var end = mutableString.Length();
+            var start = spannable.Length();
+            spannable.Append(projectSpan.ProjectName);
+            spannable.Append(unbreakableSpace);
```
juanlaube: Can you add a comment explaining why this is needed? Us in two or three months will thank you for this 😄

PR #2643, opened 2018-07-09
Trim email

### LINKED ISSUE #2561: Trim whitespace from emails in login/signup
It was reported that android's auto-completion may insert spaces after the email in the login form, which will cause the field to not validate and the login button to be disabled.

The simplest solution would be to trim the text before trying to turn it into a valid email.

sofokli1:
Closes #2561

Adds method to `Email` to trim whitespaces in the end and uses it at login

🦑  whenever


Toggl.Multivac.Tests/EmailTests.cs
```diff
@@ -67,5 +67,19 @@ public void ReturnsTheStringUsedForConstruction(string emailString)
                 email.ToString().Should().Be(emailString);
             }
         }
+
+        public sealed class TheTrimmedMethod
```
daividssilverio: The class should be named TheTrimmedEndMethod

Toggl.Multivac.Tests/EmailTests.cs
```diff
@@ -67,5 +67,19 @@ public void ReturnsTheStringUsedForConstruction(string emailString)
...
+        {
+            [Theory, LogIfTooSlow]
+            [InlineData(" email@with.spaceBothSides ")]
+            [InlineData(" email@with.spaceFront")]
+            [InlineData("email@with.spaceBack ")]
+            public void ReturnsNewEmailWithTrimmedEndSpaces(string mail)
+            {
+                var email = Email.From(mail).TrimmedEnd();
+
+                email.ToString().Should().Be(mail.TrimEnd());
```
daividssilverio: Wouldn't it be better to hide de details of the implementation and add the expected result in InlinedData as well? 
```
[InlineData(" email@with.spaceBothSides ", " email@with.spaceBothSides")]
[InlineData(" email@with.spaceFront", " email@with.spaceFront")]
[InlineData("email@with.spaceBack ", "email@with.spaceBack")]
```

Toggl.Multivac/Email.cs
```diff
@@ -22,8 +22,10 @@ private Email(string email)
 
         public override string ToString() => email;
 
-        public static Email From(string email) => new Email(email);
+        public Email TrimmedEnd()
```
daividssilverio: And why not just Trimmed? Your test was even named `TheTrimmedMethod`.

daividssilverio: Got it. 👍

PR #2644, opened 2018-07-09
Fix perceived sluggishness when starting a time entry

### LINKED ISSUE #2619: Start view text field is really slow and lags
When trying to type something in the Start view text field, it starts lagging and it's almost impossible to use. Can someone confirm?

Android
1.0 internal build

willsb:
Closes #2619 

Marshalling wasn’t enough, so I’m adding a small throttle; long enough to allow continuing typing yet small enough to feel instant once the user stops typing.

Functional testing in a real device (preferably a low-end one) is mandatory.

Squash whenever

### DIFF
--- a/Toggl.Giskard/Activities/StartTimeEntryActivity.cs
+++ b/Toggl.Giskard/Activities/StartTimeEntryActivity.cs
@@ -31,6 +31,8 @@ namespace Toggl.Giskard.Activities
               ConfigurationChanges = ConfigChanges.Orientation | ConfigChanges.ScreenSize)]
     public sealed partial class StartTimeEntryActivity : MvxAppCompatActivity<StartTimeEntryViewModel>, IReactiveBindingHolder
     {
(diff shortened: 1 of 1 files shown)


Toggl.Giskard/Activities/StartTimeEntryActivity.cs
```diff
@@ -51,8 +53,8 @@ protected override void OnCreate(Bundle bundle)
 
             editText.TextObservable
                 .SubscribeOn(ThreadPoolScheduler.Instance)
+                .Throttle(typingThrottleDuration)
```
kornelijepetak: Being accustomed to JS, I was confused to find out in the end that this is actually a Debounce named as a Throttle.
This is a bad decision on the part of Rx.NET team. 😢 
Bad confusing decision.

daividssilverio: It's Debounce on RxJava as well.

PR #2654, opened 2018-07-10
Fix issue when the continue action stays visible after swipe

### LINKED ISSUE #2649: After continuing with swipe, the row remains green
It would be great if someone could test it too and help me to figure out the exact steps how to reproduce. Atm it feels it only happens when you continue while another TE is running.

screenshot_20180709-173842

juanlaube:
Closes #2649 

Needs to be tested in a real device.

🦑 whenever.


amulware:
I can't test this, but if someone else can confirm that this fixes the issue, feel free to squash into `carbon`.

PR #2656, opened 2018-07-10
Debounce clicks in main log items to avoid EditTimeEntryActivity presented twice

juanlaube:
Closes nothing, needs testing in a device.

Squash whenever.

### DIFF
--- a/Toggl.Giskard/Views/MainRecyclerViewLogViewHolder.cs
+++ b/Toggl.Giskard/Views/MainRecyclerViewLogViewHolder.cs
@@ -1,26 +1,37 @@
 ﻿using System;
+using System.Reactive;
+using System.Reactive.Disposables;
+using System.Reactive.Linq;
+using System.Reactive.Subjects;
 using Android.Runtime;
 using Android.Views;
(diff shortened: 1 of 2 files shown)


Toggl.Giskard/Views/MainRecyclerViewLogViewHolder.cs
```diff
@@ -1,26 +1,37 @@
...
 using MvvmCross.Droid.Support.V7.RecyclerView;
+using Toggl.Foundation.MvvmCross.Extensions;
 using Toggl.Foundation.MvvmCross.ViewModels;
+using Toggl.Multivac.Extensions;
 
 namespace Toggl.Giskard.Views
 {
     public sealed class MainRecyclerViewLogViewHolder : MvxRecyclerViewHolder
     {
+        private static readonly TimeSpan editThrottleDuration = TimeSpan.FromMilliseconds(1000);
```
daividssilverio: 1 second is too much. We actually have to wait this 1 second for the click action to kick in (based on my tests, not the docs).
I'd suggest something like 250~300 ms.

PR #2657, opened 2018-07-10
Giskard 1.0 - Carbon #2

amulware:
This is an alternative version of the carbon 1.0 release which includes all changes from `develop` at time of writing, including the major Mvx update.

If testing checks out, we might go with this instead of the other one.


amulware:
We decided that this will indeed supersede the previous release. A release will be created from this branch, and the previous one will be killed.

amulware:
This has been released and can be merged.

Issue #2670, opened 2018-07-13
When using 24-hour format, it's almost impossible to select hours from the inner circle


pe0ny:
I'm afraid there's some bug when trying to select hours from the inner circle. This applies both to Start and End times. 
I compared the same component with Google Calendar's and there it works really smoothly - you can drag the pointer from inner to outer circle easily. Unfortunately, in our app, sometimes you can't select hours from the inner circle and you can't drag the pointer from the outer to the inner circle and vice versa.

This applies happens when accessing the date-time-duration popup from either the Start or Edit view - in both cases.

pe0ny:
The issue occured because we don't allow to set start time later than stop time or stop time earlier than start time. So when user tries to do so, the clock snaps back to the latest/earliest possible value.

I'm closing this at the moment but we should have a discussion to see if we can find a better solution for this. At the moment it's not clear why it snaps to other value (though we show the toast).

PR #2672, opened 2018-07-13
Swiper, no swiping.

### LINKED ISSUE #2671: Date-time-duration popup: disable swiping between the tabs
Let's disable swiping in that view because it makes really hard to change the time

kornelijepetak:
Fixes #2671 

Maybe even #2670 if I read it correctly.


daividssilverio:
It feels weird to have a view pager without horizontal scrolling. We already have to do some work around the time pickers because of the 12/24hours mode switch. 
I vote for having this view structure changed when we get the chance. We could use a simple view container, switch fragments/views with animations and it would be less problematic than what we have now.

Issue #2528, opened 2018-06-22
Investigate failing integration tests


amulware:
Our API integration tests have been failing a lot when running during the night recently, however, when I rerun them during the day, they are just fine.

Let's go over the logs or do some testing and see what tests fail, maybe it's something time-related.

If no bug that can be fixed is found, please at least post findings below.

simonrozsival:
These tests which were added recently fail repeatedly and it is probably related to being close to midnight:

```
Failed   Toggl.Ultrawave.Tests.Integration.TimeEntriesApiTests+TheGetAllBetweenStartAndEndDatesMethod.DoesNotReturnTimeEntriesWhichStartBeforeTheStartParameter
Error Message:
 Expected list to be empty, but found {Toggl.Ultrawave.Models.TimeEntry
   {
      ApiDuration = 300L
      At = <2018-06-26 01:14:12>
      Billable = False
      CreatedWith = <null>
      Description = "ea93bf35-5d31-4084-a165-b09f8323aa01"
      Duration = 300L
      Id = 907123786L
      ProjectId = <null>
      ServerDeletedAt = <null>
      Start = <2018-04-11 01:10:33>
      Stop = <2018-04-11 01:15:33>
      TagIds = <null>
      TaskId = <null>
      UserId = 4188420L
      WorkspaceId = 2817977L
   }}.
```

```
Failed   Toggl.Ultrawave.Tests.Integration.TimeEntriesApiTests+TheGetAllBetweenStartAndEndDatesMethod.ReturnsOnlyTheTimeEntriesWhichAreBetweenTheStartAndEndDates
Error Message:
 Expected list to contain 1 item(s), but found 2.
```

simonrozsival:
The other failing tests seem to be `BadGatewayException` and similar - so they are not problems in the logic or in the test themselves.

daividssilverio:
I checked the `DoesNotReturnTimeEntriesWhichStartBeforeTheStartParameter` with varying start times getting close to midnight to see what happens .

As it turns out, we are expecting time entries inside the range we requested to not be present on the response from the api.

For example: I got one failure with these values:
start time: `04/13/2018 21:17:21 -04:00 (04/14/2018 01:17:21 +00:00)`
time entry time: `04/12/2018 21:17:21 -04:00 (04/13/2018 01:17:21 +00:00)` (which is inside the range we are requesting)

The api call being made is:
```
[...]

PR #2677, opened 2018-07-13
Fix base authentication test for unauthenticated users flakiness

daividssilverio:
Closes nothing.

Part of #2528 

How does it remove the flakiness? By accepting either 401 and 403 possible status code responses when making requests with invalid credentials.

https://github.com/toggl/mobileapp/blob/develop/Toggl.Ultrawave/ApiClients/BaseApi.cs#L159 isn't enough because that piece of code expects the user to be logged in to handle 403 as an Unauthorized error, instead of an Unauthenticated error. The test affected by this PR is actually made with invalid credentials.


daividssilverio:
Integration tests: https://app.bitrise.io/build/62701b7b746ac17a

amulware:
Is our business logic also catching both exceptions? I.e. is the test in line with how the app behaves?

daividssilverio:
Well, actually, this test is kind of weird now that you've mentioned that.

Our business logic (inside Foundation and Foundation.MvvmCross) handles only `UnauthorizedException` directly. Otherwise, the code you mentioned will handle Forbidden exceptions by making another request and checking whether or not we are getting user-related status errors.

Anyway, `AuthenticatedEndpointBaseTests` is the base for a lot of api tests, and it covers all calls with ApiTokens and Simple authentication. And that's not how things are done in the code base. As far as I understood, we are only making `WithPassword` calls when doing Login/SignUp/RefreshToken, and after that api calls are made using api tokens.

Does it make sense to have the Base test as it is?

amulware:
I'm not sure I understand what you're saying. The code referenced (https://github.com/toggl/mobileapp/blob/develop/Toggl.Ultrawave/ApiClients/BaseApi.cs#L159) Should catch all `Forbidden` and turn them into `Unauthorized` if the `/me/logged` returns a 4xx error.

I just tested this locally, and that endpoint indeed returns a 403 if I use a wrong password or non-existent user.

So how can the 403 (Forbidden) ever escape Ultrawave?

daividssilverio:
I'm closing this PR because it does not actually fix anything.
As talked internally, the exception that used to make our tests fail actually happens on our tests while setting up the user for the test.
This has became quite stable and it might have been just a matter of bad timing with the staging db flush.
Since our tests are now running before the flush and after db switches, this should be fine.
We can look at it again if happens again.
This might still happen if you run your tests around the time backend flushes staging data.

daividssilverio:
Oh, just over communicating a little, but as you said, 403 (Forbidden) do not escape Ultrawave. They are wrapped around Unauthorized.

Issue #2662, opened 2018-07-11
Add build number to version in settings/about


amulware:
So we can uniquely identify builds.

amulware:
Unless overridden, I say we just include the build with the version like: `3.14 (1337)`

PR #2681, opened 2018-07-16
✨ Add build number to Setting page

### LINKED ISSUE #2663: Add build number to settings/about
So that we can uniquely identify builds.

khoi:
Closes #2663 and #2662 

Notes to reviewers:

- The build number in iOS app is set to `IOS_BUNDLE_VERSION` and later on replaced by CI. 
- Now that we have build number, should we put it in the user-agent along with the version?

screen shot 2018-07-16 at 6 07 34 pm

🐙 whenever


Toggl.Ultrawave/Network/UserAgent.cs
```diff
@@ -8,13 +8,15 @@ public sealed class UserAgent
 
         public string Version { get; }
 
-        public UserAgent(string agentName, string version)
+        public string BuildNumber { get; }
+
+        public UserAgent(string agentName, string version, string buildNumber = null)
```
willsb: I'd make this non-optional and get the build number for Droid as well, since it's not even that much work

amulware:
Have we tested/checked/thought through the consequences of including the build number in the user agent?

khoi:
@amulware I haven't changed the User Agent to use the build number. What sort of test should be run if I were to change it? 

Also I'm modifying the PR so it closes #2662 for Android also

amulware:
My bad, at a glance I thought we would include the build number in requests as well. But if we're not doing so, nothing can break.

However, I'm not really sure I like modifying Ultrawave code for something that is very app specific and has nothing to do with our API. Unless someone has a good argument here, I'd like that done differently.

amulware:
And I didn't really answer the other question, did I?

About including the build number in the user agent string... maybe? It would probably be fine, but I just don't know if I'm happy including these semantics in Ultrawave. If we would make this change however, we should make sure that something sensible is passed there with our integration tests.

khoi:
so I removed all the UserAgent thingy, and add the build number to `PlatformConstants` instead.

Toggl.Foundation.Tests/MvvmCross/ViewModels/SettingsViewModelTests.cs
```diff
@@ -500,7 +500,7 @@ public sealed class TheVersionProperty : SettingsViewModelTest
             [Fact, LogIfTooSlow]
             public void ForwardsTheValueFromTheUserAgentUsedInConstruction()
```
amulware: Please change test name appropriately.

PR #2590, opened 2018-07-03
Neptune Hotfix 2

simonrozsival:
Closes #2572 

Changelog:
```
- Fixed crashes related to the app accidentally having two running time entries
```

PR #2554, opened 2018-06-27
Remove decorators from sync

simonrozsival:
There are now no more decorators so @amulware should be happy now 😄 

Squash whenever

PR #2684, opened 2018-07-16
Merge Neptune hotfix 2 into Develop

### LINKED ISSUE #2572: Release Daneel 1.7.2
This is the fix for the repeated app-start crash that can be reproduced by tapping continue buttons quite fast.

kornelijepetak:
Closes #2572

Supersedes #2590 

This PR is actually what #2590 was supposed to do, but we had serious merge conflicts between `develop` and the `neptune-hotfix-2`.

So, these are the relevant PRs that created such conflicts:

#### On `develop`:
* #2497 Add metrics for checking status of entities push sync
* #2554 Remove decorators from sync
* #2504 Fix update with conflict resolution

#### On `neptune-hotfix-2`:
* #2573 Fix putting two running time entries into the database
* #2600 Stop one time entry if two are running

The main issue in conflicts was that the resolver now can return multiple resolution results, so the interfaces changed.

The last commit is the branch merge conflict resolution, so it's the one that needs special care during review.


ApocalypticFish:
It's looking really solid. I've tested syncing pretty thoroughly now, reports are displaying well, and settings are syncing between apps. No crashes when I spam continue either.

Issue #2673, opened 2018-07-13
Find out why Android appears to be crash-free


amulware:
There haven't been any crashes reported on App Center on Android 1.0 since the release. With 3k+ new users right now, that seems extremely unlikely.

My best guess is that something went wrong with the MVX merge into the release, or with the MVX update or other recent changes to develop in general.

PR #2688, opened 2018-07-16
🚧 Move call to external services to very beginning

sofokli1:
Doesn't close any issues. AppCenter, firebase and the rest were being called after the views were setup. It was breaking the remoteConfigure (a/b testing) and also very possible lose of app startup crashes data.

credit for the solution goes to @juanlaube 

🦑 whenever


kornelijepetak:
Maybe something similar happens on Giskard as well, causing #2673?

PR #2691, opened 2018-07-17
Hide add to duration shortcuts when editing duration

### LINKED ISSUE #2446: Editing duration and shortcut buttons not playing well together
For example:

- have TE of 10 minute duration
- go to edit it using numpad input
- type in '20' (end time does not change in popup)
- hit +30m shortcut (end time changes as if duration was 40m, duration in text field stays 20)
- hit ok (duration is set to 40)

I think it should modify the entered time when using shortcuts, or we should hide shortcuts then.

daividssilverio:
Closes #2446

This is not ideal, but I I'm hiding the shortcuts when the user clicks on the duration to start editing using the soft input keyboard.

To help understand how the widgets are organized.
`DurationEditText` is the bright white text you see after clicking the current duration TextView (grey text).

I tried binding the edition of the `DurationEditText` involved to the shortcut's commands, but as I started getting somewhere, I got confused on how this widget works right now... When editing, the XX:YY text the user sees as he/she is typing actually mean X hours and Y minutes, and clicking the duration when its current value is for example 03:00:00 (three hours), would render 03:00 on the `DurationEditText`, which looks like 3 minutes to me.

+1 for that conversation about changing the popup on Android with @kornelijepetak @pe0ny and Leez.

I rather have this easy fix, and RxFactor + re-think this view before changing its behaviour.

It looks like this:


Toggl.Giskard/Fragments/SelectTimeFragment.cs
```diff
@@ -89,6 +90,7 @@ public override View OnCreateView(LayoutInflater inflater, ViewGroup container,
 
             subscribeAndAddToDisposableBag(nameof(ViewModel.StopTime), onStopTimeChanged);
             subscribeAndAddToDisposableBag(nameof(ViewModel.StartTime), onStartTimeChanged);
+            subscribeAndAddToDisposableBag(nameof(ViewModel.IsEditingDuration), onDurationEditionStateChanged);
```
kornelijepetak: `Editing`, not `Edition`.

amulware:
You'll bring the buttons back and make them work as asked for in the issue with the rxfactor then?

kornelijepetak:
@amulware Buttons are there when duration edit input does not have focus. But yes, we want to avoid double work, so we'd like to wait until we agree what to do with that popup.

PR #2692, opened 2018-07-18
Adjust maximum modal presentation height

### LINKED ISSUE #2651: Start view is a bit hidden when "Hotspot" bar is visible
We should move it down a bit when the bar is visible.

image from ios 3

khoi:
Closes #2651 

🦑whenever.


khoi:
I've updated the code to make the presenting modal adapt to changes in the status bar frame. This is the first time we use `NSNotificationCenter` API in our code, so I'd like someone from @toggl/ios-senseis to take a look at this PR 🙇‍♂️.

Toggl.Daneel/Presentation/Transition/ModalPresentationController.cs
```diff
@@ -80,6 +85,19 @@ public override void DismissalTransitionWillBegin()
             coordinator.AnimateAlongsideTransition(_ => dimmingView.Alpha = 0.0f, null);
         }
 
+        public override void PresentationTransitionDidEnd(bool completed)
+        {
+            base.PresentationTransitionDidEnd(completed);
+            NSNotificationCenter.DefaultCenter.AddObserver(UIApplication.WillChangeStatusBarFrameNotification,
```
sofokli1: would it make sense to check if `completed` before the `notificationCenter` code?
what happens if you `...AddObserver(...` more than one time?

Toggl.Daneel/Presentation/Transition/ModalPresentationController.cs
```diff
@@ -80,6 +85,19 @@ public override void DismissalTransitionWillBegin()
...
+                OnStatusBarFrameChanged);
+        }
+
+        public override void DismissalTransitionDidEnd(bool completed)
+        {
+            base.DismissalTransitionDidEnd(completed);
+            NSNotificationCenter.DefaultCenter.RemoveObserver(UIApplication.WillChangeStatusBarFrameNotification);
```
sofokli1: same as above

Toggl.Daneel/Presentation/Transition/ModalPresentationController.cs
```diff
@@ -118,5 +136,10 @@ private void dismiss()
             PresentedViewController.DismissViewController(true, null);
             onDismissedCallback();
         }
+
+        public void OnStatusBarFrameChanged(NSNotification notification)
```
sofokli1: shouldn't this be private?

khoi:
In Swift, the observer method has to be `public`. :D Turned out you didn't need it for C# 👍

Toggl.Daneel/Presentation/Transition/ModalPresentationController.cs
```diff
@@ -118,5 +134,10 @@ private void dismiss()
             PresentedViewController.DismissViewController(true, null);
             onDismissedCallback();
         }
+
+        private void OnStatusBarFrameChanged(NSNotification notification)
```
willsb: private methods start in lowercase

Toggl.Daneel/Presentation/Transition/ModalPresentationController.cs
```diff
@@ -10,10 +11,26 @@ namespace Toggl.Daneel.Presentation.Transition
...
-
         private readonly Action onDismissedCallback;
-        private readonly nfloat maximumHeight;
+
+        private nfloat maximumHeight
+        {
+            get
+            {
+                var statusBarHeight = UIApplication.SharedApplication.StatusBarFrame.Height;
+                nfloat distanceFromTop;
```
willsb: If you `var distanceFromTop = UIApplication.SharedApplication.StatusBarFrame.Height;` you will not need the else branch and can simplify the if branch to `distanceFromTop += UIApplication.SharedApplication.KeyWindow.SafeAreaInsets.Top `

Toggl.Daneel/Presentation/Transition/ModalPresentationController.cs
kornelijepetak: The Exception being Fody/PropertyChanged `On<Prop>Changed` methods, but this does not seem to be the case here.

PR #2696, opened 2018-07-18
Override CloseFragmentDialog to check for nullity to avoid crashes

### LINKED ISSUE #2686: App crashes on restore from Edit TE screen when the duration popup is open
Steps to reproduce:

1 - Tap any existing TE from the main log
2 - Tap the Start/Stop areas to open the Edit Start/Stop/Duration popup.
3 - Press Home
4 - 💥

daividssilverio:
Closes #2686 

The source of the problem is most likely caused by a bug on the Mvx implementation used to close dialog fragments when sending the current activity to background.

What is happening is that by the time the presenter tries to close the DialogFragment used for the `SelectTimeFragment` (edit time popup), the references for the Activity and the FragmentManager are lost or set to null, causing the NullPointerException at Line 439 on the superclass `MvxAppCompatViewPresenter`.

```
var toClose = CurrentFragmentManager.FindFragmentByTag(tag);
```

Functional testing an actual devices is a must. Already tested on Nexus 5 (Android 23).

🦑 whenever


Toggl.Giskard/Presenters/TogglPresenter.cs
```diff
@@ -50,5 +52,17 @@ public override void ChangePresentation(MvxPresentationHint hint)
 
             base.ChangePresentation(hint);
         }
+
+        protected override bool CloseFragmentDialog(IMvxViewModel viewModel, MvxDialogFragmentPresentationAttribute attribute)
+        {
+            string tag = FragmentJavaName(attribute.ViewType);
```
willsb: Why not `var`?

Toggl.Giskard/Presenters/TogglPresenter.cs
```diff
@@ -50,5 +52,17 @@ public override void ChangePresentation(MvxPresentationHint hint)
...
+            var toClose = CurrentFragmentManager?.FindFragmentByTag(tag);
+            if (toClose != null && toClose is DialogFragment dialog)
```
willsb: The first null check is not needed

Toggl.Giskard/Presenters/TogglPresenter.cs
```diff
@@ -50,5 +52,17 @@ public override void ChangePresentation(MvxPresentationHint hint)
...
+            {
+                dialog.DismissAllowingStateLoss();
+                return true;
+            }
+            return false;
```
willsb: One whitespace above this line, please

Toggl.Giskard/Presenters/TogglPresenter.cs
daividssilverio: This was copy/paste from the Mvx source code. I'm fixing the style now.

Toggl.Giskard/Presenters/TogglPresenter.cs
daividssilverio: Indeed.

PR #2700, opened 2018-07-19
Improve feedback api tests

amulware:
## What?
This makes the data sent to our feedback API when running integration tests much more meaningful and useful, and also provides some instructions for support agents, so that they are not left with a big question mark.

It also standardises the emails used to be in line with the rest of our tests.

## Why?
Because yes, these emails actually get to our support people, and since they do, we might as well use them to confirm that the data we send is received correctly.

## How?
See the changes. It's very straight-forward. Note how I included a bunch of 'weird stuff' like emojis, markdown, HTML, and a link. Just to make sure that if our users do any of that, things don't explode on the other end.

## Fun fact
Coincidentally, while looking into this I found that emails are not well formatted after they get through to support. Specifically, the metadata is not deserialized into the message body well. All looks good on our side, so I'll be bugging backend about that.


amulware:
Integration tests are about to pass here: https://app.bitrise.io/build/d81499d4c26cc47d

amulware:
✅Integration tests successfully ran here: https://app.bitrise.io/build/6a0320f5dd722c32

PR #2706, opened 2018-07-19
🚧 Replace  `BindTap` with a proper `targetBinding`

### LINKED ISSUE #2623: [iOS] Description field can no longer be activated by taps
(https://toggl.desk.com/agent/case/735407)

1. Tap on an existing Time Entry on iOS 1.7.1
2. Attempt to tap repeatedly on Description field to edit it; nothing happens
3. Long-pressing on the Description field finally opens up keyboard and shows the text field cursor
4. Only long-pressing on the field to open up the magnified cursor selector lets you move the cursor around. Tapping on the text to move the cursor around doesn't work either.

Replicated on iPhone 8 Plus, running on 11.4 (15F79)

sofokli1:
Closes #2623

The issues is related to the description textview in when editing a time entry, it becomes first responder only with long press. The issues was that the binding was adding a tap gesture recogniser in order to inform the viewModel that the description textfield was selected(firstResponder) thus stilling the touch from the textfield and not allowing it to actually become first responder. I create a new target binding that checks for the new `EventHandler DidBecomeFirstResponder` in the view. Now when it actually becomes firstResponder the corresponding command in the viewModel is called.

🦑 whenever


Toggl.Daneel/ViewControllers/EditTimeEntryViewController.cs
```diff
@@ -111,8 +111,8 @@ public override void ViewDidLoad()
                       .To(vm => vm.Description);
 
             bindingSet.Bind(DescriptionTextView)
-                      .For(v => v.BindTap())
-                      .To(vm => vm.StartEditingDescriptionCommand);
+                      .For(v => v.BindDidBecomeFirstResponder())
+                      .To(vm => vm.StartEditingDescriptionCommand).TwoWay();
```
juanlaube: I'm not why sure the `.TwoWay()` binding is needed here...
btw, the `MvxTargetBinding` class has a default binding mode you can override to use a two way binding by default

sofokli1: that is a mistake i forgot to remove

Issue #2676, opened 2018-07-13
Make sure the app looks good regardless of font/display size settings


kornelijepetak:
Here is the list of items to take special care about (or the ones that some issues are already noticed with)

* [x] Time Entry cell in TE Log
* [x] Suggestions in TE Log
* [x] Running TE card
* [ ] The Time Selector (header labels)
* [ ] List of tags/projects/clients
* [ ] Calendar control in reports
* [ ] Settings screen (item's title + description label)
* [ ] Login (forgot password padding)
* [x] Login (sign in with google hidden)
* [x] Login (space between "Don't have an account/ sign up for free" texts and the bicycle icon)
* [x] Sign up (space between "Already have an account / Login" texts and the bicycle icon)
* [ ] Sign up (sign up with google hidden)
* [ ] Terms and privacy popup on sign up
Please fix each and every one of those in **separate** PR.

kornelijepetak:
In case of EditTE View I believe the best course of action would be to simply move Stop time into its own row on large font/display settings. This is how it looks like on `Font:Largest`, `Display:Largest`

![image](https://user-images.githubusercontent.com/7033216/42872111-55f9c996-8a7c-11e8-8dcc-08781604698a.png)

daividssilverio:
The problem with the bottom cards can be reproduced by having Font:Largest.
The problem with the login with google not showing can be reproduce by having Display:320 dpi.
Login (making it scrollable should be enough to show the login with google button):
<img src="https://user-images.githubusercontent.com/2173493/42908058-d29da59a-8aad-11e8-9f79-d65baee45c5a.png" width="240"/>

Sign up:
<img src="https://user-images.githubusercontent.com/2173493/42908059-d2c7ac6e-8aad-11e8-8f5e-39bce8fa7427.png" width="240"/>

amulware:
:delorean:

PR #2707, opened 2018-07-19
Fix signup hint card layout on small phones with larger fonts

daividssilverio:
Closes nothing yet.
Part of #2676 - Sign up (space between "Already have an account / Login" texts and the bicycle icon)

~It's basically the same as https://github.com/toggl/mobileapp/pull/2697, but I've also changed the hint text to make it fit smaller screens with larger fonts.~

🦑 whenever, ~but it would be nice to check if~ changing the text is ok ~first~.

what has been made on https://github.com/toggl/mobileapp/pull/2697 needs to be updated to match the fixed `dp` fonts on the bottom card here.


daividssilverio:
Updated to use the text "Have an account" only on mdpi and ldpi devices, "Already have an account" otherwise o/

Toggl.Giskard/Resources/values-mdpi/Strings.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <string name="AlreadyHaveAnAccount">Already have an account?</string>
```
kornelijepetak: In the PR description, you are saying that you are showing a short version for `ldpi` and `mdpi`. It seems that you are defining the long version on `mdpi` here. I assume all the greater densities will use this one as it best matches?

daividssilverio: You assumption is right. I did try the app on ldpi, mdpi and hdpi densities (everything was working). I'll triple check just in case.

kornelijepetak: But if feels wrong. I don't know how fallback values are used. I expected the system to work like this:

* if current density xml file exists and has a value specified, use it
* if not, use what's in the density-unspecified xml

However, in that case, I would expect the value in `values/Strings.xml` to be a fallback value. So since ldpi does not have it's own file, I would expect it to have the value from there. And it apparently does. (shows the shorter version). But then again, I would expect the same for hdpi. However, that one takes the value from values/mdpi. Hence the confusion... 🤔 

So it seems that it takes the 'best match' and this is what I'm afraid of the most. What is the 'best match'? Does it just find the greatest available density?

daividssilverio: Android tries to select the best match.
Here are some docs about that: https://developer.android.com/guide/topics/resources/providing-resources#BestMatch and https://developer.android.com/training/multiscreen/screensizes.

We can always define all files or just use `-w(enough space to fit text)` instead of the `-mdpi`. But I think it's safe this way.

daividssilverio:
As discussed internally with Leez, I've reverted the alternative string file and used fixed `dp`s for the font size on this card. It looks pretty nice on all densities; and, since this is a secondary action, it's ok to have it with a non-scalable font size.

Also, took the opportunity to convert it to a ConstraintLayout to arrange the TextView and Bike image in a more meaning way.

PR #2708, opened 2018-07-19
Make login screen scrollable

daividssilverio:
Closes nothing.
Part of #2676 - Login (sign in with google hidden)

It fix the problem by making the login screen scrollable. It looks like a lot of changes, but I'm only wrapping the main RelativeLayout in one ScrollView.

🦑 whenever, but requires tests on actual devices, preferably on multiple devices with different dimensions.


Toggl.Giskard/Resources/layout/LoginActivity.axml
```diff
@@ -1,166 +1,179 @@
...
+    android:fitsSystemWindows="true"
     android:layout_width="match_parent"
     android:layout_height="match_parent">
-    <ImageView
-        android:scaleType="centerCrop"
-        android:src="@drawable/bg_noise_login"
+    <ScrollView
+        android:scrollbars="none"
+        android:fillViewport="true"
+        android:paddingBottom="60dp"
```
juanlaube: When everything is perfectly visible, this line makes it possible to still scroll a little. However, removing this doesn't let you scroll. I think the fix could be adding bottom margin to the google login button instead of the scrollview, maybe? 🤔

daividssilverio: I'll take a look again and consider your suggestion.

daividssilverio: Placing the margin bottom on the google login button does help, the extra scroll is partially removed. But on screens where it kind of fits, it still scrolls a little. I'm committing your suggestion.

PR #2710, opened 2018-07-20
Custom Feedback UI

### LINKED ISSUE #2626: Custom feedback UI
- Use the Start view component for the main Feedback view which can be opened from the Settings view or Timer page (when asking rating).
- When the form submit succeeded, we’ll close the view and show “Done” message at the bottom of the previous page (either Settings or Timer page).
- When something went wrong, we just close the keyboard and display the error message at the bottom of the Feedback screen.
- If the user hits the back button and he/she has entered text already, we should ask for confirmation. (The logic should be the same as for the Start view.)

### Feedback UI
toggl - ios - feedback placeholder 1x

Zeplin: https://zpl.io/a8Koj7x

### Feedback UI (text entered)
toggl - ios - feedback 1x

Zeplin: https://zpl.io/V41pZzJ

### Confirmation screen
toggl - ios - feedback confirmation 1x

Zeplin: https://zpl.io/V0yQepK

### Done
toggl - ios - feedback done 1x

Zeplin: https://zpl.io/aMqJBjr

### Error states
toggl - ios - feedback error1 1x toggl - ios - feedback error2 1x

Zeplin: https://zpl.io/2ZJQ5Bd & https://zpl.io/VDzJGKr

khoi:
Closes #2626 

I've temporarily removed all `asDriver` call in my `SendFeedbackViewModel` for couple of reasons:

- I was having the problems where tests not running properly due to missing `SynchronizationContext`.

- We're currently not calling `asDriver` in the ViewModel, but during `Bind` in `ReactiveBindingHolderExtensions` which might force the ViewModel to do extra work, since the subscriptions are not `shared()`

When we sort out the `Driver` thingy, we can have another issues to update our ViewModels.


Toggl.Foundation.MvvmCross/ViewModels/Settings/SendFeedbackViewModel.cs
```diff
@@ -0,0 +1,117 @@
...
+        private ISubject<bool> isLoading = new BehaviorSubject<bool>(false);
+
+        // Inputs
+        public ISubject<Unit> CloseButtonTapped = new Subject<Unit>();
+        public ISubject<string> FeedbackText = new Subject<string>();
+        public ISubject<Unit> ErrorViewTapped = new Subject<Unit>();
+        public ISubject<Unit> SendButtonTapped = new Subject<Unit>();
+
+        // Outputs
+        public IObservable<bool> IsEdittingFeedback { get; }
```
Odrakir: The name might be a bit misleading, this property just indicates if there's text or not.

Toggl.Foundation.MvvmCross/ViewModels/Settings/SendFeedbackViewModel.cs
```diff
@@ -0,0 +1,117 @@
...
+            IDialogService dialogService
+        )
+        {
+            Ensure.Argument.IsNotNull(interactorFactory, nameof(interactorFactory));
+            Ensure.Argument.IsNotNull(navigationService, nameof(navigationService));
+            Ensure.Argument.IsNotNull(dialogService, nameof(dialogService));
+
+            IsEdittingFeedback = FeedbackText
+                .DistinctUntilChanged()
+                .Select(text => !string.IsNullOrEmpty(text))
```
Odrakir: You probably want the `DistinctUntilChanged` after this line, right?

Toggl.Foundation.MvvmCross/ViewModels/Settings/SendFeedbackViewModel.cs
```diff
@@ -0,0 +1,117 @@
...
+            var sendFeedbackObservable =
+                SendButtonTapped
+                    .WithLatestFrom(FeedbackText, (_, text) => text)
+                    .Do(_ => { isLoading.OnNext(true); })
+                    .SelectMany(text => interactorFactory
+                        .SendFeedback(text)
+                        .Execute()
+                        .Materialize()
+                    )
+                    .Share();
```
Odrakir: This is probably a matter of taste, but I would rather have only one Observable sequence, instead of sharing. If the `if nt.Kind ==` is too much to go inline here, I would try with a private method in the viewmodel.

Odrakir: I think having the two observables sequences is much less readable.

Toggl.Daneel/ViewControllers/Settings/SendFeedbackViewController.cs
```diff
@@ -0,0 +1,74 @@
...
+            this.Bind(ErrorView.Tapped(), ViewModel.ErrorViewTapped);
+            this.Bind(SendButton.Tapped(), ViewModel.SendButtonTapped);
+
+            this.Bind(ViewModel.IsEdittingFeedback.Select(e => !e), FeedbackPlaceholderTextView.BindIsVisible());
+            this.Bind(ViewModel.ErrorViewVisible, ErrorView.BindIsVisibleWithFade());
+            this.Bind(ViewModel.SendEnabled, SendButton.BindIsEnabled());
+
+            this.Bind(ViewModel.IsLoading.Select(loading => !loading), SendButton.BindIsVisible());
+            this.Bind(ViewModel.IsLoading, activityIndicatorView.BindIsVisible());
+            this.Bind(ViewModel.IsLoading, UIApplication.SharedApplication.BindNetworkActivityIndicatorVisible());
```
Odrakir: The problem with this is our bad implementation of `Bind` and the lack of `AsDriver` in the view models. I suggest using a local var:

```c#
var isLoading = ViewModel.IsLoading.AsDriver();
this.Bind(isLoading.Select...
...
```

Toggl.Foundation.MvvmCross/ViewModels/MainViewModel.cs
```diff
@@ -267,8 +267,8 @@ public void Init(string action)
                     break;
             }
 
-            ratingViewExperiment
-                .RatingViewShouldBeVisible
+            Observable.Return(true)
+                .Delay(TimeSpan.FromSeconds(5))
```
sofokli1: Why change these 2 lines. It breaks the rating view presentation logic by always show the rating view.

sofokli1: I guess for testing.

khoi: my bad, it was used for testing purposes

khoi:
in order to test the error message, modify `FeedbackApiClient`

```
public IObservable<Unit> Send(Email email, string message, IDictionary<string, string> data) {
                 return Observable.Create<Unit>(async observer =>
            {
                await Task.Delay(3000);
                observer.OnError(new Exception());
            });
}
```

Toggl.Daneel/ViewControllers/MainViewController.cs
```diff
@@ -230,6 +230,16 @@ public override void ViewDidLoad()
...
 
+            ViewModel.RatingViewModel
+                .IsFeedbackSuccessViewShowing
+                .ObserveOn(SynchronizationContext.Current)
+                .Do(showFeedback =>
+                {
+                    SendFeedbackSuccessView.Hidden = !showFeedback;
+                })
+                .Subscribe()
+                .DisposedBy(disposeBag);
```
arturs105: These lines do the same thing as would `this.Bind(ViewModel.RatingViewModel.IsFeedbackSuccessViewShowing, SendFeedbackSuccessView.BindIsVisible());`

Though for this to work, you need to change the base class to `ReactiveViewController<MainViewModel>`.

You can take a look at the `Bind` implementation to see that it's roughly the same code.

Toggl.Daneel/ViewControllers/MainViewController.cs
```diff
@@ -385,6 +395,12 @@ private void prepareViews()
             RunningEntryDescriptionFadeView.FadeLeft = true;
             RunningEntryDescriptionFadeView.FadeRight = true;
 
+            // Send Feedback Success View Setup
+            SendFeedbackSuccessView.Hidden = true;
+            var tapToDismissFeedbackGestureRegconizer =
+                new UITapGestureRecognizer(() => ViewModel.RatingViewModel.CloseFeedbackSuccessView());
+            SendFeedbackSuccessView.AddGestureRecognizer(tapToDismissFeedbackGestureRegconizer);
```
arturs105: This can be RxFactored into `this.BindVoid(SendFeedbackSuccessView.Tapped(), ViewModel.RatingViewModel.CloseFeedbackSuccessView);`. And since it's a binding now, it should be moved into `ViewDidLoad`

Toggl.Daneel/ViewControllers/Settings/SendFeedbackViewController.cs
```diff
@@ -0,0 +1,75 @@
...
+using Toggl.Foundation.MvvmCross.ViewModels;
+using System.Reactive.Linq;
+using Toggl.Daneel.Views;
+using Toggl.Multivac.Extensions;
+using UIKit;
+
+namespace Toggl.Daneel.ViewControllers.Settings
+{
+    [ModalCardPresentation]
+    public partial class SendFeedbackViewController : ReactiveViewController<SendFeedbackViewModel>
```
arturs105: `seal` this class.

Toggl.Daneel/ViewControllers/Settings/SendFeedbackViewController.cs
```diff
@@ -0,0 +1,75 @@
...
+            View.AddSubview(activityIndicatorView);
+            activityIndicatorView.TranslatesAutoresizingMaskIntoConstraints = false;
+            NSLayoutConstraint.ActivateConstraints(new[]
+            {
+                activityIndicatorView.TopAnchor.ConstraintEqualTo(SendButton.TopAnchor),
+                activityIndicatorView.BottomAnchor.ConstraintEqualTo(SendButton.BottomAnchor),
+                activityIndicatorView.LeftAnchor.ConstraintEqualTo(SendButton.LeftAnchor),
+                activityIndicatorView.RightAnchor.ConstraintEqualTo(SendButton.RightAnchor)
+            });
+            activityIndicatorView.Image = activityIndicatorView.Image.ImageWithRenderingMode(UIImageRenderingMode.AlwaysTemplate);
```
arturs105: Is this necessary?

Toggl.Foundation.MvvmCross/ViewModels/IReactiveBindingHolder.cs
```diff
@@ -38,6 +40,14 @@ public static void Bind(this IReactiveBindingHolder holder, IObservable<Unit> ob
                 .Subscribe(onNext)
                 .DisposedBy(holder.DisposeBag);
         }
+
+        public static void Bind<T>(this IReactiveBindingHolder holder, IObservable<T> observable, ISubject<T> subject)
```
arturs105: I'm not sure we should have this method, since we're not keen on exposing `ISubjects`. I think we can make this work without public `ISubjects`.

Toggl.Daneel/ViewControllers/Settings/SendFeedbackViewController.cs
```diff
@@ -0,0 +1,75 @@
...
+        {
+            base.ViewDidLoad();
+
+            prepareViews();
+            prepareIndicatorView();
+
+            this.Bind(CloseButton.Tapped(), ViewModel.CloseButtonTapped);
+            this.Bind(FeedbackTextView.Text(), ViewModel.FeedbackText);
+            this.Bind(ErrorView.Tapped(), ViewModel.ErrorViewTapped);
+            this.Bind(SendButton.Tapped(), ViewModel.SendButtonTapped);
```
arturs105: These bindings should be changed to something like: `this.BindVoid(CloseButton.Tapped(), ViewModel.Close);`

And `ViewModel.Close` is a public method that calls `NavigationService.Close(this)` and does other stuff that's needed.

arturs105: Then repeat for the other bindings.

Toggl.Foundation.MvvmCross/ViewModels/MainViewModel.cs
```diff
@@ -170,7 +170,7 @@ public bool ShouldShowWelcomeBack
             this.onboardingStorage = onboardingStorage;
 
             SuggestionsViewModel = new SuggestionsViewModel(dataSource, interactorFactory, onboardingStorage, suggestionProviders);
-            RatingViewModel = new RatingViewModel(timeService, dataSource, ratingService, feedbackService, analyticsService, onboardingStorage, navigationService);
+            RatingViewModel = new RatingViewModel(timeService, dataSource, ratingService, analyticsService, onboardingStorage, navigationService);
```
arturs105: You should remove the `feedbackService` field from `MainViewModel` as well.

Furthermore, I think you could even remove the `IFeedbackService` and `FeedbackService` all together.

Toggl.Daneel/Extensions/UIKitRxExtensions/UIViewRxExtensions.cs
```diff
@@ -30,8 +30,11 @@ public static Action<bool> BindIsVisibleWithFade(this UIView view)
                 AnimationExtensions.Animate(
                     Animation.Timings.EnterTiming,
                     Animation.Curves.EaseIn,
-                    () => view.Alpha = alpha
-                );
+                    () =>
+                    {
+                        view.Alpha = alpha;
+                        view.Hidden = !isVisible;
```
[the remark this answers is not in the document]
khoi: By just looking at the method name, I'd think that I could just pass my view where `hidden = true` and it would work. 

I'm adding that suspender for some extra safety :)

Toggl.Daneel/ViewControllers/Settings/SendFeedbackViewController.cs
khoi: yes, since this image is used by many others, forcing it to `AlwaysTemplate` mode making sure the tintcolor is set correctly

Toggl.Foundation.MvvmCross/ViewModels/IReactiveBindingHolder.cs
khoi: binding an `Observable` to a `Subject` is a common practice no? how would you do it?

Toggl.Foundation.MvvmCross/ViewModels/MainViewModel.cs
khoi: feedback service is still being used in Giskard so I can't remove it

Toggl.Foundation.MvvmCross/ViewModels/IReactiveBindingHolder.cs
arturs105: I haven't used rx in any other project, so i wouldn't know if it's common practice. We usually use public methods (like `SetText(string)`, `SomeButtonTapped()`) in the view model for view -> view model bindings.

However, @Odrakir has cooked up something in #2751 that might be useful.

Toggl.Foundation.MvvmCross/ViewModels/MainViewModel.cs
arturs105: The only usage in Giskard that i found was in `Setup.cs` and that's just for registering the singleton. So we basically register this service and then never use it. We should be able to remove the service, though it would require slight alterations to `Setup.cs` and related code in Daneel and Giskard.

Toggl.Foundation.MvvmCross/ViewModels/IReactiveBindingHolder.cs
Odrakir: Yep, I agree, exposing Subjects is not a great idea. My RxAction works for buttons, and I guess it could have an extension for text fields as well. Although, as it's in the works... I don't know, we might want to leave this and refactor after #2751 is merged.

Toggl.Foundation.MvvmCross/Helper/Color.cs
```diff
@@ -233,6 +233,12 @@ public static class Settings
             public static readonly MvxColor SyncStatusText = new MvxColor(144, 146, 147);
         }
 
+        public static class SendFeedback
```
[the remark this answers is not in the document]
arturs105: Or `FeedbackView`

Toggl.Daneel/ViewControllers/Settings/SendFeedbackViewController.cs
```diff
@@ -0,0 +1,75 @@
...
+            NSLayoutConstraint.ActivateConstraints(new[]
+            {
+                activityIndicatorView.TopAnchor.ConstraintEqualTo(SendButton.TopAnchor),
+                activityIndicatorView.BottomAnchor.ConstraintEqualTo(SendButton.BottomAnchor),
+                activityIndicatorView.LeftAnchor.ConstraintEqualTo(SendButton.LeftAnchor),
+                activityIndicatorView.RightAnchor.ConstraintEqualTo(SendButton.RightAnchor)
+            });
+            activityIndicatorView.Image = activityIndicatorView.Image.ImageWithRenderingMode(UIImageRenderingMode.AlwaysTemplate);
+            activityIndicatorView.TintColor = Color.SendFeedback.ActivityIndicator.ToNativeColor();
+            activityIndicatorView.Hidden = true;
```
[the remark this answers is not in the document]
khoi: my reasoning behind this is when another dev look at the IB, he'll know immediately that there is an `indicatorView` there. 

If I set is to `hidden` then it's kinda difficult to see, unless you look at the view structure in the left menu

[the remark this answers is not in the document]
khoi: ah, I also setup the ActivityIndicatorView purely in code huh, I guess I was being lazy at opening storyboard.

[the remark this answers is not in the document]
sofokli1: For the IB not showing hidden views, I always have enabled `Editor > Canvas > Show Bounds Rectangle`. It also shows views below other views.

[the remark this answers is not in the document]
khoi: Is it mandatory for this to be set in the storyboard? I feel like to create this view, it's easier to do it in code :D

Toggl.Daneel/ViewControllers/Settings/SendFeedbackViewController.cs
khoi: At first I thought this is the matter of style, but it's actually more than that. Let's say for the `Send` action, I move it to a separated method of the ViewModel:

```csharp
public void SendFeedback() {
    interfactor.Send(feedbackText.value) 
}
```

the `.value` part really bother me, it's a bridge between Reactive and imperative code, and I'd argue that it shouldn't be there.

Toggl.Daneel/ViewControllers/Settings/SendFeedbackViewController.cs
arturs105: Maybe it's worth moving the `ImageWithRenderingMode` call inside the `ActivityIndicator` class, since there are a few other places where we need those things colored 🤔 

Though we have to make sure that this change doesn't alter the color of other activity indicators that are being used. It's up to you whether we do it in this PR or a separate one.

khoi: Will be done in https://github.com/toggl/mobileapp/issues/2765

Toggl.Daneel/ViewControllers/Settings/SendFeedbackViewController.cs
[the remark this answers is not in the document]
sofokli1: Don't think it is mandatory, maybe ask @amulware for that.

My main reasoning is that if most of the stuff is set in IB wouldn't it look strange not to find it there when you want to make a slight tweak to the ui? You open a xib and everything is there except for something that you can only see at runtime and now you have go search in code to find it. Also consistency.

amulware:
Did another manual test. Notes:

- Trying to send in offline mode no longer takes forever (in fact it seems instant), but no error is shown
- Is there an issue about adding blur to the error/confirmation boxes?

Toggl.Giskard/Services/DialogService.cs
```diff
@@ -70,9 +70,11 @@ public IObservable<bool> ConfirmDestructiveAction(ActionType type)
                     return Confirm(null, Resources.DiscardEditingChanges, Resources.Discard, Resources.ContinueEditing);
                 case ActionType.DeleteExistingTimeEntry:
                     return Confirm(null, Resources.DeleteThisTimeEntry, Resources.Delete, Resources.Cancel);
+                case ActionType.DiscardFeedback:
```
simonrozsival: This switch is really weird. We should move this logic at least to some extension method or something. Not a problem of this PR though.

Toggl.Foundation.Tests/MvvmCross/ViewModels/BaseViewModelTests.cs
```diff
@@ -1,3 +1,4 @@
+using System.Threading;
```
simonrozsival: 👀

Toggl.Foundation.MvvmCross/ViewModels/Settings/SettingsViewModel.cs
```diff
@@ -219,8 +224,10 @@ await navigationService
             await changeDefaultWorkspace(selectedWorkspaceId);
         }
 
-        public Task SubmitFeedback()
-            => feedbackService.SubmitFeedback();
+        public async Task SubmitFeedback()
+        {
+            isFeedbackSuccessViewShowing.OnNext(await navigationService.Navigate<SendFeedbackViewModel, bool>());
```
simonrozsival: Please split this into two lines, it is unnecessarily complicated - awaiting while passing an argument. We can afford one extra LOC! :)

Toggl.Foundation.MvvmCross/ViewModels/Settings/SendFeedbackViewModel.cs
```diff
@@ -0,0 +1,100 @@
...
+            });
+
+            ErrorViewTapped
+                .Do(e => errorViewVisible.OnNext(false))
+                .Subscribe()
+                .DisposedBy(disposeBag);
+
+            SendButtonTapped
+                .WithLatestFrom(FeedbackText, (_, text) => text)
+                .Do(_ => { isLoading.OnNext(true); })
```
simonrozsival: Shouldn't we also hide the error at this point (if one is shown)?

Toggl.Foundation.MvvmCross/ViewModels/RatingViewModel.cs
```diff
@@ -125,7 +131,7 @@ private string ctaButtonTitle(bool? impressionIsPositive)
             }
             else
             {
-                await feedbackService.SubmitFeedback();
+                isFeedbackSuccessViewShowing.OnNext(await navigationService.Navigate<SendFeedbackViewModel, bool>());
```
simonrozsival: Split this into two lines please.

khoi:
@amulware 
> no error is shown
Turned out the error is hidden below the keyboard , I'm dismissing the keyboard when the user hit send
> blur in confirmation message
This will be done in a separated ticket https://github.com/toggl/mobileapp/issues/2750

Toggl.Foundation.MvvmCross/ViewModels/Settings/SendFeedbackViewModel.cs
```diff
@@ -0,0 +1,111 @@
...
+    [Preserve(AllMembers = true)]
+    public class SendFeedbackViewModel : MvxViewModelResult<bool>
+    {
+        // Internal States
+        private CompositeDisposable disposeBag = new CompositeDisposable();
+        private ISubject<bool> errorViewVisible = new BehaviorSubject<bool>(false);
+        private ISubject<bool> isLoading = new BehaviorSubject<bool>(false);
+
+        // Inputs
+        public ISubject<Unit> CloseButtonTapped = new Subject<Unit>();
```
Odrakir: I see Subjects being overused everywhere. Subjects are great for linking reactive and non-reactive code, but they shouldn't be abused.

Example of close as an RxAction:

```c#
public UIAction CloseButtonTapped;
...
CloseButtonTapped = new UIAction(close);
...
private IObservable<Unit> close()
{
    return IsFeedbackEmpty
        .SelectMany(isEmpty =>
        {
            return isEmpty
                       ? Observable.Return(true)
                       : dialogService.ConfirmDestructiveAction(ActionType.DiscardFeedback);
            })
        .SelectMany(shouldClose =>
        {
            return shouldClose
                       ? navigationService.Close(this, false).ToObservable().SelectUnit()
                       : Observable.Empty<Unit>();
            });
}
```

One thing I'm not sure about is `FeedbackText` being a Subject (which makes the need of `IsFeedbackEmpty` as an observable here). I might use a private property to store the text or just send it from the UI (Instead of `UIAction` using an `InputAction`). But that's personal preference.

Toggl.Foundation.MvvmCross/ViewModels/Settings/SendFeedbackViewModel.cs
```diff
@@ -0,0 +1,111 @@
...
+        public ISubject<string> FeedbackText = new Subject<string>();
+        public ISubject<Unit> ErrorViewTapped = new Subject<Unit>();
+        public ISubject<Unit> SendButtonTapped = new Subject<Unit>();
```
Odrakir: Another example for `SendButtonTapped`:

```c#
public InputAction<string> SendButtonTapped;
...
public IObservable<bool> ErrorViewVisible
    => SendButtonTapped.Errors.Select(_ => true);

public IObservable<bool> IsLoading
    => SendButtonTapped.Executing;
...
SendButtonTapped = new InputAction<string>(sendFeedback);
...
private IObservable<Unit> sendFeedback(string text)
{
    errorViewVisible.OnNext(false);

    return interactorFactory
        .SendFeedback(text)
        .Execute()
        .SelectMany(_ =>
        {
            return navigationService.Close(this, true).ToObservable().SelectUnit();
        });
}
```

Odrakir: Here I'm sending the text from the UI, just to show how it would be done. But as I said it's a personal preference.

The call from the VC would then be:
```c#
this.Bind(SendButton.Tapped().Select(_ => FeedbackTextView.Text), ViewModel.SendButtonTapped);
```

amulware:
With so many other's reviewing, don't wait for me to squash.

khoi:
Improvements will be done in 
https://github.com/toggl/mobileapp/issues/2775

PR #2705, opened 2018-07-19
Fix vertical alignment of project dot in running time entry

kornelijepetak:
### Description

* Fixes vertical alignment of the project dot on the running time entry.
* Fixes left margin when there is no description

### Relationships

Closes #2650

### What has been done in this PR?

~Basically, I see no reason not to use a proper unicode character for the dot. It is vertically centered and easily colorable.~ The circle is centered now by using the text-tall container and centering the dot inside it. 

### 🦑 permissions
Everyone.

PR #2713, opened 2018-07-21
Fix main running te font scale/display size friendly

daividssilverio:
### Description

* Fixes the alignment of the running time entry timer when the font is big
* Keeps the vertical alignment of the project dot on the running time entry

### Relationships

Part of #2676 (Running TE card)
Undoes part of #2705 by using the ConstraintLayout to align the project dot 

### What has been done in this PR?

* Rewrites the MainRunningTimeEntry.axml using the ConstraintLayout, which should make changes easier while keeping the layout flat as possible.

### Extra Benefits

* It introduces the ConstraintLayout 1.1 to our project, that can be used to replace the usages of RelativeLayouts (not really easy to use while keeping things flat)
* The ConstraintLayout is mature enough to be used and is being encouraged by the community and the last google IOs. It's quite easy to use as soon as you learn its properties. More on it here 
https://developer.android.com/training/constraint-layout.


daividssilverio:
I can run some checks before merging. @juanlaube and @kornelijepetak. I'm not sure myself, as far as I know the lib is pretty light, adding about ~700+ to the method count.
I'll be using this tool for checking. https://developer.android.com/studio/build/apk-analyzer.

daividssilverio:
App size:
~500KB in the app size; ~200KB in download size;

Method count (I did this on the debug build, it might be less than that):
Without the ConstraintLayout: 34473
With the ConstraintLayout: 35392
Diff: 919 methods.

PR #2718, opened 2018-07-23
Fix crash caused by spans in different order

### LINKED ISSUE #2716: Crash when creating new time entry, with tag, project and client.
The app crashes when creating a new entry with a tag, project and client assigned.

## 🐛 Bug Report
The first time this occurred it was the first time entry on the account, created a new tag using #, new project using @ and created a new client in the project menu then the app crashed.

Upon trying to reproduce this I was able to make the app crash by starting a new time entry and assigning a tag and project by either menu or @# then once the project was assigned the app would crash and restart with no record of the time entry.

## Context
Internal test build (Playstore build from slack) 1.1 from Friday 20/08/18

Build/Branch:Internal test
Phone OS: Android 7.0
Phone model:HTC U Ultra

## Expected behavior
To be able to create a new time entry.

## To reproduce
Create a new time entry, add a description, assign a tag, assign a project and the app should crash.

juanlaube:
### Description
Fixes a crash when creating a time entry with tags and project.

### Relationships
Closes #2716

### What has been done in this PR?
The issue was caused by a bug in `SpannableStringBuilder`. Spans returned by `.GetSpans` are not necessarily in the correct order, this bug was reported [here](https://issuetracker.google.com/issues/37129364) and fixed in Oreo. Sorting the spans by the start position solved the problem.

### 🦑 permissions
@amulware


amulware:
I triggered a test rerun.

PR #2669, opened 2018-07-12
:memo: Add first issue templates

amulware:

PR #2719, opened 2018-07-23
Add pull request template

amulware:
## What's this?
This adds a template for pull requests. It will automatically populate the description and all we have to do is fill in the specific details.

### Relationships

Closes no issue
Follow up to our first issue templates here: https://github.com/toggl/mobileapp/pull/2669

## Why do we want this?
Three reasons:

1. Reviewing a pull request is a lot easier if the description is clear and precise and includes all the needed information. Generally, our PR descriptions are too short, this template encourages adding a bit more info.
2. Sometimes it becomes vital to look back at past changes. Just like with 1. more information in the PR description can really help understand why and how changes were made.
3. Standardisation. If we all use the same format for our pull request descriptions, it will be easier to get an overview and get started with a review, since we know what info to expect where, and we don't have to read the entire wall of text if we're looking for specific information.

## How is it done?
I followed Github's documentation here: https://help.github.com/articles/creating-a-pull-request-template-for-your-repository/

As specified there, I added the file in the correct directory, and the rest should just happen magically. Note that in principle this supports multiple templates. Maybe we want to have those in the future, but right now I don't see much reason to complicate things.

Also, note that just like with the issue templates I added HTML comments to the different sections to explain the intended usage.

### Why not another way?
I didn't go with more than one template to not be overly confusing for now. But I did already move the template into the `PULL_REQUEST_TEMPLATE` directory to support such in the future, instead of going with the `pull_request_template.md` solution.

### Side effects
I really don't think there are any. But I'm leaving this section here as an example.

## Review hints
The best example of how PRs are going to look after this change is this PR itself. I used the exact same template.

If you like this, then you hopefully will like the template. If you have ideas for improvement, please speak up!

## :squid: Permissions
Me. Me. Me.


.github/PULL_REQUEST_TEMPLATE/default.md
```diff
@@ -0,0 +1,21 @@
...
+
+Closes #
+
+## Why do we want this?
+<!-- Describe the reason for making this change -->
+
+## How is it done?
+<!-- Describe how the changes are implemented, list the different parts the changes consist of if it's more than one -->
+
+## Side effects
```
amulware: I'm thinking of turning this into an H3 (`###`). Thoughts?

juanlaube: Agree, h3 is better, side effects are part of how it is done 🙂

juanlaube:
Maybe add 🦑 permissions?

amulware:
I'd like reviews from at least two more people, and will generally leave this up for a day or so, so everyone has a chance to comment. :)

.github/PULL_REQUEST_TEMPLATE/default.md
```diff
@@ -0,0 +1,24 @@
...
+
+### Relationships
+<!-- Mention any issues or PRs that are connected to this -->
+
+Closes #
+
+## Why do we want this?
+<!-- Describe the reason for making this change -->
+
+## How is it done?
```
kornelijepetak: Within "How is it done" I would add an `<!--optional -->`  h3 of **Why this way?** or whatever better heading you manage to conjure, for situations when something was not done in an obvious way, which calls for explanation of why.

Any thoughts on that?

PR #2725, opened 2018-07-24
Signup: make signup scrollable to support tiny screens and larger fonts

daividssilverio:
Part of #2676 Sign up (sign up with google hidden)

It's basically the same changes made #2708, but for the signup view.

Also moved the noise bg to static part of the background, making the view scrollable only when necessary (for both signup and login views)

🦑 whenever.


daividssilverio:
Relate issue: #2731

PR #2727, opened 2018-07-25
Fix clients/tasks overlapping (+ConstraintLayout on main TE log)

### LINKED ISSUE #2726: Clients are overlapping with task names
## 🐛 Bug Report

Clients names are overlapping with task names
SS (I created a test account for this test):

## Context
It happens on Giskard 1.1

Build/Branch: develop
Phone OS: Android
Phone model: Nexus 5

## Expected behavior
Client names and tasks should not overlap.

## To reproduce
1. Create a project;
2. Create a client;
3. Set the new client to the project;
4. Create a task for this new project;
5. Create a new time entry for the new project with the client and one task.
6. There you go.

daividssilverio:
Closes #2726

I took the liberty to refactor the main log TE using ConstraintLayout because RelativeLayout is annoying to handle and work with.

This change also makes the `MainLogCellDescriptionOnly.axml` and related code unnecessary because CL can handle the use case pretty well. We can most likely do the same for `MainSuggestionsCardDescriptionOnly.axml` if we ever want/need to.

🦑 whenever, but not before some testing on multiple screen sizes and TEs with all the possible configurations (description, project, task, clients, need-sync, error).


kornelijepetak:
I've tested this with various TE combinations on various font/display sizes... haven't found any problems in UI layouting.

PR #2738, opened 2018-07-25
🚧 Add snooze logic to rating view

### LINKED ISSUE #2732: Rate us snooze is not working as expected
## Context

When the _rate us_ view is visible and the user doesn't interact with it, it should be presented again after 24 hs.

Build/Branch: develop
Phone OS: iOS

## Expected behavior

After 24 hs, the _rate us_ view should appear again

## Suggestions

Check the logic in MainViewModel, starting in line 280.

sofokli1:
Closes #2732

Just adds some logic to show the view again after 24 hours if the user did not interact the first time it showed up.

🦑 whenever


Toggl.Foundation.MvvmCross/ViewModels/MainViewModel.cs
```diff
@@ -281,10 +281,27 @@ private void presentRatingViewIfNeeded(bool shouldBevisible)
         {
             if (!shouldBevisible) return;
 
-            if (onboardingStorage.RatingViewOutcome().HasValue) return;
+            var wasShownMoreThanOneTimes = onboardingStorage.NumberOfTimesRatingViewWasShown() > 1;
+            if (wasShownMoreThanOneTimes) return;
```
khoi: do we need the extra variable, I feel like
`if (onboardingStorage.NumberOfTimesRatingViewWasShown() > 1)` will suffice

amulware: I see no harm in added clarity, but it's up to you guys.

kornelijepetak: That variable is used again down there so it has to stay.
However, it should be `OneTime` and not `OneTimes`

khoi: or maybe `Once`?

PR #2739, opened 2018-07-26
Fix main log TE ConstraintLayout client/task constraints

daividssilverio:
Closes nothing.

While testing for the main log crashes, I noticed that the Client/Task labels could lose their baseline when being recycled from some TE without a project.

It would result on the main log looking like this:

You can reproduce that on develop by resuming any TE with a the full project-client-task combo a few times.

🦑 whenever; testing on a device/emulator is advised


willsb:
Tests derped, so I'm restarting them

Issue #2737, opened 2018-07-25
TimeoutException in RetryDelayService crashes app


amulware:
## 💥 Crash Report
This exception is thrown because a server-status-checking sync state took too long (this is on purpose). And somehow this seems to crash the app (this is very much not on purpose).

I'm thinking if maybe we should rethink the way we timeout in the case of the status retry. If we know we're gonna retry for longer than the timeout will allow for, why are we not simply stopping sync with a fail right there?

But even then, exceptions in the syncing algo should not crash the app.

### 👥 Impact
- over 260 crashes in 1.7.2 at time of writing
- that's 13% of crashes in 1.7.2
- all sorts of phones

- also, about 10% of all crashes in Giskard 1.2, again all sorts of phones

### 🛰 Link
https://appcenter.ms/orgs/Toggl/apps/Toggl-iOS/crashes/groups/471c4d3fec614d025f8af27f082ccbc9a385f1f9?appVersion=1.7.2&period=last90Days

### :bug: Exception & Stack Trace
`System.TimeoutException: Retry delay 119.571s exceeded the maximum delay limit 60s.`
```cs
RetryDelayService.nextDelay (System.Double factor, System.Double defaultDelay)
RetryDelayService.NextSlowDelay ()
CheckServerStatusState.delay (System.IObservable`1[T] observable)
CheckServerStatusState.Start ()
TransitionHandlerProvider+<>c__DisplayClass1_0.<ConfigureTransition>b__0 (Toggl.Foundation.Sync.ITransition _)
StateMachine.performTransition (Toggl.Foundation.Sync.ITransition transition, Toggl.Foundation.Sync.TransitionHandler transitionHandler)
StateMachine.onTransition (Toggl.Foundation.Sync.ITransition transition)
AnonymousSafeObserver`1[T].OnNext (T value)
SingleAsync`1+Sequence+_[TSource].OnCompleted ()
Catch`2+_+HandlerObserver[TSource,TException].OnCompleted ()
FirstAsync`1+Sequence+_[TSource].OnNext (TSource value)
(wrapper unknown) System.Object.gsharedvt_out()
Select`2+Selector+_[TSource,TResult].OnNext (TSource value)
(wrapper unknown) System.Object.gsharedvt_in()
(wrapper unknown) System.Object.gsharedvt_out()
Merge`1+Observables+_+InnerObserver[TSource].OnNext (TSource value)
(wrapper unknown) System.Object.gsharedvt_in()
[...]

PR #2740, opened 2018-07-26
Add crash report issue template

amulware:
## What's this?
This adds another issue template, this time for easily pulling in crash reports, for example from App Center.

## Why do we want this?
To standardize crash reports.

## How is it done?
Same as the previous issue templates in #2669. See the changeset.

### Side effects
Support people might start using this to report crashes. If they do, I'll kindly tell them not to. If they keep doing it, I can change the 'about' text to be more explicit.

## Review hints
See if the format makes sense to you. For an example of how issues with this are gonna look, see here: https://github.com/toggl/mobileapp/issues/2737

## :squid: Permissions
~Me. I want to make sure to see any comments.~
🦑 🐙 :octocat: 🐱 🌵 🍰 🍥 🐟 🎣 🍣


.github/ISSUE_TEMPLATE/c-crash-report.md
```diff
@@ -0,0 +1,27 @@
...
+<!-- Use a short and clear title, like 'ExceptionType in ClassName' -->
+<!-- Apply the [crash] and appropriate platform tags -->
+
+## 💥 Crash Report
+<!-- A short description of what this crash is about, when it seems to happen, etc. -->
+
+### 👥 Impact
+<!-- How often has this crash occured, and to how many users, and during what time span? -->
+<!-- What fraction/percentage of crashes does this account for over that time span? -->
+<!-- Any pattern in the types of devices this happens on? -->
```
juanlaube: Maybe add `<!-- Is it reproducible? If so, what are the steps? -->`?

juanlaube:
I thought as I read the PR description _that_ was the template and it didn't make any sense 😅

.github/ISSUE_TEMPLATE/c-crash-report.md
amulware: Since these issues will likely come from App Center where we won't know steps I'm not sure it's really needed, but I added it to the section above, where it belongs more than under Impact.

amulware:
Also: 🦑 if nobody else has further change requests.

PR #2744, opened 2018-07-26
Make Drivers safer

willsb:
This ensures we expose only drivers on the ViewModel. This also introduces main thread schedulers. 

I’m using an static approach for simplicity sake, but an alternative would be adding the call to `ObserveOn` calls to platform specific extensions methods. I honestly don’t like this as much, but if someone can make a point I’m okay with changing.


Toggl.Giskard/Startup/HandlerScheduler.cs
```diff
@@ -0,0 +1,70 @@
...
+            handler.Post(() => 
+            {
+                if (isCancelled)
+                    return;
+
+                innerDisp.Disposable = action(this, state);
+            });
+
+            return new CompositeDisposable(
+                Disposable.Create(() => isCancelled = true),
```
daividssilverio: You could call handler.RemoveCallbacks(null) to make sure nothing is called after being disposed.

Toggl.Giskard/Startup/HandlerScheduler.cs
```diff
@@ -0,0 +1,70 @@
...
+            handler.PostDelayed(() => 
+            {
+                if (isCancelled)
+                    return;
+
+                innerDisp.Disposable = action(this, state);
+            }, dueTime.Ticks / 10 / 1000);
+
+            return new CompositeDisposable(
+                Disposable.Create(() => isCancelled = true),
```
daividssilverio: same here.

Toggl.Giskard/Startup/HandlerScheduler.cs
```diff
@@ -0,0 +1,70 @@
...
+            bool isCancelled = false;
+            var innerDisp = new SerialDisposable { Disposable = Disposable.Empty };
+
+            handler.PostDelayed(() => 
+            {
+                if (isCancelled)
+                    return;
+
+                innerDisp.Disposable = action(this, state);
+            }, dueTime.Ticks / 10 / 1000);
```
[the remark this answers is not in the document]
willsb: 10000 is to convert ticks into millis. This is from an older C# (before digit separators) so this is just meant to make the code more readable.

Toggl.Giskard/Startup/HandlerScheduler.cs
willsb: Settings cancelled should be enough, though

Toggl.Giskard/Startup/HandlerScheduler.cs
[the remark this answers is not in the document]
amulware: Why not use TimeSpan.TicksPerMillisecond?

daividssilverio:
7 tests are failing though.

Toggl.Multivac/Extensions/ReactiveExtensions.cs
```diff
@@ -81,7 +83,7 @@ public static IObservable<T> AsDriver<T>(this IObservable<T> observable, T onErr
             => observable
                 .Replay(1).RefCount()
                 .Catch(Observable.Return(onErrorJustReturn))
-                .ObserveOn(SynchronizationContext.Current);
+                .ObserveOn(MainThreadScheduler);
```
simonrozsival: What would happen if `MainThreadScheduler` is `null`? Will we be able to notice this easily during development? I am a bit afraid that we might forget how this works in a few months and if we break the `Setup` code or when we add a different platform and forget to add this code, everything will break and we will spend a lot of time trying to figure out how to fix this.

I would say that in this case it would be nice to have a comment here with explanation.

Odrakir: I kind of agree. Maybe instead of using EBM, we could use a proper old school method and take advantage of the opportunity to check if `MainThreadScheduler` is null and throw a appropriate error if it is.

willsb: Agreed

PR #2751, opened 2018-07-27
Rx action

Odrakir:
## What's this?
This is sort of the replacement for `MvxAsyncCommand`, only simpler, better, nicer and taller.

It's based on: https://github.com/RxSwiftCommunity/Action

## Why do we want this?
This enables an easy way to bind the UI to methods on the view mode, and more specifically methods that execute interactors (or anything that returns an Observable)

We would expose this from the view model
```c#
public RxAction<TimeEntryViewModel, bool> ContinueTimeEntryAction { get; }
...
ContinueTimeEntryAction = new RxAction<TimeEntryViewModel, bool>( timeEntryViewModel => 
      interactorFactory
           .ContinueTimeEntry(timeEntryViewModel)
           .Execute()
           .Do(/*whatever*/);
);
```
And then from the ViewController we can just do:
```c#
continueButton.BindToAction(ViewModel. ContinueTimeEntryAction, timeEntryViewModel);
```

And that would disable the button while executing and take care of possible errors. 

We could listen to the errors in the VM too.
```c#
ContinueTimeEntryAction.Errors.Subscribe(/*...*/)
```
Or the result(s), though that is returned from `Execute` as well.
```c#
ContinueTimeEntryAction.Elements.Subscribe(/*...*/)
```

## What's missing

The extensions on Daneel have no tests, but none others have... 

## :squid: Permissions
Squash whenever.


Toggl.Daneel/Extensions/UIKitRxExtensions/UIButtonRxExtensions.cs
```diff
@@ -31,5 +31,27 @@ public static Action<string> BindAnimatedTitle(this UIButton button)
                     null
                 );
             };
+
+        public static void BindToAction<Input, Output>(this UIButton button, RxAction<Input, Output> action, Func<UIButton, Input> transform)
```
arturs105: I think `TInput` and `TOutput` would be better names, since this had me looking for `Input` and `Output` types.

Toggl.Multivac/Extensions/RxAction.cs
```diff
@@ -0,0 +1,107 @@
...
+using System.Reactive.Linq;
+using System.Reactive.Subjects;
+
+namespace Toggl.Multivac.Extensions
+{
+    public class RxActionNotEnabledException : Exception
+    {
+    }
+
+    public class RxAction<Input, Element> : IDisposable
```
arturs105: Same as above: `TInput`, `TElement`.

Also, is there a reason this is called `Element` and the other file has `Output`?

Toggl.Multivac/Extensions/RxAction.cs
```diff
@@ -0,0 +1,107 @@
...
+    {
+        public IObservable<Exception> Errors { get; }
+        public IObservable<Element> Elements { get; }
+        public IObservable<bool> Executing { get; }
+        public IObservable<bool> Enabled { get; }
+        public Subject<Input> Inputs { get; }
```
arturs105: 1. Why not use `ISubject`?
2. Why plural name?

Toggl.Multivac/Extensions/RxAction.cs
```diff
@@ -0,0 +1,107 @@
...
+        public CompositeDisposable DisposeBag { get; }
+
+        private IObservable<bool> enabledIf;
```
arturs105: Isn't used

Toggl.Multivac/Extensions/RxAction.cs
```diff
@@ -0,0 +1,107 @@
...
+        public RxAction(Func<Input, IObservable<Element>> workFactory, IObservable<bool> enabledIf)
+        {
+            this.enabledIf = enabledIf;
+            this.workFactory = workFactory;
+
+            DisposeBag = new CompositeDisposable();
+            Inputs = new Subject<Input>();
+
+            var enabledSubject = new BehaviorSubject<bool>(false);
+            Enabled = enabledSubject.AsObservable();
```
arturs105: Could be simplified to:
```
Enabled = Executing
    .CombineLatest(enabledIf, (executing, enabled) => !executing && enabled);
```

Toggl.Multivac/Extensions/RxAction.cs
```diff
@@ -0,0 +1,107 @@
+using System;
+using System.Reactive.Disposables;
+using System.Reactive.Linq;
+using System.Reactive.Subjects;
+
+namespace Toggl.Multivac.Extensions
+{
+    public class RxActionNotEnabledException : Exception
```
arturs105: Shouldn't this be sealed?

arturs105: The same goes for `RxAction` class

Toggl.Multivac/Extensions/RxAction.cs
```diff
@@ -0,0 +1,107 @@
...
+            var errorsSubject = new Subject<Exception>();
+            Errors = errorsSubject.AsObservable();
+
+            executionObservables = Inputs
+                .WithLatestFrom(Enabled, (i, e) => (input: i, enabled: e))
+                .SelectMany(t =>
+                {
+                    if (t.enabled)
+                    {
+                        var ob = workFactory(t.input)
```
arturs105: This piece of code could use better naming.

Toggl.Multivac/Extensions/RxAction.cs
Odrakir: I'm going with the original implementation, but I guess it's because the output is an Observable of Elements

Toggl.Multivac/Extensions/RxAction.cs
Odrakir: As with any other Observable in this file, it could receive many items, thus the plural.

Toggl.Multivac/Extensions/RxAction.cs
Odrakir: You might be right... Although, maybe it's done on purpose so everything is disposed with the `disposeBag`? 🤔

Toggl.Multivac/Extensions/RxAction.cs
Odrakir: I'm a very open guy.

Toggl.Multivac/Extensions/RxAction.cs
```diff
@@ -0,0 +1,116 @@
...
+                .SelectMany(tuple =>
+                {
+                    var enabled = tuple.enabled;
+                    var input = tuple.input;
+
+                    if (enabled)
+                    {
+                        var ob = workFactory(input)
+                            .Do(_ => { }, error => errorsSubject.OnNext(error))
+                            .Multicast(new ReplaySubject<TElement>(1)).RefCount();
```
willsb: There are specializations of Multicast that do exactly that (namely, `Replay(1)`)

Toggl.Multivac/Extensions/RxAction.cs
```diff
@@ -0,0 +1,116 @@
...
+                .SelectMany(exec =>
+                {
+                    var execution = exec
+                        .SelectMany(_ => Observable.Empty<bool>())
+                        .OnErrorResumeNext(Observable.Empty<bool>());
+
+                    return Observable.Concat(Observable.Return(true), execution, Observable.Return(false));
+                })
+                .StartWith(false)
+                .Multicast(new ReplaySubject<bool>(1)).RefCount();
```
willsb: Same as above

Toggl.Multivac/Extensions/RxAction.cs
```diff
@@ -0,0 +1,116 @@
...
+        {
+            var subject = new ReplaySubject<TElement>();
+
+            var error = Errors
+                .Select(e => Observable.Throw<TElement>(e));
+
+            executionObservables
+                .Amb(error)
+                .Take(1)
+                .SelectMany( ob => ob)
```
willsb: We have an identity function in CommonFunctions which makes this much more readable

amulware:
I'm curious: if the errors are split out as a separate observable, we probably won't listen to all of them, so if something goes wrong, is it just going to swallow the exception silently?

I don't like crashes, but I also don't like swallowed exceptions, because that means no stack traces for us :/

Odrakir:
The idea is you bind this directly to a button, which doesn't like errors. So the way to be aware of them and handle them should be either:
a) For errors that don't have to be handled in the UI (retries or whatever): In the workFactory you pass to the RxAction constructor (that workFactory could execute an interactor or whatever)
b) For errors that need some kind of user response: In the viewmodel, which should expose a `IObservable<Exception>` (or `IObservable<Error>`) to the viewcontroller.

amulware:
c) What about unexpected errors, because bugs?

Odrakir:
If unhandled by the interactor, they'll get to the viewmodel and we then decide what to do with those: ignore them or show an alert.

amulware:
But that's my question: What if we forget to handle one of them? It'll just be a silent error, right?

amulware:
@willsb Please answer that question because I'm not sure.

willsb:
Depends on how you handled it. `await`ed observables are not silent exceptions. They are full blown application crashes. Subscribed observables will fail silently

Odrakir:
What about `MvxAsynCommand`, does it handle the exception?

willsb:
Nope, it should throw

PR #2752, opened 2018-07-27
Giskard 1.2 - Oxygen

amulware:
_giving our users a breath of fresh air from crashes and broken layouts_

Did you know that oxygen is a most dangerous molecule? In its liquid form, it can kill instantly, while at higher temperatures it easily causes endothermic reactions with other chemicals due to its high electron affinity.

```
- Fixed crash when scrolling the time entry log while syncing

- Fixed descriptions being cut off with RTL layout
- Fixed clients and tasks overlapping in time entry log
- Fixed timer display with varying font and display sizes
- Fixed login/signup screen layout on smaller phones
- Removed broken landscape mode in time entry edit view

```


amulware:
This has been released to users. Can I get a ✅?

PR #2760, opened 2018-07-27
Add IsGhost to database entities

juanlaube:
## What's this?

This is part of the effort to keep inaccessible data when the user lost access to a workspace. In this PR, only a property to know if an entity is ghost was introduced.

### Relationships

Ref #1337

## How is it done?

Workspace now has a new `IsGhost` property and other entities (i.e: TimeEntry, Project, Task, Tag, Client) delegate that knowledge to their workspace.

### Why not another way?

We could potentially add the same `IsGhost` property to all the realm objects, but right now this seems to be easier. If performance is a problem, we should consider this.

## Review hints

Is there anything that could be covered by unit tests? If so, let me know.

## :squid: Permissions

Whenever


Toggl.PrimeRadiant/Models/IDatabaseGhostSyncable.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace Toggl.PrimeRadiant.Models
+{
+    public interface IDatabaseGhostSyncable : IDatabaseSyncable
```
simonrozsival: I don't think we need this inheritance. I would vote for removing this dependency.

Toggl.Foundation/TogglSyncManagerFactory.cs
```diff
@@ -140,7 +142,9 @@ public static class TogglSyncManager
             var deleteNonReferencedGhostProjects = new DeleteNonReferencedProjectGhostsState(dataSource.Projects, dataSource.TimeEntries);
 
             transitions.ConfigureTransition(entryPoint, fetchAllSince);
-            transitions.ConfigureTransition(fetchAllSince.FetchStarted, persistWorkspaces);
+            transitions.ConfigureTransition(fetchAllSince.FetchStarted, ghostWorkspaces);
```
simonrozsival: Does this always mark all workspaces as ghosts? Is this just for testing (I am aware that this is still `wip`)?

amulware:
Please be aware that we have a bunch of split up issues in an appropriate project. Let's have similarly split up PRs if at all possible. If the sync would not work with a partial solution, we can simply not hook it up yet.

Toggl.PrimeRadiant/Models/IDatabaseGhostSyncable.cs
juanlaube: Removing this means other entities like User, Preferences or WorkspaceFeature can be ghosts too, does that make sense? 🤔

Toggl.Foundation/TogglSyncManagerFactory.cs
juanlaube: I need to look for a better way to do this (I'm all ~ears~ eyes if you have an idea), but for the moment this works because all workspaces are _ghosted_ and those that are fetched later are _unghosted_.

simonrozsival: If the app crashes between the transitions and user then goes offline, he/she won't be able to use the app, until he/she fetches the data at some point.

Toggl.PrimeRadiant/Models/IDatabaseGhostSyncable.cs
simonrozsival: I am talking just about having a separate `IDatabaseSyncable` and `IGhostable` interfaces. The entities which need it can implement both.

Toggl.PrimeRadiant.Realm/Database.cs
```diff
@@ -98,6 +98,15 @@ private RealmConfiguration createRealmConfiguration()
                     {
                         // nothing needs explicit updating when updating from schema 4 up to 5
                     }
+
+                    if (oldSchemaVersion < 6)
+                    {
+                        var workspaces = migration.NewRealm.All<RealmWorkspace>();
+                        foreach (var workspace in workspaces)
+                        {
+                            workspace.IsGhost = false;
```
amulware: Since false is the default value, I don't think this code is needed and the migration can be empty.

Toggl.PrimeRadiant/Models/IDatabaseGhostSyncable.cs
amulware: If what Simon says is possible, I'd prefer that too.

amulware:
In fact, if we only update the models first, we can start work on sync and UI separately, and I'd like that. :)

Toggl.PrimeRadiant.Realm/Database.cs
juanlaube: That's what I thought, but for some reason this didn't work. I can investigate more.

Toggl.PrimeRadiant/Models/IDatabaseGhostSyncable.cs
juanlaube: It is possible 🙂

juanlaube:
Since we're going with the small PRs approach, I removed all the things that are not part of adding the IsGhost property, should be easier to review.

Toggl.PrimeRadiant.Realm/Database.cs
```diff
@@ -98,6 +98,11 @@ private RealmConfiguration createRealmConfiguration()
                     {
                         // nothing needs explicit updating when updating from schema 4 up to 5
                     }
+
+                    if (oldSchemaVersion < 6)
```
simonrozsival: You can merge these two last ifs and update the comment accordingly (see the first `if`).

Toggl.PrimeRadiant.Realm/Models/RealmClient.cs
```diff
@@ -17,5 +17,7 @@ internal partial class RealmClient : RealmObject, IDatabaseClient
         public long WorkspaceId => RealmWorkspace?.Id ?? 0;
 
         public IDatabaseWorkspace Workspace => RealmWorkspace;
+
+        public bool IsGhost => Workspace.IsGhost;
```
simonrozsival: I am a bit hesitant about these properties in the `Realm` object - it might be tempting for somebody to use them in a query, but then we lose all the advantages of Realm being fast and everything will be loaded into memory and the search would run there.

Could we maybe have the `IsGhost` just in `RealmWorkspace` and then all the derived properties just in the foundation models?

amulware: I'm not sure I agree here. Won't having the property only in Foundation be what causes it to run in memory? Wouldn't these properties work in a realm query?

If they don't, then I'd ere on the side of not having them at all, and checking entity.Workspace.IsGhost everywhere, to not confuse anybody.

juanlaube: I'm not familiar with Realm and I can't find anything in their docs about performance. What's the best approach for this? Should we implement `IGhostable` only in the threadsafe objects, keep it like this or also add `IsGhost` to the other entities in the database?

Toggl.PrimeRadiant.Realm/Database.cs
juanlaube: 🙈

Toggl.PrimeRadiant.Realm/Models/RealmClient.cs
simonrozsival: You can use only properties which are stored in the database in queries. These properties are not stored and so it can't run in realm (over the raw data), but it will instantiate all the objects and run the linq over them.

If we wanted the queries to run in Realm, we would have to store the ghostness of the entities in the DB directly.

juanlaube: From your comment, I guess we need to store the property in the realm object? 🤔

amulware: @simonrozsival Are you sure? Why wouldn't `.Get(project => project.Workspace.IsGhost)` work?

simonrozsival: See https://realm.io/docs/dotnet/2.1.0/api/linqsupport.html#predicate-operations - last time I checked acessing child entities in a query it did not work and I found some discussions on GH which said it can’t be done at that time. I can’t find it now, maybe I will do some more research later and do some more testing.

amulware: But I'm really liking only having the properties in the workspace... It seems to me that'll make the syncing a lot simpler, except for maybe the cleanup code.

Then again, we won't have terrible amounts of ghost entities, in most cases none, so maybe it's just fine?

Do we need a full list of operations that we'll have to do that check for ghostness and see how it adds up?

Toggl.PrimeRadiant/Models/IGhostable.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace Toggl.PrimeRadiant.Models
+{
+    public interface IGhostable
```
Odrakir: Does it really need another interface? Couldn't it be part of `IDatabaseSyncable`?

amulware: I'm hoping we don't need this property in any entities except for workspaces, and we definitely don't need ti in users and preferences, so it can't be in the same interface.

amulware:
This is moving really slowly, here is what we need:

- a definite answer as to what we can do in realm queryables - specifically can we resolve references like project.Workspace.IsGhost?
- if we can, can we do so with a computed property like project.IsGhost => this.Workspace.IsGhost - probably not, but I'd like a definite answer
- if neither of these work, I want a trade-off analysis between having the property everywhere or only on the workspace - what operations will be affected in what ways? What are the potential things we can optimise for? recommendation?

amulware:
Ok, the answers are (thanks @kornelijepetak )
- no, not in the queryable
- yes, in the enumerable, still on realm objects before converting to foundation objects so should be reasonably fast
- not necessary

amulware:
After lengthy discussion internally, we decided that we do not want the property on all entities for the time being. If we really run into issues with it, we can add them later, but for now, they just complicate things needlessly.

Toggl.Foundation/Models/Workspace.cs
```diff
@@ -52,6 +52,7 @@ private Workspace(Builder builder)
             Name = builder.Name;
             SyncStatus = builder.SyncStatus;
             At = builder.At;
+            IsGhost = false;
```
amulware: Why set the default value?

PR #2762, opened 2018-07-29
No workspace UI

### LINKED ISSUE #1342: No workspace UI

arturs105:
## What's this?
This adds the UI and (commented out) presentation parts of no workspace view. The view isn't added properly, because it's still bugy. There's an issue to fix the bugs in the view model and enable the view presentation (#2769), so this can be merged as is - just the UI parts.

### Relationships
Closes #1342 

## :squid: Permissions
✅  🦑


amulware:
Currently Toggl backend closes accounts with no workspaces, 'for reasons'. The workaround is to have a pending invite to another workspace on the same user, that will keep it alive.

So:
- invite test user to two workspaces
- accept one of the invites
- delete previous workspace of user
- remove user from the accepted workspace
- now it should happen

Toggl.Foundation.MvvmCross/Services/ErrorHandlingService.cs
```diff
@@ -63,7 +63,7 @@ public bool TryHandleNoWorkspaceError(Exception error)
         {
             if (error is NoWorkspaceException)
             {
-                // Show no workspace UI here. Also don't forget to call `DataSource.StartSyncing` when the user creates the workspace
+                navigationService.Navigate<NoWorkspaceViewModel>();
```
juanlaube: Can you undo this until we fix the bugs in the view model?

PR #2763, opened 2018-07-30
Rxaction extensions

Odrakir:
## What's this?
Adds some subclasses and extensions to RxAction to make it more useful.

### Relationships
It'll hopefully be used in a lot of the refactoring going on right now, to get rid of MvxAsynAction.

## How is it done?

In the view model
```c#
DeleteTimeEntry = new InputAction<TimeEntryViewModel>(deleteTimeEntry);
...
private IObservable<Unit> deleteTimeEntry(TimeEntryViewModel timeEntry)
{
    return interactorFactory
        .DeleteTimeEntry(timeEntry.Id)
        .Execute()
        .Do( _ => {
            analyticsService.DeleteTimeEntry.Track();
            dataSource.SyncManager.PushSync();
        });
}
```
In the view controller (`SwipeToDelete` is an `IObservable<TimeEntryViewModel>`) 
```
this.Bind(tableViewSource.SwipeToDelete, ViewModel.DeleteTimeEntry);
```

You could also bind the action directly to a button, which would handle disabling the button when the action is disabled or running.

```c#
EditTimeEntryButton.BindToAction(ViewModel.EditCurrentTimeEntry);
```

### Side effects
We are ignoring all unhadled errors, but there's an issue to do something about it.

## :squid: Permissions
Do squash whenever.


Toggl.Multivac/Extensions/RxAction/InputAction.cs
```diff
@@ -0,0 +1,24 @@
+using System;
+using System.Reactive;
+
+namespace Toggl.Multivac.Extensions
+{
+    public class InputAction<TInput> : RxAction<TInput, Unit>
```
khoi: What is the purpose of this class? Isn't `RxAction<Unit, Unit>` what we usually use?

Odrakir: I’m using this one in my example. It’s for those times when there’s an input but we don’t care about the output and just need to know when it finished and if it was successful.

khoi: I feel like `RxAction<SomeInput, Unit>` would suffice. But there's nothing wrong with this class

PR #2770, opened 2018-07-30
Replace those funny `TwentySevenParameterConstructorTestData` classes with a custom xunit data attribute

simonrozsival:
## What's this?
This PR makes writing and changing unit tests for constructors' null-checks easier for us.

### Relationships
This PR is not related to any issue or PR.

## Why do we want this?
Every time somebody changes the constructor params of a service whose constructor is tested for `null`s, he/she has to update the `ClassData` attribute to match the number of parameters. I don't like that. We can easily count the number of params with reflection.

## How is it done?
`XUnit` allows us to create a custom data attribute and use reflection to examine the method's parameters.

### Why not another way?
This way is very straightforward, I don't see a simpler solution.

### Side effects
No side effects. There will be merge conflicts in other PRs thought, because many classes were 🔥.

## Review hints
Just look if the unit tests passed (they did).

## :squid: Permissions
Squash whenever


simonrozsival:
Sure, I can add such test, it is trivial 👍

simonrozsival:
I am aware of that. I think this could be OK and that we don't have to randomize the order of the test data and therefore the test is fine now.

amulware:
I agree.

PR #2771, opened 2018-07-30
✨ Add TabBarController

### LINKED ISSUE #2585: Add a tab bar
- [x] New tab bar should be displayed only in these three views: Timer page, Reports and Settings
- [x] Remove Reports and Settings icons from the current nav bar
- [x] In Reports and Settings page, remove "Back" option from the nav bar

### Timer
toggl - ios - timer tab bar

Zeplin: https://zpl.io/VxYvxq3

### Running entry
toggl - ios - timer - running entry tab bar 1x

Zeplin: https://zpl.io/2j4RAo6

### Reports
toggl - ios - reports tab bar 1x

Zeplin: https://zpl.io/bAq5JPp

### Settings
toggl - ios - settings tab bar 1x

Zeplin: https://zpl.io/2vYo9Gr

sofokli1:
## What's this?
This adds a tabBar to the app for easy access to the main log, reports and settings.

### Relationships

Closes #2585

## Why do we want this?

Not sure why. Maybe it looks good and it makes it easier to switch between the 3 views.

## How is it done?

Added a tabBarVC and a VM where the 3 tabs VC are created from their VMs and added into navigationBars before being added to the tabBar. Our presenter was a bit hardcoded for navigationControllers and needed some changes to support tabBarController.

### Why not another way?

I tried make the presentation of the tabs a bit more MVX style but failed. Feedback is welcome.

### Side effects

Not that I know of.
## Review hints

This needs as much functional test as possible since it changes the way some of our VMs are presented.

## :squid: Permissions

Whenever


amulware:
Note: We want this because we'll add the calendar screen soon, and the top bar is getting full. :D

Toggl.Daneel/ViewControllers/MainTabBarController.cs
```diff
@@ -0,0 +1,38 @@
...
+    public partial class MainTabBarController : MvxTabBarViewController<MainTabBarViewModel>
+    {
+        public MainTabBarController()
+        {
+            setupViewControllers();
+        }
+
+        private void setupViewControllers()
+        {
+            var viewControllers = ViewModel.ViewModelTuples.Select(tuple => createTabFor(tuple.Item1, tuple.Item2)).ToArray();
```
simonrozsival: I don't like the fact that the VM knows what the name of the image is. I think that the mapping of the vm type => image should be in the UI layer.

Toggl.Foundation.MvvmCross/ViewModels/MainTabBarViewModel.cs
```diff
@@ -0,0 +1,103 @@
...
+                platformConstants,
+                onboardingStorage,
+                navigationService);
+        }
+
+        public override async Task Initialize()
+        {
+            await base.Initialize();
+
+            ViewModelTuples = new(MvxViewModel, String)[] { (mainViewModel, "icTime"), (reportsViewModel, "icReports"), (settingsViewModel, "icSettings") }.Do(async tuple => await tuple.Item1.Initialize());
```
simonrozsival: Should we initialize all the view models right away, even though only one of them will be shown and the others will be hidden? Is there another way of doing this?

Toggl.Daneel/ViewControllers/MainTabBarController.cs
sofokli1: The way I see it, is that the tabBar VM knows about some files that are relate to the other VMs, same way as any other non UI resource. If it was a sound file name or api endpoint or a database path or any other file for that matter, would you want it in the VC or in the VM? Also the UI should be as simple as possible IMO, just show what data the VM gives it.

Let me know what you think. If you still fill strongly about it I will change it.

Toggl.Foundation.MvvmCross/ViewModels/MainTabBarViewModel.cs
sofokli1: Can't see a way around this one

Toggl.Daneel/ViewControllers/MainTabBarController.cs
simonrozsival: The thing is that Android uses the same VM and this is platform specific so it should not be in the VM. I think the easiest thing would be a dictionary with items like `[typeof(MainViewModel)] = "icTime"` and you would simply get the image name by `icons[vm.GetType()]`.

Toggl.Foundation.MvvmCross/ViewModels/MainTabBarViewModel.cs
simonrozsival: OK. Let's check that there are no leaks in the observables which are created in those `Initialize` methods and that the subscriptions aren't disposed before the view is displayed for the first time. The more I think about this, the more I think I am overthinking it. I will think about this a bit more just to be sure :D

Toggl.Daneel/Presentation/TogglPresenter.cs
```diff
@@ -140,15 +140,29 @@ protected override void SetWindowRootViewController(UIViewController controller,
...
                 ? null
-                : getPresentedViewController(navigationController.TopViewController) as MvxViewController;
+                : getPresentedViewController(tabBarController.SelectedViewController) as MvxViewController;
 
             //Don't show the same view twice
             if (topViewController?.ViewModel?.GetType() == request.ViewModelType)
                 return;
-                
+
+            if (request.ViewModelType == typeof(MainViewModel))
```
simonrozsival: Why are there two ifs for `MainViewModel` and for `ReportsViewModel`, but none for `SettingsViewModel`?

Toggl.Daneel/Presentation/TogglPresenter.cs
```diff
@@ -178,18 +192,20 @@ public override void ChangePresentation(MvxPresentationHint hint)
             {
                 case ReloadLogHint _:
                 {
-                    var mainViewController =
-                        MasterNavigationController
-                            ?.ChildViewControllers
-                            .FirstOrDefault(vc => vc is MainViewController) as MainViewController;
+                    var mainViewController = mainTabBarController
+                        .ViewControllers
+                        .Select(vc => vc as UINavigationController)
```
simonrozsival: It would be nicer to use `.Cast<UINavigationController>()` here.

Toggl.Daneel/Startup/AppDelegate.cs
```diff
@@ -129,11 +131,18 @@ public override void PerformActionForShortcutItem(UIApplication application, UIA
                     break;
```
simonrozsival: The shortcut to reports crashes the app.

Toggl.Daneel/Presentation/TogglPresenter.cs
sofokli1: Those are only used for the shortcuts I will add the settings one just in case we want to use `Navigate<...>` to change tabs

Toggl.Daneel/Startup/AppDelegate.cs
sofokli1: I test this on simulator and device and can't reproduce the crash. any error messages?

Toggl.Daneel/Presentation/TogglPresenter.cs
```diff
@@ -140,15 +140,36 @@ protected override void SetWindowRootViewController(UIViewController controller,
 
         public override void Show(MvxViewModelRequest request)
         {
-            var navigationController = UIApplication.SharedApplication.KeyWindow?.RootViewController as MvxNavigationController;
-            var topViewController = navigationController == null
+            var tabBarController = UIApplication.SharedApplication.KeyWindow?.RootViewController as UITabBarController;
```
Odrakir: Why don't you use the `mainTabBarController` created down there?

By the way, while on it, can you correct the order of this class? First properties then methods.

sofokli1:
@amulware @pe0ny regarding @Odrakir above comment about the reports view. What should be the correct behaviour?

amulware:
Good question. I can see what @Odrakir means. It makes sense to me to keep the reports state at least while the user is using the app. Maybe only reset it if they were gone for a while or the app was restarted properly. Final decision is up to @pe0ny, but if we're gonna do this, I'd say let's do it in a separate issue & pr.

Odrakir:
I brought it up here because it seems we are recreating the VC every time we change tabs, which is not the default behavior of the tab bar as far as I know.

Toggl.Daneel/Startup/AppDelegate.cs
sofokli1: @simonrozsival any news on this?

khoi:
There is a crash happening when signing up for a new user

khoi:
I'm logged in inside my simulator, when switching over this branch, I get the login screen when open the app. 

And if I go back to `develop` then run the app, I'm still logged in. Could this be a bug?

willsb:
How I'd go about and deal with this:

- Since state needs to be kept, I wouldn't recreate views whenever needed
- I'd create a TabViewModel, which is the main view and holds everything
- The TabViewModel would have instances of all its child ViewModels to prevent them from losing state
- I'd create a TabViewController that handle showing its children
- Children would in turn use the  `[TabPresentationAttribute]` and the presenter would do whatever is needed to show them (either create a new VC or just change the tabs).

That's the cleanest solution, in my opinion. If this can wait long enough, I can implement it myself.

amulware:
Create all the view models for all tabs on app-start? That doesn't strike me as a performant idea. Better to create them lazily when needed I think.

sofokli1:
@willsb I have a couple of questions:

- Did you read the code at all or just the comments? I am asking because all the points you mention except for the last one are there already.
- Did you run the app at all? I am asking because if you run the app you can clearly see that `Main` and `Settings` do not lose state. The issue is not with the tab bar but the `Reports VC/VM` and how we show the `ReportsCalendar VC/VM` that refreshes the view thus reseting the state.

I do not see any reason for this issue to be taken over by someone else since I am not on vacation.

Constructive feedback is always appreciated.

sofokli1:
@amulware it is not easy to create them lazily since the tap bar requires all the VCs to show the appropriate tab in the bar

amulware:
Let's at least keep a close eye on startup time then.

sofokli1:
I would not expect any special delays since the views are loaded when selecting the tab the first time, so that is lazy by default. The main view is the heaviest of them all and that is visible first so it will load as before. The only loading that is done non lazily is the VM creations for reports and settings which are relatively light VMs.

sofokli1:
To sum up the reports refreshing will be fixed in a different PR. This PR is looking for feedback/approval regarding the tabBar only not the individual tabs.

sofokli1:
@khoi I tried fresh install from develop, login, change branch to here and run. It logged in automatically. Can you take a look again when possible?

khoi:
@sofokli1 I tried again and it worked this time 🤷‍♂️

Toggl.Daneel/Startup/AppDelegate.cs
simonrozsival: I will test this asap.

simonrozsival: I can't reproduce the crash anymore.

sofokli1: that looks like a good reason to approve 😄

Toggl.Foundation.MvvmCross/ViewModels/MainTabBarViewModel.cs
```diff
@@ -0,0 +1,82 @@
...
+using Toggl.Ultrawave.Network;
+
+namespace Toggl.Foundation.MvvmCross.ViewModels
+{
+    [Preserve(AllMembers = true)]
+    public sealed class MainTabBarViewModel : MvxViewModel
+    {
+        private MainViewModel mainViewModel;
+        private ReportsViewModel reportsViewModel;
+        private SettingsViewModel settingsViewModel;
```
simonrozsival: 🔥

sofokli1:
Just trying to make sure all here are ok with this PR:

@khoi do you still find something wrong with this?
@willsb are you going to make an other PR agains this or should i merge and you can go on from develop?
@pe0ny now there are only 2 tabs.
@amulware you ok with this?

Just `comment`, `approve` or `give a thumbs up` so that i can merge this.

willsb:
@sofokli1 I'll create an issue to use attributes, but let's merge this here first.

pe0ny:
@sofokli1 Just tested it and it looks nice 😻

PR #2784, opened 2018-07-31
Fix onboarding phase 3 tooltip positioning when ime changes

### LINKED ISSUE #2687: Droid Onboarding Phase 3: "Add a project and tags on the fly" tooltip position
Tested on my Nexus 5, both on develop and production.

The tooltip appears as if the keyboard were closed and the bottom bar was still in the bottom of the screen. I thought this was already fixed. Maybe a bad merge?

daividssilverio:
### Relationships
Root issue: #1649
PR that introduced the tooltip: #2594 

Closes #2687

## Why do we want this?
Because it looks bad. This was a visual glitch.

## How is it done?
The PopupWindow doesn't work as nice as I thought and it won't update its position based on the position of the anchor when the later is not inside a scrolling view.
So, I'm checking for layout changes on the anchor, which is triggered every time the keyboard is opened or closed, dismissing an existing tooltip if there, creating and showing a new tooltip with updated positions.

### Why not another way?
This was simple and easy to do based on the code we already created for handling tooltips.
We could use an static view instead of the PopupWindow too, but that would require more work and changes to the current layout.

### Side effects
It adds an event listener to the anchor of the tooltip that will be running even if not needed (the tooltip doesn't have to be shown).
This is a bigger problem with the current implementation of the on-boarding steps across the app, and we should write another issue for that.

## Review hints
Change the code to always show the tooltip so you won't have to clear your data for the tests.

## :squid: Permissions
Anyone. 🦑 whenever


Toggl.Giskard/Activities/StartTimeEntryActivity.cs
```diff
@@ -34,6 +34,7 @@ public sealed partial class StartTimeEntryActivity : MvxAppCompatActivity<StartT
         private static readonly TimeSpan typingThrottleDuration = TimeSpan.FromMilliseconds(300);
 
         private PopupWindow onboardingPopupWindow;
+        private IDisposable onboardingDisposable;
```
kornelijepetak: Couldn't you have used the `DisposeBag?`

kornelijepetak: Oh, no I see. It needs to be disposed at other times as well. My bad.

PR #2786, opened 2018-08-01
Switch workspace report

### LINKED ISSUE #2702: Reports: Allow selecting workspace
If users have more than one workspace, we want to allow them to select one on the reports page.

It'll default to the default workspace, and not hide the selection at all when there is only a single one.

Note that selecting a different workspace should re-fetch the report, just as if a different date range was picked.

khoi:
## What's this?
Add the ability to change the Workspace for the Reports.

### Relationships
Closes #2702

## How is it done?
A new command is added to the ViewModel
`public IMvxCommand<long> ChangeWorkspaceCommand`

### Side effects
🤷‍♂️ 

## :squid: Permissions
Anyone


Toggl.Foundation.Tests/MvvmCross/ViewModels/ReportsViewModelTests.cs
```diff
@@ -19,6 +19,7 @@
 using Toggl.PrimeRadiant.Models;
 using Xunit;
 
+
```
simonrozsival: 👀

Toggl.Foundation.Tests/MvvmCross/ViewModels/ReportsViewModelTests.cs
```diff
@@ -487,5 +488,25 @@ public sealed class TheSegmentsProperty : ReportsViewModelTest
                     .ForEach(percentage => percentage.Should().BeGreaterOrEqualTo(5));
             }
         }
+
+        public sealed class TheChangeWorkspaceCommand : ReportsViewModelTest
+        {
+            [Fact, LogIfTooSlow]
+            public async Task Something()
```
simonrozsival: Please structure this test as AAA (arrange, act, assert) - take a look at the other tests in the codebase to see how it looks like.

Toggl.Foundation.Tests/MvvmCross/ViewModels/ReportsViewModelTests.cs
khoi: My Rider doesn't like whitespaces, can I get away with this, or should I revert the change

Issue #2750, opened 2018-07-27
Reusable Toast view


khoi:
📑Description

We use toast view quite frequently in our app. Currently, we're constructing each toast individually and add it to the view we want to display it.

Let's create one that's reusable and allow us to customize it. 

✍️Suggestions: 

- There should be 2 type of toasts (success and error).
https://app.zeplin.io/project/5728ce853dd1b182475f25e9/screen/5b43160025e2c904a0a14b36
https://app.zeplin.io/project/5728ce853dd1b182475f25e9/screen/5b4315ffb3366bcd6a0546f2

- The title and the description of the toast can be changed

- Some toast might have an button attached to it. https://app.zeplin.io/project/5728ce853dd1b182475f25e9/screen/5b31bb693ea3532eaabd8084

The toast has slight blur effect, @arturs105 has played around with the code for iOS below that could be used for references:

```csharp
var frame = new CGRect(10, 10, 300, 300);
var containerView = new UIView(frame);
var blurEffect = UIBlurEffect.FromStyle(UIBlurEffectStyle.Light);
var visualEffectView = new UIVisualEffectView(blurEffect);
visualEffectView.Frame = frame;
containerView.AddSubview(visualEffectView);

var blue = new UIView(frame);
blue.BackgroundColor = UIColor.Blue;
blue.Alpha = 0.5f;

//Add blue below containerView
View.AddSubview(blue);
View.AddSubview(containerView);
```

Odrakir:
#2801 Get's us close to this, but still some customization has to be done. So this issue should not be closed until that's done.

PR #2801, opened 2018-08-02
Undo deletion iOS UI

### LINKED ISSUE #2540: Undo time entry deletion UI
- Use the same animation to show and hide the snackbar: #1538

### Undo when the timer is not running
toggl - ios - timer - undo 1x

Zeplin: https://zpl.io/b6MDQrR

### Undo when the timer is running
toggl - ios - timer - undo running entry 1x

Zeplin: https://zpl.io/agvk9AZ

### Tappable area: 8pt padding from left and right; 10pt from to and bottom
screen shot 2018-07-02 at 13 07 48

Odrakir:
## What's this?
This adds the Snack bar for UNDO deletion. It's reusable enough so it can also be used in the feedback ui and password reset (or wherever else needed) with minimal or non modification. Ping @khoi & @arturs105 if interested.

It decides where to put the buttons dynamically depending on the text length and view size. So you can have both the following views with the same class.

captura de pantalla 2018-08-02 a las 11 40 51
captura de pantalla 2018-08-02 a las 11 42 02

Also, it has a public property to set the bottom anchor, which can be changed in run time and it adapts to the new anchor.

### Relationships
Closes #2540 #2750 

## Why do we want this?
The next step is the whole undo logic, but this PR brings the UI functionality needed for that.

## How is it done?
It's just a UIView subclass with some extra methods. The default colors and fonts are taken from the xib file, but they can be changed programmatically.

## Review hints
In order to test it you can replace the Start button behavior so it shows this bar.

In line 148 of `MainViewController.cs` do:
```c#
//bindingSet.Bind(StartTimeEntryButton).To(vm => vm.StartTimeEntryCommand);
var snackBar = SnackBar.Create();
snackBar.SnackBottomAnchor = StartTimeEntryButton.TopAnchor;
snackBar.AddButton("UNDO", () =>
{
    Console.WriteLine("Tap");
});
snackBar.SetTimer(3, () =>
{
    Console.WriteLine("Fin");
});
StartTimeEntryButton.Tapped()
    .VoidSubscribe(() => { snackBar.Show("Time entry was deleted", View); });

ViewModel.IsTimeEntryRunning
    .Subscribe(running =>
    {
        snackBar.SnackBottomAnchor = running ? CurrentTimeEntryCard.TopAnchor : StartTimeEntryButton.TopAnchor;
    })
    .DisposedBy(disposeBag);
```

## :squid: Permissions
Go ahead.


Toggl.Daneel/Views/SnackBar.cs
```diff
@@ -0,0 +1,189 @@
...
+                if (!showing) return;
+                Hide();
+                onTimer();
+            };
+        }
+
+        public void Show(string text, UIView superView)
+        {
+            if (Superview != null)
+            { //Only show it once
```
simonrozsival: 👀

Toggl.Daneel/Views/SnackBar.cs
```diff
@@ -0,0 +1,189 @@
...
+        }
+
+        public void Hide()
+        {
+            showing = false;
+            UIView.Animate(
+                0.3,
+                () =>
+                {
+                    Transform = CGAffineTransform.MakeTranslation(0, 100);
```
simonrozsival: Magic 💯 - you use this constant twice for translation, please give it a name.

Toggl.Daneel/Views/SnackBar.cs
```diff
@@ -0,0 +1,189 @@
...
+                    buttonPosition = ButtonPositionType.Bottom;
+                    configureLayout();
+                    LayoutIfNeeded(); // Seems silly, but this one is needed too
+                }
+            }
+        }
+
+        private void configure()
+        {
+            Layer.CornerRadius = 10;
```
simonrozsival: A named constant please!

Toggl.Daneel/Views/SnackBar.cs
```diff
@@ -0,0 +1,189 @@
...
+
+        public override void LayoutSubviews()
+        {
+            base.LayoutSubviews();
+
+            if (buttonPosition == ButtonPositionType.Right)
+            {
+                LayoutIfNeeded(); // Seems silly, but I need this line so the label returns the correct width
+                var labelWidth = label.Frame.Width;
+                var requiredWidth = ((NSString)label.Text).GetBoundingRect(new CGSize(2000, 2000),
```
simonrozsival: Why 2000x2000?

Toggl.Daneel/Views/SnackBar.cs
Odrakir: I'm not a big fan of commenting everything, but I think that comment makes sense.

simonrozsival: I don't mind the comment, the formatting seems off though. I should have been more specific the first time.

khoi:
Probably going to close #2750 as well

simonrozsival:
How is this gonna work when you delete several TEs while the snack bar is visible - will tapping the UNDO button reverse the deletion of the first or the last TE?

Odrakir:
According to the issue deleting a new entry will delete the previous one for real. Still not sure how I'm going to handle that, but probably not in the UI, so the old snack bar will just dismiss some how.

Issue #2703, opened 2018-07-19
Reports: WS select UI


amulware:
Requires #2702
iOS version here: #2624

lizaPlotnikova:
Hey guys! I'd like to hear some of your thoughts on this layout:

<img width="1189" alt="screen shot 2018-07-20 at 3 24 11 pm" src="https://user-images.githubusercontent.com/24964497/43021313-335b687a-8c31-11e8-8d81-ec33482dc255.png">

My thoughts on the designs: 
### v1 (leftmost screenshot)
This is parodying what is designed for iOS already – The WS select UI would be a custom fab that floats on top of the content. As in iOS because we don't want it to obscure any content permanently, extra padding would be added below the projects table to accommodate the height of the fab

I'm not a big fan of this design for a couple of reasons. 1) I don't think this is a very discoverable position for the element... you basically need to read the whole screen first before even seeing the button 2) I don't think it supports the hierarchy well because it's describing what the content of the whole screen is, but is positioned at the very bottom 3) its using a really modified version of a fab, yet a new custom component

So I propose the two alternatives: 
### v2
put a "single-cell" element (same one as in the "Edit" view) into a card at the top of the content
this might have better affordance (for it being clickable) than v3

### v3 (rightmost... sorry about the typo in the screenshot again)
put a "single-cell" element at the top of the content
this seems less cluttered to me, and more like a header (which I think is appropriate for the content)

amulware:
@pe0ny @lizaPlotnikova Please confirm if the design above is final or not. Discuss internally if needed. Thanks!

lizaPlotnikova:
Updated / Finalized designs:

- If a user belongs to multiple WSs, display the name of the selected WS at the top of the Reports page
- A user can change the selected WS by clicking the fab in the bottom-right which launches the Workspace picker dialog

![reports multiple workspaces](https://user-images.githubusercontent.com/24964497/44749975-297a1e80-aae2-11e8-9ef5-09d99d44dce9.png)
Zeplin: https://zpl.io/a3ljJG8

![reports ws selector](https://user-images.githubusercontent.com/24964497/44750238-d2c11480-aae2-11e8-9348-2f306017e90c.png)
Zeplin: https://zpl.io/aBNKn8A

kornelijepetak:
@lizaPlotnikova @pe0ny 

1. Zeplin needs a briefcase white icon as an image resource, not a shape.
[...]

PR #2803, opened 2018-08-02
Switching workspaces in reports

### LINKED ISSUE #2624: Reports: WS select UI
- The user should be able to switch Workspace in Reports view to see data from another Workspace. This, however, won't change the default workspace.
- Show the WS button only if the user has more than one workspaces. If the user has only one, don't show it at all.
- WS select button should float above the content.
- Max width of the button: 192pt. If the name doesn't fit, fade the name out.
- When the user taps the WS select button, use iOS default actions to list all the WS the user has and let him/her choose between those. Display also the selected WS name.
- Make sure when the WS button is visible, the padding between the content (list) and bottom should be higher.

### Reports view with WS select
toggl - ios - reports ws select 1x

Zeplin: https://zpl.io/29Z0AzY

### WS select with longer name
toggl - ios - reports ws select with longer name 1x

Zeplin: https://zpl.io/2vYZqA5

### WS select actions
toggl - ios - reports workspace select with actions 1x

Zeplin: https://zpl.io/V41BY94

simonrozsival:
## What's this?
This PR adds a floating button to the reports view which shows the current workspace and allows the user to switch between different workspaces. It is visible only when user has two or more workspaces.

### Relationships
Closes #2624 

## How is it done?
- I extended the dialog service to allow selecting from a list of options.
- The reports VM was extended with the necessary properties and the command for changing the workspace was modified a bit.
- Most of the UI is defined in the XIB file, just the drop shadow is in the controller
- I simplified the way we open the reports - it is now parameterless, it will always use the default workspace when opened

### Side effects
I hope this does not break Android in any way. I would be glad if somebody from the @toggl/androids checked this and confirmed it.

## Review hints
- Start with a user with just one workspace, check that the button is not there at all.
- Add a few workspaces using the web app and sync the mobile app - now you should see the button and see that you can swap the workspaces and see different reports
- Add a workspace with a very long name and check how the fade out works

## :squid: Permissions
Squash whenever


Toggl.Daneel/Services/DialogService.cs
```diff
@@ -109,6 +110,43 @@ public IObservable<Unit> Alert(string title, string message, string buttonTitle)
...
+                        observer.OnNext(value);
+                        observer.OnCompleted();
+                    });
+
+                    actionSheet.AddAction(action);
+                }
+
+                var cancelAction = UIAlertAction.Create(Resources.Cancel, UIAlertActionStyle.Cancel, _ =>
+                {
+                    observer.OnNext(null);
```
khoi: Me not reviewing but learning c#: 
`await`ing an Observable which contains no element causes an exception

Is there any other `await` syntax so that an `empty` Observable returns null instead of throwing?

Toggl.Daneel/ViewControllers/ReportsViewController.cs
```diff
@@ -47,6 +52,14 @@ public override void ViewDidLoad()
             source.OnScroll += onReportsTableScrolled;
             ReportsTableView.Source = source;
 
+            var areThereEnoughWorkspaces = new LambdaConverter<IDictionary<string, IThreadSafeWorkspace>, bool>(workspaces => workspaces.Count > 1);
+            var isWorkspaceNameTooLong = new LambdaConverter<string, bool>(_ =>
+            {
+                WorkspaceLabel.SetNeedsLayout();
```
khoi: either `SetNeedsLayout` or `LayoutIfNeeded`, not both

Toggl.Daneel/ViewControllers/ReportsViewController.xib
```diff
@@ -41,11 +44,57 @@
...
+                                    <nil key="highlightedColor"/>
+                                </label>
+                            </subviews>
+                            <color key="backgroundColor" white="1" alpha="1" colorSpace="custom" customColorSpace="genericGamma22GrayColorSpace"/>
+                            <constraints>
+                                <constraint firstItem="m1j-gc-5S9" firstAttribute="leading" secondItem="aWZ-cY-YUv" secondAttribute="leading" id="2UY-AA-6pF"/>
+                                <constraint firstItem="m1j-gc-5S9" firstAttribute="top" secondItem="aWZ-cY-YUv" secondAttribute="top" id="6L5-zQ-Kiz"/>
+                                <constraint firstAttribute="trailing" secondItem="m1j-gc-5S9" secondAttribute="trailing" id="SCc-MH-0v7"/>
+                                <constraint firstAttribute="bottom" secondItem="m1j-gc-5S9" secondAttribute="bottom" id="aRP-29-x8E"/>
+                                <constraint firstAttribute="width" relation="lessThanOrEqual" constant="144" id="hJE-N6-4K3"/>
```
sofokli1: Here you set the max width to 144 but in the VC you check for 140. Any reason for this?

Toggl.Daneel/Services/DialogService.cs
simonrozsival: I could probably write my own awaiter or I could just use try-catch block and if it throws, I know nothing was emitted before the observable completed.

Do you think that returning null is weird? That it could be a valid value sometimes?

Toggl.Daneel/ViewControllers/ReportsViewController.cs
simonrozsival: Using just one does not work. I need to force the UI to update the layout when the text was changed so I can look how wide the label is now with the new text. Why should I not use both? I know that it would be better to wait for the next cycle... but I need the width right at the line 60! :) Is there a better way to force layout or how to measure the text of the text with the given font settings?

Toggl.Daneel/ViewControllers/ReportsViewController.xib
simonrozsival: I will change this to 144.

Toggl.Daneel/ViewControllers/ReportsViewController.cs
simonrozsival: I will give this a try, it looks much better than my code!

daividssilverio:
Yeah, there is one :) #2703

PR #2804, opened 2018-08-02
Add IQ/Realm docs

kornelijepetak:
## What's this?

This PR contains a preliminary docs about some of the inner workings of Realm database, as well as an introduction into IQueryable interface and its purpose and its lifestyle.

The consequences of these ideas on our app will come tomorrow under `Prime Radiant` header. Come on! Even Asimov told the stories in sequels... Be patient.

## Relationships

:broken_heart: None!

## Why do we want this?

**Who we are?** Developers!
**What do we want?** More docs!
**When do we want?** After approval!

But basically, it helps people who did not grow up with .NET in their yard get into ideas of IQueryables and also having enumerated Realm gotchas in one place could be helpful.

## Review hints

Well, read and criticize.

Too verbose? Speak up!
Too boring? Speak up!
Irrelevant? Speak up!
Already knew all of this? Speak up!
Learned anything? Speak up!
You have more insight into the subject and want to share, speak up! Especially you!

Just, say it!

## :squid: Permissions

Please, no seafood. This is a WIP.


docs/realm/realm.md
```diff
@@ -0,0 +1,246 @@
...
+    .Select(group => new { Letter: g.Key, Count: g.Count() })
+    .ToList();
+```
+The problem with `IQueryable` though is that for the same query, a query provider can't just be switched around with the expectation of the same behavior. And this is because the query provider can support, or not support, certain operators.
+
+And that leads us to Realm Database Engine.
+
+## :package: Realm
+
+To make things clear from the start, Realm is a **pure object database**. No relations, tables, attributes, foreign keys, normal forms and all other typical RDB fuss.
```
juanlaube: Realm has some sort of relationships, but maybe that's not available in the .Net wrapper 🤔

docs/realm/realm.md
```diff
@@ -0,0 +1,246 @@
...
+
+POCO all the way (almost). 
+
+:warning: A few important considerations about the data model, though:
+
+* Classes that need to be persisted **must** inherit from `RealmObject`
+  * This sucks in general case because it prevents us from creating our own inheritance hierarchies, but since we are trying to avoid needless inheritance chains, we don't really care about this.
+* Only classes that inherit from `RealmObject` will be persisted. Classes that inherit from classes that inherit from `RealmObject` are ignored and will not work with Realm.
+  * Well, we don't care much about this either.
+* Every class that needs to be persisted **must** have a public, parameterless constructor (implicit, or explicit).
```
khoi: I'm looking at `RealmWorkspace.cs` and couldn't find its public constructor. Is the constructor synthesized by the compiler?

docs/realm/realm.md
```diff
@@ -0,0 +1,246 @@
...
+
+Those are enumerables so the result is the same.
+However, if we try to do the same with Realm, things will go :boom:.
+
+```C#
+var elders1 = realmInstance.All<Person>().Where(p => p.Age > 65).ToList();
+var elders2 = realmInstance.All<Person>().Where(p => 65 < p.Age).ToList();
+```
+
+The second query will throw: `System.NotSupportedException: 'The left-hand side of the LessThan operator must be a direct access to a persisted property in Realm.'`
```
khoi: 😱whatt?

docs/realm/realm.md
```diff
@@ -0,0 +1,246 @@
...
+In the case of `IEnumerable`, the getter of the `Text` property of every object in the collection will be accessed. However, in the case of `IQueryable` we actually have no idea what can happen. That depends on the _query provider_ used.
+
+If we look at the `Where` extension methods, we can see the following signatures:
+
+```C#
+IEnumerable<T> Where<T>(this IEnumerable<T> collection, Func<T, bool> predicate);
+IQueryable<T> Where<T>(this IQueryable<T> collection, Expression<Func<T, bool>> predicate);
+```
+
+So instead of the `Func<T, bool>` we have `Expression<Func<T, bool>>`. That means that our `item => item.Text.StartsWith("Data")` lambda is actually data, and not the code to be executed as is.
```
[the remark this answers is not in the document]
kornelijepetak: It is somewhat convoluted, yes. I've tried once, but I didn't have a good use case, so I dropped it. But basically, you just store your query info until the time that query is enumerated over, at that point, you just use the collected info to determine what to return as the enumerated data.

docs/realm/realm.md
```diff
@@ -0,0 +1,246 @@
...
+
+Under the hood, there is a very optimized C++ core engine and different platforms simply bind to that engine. In part, this is a source of Realm's speed. 
+
+The other part of its speed lies in the fact that its memory model and the file format model are the same. This memory-mapping of the files allows for less searching thru files and loading less data into memory. :horse_racing: Speed!
+
+It is **ACID**-compliant and uses [MVCC for concurrency control](https://en.wikipedia.org/wiki/Multiversion_concurrency_control).
+
+### :blue_book: Data Model
+
+POCO all the way (almost).
```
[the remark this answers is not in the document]
kornelijepetak: POCO = Plain Old CLR Object

docs/realm/realm.md
kornelijepetak: If I remember correctly, all those constructors are generated by a `.tt` file (T4 template).

docs/realm/realm.md
kornelijepetak: Yeah, they could've supported that as well, but somehow they just didn't. :)

daividssilverio:
Looking good so far o/

docs/realm/realm.md
kornelijepetak: Not `relationship` in a relational-database sense of the word. But yes, there are linked entities (which constitute a kind of relationship).

kornelijepetak:
We are doing a lots of discussions regarding Realm and Prime Radiant internally so to avoid potentially useless information, I will merge this PR with what we have here. 

I just need a single approval.

PR #2805, opened 2018-08-02
Undo deletion

### LINKED ISSUE #2363: Undo time entry deletion
When swipe-to-deleting a time entry, we should show a little toast that allows the user to undo.

Technically, we will only hide the TE from the UI, and only delete it if:
- [ ] the user doesn't interact with the TOAST fast enough
- [ ] the user deletes another TE (only one undo at a time)
- [ ] the user navigates to another screen (start, edit(?), reports, settings..)
- [ ] the user leaves the app/the app becomes inactive
- [ ] the user dismisses the toast (should we add this functionality)

@pe0ny What we need are:
- [ ] designs for the toast for both platforms (please make one issue per platform)
- [ ] the amount of time we give the user to undo (5 seconds? 10? 20? infinite (until they do something else))
- [ ] whether the user can dismiss the toast themselves to make it go away (without undoing)
- [ ] whether going to edit an entry dismisses the toast and deletes the entry
- [ ] whether to use this also when deleting entries from the edit dialogue (and if so, do we still need to ask for confirmation?)

Odrakir:
## What's this?
Adds UNDO functionality for the deletion of time entries from the timeline.

### Relationships
Closes #2363 

## Why do we want this?
Because 💩 happens and users may slide without actually wanting to delete an entry.

## How is it done?
We basically delete it from the Time entries view model, not the DB. If the user UNDOs we add it again. If the app crashes or he refreshes or whatever, the entry is still there. It only gets actually deleted when the specified time passes.

### Side effects
The app is much cooler now.

## :squid: Permissions
This is a WIP. Don't.


Toggl.Foundation.MvvmCross/ViewModels/MainViewModel.cs
```diff
@@ -76,6 +76,8 @@ public sealed class MainViewModel : MvxViewModel
         // Inputs
         public InputAction<TimeEntryViewModel> ContinueTimeEntry { get; }
         public InputAction<TimeEntryViewModel> DeleteTimeEntry { get; }
+        public InputAction<TimeEntryViewModel> PreDeleteTimeEntry { get; }
```
daividssilverio: wouldn't SoftDeleteTimeEntry sound a little better?

Odrakir: Yes, I like that.

Toggl.Daneel/Views/SnackBar.cs
```diff
@@ -188,5 +190,13 @@ private void configureLayout()
 
             SetNeedsLayout();
         }
+
+        public static SnackBar Undo(Action onTap, Action onTimer)
+        {
+            var snackBar = SnackBar.Create("Time entry was deleted");
```
amulware: Strings should be in resource files.

Toggl.Daneel/Views/SnackBar.cs
```diff
@@ -188,5 +190,13 @@ private void configureLayout()
 
             SetNeedsLayout();
         }
+
+        public static SnackBar Undo(Action onTap, Action onTimer)
```
amulware: I'm not sure what I think about having the specific use case of the snackbar in the same class itself. There might be more here soon, and that's an odd pattern to me. Shouldn't the class be agnostic to the various use cases?

amulware:
Another edge case: What if:
- user deletes
- TE is updated from sync
- user undos?
Will it show the TE with it's new proper state?

Toggl.Daneel/Views/SnackBar.cs
Odrakir: I think it's useful to have a way to instantiate and configure common SnackBars in an easy and convenient way. And in my opinion it makes sense to have that tied to the class itself. Any other ideas?

amulware: You say 'common', but it's only used in one place. :D I think it would be better to have that code in that place, but I can also see why you didn't want that to keep it cleaner. I'm not sure.

Odrakir:
Edge cases comments and questions:
1 - Soft delete -> new server update to that entry (before 5 seconds) -> User does nothing: Entry will be deleted.
2 - Soft delete -> User leaves the app (before 5 seconds): Entry will be deleted (probably not in current pushed code, but that's the idea, deletion will be executed on dispose)
3 - Soft delete -> new server update to that entry (before 5 seconds) -> User undos: Right now entry will be duplicated.

I think 2 is correct. Not sure about what the correct behavior should be for 1 and 3.

Toggl.Daneel/Views/SnackBar.cs
simonrozsival: I like the factory method, but it could be used in some static `SnackBarFactory` class.

Odrakir: Yeah, "common" is not the word. In Swift I would make an extension with the static methods to create those SnackBars. I could create a SnackBarFactory if you don't feel confortable with this. I'm totally ok with it, though.

Toggl.Daneel/ViewControllers/MainViewController.cs
```diff
@@ -119,9 +122,22 @@ public override void ViewDidLoad()
                 tableViewSource.SwipeToContinue
             );
             this.Bind(continueTimeEntry, ViewModel.ContinueTimeEntry);
-            this.Bind(tableViewSource.SwipeToDelete, ViewModel.DeleteTimeEntry);
+            this.Bind(tableViewSource.SwipeToDelete, ViewModel.PreDeleteTimeEntry);
             this.Bind(tableViewSource.ItemSelected, ViewModel.SelectTimeEntry);
 
+            this.Bind(tableViewSource.SwipeToDelete, timeEntry =>
+            {
```
kornelijepetak: I would appreciate if this was a named function instead of the lambda, so that bindings are not interwoven with behavior.

Toggl.Daneel/Views/SnackBar.cs
kornelijepetak: I wanted to suggest a factory, and then I saw @simonrozsival already did.

amulware:
Edge cases:
2 sounds fine, yeah.
I'm unsure why 1 and 2 are so different though. Will the update from the server cause the entry to reappear? If so, it shouldn't. The deletion should still go through.
And of course, in 3, it should show the correct new entry, and not the old state, and certainly no duplicate.

Odrakir:
In the case of one the actual deletion hasn't happened yet, so a new entry might appear with the same ID, then syncing will do the right stuff and there will only be one of those. But when the deletion kicks in, it'll disappear.

In the case of three the same thing happens, but when the user hits UNDO (after the reappearing of the entry because of sync), the entry will go back to the view model (not the DB). So we will have two entries in the table, but only one in the DB. In next refresh it should be fixed.

But anyway, I'm giving this another though, because I think it makes sense to move some of the logic out of the snack bar and vc and into the vm.

Odrakir:
I changed this a lot. I moved the logic to the view model in an effort to make this simple (I hope I haven't made it more complicated).

Toggl.Daneel/Extensions/Binders/ReactiveTableViewBinder.cs
```diff
@@ -40,6 +41,12 @@ public void handleCollectionChanges(IEnumerable<CollectionChange> changes)
         {
             lock (animationLock)
             {
+                if (changes.Any( c => c.Type == CollectionChangeType.Reload))
```
kornelijepetak: some strange whitespace... nothing alarming.

Toggl.Daneel/Views/SnackBar.cs
```diff
@@ -115,12 +106,12 @@ public void Show(string text, UIView superView)
                     Alpha = 1;
                 }
             );
-
-            startTimer?.Invoke();
         }
 
-        public void Hide()
+        public void Hide(bool execute = true)
```
juanlaube: `execute` is not being used, is this something from a previous commit? :eyes:

Toggl.Foundation.MvvmCross/ViewModels/MainViewModel.cs
```diff
@@ -398,15 +405,49 @@ private IObservable<Unit> timeEntrySelected(TimeEntryViewModel timeEntry)
...
 
-        private IObservable<Unit> deleteTimeEntry(TimeEntryViewModel timeEntry)
+        private IObservable<Unit> delayDeleteTimeEntry(TimeEntryViewModel timeEntry)
+        {
+            timeEntryToDelete = timeEntry;
+
+            timeEntriesViewModel.RemoveTimeEntryFromVieWModel(timeEntry);
+            showUndoSubject.OnNext(true);
+
+            delayedDeletionDisposable = Observable.Merge( // If 5 seconds pass or we try to delete another TE
```
juanlaube: `delayedDeletionDisposable` needs to be disposed also in the `Dispose` implementation of this class

Odrakir:
The timer is still running, so it'll get deleted in less than 5 seconds. I think that's enough, otherwise we would have to find a way to communicate that from the vc to the vm.

Toggl.Foundation.MvvmCross/ViewModels/MainViewModel.cs
Odrakir: Actually... will this VM ever get disposed?

simonrozsival:
> 1 - Soft delete -> new server update to that entry (before 5 seconds) -> User does nothing: Entry will be deleted.

Yes, this is how it works right now. I will try to make sure that a soft-deleted TE does not appear in the log when it is pulled from the server.

> 2 - Soft delete -> User leaves the app (before 5 seconds): Entry will be deleted (probably not in current pushed code, but that's the idea, deletion will be executed on dispose)

Yes, this works well.

> 3 - Soft delete -> new server update to that entry (before 5 seconds) -> User undos: Right now entry will be duplicated.

No, there is no duplication (Ricardo already fixed this).

I don't know about any other edge cases right now. I will fix case `1` and then I think this will be ready for merging.

simonrozsival:
I fixed the expected behavior of this PR and now this is missing just unit tests which I will write today and tomorrow. Please continue testing this PR so we can merge it as soon as there are unit tests for the new commands and they are passing.

khoi:
According to the spec, the snackbar should be dismissed and TE deleted when users navigate to other screens. But that doesn't seem to be the case now.

simonrozsival:
It is true that if you delete a time entry, go to settings/reports/start a new time entry and then you _quickly_ go back, you can still undo deleting the TE - but only if you do that in under 5 seconds. IMO it doesn't make any sense to detect if the user navigates to another view and hide it at that moment, when it will eventually be deleted anyway after a few seconds.

Toggl.Foundation.MvvmCross/ViewModels/MainViewModel.cs
```diff
@@ -131,7 +132,8 @@ public sealed class MainViewModel : MvxViewModel
             IInteractorFactory interactorFactory,
             IMvxNavigationService navigationService,
             IRemoteConfigService remoteConfigService,
-            ISuggestionProviderContainer suggestionProviders)
+            ISuggestionProviderContainer suggestionProviders,
+            IScheduler scheduler)
```
daividssilverio: There is a new `ISchedulerProvider` available now and it's already being injected in the MainViewModel.  If you merge develop into this branch, you can use that instead.

Toggl.Foundation.MvvmCross/ViewModels/TimeEntriesViewModel.cs
```diff
@@ -30,21 +33,39 @@ public sealed class TimeEntriesViewModel
...
+        public IObservable<bool> ShouldShowUndo => showUndoSubject.AsObservable();
+
+        public InputAction<TimeEntryViewModel> DelayDeleteTimeEntry { get; }
+        public UIAction CancelDeleteTimeEntry { get; }
+
         public TimeEntriesViewModel (ITogglDataSource dataSource,
-                                     IInteractorFactory interactorFactory)
+                                     IInteractorFactory interactorFactory,
+                                     IAnalyticsService analyticsService,
+                                     IScheduler scheduler)
```
daividssilverio: I'd give preference to the ISchedulerProvider as well.

Toggl.Foundation.Tests/MvvmCross/ViewModels/MainViewModelTests.cs
```diff
@@ -83,7 +84,8 @@ public sealed class TheConstructor : MainViewModelTest
                 bool useInteractorFactory,
                 bool useNavigationService,
                 bool useRemoteConfigService,
-                bool useSuggestionProviderContainer)
+                bool useSuggestionProviderContainer,
+                bool useScheduler)
```
daividssilverio: This was also updated to use the ISchedulerProvider on develop. Please check.

Toggl.Foundation.Tests/MvvmCross/ViewModels/TimeEntriesViewModelTests.cs
```diff
@@ -45,13 +55,17 @@ public sealed class TheConstructor : TimeEntriesViewModelTest
             [ConstructorData]
             public void ThrowsIfAnyOfTheArgumentsIsNull(
                 bool useDataSource,
-                bool useInteractorFactory)
+                bool useInteractorFactory,
+                bool useAnalyticsService,
+                bool useScheduler)
```
daividssilverio: Here too. ISchedulerProvider.

Toggl.Foundation.Tests/MvvmCross/ViewModels/MainViewModelTests.cs
```diff
@@ -386,7 +384,7 @@ public TheStopTimeEntryCommand()
                 DataSource.TimeEntries.CurrentlyRunningTimeEntry.Returns(observable);
 
                 ViewModel.Initialize().Wait();
-                Scheduler.AdvanceBy(TimeSpan.FromMilliseconds(50).Ticks);
+                SchedulerProvider.TestScheduler.AdvanceBy(TimeSpan.FromMilliseconds(50).Ticks);
```
daividssilverio: The TestSchedulerProvider have both Default and MainScheduler set to the property `TestScheduler` created for the `BaseMvvmCrossTests`
There is a property TestScheduler for MvxTests that holds that, you can use that shortcut if you want. Up to you.

This would become: `TestScheduler.AdvanceBy(TimeSpan.FromMilliseconds(50).Ticks);`

Toggl.Giskard/Toggl.Giskard.csproj
```diff
@@ -262,7 +262,6 @@
     <Compile Include="Resources\Resource.designer.cs" />
     <Compile Include="Properties\AssemblyInfo.cs" />
     <Compile Include="LinkerPleaseInclude.cs" />
-    <Compile Include="Startup\AndroidSchedulerProvider.cs" />
```
daividssilverio: Oops, bad merge it seems.

Toggl.Giskard/Toggl.Giskard.csproj
```diff
@@ -415,7 +414,6 @@
     <Compile Include="Views\AutocompleteEditText.cs" />
     <Compile Include="Views\NonSwipingViewPager.cs" />
     <Compile Include="Activities\MainActivity.ViewBinds.cs" />
-    <Compile Include="Startup\HandlerScheduler.cs" />
```
daividssilverio: Here too.

Toggl.Giskard/Startup/TogglApplication.cs
```diff
@@ -13,17 +13,5 @@ public class TogglApplication : MvxAppCompatApplication<Setup, App<LoginViewMode
         public TogglApplication(IntPtr javaReference, JniHandleOwnership transfer) : base(javaReference, transfer)
         {
         }
-
-        public override void OnCreate()
```
daividssilverio: Oops, bad merge it seems too.

PR #2807, opened 2018-08-02
Fix incorrect behaviour when closing VMs

### LINKED ISSUE #1861: After creating a new project, the app should direct user back to the time entry
Currently, it takes user back to the Projects view with the new project name typed in.

daividssilverio:
## What's this?
This is a workaround for the current implementation of the relationship between the `MvxPresenter:CurrentActivity` property and the lifecycle callback management that handles the process of updating the current activity, provided by IMvxAndroidCurrentTopActivity.

### Relationships
It seems the fix here doesn't fix the problem in our case: https://github.com/MvvmCross/MvvmCross/pull/2513

Closes #1861

## Why do we want this?
To provide the expected experience when creating projects when editing TEs.
Also, this would fix other interactions problems we might have when the hierarchy of screens is bigger than 1.

## How is it done?
I copied (yes...) the current `MvxApplicationCallbacksCurrentTopActivity` implementation, modified the visibility of the list of activities it keeps and added a method to look for activities based on the viewmodel (and tried to make this class less leak prone).
Then, when presenter tries to close the current VM, if it current activity is not updated yet (nulled), asking the new implementation of the IMvxAndroidCurrentTopActivity will look for a matching activity that holds the VM that is being closed, finishing it if existent.

### Why not another way?
I couldn't just override the CurrentActivity property because the MvxApplicationCallbacksCurrentTopActivity does not expose the list of activities it keeps.
I'll submit an Issue and PR to MvvmCross suggesting to modify the visibility of MvxApplicationCallbacksCurrentTopActivity. We can then just override what is necessary in our case.

In the case of #1861, I tried adding a Task.wait to wait for the Presenter to trigger `finish` on SelectProjectActivity, but everything happens so fast the SelectProjectActivity doesn't even have a chance to become the CurrentActivity.

### Side effects
There might be hidden side effects from using the `MvxAppCompatApplication`, but it didn't get anything during my tests.
Overriding the CloseActivity on the `TogglPresenter` also seems pretty safe, and I didn't find anything wrong in the app with the new implementation.

## Review hints
* Testing all transitions in the app would be perfect to check if something is broken.
* Avoid testing with the "Don't keep activities" mode on. There is another bug/crash happening when you return to the MainActivity from the EditTimeEntryActivity when that mode is enabled, which also happens on develop.

## :squid: Permissions
🦑 whenever


Toggl.Giskard/Startup/TogglApplication.cs
```diff
@@ -0,0 +1,17 @@
...
+using Android.App;
+using Android.Runtime;
+using MvvmCross.Droid.Support.V7.AppCompat;
+using Toggl.Foundation.MvvmCross;
+using Toggl.Foundation.MvvmCross.ViewModels;
+
+namespace Toggl.Giskard
+{
+    [Application]
+    public class TogglApplication : MvxAppCompatApplication<Setup, App<LoginViewModel>>
```
willsb: Is this needed?

Toggl.Giskard/Startup/QueryableMvxLifecycleMonitorCurrentTopActivity.cs
```diff
@@ -0,0 +1,96 @@
...
+using System.Linq;
+using Android.App;
+using Android.OS;
+using MvvmCross.Droid.Support.V7.AppCompat;
+using MvvmCross.Platforms.Android;
+using MvvmCross.ViewModels;
+
+namespace Toggl.Giskard
+{
+    public class QueryableMvxLifecycleMonitorCurrentTopActivity : Java.Lang.Object, Application.IActivityLifecycleCallbacks, IMvxAndroidCurrentTopActivity
```
willsb: As much as I understand this, it's really gross. Could you also submit a PR to MvvmCross to make the property you need available?

Toggl.Giskard/Startup/TogglApplication.cs
daividssilverio: Yeah, I need the application to be MvxAppCompatApplication to register lifecycle callbacks here:
https://github.com/toggl/mobileapp/blob/f6565ab2c58c2d308dee61e0970dfd780788bef6/Toggl.Giskard/Startup/Setup.cs#L123

Toggl.Giskard/Startup/QueryableMvxLifecycleMonitorCurrentTopActivity.cs
daividssilverio: I'm doing that.

daividssilverio: https://github.com/MvvmCross/MvvmCross/issues/3048

daividssilverio: The PR got merged. We can remove this class when we update MvvmCross.

Issue #2628, opened 2018-07-09
WS removal: Detect workspaces the user lost access to


amulware:
We need a sync-state that does the following:

- [x] compare the list of workspaces fetched on pull sync with the workspaces in the app that have been synced (ignore local workspaces that have not been synced)
- [x] those that we know locally but did not get pulled are workspaces the user got removed from
- [x] pass those states on to the state which will remove data from those workspaces

We have to think about how exactly the flow works here. E.g. does the state pass on the list of all workspaces we were removed from? Or do we handle them one by one?

Also, at what point during the follow-up process do we mark the workspace itself as ghost/remove it? We have to make sure that if the app gets killed/syncing stops at any point, the local data remains in a valid state and will continue properly on next sync.

simonrozsival:
I would say that the result should be a list of workspaces. That will be much easier and is more straightforward as the HTTP response gives us a list of workspaces that the user has access to.

The follow-up process will contain
1. removing all synced data
2. setting the workspace's `IsGhost` property to true

If the app is terminated while we delete entities in 1., it is fine - next time sync will run we will finish it, because we hadn't marked the ws as ghost yet so it will be detected next time. Also we deleted only synced stuff, which we didn't need in the meantime.

It might look weird in some edge cases:
- we delete half of the stuff we want to delete, user kills the app, goes offline, goes back to the app, sees only half of the data
- we delete half of the stuff, user kills the app, user gains back access to the workspace, goes back to the app, sees only half of the data and he/she can't pull the missing ones (because of `since` dates)

amulware:
What if we set IsGhost first to avoid the second edge case? We only have to make sure to include time entries in the cleanup loop then.

simonrozsival:
Or we just set the workspace to `IsGhost` and in a separate `clean-up loop` we will look for all ghost workspaces and try to prune as much data as possible. That would be immune to any interruptions to the clean-up process.

amulware:
Agreed. Though we should at least trigger time entry clean up as soon as possible to make sure we don't show those for any longer than needed.

amulware:
Looks to me like this was closed by the PR above.

PR #2813, opened 2018-08-03
Detect losing access to workspace

simonrozsival:
## What's this?
This PR adds a sync state which detects that user lost access to a workspace.

### Relationships
Ref #2628 

## How is it done?
A simple pull state which compares the local data and the data received from the server.

## :squid: Permissions
Squash whenever


Toggl.Foundation/Sync/States/Pull/DetectLosingAccessToWorkspacesState.cs
```diff
@@ -0,0 +1,43 @@
...
+                    storedWorkspaces
+                        .Where(storedWorkspace =>
+                            fetchedWorkspaces.None(fetchedWorkspace => fetchedWorkspace.Id == storedWorkspace.Id))
+                        .ToList())
+                .Select(missingWorkspaces => missingWorkspaces.Any()
+                    ? LostAccessTo.Transition(missingWorkspaces)
+                    : (ITransition)NoAccessLost.Transition());
+
+        private IObservable<IEnumerable<IThreadSafeWorkspace>> allStoredWorkspaces()
+            => dataSource.GetAll(ws => ws.Id > 0);
```
amulware: This should probably filter out ghosts as well? Otherwise, we'll continue to ghost the same workspaces.

Toggl.Foundation/Sync/States/Pull/DetectLosingAccessToWorkspacesState.cs
```diff
@@ -0,0 +1,43 @@
...
+        public DetectLosingAccessToWorkspacesState(IDataSource<IThreadSafeWorkspace, IDatabaseWorkspace> dataSource)
+        {
+            Ensure.Argument.IsNotNull(dataSource, nameof(dataSource));
+
+            this.dataSource = dataSource;
+        }
+
+        public IObservable<ITransition> Start(IFetchObservables fetchObservables)
+            => fetchObservables.GetList<IWorkspace>()
+                .Zip(allStoredWorkspaces(), (fetchedWorkspaces, storedWorkspaces) =>
```
amulware: I'm not really convinced by the Zip here. We're zipping only one item each. Why not use imperative code instead? Or do something like `allStoredWorkspaces().Except(...)/Where(...)`?
Or if you really only wanna GetList<> once, `GetList<>().Select(fetchedWorkspaces => allStoredWorkspaces().Except(...)/Where(...))`

Toggl.Foundation.Tests/Sync/States/Pull/DetectLosingAccessToWorkspacesStateTests.cs
```diff
@@ -0,0 +1,118 @@
...
+                var state = new DetectLosingAccessToWorkspacesState(dataSource);
+
+                var transition = await state.Start(fetchObservables);
+
+                transition.Result.Should().Be(state.LostAccessTo);
+                ((Transition<IEnumerable<IThreadSafeWorkspace>>)transition).Parameter.Should().HaveCount(1);
+            }
+
+            [Fact]
+            public async Task ReturnsDifferentResultWhenNoAccessIsLost()
```
amulware: 'DifferentResult'? Why not be specific?

Toggl.Foundation.Tests/Sync/States/Pull/DetectLosingAccessToWorkspacesStateTests.cs
```diff
@@ -0,0 +1,118 @@
...
+                prepareFetch(new List<IWorkspace>
+                {
+                    new MockWorkspace { Id = 1 }
+                });
+                var state = new DetectLosingAccessToWorkspacesState(dataSource);
+
+                var transition = await state.Start(fetchObservables);
+
+                transition.Result.Should().Be(state.LostAccessTo);
+                ((Transition<IEnumerable<IThreadSafeWorkspace>>)transition).Parameter.Should().HaveCount(2);
```
amulware: I'd like it to check for the ids as well, just to be sure.

Toggl.Foundation.Tests/Sync/States/Pull/DetectLosingAccessToWorkspacesStateTests.cs
```diff
@@ -0,0 +1,118 @@
...
+                prepareFetch(new List<IWorkspace>
+                {
+                    new MockWorkspace { Id = 1 }
+                });
+                var state = new DetectLosingAccessToWorkspacesState(dataSource);
+
+                var transition = await state.Start(fetchObservables);
+
+                transition.Result.Should().Be(state.LostAccessTo);
+                ((Transition<IEnumerable<IThreadSafeWorkspace>>)transition).Parameter.Should().HaveCount(1);
```
amulware: Check for ids here too?

amulware:
I want to make sure it's not only us knowing how it works again afterwards, so I'd like at least one more review. Maybe from @juanlaube, @sofokli1 or @khoi? The more the merrier!

simonrozsival:
@sofokli1 I don't think we should use this in the state machine until there is a state which processes the results. It will be in a separate PR.

Toggl.Foundation.Tests/Sync/States/Pull/DetectLosingAccessToWorkspacesStateTests.cs
```diff
@@ -0,0 +1,122 @@
...
+                    new MockWorkspace { Id = 1 }
+                });
+                var state = new DetectLosingAccessToWorkspacesState(dataSource);
+
+                var transition = await state.Start(fetchObservables);
+                var parameter = ((Transition<IEnumerable<IThreadSafeWorkspace>>)transition).Parameter.ToList();
+
+                transition.Result.Should().Be(state.LostAccessTo);
+                parameter.Should().HaveCount(2);
+                parameter.Should().Contain(ws => ws.Id == 2 || ws.Id == 3);
```
amulware: It should contain 2 OR 3? (1, 2) would pass that check while it should not. I think that should be an All call or maybe just have this line twice, once for each id?

Toggl.Foundation/Sync/States/Pull/DetectLosingAccessToWorkspacesState.cs
```diff
@@ -0,0 +1,44 @@
...
+            Ensure.Argument.IsNotNull(dataSource, nameof(dataSource));
+
+            this.dataSource = dataSource;
+        }
+
+        public IObservable<ITransition> Start(IFetchObservables fetchObservables)
+            => fetchObservables.GetList<IWorkspace>()
+                .SelectMany(fetchedWorkspaces =>
+                    allStoredWorkspaces()
+                        .SelectMany(CommonFunctions.Identity)
```
amulware: Maybe put the SelectMany into allStoredWorkspaces?

Toggl.Foundation/Sync/States/Pull/DetectLosingAccessToWorkspacesState.cs
```diff
@@ -0,0 +1,44 @@
...
+                        .Where(storedWorkspace =>
+                            fetchedWorkspaces.None(fetchedWorkspace => fetchedWorkspace.Id == storedWorkspace.Id))
```
amulware: We're four => deep here. Kornelije is going to have a heart attack. :D 

I think semantically it makes a lot of sense to put the `fetchedWorksaces =>` lambda into a function, so that `Start` doesn't have to deal with all these details. (Layers of abstraction and all that)

Issue #2782, opened 2018-07-31
Select workspace view crashes when default workspace isn't set


arturs105:
## 🐛 Bug Report
When the user doesn't have a default workspace set, nothing in the workspace row in the settings is shown and pressing the row to select a workspace crashes the app.

## Expected behavior
After inspecting the code i found out that if there is no default workspace, then workspaces should be ordered alphabetically and then first one chosen as the active workspace. But that code isn't working properly for some reason.

## To reproduce
1. Create a few workspaces on the web app
3. Delete the default workspace on the web app
4. Log in with the mobile app.

simonrozsival:
I am not able to reproduce having no default workspace. When I leave the default workspace, I automatically get a different ws as the default one. Is there some trick how to do it?

daividssilverio:
There are two tricks to reproduce the state of "workspace-less":

1. With two accounts
Create two test accounts. `A` and `B`.
`A` creates 2 workspaces, `a1` and `a2`.
`A` invites `B` to both workspaces.
`B` accepts only the invite to `a1`.
`B` leaves its own default workspace.
`A` deletes `B` from workspace `a1`.

or

2. With three accounts (thanks @juanlaube )
`A` invites `C`
`B` invites `C`
`C` accepts `A`
`C` leaves `C`’s workspaces
`A` removes `C`
The invite from `B` is what keeps `C` alive

I Just realized this is not in question here. But the trick is there.

daividssilverio:
Actually, after becoming workspace-less from the steps above I was able to reproduce the state of having workspaces but without having a default workspace.

After becoming workspace-less:
1. Login on the web app.
2. You will be requested to create a new workspace.
3. Create one. (don't set it as default)
4. Create some more.
5. Logout and login on the mobile app.

daividssilverio:
But selecting a row the list won't crash giskard.

simonrozsival:
Well I think this issue is not related to being in no workspace, but rather having `null` as the default workspace. I came to the conclusion that I can set this manually in Realm studio.

daividssilverio:
So, the expected behaviour would be having the user `DefaultWorkspaceId` set to the result of `GetDefaultWorkspaceInteractor` when there's no `DefaultWorkspaceId` is null?

simonrozsival:
When the user has no workspace, then the code in the `PickDefaultWorkspace` method of the  `SettingsViewModel` will crash the app (null reference).

[...]

PR #2816, opened 2018-08-03
Fix bugs in NoWorkspaceViewModel related to syncManager

### LINKED ISSUE #2769: Bugs in NoWorkspaceViewModel
## 🐛 Bug Report
Some bugs were discovered while testing the NoWorkspace UI.

- [x] `dataSource.User.Current` must call `FirstAsync()` otherwise it will never complete
- [ ] It doesn't sync the new workspace after creation. This seems simple, but it requires that a new `SyncManager` gets created and the old one gets disposed. The old one isn't usable because it has encountered an error (`NoWorkspaceException`) and is 💀.
- [ ] `TryAgain` method needs to be fixed. Due to the same SyncManager failure, a new one is required each time `TryAgain` is executed.
- [ ] When the issues are solved, the view presentation should be re-enabled in `ErrorHandlingService`.

To the person, who works on this: while you're at it, you might as well convert the view model's public methods to the new `RxAction`s (#2751).

## Expected behavior

No crashes 🙂

sofokli1:
## What's this?
Bug fixes to the `NoWorkspaceViewModel`. I am not addressing the part of the issue about `BindToAction` since i am not very familiar with it.

### Relationships

Closes #2769

## Why do we want this?
When the syncManager fails it is not usable anymore so we need to create a new one.

## How is it done?
It adds the to the `TogglDataSource` the ability to create a new sync manager. And this is used in the `NoWorkspaceViewModel`. It is quite tricky to test and reproduce for testing purposes. I myself am not 100% sure it works correctly. On the other hand I am not sure it is even needed since (at least from I see in the code) the sync manager is not frozen if no workspace exception is fired so in a way it should work without creating a new syncManager.

### Why not another way?

This is the only way I thought off.

### Side effects
Not sure. Either none or everything will break

## Review hints
I call on brave souls to try and test this in a functional way.
Would advise to try the steps from 
https://github.com/toggl/mobileapp/issues/2782#issuecomment-410079411
I myself am still trying to make this way to work

## :squid: Permissions
Whenever


Toggl.Foundation.MvvmCross/ViewModels/NoWorkspaceViewModel.cs
```diff
@@ -35,6 +35,8 @@ public async Task TryAgain()
         {
             isLoading.OnNext(true);
 
+            dataSource.CreateNewSyncManager();
+
             var workspaces = await dataSource
```
juanlaube: There's a crash when using `workspaces` related to accessing Realm from a different thread. It's easy to solve by adding the `.Any()` usage in this expression.

Toggl.Foundation.MvvmCross/ViewModels/NoWorkspaceViewModel.cs
```diff
@@ -53,6 +55,8 @@ public async Task CreateWorkspaceWithDefaultName()
         {
             isLoading.OnNext(true);
 
+            dataSource.CreateNewSyncManager();
+
```
juanlaube: You should kick a full sync after creating the workspace.

Toggl.Foundation/DataSources/TogglDataSource.cs
```diff
@@ -90,12 +93,18 @@ public sealed class TogglDataSource : ITogglDataSource
...
+        public ISyncManager SyncManager { get; private set; }
 
         public IReportsProvider ReportsProvider { get; }
 
         public IFeedbackApi FeedbackApi { get; }
 
+        public void CreateNewSyncManager()
+        {
+            SyncManager = createSyncManager(this);
+            errorHandlingDisposable = SyncManager.ProgressObservable.Subscribe(onSyncError);
```
arturs105: Don't you need to dispose the previous disposable?

Toggl.Foundation.MvvmCross/ViewModels/NoWorkspaceViewModel.cs
sofokli1: @juanlaube isn't it used in line 48?

Toggl.Foundation/DataSources/TogglDataSource.cs
sofokli1: @arturs105 take a look again when possible

sofokli1:
@arturs105 and @juanlaube can i get some love here?

Toggl.Foundation.MvvmCross/ViewModels/NoWorkspaceViewModel.cs
```diff
@@ -57,6 +57,10 @@ public async Task CreateWorkspaceWithDefaultName()
 
             dataSource.CreateNewSyncManager();
 
+            await dataSource
+                .SyncManager
+                .ForceFullSync();
+
```
juanlaube: the ForceFullSync should be fired after creating the workspace.

Also, I've found another problem! 😅 But rejoice, because I may have the solution.
The default workspace id in the user profile still points to an old inaccessible workspace, we need to update that id.

Something like this should work...
```
var workspace = await dataSource.Workspaces.GetAll().FirstAsync().First();
user.DefaultWorkspaceId = workspace.Id;
await dataSource.User.Update(user);
```

simonrozsival: Well that might be the same workspace as the default one. Also I don't think we should change the user's preferences. Maybe try using the `GetDefaultWorkspaceInteractor` + maybe add some validation into the interactor if we should filter inaccessible workspaces there.

sofokli1: @juanlaube Line 70 is setting the default workspace for the user to the newly created.
@simonrozsival This is the wright place to set the default workspace (line 70) since the whole point of this code is for cases where the user does not have any workspace anywhere.

simonrozsival: @sofokli1 you are right, I didn't get it the first time 👍

Issue #2817, opened 2018-08-03
Duration editing in the creation view looks weird


simonrozsival:
## 🐛 Bug Report
There is something with the constraints of the duration input. When you focus the duration input, it moves to the top left a bit. See the screenshot for details.

Build/Branch: develop, 1.8
Phone OS: iOS
Phone model: iPhone X, iPhone 7

## Expected behavior
The input should be aligned with the elapsed time label.

## To reproduce
Go to the app and start a new time entry, tap on the elapsed time label to focus the underlying input.

![duration_glitch](https://user-images.githubusercontent.com/374616/43640305-bc97c306-971f-11e8-89b5-142840eadd50.png)

dshr:
I’ve encountered this while doing #2305. I’ve fixed it too, while going through all the constraints in that view.

Issue #2305, opened 2018-05-28
X in onboarding hint in start view animates from left to right when the start view opens


amulware:
Either the whole tooltip should grow in size as well, or the X should not move.

In general, several other elements animate from left to right as well (confirm button, time at the top).

I like the view being a bit more dynamic looking, and not totally static, but we should maybe have this looked over by @pe0ny to decide properly how it should all behave.

pe0ny:
Yes, the whole tooltip should grow in size together with its content.

dshr:
Looks like the "dynamic look" is not intended — `EditTimeEntryViewController` does not exhibit the same behaviour, and I couldn't find any explicit code that does the resizing. So far it seems to be some kind of Autolayout side-effect, which I'll be looking into further.

dshr:
I'm completely at loss as to what's causing the strange constraint behaviour on this screen, at any point before and after the show transition everything looks fine. Trying to capture the UI state midway throws an error 😭. I'm leaving this for someone with better AutoLayout chops to solve.

@khoi I'm removing the `pending-review` label, since this issue is not being reviewed right now.

amulware:
Closing this for now, since we might change a lot about our onboarding anyways and it's a minor glitch.

PR #2823, opened 2018-08-03
Fix duration input in the creation view

dshr:
## What's this?
This PR fixes the constraints of the duration input and duration label so they look right when editing the duration/opening the view.

### Relationships
Initially started as a solution to #2305, but in the process became a fix to #2817.

## Why do we want this?
Things looked weird. Like proper Annihilation (2018).

## How is it done?
Now there is no height constraint to the input, instead it's just vertically centred. Also, the duration label now has a trailing constraint that matches the inputs constraint, which fixes the glitchy animation when opening the view.

### Why not another way?
Can't think of a different way tbh.

### Side effects
None that I've noticed.

## Review hints
simulator screen shot - iphone x - 2018-08-03 at 19 36 13

## :squid: Permissions
Go for it!


simonrozsival:
There is too much empty space after the input - could we change it so it is aligned to the right or make the background view narrower?

trailing-margin

dshr:
True. At the moment the background is stretched bc the label is longer than the input content. Will add some constraint disabling for bot the label and the input, so we never have weird effects when one is larger than the other.

dshr:
Done

PR #2826, opened 2018-08-06
Refactor schmrfactor

kornelijepetak:
## What's this?

This is an attempt to reduce the number of build warnings to 0.

We should really try to avoid those as those are there for a reason.
The only ones I've left untouched are `[Obsolete]` ones that require some action (mostly refactoring into a new way of doing things).

The rest I was able to fix with various modifications that does not change the behaviour of the code.

## What this is not?

This does not handle project/dependencies warnings. These are more tricky and possibly opening a can of worms. However, we should fix the references as well as more and more libraries get implemented for .net standard version we are using.

## Why do we want this?

Warnings exist for a reason and ignoring them can cause problem in the future.

## How is it done?

Well, mostly by doing the obvious.

Also, I have added a streamlined way to deal with hash codes (`GetHashCode`), please take a look at it and comment whether it's useful.

### Side effects

This pertains mostly to @simonrozsival... Some of the storage/sync parts were the `hides inherited member`. In those cases I have added `new` keyword in order to prevent changing the behaviour. However, I believe that in certain cases (like `BaseObjectStorage`) the `virtual/override` behaviour was intended, so please do comment when you see those.

## Review hints

Well, take it commit by commit, it will be much easier.
I have also tried to do things in commits for another reason and that is commit reversal in case some parts other devs don't agree with.

## :squid: Permissions

Since I am on vacations, everyone can comment, push and merge.


simonrozsival:
I think we should avoid using `new` whenever possible. I will try to look into why there is such a problem and probably suggest making the method of the base class virtual. Having it not virtual will make it possibly buggy.

Toggl.Multivac/HashCode.cs
```diff
@@ -0,0 +1,35 @@
...
+
+        public static int From(params object[] items)
+        {
+            unchecked
+            {
+                int hash = 100279;
+
+                for (int i = 0; i < items.Length; i++)
+                {
+                    hash ^= items.GetHashCode() * primeNumbers[i % primeNumbers.Length];
```
willsb: Shouldn't this be `items[i].GetHashCode()`?

Toggl.Foundation.MvvmCross/ViewModels/SuggestionsViewModel.cs
```diff
@@ -71,10 +71,10 @@ private void fetchSuggestions(bool databaseIsEmpty)
                 .Providers
                 .Select(provider => provider.GetSuggestions())
                 .Aggregate(Observable.Merge)
-                .Subscribe(addSuggestions);
+                .Subscribe(addSuggestion);
         }
 
-        private void addSuggestions(Suggestion suggestions)
+        private void addSuggestion(Suggestion suggestions)
```
willsb: The parameter name is still in the plural

Toggl.Foundation.MvvmCross/ViewModels/EditTimeEntryViewModel.cs
```diff
@@ -295,11 +295,12 @@ private async Task delete()
                 await interactorFactory.DeleteTimeEntry(Id).Execute();
 
                 analyticsService.DeleteTimeEntry.Track();
-                dataSource.SyncManager.PushSync();
+                var pushSync = dataSource.SyncManager.PushSync();
```
juanlaube: Is the assignment really needed?

Toggl.Daneel/Views/CountrySelection/CountryViewCell.cs
```diff
@@ -19,7 +19,7 @@ static CountryViewCell()
             Nib = UINib.FromName(nameof(CountryViewCell), NSBundle.MainBundle);
         }
 
-        protected CountryViewCell(IntPtr handle) : base(handle)
+        private CountryViewCell(IntPtr handle) : base(handle)
```
willsb: Make these two public instead for consistency with the rest

Toggl.Foundation.Tests/MvvmCross/ViewModels/CalendarPageViewModelTests.cs
```diff
@@ -44,6 +44,7 @@ public sealed class TheDaysProperty
                 viewModel = new CalendarPageViewModel(calendarMonth, beginningOfWeek, today ?? DateTimeOffset.Now);
             }
 
+#pragma warning disable xUnit1026
```
willsb: Why disable it here?

Toggl.Foundation.MvvmCross/ViewModels/TimeEntriesLogViewModel.cs
```diff
@@ -226,7 +226,7 @@ private async Task delete(TimeEntryViewModel timeEntryViewModel)
             await interactorFactory.DeleteTimeEntry(timeEntryViewModel.Id).Execute();
 
             analyticsService.DeleteTimeEntry.Track();
-            dataSource.SyncManager.PushSync();
+            var pushSync = dataSource.SyncManager.PushSync();
```
willsb: Either use `_` :kornemisis: or introduce a method that returns `void` instead

Toggl.Foundation.MvvmCross/ViewModels/MainViewModel.cs
```diff
@@ -64,7 +64,9 @@ public sealed class MainViewModel : MvxViewModel
         public SuggestionsViewModel SuggestionsViewModel { get; }
         public RatingViewModel RatingViewModel { get; }
         public IOnboardingStorage OnboardingStorage => onboardingStorage;
-        public IMvxNavigationService NavigationService => navigationService;
+
+        public new IMvxNavigationService NavigationService => navigationService;
```
juanlaube: TIL about the new modifier and I don't like it 😞

Toggl.Foundation.MvvmCross/ViewModels/Settings/SettingsViewModel.cs
```diff
@@ -293,7 +293,7 @@ public async Task SelectBeginningOfWeek()
                 return;
 
             await dataSource.User.Update(new EditUserDTO { BeginningOfWeek = newBeginningOfWeek });
-            dataSource.SyncManager.PushSync();
+            var pushSync = dataSource.SyncManager.PushSync();
```
juanlaube: Same thing, is the assignment needed? 🤔

Toggl.Foundation.Tests/MvvmCross/ViewModels/CalendarPageViewModelTests.cs
kornelijepetak: Because the two `Theories` here do not use one of the parameters, issuing a warning. Seemed better to disable the warning than to

1. Create another type of `CalendarPageTestData` with less parameters
2. Use the variable just for the sake of removing the warning even though the test does not care about it

But I've found a rather simple way to do the 1. And it doesn't seem to be such a bad idea. Still more verbose than the pragma warning disable, but maybe more reasonable.

Toggl.Foundation.MvvmCross/ViewModels/EditTimeEntryViewModel.cs
kornelijepetak: The assignment is Microsoft's recommended way of removing the warning when an awaitable thing (`Task`/`IObservable`) is not awaited . However, as @willsb pointed out, it may make more sense to name it `_` in this case to denote that it is not used.

Toggl.Foundation.MvvmCross/ViewModels/MainViewModel.cs
kornelijepetak: Yes. I don't like it either in this case. This `new` modifier makes sense for some uses, but this is not the one. However, I have chosen to put it here in order not to change any behaviour. Without the `new` keyword, it will behave just as with it. That is why all instance of `new` modifier should be checked (preferably by someone not on the vacations 😄) and checked whether virtual/override was actually needed instead of hiding.

For this PR I would rather not get into that.

For this specific instance, I am not sure MvxViewModel's default navigation service is the same that we are injecting. If it is, then this whole property is not needed. @willsb ?

Toggl.Multivac/HashCode.cs
kornelijepetak: Indeed. Fixed.

Toggl.Foundation.MvvmCross/ViewModels/SuggestionsViewModel.cs
kornelijepetak: Fixed.

Toggl.Foundation.MvvmCross/ViewModels/EditTimeEntryViewModel.cs
amulware: One can `_ = expression` without the var keyword as well.

amulware:
I noticed that new way of generating hash codes, and I'm not a big fan of the allocation and boxing that happens in it. Calculating hash codes should be as cheap as possible. If we think we need to unify this, I'd rather have us have a set of methods with generic parameters.

Each of those methods could be as simple as doing only one step of an algorithm and calling down to lower parameter versions.

```
int From<T1, T2, T3>(T1 v1, T2, v2, T3, v3)
{
    return v1.GetHashCode() + 397 * From(v2, v3);
}
```

Toggl.Foundation.MvvmCross/ViewModels/MainViewModel.cs
simonrozsival: Let's keep that for a different refactoring as it would break too many things at this moment.

simonrozsival:
Please have a look at the last two commits (ignore the develop merge), @juanlaube and @willsb.

Toggl.Multivac/HashCode.cs
```diff
@@ -1,34 +1,28 @@
-﻿using System;
-namespace Toggl.Multivac
+﻿namespace Toggl.Multivac
 {
     public static class HashCode
     {
-        private static int[] primeNumbers = new[]
+        public static int From<T1, T2, T3, T4>(T1 itemA, T2 itemB, T3 itemC, T4 itemD)
```
sofokli1: wouldn't it make sense to do this recursively. with one func that accepts an array as parameter and internally call itself with the array-1 element and have an if statment if count is 1 to return?

sofokli1:
Not sure why my comment is hidden ☝️

amulware:
You made the comment while looking at a partial diff instead of the whole, not sure why it hides them then either.

Toggl.Multivac/HashCode.cs
amulware: That would reintroduce both boxing and make performance even worse, since it would have to copy the entire array every call down.

amulware:
Don't wait for my review. I'm happy with the new hash code code.

PR #2832, opened 2018-08-06
Fix crash on Remote Config fetching

### LINKED ISSUE #2828: RemoteConfigFetchFailedException
## 💥 Crash Report
Looks that we don't handle failure to fetch remote config from Firebase correctly. Obviously, that failure should not crash our app under any circumstances. It should simply use the previous/default values and try again later (I'm assuming this is implemented by the Firebase SDK).

https://github.com/toggl/mobileapp/blob/f2c2f0271b6da21cef52f95062902a75a38b8ef7/Toggl.Daneel/Services/RemoteConfigService.cs#L13-L32

### 👥 Impact
- this has happened 558 times to 84 users in 1.8 so far
- that's over 90% of all crashes in 1.8
- all sorts of devices

### 🔗 Context
This was introduced with the recent ratings dialog.

### 🛰 Link
https://appcenter.ms/orgs/Toggl/apps/Toggl-iOS/crashes/groups/09c114c97fd0201409c7fde9cb0d116711aa2782?appVersion=1.8

`Toggl.Foundation.Exceptions.RemoteConfigFetchFailedException: The operation couldn’t be completed. (com.google.remoteconfig.ErrorDomain error 8003.)`
```cs
Select`2+Selector+_[TSource,TResult].OnError (System.Exception error)
SelectMany`2+ObservableSelector+_[TSource,TResult].OnError (System.Exception error)
AutoDetachObserver`1[T].OnErrorCore (System.Exception exception)
ObserverBase`1[T].OnError (System.Exception error)
RemoteConfigService+<>c__DisplayClass1_0.<get_RatingViewConfiguration>b__1 (Firebase.RemoteConfig.RemoteConfigFetchStatus status, Foundation.NSError error)
Trampolines+SDRemoteConfigFetchCompletionHandler.Invoke (System.IntPtr block, System.nint status, System.IntPtr error)
(wrapper native-to-managed) ObjCRuntime.Trampolines+SDRemoteConfigFetchCompletionHandler.Invoke(intptr,System.nint,intptr)

```

dshr:
## What's this?

A WIP fix to RemoteConfigFetchFailedException problem we're having. I'm unsure of some of the choices, hence the WIP tag, so a PR would be a nice place to have a discussion.
### Relationships

Will close #2828 once done

## Why do we want this?

To not get crashes
## How is it done?
By adding a .plist file with some default remote config values we can solve the case when the initial load fails and there is no config to work with. 

I've also refactored the way we handle the load error — now the app will use the cached config if the fetch fails.

### Why not another way?

Another option would be not trying to show anything if the fetch fails. The problem I see with this approach is that we'd lose a lot of potential reviews this way.
### Side effects

None that I can think of.
## Review hints

I'm not sure about `RemoteConfigService.cs:13-17` and `Setup.cs:90-91`. Let me know if there is a better way to do this.
## :squid: Permissions

Fire away 👍🏻


Toggl.Daneel/RemoteConfigDefaults.plist
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+    <key>day_count</key>
+    <integer>5</integer>
+    <key>criterion</key>
+    <string>stop</string>
```
juanlaube: This should be something different from `start|stop|continue` (maybe `none` is a good value). From today's discussion, we don't want to show this to every user.

dshr: Will do! I'm thinking that it might be a good idea to formalise some of the other rules as well, like the interval of inactivity after which we hide the review prompt, etc.

Toggl.Daneel/Services/RemoteConfigService.cs
```diff
@@ -10,17 +10,21 @@ namespace Toggl.Daneel.Services
...
             => Observable.Create<RatingViewConfiguration>( observer =>
             {
                 var remoteConfig = RemoteConfig.SharedInstance;
                 remoteConfig.Fetch((status, error) =>
                 {
-                    if (error != null)
-                        observer.OnError(
-                            new RemoteConfigFetchFailedException(error.ToString()));
+                    if (error == null)
+                        remoteConfig.ActivateFetched();
```
sofokli1: I do not get why you` activate` the config on firebase if there is no error but still push those values to `observer.OnNext` regardless of error. I say always activate since we use the local default values.

dshr: `activateFetched` sets the newly fetched config as the current one. If we're using the defaults, then there is no need to activate.

Yeah, completely blanked out about the `hotfix`. I'll try and rebase these in a sec.

sofokli1: - Isn't the default local config a new config? (because there was no configService in 1.7.2)
- Isn't the activateFetched a way to say that we accept the config and that we are using it now?

dshr: We're not fetching the default, but loading it from a .plist on app launch (`Setup.cs:90-91`), so it's active by default.

It does not look like that to me: https://firebase.google.com/docs/reference/ios/firebaseremoteconfig/api/reference/Classes/FIRRemoteConfig#/c:objc(cs)FIRRemoteConfig(im)activateFetched

sofokli1: ok. fix the branch and you are good to go

amulware:
I don't have time to check this in detail, but if you're all good, you have my permissions to squash into the hotfix branch.

dshr:
Cherry-picked the commits onto a new branch. I will create a new PR for it, while closing this one.

PR #2835, opened 2018-08-06
Add CalendarViewController

### LINKED ISSUE #2796: Add the CalendarViewController
Depends on #2794

Simple issue. Add the classic Mvx boilerplate. Paint the body of the view some bright color (I personally like #FF00FF, but YMMV) so we know that the view is there.

arturs105:
## What's this?
This adds just the view controller boilerplate.

### Relationships

Closes #2796 

## :squid: Permissions

Whenever.


Toggl.Foundation.MvvmCross/ViewModels/MainViewModel.cs
```diff
@@ -347,7 +347,7 @@ private void onMidnight(DateTimeOffset midnight)
         }
 
         private Task openSettings()
-            => navigate<SettingsViewModel>();
+            => navigate<CalendarViewModel>();
```
juanlaube: 👀 is this from your tests?

juanlaube: Or... if we want to navigate quickly to the calendar view while it is WIP, can we have another button for that instead or replacing the settings? 😄

arturs105: 🤦‍♂️ Yes

PR #2836, opened 2018-08-06
🚧 Fix relative date (today, yesterday) not shown in TE header

### LINKED ISSUE #2834: Main TE list section headers format
## 🐛 Bug Report

The section headers in the main screen time entry table view don't show up as title strings like "Today" and "Yesterday".
## Context

Build/Branch: develop
Phone OS: iOS 11
Phone model: iPhone any

## Expected behavior

For the current day and the day before the labels should be "Today" and "Yesterday" respectively.
## To reproduce

1. run the app
2. create a new time entry
3. stop it 

## Suggestions

Looks like `DateToTitleStringValueConverter.cs` has been disconnected from the whole process, which now uses a new converter in `DateToTitleString.cs`

khoi:
## What's this?
The fix for the main TE table view list header title doesn't show relative date 'Today', 'Yesterday'

### Relationships
Closes #2834

## How is it done?
I'm moving the `DateToTitleString` from Daneel to Foundation so it can be reused by Giskard. According to @willsb, Giskard is still using `DateToTitleStringValueConverter` using reflection :magic: so that class can't be removed yet.

### Side effects
Not that I know of

## Review hints
Manually create time entries yesterday, today

## :squid: Permissions
Anyone


khoi:
@willsb can you help to review this to make sure I didn't break any reflection code anywhere

khoi:
rebasing, sec

Issue #2620, opened 2018-07-06
Test rate-us view


amulware:
We need to verify that:

- [x] the remote config works as expected
- [x] the rate-us view pops up based on the remote config as specified
- [x] all four paths have the expected outcome
- [ ] it behaves correctly if the user ignores it and leaves the app for short/longer durations
- [x] the final state is store correctly, and the view is never shown again
- [ ] metrics events are being sent off

I'd also like someone to figure out how @pe0ny and @ApocalypticFish to test the UX side of this without having to dig into the code.

sofokli1:
- For the UX part I would propose a branch fixed in a way that always shows the rating view.
- For the rest I would propose to set up everything the same way as in app store app id but for develop and adhoc. Everything will be separate by app id (analytics crashes etc.) so no noise from the development app id in the app store one. This way we can play with a/b testing in develop and adhoc (bitrise) and in the end just copy the experiment in the app store app id. (same goes for testing new analytics events and what not)

amulware:
Do we then need a separate configuration in firebase/app center? I'm pretty sure those are tied to app id.

sofokli1:
I don't think so. I think it is enough to just remove the` #if` in the appDelegate

amulware:
Let's do it then.

sofokli1:
@amulware For the firebase we need at least the googl-app-id and some other keys for appCenter and facebook that it is usually only set in bitrise. I would suggest to have 3 scripts. 1 shared only internally to set the env variable in our local machines and 2 scripts (or one with parameters) to set and unset the values from project (set values before build and unset after build automatically). So the varaibels will be set and unset without affecting git. Does this make sense? Should i move forward with this?

amulware:
Why not just build a test build on bitrise to replace secrets? And I'm not sure of the exact state, but what about Arturs' test branch? (Ricardo mentioned it earlier)

sofokli1:
I would start only with the google/firebase thing and then see about appCenter and facebook in a different PR based on needs

sofokli1:
My main issues with the rating view and the firebase remote config (a/b testing) are:
- When there is an error fetching the configuration, It is not handled and it will crash. (at least on the branch for testing)
- If the view is presented and the app restarts the view will not show up again even if the users never interacted with it (not sure if this is by design).

- It is hard to test if the metrics events are sent to firebase since it takes around 24 h for them to appear.


[...]

PR #2846, opened 2018-08-06
🚧 Fix date checking for experiment based on first install instead of last update

### LINKED ISSUE #2827: Do not show ratings dialog by default
## 🐛 Bug Report
The ratings dialog is set to show on one of the triggers by default.

## Expected behavior
It should never be shown to user unless we activate an appropriate experiment.

That's the only way we can gather meaningful data on small samples of users

## To reproduce
- Have 1.8 installed
- Wait five days
- Start/stop a time entry

## Suggestions
You could set the 5-day limit lower for testing purposes.

sofokli1:
## What's this?

Is a fix for the day count calculations of the rating view experiment.

### Relationships

Closes #2827

## Why do we want this?

So that calculations do not start from install date.

## How is it done?

It keeps track of the first date the app was run after being updated and uses it to calculate days in the rating view experiment.

### Why not another way?

-

### Side effects

-

## Review hints

Use tutorial in from [this issue](https://github.com/toggl/mobileapp/issues/2620) to test it.

## :squid: Permissions

Whenever

### DIFF
--- a/Toggl.Daneel/Startup/AppDelegate.cs
+++ b/Toggl.Daneel/Startup/AppDelegate.cs
@@ -14,6 +14,9 @@ using Toggl.Foundation.MvvmCross.ViewModels;
 using Toggl.Foundation.Services;
 using Toggl.Foundation.Shortcuts;
 using UIKit;
+using Toggl.PrimeRadiant.Settings;
+using Toggl.Foundation;
+
 
 #if USE_ANALYTICS
 using System.Linq;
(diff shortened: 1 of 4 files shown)


amulware:
I don't understand why we want this. Please elaborate.

sofokli1:
At the moment we use the install date to calculate the days count since now to evaluate if we are going to show the rating view or not. For someone that has the app installed for 100 days and we start an experiment for 10, 30, 50 days variant they pass it immediately thus making the day count useless.

amulware:
I don't see how this makes any difference. Wether they see it after 100 or after 110 days of using the app shouldn't make a difference. At first we won't do much with the time parameter. That will really only become meaningful once we start running the experiment for all new users.

I don't think we need this change.

sofokli1:
sure, but when starting an experiment we do not pick and choose who is part of it and who isn't (new or old users) we just choose a percentage of the user base.

amulware:
Yes, we have to be especially careful with that and make sure to evaluate results accordingly.

PR #2847, opened 2018-08-06
Rename the reports calendar

### LINKED ISSUE #2830: Rename one of the calendars
With the new calendar that we're working on, we now have two calendars:
* In reports
* The new log/calendar.

We should rename one of those, so it's easier to distinguish which classes are used for which calendar. We could prefix all the reports calendar stuff with `ReportsCalendar`. That includes quick select shortcuts, view models, view controllers, tests and anything else that has `Calendar` in it's name and is referring to the reports calendar.

dshr:
## What's this?
Renames every calendar-related thing on the reports page from `Calendar...` to `ReportsCalendar...`

### Relationships
Closes #2830

## Why do we want this?
To separate the calendar on the reports page from the big calendar integration.

## How is it done?
Just renaming things, lol.

### Why not another way?
Not really.

### Side effects
Couldn't deploy to Giskard, but it built just fine, so it all should be good 😅 

## Review hints
Build and run both on Daneel and Giskard and go through the reports page. 

## :squid: Permissions
All yours!


amulware:
So I'm wondering: Why aren't we using namespaces for this?

arturs105:
I thought about using namespaces before creating the issue, but i think renaming one of the calendars would greatly improve fuzzy search. Also: i don't like that 2 different things have the same name.

PR #2850, opened 2018-08-07
Make [Making Drivers safer] safer

daividssilverio:
## What's this?
This a refactory to fix the tests on #2744.

### Relationships
Parent: #2744

Closes Nothing

## Why do we want this?
It fixes all the breaking tests that makes the asDriver extension safer.
It also gives the possibility to have more control on where observables are running on VMs using `SubscribeOn` and `ObserveOn` as necessary.

## How is it done?
It changes how the main scheduler is made available for ViewModels and other components, making it injectable.

### Why not another way?
This way feels natural to me and it's "easy" to test. 
I'd rather have the
```
public static IObservable<T> AsDriver<T>(this IObservable<T> observable, T onErrorJustReturn, ISchedulerProvider schedulerProvider)
```
without the schedulerProvider parameter, but this could be changed later.

### Side effects
Tests that somehow involve drivers have to call `TestScheduler.Start()` before checking observers messages.

## Review hints
Please test functionally on iOS and Android as well to see if everything is still ok.

## :squid: Permissions
Only @willsb. This merges into an open PR.


Toggl.Daneel/Startup/Setup.cs
```diff
@@ -88,8 +88,9 @@ protected override void InitializeApp(IMvxPluginManager pluginManager, IMvxAppli
             var userAgent = new UserAgent(clientName, version);
             var keyValueStorage = new UserDefaultsStorage();
             var settingsStorage = new SettingsStorage(Version.Parse(version), keyValueStorage);
+            var schedulerProvider = new IOSSchedulerProvider();
 
-            ReactiveExtensions.MainThreadScheduler = new NSRunloopScheduler();
+            ReactiveExtensions.SchedulerProvider = schedulerProvider;
```
willsb: Stop using this static guy and inject it everywhere instead.

PR #2853, opened 2018-08-07
Remote config fetching

### LINKED ISSUE #2828: RemoteConfigFetchFailedException

dshr:
## What's this?
A cherry-pick of the commits for #2832, which is a fix to the RemoteConfigFetchFailedException we're having.

### Relationships
Closes #2828

## Why do we want this?
To not get crashes when remote config fetching does not work and to have all the changes against `hotfix-1.8.1`.
## How is it done?
By adding a .plist file with some default remote config values we can solve the case when the initial load fails and there is no config to work with.

I've also refactored the way we handle the load error — now the app will use the cached config if the fetch fails.

### Why not another way?
This is the simplest so far.

### Side effects
None that I can think of.

## Review hints
On 1.8 try replicating the crash using the steps described by @sofokli1 in #2828
## :squid: Permissions
Go for it.


Toggl.Daneel/Startup/Setup.cs
```diff
@@ -87,6 +87,8 @@ protected override void InitializeApp(IMvxPluginManager pluginManager, IMvxAppli
             var userAgent = new UserAgent(clientName, version);
             var keyValueStorage = new UserDefaultsStorage();
             var settingsStorage = new SettingsStorage(Version.Parse(version), keyValueStorage);
+            var remoteConfigService = new RemoteConfigService();
+            remoteConfigService.setupDefaults("RemoteConfigDefaults");
```
simonrozsival: This file name could go into a `const string` property and the code of this PR would be perfect 👍

Toggl.Daneel/RemoteConfigDefaults.plist
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+    <key>day_count</key>
+    <integer>5</integer>
```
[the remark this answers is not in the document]
amulware: I think it's ok to block it for new users for a few days. We don't want to accidentally show it to them on their first day, when they don't have a good opinion yet... Then again, as long as the other parameter is set to `none` it doesn't matter since we can always override both in firebase.

So... I don't think it matters what the value is here.

Toggl.Daneel/Startup/Setup.cs
willsb: `setupDefaults ` should begin with a capital as well 👀

PR #2854, opened 2018-08-07
Add a way to navigate to CalendarViewModel

### LINKED ISSUE #2795: Introduce a way to navigate to the CalendarViewModel from the MainViewModel
Ref #2754 for architectural details
Depends on #2793 and #2794

Either way, it should be shown/hidden as needed after checking the results of the FeatureService. Navigation to the CalendarViewModel should only happen it the service allows it.

This is not an UI issue. This only concerns the ViewModel layer. Tests should be added as applicable.

arturs105:
## What's this?
This adds an action in the `MainViewModel` (and tests) to open calendar.

### Relationships

Closes #2795 

## :squid: Permissions
Whenever.


Toggl.Foundation.MvvmCross/ViewModels/MainViewModel.cs
```diff
@@ -358,6 +360,11 @@ private async Task openReports()
         private Task openSyncFailures()
             => navigate<SyncFailuresViewModel>();
 
+        private IObservable<Unit> openCalendarAction()
+            => Observable
+                .FromAsync(async () => await navigationService.Navigate<CalendarViewModel>())
+                .Select(_ => Unit.Default);
```
willsb: Change this into `SelectUnit()`

PR #2856, opened 2018-08-07
:warning: Fix signup TOS acceptance test after API update

### LINKED ISSUE #2800: Update TheSignUpMethod.SucceedsEvenIfUserDidNotAcceptTermsAndConditions
Backend has made TOS acceptance on signup mandatory.

This test was checking the previous behaviour and is now out of date. It should be changed to test that signup without accepting TOS will fail.

amulware:
## What's this?
Backend has made accepting TOS on signup mandatory. I.e. we always have to send the appropriate flag and it has to be set to true, or signup will fail.

This updates our test verifying the previous behaviour to now verify the new one.

### Relationships
Closes #2800
Relevant backend issue: https://github.com/toggl/backend/issues/6505

## Why do we want this?
Because this is the only test failing in the majority of integration test runs, and having to check it manually every time is silly.

## How is it done?
See the changed code. Nothing surprising there.

### Why not another way?
It's as straightforward as it goes. Note that we're not testing what happens when the TOS field isn't sent to the API at all, but since Ultrawave doesn't support that, I don't think we care.

### Side effects
Integration tests will mostly ✅ again!

## Review hints
Check the integration tests here: https://app.bitrise.io/build/e65b2169a6593abc

## :squid: Permissions
You, you and you. Everyone gets a 🦑!


Toggl.Ultrawave.Tests.Integration/UserApiTests.cs
```diff
@@ -275,20 +275,19 @@ public async Task CreatesADefaultWorkspaceWithCorrectName(string emailPrefix, st
             }
 
             [Fact, LogTestInfo]
-            public async Task SucceedsEvenIfUserDidNotAcceptTermsAndConditions()
+            public async Task FailsIfUserDidNotAcceptTermsAndConditions()
```
simonrozsival: Oh, this should be just a `void` now.

amulware: I don't have msbuild on my new machine yet so I'm writing code in Sublime. Clearly I'm too reliant on IDEs :D

amulware:
Integration tests passed.

PR #2860, opened 2018-08-07
Add a way to store enabled calendars

### LINKED ISSUE #2843: Add a way to persist the list of calendars in the storage
We don't need to change Realm for this. We just need a way to read/persist the list of calendars the user wants to use. 

If we do it in the [`SettingsStorage`](https://github.com/toggl/mobileapp/blob/develop/Toggl.PrimeRadiant/Settings/SettingsStorage.cs) we won't even need to have a droid issue because it'll leverage on our existing storage mechanisms! 

Regardless of using the existing base class, I'd add another interface to keep separation of concerns in check

arturs105:
## What's this?
This adds a way to store, which user calendars should be used in the new calendar view.

### Relationships

Closes #2843

## Why do we want this?
We need to keep track of which user calendars to use in the new calendar view.

## How is it done?
We store just the ids as a single string using `;` as a separator. I checked and the calendar ids look something like this: `DSG84-GBFI2-H24T2Q4T-ET34G`, so there should be no problems with the separator character.

### Why not another way?
I don't think we need to store any other info, because we can access anything else that's needed through the calendar APIs of each platform.

## :squid: Permissions
Whenever.


Toggl.PrimeRadiant/Settings/SettingsStorage.cs
```diff
@@ -315,6 +320,38 @@ void IUserPreferences.Reset()
...
+
+        public void EnableCalendars(params string[] ids)
+        {
+            if (ids == null)
+            {
+                keyValueStorage.Remove(enabledCalendarsKey);
+                return;
+            }
+
+            if (!ids.Any())
```
willsb: `ids.None()`

Toggl.PrimeRadiant/Settings/SettingsStorage.cs
```diff
@@ -315,6 +326,40 @@ void IUserPreferences.Reset()
...
+            var aggregatedIds = keyValueStorage.GetString(enabledCalendarsKey);
+            if (string.IsNullOrEmpty(aggregatedIds))
+                return new List<string>();
+
+            return aggregatedIds
+                .Split(calendarIdSeparator)
+                .ToList();
+        }
+
+        public void EnableCalendars(params string[] ids)
```
juanlaube: `EnableCalendars` hints there's another method to disable calendars. I think a better name would be `SetEnabledCalendars`.

PR #2863, opened 2018-08-07
Add a button to navigate to the calendar view controller

### LINKED ISSUE #2797: Add the means of navigating to the CalendarViewController
Depends on #2795 and #2796

If #2771 is merged, please add a new tab item. Otherwise, add the calendar icon to the navigation bar and open an issue to move it to the Tab bar later.

Either way, the tab should respect whether the icon should be visible or not (this will already be properly implemented in the VM) and only allow navigation when the FeatureService says it's possible.

juanlaube:
## What's this?

Adds a new button in the navigation bar to navigate to the calendar view model.

### Relationships

Closes #2797

## :squid: Permissions

Whenever


Toggl.Foundation.MvvmCross/ViewModels/MainViewModel.cs
```diff
@@ -355,6 +357,9 @@ private async Task openReports()
             await navigate<ReportsViewModel, long>(workspace.Id);
         }
 
+        private Task openCalendar()
+            => navigate<CalendarViewModel>();
+
```
arturs105: Since #2854 was merged, you can use the new rx action from that PR.

PR #2864, opened 2018-08-07
Make FeedbackService optional

simonrozsival:
## What's this?
The iOS app on the `develop` branch is broken right now - the bug was introduced in #2859 where we brought back the `FeedbackService`, but we forgot to fix services' initialization on iOS, where this service is not used.

## Why do we want this?
We want the iOS app to work.

## How is it done?
The `FeedbackService` is now optional so that on iOS we can ignore it and on Android it works well until it will be replaced soon with the custom feedback UI.

### Why not another way?
It would not make any sense to register the service in the setup code for iOS, because it would not be used anywhere.

## :squid: Permissions
Squash whenever.


simonrozsival:
I would like some Android dev to verify that this does not break Android in any way (it shouldn't, but I want to be sure and end this bug ping-pong).

Issue #2833, opened 2018-08-06
Determine cause of android startup crashes


amulware:
## 💥 Crash Report
We did great fixing all very common crashes that were tracked on App Center with 1.2

However, users are still reporting common crashes on startup, and Google Play shows crash rates of between 25% and 30% (users per day experiencing at least one crash).

![android-crashes-comparison](https://user-images.githubusercontent.com/3987892/43717073-45f6cb6e-997e-11e8-99b3-85f1ea99046d.png)

We gotta figure out what is going on and fix it.

### 👥 Impact
- somewhere around 25%-30% of android users
- has been around since first release
- responsible for 90%+ of all our crashes right now

### 🛰 Link
No link. While Google reports it in the total crash numbers, it doesn't actually show any other information anywhere.

### 📝  Exception & Stack Trace
No such info is known yet. Neither Google nor Microsoft have any data for us here.

amulware:
I asked support to collect data from users who have crashes on startup, specifically:
- their device/android version
- whether the crash happens on every start, or every first start that day, or randomly, or...

Maybe that'll allow us to find a pattern.

amulware:
It very much looks like the change in #2865 has solved this. We get a lot more crashes recorded now, close to 500 a day, which seems to close the gap detailed above almost, or maybe in fact, entirely.

There may be more than one thing going on, but #2923 covers 84% of crashes in 1.3, so let's deal with that and see!

PR #2865, opened 2018-08-07
Move appcenter startup from SplashScreen to TogglApplication

daividssilverio:
## What's this?
Pretty straightforward, it moves the initialization of the appcenter service to the application onCreate method.

### Relationships
An effort to make #2833 easier

## Why do we want this?
We still have crashes on startup and it's hard to understand what is going on. Hopefully, this will help catching some of the causes.

## How is it done?
The initialization block is moved from the `SplashScreen` `onCreate` to the `TogglApplication` `onCreate`.
The cake script is updated to make the replacements in the right file.

### Why not another way?
There is no earlier startup point. Maybe using NDK... 🤔

### Side effects
Hopefully, we will get more side effects.
This might affect the initialization just a tad.

## Review hints
Try removing the compiler macros to test it on your device with a debug build.
Otherwise replace all secrets and test it as a release build :)
Nothing should be different. 

## :squid: Permissions
Anyone, as long as you test it.


Toggl.Giskard/Startup/TogglApplication.cs
```diff
@@ -13,5 +13,17 @@ public class TogglApplication : MvxAppCompatApplication<Setup, App<LoginViewMode
         public TogglApplication(IntPtr javaReference, JniHandleOwnership transfer) : base(javaReference, transfer)
         {
         }
+
+        public override void OnCreate()
+        {
+            base.OnCreate();
+#if USE_ANALYTICS
```
willsb: Why create an empty override? I'd rather have the entire method inside the #if...

PR #2866, opened 2018-08-07
Giskard 1.3 - Fluorine

amulware:
_Fluorine is a somewhat sneaky element. It makes space for other elements like [Oxygen](https://github.com/toggl/mobileapp/pull/2752) in stellar nucleosynthesis and could this be fairly blamed for much of the trouble caused by the latter._

```
- Improved app startup time
- Reduced app download size

- Enabled keyboard suggestions in edit view

- Fixed tooltip positioning in start view
- Fixed navigation error after creating project
```


Toggl.Giskard/Toggl.Giskard.csproj
```diff
@@ -46,6 +46,9 @@
     <AndroidSigningStorePass>{KEYSTORE_PASSWORD}</AndroidSigningStorePass>
     <AndroidSigningKeyAlias>{KEYSTORE_ALIAS}</AndroidSigningKeyAlias>
     <AndroidSigningKeyPass>{KEYSTORE_ALIAS_PASSWORD}</AndroidSigningKeyPass>
+    <AndroidCreatePackagePerAbi>true</AndroidCreatePackagePerAbi>
+    <AotAssemblies>true</AotAssemblies>
+    <EnableLLVM>true</EnableLLVM>
```
amulware: We should probably remove these changes to the adhoc config.

willsb: I, OTOH, see the value of testing the same binaries for both AdHoc and release

daividssilverio:
oh... the element name is wrong :0. It was supposed to be Fluorine.
I guess the branch I created made the confusion.

Toggl.Giskard/Toggl.Giskard.csproj
amulware: Except it makes it really annoying downloading the right binary from bitrise, if one wants to test an adhoc build - no?

amulware: Also, bitrise is no good with serving multiple downloads I think, we'll have to see. I'll remove these now so we can merge this branch with minimal changes for the time being. We can always redo this properly down the line.

amulware:
We started releasing this to users. Let's get this PR merged. Please do pay attention that the changes to the project file make sense. The intention is to have multiple abis enabled only for playstore release builds.

PR #2873, opened 2018-08-07
Move compiler macro to hide override when not on release

daividssilverio:
## What's this?
Just an improvement on the TogglApplication to avoid the override on debug builds.

### Relationships
Changes requested here #2865  

Closes Nothing

## Why do we want this?
To avoid the extra override

## How is it done?
The macro is moved to wrap the whole onCreate override

### Why not another way?
To avoid the extra override

### Side effects
None.

## Review hints
This is pretty straight forward.

## :squid: Permissions
Anyone, but talk to @amulware first.


daividssilverio:
We can rebase this after fluorine is merged.

amulware:
Please do so, and after doing so delete the release branch. THX

PR #2883, opened 2018-08-08
Add Interactor to read/fetch a list of all calendars

### LINKED ISSUE #2844: Add Interactor to read/fetch a list of all calendars the user wants to see
Depends on #2838
Depends on #2843

This will need a method in the calendar service that returns a list of `UserCalendar`s. The data will mostly come from the platform specific service, but in order to fill the `Selected` property we'll need to cross-reference the info from the settings storage in order to define whether a calendar is selected or not.

A second method that gets a list of calendars and persists them in the SettingsStorage is also needed. This is so that we don't expose the `ICalendarSettingsStorage` to the ViewModel, but instead use interactors for that.

arturs105:
## What's this?
This adds the necessary interactors to get a list of user calendars and set them as enabled.

### Relationships

Closes #2844

## :squid: Permissions
Whenever.


Toggl.Daneel/Services/CalendarService.cs
```diff
@@ -1,14 +1,65 @@
...
+            Ensure.Argument.IsNotNull(permissionsService, nameof(permissionsService));
+
+            this.userPreferences = userPreferences;
+            this.permissionsService = permissionsService;
+        }
+
         public IObservable<IEnumerable<CalendarItem>> GetEventsForDate(DateTime date)
             => Observable.Return(new List<CalendarItem>());
+
+        public void SetEnabledCalendars(params string[] ids)
```
willsb: Why is this needed?

Toggl.Daneel/Services/CalendarService.cs
```diff
@@ -1,14 +1,65 @@
...
+            => userPreferences.SetEnabledCalendars(ids);
+
+        public IObservable<IEnumerable<UserCalendar>> UserCalendars
```
willsb: I this could be in a base class, since the logic for Android will be the same.But we can always change later

Toggl.Daneel/Services/CalendarService.cs
willsb: Since it's only forwarding, this should be only in the interactor

Toggl.Daneel/Services/CalendarService.cs
```diff
@@ -1,14 +1,65 @@
...
+            => permissionsService
+                .CalendarAuthorizationStatus
+                .Do(throwIfPermissionDenied)
```
willsb: I think something like:

```csharp
Observable.DeferAsync(async () =>
{
    var isAuthorized = await permissionsService.CalendarAuthorizationStatus;
    if (!isAuthorized)
        return Observable.Throw<Type>();
    
    var hash = userPreferences.EnabledCalendarIds().ToHashSet();
    return Observable.Return(selectCalendars(hash));
});
```

Feels more reactive and easier to read/reason about than a `.Do`

Toggl.Foundation/Interactors/Calendar/SetEnabledCalendarsInteractor.cs
```diff
@@ -0,0 +1,28 @@
...
+            Ensure.Argument.IsNotNull(calendarService, nameof(calendarService));
+            Ensure.Argument.IsNotNull(selectedCalendarIds, nameof(selectedCalendarIds));
+
+            this.calendarService = calendarService;
+            this.selectedCalendarIds = selectedCalendarIds;
+        }
+
+        public Unit Execute()
+        {
+            calendarService.SetEnabledCalendars(selectedCalendarIds);
```
willsb: I think the `CalendarService` doesn't need to know about the ids in this case, like I mentioned above.

Toggl.Daneel/Services/CalendarService.cs
arturs105: I think we can do this when we need this stuff in android `CalendarService`.

Toggl.Daneel/Services/CalendarService.cs
```diff
@@ -1,14 +1,70 @@
...
             => Observable.Return(new List<CalendarItem>());
+
+        public IObservable<IEnumerable<UserCalendar>> UserCalendars
+            => Observable.DeferAsync(async (cancellactionToken) =>
+                {
+                    var isAuthorized = await permissionsService.CalendarAuthorizationStatus;
+                    if (!isAuthorized)
+                    {
+                        return Observable.Throw<IEnumerable<UserCalendar>>(
+                            new NotAuthorizedException("Don't have permission to access calendars")
```
willsb: _You_ don't have permission

Toggl.Daneel/Services/CalendarService.cs
```diff
@@ -1,14 +1,70 @@
...
+                        );
+                    }
+
+                    var enabledIds = userPreferences.EnabledCalendarIds().ToHashSet();
+                    return Observable.Return(selectCalendars(enabledIds));
+                });
+
+        private void throwIfPermissionDenied(bool havePermission)
```
willsb: Unused method

willsb: Also it should be `hasPermission`, but it'll be deleted so whatever

Issue #2631, opened 2018-07-09
WS removal: remove synced TEs of lost workspace


amulware:
If we lose access to a workspace, we need to get rid of all the synced data we have in it. Unsynced data will be kept so that the user can regain access and sync it, or move it to a different workspace.

~This state will probably run after the one from #2628.~

Since time entries are not referred to by anything else, we can easily remove those that are synced from the DB.

Other entities have to be checked especially since they are referred to by time entries.

simonrozsival:
We don't need to run this right after the access is lost, we can do this in the `CleanUp` loop instead. We should schedule a clean-up right after we detect losing access to the workspace though.

PR #2886, opened 2018-08-08
Clean up sync entry point

simonrozsival:
## What's this?
This PR adds a new entry point for the syncing machine for the _clean up_. This part of syncing is meant mainly for pruning old data and removing data which were pulled but should no longer be stored on the device (like when the user loses access to a workspace).

### Relationships
Ref #2631

## Why do we want this?
Checking the database for items which are no longer needed might potentially include several complex queries which should not run on every pull or push sync. It is OK to prune the data every day or so. That is why a separate syncing entry point comes into play.

## How is it done?
The sync state queue is modified to include a new type of a sync state. The sync manager, state machine, and the state machine orchestrator are modified so that they support this new entry point. Two states which were part of the pull-sync are now moved to the sync clean-up.

### Side effects
We won't prune old data as often as we did so far, but that should not be a problem at all as we are doing that too often in the current release.

## Review hints
Change your time settings to see if the clean-up sync kicks in right after the date changes ("midnight" observable). Feel free to ask me how I test this myself in the simulator.

## :squid: Permissions
This is still a WIP (more testing is needed) and is not ready for merging.

### DIFF
--- a/Toggl.Foundation.Tests/Sync/SyncManagerTests.cs
+++ b/Toggl.Foundation.Tests/Sync/SyncManagerTests.cs
@@ -373,6 +373,24 @@ namespace Toggl.Foundation.Tests.Sync
             }
         }
 
+        public sealed class TheCleanUpMethod : SyncMethodTests
+        {
+            protected override IObservable<SyncState> CallSyncMethod()
+                => SyncManager.CleanUp();
+
+            [Fact, LogIfTooSlow]
+            public void TellsQueueToStartSyncAfterQueingPush()
+            {
+                CallMethod();
+
+                Received.InOrder(() =>
+                {
+                    Queue.QueueCleanUp();
+                    Queue.Dequeue();
+                });
+            }
+        }
+
         public sealed class TheFreezeMethod : SyncManagerTestBase
         {
             [Fact, LogIfTooSlow]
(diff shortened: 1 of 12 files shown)


simonrozsival:
I am now confident this works well, please review.

Toggl.Foundation.Tests/Sync/SyncManagerTests.cs
```diff
@@ -373,6 +373,24 @@ public void SavesTheTimeOfSuccessfulFullSync(DateTimeOffset now)
             }
         }
 
+        public sealed class TheCleanUpMethod : SyncMethodTests
+        {
+            protected override IObservable<SyncState> CallSyncMethod()
+                => SyncManager.CleanUp();
+
+            [Fact, LogIfTooSlow]
+            public void TellsQueueToStartSyncAfterQueingPush()
```
sofokli1: I don't get this test. I can see that it checks that it queues cleanup and it dequeues it but where do we check that it is queueing push?
Also i think there is a typo in the test name `...Queing...`, I think it should be `...Queuing...`

simonrozsival: This is a bad naming due to ©️ 🍝 . I will change the name to make some sense.

Toggl.Foundation/Sync/SyncStateQueue.cs
```diff
@@ -29,14 +35,21 @@ public SyncState Dequeue()
             if (pushSyncQueued)
                 return push();
 
-            return sleep();
+            if (cleanUp)
+            {
+                cleanUp = false;
```
amulware: I'd prefer these side effects to stay hidden away like they are in the pull/push methods. Having a consistent level of abstraction is important I think, and this method is already getting long.

simonrozsival: OK.

PR #2892, opened 2018-08-09
Add Feedback service back to Daneel

khoi:
## What's this?
The iOS app is still crashing when users navigate to the setting screen since `SettingsViewModel` cannot be constructed without a `FeedbackService`

The bug was first introduced in #2859 and another fix attempted in #2864  but it's still failing.

### Relationships

Revert #2864 

Related to #2859 #2864 #2710 

## Why do we want this?
so the app can run

## How is it done?
Adding back the FeedbackService to `Daneel`'s setup code even though it's not used by the iOS app. 

### Why not another way?
I think this is the easiest way without having to customize the IOC

## Review hints

Run both Android and iOS app, go to settings, feedback, and make sure you can actually send one.

## :squid: Permissions
Anyone


simonrozsival:
I think we should revert #2864 as a whole. That PR does not make any sense anymore. Let's also mark feedback service as `[Obsolete]`.

khoi:
I need someone from @toggl/androids to verify that it doesn't break Giskard 🗡

PR #2899, opened 2018-08-09
Mark workspaces as ghosts

### LINKED ISSUE #2639: WS removal: mark workspace as ghost
The workspace we got removed from should be marked as ghost according to #1337.

simonrozsival:
## What's this?
This PR modifies the sync state which detects workspaces the user can't access anymore as ghosts. That is all.

### Relationships
Closes #2639 

## How is it done?
I had to add a new constructor to the `Toggl.Foundation.Models.Workspace` class which allows to change the workspace to a ghost-workspace.

I also had to fix a bug where we did not copy the state of the `IsGhost` property which caused the information to be lost.

I updated the state we already had to mark the workspaces as ghosts directly instead of passing a list of these workspaces. It will make code simpler and easier to maintain.

### Why not another way?
I could add another separate state which would take a list of workspaces and mark them as ghosts but that would add so much boilerplate code that it would not make any sense to do it that way.

### Side effects
No side effects. The sync state is not used by the state machine yet.

## Review hints
You may try this:
- reverte the `18a8fdb` commit
- create several extra workspaces using the web app
- sync the app
- leave some of the workspaces in the web app
- put a breakpoint in `Toggl.Foundation/Models/Workspace.cs` on line 61
- sync the app (you should hit the breakpoint once for every WS you left
- sync the app again (you should not hit the breakpoint again)

Also check Realm Studio to see that the workspaces have the `IsGhost` property set to true.

## :squid: Permissions
Squash whenever


Toggl.PrimeRadiant.Realm/Models/RealmConstructors.cs
```diff
@@ -397,6 +397,7 @@ public void SetPropertiesFrom(IDatabaseWorkspace entity, Realms.Realm realm)
             Rounding = entity.Rounding;
             RoundingMinutes = entity.RoundingMinutes;
             LogoUrl = entity.LogoUrl;
+            IsGhost = entity.IsGhost;
```
amulware: That's what we get for removing templates, huh? 🙈

Toggl.Foundation/Models/Workspace.cs
```diff
@@ -53,5 +54,17 @@ private Workspace(Builder builder)
...
+        internal Workspace(IThreadSafeWorkspace workspace, bool isGhost)
+            : this(workspace)
+        {
+            IsGhost = isGhost;
+        }
+    }
+
+    public static class ThreadSafeWorkspaceExtensions
+    {
+        public static IThreadSafeWorkspace AsGhost(this IThreadSafeWorkspace workspace)
```
amulware: Thank you for this. Having the constructor public would have upset me greatly.

Toggl.Foundation/Sync/States/Pull/DetectLosingAccessToWorkspacesState.cs
```diff
@@ -41,5 +40,15 @@ private IObservable<IList<IThreadSafeWorkspace>> workspacesWhichWereNotFetched(L
...
+
+        private IObservable<Unit> markAsGhosts(IList<IThreadSafeWorkspace> workspacesToMark)
+            => Observable.Return(workspacesToMark)
+                .SelectMany(CommonFunctions.Identity)
+                .SelectMany(markAsGhost)
+                .ToList()
+                .Select(Unit.Default);
+
+        private IObservable<IThreadSafeWorkspace> markAsGhost(IThreadSafeWorkspace workspaceToMark)
+            => dataSource.Update(workspaceToMark.AsGhost());
```
amulware: So this overrides them with the thing fetched some lines above just with the property set to true... Since we don't have another way of editing workspaces it's probably fine, but otherwise, there would be a race condition here. Not sure if it is worth complicating this significantly to prevent that, however.

Toggl.PrimeRadiant.Realm/Models/RealmConstructors.cs
simonrozsival: Yes 😢

amulware: Let's be extra watchful when mutating models down the line.

Toggl.Foundation/Sync/States/Pull/DetectLosingAccessToWorkspacesState.cs
simonrozsival: No - this changes just the `IsGhost` property. The workspaces which are marked are those which are missing in the list of fetched workspaces.

amulware: It still overrides all properties, so if someone were to edit the name in just the wrong moment (which as mentioned isn't supported) it could go wrong.

Toggl.Foundation.Tests/Sync/States/Pull/DetectLosingAccessToWorkspacesStateTests.cs
```diff
@@ -53,12 +53,33 @@ public async Task ReturnsListOfWorkspacesWhichAreStoredLocallyButAreNotInTheList
...
-                parameter.Should().HaveCount(2);
-                parameter.Should().Contain(ws => ws.Id == 2);
-                parameter.Should().Contain(ws => ws.Id == 3);
+                parameter.Should().Be(fetchObservables);
+            }
+
+            [Fact]
+            public async Task MarksWorkspacesWhichAreStoredLocallyButAreNotInTheListFromTheServerAsGhosts()
+            {
+                prepareDatabase(new[]
```
sofokli1: if `prepareDatabase` is not called at all here, will it return what the previous test set it to return?

simonrozsival: It should not do that - each test should be run in an setup which is not dependent on other tests.

The only exceptions are static classes (tests run in parallel access the same static classes, it cannot be isolated) and property tests (which are run repeatedly with the same setup and have to be reset manually). Neither of these cases apply here.

amulware:
What's this waiting for?

amulware:
Do go on without me.

PR #2908, opened 2018-08-09
Expose the calendar items from the view model

### LINKED ISSUE #2867: Consuming the calendar interactor in the CalendarViewModel
We need to consume the Interactor introduced in #2849 (the one that fetches the `CalendarItem`s from the database/calendar) and expose its results as an observable in the CalendarViewModel. Pretty simple. This doesn't need to consume anything in the ViewController, just the Foundation changes and their appropriate tests.

The other interactor is meant to be consumed in the Onboarding issue, so this covers everything we need for now.

juanlaube:
## What's this?

Exposes the calendar items from the view model to be used in the view controller.

### Relationships

Closes #2867

## How is it done?

Exposes a ObservableGroupedOrderedCollection<CalendarItem>, however, the grouping is not needed.

### Why not another way?

We could have used an IObservable, but we get a lot for free using the ObservableGroupedOrderedCollection.

### Side effects

None.

## :squid: Permissions

Whenever.


Toggl.Foundation.Tests/MvvmCross/ViewModels/CalendarViewModelTests.cs
```diff
@@ -43,5 +49,30 @@ public sealed class TheConstructor : CalendarViewModelTest
                 tryingToConstructWithEmptyParameters.Should().Throw<ArgumentNullException>();
             }
         }
+
+        public sealed class TheCalendarItemsProperty : CalendarViewModelTest
+        {
+
```
willsb: Extra line

PR #2911, opened 2018-08-10
## What's this? Fix the done button in start TE doesn't turned ✅

### LINKED ISSUE #2809: Check mark isn't changing to green when starting a time entry.
When starting a new timer in app the check mark in the time entry screen isn't changing to green it is greyed out but functional.

## 🐛 Bug Report
When creating a new time entry the check mark (tick) should become green once it is possible to create the time entry but this stays greyed out.

## Context
Occurred first time creating a new entry and following 3 times.

Build/Branch:1.8 Constant feedback build
Phone OS:
Phone model:

## Expected behavior
The check mark should be green

## To reproduce
Try to create a new time entry.

## Suggestions
I had freshly installed the app and created a new account

khoi:
### Relationships

Closes #2809

## How is it done?
Moving away from using .NET Expression and replace it with Reactive Observable

## :squid: Permissions

Squash whenever


Toggl.Daneel/ViewControllers/StartTimeEntryViewController.cs
```diff
@@ -195,6 +195,11 @@ public override void ViewDidLoad()
             // Reactive
             this.Bind(ViewModel.TextFieldInfoObservable, onTextFieldInfo);
 
+            this.Bind(DescriptionTextView
```
simonrozsival: The formatting seems off - please break line after the `(` and it will be nicer.

PR #2912, opened 2018-08-10
Fix spacing on the main view

### LINKED ISSUE #2810: Timer page: the padding between the nav bar and the log is too big
## 🐛 Bug Report
After signing up as a new user (and not having suggestions at the top), the padding between the nav bar and the log is too big. I think this might be caused by the Rating popup, no (maybe worth checking out this #2781)?

Please, check Zeplin for correct designs: https://zpl.io/VDDdy8l

Build/Branch: 1.8 (1616)
Phone OS: SE
Phone model: 11.4.1

## To reproduce
1. Sign up as a new user
2. And create a new entry - bang - there's a lot of space between the nav bar and log

simonrozsival:
## What's this?
There was extra top margin on the main view. This PR removes it.

### Relationships
Closes #2810 

## Review hints
- Sign up using the app.
- Compare with the designs in Zeplin (empty view is shown).
- Start a new time entry and stop it.
- Compare with the designs in Zeplin (the regular log is shown).
- Stop the app. Start the app again.
- Compare with the designs in Zeplin (now the suggestions are shown).

## :squid: Permissions
Squash whenever


simonrozsival:
I think that the designs were  a bit different, I don't remember though.

PR #2723, opened 2018-07-24
Rxfactor the calendar

willsb:
Ref #2524

Please test this thoroughly and report any bugs you can find.

PR #2926, opened 2018-08-13
Remove animation to fix glitch in the reports calendar

### LINKED ISSUE #2083: Reports: A black bar appears between the calendar and the charts
It happens when you scroll the calendar half way up and then tries to scroll between months.

Tested on my iPhone SE. (Toggl 1.5)

simonrozsival:
## What's this?
The Animated constraint binding did not force the layout of the superview which caused the view to look weird.

### Relationships
Closes #2083 

The rxfactoring of the calendar (#2723) might re-introduce this bug so please check this out once the PR is ready, @willsb.

## Why do we want this?
We don't want weird looking UI.

### Side effects
When the number of rows changes while the calendar is half-hidden, it will reveal the whole calendar. The offset is reset for some reason. I don't know why or how to counter that, but I think it is not a problem.

## :squid: Permissions
Squash whenever


Toggl.Daneel/Binding/LayoutConstraintAnimatedConstantTargetBinding.cs
```diff
@@ -25,8 +25,12 @@ protected override void SetValue(nfloat value)
             AnimationExtensions.Animate(
                 Animation.Timings.EnterTiming,
                 Animation.Curves.SharpCurve,
-                () => ((UIView)Target.FirstItem).Superview.LayoutSubviews()
-            );
+                () =>
+                {
+                    var superview = ((UIView)Target.FirstItem).Superview;
+                    superview.LayoutSubviews();
```
khoi: according to the doc
```
You should not call this method directly. If you want to force a layout update, call the setNeedsLayout() method instead to do so prior to the next drawing update. If you want to update the layout of your views immediately, call the layoutIfNeeded() method.
```

PR #2927, opened 2018-08-13
Bind the collection view with the calendar items collection

### LINKED ISSUE #2868: Bind the calendar item observable to the calendar collection view
Depends on #2799 and #2867 

This is the iOS counterpart of #2867. We need to consume the exposed observable and bind it to the existing collection view.

⚠️This concerns only displaying the calendar entries! No need to worry about the additional functionality like TimeEntry creation by tapping calendar entries or the draggable time entries yet. ⚠️

juanlaube:
## What's this?

Binds the calendar items collection with the view. Does not bind each item with the cells though, that will happen in another pull request.

### Relationships

Ref #2868

## How is it done?

This adds the collection view to the view controller, and adds a new data source that conforms to the collection and layout data source interfaces.

### Side effects

None.

## Review hints

Pay attention to the `calculateLayoutAttributes` method, there's some logic there to calculate how many items are at the same time.

## :squid: Permissions

No u.


Toggl.Daneel/ViewSources/CalendarCollectionViewSource.cs
```diff
@@ -0,0 +1,170 @@
...
+{
+    public sealed class CalendarCollectionViewSource
+        : MvxCollectionViewSource, ICalendarCollectionViewLayoutDataSource
+    {
+        private readonly string itemReuseIdentifier = nameof(CalendarItemView);
+        private readonly string hourReuseIdentifier = nameof(HourSupplementaryView);
+        private readonly string currentTimeReuseIdentifier = nameof(CurrentTimeSupplementaryView);
+
+        private readonly ObservableGroupedOrderedCollection<CalendarItem> collection;
+        private IList<CalendarItem> calendarItems => collection[0].ToList();
```
willsb: `collection.First().ToList()` is easier to read. 

Also: Shouldn't you check for null?
Also 2: Shouldn't you cache the call to `.ToList()` to avoid excessive enumerations? (this one is just a thought)

Toggl.Daneel/ViewSources/CalendarCollectionViewSource.cs
```diff
@@ -0,0 +1,170 @@
...
+                .VoidSubscribe(onCollectionChanges)
+                .DisposedBy(disposeBag);
+
+            onCollectionChanges();
+        }
+
+        protected override UICollectionViewCell GetOrCreateCellFor(UICollectionView collectionView, NSIndexPath indexPath, object item)
+        {
+            var cell = collectionView.DequeueReusableCell(itemReuseIdentifier, indexPath) as CalendarItemView;
+            cell.Item = calendarItems[(int)indexPath.Item];
```
willsb: Yeah, that `.ToList()` might get expensive here.

Toggl.Daneel/ViewSources/CalendarCollectionViewSource.cs
```diff
@@ -0,0 +1,170 @@
...
+            if (collection.IsEmpty)
+                return new List<CalendarCollectionViewItemLayoutAttributes>();
+
+            var firstItem = calendarItems[0];
+            var attributes = new List<CalendarCollectionViewItemLayoutAttributes>();
+            var currentGroup = new List<CalendarItem>();
+            var currentGroupStartTime = firstItem.StartTime;
+            var currentGroupEndTime = firstItem.EndTime;
+
+            foreach (var calendarItem in calendarItems)
```
willsb: This is pretty hard to read because of its imperative style. A reduce (aka Aggregate) would be much appreciated here.

PR #2945, opened 2018-08-14
Update CalendarItem to allow for error states

### LINKED ISSUE #2931: Update CalendarItem to allow for error states
Check #2877 to understand the properties that are needed.

We need to support these properties. Needless to say, these should only be visible when the CalendarItem was created from a TimeEntry, so the other constructor should set them to whichever value makes the icons invisible.

arturs105:
### Relationships

Closes #2931

## :squid: Permissions
Whenever.


arturs105:
I completely agree about the tests and will add those, though i thought of something. Maybe we should make these new `bool` properties nullable.  Does it even make sense for calendar events to have properties related to sync?

PR #2948, opened 2018-08-14
Daneel 1.9 - Zoranel

amulware:
Changelog:
```
- Added UNDO prompt when deleting time entries by swiping
- Added workspace selection to reports
- Added tabs to navigate between time entry log and reports

- Fixed crash when user does not have any workspaces
- Improved performance of main time entry log view

- 14 Other fixes and improvements
```


amulware:
Github. Y u close another PR just because the branch is mostly identical?

PR #2956, opened 2018-08-15
📐 Colorful Activity Indicator 

### LINKED ISSUE #2765: Colorful Activity Indicator
Make changing the color of `ActivityIndicatorView` easier 🎏 

Suggestions:
- Allow passing in a `UIColor` to the view constructor
- Force the image render mode to `.AlwaysTemplate`

khoi:
## What's this?
Some changes made to the spinning indicator in iOS
- Rename the `StartAnimation` and `StopAnimation` to `StartSpinning` and `StopSpinning`
- Allow passing in a UIColor to change the indicator color

### Relationships
Closes #2765 and #2960

## Why do we want this?
Make our life a little bit easier 

## :squid: Permissions

Anyone


Toggl.Daneel/Views/ActivityIndicatorView.cs
```diff
@@ -31,6 +26,15 @@ public ActivityIndicatorView(IntPtr handle)
         {
         }
 
+        public UIColor IndicatorColor
+        {
+            set
+            {
+                Image = Image.ImageWithRenderingMode(UIImageRenderingMode.AlwaysTemplate);
+                TintColor = value;
+            }
```
arturs105: I don't see a reason to not have a getter, but that is easy to add whenever it's needed, so this can remain like this as well.

Toggl.Daneel/Views/ActivityIndicatorView.cs
```diff
@@ -41,37 +45,26 @@ public override void AwakeFromNib()
         /// Starts the spinning animation of the activity indicator.
         /// NOTE: may not work when called from ViewDidLoad, use ViewWillAppear or ViewDidAppear.
         /// </summary>
-        public void StartAnimation()
+        public void StartSpinning()
         {
+            StartAnimating();
```
arturs105: Why do you call this?

khoi: 🙈

Toggl.Daneel/Views/ActivityIndicatorView.cs
khoi: I can't think of a reason that we would ever want to get the indicator color :D

khoi:
@arturs105 please take a look again when u can 🙇

PR #2957, opened 2018-08-15
🐞 Fix crashes caused by Mvx not being initialised yet

### LINKED ISSUE #2923: NullReferenceException in MvxAndroidBindingContextHelpers.Current
## 💥 Crash Report
By initialising app center earlier, we are getting a lot more crash reports which hopefully will account for the so far unexplained crashes (see #2833).

### 👥 Impact
- about 1600 crashes, over 500 users, for 84% of all crashes in 1.3
- all sorts of devices and OS versions

### 🛰 Link
https://appcenter.ms/orgs/Toggl/apps/Toggl-Android/crashes/groups/96ea405eecd1afb1ffe35f417763a4f48b2a2ba5?appVersion=1.3
https://appcenter.ms/orgs/Toggl/apps/Toggl-Android/crashes/groups/8153f92a96249266df0bb5c054e7b125e7947458?appVersion=1.3

```cs
Mvx.Resolve[TService] ()
MvxAndroidBindingContextHelpers.Current[T] ()
MvxAndroidBindingContextHelpers.Current ()
MvxLayoutInflater.Inflate (System.Int32 resource, Android.Views.ViewGroup root, System.Boolean attachToRoot)
LayoutInflater.n_Inflate_ILandroid_view_ViewGroup_Z (System.IntPtr jnienv, System.IntPtr native__this, System.Int32 resource, System.IntPtr native_root, System.Boolean attachToRoot)
```

dshr:
## What's this?
This is a fix to the numerous crashes we've had with 1.3.

### Relationships
Closes #2923

## Why do we want this?
To not have crazy crashes.

## How is it done?
After looking through the stack trace, I came across several issues in the MvvmCross repo which all had similar stacktraces and symptoms — the crash came either on startup, or when the app was sent to bg, partially gc'ed, and then called back to the foreground. The conclusion was that calls to Mvx bindings for adapters were done at a time where mvx was not yet initialised.

There were two proposed fixes there:
1.  Pass an explicit binding context to the adapter
2. Ensure that the mvx singleton is setup on startup

As it turned out, the first approach had some problems — the newly created binding contexts were then not put into the mvx binding context stack, which made things just more confusing, so I went with the second one.

Another piece of advice I got from @willsb was to change the SplashScreen from a `MvxSplashScreenAppCompatActivity` to `MvxSplashScreenAppCompatActivity<Setup, App<LoginViewModel>>` which should ensure that mvx is loaded.

### Why not another way?
Couldn't really come up with anything else, so all suggestions are welcome.

### Side effects
Slightly slower startup times since we're waiting for `EnsureInitialized`.

## Review hints
Couldn't come up with a reliable way to replicate this, but my steps were:
1. Run app
2. Background it
3. Sleep sim or load a heavy app/web page
4. Bring back app from the bg or open it again from the apps list

## :squid: Permissions
Go for it!


Toggl.Giskard/Startup/SplashScreen.cs
```diff
@@ -23,7 +24,7 @@ namespace Toggl.Giskard
         Categories = new[] { "android.intent.category.BROWSABLE", "android.intent.category.DEFAULT" },
         DataSchemes = new[] { "toggl" },
         DataHost = "*")]
-    public class SplashScreen : MvxSplashScreenAppCompatActivity
+    public class SplashScreen : MvxSplashScreenAppCompatActivity<Setup, App<LoginViewModel>>
```
daividssilverio: I'm not sure about this. I thought it would be enough to have the `TogglApplication : MvxAppCompatApplication<Setup, App<LoginViewModel>>` as we have right now.
Will this make the initialize itself twice? 🤔 @willsb mastery required.

willsb: It won't do it twice, no. It always checks whether it has been inited or not

PR #2962, opened 2018-08-15
🚧 Fix signup password doubling

### LINKED ISSUE #2959: Sign up "doubles" the password
## 🐛 Bug Report
When I signed up using the app I entered some password, but when I tapped the "sign up" button, the number of dots doubled. When I then had to log in to the web app, I had to repeat the password twice to log in.

Also the same happens when I enter some password in the login view and then switch to the sign up view - the original password is repeated twice.

## Context

Build/Branch: 1.9

## Expected behavior
Use the password the user typed just once, not twice.

## Suggestions
I guess there are some misbehaving bindings, but I can't really tell without looking at the code.

sofokli1:
## What's this?
Fix for the problem of the signup password UITextField repeating the password every time it became firstResponder and every time it toggled the show hide password.

### Relationships
Closes #2959

## Why do we want this?
We want people to signup with the password that they give and some other string that we give insert by error.

## How is it done?
There was some code to insert the `Text` to the UITextfield again (thus repeating the already available text) for some unknown reason.

### Why not another way?
Can't see an other way.

### Side effects
None that I know off but will see from functional testing from reviewers.

## Review hints
To reproduce in develop, just go to signup page type a couple of characters and toggl the show/hide apps or resignFirstResponder and becomeFirstResponder for the password field.

## :squid: Permissions
Whenever


khoi:
The doubling issue is fixed, but I found one weird behavior that could be improved. 

- In the login screen, type in a valid email and a password that's long enough
- Switch to Sign up, the fields are pre-filled, but the signup button is still disabled. 

Only when the user focus on the password field again, then the button is enabled.

simonrozsival:
Please create an issue, @khoi, let's keep this simple!

amulware:
I believe this should be rebased against the 1.9 release banch?

sofokli1:
I took a look at @khoi's comment and the solution is not trivial. It will need rxfactoring of the signup vm/vc as the login had. Will make an other PR for that

khoi:
Let's do it in another PR. I've already created an issue https://github.com/toggl/mobileapp/issues/2973

PR #2967, opened 2018-08-15
Giskard Hotfix 1.3.1

dshr:
## What's this?
A hotfix for the abundant crashes we've encountered in 1.3. 

## Why do we want this?
Crashes are bad, mkay?

## How is it done?
It's just a hotifx.

## :squid: Permissions
Only merge, since this is a hotfix.


amulware:
Released to a percentage of user, so might as well merge it.

PR #2972, opened 2018-08-16
Fix bottom bar covering TE log

### LINKED ISSUE #2954: Bottom bar is covering the last TE on the main log
## 🐛 Bug Report
The last TE on the main log is being covered by the running time entry card because of the bottom nav bar.

## Context
This is a simple bug. It was introduced when the bottom bar was added to iOS.

Build/Branch:
Phone OS: iOS
Phone model: iPhone SE

## Expected behavior
The main log should scroll to show the last item when the running time entry card is visible.

## To reproduce
1. Have enough TEs to fill the whole screen.
2. Start a new TE.
3. Check that the last item is covered by the running te card and you can't scroll over it over the card (without holding your finger to keep the list scrolled past its bounds).

khoi:
## What's this?
Address the 🐛 where the TE bug is covered by the running TE.

### Relationships
Closes #2954

## How is it done?

1. The bottom constraint of the tableview is now anchored to safeareabottomguide's top with priority 750
2. Tableview's bottom to top current running time entry constraint added with higher priority of 1000

When there is a running TE, enable the `2` constraint, and disable it otherwise.

### Side effects
🤷🏻‍♂️

## :squid: Permissions
Whenever


simonrozsival:
But please rebase this so we can (well, Paul can) merge into the 1.9 release branch.

amulware:
:doit:

PR #2977, opened 2018-08-16
Calendar y u so mean

### LINKED ISSUE #2975: Calendar service does not return a list of calendars
Steps to reproduce (works best on a fresh install):
1. Go to the calendar tab
2. Press "Get started"
3. Press "OK" in the permission dialog
4. 💥 Calendar list is empty

If you then restart the app and press "Get started" again, the everything works as it should.

My theory was that, this was caused by using different instances of `EKEventStore` in `PermissionsService` and `CalendarService`, but trying to inject a single instance into those service yielded no results. However, we should still try to use only a single instance of the event store, so making the changes to inject it should be part of this PR.

@juanlaube Suggested that we might try to recreate the event store after the permission has been granted. If this works, the above paragraph should be ignored.

juanlaube:
## What's this?

This instantiate the event store again after the calendar authorization status changes. We need this because for some reason (a bug from Apple, Xamarin, us?) the GetCalendars method returns an empty array.

### Relationships

Closes #2975

## How is it done?

The permissions service exposes an observable that emits when the authorization status changes and the calendar service subscribes to that, reloading the event store.

## :squid: Permissions

Whenever.


willsb:
Why not simply instantiate a new one every time?

PR #2980, opened 2018-08-17
📐 Fix bottom spacing in reports

### LINKED ISSUE #2893: Reports: WS select button hides the content
## 🐛 Bug Report
Currently, WS select button hides the content permanently when you scroll at the bottom of the Reports list.

## Context

Build/Branch: I'm testing the current `develop` branch

## Expected behaviour
As we don't want the button to obscure any content permanently, extra padding should be added below the projects table only when the WS button is visible. When user has only one WS and we don't show the button, extra padding is not needed.

Design how this should look like is available in Zeplin: https://zpl.io/2vYZqA5

sofokli1:
## What's this?
Small ui tweak in reports.

### Relationships
Closes #2893

## Why do we want this?
We don't want to hide anything to our users.

## How is it done?
I just add an empty view to as a table footer view with the desired height.

### Why not another way?
This is the simplest.

### Side effects
0

## Review hints
Go to reports and scroll to the bottom. If the WS button is still hiding the table view than i didn't do a good job.

## :squid: Permissions
Whenever


simonrozsival:
The issue says:

> When user has only one WS and we don't show the button, extra padding is not needed.

It should be easy to hide the empty view when the btn is not visibile (or make the height 0 probably?).

khoi:
Any difference between this and using table view insets?

sofokli1:
@khoi Usually I have found that insets bite me later on. This is straight forward.
@simonrozsival addressed

Toggl.Daneel/ViewSources/ReportsTableViewSource.cs
```diff
@@ -16,17 +16,20 @@ public sealed class ReportsTableViewSource : MvxTableViewSource
         private const string cellIdentifier = nameof(ReportsLegendViewCell);
         private const string headerCellIdentifier = nameof(ReportsHeaderView);
 
-        public ReportsViewModel ViewModel { get; set; }
+        private ReportsViewModel viewModel { get; set; }
```
simonrozsival: If we are making this change, let's make it properly - please make this a `readonly` field, there is no need to have default getters or setters on a private member.

PR #2988, opened 2018-08-17
Calendar event tracking

### LINKED ISSUE #2942: Track calendar events
We need to track the events to our analytics service in order to obtain insights on how the calendar is being used. 

### Absolute must haves:

- [x] Edit view opened from a calendar time entry
- [x] Time entry created from a calendar event
- [x] Time entry created from a time span
- [x] Time entry was offset using the long press and drag feature
- [x] Time entry changed duration using the slider that changes duration
- [x] User clicked the Get Started button to kickoff the onboarding

willsb:
## What's this?

This adds and tracks the events that happen when the user is using the calendar view 

### Relationships

Closes #2942

## Why do we want this?

These stats can help us shape the feature based on usage.

## Review hints

Check that everything is covered and that the events contain actual relevant info

## :squid: Permissions

Anyone


Toggl.Foundation/Analytics/IAnalyticsService.cs
```diff
@@ -92,6 +92,16 @@ public interface IAnalyticsService
 
         IAnalyticsEvent TwoRunningTimeEntriesInconsistencyFixed { get; }
 
+        IAnalyticsEvent CalendarOnboardingStarted { get; }
+
+        IAnalyticsEvent EditViewOpenedFromCalendar { get; }
+
+        IAnalyticsEvent TimeEntryCreateFromCalendarEvent { get; }
+
+        IAnalyticsEvent TimeEntryCreatedFromCalendarTapping { get; }
```
amulware: Create/Created - let's stick to one.

Toggl.Foundation/Analytics/IAnalyticsService.cs
```diff
@@ -92,6 +92,16 @@ public interface IAnalyticsService
 
         IAnalyticsEvent TwoRunningTimeEntriesInconsistencyFixed { get; }
 
+        IAnalyticsEvent CalendarOnboardingStarted { get; }
+
+        IAnalyticsEvent EditViewOpenedFromCalendar { get; }
+
+        IAnalyticsEvent TimeEntryCreateFromCalendarEvent { get; }
```
amulware: This doesn't open the start view to create, does it? If so, we should add another origin to the general TE creation event.

Not sure if we need both events then?

Toggl.Foundation/Analytics/IAnalyticsService.cs
amulware: Also, how does this work with respect to the general TE creation event?

Toggl.Foundation/Analytics/IAnalyticsService.cs
```diff
@@ -92,6 +92,16 @@ public interface IAnalyticsService
...
+
+        IAnalyticsEvent TimeEntryCreatedFromCalendarTapping { get; }
+
+        IAnalyticsEvent TimeEntryDurationChangedFromCalendar { get; }
```
amulware: What about moving an entry up and down? And changing the start time? Are these not supported (I saw designs)?

Maybe a general "ChangedTime" event with a property that can have the three different values?

PR #2990, opened 2018-08-17
Add more rating view events

amulware:
## What's this?
This adds more rating view events. This was already done in 1.8.2 and released to users, so we better have the same change here too!

### Relationships
Cherry pick of #2987

## Why do we want this?
Because this was released with 1.8.2, while 1.9 was branched from 1.8.1 and we don't want to lose the events until 1.10 again.

## How is it done?
git fu

## Review hints
Make sure I didn't do git poo.

## :squid: Permissions
Whoever is handling 1.9.

### DIFF
--- a/Toggl.Foundation.MvvmCross/ViewModels/RatingViewModel.cs
+++ b/Toggl.Foundation.MvvmCross/ViewModels/RatingViewModel.cs
@@ -94,10 +94,16 @@ namespace Toggl.Foundation.MvvmCross.ViewModels
             impressionSubject.OnNext(isPositive);
             analyticsService.UserFinishedRatingViewFirstStep.Track(isPositive);
 
-            var outcome = isPositive
-                ? RatingViewOutcome.PositiveImpression
-                : RatingViewOutcome.NegativeImpression;
-            onboardingStorage.SetRatingViewOutcome(outcome, timeService.CurrentDateTime);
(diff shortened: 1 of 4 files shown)


Toggl.Foundation.MvvmCross/ViewModels/RatingViewModel.cs
```diff
@@ -164,11 +172,13 @@ public void Dismiss()
             if (impressionSubject.Value.Value)
             {
                 onboardingStorage.SetRatingViewOutcome(RatingViewOutcome.AppWasNotRated, timeService.CurrentDateTime);
+                analyticsService.RatingViewSecondStepDontRate.Track();
```
juanlaube: When tapping the `CtaButton`, it calls `PerformMainAction` and `Dismiss`. This event and next one are tracked regardless of the action selected.

amulware: I'm not sure I follow. If it calls both, does that mean the previous events are also tracked too often?

amulware: The answer appears to be yes. Since this PR reflects the changes of 1.8.2 I would still like to merge it as is, and we can then fix this separately.

PR #2991, opened 2018-08-17
Zoranel intermediate merge

amulware:
## What's this?
This is an intermediate merge of Daneel 1.9 - Zoranel.

### Relationships
Zoranel lives here: #2948

## Why do we want this?
So we can already use the fixes to the RX main view on Android as well.

## How is it done?
I branched from `zoranel`. :magic:

### Why not another way?
This leaves the original release PR intact, it's the easiest.

### Side effects
A slightly more locally complicated git graph? Can't think of anything else.

## Review hints
Check that this is actually the release branch and not some random stuff.

## :squid: Permissions
Just me, and proper merge, no 🦑.


amulware:
Ping @daividssilverio who wants these changes for his shenanigans.

Issue #2932, opened 2018-08-13
Update Calendar cell to display error states


willsb:
Check #2877 for designs. Depends on #2931

We now need to consume the properties introduced in #2931 in the View layer. Not much else to add.

juanlaube:
Closed by #2976

PR #2994, opened 2018-08-17
Giskard 1.3.2 hotfix

### LINKED ISSUE #2923: NullReferenceException in MvxAndroidBindingContextHelpers.Current

daividssilverio:
## What's this?
This is another attempt to fix the startup crashes on giskard

### Relationships
This undoes what was done on giskard-1.3.1-hotfix.

Closes #2923

## Why do we want this?
Because it stops our startup crashes o/

## How is it done?
1. The calls to `ChangeStatusBarColor` were all removed in favour to xml theming configuration.
2. The method `ensureDataSourceInitializationIfLoggedIn` was added to giskard's setup. This makes sure the ITogglDataSource is registered on the Mvx IoC handler. This removes the second crash unveiled by (1). More on #2932
 
### Why not another way?
We could do 
```
var mvxSetup = MvxAndroidSetupSingleton.EnsureSingletonAvailable(ApplicationContext); mvxSetup.EnsureInitialized();
```
before each call to `ChangeStatusBarColor` on every activity that uses it. But that's just ugly. Using xml to configure themes is the standard way of doing this kind of stuff on Android. We were not doing anything magical there, xml is fine for that.

### Side effects
Rehydration will take more time execute, but it works.
The initial startup should be a tad slower because of the `ensureDataSourceInitializationIfLoggedIn` call.

This will unveil other bugs related to rehydration, now the app shouldn't crash before the activities can be resumed, ViewModel saving and restoring has to be done at some point.

## Review hints
Enable `Don't keep activities` and set `Limit background processes` to `No background processes` and you should be able to cause tombstoning basically every time you change screens.

With that configuration done, you should see some bugs/crashes when creating TEs, Projects, Clients, etc.

Disable the said configurations and you should able to use the app as it was before, no crashes and all status bars still have their right colours, pre-lollipop inclusive.
 
## :squid: Permissions
Don't. I'm not sure I did the right branching/PR process here. I'll wait for @amulware for some guidance. Also `giskard-1.3.1-hotfix` is @dshr's.

### DIFF
--- a/Toggl.Giskard/Activities/AboutActivity.cs
+++ b/Toggl.Giskard/Activities/AboutActivity.cs
@@ -1,6 +1,5 @@
 ﻿using Android.App;
 using Android.Content.PM;
-using Android.Graphics;
 using Android.OS;
 using Android.Support.V7.Widget;
 using MvvmCross.Platforms.Android.Presenters.Attributes;
@@ -17,8 +16,6 @@ namespace Toggl.Giskard.Activities
     {
         protected override void OnCreate(Bundle bundle)
         {
-            this.ChangeStatusBarColor(Color.ParseColor("#2C2C2C"));
-
             base.OnCreate(bundle);
             SetContentView(Resource.Layout.AboutActivity);
             OverridePendingTransition(Resource.Animation.abc_slide_in_right, Resource.Animation.abc_fade_out);
--- a/Toggl.Giskard/Activities/BrowserActivity.cs
+++ b/Toggl.Giskard/Activities/BrowserActivity.cs
@@ -1,14 +1,12 @@
 ﻿using System;
 using Android.App;
 using Android.Content.PM;
-using Android.Graphics;
 using Android.OS;
 using Android.Support.V7.Widget;
 using Android.Webkit;
 using MvvmCross.Droid.Support.V7.AppCompat;
 using MvvmCross.Platforms.Android.Presenters.Attributes;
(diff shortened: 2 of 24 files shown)


daividssilverio:
Closing this as the branching was done wrong.

Issue #2777, opened 2018-07-31
Scrollability in time view calendar


amulware:
## 🐛 Bug Report
On my Xiaomi 5A (running Android 7), there is a scrollbar in the calendar, so I can scroll to see the 6th week of a month.

On the random Samsung Galaxy J5 (SM-J510FN) (running Android 6) there is no scrollbar. There is also no separator line between the calendar and buttons below.

Xiaomi 5A / Samsung J5

<img width="1121" alt="screen shot 2018-07-31 at 12 21 36" src="https://user-images.githubusercontent.com/3987892/43456677-52896272-94bc-11e8-863e-b3be20231efa.png">

## Expected behavior
This should work the same on all phones. Either be scrollable or big enough/adapt to the size of the calendar. Ping @pe0ny for intended behaviour.

## To reproduce
Check different phones/OSes(?), and check with a month that spans six weeks.

pe0ny:
Since it should fit to all screen sizes it should be big enough - scrolling is not intended behaviour.

PR #3002, opened 2018-08-20
Fix last line of date picker being unresponsive

### LINKED ISSUE #2745: Dates from last week of each month cannot be selected
## 🐛 Bug Report
When you want to edit a time entry and select a different start or end date, you are not able to select any date that is from the last week of each month. It does not matter if the current date falls into the last week of a month, you still cannot select any date from that last week.

It is also valid for previous months, not only for the current one.

This seems to be an issue for Huawei P9 Lite.

## Context
This issue was first encountered by a user (using Huawei P9 Lite) and later confirmed by me once I've tested it on the same phone.

I've used new time entries (created for this issue) as well as old time entries, same behavior.

I've also manually changed the time on my phone (to May & April) to create time entries in previous months - results were the same, you cannot select any date that is in the last week of a month.

That week is clearly shown on the actual screen, there are no clipping issues, it is just unclickable (screenshots below).
image
_June - can only select dates from 14 - 24_

image
_July - can only select dates from 1 - 22_ || July 26 is the current date and is selected by default; if I click any other date, i.e. July 19, I won't be able to click July 26 again

Build/Branch: **Toggl Giskard - 1.1** - downloaded from the Play Store
Phone OS: **Android**
Phone model: **Huawei P9 Lite**

## Expected behavior
Select any date from the calendar when editing a time entry.

## To reproduce
This might be related only to particular phone model(s); I just tried to select/edit a date from last week of a month and it just wouldn't let me - those dates are just not reacting.

dshr:
## What's this?
A little fix that helps the native date picker UI — when a month over spans 5 weeks, the last week becomes unresponsive.

### Relationships
Closes #2745

## Why do we want this?
So people can use the date picker.

## How is it done?
By adding a negative bottom margin to the date picker equal to the height of the day cell.

### Why not another way?
The date picker is a native android one, so the only other option is writing one from scratch, which is way too much effort.

### Side effects
None that I know of.

## Review hints
1. Open the Edit TE screen
2. Try changing the start date or the end date to a day that falls on a 5th week of a month.

## :squid: Permissions
Go for it!


daividssilverio:
#2950 should be merged before this. This pr's branch is broken.

daividssilverio:
This fix was meant to make the last visible line of the calendar clickable, right?
It's working, but I can't see the last line of the calendar of December 2017 for example. Dec 31 is missing.

Toggl.Giskard/Resources/layout/SelectTimeFragment.axml
```diff
@@ -53,4 +53,4 @@
             android:layout_height="36dp"
             android:layout_marginRight="16dp" />
     </LinearLayout>
-</RelativeLayout>
\ No newline at end of file
+</RelativeLayout>
```
daividssilverio: Also, discard this changes that don't change anything.

daividssilverio:
Hey, I think you can also fix #2777 together with this one by changing the fragment dialog height.

dshr:
Yeah, working on #2777 right now. Tweaking the height makes "shorter" months look weird, so I'm adding a stretchy pager view.

daividssilverio:
Don't forget to assign yourself o/

dshr:
Didn't work, the `DatePicker` always says it's the same height. I'm doing the easy thing and simply add more space.

Also closing this, since the new PR will handle it.

PR #3003, opened 2018-08-20
Add calendar settings view model

### LINKED ISSUE #2888: Create the CalendarSettingsViewModel
This is the ViewModel that will display the configurations of the calendar. 

It'll basically contain:

- Some way to enable/disable the calendar integration. 
  - It shows on if we have permission and off otherwise
  - Tapping this item opens the iOS settings so users can give/revoke permissions there
- A way to select which user calendars should be included in the calendar integration
  - The list of user calendars is only visible when the integration is on
  - Its functionality is literally the same as the one in the calendar onboarding
  - We can probably reuse the view source if we make the cells injectable

arturs105:
### Relationships
Closes #2888

## How is it done?
I decided to extract some functionality from `SelectUserCalendarsViewModel`, since both of these require getting a list of user calendars and selecting calendars.

## :squid: Permissions
Whenever.


Toggl.Foundation.MvvmCross/ViewModels/Calendar/SelectUserCalendarsViewModel.cs
```diff
@@ -1,86 +1,48 @@
...
+        protected override IObservable<Unit> SelectCalendar(SelectableUserCalendarViewModel calendar)
         {
-            if (selectedCalendarIds.Contains(calendar.Id))
-                selectedCalendarIds.Remove(calendar.Id);
-            else
-                selectedCalendarIds.Add(calendar.Id);
+            var unit = base.SelectCalendar(calendar);
 
-            doneActionEnabledSubject.OnNext(selectedCalendarIds.Any());
+            doneActionEnabledSubject.OnNext(SelectedCalendarIds.Any());
```
willsb: This looks really weird. I think that it'd be better if you, instead, used the very action's observable instead of doing this override.

Toggl.Foundation.MvvmCross/ViewModels/Calendar/SelectUserCalendarsViewModelBase.cs
```diff
@@ -0,0 +1,64 @@
...
+                .Select(calendars => calendars.Select(toSelectable))
+                .Do(calendars => calendars.ForEach(calendar => Calendars.InsertItem(calendar)));
+        }
+
+        private SelectableUserCalendarViewModel toSelectable(UserCalendar calendar)
+            => new SelectableUserCalendarViewModel(calendar, false);
+
+        protected virtual IObservable<Unit> SelectCalendar(SelectableUserCalendarViewModel calendar)
+        {
+            if (SelectedCalendarIds.Contains(calendar.Id))
```
willsb: The lack of curly braces here looks odd

Toggl.Foundation.MvvmCross/ViewModels/Settings/CalendarSettingsViewModel.cs
```diff
@@ -0,0 +1,41 @@
...
+using Toggl.Multivac;
+using Toggl.Multivac.Extensions;
+
+namespace Toggl.Foundation.MvvmCross.ViewModels.Settings
+{
+    public sealed class CalendarSettingsViewModel : SelectUserCalendarsViewModelBase
+    {
+        private readonly IPermissionsService permissionsService;
+
+        public bool PermissionGranted { get; }
```
willsb: Does revoking also reset the app state? If not, this needs to be an observable

Toggl.Foundation.MvvmCross/ViewModels/Settings/CalendarSettingsViewModel.cs
```diff
@@ -0,0 +1,41 @@
...
+
+        public UIAction RequestAccessAction { get; }
+
+        public CalendarSettingsViewModel(
+            IInteractorFactory interactorFactory,
+            ISchedulerProvider schedulerProvider,
```
willsb: Why are you injecting this?

Toggl.Foundation.MvvmCross/ViewModels/Calendar/SelectUserCalendarsViewModelBase.cs
```diff
@@ -0,0 +1,64 @@
...
+                .GetUserCalendars()
+                .Execute()
+                .Select(calendars => calendars.Select(toSelectable))
+                .Do(calendars => calendars.ForEach(calendar => Calendars.InsertItem(calendar)));
+        }
+
+        private SelectableUserCalendarViewModel toSelectable(UserCalendar calendar)
+            => new SelectableUserCalendarViewModel(calendar, false);
+
+        protected virtual IObservable<Unit> SelectCalendar(SelectableUserCalendarViewModel calendar)
```
willsb: Should be made private after the changes I requested

Toggl.Foundation.MvvmCross/ViewModels/Calendar/SelectUserCalendarsViewModel.cs
arturs105: I'm not sure, i understand.

Toggl.Foundation.MvvmCross/ViewModels/Calendar/SelectUserCalendarsViewModelBase.cs
arturs105: I think having no curly braces is fine, since there is only one method call per `if` branch. But if you insist on having curly bois, then i'll add them.

Toggl.Foundation.MvvmCross/ViewModels/Settings/CalendarSettingsViewModel.cs
arturs105: Yes, any change to app permissions resets the app state.

Toggl.Foundation.MvvmCross/ViewModels/Settings/CalendarSettingsViewModel.cs
arturs105: That's legacy code, from the time when i needed to call `AsDriver()`. Will 🔥.

PR #3007, opened 2018-08-20
🚧 Refcatore signup

### LINKED ISSUE #2973: Sign up button disabled after coming from Sign in
## 🐛 Bug Report

In Sign in screen, if the user enters an email address and a password, then hit "Sign Up". The prefilled fields are taken over to the Signup screen, but the "Sign up" button state is disabled. Only when the user refocuses to the password field, the button is enabled again.

screen shot 2018-08-16 at 16 49 18

## Context
`develop` branch

## Expected behavior
The sign-up button is enabled.

sofokli1:
## What's this?
This PR refactors the signup VC,VM,activity to use the new RX way we are doing things.

### Relationships

Closes #2973

## Why do we want this?
Apart from the refactoring it should and close an issues it moves the code one step further away from the mvx.

## How is it done?
By replacing old code with new one.

### Why not another way?
The other way is to do nothing

### Side effects
There shouldn't be any, testing will tell.

## Review hints
Apart from taking a quick look at the code I suggest rigorous functional testing since it will hinder new users from signing up if there is any problem.

## :squid: Permissions
🦑


Toggl.Foundation.MvvmCross/ViewModels/SignupViewModel.cs
```diff
@@ -35,48 +45,45 @@ public sealed class SignupViewModel : MvxViewModel<CredentialsParameter>
...
-        public bool IsShowPasswordButtonVisible { get; set; }
+        public IObservable<ShakeTargets> Shake { get; }
 
-        public IMvxCommand GoogleSignupCommand { get; }
+        public IObservable<string> ErrorMessage { get; }
 
-        public IMvxCommand TogglePasswordVisibilityCommand { get; }
+        public IObservable<bool> IsPasswordMasked { get; }
 
-        public IMvxAsyncCommand SignupCommand { get; }
```
daividssilverio: This guy too, deprecate please.

Toggl.Foundation.MvvmCross/ViewModels/SignupViewModel.cs
```diff
@@ -35,48 +45,45 @@ public sealed class SignupViewModel : MvxViewModel<CredentialsParameter>
...
-
-        public IMvxAsyncCommand LoginCommand { get; }
```
daividssilverio: This guy too, deprecate please.

Toggl.Foundation.MvvmCross/ViewModels/SignupViewModel.cs
```diff
@@ -35,48 +45,45 @@ public sealed class SignupViewModel : MvxViewModel<CredentialsParameter>
...
 
         public bool IsCountryErrorVisible { get; private set; } = false;
 
         public bool IsCountryValid => countryId.HasValue;
 
-        public Email Email { get; set; } = Email.Empty;
-
-        public Password Password { get; set; } = Password.Empty;
-
-        public bool IsLoading { get; private set; }
```
daividssilverio: This guy too, deprecate instead please.

Toggl.Foundation.MvvmCross/ViewModels/SignupViewModel.cs
```diff
@@ -250,22 +320,16 @@ private async Task pickCountry()
...
         {
-            var parameter = CredentialsParameter.With(Email, Password);
+            if (isLoadingSubject.Value)
+                return Task.CompletedTask;
+
+            var parameter = CredentialsParameter.With(emailSubject.Value, passwordSubject.Value);
             return navigationService.Navigate<LoginViewModel, CredentialsParameter>(parameter);
         }
-
-        private void OnEmailChanged()
```
daividssilverio: These three methods are also necessary for mvx bindings. Please deprecate them instead OnEmailChanged; OnPasswordChanged; OnIsLoadingChanged.

daividssilverio:
Also, please merge develop into your branch. Android crashes at runtime because of the `MainTabBarViewModel` not being there for the OS. This problem has been fixed on develop.

Toggl.Giskard/Activities/SignUpActivity.cs
```diff
@@ -15,15 +17,43 @@ namespace Toggl.Giskard.Activities
...
+            this.Bind(ViewModel.IsLoading, progressBar.BindIsVisible());
+            this.Bind(ViewModel.SignupEnabled, signupButton.BindEnabled());
+            this.Bind(ViewModel.IsCountryErrorVisible, countryErrorView.BindIsVisible());
+
+            //Commands
+            this.Bind(loginCard.Tapped(), ViewModel.Login);
+            this.Bind(signupButton.Tapped(), ViewModel.Signup);
+            this.Bind(passwordEditText.EditorActionSent(), ViewModel.Signup);
+            this.BindVoid(googleSignupButton.Tapped(), ViewModel.GoogleSignup);
+            this.Bind(countryNameTextView.Tapped(), ViewModel.PickCountry);
```
daividssilverio: Please change the target of this bind to the view with Id `SignUpCountrySelection`. It's hard to click with bind on the TextView.

Toggl.Foundation.MvvmCross/ViewModels/SignupViewModel.cs
```diff
@@ -27,6 +29,15 @@ namespace Toggl.Foundation.MvvmCross.ViewModels
     [Preserve(AllMembers = true)]
     public sealed class SignupViewModel : MvxViewModel<CredentialsParameter>
     {
+        [Flags]
+        public enum ShakeTargets
+        {
+            None = 0,
+            Email = 1,
+            Password = 2,
+            Country = 3
```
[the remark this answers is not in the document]
khoi: Use the bit-shift operator for readability
```
None = 0,
Email = 1 << 0,
Password = 1 << 1,
Country = 1 << 2,
...
```

Toggl.Foundation.MvvmCross/ViewModels/SignupViewModel.cs
```diff
@@ -35,48 +46,45 @@ public sealed class SignupViewModel : MvxViewModel<CredentialsParameter>
...
 
-        public bool SignupEnabled
-            => Email.IsValid
-            && Password.IsValid
-            && !IsLoading
-            && countryId.HasValue;
+        public IObservable<bool> IsLoading { get; }
 
-        public bool IsShowPasswordButtonVisible { get; set; }
+        public IObservable<bool> SignupEnabled { get; }
```
khoi: We can make use of `RxAction` to simplify the code quite a bit.

Toggl.Foundation.MvvmCross/ViewModels/SignupViewModel.cs
```diff
@@ -145,29 +211,44 @@ private void setCountryIfNeeded(ICountry country)
...
+            isCountryErrorVisibleSubject.OnNext(true);
         }
 
-        private async Task signup()
+        public async Task Signup()
         {
+            var shakeTargets = ShakeTargets.None;
+            shakeTargets |= emailSubject.Value.IsValid ? ShakeTargets.None : ShakeTargets.Email;
+            shakeTargets |= passwordSubject.Value.IsValid ? ShakeTargets.None : ShakeTargets.Password;
+            shakeTargets |= countryId.HasValue ? ShakeTargets.None : ShakeTargets.Country;
```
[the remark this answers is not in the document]
khoi: I feel like 
```csharp
if  !emailSubject.Value.IsValid  {
     shakeTargets |= ShakeTargets.Email;
}
if !passwordSubject.Value.IsValid {
   shakeTargets |= ShakeTargets. Password;
}
```
is easier to read

Toggl.Foundation.Tests/MvvmCross/ViewModels/SignupViewModelTests.cs
```diff
@@ -655,65 +754,121 @@ protected override void AdditionalViewModelSetup()
...
                 var parameter = CredentialsParameter.With(email, password);
+                var expectedValues = new[] { Password.Empty.ToString(), password.ToString() };
+                var actualValues = new List<string>();
+                viewModel.Password.Subscribe(actualValues.Add);
+
+                viewModel.Prepare(parameter);
+
+                TestScheduler.Start();
+                CollectionAssert.AreEqual(expectedValues, actualValues);
+                //var observer = TestScheduler.CreateObserver<string>();
```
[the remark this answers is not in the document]
sofokli1: 🔥

khoi:
[0001-wip.patch.txt](https://github.com/toggl/mobileapp/files/2327092/0001-wip.patch.txt)
This is how I would use RxAction for the SignUp action. 

I still have to keep
```
this.Bind(ViewModel.SignupAction.Enabled.Select(signupButtonTitleColor), SignupButton.BindTitleColor());
```

since we're changing the title color manually. 

If it was done by setting the `UIControl.State` property, then 
```
            SignupButton.BindToAction(ViewModel.SignupAction);
```
would also take care of it for us

sofokli1:
@khoi The main problem that I see with the `SignupAction` is the use of the `SignupEnabled` in the constructor. If we do that we lose the shaking of the field when taping the signup button and also breaks many tests that relay on manually calling the signup method. I would say to leave it as is for now.

Toggl.Daneel/ViewControllers/SignupViewController.cs
```diff
@@ -29,79 +28,68 @@ public sealed partial class SignupViewController : MvxViewController<SignupViewM
...
-            bindingSet.Bind(SelectCountryButton).To(vm => vm.PickCountryCommand);
-            bindingSet.Bind(LoginCard)
-                      .For(v => v.BindTap())
-                      .To(vm => vm.LoginCommand);
+            this.Bind(ViewModel.Email, EmailTextField.BindText());
+            this.Bind(ViewModel.ErrorMessage, ErrorLabel.BindText());
+            this.Bind(ViewModel.Password, PasswordTextField.BindText());
+            this.Bind(EmailTextField.Text().Select(Email.From), ViewModel.SetEmail);
+            this.Bind(PasswordTextField.Text().Select(Password.From), ViewModel.SetPassword);
+            this.Bind(ViewModel.IsLoading.Select(signupButtonTitle), SignupButton.BindAnimatedTitle());
```
khoi: Having this `IsLoading` thing stream, you can do one little extra nice thing
```
this.Bind(ViewModel.IsLoading, UIApplication.SharedApplication.BindNetworkActivityIndicatorVisible());
```

sofokli1: I gave that a try but:
- after signup the spinner remains there
- this is not widely used in the app so it is a bit inconsistent I only see it in the `Send feedback VC` and it would make way more sense to have it bind to the network layer and not each VC action.
- I think this spinner is not visible in iPhone X (at least I can't see it in this app or other) which depending on what the new generation of phones will look like it might not be needed at all in the end

(I do agree it is a nice feedback for the user to know that the app is networking)

amulware:
Note that we'll probably have to redo the login/signup UI a bit in the coming months anyways. The communication about what's going on, especially if something fails or takes a bit longer is really bad right now, and we'll have to take another look at that.

Toggl.Daneel/ViewControllers/SignupViewController.cs
khoi: yeah, the `IsLoading` has to emit false when everything is done for it to disappear. I don't know what Apple's done with the global loading indicator on iPhone X. freaking notch ⁉️

Toggl.Foundation.Tests/MvvmCross/ViewModels/SignupViewModelTests.cs
```diff
@@ -461,6 +462,71 @@ public void PassesTheCredentialsToLoginViewModelIfUserTriedToSignUpWithExistingE
...
+                ViewModel.Shake.Subscribe(observer);
+
+                var currentEmail = Email.From(email);
+                var currentPassword = Password.From(password);
+                var expectedShakeTargets = ShakeTargets.None;
+
+                if (!countryIsSelected)
+                {
+                    Api.Location.Get().Returns(Observable.Throw<ILocation>(new Exception()));
+                    expectedShakeTargets |= ShakeTargets.Country;
```
daividssilverio: Adding the same logic used in the SignupViewModel feels wrong.
Please simply add another parameter to your theory with the expected flags.
Only the if logic to call `Api.Location.Get().Returns(Observable.Throw<ILocation>(new Exception()));` seems unavoidable.

PR #3020, opened 2018-08-21
Fix glitch when scrolling in the calendar view

### LINKED ISSUE #3017: Glitchy calendar items when scrolling the calendar view
Sometimes (specially when you have multiple calendar items and they overlap, but there's no formula) calendar items start flickering as you scroll. Scrolling slowly helps one perceive that they simply disappear for a certain portion of the scrolling and then reappear. 

I tried tapping the position where the item would be to see if it was invisible or gone, and since nothing happened I can assume that it was not being displayed instead of being made invisible.

juanlaube:
## What's this?
Fixes a glitch when scrolling in the calendar view, caused by not considering the timezone when calculating the visible items in the collectionview.

### Relationships

Closes #3017

## How is it done?
It's done by converting the start and end offsets to the local time when calculating the index paths of visible items between the visible hours.

### Side effects
None.

## Review hints
It's a straightforward review.

## :squid: Permissions
Whenever


willsb:
Should we maybe consider returning all of the CalendarItems in LocalTime?

I ask because they are meant to be displayed for the user, they are ephemeral and they are never ever persisted.

juanlaube:
I think that's a good idea, but I'd do that in another PR or wait for things that depend on the start/end times to get merged.

PR #3023, opened 2018-08-21
Fix double events in rating view

### LINKED ISSUE #3001: Duplicate event tracking in rating view
## 🐛 Bug Report
As was noted in #2990, the dismiss/unsuccessful events of the rating view are tracked even when the user agrees to rate/send feedback.

This is because both the `PerformMainAction` and `Dismiss` methods are called from the view.

https://github.com/toggl/mobileapp/blob/b22198d08163d9dd0738b7aa9b19e748919eff5a/Toggl.Daneel/Views/Rating/RatingView.cs#L88-L89

## Expected behavior
Instead, the events in the `Dismiss` method should only be tracked if the user taps the dismiss button in the second step.

## Suggestions
The easiest is probably to add the dismissal navigation to the main action method as well, and remove the dismiss binding.

A renaming of the methods to be clearer, so that it doesn't get called accidentally might be nice too.

Also, maybe this should be refactored, so that it cannot be abused by the view. How about a check in both methods that prevents both from being called?

amulware:
## What's this?
This fixes unsuccessful-action events from being fired on the success of the second rating view step.

It also cleans up the related code a bit.

### Relationships
Closes #3001
The issue was first noticed here: 2990

## Why do we want this?
Because having double events makes statistical analysis more difficult.

## How is it done?
The functional changes are all in the first commit. I duplicated the navigation call to the second action method and removed the double binding.

The other two commits extract shared code into reused methods, which I think is nicer. If you think that's too much for this PR, I can always revert that and do a proper cleanup PR of this down the line though.

Other things that could be done would be throwing if both methods are called, but I'm not sure if that is necessary.

### Why not another way?
This seemed the easiest.

### Side effects
I did not duplicate the line `impressionSubject.OnNext(null);` from the Dismiss method, because I don't know what it does. Maybe @dshr or @arturs105 (git blamed you) can comment on that.

Tests pass without that line by the way, so I was considering just removing it

## Review hints
*I encourage review per commit*, that makes the changes very clear. Please double check that I did not mess up any of the logic since I don't know MVX very well at all. Similar with the line I did not duplicate mentioned above.

Note: I did not do a functional test. Someone else please do. kthx.

## :squid: Permissions
I'll do it.


amulware:
Darp, it derped. Still gonna wait for @dshr or @arturs105 to comment on the question above though.

Toggl.Foundation.MvvmCross/ViewModels/RatingViewModel.cs
```diff
@@ -32,6 +32,9 @@ public sealed class RatingViewModel : MvxViewModel
         private readonly BehaviorSubject<bool?> impressionSubject = new BehaviorSubject<bool?>(null);
         private readonly ISubject<bool> isFeedbackSuccessViewShowing = new Subject<bool>();
 
+        private bool impressionWasRegistered => impressionSubject.Value != null;
+        private bool impressionIsPositive => impressionSubject.Value.Value;
```
arturs105: Won't this throw a `NullreferenceException` if `impressionSubject.Value` is null?

dshr: Line 36 def will

amulware: Yes, but it is never called that way. Do you have a better way of doing this?

amulware:
@dshr @arturs105 I'd still like an explicit comment on the `impressionSubject.OnNext(null);` statement.

Toggl.Foundation.MvvmCross/ViewModels/RatingViewModel.cs
arturs105: `impressionSubject.Value?.Value ?? false;`

arturs105: Reason to change: potential `NullReferenceException` (better safe than sorry).

Another argument for change: if there is no impression yet (`impressionSubject.Value` is null), then the impression is definitely not positive, which means `impressionIsPositive` variable should be false (or something else), but it should definitely not throw.

Another option would be to use a nullable field: `bool? impressionIsPositive => impressionSubject.Value?.Value`

amulware: I don't see the exception as a problem. This is private code that shouldn't be used without understanding it, and it's used in a way where it specifically checks if an impression was left first.

If we make it nullable, we then have to again complicate the usage of this property, which I'm against.

If we can't agree on this, I rather revert the changes for now than make the code less readable.

arturs105: >This is private code that shouldn't be used without understanding it

Can we ensure that people, who make changes to these files know that the property might throw?

If you strongly disagree about using a default value, then i'd be fine with keeping it like it is now, if we added a comment above the property. Something like: `//❗️WARNING❗️ : this property might throw if held wrong, please ensure that impressionSubject.Value is not null`.

amulware: I was assuming people would read the code before changing it, yeah. But alright, I'll add that comment.

amulware:
@arturs105 I added a comment to the property, and as discussed internally, I removed the cleanup entirely since it's not consistent and serves no purpose for the time being.

dshr:
@amulware just tested this and while the rating view does hide on dismiss, it leaves a big empty space. Funnily enough, the `OnNext(null)` is not the culprit here.

amulware:
I'm not sure I follow. Does it work properly when the `OnNext(null)` is there, but not otherwise? Or does it leave the empty space in both cases?

dshr:
In both cases. We might've accidentally removed a constraint binding somewhere.

amulware:
Can you check if the same issue exists in 1.8.2?

amulware:
For the time being, I'll go ahead and squash this.

dshr:
screen shot 2018-08-22 at 14 47 17
It is happening on 1.8.2 indeed. Looks to me like it's a problem with the `ratingViewContainer`. I'll keep on looking.

PR #3025, opened 2018-08-21
Switch to non-localised strings

### LINKED ISSUE #2776: Longer button names break into multiple lines
## ⛑️ User Support
We use localized button names, which vary in length. Apparently, at least some of our buttons do not resize themselves accordingly and break the text into multiple lines.

This happens at least in the time popup, but possibly also other places.

## About the user

Version: Giskard/1.0 
Phone: samsung SM-J510FN 
OS: 6.0.1 LOLLIPOP_MR1
OS Language: German

## Expected behaviour
We should either not localize these buttons, or we should make their size adjust to their content.

This decision should be made by @pe0ny.

## To Reproduce
Set your phone to a language that has long words. The screenshot is in German. I'm not sure if the size of the screen matters as well.

screen shot 2018-07-31 at 11 54 25

dshr:
## What's this?
We were using localised strings in the TE Time edit views, when we shouldn't. Switched them to normal.

### Relationships
Closes #2776

## Why do we want this?
Consistency.

## How is it done?
By setting a proper `android.text`

### Why not another way?
Seems like the simplest option.

### Side effects
None

## Review hints
Set your system language to something else, and then try editing the TE time.

## :squid: Permissions
All yours!


Toggl.Giskard/Resources/layout/SelectTimeFragment.axml
```diff
@@ -34,7 +34,7 @@
             android:gravity="center"
             android:textAllCaps="true"
             android:textColor="@color/popupBlueBackground"
-            android:text="@android:string/cancel"
+            android:text="Cancel"
```
juanlaube: Instead of hardcoding the strings in the xml, please add a new non-localized string resource.

daividssilverio: Or, you can mark the strings with `<string name="stringResName" translatable="false">`.

daividssilverio: Oh... I see, we were using an `@android` resource. Gotta create a new one. :face_palm:

PR #3037, opened 2018-08-22
Prevent rating view from leaving blank space after dismissal

### LINKED ISSUE #3036: Rating view leaves blank space after dismissal
## 🐛 Bug Report
@dshr reported in #3023 that the rating view leaves behind an empty area when the user is done with it.

## Context
This happened specifically on the 1.9 release branch `zoranel`. We should check if it also happens in 1.8.2. It might well be a bug introduced with the main view rxfactor that's part of 1.9.

## Expected behavior
The rating view should not leave behind any empty space. The main view should look just like before.

## To reproduce
- hardcode the rating view to always show up
- make it show up
- step through the steps to make it disappear again
- observe empty space left behind

## Suggestions
As mentioned above, please check 1.8.2 to see if there is a difference in behaviour. That might not only help for debugging but might affect our decision for whether we can go ahead with A/B testing the rating view.

dshr:
## What's this?
A fix to the blank space problem we've had since 1.8.2. 

### Relationships
Closes #3036 

## Why do we want this?
So things look pretty when our users dismiss the rating view.

## How is it done?
~~By changing the UITableView header frame. They don't support AL, so this is the only way. We then make the UITableView re-render itself using `ReloadData()`.~~

This has been refactored. Now the TableHeader height is set on every ViewDidLayoutSubviews, which takes care both of showing and hiding the ratings view.

### Why not another way?
This seems like the simplest way to me.

### Side effects
None that I know of.

## Review hints
Make the review view always-on by:
1. Commenting out `MainViewModel.cs:292-309` and rewriting `RatingViewExperiment.cs:18` to `public IObservable<bool> RatingViewShouldBeVisible => Observable.Return(true).Delay(30)` to simulate the async fetching.
2. Run the app
3. Click on one of the first actions and then dismiss the rating view
4. .....
5. 🎉 

## :squid: Permissions
All yours!


khoi:
Dismissal looks good now. 
But when the rating view appears, there is a blank space at the bottom of the rating view
screen shot 2018-08-23 at 15 32 53

dshr:
@khoi how many TEs do you have on the main view?

PR #3043, opened 2018-08-22
Fix issue when dragging up

### LINKED ISSUE #3030: Tap and drag upwards logic has flaws
When you tap and drag your thumb down to create a time entry, everything works flawlessly. However, when doing the same and dragging your thumb upwards, it miscalculates overlapping CalendarItems.

Video evidence [here](https://toggl.slack.com/archives/C02APMYTM/p1534934724000100)

juanlaube:
## What's this?
Fixes the issue when dragging an item up caused by some flaws in the overlapping group logic.

### Relationships

Closes #3030

## How is it done?
Until now, when a calendar item was updated by dragging up/down, the item was removed and inserted in order, this caused the index paths to change. With this change, the index paths are fixed, solving the problems with the layout calculations and changing index paths.

### Side effects
None.

## Review hints
It's a straightforward review, but please test.

## :squid: Permissions
Whenever.


Toggl.Daneel/ViewSources/CalendarCollectionViewSource.cs
```diff
@@ -239,32 +239,44 @@ private IList<CalendarCollectionViewItemLayoutAttributes> calculateLayoutAttribu
...
-                        else
-                            buckets.Add(new List<CalendarItem>() { item });
-                    }
-
-                    return buckets;
-                })
-                .Select(bucket => bucket.Select((item, index) => attributesForItem(item, bucket.Count, index)))
+            var pairs = calendarItems
+                .Select((item, index) => (item, index))
+                .OrderBy(pair => pair.Item1.StartTime)
```
arturs105: You could use named tuples here: `pair.item.StartTime`.

Toggl.Daneel/ViewSources/CalendarCollectionViewSource.cs
```diff
@@ -239,32 +239,44 @@ private IList<CalendarCollectionViewItemLayoutAttributes> calculateLayoutAttribu
...
+                .Aggregate(initialValue, groupOverlappingItems)
+                .Select(toLayoutAttributes)
                 .SelectMany(CommonFunctions.Identity)
+                .OrderBy(pair => pair.Item2)
+                .Select(pair => pair.Item1)
                 .ToList();
 
             return attributes;
+
+            List<List<(CalendarItem, int)>> groupOverlappingItems(List<List<(CalendarItem, int)>> previous, (CalendarItem, int) pair)
```
arturs105: You should use named tuples here as well.

PR #3058, opened 2018-08-23
Calendar supplementary hours

### LINKED ISSUE #3034: Labels in calendar view need some tweaking
- [x] Scrolling all the way up should comfortably show midnight. Right now midnight is capped and only visible when you overscroll
- [ ] We need to pad the times from midnight to 09:00 with a zero when in 24-hour mode
image

- [x] We should not show more times after 23:00
- [ ] When in ~wrong~ 12-hour mode, we don't need to prepend a `:00` to every hour nor pad the hours from midnight to 9am with a zero
image

For reference, you can see what the view should look like in #2874

juanlaube:
## What's this?
Fixes various issues with the hours supplementary views:
- First hour was cut
- Was showing hours after midnight
- Hour formatting

### Relationships

Closes #3034 

## How is it done?

The first two problems are solved in the first commit by adding some insets and calculating correctly how many hours are displayed.
The last problem is fixed in the second commit by using specific formats for the hour labels.

## :squid: Permissions
Whenever.


Toggl.Daneel/Views/Calendar/CalendarCollectionViewLayout.cs
```diff
@@ -128,13 +128,23 @@ private IEnumerable<NSIndexPath> indexPathsForVisibleItemsInRect(CGRect rect)
...
+            var minHour = clamp((int)Math.Floor(rect.GetMinY() / hourHeight), 0, 24);
+            var maxHour = clamp((int)Math.Floor(rect.GetMaxY() / hourHeight), 0, 25);
 
             return Enumerable
-                .Range(minHour, maxHour)
+                .Range(minHour, maxHour - minHour)
                 .Select(hour => NSIndexPath.FromItemSection(hour, 0))
                 .ToArray();
+
+            int clamp(int value, int min, int max)
```
willsb: We have a clamp extension method

PR #3059, opened 2018-08-24
👩‍🏫 Siri project setup

### LINKED ISSUE #3042: Create first intent
We need to create the first intent, it'll be the most complex as it requires the creation of the intent definition file and setup everything for this to work in Xamarin.

Let's start with a simple one, for example "Stop current entry" (name not final). It'll return a response with the description of the entry and the time it was running if it succeeds. And an appropriate error message if it fails.

The intent should be donated every time a time entry is stopped. My idea is this should be done in Foundation but through an iOS service, or a service with an empty (for now) android implementation, but I'm open to suggestions.

This, of course, needs the beta version of Xamarin to work.

I suggest Apple docs are read first, specifically the Soup Chef example shows how to create Intents in a native app: https://developer.apple.com/documentation/sirikit/soup_chef_accelerating_app_interactions_with_shortcuts

And also take a look at Xamarin version, which adds some overhead: https://developer.xamarin.com/samples/monotouch/ios12/SoupChef/

khoi:
## What's this?
This PR contains some initial configurations needed to interact with the "Custom Intent Definition" file. It also contains the `StopTimerIntent`, but it is merely a placeholder and subject to change.

There is no intent handling logic in this PR.

**Note: Build is broken since Bitrise doesn't support iOS 12 yet.**

### Relationships
Closes #3042 

## How is it done?
The process to use Custom Intents in Xamarin can be summarized into these steps

1. Have an Objective-C Xcode project to edit the Intent Definition file. 

It's `Toggl.Daneel.IntentCodeGen` in our case. Building this will generate `.h` and `.m` files for 
each intent.

Locating the generated file is quite tricky. Check `AppDelegate.m` for instruction.

The output of this steps are: `StopTimeEntryIntent.{h,m}` 

2. Have another Objective-C Xcode project to compile all the generated Intent files above to a static library 

`Toggl.Daneel.IntentStaticLib` is the one. The generated `.h` and `.m` files above are copied to this project 

This also contains a `Makefile` in which will generate the fat binary, and also use `sharpie` to generate the C# binding files that is going to be used in the next step.

*Note*: The generated C# files need some human alteration in order to be used. 🤯

Run `make` in the folder will produce the output:

- `libToggl.Daneel.IntentStaticLib.a` the fat library file
- `/bo/ApiDefinitions.cs` and `StructsAndEnums.cs` generate bridge C# code

3. A new `Bindings Library` project named `Toggl.Daneel.IntentBinding` is added to bind the static lib above to our Xamarin app

- The fat `libToggl.Daneel.IntentStaticLib.a` is added as a native references
- The content of `ApiDefinition.cs` and `StructsAndEnums.cs` are copied here.

That's it, we now have our Custom Intents ready to use.

Some references: 

- https://docs.microsoft.com/en-us/xamarin/ios/platform/binding-objective-c/walkthrough?tabs=vsmac#creating-a-static-library
- https://docs.microsoft.com/en-us/xamarin/ios/platform/introduction-to-ios12/siri-shortcuts
- https://developer.xamarin.com/samples/monotouch/ios12/SoupChef/

### Why not another way?
This follows the documentation from Xamarin.

But the process for adding new intent is quite troublesome, there are many manual steps involved. I'd love to hear feedback on how we could improve this.

### Side effects
Not that I know of

## Review hints
In order to build it

- Have Xcode 10 setup and 
```
In Xcode, choose Xcode > Preferences > Locations and set Command Line Tools to the most current Xcode 10 installation available on your system
```
- Install [Sharpie](https://docs.microsoft.com/xamarin/cross-platform/macios/binding/objective-sharpie/get-started?context=xamarin/mac#installing-objective-sharpie)
- Go to `Toggl.Daneel.IntentStaticLib` and run `make`
- Check if the project can be built

Review the PR commit by commit would probably be easier.

## :squid: Permissions
Me


Toggl.Daneel.IntentBinding/ApiDefinition.cs
```diff
@@ -0,0 +1,58 @@
+﻿using System;
+using Foundation;
+using Intents;
+using ObjCRuntime;
+
+namespace Toggl.Daneel.Intents
+{
+    // @interface StopTimerIntent : INIntent
+//    [Watch(5, 0), iOS(12, 0)]
```
khoi: In the Soup Chef Xamarin sample project, they commented these out. WHYYYYYYYYYYYYY

Toggl.Daneel.IntentBinding/StructsAndEnums.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System;
+using ObjCRuntime;
+
+namespace Toggl.Daneel.Intents
+{
+//    [Watch (5,0), iOS (12,0)]
+    [Native]
+    public enum StopTimerIntentResponseCode : long
```
khoi: `sharpie` generated the file as `public enum StopTimerIntentResponseCode : uint` which can't be compiled. I followed the `SoupChef` sample app and changed it to `long` 🤷🏻‍♂️

Toggl.Daneel.IntentBinding/ApiDefinition.cs
khoi: Found the answer for it 
```
Both ApiDefinitions.cs and StructsAndEnums.cs contain attributes such as [Watch (5,0), iOS (12,0)]. These attributes, generated by Objective Sharpie, have been commented out since they aren't necessary for this project.
```
https://docs.microsoft.com/en-us/xamarin/ios/platform/introduction-to-ios12/siri-shortcuts

willsb: Why not remove these instead of commenting them out?

Toggl.Daneel.IntentCodeGen/.gitignore
```diff
@@ -0,0 +1,23 @@
+# Xcode
```
willsb: Please add these to the main .gitignore; there's no reason to keep both

Toggl.Daneel.IntentCodeGen/Toggl.Daneel.IntentCodeGen.xcodeproj/project.xcworkspace/xcshareddata/IDEWorkspaceChecks.plist
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>IDEDidComputeMac32BitWarning</key>
```
willsb: new project who dis?

Toggl.Daneel.IntentStaticLib/.gitignore
```diff
@@ -0,0 +1,23 @@
+# Xcode
```
willsb: Similar to the above

khoi:
>Do we need that .a file to be versioned?

I have thought about this and I think it's best that we add it into the version control. The reasoning is that newcomers don't need to care about how all of this work. They could expect just to clone the project and run it normally.

Unless we can somehow add `make` as one of the build step?

Toggl.Daneel.IntentCodeGen/Toggl.Daneel.IntentCodeGen.xcodeproj/project.xcworkspace/xcshareddata/IDEWorkspaceChecks.plist
khoi: i dont get dis ???

Odrakir:
I managed to make it compile. So I guess this is ok. Although we will do further testing when actually trying out the intents.

A guide on the steps to add new intents would be very useful.

PR #3070, opened 2018-08-24
Fix duplicate workspace name crash

### LINKED ISSUE #3060: Duplicate workspace names crash app on start
## 💥 Crash Report
The app crashes on startup if the user has two workspaces with the same name. This happens because the ReportsViewModel expects names to be unique here:

https://github.com/toggl/mobileapp/blob/5dc6066acde49f16d853cc930829ae1f3d28390e/Toggl.Foundation.MvvmCross/ViewModels/ReportsViewModel.cs#L141-L144

### 👥 Impact
Potentially a bunch of users, since duplicate workspaces names can easily happen with the default name being based on the user's email address.

### 🛰 Link
https://appcenter.ms/orgs/Toggl/apps/Toggl-iOS/crashes/groups/f43a6cced0f4c877ede22bd81d946dfedb41cf2a?appVersion=1.9&period=last7Days

`System.ArgumentException: An item with the same key has already been added. Key: <REDACTED>`
```cs
ThrowHelper.ThrowAddingDuplicateWithKeyArgumentException (System.Object key)
Dictionary`2[TKey,TValue].TryInsert (TKey key, TValue value, System.Collections.Generic.InsertionBehavior behavior)
Dictionary`2[TKey,TValue].Add (TKey key, TValue value)
ToDictionary`3+_[TSource,TKey,TElement].OnNext (TSource value)
ExceptionHelpers.ThrowIfNotNull (System.Exception exception)
AsyncSubject`1[T].GetResult ()
ReportsViewModel+<Initialize>d__73.MoveNext ()
TaskAwaiter.ThrowForNonSuccess (System.Threading.Tasks.Task task)
TaskAwaiter.HandleNonSuccessAndDebuggerNotification (System.Threading.Tasks.Task task)
TaskAwaiter.ValidateEnd (System.Threading.Tasks.Task task)
TaskAwaiter.GetResult ()
MainTabBarViewModel+<>c+<<Initialize>b__7_0>d.MoveNext ()
AsyncMethodBuilderCore+<>c.<ThrowAsync>b__6_0 (System.Object state)
UIKitSynchronizationContext+<>c__DisplayClass1_0.<Post>b__0 ()
NSAsyncActionDispatcher.Apply ()
```

## ❓Other questions

What this also makes me wonder:

1. Apparently, the reports are initialised on app start - why? Does it also fetch reports right away? It makes no sense to me to initialise them before first viewing.
2. Is the initialisation called more than once? If so, does that make sense? If not, how will it know about workspaces being added/removed?

amulware:
## What's this?
This removes the unnecessary use of dictionaries from the dialog service and reports view model.

This caused a crash because workspace names do not have to be unique.

### Relationships
Closes #3060

## Why do we want this?
Because crashing is no good. In fact, since reports are right now initialised right on app start, once this would happen the app would never open again until reinstall.

## How is it done?
Pretty simply, check the code. Note how I expose ICollection, which can do everything required (it can be enumerated by the dialog and it has a Count property used in other places).

### Why not another way?
The dictionary really isn't needed here and should have never been used. Note that I put a bit of effort into getting the publicly exposed collection to be guaranteed to be readonly. I like it that way, though I wish there was an easier way. Observables make it hard. Maybe a custom rxtension would be nice for this.

Also, I am of the opinion that the workspaces collection exposed in the view model doesn't have to be - it's certainly not used right now. So maybe it shouldn't be. It is also not tested, and it doesn't update appropriately when workspaces change, so that needs improvement. I'll create a follow-up issue for that.

### Side effects
None that I can think of.

## Review hints
You can look at individual commits, but you don't have to.

## :squid: Permissions
Me.


Toggl.Foundation.MvvmCross/ViewModels/ReportsViewModel.cs
```diff
@@ -141,7 +141,9 @@ public override async Task Initialize()
             Workspaces = await dataSource.Workspaces
                 .GetAll()
                 .SelectMany(CommonFunctions.Identity)
-                .ToDictionary(ws => ws.Name, ws => ws);
+                .Select(ws => (ws.Name, ws))
+                .ToList()
+                .Select(list => list.ToList().AsReadOnly());
```
Odrakir: Another way of doing this, not sure if more or less clear:

```c#
Workspaces = await dataSource.Workspaces
    .GetAll()
    .Select(list => list
        .Select(ws => (ws.Name, ws))
        .ToList()
        .AsReadOnly()
    );
```

Everything inside the `Select` could be put in a private method.

khoi:
I can no longer see the Select Workspace button in the Daneel Report View.

Toggl.Foundation.MvvmCross/ViewModels/ReportsViewModel.cs
```diff
@@ -97,7 +97,7 @@ public bool IsCurrentWeek
 
         public string WorkspaceName { get; private set; }
 
-        public IDictionary<string, IThreadSafeWorkspace> Workspaces { get; private set; }
+        public ICollection<(string ItemName, IThreadSafeWorkspace Item)> Workspaces { get; private set; }
```
amulware: @khoi It looks like this change breaks bindings somehow. I do not know why. I don't understand bindings very well. I'll try refactoring it a little bit.

amulware: Oh, I found the issue, it's a type issue in the actual binding in the VC. But of course it doesn't crash, it just fails silently.

amulware:
I did as @Odrakir suggested. This looks much nicer, thanks!

And I also fixed the binding for you @khoi. Please test again.

khoi:
Working well now 👍

amulware:
I'll take that as approval and will squash as soon as the tests pass.

PR #3075, opened 2018-08-24
Add notification service interface

### LINKED ISSUE #3066: Add the interface for the notification service
We need a simple notification service that allows us to:

- Schedule a notification
- Unschedule a notification based on some id
- Get a list of the scheduled notifications

This will also need a model for the notification (Title, Description and ID).

⚠️ Do not implement this service yet. This will happen in another PR ⚠️

arturs105:
## What's this?
This adds just the interface and `Notification` struct.

### Relationships

Closes #3066

## :squid: Permissions
Whenever, wherever.


Toggl.Foundation/Services/INotificationService.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System.Collections.Generic;
+using Toggl.Multivac;
+
+namespace Toggl.Foundation.Services
+{
+    public interface INotificationService
+    {
+        void UnscheduleNotification(long id);
+
+        Notification SchedudleNotification(string title, string description = null);
```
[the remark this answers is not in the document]
willsb: I think we should pass a `Notification` object to this method, because we want the `Id` property to be provided by us. Also, it should return a bool to indicate if the operation succeeded or not.

Toggl.Foundation/Services/INotificationService.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System.Collections.Generic;
+using Toggl.Multivac;
+
+namespace Toggl.Foundation.Services
+{
+    public interface INotificationService
+    {
+        void UnscheduleNotification(long id);
```
willsb: Like the method below, this one should return a bool to indicate success or failure

Toggl.Foundation/Services/INotificationService.cs
```diff
@@ -0,0 +1,14 @@
...
+
+        Notification SchedudleNotification(string title, string description = null);
+
+        List<Notification> ScheduledNotifications();
```
willsb: `IImmutableList<Notification>` for consistency with the whole Immutable trend.

PR #3076, opened 2018-08-24
Add idle/running notification settings

### LINKED ISSUE #3054: Add idle/running notification settings
Two checkboxes/toggles in the settings. Maybe needs a whole new notification section as well.

The two settings are stored locally and cleared on logout.

Both default to DONT SHOW.

dshr:
## What's this?
We're adding the running timer/stopped timer notifications from the old app. This is the first step, where we add the settings checkboxes for them.

### Relationships
Closes #3054

## Why do we want this?
So the users can turn them on and off.

## How is it done?
By adding the appropriate fields to the `SettingsStorage` in `PrimeRadiant`

### Why not another way?
This is the simplest way tbh.

### Side effects
None that I've noticed.

## Review hints
Go to settings, click the switches around; close the app, open it — they should be there; log out — the settings should be forgotten.

## :squid: Permissions
All yours!

### DIFF
--- a/Toggl.PrimeRadiant/Settings/IUserPreferences.cs
+++ b/Toggl.PrimeRadiant/Settings/IUserPreferences.cs
@@ -6,11 +6,27 @@ namespace Toggl.PrimeRadiant.Settings
     {
         IObservable<bool> IsManualModeEnabledObservable { get; }
 
+        IObservable<bool> AreRunningTimerNotificationsEnabledObservable { get; }
+
+        IObservable<bool> AreStoppedTimerNotificationsEnabledObservable { get; }
+
         bool IsManualModeEnabled { get; }
+
+        bool AreRunningTimerNotificationsEnabled { get; }
+
+        bool AreStoppedTimerNotificationsEnabled { get; }
 
         void EnableManualMode();
 
         void EnableTimerMode();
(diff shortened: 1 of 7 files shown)


daividssilverio:
I guess we can create another branch for the whole notifications's feature. We don't want just the settings part, or we can comment or hide the views out.

Toggl.PrimeRadiant/Settings/IUserPreferences.cs
```diff
@@ -6,12 +6,28 @@ public interface IUserPreferences
...
 
+        bool AreRunningTimerNotificationsEnabled { get; }
+
+        bool AreStoppedTimerNotificationsEnabled { get; }
+
         void EnableManualMode();
 
         void EnableTimerMode();
 
+        void EnableRunningTimerNotifications();
```
kornelijepetak: Why not do `SetRunningTimerNotifications(bool state)`?

You have reasonable amount of code repetition at the definition site that differ only in that boolean value.

kornelijepetak: `Toggle*` functions in `SettingsViewModel` would also benefit in clarity and conciseness.

PR #3077, opened 2018-08-24
Display blue time labels when in edit mode

### LINKED ISSUE #3035: Display blue time labels when in edit mode
During tap & drag time entry creation and during duration edit mode, as described in #2879 and #2878, we need to show an additional view, similar to the ones that display the time, on the left side so the users know the start and end times of the view they are editing.

These are probably more supplementary views and @juanlaube can help with his collection view knowledge

juanlaube:
## What's this?
Adds the blue labels in the left side to display the start/end time of the item being edited.

### Relationships

Closes #3035

## How is it done?
It's more collection view magic, but the important things are a new kind of supplementary view and the usage of invalidation context, to invalidate pieces of the layout.

### Side effects
A good one. Now cells are not recycled when updated, instead their layout is invalidated.

## Review hints
Test and try to break it.

## :squid: Permissions
Whenever.


Toggl.Daneel/Views/Calendar/CalendarCollectionViewLayout.cs
```diff
@@ -176,5 +232,10 @@ private CGRect frameForCurrentTime()
 
             return new CGRect(x, y, width, height);
         }
+
+        private UICollectionViewLayoutInvalidationContext invalidationContextForEditingMode()
+        {
+            throw new NotImplementedException();
```
willsb: What's this?

juanlaube: something I tried and forgot to delete 😂

dshr:
Looks and works great! Notice one very tiny thing — the BG of the blue time label sometimes overlaps with the gray time labels and cuts them off a bit: 
screen shot 2018-08-27 at 12 57 40
Obviously we have to have the white bg so the labels don't overlap, but maybe we can fake a blur using the CGLayer shadow with it's colour set to white, so at least the overlap doesn't feel this abrupt? I'd be more than happy to add that myself.

juanlaube:
Good catch @dshr!

PR #3086, opened 2018-08-27
Donating first Siri intent

### LINKED ISSUE #3069: Create first intent
Let's start with a simple one, for example "Stop current entry" (name not final). It'll return a response with the description of the entry and the time it was running if it succeeds. And an appropriate error message if it fails.

The intent should be donated every time a time entry is stopped. My idea is this should be done in Foundation but through an iOS service, or a service with an empty (for now) android implementation, but I'm open to suggestions.

khoi:
## What's this?

- Donates the `Stop current running time entry` intent to Siri. 
- Add documentation on how to modify intent definition file.

### Relationships
Closes #3069 

## Why do we want this?
Donating the action so later on Siri can intelligently suggest the user to stop a time entry.

## How is it done?
A new service `IIntentDonationService` is added to Foundation. The implementation resides in Daneel, also a dummy one is added for Giskard so the app doesn't crash.

### Why not another way?
Dummy service seems to be the easiest way. 

### Side effects
Not that I know of

## Review hints

Since CI doesn't work for this, ensure you have passing unit tests.

- Enable `Display Donations on Lock screen` in `Developer` setting of the Simulator
- Start and stop a time entry
- Check the home screen to see whether the donation is listed there.

## :squid: Permissions
me or @Odrakir


Toggl.Foundation/IIntentDonationService.cs
```diff
@@ -0,0 +1,10 @@
+using Toggl.Foundation.Models;
```
Odrakir: Can you move this file to the `Services` folder?

Odrakir:
@khoi About the error you are getting, I think it might have something to do with the parameters. Please try with this configuration for the intent:

captura de pantalla 2018-08-27 a las 18 09 43
captura de pantalla 2018-08-27 a las 18 09 50

Toggl.Daneel/Services/IntentDonationService.cs
```diff
@@ -0,0 +1,31 @@
...
+using Toggl.Multivac.Models;
+
+namespace Toggl.Daneel.Services
+{
+    public class IntentDonationService: IIntentDonationService
+    {
+        public void DonateStopCurrentTimeEntry()
+        {
+            var intent = new StopTimerIntent();
+            intent.SuggestedInvocationPhrase = "Stop timer.";
```
Odrakir: I think we should drop the "." here.

PR #3098, opened 2018-08-27
Check for permission before reading calendar events

### LINKED ISSUE #3080: Check for permission before reading calendar events
In the [`BaseCalendarService`](https://github.com/toggl/mobileapp/blob/calendar-integration/Toggl.Foundation.MvvmCross/Services/BaseCalendarService.cs) we are checking for permissions (and throwing the adequate exceptions) before reading the user calendars.  We should do the same for the calendar events.

This change also requires changing the [interactor](https://github.com/toggl/mobileapp/blob/calendar-integration/Toggl.Foundation/Interactors/Calendar/GetCalendarItemsForDateInteractor.cs) to handle the error and return an empty list as the list of events so that everything works without permissions.

arturs105:
### Relationships

Closes #3080 

## :squid: Permissions
Whenever


Toggl.Foundation.MvvmCross/Services/BaseCalendarService.cs
```diff
@@ -33,10 +34,31 @@ public IObservable<IEnumerable<UserCalendar>> GetUserCalendars()
...
+            {
+                var isAuthorized = permissionsService.CalendarPermissionGranted;
+                if (!isAuthorized)
+                {
+                    return Observable.Throw<IEnumerable<CalendarItem>>(
+                        new NotAuthorizedException("You don't have permission to access calendars")
+                    );
+                }
+
+                return Observable.Return(NativeTetEventsInRange(start, end));
```
sofokli1: what does `Tet` mean?

PR #3099, opened 2018-08-28
Shared private storage

### LINKED ISSUE #3081: Set app state in shared storage
In order to keep the extension aware of the state of the app we will set a shared the following properties in the NSUserDefaults shared with the App Group

- api token
- timestamp of last sync
- running entry id (or nil if none)

The api token will be used to construct the http requests, the timestamp of the last sync could be used to force the user to open the app in case it's been "too long" since the last sync.

The id of the running time entry could be used to avoid having to do an extra request to get the latest entries.

The code that manages this data should be shared between the app and the extension, as both will need read/write access.

Odrakir:
## What's this?
This adds a service that enables saving data in a private shared location, so apps in the same app group can share some data between them. This is needed for the iOS app to communicate some info with the Siri Extension and back.

Closes #3081 

## How is it done?
I created the service interface in Foundation and the implementation in Daneel, there's also a dummy implementation in Giskard. This just makes things easier from the xplatform side of things, and maybe in the future it'll come useful for Giskard as well.

## :squid: Permissions
Do it.


Odrakir:
@willsb Suggested not using a service (with the dummy version for Android) for this, but his alternative (subclassing `LocationManager` in Daneel) doesn't seem possible for me.

Any other ideas are welcome, otherwise this will be merged.

Odrakir:
@sofokli1 Cake should work and android build should pass tests. But iOS won't because the stack doesn't work on bitrise yet.

You could download and test locally with iOS 12 sdk, though.

PR #3103, opened 2018-08-28
iOS Notification Service

### LINKED ISSUE #3078: Implement the schedule service in iOS
Once the interface from #3075 is merged, we need to implement it in the iOS layer, mock it on the Droid layer and register it during app startup

willsb:
## What's this?
This implements the notification service on iOS. It refactors some parts of the codebase to make everything fit in nicely.

### Relationships
Closes #3078

## How is it done?
Basically a blast from my past where I revisit the [DefaultNotificationService](https://github.com/toggl/superday/blob/develop/teferi/Services/Implementations/Interaction/DefaultNotificationService.swift) and reimplement it in C#.

### Why not another way?
:superday:

### Side effects
This does not care for the modifications needed in the .csproj to make our app aware of permissions. This is to be done in another issue, where we change the onboarding to request permissions.

## Review hints
Read docs, check the existing and working superday code and compare it with this implementation

## :squid: Permissions
Only me


sofokli1:
when i do a fresh install > go to calendar > don't allow > go to settings and allow > reopen the app (closed by the OS) > it does not ask me to give access to any calendar. so I do not see any event from the calendar in the app. I s this the correct behaviour?

PR #3105, opened 2018-08-29
Siri show report intent

### LINKED ISSUE #3064: Create show reports intents
We can start by creating one to just open the reports tab, and then create different ones for last week, last month... views.

khoi:
📊 Donating intents when users open a report page

## What's this?
This donates the intents to Sirikit when the user interacts with the report.

### Relationships

Closes #3064

## Why do we want this?
So Siri can intelligently predict and suggest shortcuts to users.

## How is it done?
A new intent is added to the intent definition file. And the interactions with the report is donated in Daneel

### Side effects
🤷🏻‍♂️

## Review hints

- Open reports
- Select various report durations
- See if the donations are listed in home screen (must enable Developer settings)

## :squid: Permissions
Go for it


Toggl.Daneel/Services/IntentDonationService.cs
```diff
@@ -7,25 +8,75 @@
...
+            var intent = new ShowReportPeriodIntent();
+            switch (period)
             {
-                if (!(error is null))
-                {
-                    Console.WriteLine($"Interaction donation failed: {error}");
-                }
-                else
+                case DonationReportPeriod.Today:
+                    intent.Period = ShowReportPeriodReportPeriod.Today;
```
khoi: This class is generated by Sharpie :( I have no control over how it's named. So it's a little bit silly

PR #3107, opened 2018-08-29
Update to RXFactoring of Main Log

### LINKED ISSUE #2845: Rxfactor the main log on giskard
## What is this issue about?
🚧 Rxfactor the main log on Android!

## Why?
We are slowly moving away from Mvx bindings, refactoring the app using Rx, and this is one next step in that direction. 
Also, this should make changes on the main log easier for people with Android and just plain Rx backgrounds.

More specifically, this should make #2541 easier to implement.

## This issue blocks
#2541 
#2665

## This issue affects
1. The main time entry log
2. Suggestions on the main log
3. Onboarding tooltips and animations implementations that depend on the MainRecyclerView and its adapter.
4. The running TE card
5. The FAB to start a TE
6. The swipe to refresh component on the main screen
7. Possibly something else I don't know about.

## Expected behaviour
Everything should work the way it worked before after the refactoring finishes.

## Related PRs
#2741 introduced the `GroupedOrderedCollection.cs` that can be used to rewrite the `SegmentedRecyclerAdapter`.
#2743 is the equivalent on Daneel. @Odrakir was kind enough to introduce `Obsolete` annotations and they should help making the changes on Giskard as well. 

## Thinks to remember
The `SegmentedRecyclerAdapter` is used on the MainRecyclerView, its replacement will have to be well thought. 
Pay attention to adapter updates (adapter `notify*Changed` methods) as they were causing a lot of crashes recently.

kornelijepetak:
## What's this?

* Fixes most of the changes requested in #2971 
* Fixes the misbehaving ♻️ indicator
* Small refactorings

### Relationships
Based upon #2971 `giskard-rx-main-log`

Closes #2845 
(but let this be merged into #2971 and then from there into `develop`)

## Why do we want this?
So we can finally release 1.4

## How is it done?
By replacing some MVX bindings with RX bindings, by 🔥, and pure willpower.

### Why not another way?
I dunno. This is the only way I know 😄 

### Side effects
* The undo feature depends on delayed deletion of the time entry. With this PR, if an entry is deleted by swiping left, it will not show undo, but the TE will be deleted only after 5 seconds. That means that if the app is killed during those 5 seconds, no 🔥for you!

## Review hints
Commit by commit.

## :squid: Permissions
@daividssilverio , @dshr or me. Also @amulware. 
After thorough review.


Toggl.Giskard/Resources/layout/MainActivity.axml
```diff
@@ -7,7 +7,6 @@
     android:layout_height="match_parent">
     <mvvmcross.droid.support.v4.MvxSwipeRefreshLayout
```
daividssilverio: Can't you just use the vanilla RefreshLayout now?

Toggl.Giskard/Activities/MainActivity.cs
```diff
@@ -56,40 +56,59 @@ protected override void OnCreate(Bundle bundle)
...
         }
 
         private void updateRecyclerViewPadding(bool isRunning)
         {
             var newPadding = isRunning ? 104.DpToPixels(this) : 70.DpToPixels(this);
-            mainRecyclerView.Post(() =>
-            {
-                mainRecyclerView.SetPadding(0, 0, 0, newPadding);
-            });
+            mainRecyclerView.SetPadding(0, 0, 0, newPadding);
```
daividssilverio: I'll approve the PR as soon as I test this. It was breaking when you stop/start time entries at the bottom of the main log list.

dshr:
Looks and works good, but I have to second @daividssilverio's point — if we're going Rx, we should switch to the vanilla SwipeRefreshView

Toggl.Giskard/Activities/MainActivity.cs
dshr: Not anymore though, works flawlessly.

kornelijepetak: Well, it's observed on UI thread, so should be fine. @daividssilverio can you confirm?

kornelijepetak:
Yes, indeed. I will fix remove the MVX version.

PR #3110, opened 2018-08-29
Add ultrawave to the Siri Extension

### LINKED ISSUE #3082: Add Ultrawave to the extension
We need to make requests to the Toggl API from the extension. If possible, we should reuse Ultrawave for this. Using the API token shared in the App group by the extension we should be able to create the appropriate requests.

Needed properties, like the API token, will come from the private shared storage implemented for #3081

Odrakir:
## What's this?
Adds Ultrawave to the extension and instantiates it to make sure it works. It doesn't do anything with it yet, that's coming in the next PR.

Closes #3082 

## Why do we want this?
We need to make requests to the Toggl API.

## How is it done?
We import the package and use the `api-token` in the shared storage to instantiate the Toggl API.

## Review hints
I'm not sure about the `UserAgent`, what's that for?

## :squid: Permissions
Please do.


amulware:
The user agent is us telling our api what app is using it. Let's make sure we use something meaningful there. I'd say... `Daneel/VERSION.extension` maybe (the app itself uses `Daneel/VERSION`)? Thoughts?

Odrakir:
But those are two different things, right? The name and the version, I'm setting the version from the `info.plist` already.

amulware:
Yes true. But it would be good if the extension was sort of grouped under the overall app client. I think adding the '.extension' bit to the version is the cleanest. Backend often uses filters like 'Daneel/*' for things. Otherwise we could also go with `Daneel-SiriExtension/..` maybe... but that feels less clean to me.

Odrakir:
`Daneel.SiriExtension`?

amulware:
Nope. If the API accepts it, let's go with `Daneel` and append the siri bit to the version. Otherwise our internal stats will be messed up, they specifically group user agents by the string before the slash, and we want data from the extension to show up with the rest.

PR #3111, opened 2018-08-29
Editing time constraints

### LINKED ISSUE #3097: Weird behaviour when creating a time entry
When dragging to create a new entry, if I start dragging at 1200 and drag down, the created TE spans from 12:00-X,. When dragging upwards it spans from X-12:15. That's inconsistent and should be fixed.

juanlaube:
## What's this?
Prevents creating or editing time entries before 0:00 and after 23:59

### Relationships

Closes #3097

## :squid: Permissions
Whenever


willsb:
I fail to see how this closes the issue

juanlaube:
You're right, this is what happens when I write code without my coffee 😅 The first commits solves a different problem, it was possible until now create or edit a TE before 0:00 and after 23:59. The bug from the issue is solved in the last commit.

PR #3119, opened 2018-08-30
Implement stop intent

### LINKED ISSUE #3083: Implement stop entry intent
This is here for discussion, I want @amulware to give his opinion on some points.

The entry should be stopped from the extension, this should be the logic:
- Check if there's an api token and last sync timestamp in the shared space, other wise redirect the user to the app.
- Check the last sync time, if it's been "too long", redirect to the app. (What's too long?)
- Get the latest entries from the API endpoint, if error redirect to app.
- Compare time of last sync with start time of latest entry:
  - If there's a server entry running
    - If the time of last sync >= last entry start time -> send stop request. Set a request for sync timestamp in the shared space.
    - If the time of last sync < last entry start time -> ... What should we do here? Error? Stop it anyway?
  - If there is not a server entry running.
    - If the time of last sync >= last entry start time -> Error: there's no time entry running.
    - If the time of last sync < last entry start time -> Error: there's no time entry running (And advice the user to open the app?)

If the intent is successful Siri should respond with: "Time entry stopped, you've been {TE_DESCRIPTION} for {TE_DURATION}" 

If there's an error Siri should respond with the appropriate error and opening the app if necessary.

When the user goes back to the app, it should check for the request for sync timestamp and see if it's newer than the last sync, if it is, it should try to sync.

Odrakir:
## What's this?
Adds the implementation of the Stop Intent to the extension.

Closes #3083 

## Why do we want this?
This enabled the extension to stop the currently running time entry when the stop intent Siri shortcut gets executed.

## How is it done?
Through the extension using ultra wave to stop the entry in the server.

## Review hints
There are two things missing here:
- The app should sync after this works successfully, there's another issue for that
- Siri won't say the custom messages due to a bug in Xamarin, we don't know when that will be fixed, so we will implement some UI to at least show something meaningful.

## :squid: Permissions
Be my guest.


Toggl.Daneel.SiriExtension/StopTimerIntentHandler.cs
```diff
@@ -1,18 +1,122 @@
...
+                        var response = new StopTimerIntentResponse(StopTimerIntentResponseCode.Success, null);
+                        completion(response);
+                    },
+                    exception =>
+                    {
+                        completion(responseFromException(exception));
+                    });
+        }
+
+        private TogglApi getTogglAPI()
```
khoi: this can be abstracted out to other class, so other intent handlers can use it. Let's do it once we have more intent.

Toggl.Daneel.SiriExtension/StopTimerIntentHandler.cs
```diff
@@ -1,18 +1,122 @@
...
         {
-            throw new NotImplementedException();
+            StopTimerIntentHandler.togglApi.TimeEntries.GetAll()                
+                    .Select(getRunningTimeEntry)
+                    .SelectMany(stopTimeEntry)
+                    .Subscribe(
+                    te =>
+                    {
+                        /*
+                        var response = StopTimerIntentResponse.SuccessIntentResponseWithEntry_description(
```
khoi: why don't we just use this response, and when xamarin fixes it we don't have to do anything?

Toggl.Daneel.SiriExtension/StopTimerIntentHandler.cs
Odrakir: In a next PR I'm finally creating a library to share this kind of code.

Toggl.Daneel.SiriExtension/StopTimerIntentHandler.cs
Odrakir: Because now it makes Siri respond with an error.

khoi:
I still can't build, but it has nothing to do with this PR. `siri-shortcuts` is currently broken.
I also couldn't do functional test for this, maybe we can merge this and fix the build problem in another issue?

sofokli1:
Shouldn't there be some `if` logic for comparing the `last update time/At` of the time entry as described in the issue?

sofokli1:
also, shouldn't `android build` work?

Odrakir:
@sofokli1 About the `At` property... there's not yet a decision on what comparison to make, so that'll come in a different PR.

PR #3129, opened 2018-08-31
Create the calendar notification properties in the user preferences storage

### LINKED ISSUE #3115: Create the calendar notification properties in the user preferences storage
We'll need:

- One `IObservable<bool>` to indicate whether or not the user wants to receive notifications
- One method to set this bool (this will emit an event in the aforementioned observable)
- One method to set an integer representing the amount of minutes to offset the start date of the calendar when scheduling notifications
- One `IObservable<int>` that emits whenever this integer value changes

No need to consume these just yet. We only need the properties in place for later issues.

arturs105:
### Relationships

Closes #3115

## :squid: Permissions
Whenever.


Toggl.PrimeRadiant/Settings/IUserPreferences.cs
```diff
@@ -9,6 +9,10 @@ public interface IUserPreferences
 
         IObservable<List<string>> EnabledCalendars { get; }
 
+        IObservable<bool> CalendarNotificationsEnabled { get; }
+
+        IObservable<TimeSpan?> TimeSpanBeforeCalendarNotifications { get; }
```
willsb: I think defaulting this to some value could be useful

PR #3130, opened 2018-08-31
Show the Toggl logo in the nav bar title

### LINKED ISSUE #3124: Add a title to the calendar view
The toggle logo only appears in the log view. It needs to appear in the calendar as well.

juanlaube:
## What's this?
Adds the Toggl logo in the navigation bar title, just like the main log.

### Relationships

Closes #3124 

## How is it done?
©️ 🍝 🙈

### Side effects
None.

## :squid: Permissions
Whenever.


Toggl.Daneel/ViewControllers/Calendar/CalendarViewController.cs
```diff
@@ -50,6 +52,8 @@ public override void ViewDidLoad()
             this.Bind(dataSource.ItemTapped, ViewModel.OnItemTapped);
             this.Bind(createFromSpanHelper.CreateFromSpan, ViewModel.OnDurationSelected);
             this.Bind(editItemHelper.EditCalendarItem, ViewModel.OnUpdateTimeEntry);
+
+            NavigationItem.TitleView = titleImage;
```
willsb: Shouldn't this be set on `ViewWillAppear`?

juanlaube: 🤔
Although there's no need for that, it's safer to do that if we are going to push other screens in the same stack later. Good catch.

PR #3132, opened 2018-08-31
Undo delete UI

### LINKED ISSUE #2541: Undo time entry deletion UI
- Use the same animation to show and hide the snackbar: #1538 

### Undo 

timer - undo deletion

Zeplin: https://zpl.io/a74vwZ8

dshr:
## What's this?
The time entry deletion undo UI. Using the built-in Android Snackbar.

### Relationships
Closes #2541

## Why do we want this?
So people can undo deleting time entries that they didn't want to delete.

## How is it done?
By using the built-in Snackbar and wiring it into the existing undo logic.

### Why not another way?
Seems to be the simplest way.

### Side effects
None.

## Review hints
This is WIP, I'm still struggling with positioning the Snackbar according to the Material guidelines (and so does Google, apparently).

## :squid: Permissions
All yours!


Toggl.Giskard/Activities/MainActivity.cs
```diff
@@ -171,6 +172,16 @@ private async void onTimeEntryCardVisibilityChanged(bool visible)
             }
         }
 
+        private void toggleUndoDeletion(bool show)
```
kornelijepetak: it's actually `setUndoDeletion` since it's not actually being toggled. Depending on `show` it's showing the snackbar.

kornelijepetak: or better yet 'showUndoDeletion'

PR #3133, opened 2018-08-31
🚧 Report workspace button refresh

### LINKED ISSUE #2982: Switching workspaces in Reports is not updated after sync
The floating workspace switching button does not change when user gains or loses access to a workspace and it shows the option to switch to the previously known workspaces.

It is not shown or hidden when the number of workspaces increases from 1 to 2 or more or decreeses back to 1.

If a workspace which is no longer accessible was selected before, it will be still shown and used to fetch reports (although the reports won't be loaded anymore).

## Context
The floating button was added in #2803 

## Expected behavior
When the list of available workspaces changes, the button is ReportsViewModel adjusts its state accordingly and visual elements are shown or hidden correctly with up-to-date data.

## Suggestions
We need to change the workspace data source to an observable data source and observe the changes of workspaces. Don't forget that workspaces which are marked as ghosts should not be accessible as well.

sofokli1:
## What's this?
This makes the switch workspace button show the changes in the workspaces after sync and hides the button if there is only 1 workspace.

### Relationships
Closes #2982

## Why do we want this?
We want the button and what it shows to update as the data changes.

## How is it done?
Partial rxfactoring of the iOS VM/VC and change of the workspace from datasource to observableDatasource.

### Why not another way?
Can't think of any

### Side effects
Only functional tests can tell

## Review hints
Just test a lot. When leaving a workspace in web it never gets removed in the app, not sure why but i don't think this is introduced in this PR. Also i am having trouble with 1 test, even thought the data are correct it fails 😤 crazy stuff (help is very welcome)

## :squid: Permissions
:squid:


Toggl.Foundation.Tests/MvvmCross/ViewModels/ReportsViewModelTests.cs
```diff
@@ -509,14 +527,22 @@ public async Task ShouldTriggerAReportReload()
...
                 await ViewModel.Initialize();
                 var mockWorkspace = new MockWorkspace { Id = WorkspaceId + 1, Name = "Selected workspace" };
                 DialogService.Select(Arg.Any<string>(), Arg.Any<IEnumerable<(string, IThreadSafeWorkspace)>>())
                     .Returns(Observable.Return(mockWorkspace));
 
-                await ViewModel.SelectWorkspace.ExecuteAsync();
+                await ViewModel.SelectWorkspaceAction();
 
-                ViewModel.WorkspaceName.Should().Be(mockWorkspace.Name);
+                TestScheduler.Start();
```
khoi: This is the failing test I assume? 

The problem is the `observer.Messages` value is `"", null, "Selected Workspace"`.

Notice how the second value is null, this value is emitted in `ReportViewModel`'s `Initialize

```csharp
{ 
     var workspace = await interactorFactory.GetDefaultWorkspace().Execute();
     workspaceNameSubject.OnNext(workspace.Name);
}
```

In the test, this is not mocked which lead to the null value.

Toggl.Foundation.Tests/MvvmCross/ViewModels/ReportsViewModelTests.cs
```diff
@@ -509,14 +527,22 @@ public async Task ShouldTriggerAReportReload()
...
+                observer.Messages.AssertEqual(
+                    ReactiveTest.OnNext(1, ""),
```
khoi: This is handy, I didn't know that you could do this 👍

Toggl.Daneel/ViewControllers/ReportsViewController.cs
```diff
@@ -50,42 +51,32 @@ public override void ViewDidLoad()
             source.OnScroll += onReportsTableScrolled;
             ReportsTableView.Source = source;
 
-            var areThereEnoughWorkspaces = new LambdaConverter<ICollection<(string, IThreadSafeWorkspace)>, bool>(workspaces => workspaces.Count > 1);
```
simonrozsival: We can now also 🔥 the `LambdaConverter` itself. Please delete it so we don't forget about it and don't have the unnecessary class in the codebase.

Toggl.Foundation.MvvmCross/ViewModels/ReportsViewModel.cs
```diff
@@ -54,6 +56,9 @@ public sealed class ReportsViewModel : MvxViewModel
         private DateFormat dateFormat;
         private IReadOnlyList<ChartSegment> segments = new ChartSegment[0];
         private IReadOnlyList<ChartSegment> groupedSegments = new ChartSegment[0];
+        private readonly BehaviorSubject<ICollection<(string ItemName, IThreadSafeWorkspace Item)>> workspacesSubject = new BehaviorSubject<ICollection<(string ItemName, IThreadSafeWorkspace Item)>>(new Collection<(string ItemName, IThreadSafeWorkspace Item)>());
```
simonrozsival: 🔥

Toggl.Foundation.MvvmCross/ViewModels/ReportsViewModel.cs
```diff
@@ -38,6 +39,7 @@ public sealed class ReportsViewModel : MvxViewModel
         private readonly IInteractorFactory interactorFactory;
         private readonly IAnalyticsService analyticsService;
         private readonly IDialogService dialogService;
+        private readonly ISchedulerProvider schedulerProvider;
```
simonrozsival: The `schedulerProvider` is used only in the constructor, we don't need the reference to it outside of it and so we don't have to store it in a field. 🔥 please.

Toggl.Foundation.MvvmCross/ViewModels/ReportsViewModel.cs
```diff
@@ -360,14 +401,14 @@ private IReadOnlyList<ChartSegment> groupSegments()
                 .AsReadOnly();
         }
 
-        private async Task selectWorkspace()
+        public async Task SelectWorkspaceAction()
```
arturs105: Why add the "action" part to the name? 🤔

Toggl.Daneel/ViewControllers/ReportsViewController.cs
```diff
@@ -50,42 +51,32 @@ public override void ViewDidLoad()
...
                 return size.Width >= maximumWorkspaceNameLabelWidth;
-            });
+            };
+
+            //Text
+            this.Bind(ViewModel.WorkspaceNameObservable, WorkspaceLabel.BindText());
+            this.Bind(ViewModel.CurrentDateRangeStringObservable, titleButton.BindTitle());
+
+            //Visibility
+            this.Bind(ViewModel.WorkspacesObservable.Select(areThereEnoughWorkspaces), WorkspaceButton.BindIsVisible());
```
simonrozsival: When you open the reports view with a single workspace, the button is visible for a brief moment and then it is hidden. An easy fix: add `.StartWith(false)` to the observable.

Toggl.Foundation.Tests/MvvmCross/ViewModels/ReportsViewModelTests.cs
```diff
@@ -266,14 +269,22 @@ public async Task IsInitializedToEmptyOrNull()
...
+
                 var currentDate = new DateTimeOffset(currentYear, currentMonth, currentDay, 0, 0, 0, TimeSpan.Zero);
                 var start = new DateTimeOffset(startYear, startMonth, startDay, 0, 0, 0, TimeSpan.Zero);
                 var end = new DateTimeOffset(endYear, endMonth, endDay, 0, 0, 0, TimeSpan.Zero);
                 TimeService.CurrentDateTime.Returns(currentDate);
                 ViewModel.ChangeDateRangeCommand.Execute(
                     ReportsDateRangeParameter.WithDates(start, end).WithSource(ReportsSource.Calendar));
 
-                ViewModel.CurrentDateRangeString.Should().Be($"{Resources.ThisWeek} ▾");
+                //ViewModel.CurrentDateRangeString.Should().Be($"{Resources.ThisWeek} ▾");
```
arturs105: This should be either uncommented or removed.

Toggl.Daneel/ViewControllers/ReportsViewController.cs
sofokli1: What do you think about having the `WorkspaceButton` hidden in IB by default?

Toggl.Foundation.MvvmCross/ViewModels/ReportsViewModel.cs
sofokli1: there is something else called `SelectWorkspace`. Will rename `SelectWorkspaceAction` to `SelectWorkspaceMethod`. How does this sound?

Toggl.Foundation.Tests/MvvmCross/ViewModels/ReportsViewModelTests.cs
sofokli1: thanx, I was going crazy with this one 😤

sofokli1:
@khoi since the app does not detect when it loses access to a workspace there isn't much to test in that situation until that feature is available.

Toggl.Daneel/ViewControllers/ReportsViewController.cs
simonrozsival: That would work as well, I am fine with that.

PR #3135, opened 2018-08-31
Remove needless calls to reloadData

willsb:
## What's this?
This removes two needless calls to reload data. Since we're subscribing to the time entry changes, reload data will already be called via our subscriptions

## :squid: Permissions
Anyone


willsb:
Obsolete

PR #3136, opened 2018-08-31
Notification settings view model

juanlaube:
## What's this?
Adds a new notification settings view model, and an auxiliary generic view model to select an item from a list

### Relationships

Closes # 3127

## How is it done?
It's a new view model, very similar to the calendar settings view model, the new thing is the `SelectFromListViewModel` which can be reused everywhere we need to pick an item from a list

### Why not another way?
The `SelectFromListViewModel` could be something specific for the upcoming events config, but that's boring 😄 

### Side effects
None

## Review hints
Go commit by commit

## :squid: Permissions
Only myself


Toggl.Foundation.MvvmCross/Parameters/SelectFromListParameters.cs
```diff
@@ -0,0 +1,19 @@
...
+        public string Title { get; set; }
+
+        public TValue Value { get; set; }
+    }
+
+    public struct SelectFromListParameters<TValue>
+    {
+        public IList<SelectableItem<TValue>> Items { get; set; }
+
+        public int SelectedIndex { get; set; }
```
sofokli1: Why not make this usable without a default selection? Ex: when you want to make the user make a choice, when there is no default choice.

Toggl.Foundation.MvvmCross/ViewModels/Settings/NotificationSettingsViewModel.cs
```diff
@@ -0,0 +1,140 @@
...
+                saveUpcomingEventsOption(selection);
+            }).SelectUnit();
+        }
+
+        private IObservable<int> currentUpcomingEventsOptionIndex()
+        {
+            return Observable.FromAsync(async () =>
+            {
+                var notificationsEnabled = await userPreferences.CalendarNotificationsEnabled;
+                var timespan = await userPreferences.TimeSpanBeforeCalendarNotifications;
```
sofokli1: Shouldn't `timespan` be `timeSpan`?

Toggl.Foundation.MvvmCross/Parameters/SelectFromListParameters.cs
juanlaube: That's a good idea, but I will let someone who needs that particular use case to do it 😂

Toggl.Foundation.MvvmCross/ViewModels/Settings/NotificationSettingsViewModel.cs
juanlaube: Good catch

PR #3138, opened 2018-09-03
Fix te stopping bug on slow networks

### LINKED ISSUE #3068: The TEs log can contain rows referencing entries with negative ID even after sync suceeds
Build: 1.9 - 1671

Steps to reproduce:
- make sure the app has no TEs and is synced (eg after new sign up)
- start creating TEs with unique descriptions (eg “A”, “Aa”, “Aaa”,...)
- as soon as the TE is started, stop it and proceed to creation of another TE
- once you see a duplicite row with an unsynced icon, swipe to delete
- wait 5s, then you should get a :boom: 

Recommendation: use a shitty network

The bug is the visibility of the row whith the “unsynced entity” icon after a successful sync

simonrozsival:
### Relationships

Closes #3068

## Why do we want this?
We don't want the app to crash even for users who don't have a fast network connection (3G or better).

## How is it done?
There was a problem with conflict resolution because we didn't mark the time of the last change when stopping a time entry. That resulted in a time entry being displayed in the log although in fact it was still running due to IDs mismatch (ID from the server vs. the original local negative ID).

### Why not another way?
This is the best way to get rid of this problem I can think of.

### Side effects

The same bug was probably present in Giskard as well, but I didn't test the Android version. Now both platforms should be OK and immune to this problem.

## Review hints
Use Network Line Conditioner from Apple (see https://nshipster.com/network-link-conditioner/ for instructions) to simulate Edge for the simulator. Start a time entry and immediately stop it (you have a few seconds before the app syncs). The TE should be stopped and appear in the log. When the app finishes sync it should be displayed as synced and you can edit or delete it without any crash.

Please try the same on Android if possible.

## :squid: Permissions
🦑 whenever


Toggl.Foundation/DataSources/TimeEntriesDataSource.cs
```diff
@@ -103,6 +103,7 @@ public IObservable<IThreadSafeTimeEntry> Stop(DateTimeOffset stopTime)
                 .Select(timeEntries => timeEntries.SingleOrDefault() ?? throw new NoRunningTimeEntryException())
                 .SelectMany(timeEntry => timeEntry
                     .With((long)(stopTime - timeEntry.Start).TotalSeconds)
+                    .UpdatedAt(timeService.CurrentDateTime)
```
amulware: Not happy about creating two copies here, but we can hopefully refactor that away easily once we take another look at DTOs etc.

Toggl.Foundation/Models/TimeEntry.cs
```diff
@@ -170,6 +170,12 @@ public TimeEntry(IDatabaseTimeEntry timeEntry, long duration)
             Duration = duration;
         }
 
+        public TimeEntry(IDatabaseTimeEntry timeEntry, DateTimeOffset at)
```
amulware: Also not happy about this being public, but it'll do for now maybe.

dshr:
Still happens on Android, I'm afraid :( Steps to reproduce:

simonrozsival:
The #3162 issue describes the bug @dshr found. It will be fixed in a separate PR.

PR #3143, opened 2018-09-04
Fix integration tests

### LINKED ISSUE #3048: Update integration tests for empty me/time_entries
## 🐛 Bug Report
Backend changed `me/time_entries` to return an empty list `[]` instead of `null`. We have two tests expecting the `null` that have to be updated.

## Context
The tests in question are:
`Toggl.Ultrawave.Tests.Integration.TimeEntriesApiTests+TheGetAllMethod.ReturnsNullInsteadOfAnEmptyListWhenThereIsNoTimeEntryOnTheServer`
`Toggl.Ultrawave.Tests.Integration.TimeEntriesApiTests+TheDeleteMethod.DeletesTheTimeEntryFromTheServer`

The first documented the previous bugged behaviour on purpose, and should probably be rewritten and renamed accordingly. The second just checked that nothing is returned after deleting a time entry and the change should be trivial.

## Expected behavior
No failing tests!

## To reproduce
Run those integration tests.

simonrozsival:
## What's this?
The goal of this PR is to make the integration tests work again so we can get some ✅s again and have more confidence in our networking layer.

There were two changes recently:

1. the response body of the `me/time_entries` endpoint for the case when the user has no time entries changed (as described in #3048)
2. the enterprise plans were removed

### Relationships
Closes #3048

## Why do we want this?
We want to be sure our networking is implemented correctly.

## How is it done?
The changes with regards to the pricing plans changes are simple. I just removed the items from the enum and removed some test cases of a theory.

The changes of the `me/time_entries` endpoint are a bit more complex. It could be done much more simpler, but I didn't want to do that - specifically the changes in `PersistListState.cs` on line 47 might not be necessary, but it would leave us with code which will never be executed and will be confusing for anyone who starts looking into this syncing state.

These changes meant that I had to add integration tests which ensure that the case of an empty list for clients, tags, projects, tasks, time entries, and workspaces is consistent and the body of the response is never null (for both `GetAll` and `GetAllSince` methods). This means more changes in integration tests (most notably the `v8` endpoint for deletion of a workspace).

### Why not another way?
I wanted not just to make the tests pass again, but also make the code and tests the way it should be.

### Side effects

## Review hints
Check the integration tests: https://app.bitrise.io/build/112415caa011024c

## :squid: Permissions
Squash whenever


Toggl.Multivac.Tests/EnsureTests.cs
```diff
@@ -143,7 +142,7 @@ public void ThrowsWhenTheArgumentIsNotADefinedEnumValue()
             public void DoesNotThrowWhenTheArgumentIsNotNull()
             {
                 Action whenTheCalledArgumentIsNotADefinedEnumValue =
-                    () => Ensure.Argument.IsADefinedEnumValue<SyncStatus>(SyncStatus.SyncNeeded, "argument");
```
amulware: These changes seem unrelated.

Toggl.Multivac.Tests/Toggl.Multivac.Tests.csproj
```diff
@@ -25,6 +25,5 @@
   </ItemGroup>
   <ItemGroup>
     <ProjectReference Include="..\Toggl.Multivac\Toggl.Multivac.csproj" />
-    <ProjectReference Include="..\Toggl.PrimeRadiant\Toggl.PrimeRadiant.csproj" />
```
amulware: Also unrelated to the PR.

Toggl.Ultrawave.Tests.Integration/WorkspacesApiTests.cs
```diff
@@ -20,6 +24,17 @@ public sealed class TheGetMethod : AuthenticatedEndpointBaseTests<List<IWorkspac
...
+            public async Task ReturnsOneWorkspaceForANewUser()
+            {
+                var (togglClient, user) = await SetupTestUser();
+
+                var workspaces = await CallEndpointWith(togglClient);
+
+                workspaces.Should().HaveCount(1);
+                workspaces.Should().Contain(ws => ws.Id == user.DefaultWorkspaceId);
+            }
+
```
amulware: Also not really part of the stated purpose.

Toggl.Multivac.Tests/EnsureTests.cs
simonrozsival: This is not unrelated. `SyncStatus` is in PrimeRadiant, I wasn't able to build and run integration tests locally when `Multivac` depended on `PrimeRadiant` for some reason. I had to remove this dependency to make things work locally (I have no idea why it works/worked on Bitrise though).

Toggl.Multivac.Tests/Toggl.Multivac.Tests.csproj
simonrozsival: For the same reason as above.

Toggl.Ultrawave.Tests.Integration/WorkspacesApiTests.cs
simonrozsival: I agree that this test is not necessary. I would say that we should have this test though and since it is so simple, I would keep it in this PR.

PR #3156, opened 2018-09-04
Logout changes

willsb:
## What's this?
This unschedules notifications when logging out

## Why do we want this?
Because reminders should be handled only when the app is in the logged in state

## How is it done?
We simply run the service directly. Ideally, this should all be an interactor. I'll push for a bigger refactor later.

## :squid: Permissions
Anyone, anytime


Toggl.Foundation.Tests/DataSources/TogglDataSourceTests.cs
```diff
@@ -89,6 +91,19 @@ public void DoesNotClearTheDatabaseBeforeTheSyncManagerCompletesFreezing()
                 Database.DidNotReceive().Clear();
             }
 
+            [Fact, LogIfTooSlow]
+            public void UnscheduleAllNotifications()
```
sofokli1: Based on the above tests, shouldn't this be called `UnschedulesAll...`?

Toggl.Foundation.Tests/DataSources/TogglDataSourceTests.cs
```diff
@@ -89,6 +91,19 @@ public void DoesNotClearTheDatabaseBeforeTheSyncManagerCompletesFreezing()
...
+            {
+                var scheduler = new TestScheduler();
+                SyncManager.Freeze().Returns(Observable.Never<SyncState>());
+
+                var observable = DataSource.Logout().SubscribeOn(scheduler).Publish();
+                observable.Connect();
+                scheduler.AdvanceBy(TimeSpan.FromDays(1).Ticks);
+
+                NotificationService.DidNotReceive().UnscheduleAllNotifications();
```
sofokli1: I don't get this. why `DidNotReceive`? shouldn't it receive `UnscheduleAllNotifications` based on the name of the test?

Toggl.Foundation/DataSources/TogglDataSource.cs
```diff
@@ -146,6 +144,10 @@ public IObservable<Unit> Logout()
                 .Do(_ => stopSyncingOnSignal())
                 .SelectMany(_ => database.Clear())
                 .Do(_ => shortcutCreator.OnLogout())
+                .SelectMany(_ => 
+                    notificationService
+                        .UnscheduleAllNotifications()
+                        .Catch(Observable.Return(Unit.Default)))
```
juanlaube: In what particular situation can this throw? 🤔

willsb: If we don't have the permission for notifications

Toggl.Foundation.Tests/DataSources/TogglDataSourceTests.cs
willsb: ©️ 🍝

PR #3163, opened 2018-09-05
Fix incorrect transitions when entity changes during creation sync

### LINKED ISSUE #3162: Time entry is created repeatedly when it changes during sync
## 🐛 Bug Report
There is a weird bug which causes our app to create duplicite TEs. The main reason is that if the network is slow and the POST request takes too long (several seconds), the user can easily change the time entry - e.g., stop it. When the change is detected by the `CreateEntityState` it changes the ID to the one given by the server and returns the `EntityChanged` result. This result is then looped back to the `CreateEntityState`. This time we push the entity again, even though it already has a positive ID and there is no reason to push it 🤦‍♂️

## Expected behavior
What should happen instead is that the `EntityChanged` result of the `CreateEntityState` should be connected to the `UpdateEntityState`.

## To reproduce
- Use Network Link Conditioner to simulate Edge connectivity
- start a time entry and stop it immediately before it syncs
- wait for the app to sync
- force-sync the app again
- you will now see an extra running TE appearing from nowhere

## Suggestions
The solution is already described in this bug report.

simonrozsival:
## What's this?
This PR simply implements the fix described in the issue.

### Relationships
Closes #3162, should be merged after #3138 

## Why do we want this?
We don't want duplicate time entries (or other entities).

## How is it done?
It is enough to change the transitions.

### Why not another way?
This is the simplest way.

### Side effects
No side effects.

## Review hints
Read the steps in the issue and try reproducing it. You should not get the incorrect output anymore with this branch. Please note that the the PR #3138 should be merged first so the UI works well.

## :squid: Permissions
🦑


Toggl.Foundation/TogglSyncManagerFactory.cs
```diff
@@ -276,7 +276,7 @@ public static class TogglSyncManager
             transitions.ConfigureTransition(checkServerStatus.Retry, checkServerStatus);
             transitions.ConfigureTransition(checkServerStatus.ServerIsAvailable, push);
 
-            transitions.ConfigureTransition(create.EntityChanged, create);
+            transitions.ConfigureTransition(create.EntityChanged, update);
```
amulware: I disagree with this. The entity could have been deleted. We should instead go all the way back to the top and make sure we make the correct decision on what to do, or in fact even refetch the entity to be synced from the DB.

Toggl.Foundation/TogglSyncManagerFactory.cs
```diff
@@ -276,8 +276,8 @@ public static class TogglSyncManager
             transitions.ConfigureTransition(checkServerStatus.Retry, checkServerStatus);
             transitions.ConfigureTransition(checkServerStatus.ServerIsAvailable, push);
 
-            transitions.ConfigureTransition(create.EntityChanged, create);
-            transitions.ConfigureTransition(update.EntityChanged, update);
+            transitions.ConfigureTransition(create.EntityChanged, pushOne);
+            transitions.ConfigureTransition(update.EntityChanged, pushOne);
```
amulware: This transition will still pass the entity right? Does it pass the original, or the new version that was found?

simonrozsival: It passes the entity. I re-checked it, because I was quite sure the `create.EntityChanged` passed the updated entity, but I did not remember what `update` did. The `create` state does what it should, the `update` passes the _old_ entity though.

That means that doing it the way this PR does right now would not help us in the case of updating (it would override any local changes).

That means either we must edit the updating state so that it fetches the up-to-date entity and passes it through the transition, or configure the transition to the `push` state.  This state would then pick the next entity to push - this will be the same entity with high probability, but it might be a different one. It is probably not important in which order these entities will be processed IMO.

I am not sure which of the approaches is better. I will think about it.

amulware: I'd say going to the `push` state is easiest, and moves us closer to making the loops bigger as discussed internally.

amulware: Generally, I think continuing to use the data passed in this state result is tricky since it's not really clear what data is passed exactly. That's bound to cause confusion or bugs.

Toggl.Foundation/TogglSyncManagerFactory.cs
```diff
@@ -335,7 +335,7 @@ public static class TogglSyncManager
             transitions.ConfigureTransition(checkServerStatus.Retry, checkServerStatus);
             transitions.ConfigureTransition(checkServerStatus.ServerIsAvailable, push);
 
-            transitions.ConfigureTransition(create.EntityChanged, create);
+            transitions.ConfigureTransition(create.EntityChanged, new InvalidTransitionState($"Entity cannot changed since updating is not supported for {typeof(TModel).Name} during Push sync."));
```
amulware: Grammar. `Entity cannot changed` -> `Entity should not change/cannot have changed`

PR #3168, opened 2018-09-05
Add 3D touch actions to calendar notification

juanlaube:
## What's this?
Adds 3D touch actions to calendar notification, because we are cool

### Relationships

Closes nothing

## Why do we want this?
Because it's dope

### Side effects
None

## Review hints
Please test

## :squid: Permissions
Whenever


Toggl.Foundation/Extensions/TimeEntryExtensions.cs
```diff
@@ -24,9 +24,21 @@ public static ITimeEntryPrototype AsTimeEntryPrototype(this TimeSpan timeSpan, D
...
-                projectId: null, 
-                taskId: null, 
-                tagIds: null, 
+                projectId: null,
+                taskId: null,
+                tagIds: null,
+                isBillable: false
+            );
+
+        public static ITimeEntryPrototype AsRunningTimeEntryPrototype(this DateTimeOffset startTime, long workspaceId, string description = "")
```
willsb: I don't think this is necessary. We should track the time chunk, not start a time entry.

Toggl.Foundation/Interactors/Notifications/ScheduleEventNotificationsInteractor.cs
```diff
@@ -59,7 +59,8 @@ private Notification eventNotification(CalendarItem calendarItem)
                 calendarItem.Id,
                 Resources.EventReminder,
                 calendarItem.Description,
-                calendarItem.StartTime - intervalBeforeEvent
+                calendarItem.StartTime - intervalBeforeEvent,
+                Notification.CalendarEventCategory
```
willsb: Why is this needed? We only have this one kind of notification. We can care about that later

willsb: Also: This shouldn't be a string, but rather an enum so that Android can benefit from it as well. But not now. Right now just 🔥 this

Toggl.Daneel/Services/NotificationService.cs
```diff
@@ -13,12 +13,20 @@
...
 using UserNotifications;
 using Notification = Toggl.Multivac.Notification;
+using FoundationResources = Toggl.Foundation.Resources;
 
 namespace Toggl.Daneel.Services
 {
     public sealed class NotificationService : PermissionAwareNotificationService
     {
         public const string CalendarEventIdKey = "Id";
+        public const string DescriptionKey = "Description";
```
willsb: This can be 🔥, can't it?

Toggl.Foundation/Extensions/TimeEntryExtensions.cs
```diff
@@ -24,9 +24,9 @@ public static ITimeEntryPrototype AsTimeEntryPrototype(this TimeSpan timeSpan, D
                 description: "",
                 duration: timeSpan,
                 startTime: startTime,
-                projectId: null,
```
willsb: ♻️ this

Toggl.Multivac/Notification.cs
```diff
@@ -12,7 +12,11 @@ public struct Notification
 
         public DateTimeOffset ScheduledTime { get; }
 
-        public Notification(string id, string title, string description, DateTimeOffset scheduledTime)
+        public Notification(
```
willsb: ♻️ this as well

PR #3169, opened 2018-09-05
Add two Do extension methods

kornelijepetak:
## What's this?
Adds 2 extension methods with the same behaviour.

1. `Do` on `IEnumerable<T>`
2. `Do` on `IObservable<T>`

### Relationships
Alone and lonely.

## Why do we want this?
To reduce the amount of surrogate `_` identifiers when they are not really needed (such as ambiguities or functions that do not return `void`).

## How is it done?
By just wrapping the usual `Do`.

### Side effects
Wouldn't say so.

## Review hints
The usuals. This one's not hard.

## :squid: Permissions
Do! 

(_pun intended_)


Toggl.Multivac/Extensions/FunctionalExtensions.cs
```diff
@@ -57,5 +58,11 @@ public static IEnumerable<T> Do<T>(this IEnumerable<T> self, Action<T> action)
             self.ForEach(action);
             return self;
         }
+
+        public static IEnumerable<T> Do<T>(this IEnumerable<T> self, Action action)
+        {
+            self.Do(x => action());
```
willsb: `_` 👀

kornelijepetak: Oh, yeah, I was just testing something and then forgot to change it. Nice find.

PR #3170, opened 2018-09-05
Add running time entry persistent notification

### LINKED ISSUE #3055: Add running timer notification
This is an optional feature that can be enabled in the settings. When it is enabled, as long as the user is logged in and has a running timer we should show a permanent notification that will inform the user that the timer is running.

This notification shows data from the time entry, including a real-time duration. It allows the user to go straight back to the app, and may also include a stop button.

If the user changes the appropriate setting, the notification should appear/disappear right away if needed.

daividssilverio:
## What's this?
We are bringing back the persistent notifications o/

### Relationships
Closes #3055 

## Why do we want this?
Because our clients miss the notifications from the old app!

## How is it done?
An observable to the running time entry is added to the main vm and observed in the MainActivity. Every time the current running TE changes, the notification will be updated.

I removed vibration and there should be no notification sound played when the notification is created/updated.

### Why not another way?
This is simple enough. We don't have background sync yet, so there's no need for a service and the notification being updated only from the MainActivity works well enough.

### Side effects
0️⃣

## Review hints
Just enable the running notifications on settings and you will see the notifications there.
Disable it will disappear.

## :squid: Permissions
Go ahead. 🦑


Toggl.Giskard/Services/GiskardNotificationsPostman.cs
```diff
@@ -0,0 +1,90 @@
...
+using Android.Content;
+using Java.Lang;
+using Toggl.Foundation.Models.Interfaces;
+using Toggl.Foundation.MvvmCross.ViewModels;
+using Toggl.Giskard.Activities;
+using Toggl.Giskard.Extensions;
+
+namespace Toggl.Giskard
+{
+    public static class GiskarNotificationsPostman
```
willsb: Postman is not a good name (it's pretty cool, but not a good name)

Toggl.Giskard/Extensions/ActivityExtensions.cs
```diff
@@ -4,12 +4,15 @@
...
 using Toggl.Giskard.Helper;
+using Android.Support.V4.App;
+
 
 namespace Toggl.Giskard.Extensions
 {
     public static class ActivityExtensions
     {
         private static readonly Color lollipopFallbackStatusBarColor = Color.ParseColor("#2C2C2C");
+        private static readonly long[] noNotificationVibrationPattern = {0L, 0L};
```
willsb: Put a space after the first curly brace and before the last one

Toggl.Giskard/Extensions/ActivityExtensions.cs
```diff
@@ -62,5 +65,35 @@ public static void SetDefaultDialogLayout(this Window window, Activity activity,
 
             window.SetLayout(width, height);
         }
+
+        public static NotificationCompat.Builder CreateNotificationBuilderWithDefaultChannel(this Context context,
```
willsb: Why say `WithDefaultChannel` and then allow the channel to be supplied? That's a bit odd

willsb: I'd remove these optional strings (make them `private const`s instead).

Toggl.Giskard/Services/GiskardNotificationsPostman.cs
```diff
@@ -0,0 +1,90 @@
...
+
+        private static void updateRunningNotification(IThreadSafeTimeEntry timeEntryViewModel, Activity activity, NotificationManager notificationManager)
+        {
+            if (notificationManager == null) return;
+
+            if (timeEntryViewModel != null)
+            {
+                var startTime = timeEntryViewModel.Start.ToUnixTimeMilliseconds();
+                var timeEntryDescription = string.IsNullOrEmpty(timeEntryViewModel.Description)
+                    ? activity.GetString(Resource.String.NoDescription)
```
willsb: I think we already have a proper resource in Foundation for `NoDescription`. Even if we don't, I'd rather have this on foundation instead of here in Giskard.

Toggl.Giskard/Services/GiskardNotificationsPostman.cs
```diff
@@ -0,0 +1,90 @@
...
+                    : timeEntryViewModel.Description;
+                var projectDetails = extractProjectDetails(activity, timeEntryViewModel);
+                var startMainActivityIntent = new Intent(activity, typeof(MainActivity))
```
willsb: Why not make this a a URI intent that opens the main activity using shortcuts? I think that's a tad better because it relies on Mvx. Did you test that this correctly initializes MvvmCross, even though it's skipping the SplashScreen?

Toggl.Giskard/Services/GiskardNotificationsPostman.cs
```diff
@@ -0,0 +1,90 @@
...
+            if (timeEntryViewModel.Project.ClientId != null)
+            {
+                projectDetails.Append($" \u00b7 {timeEntryViewModel.Project.Client.Name}");
+            }
+            return projectDetails.ToString();
+        }
+
+        private static void cancelRunningTimeEntryNotification(NotificationManager notificationManager)
+        {
+            notificationManager.Cancel(runningTimeEntryNotificationId);
```
willsb: I don't see the need for a one line method when `notificationManager.Cancel(runningTimeEntryNotificationId);` is pretty clear

Toggl.Giskard/Services/GiskardNotificationsPostman.cs
```diff
@@ -0,0 +1,90 @@
...
+            {
+                cancelRunningTimeEntryNotification(notificationManager);
+            }
+        }
+
+        private static string extractProjectDetails(Activity activity, IThreadSafeTimeEntry timeEntryViewModel)
+        {
+            if (timeEntryViewModel.ProjectId == null)
+            {
+                return activity.GetString(Resource.String.NoProject);
```
willsb: Same thing about Foundation Resources.

Toggl.Giskard/Services/GiskardNotificationsPostman.cs
```diff
@@ -0,0 +1,90 @@
...
+
+            var projectDetails = new StringBuilder(timeEntryViewModel.Project.Name);
+            if (timeEntryViewModel.TaskId != null)
+            {
+                projectDetails.Append($": {timeEntryViewModel.Task.Name}");
+            }
+
+            if (timeEntryViewModel.Project.ClientId != null)
+            {
+                projectDetails.Append($" \u00b7 {timeEntryViewModel.Project.Client.Name}");
```
willsb: Please put `\u00b7` in a properly named constant

Toggl.Giskard/Extensions/ActivityExtensions.cs
```diff
@@ -62,5 +65,35 @@ public static void SetDefaultDialogLayout(this Window window, Activity activity,
...
+            string channelDescription = "Toggl notifications",
+            bool vibrationEnabled = false,
+            NotificationImportance notificationImportance = NotificationImportance.Default)
+        {
+            if (OreoApis.AreAvailable)
+            {
+                var channel = new NotificationChannel(channelId, channelName, notificationImportance);
+                channel.Description = channelDescription;
+                channel.EnableVibration(vibrationEnabled);
+                if (!vibrationEnabled)
```
willsb: Same for this vibrate thing. This is our only notification and vibration is not on. IMO, YAGNI applies and you can just make everything silent until someone needs vibration.

Toggl.Giskard/Services/GiskardNotificationsPostman.cs
daividssilverio: Ohh... come one, it was nice. And I forgot to move it to the helper package.
`PersistentNotificationsHelper` is any better?

Toggl.Giskard/Extensions/ActivityExtensions.cs
daividssilverio: I thought about using this later for the calendar notifications. But we can make it generic later when we need.

Toggl.Giskard/Services/GiskardNotificationsPostman.cs
daividssilverio: It works fine. But I can do the other way if it you think it would play better with mvx.

Toggl.Giskard/Services/GiskardNotificationsPostman.cs
daividssilverio: Indeed o/

Toggl.Giskard/Services/GiskardNotificationsPostman.cs
daividssilverio: No, I added one now though.

PR #3171, opened 2018-09-06
📊 Donating intents when users start a new timer

### LINKED ISSUE #3046: Create start entry intent
When ever the user hits the start button we should create and donate an intent "Start time tracking" for an empty task.

It won't get or return anything, the response should be something like "Time tracking started" or similar.

khoi:
## What's this?
This donates the intents to Sirikit when user start a new timer.

### Relationships

Closes #3046

## Why do we want this?
So Siri can intelligently predict and suggest shortcuts to users.

## How is it done?
New intent is added to the intent definition file. And the interactions with report is donated in Daneel

### Side effects
🤷🏻‍♂️

## Review hints

- Start a timer
- See if the donations are listed in home screen (must enable Developer settings)

## :squid: Permissions
Go for it


Odrakir:
I'm not sure about that, because continuing an existing entry is the only way the user has to start the timer for a specific task. If every morning you "Code review" the user should be able to have a shortcut to "start" a timer for that.

khoi:
we can separate out
- start a new timer intent
- continue existing entry intent
would that work?

PR #3172, opened 2018-09-06
Listen to TE changes in Edit TE view

### LINKED ISSUE #3165: Edit time entry view cannot save changes when ID of the time entry changes because of sync
## 🐛 Bug Report
If a time entry is created and user is navigated to the edit view before the TE syncs and then the syncing finishes, which leads to the change of the ID of the time entry from temporary ID to the proper value given by backend, the changes made in the edit view cannot be saved.

## Context

Build/Branch: `develop`
Phone OS: both

## Expected behavior
The changes can be saved.

simonrozsival:
## What's this?
This PR introduces a way of closing the issue. This is one of the two ways we discussed (the other one is changing the repository to match entities by `Id` and `OriginalId`) and we don't have a consensus about the correct way of dealing with this problem. This PR can be a place where we can discuss this matter.

### Relationships

Closes #3165 

## Why do we want this?
We want to be able to edit TEs right after we create them (we need this for a specific use case in the calendar).

## How is it done?
This implementation listens for changes to the edited time entry and updates the internal fields accordingly, unless they are dirty. The ID is always updated (it cannot be edited manually and be made dirty).

### Why not another way?
I don't want to change the way we work with the storage. It would add even more logic to `PrimeRadiant` and it would make it even more complex and error prone and scary.

### Side effects
There are no side effects.

## Review hints
Use a slow network (Network Link Conditioner) to simulate a slow network (Edge). Start a new time entry and tap the card to edit it before it is synced. Make some changes and save it. It should work without any problem.

## :squid: Permissions
Let's discuss this first and only merge this if we can all agree that this is a good way of fixing this problem. If we decide to take the other approach, then we will close this.

## WIP
Even if we decide to go with this approach, more tests should be added.


amulware:
I'm sort of VETOing this for being complicating and finicky and different from the way we handle similar problems in all other places. We added the local ids for a reason, and should be using them.

simonrozsival:
Superseded by #3176

PR #3175, opened 2018-09-06
Extension ui

### LINKED ISSUE #3122: Add UI Extension to have some kind of response
While they fix not having custom Siri responses (more info here: https://github.com/xamarin/xamarin-macios/issues/4581) we are going to use a UI Extension to show some kind of info on success (I'm afraid we can't do anything about the error cases)

So we need the following designs:
- [x] UI for "stop current entry" intent. I'd say it should show the description of the entry and how long it has been running.
- [x] UI for "start entry" intent. This is for empty entries with no description.
- [ ] UI for "start specific entry" intent. This is for existing entries, we know the description.

Odrakir:
## What's this?
This PR adds a UI for Siri responses.

Closes #3122 

## Why do we want this?
There's a bug in Xamarin preventing Siri from saying custom responses, that's why we need some UI to show something to the user.

## How is it done?
Adds yet another extension.

## Review hints
This is not working for now. I'm getting this error:
`Foundation.MonoTouchException: Objective-C exception thrown.  Name: NSUnknownKeyException Reason: [<StopTimerIntentResponse 0x60000235f680> valueForUndefinedKey:]: this class is not key value coding-compliant for the key entry_description.`

Seems like it doesn't know about the intent's response properties.

## :squid: Permissions
Not yet.


Toggl.Daneel.SiriExtension.UI/IntentViewController.cs
```diff
@@ -0,0 +1,93 @@
...
+using Foundation;
+using Intents;
+using IntentsUI;
+using UIKit;
+using CoreFoundation;
+using Toggl.Daneel.Intents;
+
+namespace Toggl.Daneel.SiriExtension.UI
+{
+    // As an example, this extension's Info.plist has been configured to handle interactions for INSendMessageIntent.
```
khoi: 🔥 these comments

PR #3180, opened 2018-09-06
Pick item foreground color based on Luma

### LINKED ISSUE #3167: Handle cases when background of calendar entries has light color.
Currently time entries in the calendar can be barely readable if the project color is too light (see screenshot).

We should use varying text colors based on the background color.

screen shot 2018-09-05 at 17 40 32

willsb:
## What's this?
Cool colours for our calendar items

### Relationships
Closes #3167

## How is it done?
Luma calculation (weighted sums of R, G and B) and then checking against a threshold.

### Why not another way?
This is simple enough

### Side effects
The app is 2 cool for school now

## Review hints
Use the `YELLOW` project of our test account, create by yours truly.

## :squid: Permissions
Anyone, anytime


Toggl.Foundation.MvvmCross/Extensions/ColorExtensions.cs
```diff
@@ -60,5 +61,12 @@ public static float GetHue(this MvxColor self)
             }
             return hue / 360;
         }
+
+        public static MvxColor ForegroundColor(this CalendarItem calendarItem)
+        {
+            var color = MvxColor.ParseHexString(calendarItem.Color);
+            var luma = 0.2126 * color.R + 0.7152 * color.G + 0.0722 * color.B;
```
juanlaube: This formula works only with normalized values, divide by 255 and we're good.

Two screenshots to compare:
simulator screen shot - iphone x ios 11 4 - 2018-09-06 at 10 57 28
simulator screen shot - iphone x ios 11 4 - 2018-09-06 at 11 02 38

PR #3181, opened 2018-09-06
Set `At` date when soft deleting a time entry

simonrozsival:
## What's this?
This PR changes the `At` date of a time entry when soft-deleting the time entry (soft-deleting is just an update which sets the `IsDeleted` flag to `true`). Without this change, the `IsDeleted` flag might be ignored and overwritten and effectively un-deleting it. This happened to me a few times when testing over a slow network.

### Relationships

Similar to #3138 

## Why do we want this?
We want to know when the entity was updated the last so we can run conflict resolution correctly.

### Side effects
No side effects. The thing that worries me is that the problem which was already present in #3138 is present in some other places. Namely I am not sure what should happen in `stopMultipleRunningTimeEntries`? Should we change the `At`s there too?

## :squid: Permissions
Squash whenever


simonrozsival:
This is probably not an urgent bug, even though we can't really be sure (it does not crash the app, we don't know if this happens to the users and how often). I would not wait for this for the 1.10 release.

kornelijepetak:
@simonrozsival do you think any tests for this would be useful?

simonrozsival:
I should add tests, you are right 😄  I will try to add them ASAP.

Issue #3056, opened 2018-08-23
Add idle timer notification


amulware:
This is an optional feature that can be enabled in the settings. When it is enabled, as long as the user is logged in and has no running timer we should show a permanent notification that will inform the user that no timer is running.

It will also allow them to get to the app easily. Perhaps there will even be a shortcut button to start a timer and go straight to the start view?

If the user changes the appropriate setting, the notification should appear/disappear right away if needed.

pe0ny:
<img width="369" alt="screen shot 2018-08-24 at 10 14 22" src="https://user-images.githubusercontent.com/5064445/44570432-809c8e00-a786-11e8-8926-abe68cdf6e1c.png">

### The header area is comprised of:

**1. The app icon:**
The app icon that appears in monochrome in the status bar.
Question: Is it possible to show a different icon there when the timer is not running? If not, show the same like in here #3055

**2. The app name:**
The name of the app automatically appears in the notification.

**3. Header text:**
We won't display anything here.

**4. Timestamp:**
What the title says.

### The content area:

**1. A content title:**
Since both a content title and text are needed, we'll display just "Toggl" here.

**2. Content text:**
And "What are you working on?" here.

Tapping the notification itself will open the app in the Timer page.

The expandable option will be added in the future but not in this phase.

dshr:
Putting this on hold until we merge #3170

dshr:
@pe0ny I'm a bit confused about the timestamp — do we show the time since the app was opened or the time since the last time entry has ended?

daividssilverio:
🤔 It looks like it's the time since the last time entry has ended. But... this might be weird if the user don't use the app too often. The nice thing about it is that whomever doesn't need it, can disable it.

kornelijepetak:
Closed by #3189

PR #3182, opened 2018-09-06
Remove notifications on token reset

### LINKED ISSUE #3057: Remove idle/running timer notifications on logout and reset api token
Let's make sure we cover those two cases. The logout is pretty straightforward.

With regards to the reset:
- when we detect that the api token is no longer valid, remove the notifications
- do not however clear the settings here yet, unless the user actually logs out
- when the user logs back in on the api reset screen the correct notification should be shown again (if enabled)

daividssilverio:
## What's this?
It removes the notifications currently shown by the app when there's a api token reset.

### Relationships
All notifications should be closed, but the ones being created are coming from: #3055 and #3056 
Closes #3057

## Why do we want this?
We can't show user data if he/she has an invalid api token.

## How is it done?
All that has to be done is to clear any existing notifications.
The settings are reset when logout happens anyway.
After the api token reset logic, the navigation flow should take care of re-sending previously shown notifications.

### Why not another way?
This is simple enough.

### Side effects
If we had any notifications that should stay there when the user is logged out, they would have been removed as well. But we don't have them.

## Review hints
Enable notifications in Settings.
Make sure you have a running TE or wait for #3056 to be done to check the idle notification.
Make/simulate an api token reset.

## :squid: Permissions
🦑 it


kornelijepetak:
@daividssilverio I see you have put the `Cancel` on login screen, which is of course the screen the app reaches the moment you log out.
But what happens if:

- User clicks logout
- App crashes or user manages to kill it before reaching Login View?

daividssilverio:
Indeed @kornelijepetak, It might happen. I'll see what I can do to make it better.

PR #3183, opened 2018-09-06
Daneel 1.10 - Cil

simonrozsival:
Changelog:

```
- Fixed several bugs which can occur on slow networks (e.g., EDGE)
- Fixed a glitch with the workspace selection button in reports
- Fixed several glitches on the login and sign-up screens
- Fixed the calculation of total time tracked for each day on the main screen

- Improved stability and the synchronization with our servers
```


amulware:
Potentially useful, these are the changes since last release: https://github.com/toggl/mobileapp/compare/daneel-1.9-5...cil

PR #3189, opened 2018-09-06
Add idle time notifications

### LINKED ISSUE #3056: Add idle timer notification

dshr:
## What's this?
We're bringing back the persistent notifications! This one ties in with @daividssilverio's code for the running time entry notifications.

### Relationships
Closes #3056

## Why do we want this?
Clients miss them very much 😢 

## How is it done?
We're binding to the `IsTimeEntryRunning` observable on the MainViewModel and build the appropriate notification.

### Why not another way?
Seems like the simplest option to me 🤷‍♂️ 

### Side effects
Zilch

## Review hints
Enable this notification in your settings, lock your phone and rejoice ✨ . Start a new TE then, lock your phone again and it will be gone.

## :squid: Permissions
Go for it!


Toggl.Giskard/Helper/PersistentNotificationsHelper.cs
```diff
@@ -31,6 +32,18 @@ public static class PersistentNotificationsHelper
             activity.Bind(runningTimeEntryNotificationSource, te => updateRunningNotification(te, activity, notificationManager));
         }
 
+        public static void BindStoppedTimeEntry<T>(this T activity,
```
daividssilverio: This name feels odd for the functionality. It should be something related to being idle, right?

Toggl.Giskard/Helper/PersistentNotificationsHelper.cs
```diff
@@ -67,6 +80,36 @@ private static void updateRunningNotification(IThreadSafeTimeEntry timeEntryView
             }
         }
 
+        private static void updateStoppedNotification(bool shouldShow, Activity activity, NotificationManager notificationManager)
```
daividssilverio: Same, name it as something related to being idle.

Toggl.Giskard/Helper/PersistentNotificationsHelper.cs
```diff
@@ -67,6 +80,36 @@ private static void updateRunningNotification(IThreadSafeTimeEntry timeEntryView
...
+                var pendingIntent = PendingIntent.GetActivity(activity, 0, startMainActivityIntent, 0);
+
+                var notification = activity.CreateNotificationBuilderWithDefaultChannel(notificationManager)
+                    .SetShowWhen(true)
+                    .SetAutoCancel(false)
+                    .SetOngoing(true)
+                    .SetContentTitle(Resources.AppTitle)
+                    .SetContentText(Resources.IdleTimerNotification)
+                    .SetContentIntent(pendingIntent)
+                    .SetSmallIcon(Resource.Drawable.ic_icon_running)
```
daividssilverio: Resource icon is wrong, please import the svg `icon_notrunning` from Zeplin. You can use Android Studio to make your life easier.

Toggl.Giskard/Helper/PersistentNotificationsHelper.cs
```diff
@@ -17,6 +17,7 @@ public static class PersistentNotificationsHelper
     {
         private static Uri togglMainNavigationUri  = Uri.Parse("toggl://main");
         private static int runningTimeEntryNotificationId = 111;
+        private static int stoppedTimeEntryNotificationId = 112;
```
daividssilverio: rename it to suggest being idle.

daividssilverio:
Other than my comments, it works fine o/
We just have to clarify what that time should show. Right now it's showing the time the notification was put in the notification tray. If that's what we want, we don't need to set the `when` value. The default value should just work.

Toggl.Giskard/Helper/PersistentNotificationsHelper.cs
dshr: Good point, I went with `stopped` just because that's how I named it initially, but something along the lines of `IdleTimer` as in `BindIdleTimer` or `UpdateIdleTimerNotification` will be much clearer.

Toggl.Giskard/Helper/PersistentNotificationsHelper.cs
```diff
@@ -93,7 +93,6 @@ private static void updateStoppedNotification(bool shouldShow, Activity activity
                 var pendingIntent = PendingIntent.GetActivity(activity, 0, startMainActivityIntent, 0);
 
                 var notification = activity.CreateNotificationBuilderWithDefaultChannel(notificationManager)
-                    .SetShowWhen(true)
```
daividssilverio: Oh, that's not what I meant by not needing to set the `when` value. Sorry for not being clear.
You don't need to call `SetWhen`, but `SetShowWhen` is still necessary as it won't show anything in newer android versions. (it defaults to false in N and above).

Toggl.Giskard/Helper/PersistentNotificationsHelper.cs
```diff
@@ -67,6 +80,35 @@ private static void updateRunningNotification(IThreadSafeTimeEntry timeEntryView
             }
         }
 
+        private static void updateIdleTimerNotification(bool shouldShow, Activity activity, NotificationManager notificationManager)
+        {
+            if (notificationManager == null) return;
+
+            if (shouldShow)
+            {
+                var startTime = DateTime.Now.Ticks / TimeSpan.TicksPerMillisecond;
```
daividssilverio: 🔥 not being used

Toggl.Giskard/Helper/PersistentNotificationsHelper.cs
```diff
@@ -67,6 +80,35 @@ private static void updateRunningNotification(IThreadSafeTimeEntry timeEntryView
...
+                var startMainActivityIntent = new Intent(Intent.ActionView, togglMainNavigationUri)
```
daividssilverio: Oh, and please extract the intent creation logic to another method now that's it's being used in the same way both here and in the running te notification.

PR #3195, opened 2018-09-07
📨 receive notification when significant time changes

### LINKED ISSUE #3194: Timer page not refreshed when significant time changes
# 🐛 Bug Report
Timer page doesn't reload when significant time changes:

- Overnight
- Time zone changes

## Context

Build/Branch: `develop`
Phone OS: Any
Phone model: Any

## Expected behavior
- The te log should be updated properly

## To reproduce
- Change the timezone of the phone to Honolulu
- Create some time entries 
- Change the timezone to Brisbane

khoi:
## What's this?
Well this is supposed to be fixing the bug, but it's crashing due to some Xamarin issue 🤷🏻‍♂️. 

### Relationships

Closes #3194

## Why do we want this?
So the TEs display properly

## How is it done?
~Listen to a system event called `UIApplication.SignificantTimeChangeNotification` and then force a sync when it happens.~

A function that returns timezone's id is injected into `TimeService` for each platform. If the timezone changes the `SignificantTimeChange` will emit event and force reload the main TE log screen.

### Side effects
Not that I know of

## Review hints
You'll need a real device for this one.

- Switch the timezone to Honolulu
- Create some time entries
- Put the app to background
- Go to settings and change the timezone to Brisbane Australia
- Go back to the app and the TEs should be refreshed properly.

## :squid: Permissions
🐙


khoi:
Since the API is currently broken in Xamarin, I was thinking maybe we could force sync whenever the users put the app to foreground. 

Thoughts? @toggl/ios-senseis

simonrozsival:
I think what might be the easiest solution is to change what the `TimeService.MidnightObservable` does (and rename it): now it emits a value whenever the date between two ticks of the current time changes. It could check for that + if the difference between the two values is > 1s + some epsilon. If so, we should reload certain visual elements, because the displayed date/time might be incorrect.

Toggl.Foundation/TimeService.cs
```diff
@@ -12,33 +13,43 @@ namespace Toggl.Foundation
...
+
+            CurrentDateTimeObservable = ticker
                     .Select(_ => CurrentDateTime)
                     .Publish().RefCount();
 
             var localTimeObservable = CurrentDateTimeObservable
                 .Select(t => t.ToLocalTime());
 
-            MidnightObservable = localTimeObservable
+            var timeZoneId = ticker.Select(_ => timeZoneIdResolver());
```
juanlaube: Checking for timezone changes every second doesn't look like a good idea. I'd make the time service depend on the background service and do this check only when the app enters foreground.

khoi: Background service currently depends on time service, doing that would lead to circular dependency. But you're right, checking the timezone every tick is not ideal.

PR #3196, opened 2018-09-07
Use swipe gestures to close modal views and open edit view for currently running TE

simonrozsival:
## What's this?
This PR adds a better way of interacting with the iOS app. Swipe up on the currently running TE card to open the edit view. Swipe down to close it. It works the same way as with the close button and tapping on the background - it asks to discard/continue editing.

Swiping to close works with all of the other modal views correctly - by closing the same way as if you tapped on the close button.

### Relationships

There is no issue for this change - it has therefore low priority.

## Why do we want this?
It is really intuitive and we already discussed it a while back.

## How is it done?
Gesture recognizers.

The `Dismiss` function can return `false` to cancel the dismissing. That is necessary to cancel dismissing in the edit TE view when the user changes something and he does not want to lose his/her changes.

### Side effects

There are no side effects.

The problem might be in the future if some new modal view does not implement `IDismissableViewController` and then the view might not be closed properly.

## Review hints
Try this on a physical device.

## :squid: Permissions
As there is no issue for this change, let's not merge this until @pe0ny or @amulware approve it.


sofokli1:
The idea is a good one but UX feels a bit strange. I would expect the view to be draggable for as long as my finger is still touching the screen and animate all the way to the bottom of the screen when i remove the finger from the screen. I know this complicates things but it would make the interaction way smoother and intuitive. View disappearing under my finger as i am dragging it, flees off. Just my 2 cents, carry on.

simonrozsival:
I agree that the UX is far from perfect and that it would be much better if you could drag the modal and it would only be closed if you drag it far enough. It would require a bit more work to achieve that, so this could be a first step ("better than nothing") and later we wouldn't have to deal with the correct closing of the VMs.

There is no issue for this PR and it is just my initiative to implement this type of interaction with the app. We can close this and leave it until we have a better specification of how these gestures should work.

Odrakir:
There's a system API to do this properly, it's really not that complicated. Here's a tutorial: https://medium.com/@ludvigeriksson/custom-interactive-uinavigationcontroller-transition-animations-in-swift-4-a4b5e0cefb1e

simonrozsival:
I will close the PR for now and reopen it once I make the UX better.

simonrozsival:
Please check this PR now, if it is any better :)

amulware:
:doit:

PR #3205, opened 2018-09-10
Make UI switches larger

### LINKED ISSUE #3204: Make on/off switches larger
The settings screen switches are too small and often hard to hit after the #3151 change. They should be made larger.

The same applies to the "Billable" switch in the time entry edit view and the "Private project" switch in the project creation screen.

## Context
We are currently making the switches smaller to make them nicer. Apple normally uses bigger switches, which are easier to tap. These switches are not meant to be customizable, so it is not possible to make them smaller and increase the hit area of the switch without any hacky solutions.

Build/Branch: `develop`
Phone OS: `iOS`

## Expected behavior
The switches should be easier to hit.

simonrozsival:
## What's this?
This PR removes the shrinking of the on/off switches in the settings view (24/12 hours, manual mode), in the edit view (the billable flag), and in the project editing view (the private flag).

To make the switches behave consistently (follow-up of #3151), just the billable TE and private project switches are tappable instead of the whole rows (the normal Apple way).

### Relationships

Closes #3204

## Why do we want this?
We want the on/off switches be larger so they are easier to tap and to make them behave consistently throughout the app.

### Side effects
No side effects.

## :squid: Permissions
🦑


simonrozsival:
| Before | After |
| :-----: | :----: |
| screen shot 2018-09-10 at 12 24 16 pm | screen shot 2018-09-10 at 12 29 56 pm |

kornelijepetak:
In all honesty, the `before` looks much better IMO.

simonrozsival:
It looks better. But it looks different from other apps (e.g., Settings) and it is harder to tap it.

screen shot 2018-09-10 at 12 38 42 pm

kornelijepetak:
Hm... not an iPhone user. Strange decision on Apple's part. I always thought they did large paddings 😄 
Well, in any case, consistency is good. So, ... ok.

PR #3206, opened 2018-09-10
Siri cleanup logout

### LINKED ISSUE #3101: Clean up on logout
The following needs to be deleted:
- All donated shortcuts (even if they are not user specific I said it's better to be safe)
- All content in the shared space

Odrakir:
## What's this?
This takes care of cleaning up everything related to the Siri Extension when user logs out. This includes: 
- All data stored in the private shared storage
- All intents created

Closes #3101 

## Why do we want this?
We don't want a new user to accidentally use intents from the previous one.

## How is it done?
For the private storage is just an NSUserDefaults instance, so that's easy. For the intents there's a static method that deletes all of them, so that's even easier.

## :squid: Permissions
Do it!


khoi:
looks good, but add the tests to ensure the deletion is called

PR #3213, opened 2018-09-10
Stop auto scrolling when long press ends

### LINKED ISSUE #3201: Auto scrolling in the calendar view goes crazy
When the auto scrolling is happening and I release the pan gesture, sometimes it keeps scrolling. This is something I noticed at least twice today, however I don't have specific steps to reproduce the problem.

juanlaube:
## What's this?
Fixes problem with the auto scrolling in the calendar view. This only happened when editing a time entry and dragging the item when beginning the edit mode (i.e: in the same gesture to enter editing) and dragging up or down.

### Relationships

Closes #3201

## How is it done?
Adds the missing `StopAutoScroll` 🙈 

### Side effects
None.

## Review hints
Please test, I couldn't find another scenario where something like this is needed.

## :squid: Permissions
Whenever, anyone.


willsb:
Can you explain the line you 🔥 tho?

juanlaube:
I can. It was useless 😄

PR #3216, opened 2018-09-11
Increase navigation bar settings button hit area

### LINKED ISSUE #3178: The settings ⚙️ hitbox is too damn small
The settings ⚙️ hitbox is too damn small at least on non-phablet devices

## 🐛 Bug Report
It is very difficult to press the ⚙️ usually takes a few attempts to hit the right spot and there seems to be free real estate in the vicinity.

## Context
Calendar build from 05/09 but is consistent across versions I believe.

Build/Branch:
Phone OS: iOS
Phone model: 5E

## Expected behavior
To access settings

## To reproduce
Try to access settings

## Suggestions
Make it easier to access settings.
Bigger hitbox/es, please

simonrozsival:
## What's this?
This PR tries to increase the hit size of settings button in the navigation bar. The question is, how large should it be.. I decided to make it a square and set it to 40x40 (increase from 30x40).

### Relationships
Closes #3178 

## Why do we want this?
Increase the UX of the app.

## How is it done?
Simply change the dimensions of the button.

### Side effects
The image is centered in the button, which is now 40x40 instead of 30x40, so it appears shifted slightly towards the center (by 5pt) which is barely visible, but starts becoming more and more obvious for greater widths (50+).

## Review hints
Try it on a physical device. Is it easier to hit the btn?

## :squid: Permissions
🦑


amulware:
Square is correct in my estimation.

simonrozsival:
I tested increasing the size of the button vertically and it seems that setting it to 50 yields better results without changing the appearance of the btn.

willsb:
Could you instead target this at the `calendar-integration` branch and change the size of the calendar settings button as well? I will (probably) be included in the same release :)

juanlaube:
+1 to target `calendar-integration` and doing this for the _not-a-gear_ button in the calendar tab

simonrozsival:
Sure, I can do that. I will close this PR and open a new one.

PR #3217, opened 2018-09-11
Siri start intent handler

### LINKED ISSUE #3085: Implement start empty entry intent
The entry should be created from the extension, this should be the logic:
- Check if there's an api token and last sync timestamp in the shared space, other wise redirect the user to the app
- Check the last sync time, if it's been "too long", redirect to the app
- Get the latest entries from the API endpoint (if error redirect to app)

Open for discussion: What should happen if there's an entry running? Should it be stopped and then create the new one, or should we respond with an error? @amulware

khoi:
## What's this?
Add a siri handler for start timer intent.

### Relationships
Closes #3085

## Why do we want this?
Start timer without opening the app 💃 

## How is it done?

- New intent handler is added, which called the API to create a new TE
- In the case where user click the shortcut, the app will open the start te so users can fill out other TE information

### Why not another way?
Dazzle me

### Side effects
Not that I know of

## Review hints
- Setup the shortcuts for multiple workspaces
- Invoke the shortcuts by voice and by clicking on it.
- Test the case where the shortcut's workspace is not the default workspace.

## :squid: Permissions
Go for it


khoi:
tests are failing, checking

khoi:
test now passing

Odrakir:
Fix conflicts, please.

Toggl.Daneel.ExtensionKit/SharedStorage.cs
```diff
@@ -34,8 +35,16 @@ public void setNeedsSync(bool value)
             userDefaults.Synchronize();
         }
 
+        public void setUserId(double userId)
+        {
+            userDefaults.SetDouble(userId, UserIdKey);
```
Odrakir: It should also be removed in `DeleteEverything`

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
```diff
@@ -54,6 +56,7 @@ public sealed class StartTimeEntryViewModel : MvxViewModel<StartTimeEntryParamet
         private bool hasAnyTags;
         private bool hasAnyProjects;
         private long defaultWorkspaceId;
+        private string defaultWorkspaceName;
```
Odrakir: This would be better as a single property `defaultWorkspace` instead of two `defaultWorkspaceId` and `defaultWorkspaceName

PR #3220, opened 2018-09-11
Fix app freeze when opening from notification while app is in foreground

### LINKED ISSUE #3212: App freezes on Splash Screen when tapping the notification
## 🐛 Bug Report
App freezes when tapping the notification if it was already in the foreground.

## Context

Build/Branch: `giskard-notifications`
Phone OS: **Android 7.1.1** (not sure on other versions)
Phone model: **Android Emulator**

## To reproduce
* Open the app
* Make sure that both notification options are enabled in settings
* Start a TE (_the notification appears_)
* Pull down the notification/status bar
* Tap on the Toggl notification, making sure that Toggl app is currently in the foreground
* A new activity opens on top of the running app, starting with the splash screen
* The app freezes on Splash screen

## Suggestions
- Maybe an option with the PendingIntent that will make the intent reuse the app instead of opening a new one.
- Otherwise, maybe prevent the intent if the app is already in the foreground.

dshr:
## What's this?
Fixes the weird app freezes that we got when tapping on notifications while the app was in the foreground.

### Relationships
Closes #3212

## Why do we want this?
To cover the case where users decide to tap on a notification from us while the app is still open.

## How is it done?
The problem was related to the ApplicationUrls.Main Regex — it was always expecting an Action param, which we were not supplying. Since we don't want to take any action as well, it only made sense to make that param optional.

~I also tweaked the Intent flags a bit to try and skip the splash screen when opening from a notification — we don't need it since the notifications are shown only if the app is in memory.~
Keeping the flags as is, since the notification stays even if we kill the app.

### Why not another way?
Can't really think of any other options.

### Side effects
Nope.

## Review hints
Run the app, turn on the running timer notification and start a TE. Now swipe down to reveal the notifications bar, and click on our notification. The app should work fine. Then compare it to `neon`

Edit: Run the app via launcher and not VS — debugging breaks this for some weird reason.

## :squid: Permissions
All yours!


Toggl.Giskard/Helper/PersistentNotificationsHelper.cs
```diff
@@ -124,7 +124,7 @@ private static string extractProjectDetails(IThreadSafeTimeEntry timeEntryViewMo
         private static PendingIntent getIntentFor(Activity activity)
         {
             var startMainActivityIntent = new Intent(Intent.ActionView, togglMainNavigationUri)
-                .SetFlags(ActivityFlags.ClearTop | ActivityFlags.ClearTask | ActivityFlags.NewTask);
+                .SetFlags(ActivityFlags.NoAnimation);
```
daividssilverio: This doesn't seem to be fixing the problem.
It works if you keep clicking the notification with the application open.
But if you go back to home and open the application again from the launcher, it will freeze.

daividssilverio: The intent flags to clear any existing application tracer are there to prevent two MainActivities to be running at the same time. If you open the application from the launcher, then click the notification and press back again, it will reopen the MainActivity that was sent to the backstack.

dshr: Aw man, completely missed that 😞

daividssilverio:
And I just glanced my eyes again over your PR's description, and `we don't need it since the notifications are shown only if the app is in memory` isn't true. The ongoing notification will stay there even if you kill the app manually.

kornelijepetak:
After latest commit, currently the only way I managed to freeze the app was by attempting to open it by using the standard shortcut in the drawer, when the app was in the background and a time entry was running (with running notification being displayed). However, I have not managed to reproduce this consistently. Actually, I encountered this only once.

dshr:
> After the latest commit, the only way I managed to freeze the app was by attempting to open it by using the standard shortcut in the drawer, when the app was in the background and a time entry was running (with running notification being displayed). However, I have not managed to reproduce this consistently. Actually, I encountered this only once.

I'm getting this fairly consistently on my simulator if I start the app from the launcher, but not 100% of the time. If I start it in debug mode from VS, I get the freeze all the time. Very weird, and I would feel much more confident pushing this if we get some clarity on the cause.

OTOH, we want 1.4 out ASAP, so overthinking is not something we can do. Let's wait for @willsb to come online with the MVX mastery, have him have a look, and then we'll make a decision.

daividssilverio:
@dshr It seems that changing `RunAppStart` in the `SplashScreen.cs` does the job.
I'm not getting any errors related to navigation, nor freezes either starting from the notification or the launcher.
Maybe @willsb can tell if there's any harm in this approach.
```
protected override void RunAppStart(Bundle bundle)
        {
            base.RunAppStart(bundle);
            var navigationUrl = Intent.Data?.ToString();
            var navigationService = Mvx.Resolve<IMvxNavigationService>();
            if (string.IsNullOrEmpty(navigationUrl))
            {
                navigationService.Navigate<MainViewModel>();
                return;
            }

            navigationService.Navigate(navigationUrl);
        }

File: Toggl.Giskard/Activities/MainActivity.cs [hunk 2]
```diff
@@ -29,7 +29,8 @@ namespace Toggl.Giskard.Activities
     [MvxActivityPresentation]
     [Activity(Theme = "@style/AppTheme",
               ScreenOrientation = ScreenOrientation.Portrait,
-              ConfigurationChanges = ConfigChanges.Orientation | ConfigChanges.ScreenSize)]
+              ConfigurationChanges = ConfigChanges.Orientation | ConfigChanges.ScreenSize,
+              LaunchMode = LaunchMode.SingleTask)]
```
daividssilverio: You see, single task is not commonly used and is not really recommended in most scenarios. This makes the activity start a new task, becoming its root. Other than that, the main activity won't be recreated (new Intents won't go to onCreate, will go to onNewIntent instead) and that might change how our application works.
I'm not sure how this is gonna play with our architecture, mvx usage.

Have you tested this extensively? navigating a lot, letting the application be tombstone in other screens, moving to reports, settings, etc?

Toggl.Giskard/Helper/PersistentNotificationsHelper.cs
```diff
@@ -123,10 +123,10 @@ private static string extractProjectDetails(IThreadSafeTimeEntry timeEntryViewMo
 
         private static PendingIntent getIntentFor(Activity activity)
         {
-            var startMainActivityIntent = new Intent(Intent.ActionView, togglMainNavigationUri)
-                .SetFlags(ActivityFlags.NoAnimation | ActivityFlags.ClearTop | ActivityFlags.ClearTask);
-
-            return PendingIntent.GetActivity(activity, 0, startMainActivityIntent, 0);
+            var notificationIntent = activity.PackageManager.GetLaunchIntentForPackage(activity.PackageName);
```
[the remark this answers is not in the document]
dshr: This is basically to bypass the Splashscreen and not get the weird flicker when we click on a notification while the app is in the foreground.

Toggl.Giskard/Activities/MainActivity.cs
dshr: Yeah, true. Just tried without it, and it's working just as good.

daividssilverio: Then 🔥 please.

PR #3223, opened 2018-09-11
Reload calendar whenever user resumes the app from background

arturs105:
## What's this?
This adds code to refresh the calendar whenever the app is resumed from background

## Why do we want this?
Potential scenario:
1. User is using our app
2. He then opens up his iOS calendar to add/remove/edit some events
3. User goes back to Toggl app (which was sitting in the background nicely)
4. He doesn't see his new events in our app
5. The user goes to App store and gives us "1/5 Worst. App. Ever."

We don't want that.

## :squid: Permissions
Whenever.


Toggl.Foundation.MvvmCross/ViewModels/Calendar/CalendarViewModel.cs
```diff
@@ -159,29 +159,32 @@ public async override Task Initialize()
                 .EnabledCalendars
                 .SelectUnit();
 
-            var foregroundObservable = backgroundService
+            var appResumedFromBackgroundEachHourObservable = backgroundService
```
juanlaube: I don't think we should have these two separated observables. As a user, if we reload data in the UI, I want the notifications for the new events I see to be scheduled.

willsb: I think what Juan says actually makes sense. Since there's no way to know whether new events were added via observing, we might as well drop this one hour constraint.

PR #3225, opened 2018-09-11
Fix ugly bug with ugly fix

### LINKED ISSUE #3224: Calendar - Something is wrong with cells reusing
## 🐛 Bug Report

Sometimes, after I finish moving my time entries around, data is displayed in the wrong cells. This is probably some bad cells reusing or we're picking incorrect indices for a short period of time.

## Context

Build/Branch: calendar-integration
Phone OS: iOS
Phone model: Any

juanlaube:
## What's this?
Sometimes, after editing a TE in the calendar view, the data is reloaded multiple times, but the items are not always in the same order (because insertions, deletions, different start times, whatever...) causing the items in the UI to be rendered in different positions. This introduces a short delay to avoid this problem.

### Relationships

Closes #3224 

## How is it done?
By throttling 300 ms on items changed in the data source.

### Why not another way?
I couldn't find a better way to do this, if you have a better idea please let me know because this is ugly.

### Side effects
None.

## :squid: Permissions
Only for myself.


simonrozsival:
Throttling might not be a bad idea in this case 🤔 I can't really think of a better solution to the problem in general, if you can't make the order of the operations deterministic (which you probably can't). How did you come to 300 ms though? Isn't that too long? It must be noticeable to the user.

simonrozsival:
Sorry, closed this unintentionally 🤦‍♂️

juanlaube:
300 ms was enough to not notice the issue and still reload properly. I can probably reduce this time a little more.

juanlaube:
I knew there was a better way to do this! 😄 
Layout attributes were not correct when leaving edit mode, now the glitch is fixed by recalculating the layout attributes and invalidating the layout for all the visible items when edit is done.

willsb:
How do I test this?

juanlaube:
You can see the glitch easily when the items have different title or background colors. Have three different items and move one of them around. You can see the items "jumping" when ending the edit.

dshr:
> I knew there was a better way to do this! 😄
> Layout attributes were not correct when leaving edit mode, now the glitch is fixed by recalculating the layout attributes and invalidating the layout for all the visible items when edit is done.

Hm, the new way is def cleaner.

PR #3228, opened 2018-09-12
Siri start with description

### LINKED ISSUE #3227: Modify start intent to get a description parameter
After some thought I think this is going to be the best way to go. To start a timer for an already existing entry we will reuse the start timer intent adding the optional description parameter.

Odrakir:
## What's this?
Donates and handles an intent for starting a specific time entry with description

Closes #3227 

## Why do we want this?
So the user can start specific time entries with Siri.

## How is it done?
I'm just modifying the existing empty time entry to include an optional description parameter.

I also modify a bit the ‘StartTimeEntryViewModel’ so it can take an initial description, for when passing the intent to the app. 

## :squid: Permissions
Go ahead.


khoi:
now that we have description supported, we can also donate the interactions where user hit continue on an existing te

Odrakir:
Yes. That’s the idea. Push pending until I charge my laptop.

Toggl.Daneel.IntentBinding/StructsAndEnums.cs
```diff
@@ -3,7 +3,7 @@
 
 namespace Toggl.Daneel.Intents
 {
-    // [Watch (5,0), iOS (12,0)]
+    [Watch (5,0), iOS (12,0)]
```
khoi: I can build the project fine, do we need to remove this?

Toggl.Daneel/Startup/AppDelegate.cs
```diff
@@ -145,7 +145,12 @@ public override void PerformActionForShortcutItem(UIApplication application, UIA
         public override bool ContinueUserActivity(UIApplication application, NSUserActivity userActivity,
             UIApplicationRestorationHandler completionHandler)
         {
-            var intent = userActivity.GetInteraction()?.Intent;
+            var interaction = userActivity.GetInteraction();
+            if (interaction == null || interaction.IntentHandlingStatus != INIntentHandlingStatus.DeferredToApplication) {
+                return false;
```
khoi: nice

khoi:
when the users start an entry with description and also a project, the donation's desc is set to null

Odrakir:
🤔 Can't reproduce that, what's true is we are losing the project, maybe we should add more params...

khoi:
kapture 2018-09-14 at 13 56 34
here is the step that I did

khoi:
The bug I mentioned above is fixed now. Just found out another, hitting the "Working on these" suggestion doesn't trigger intent donation

PR #3237, opened 2018-09-12
UI Tests Groundwork

heytherewill:
## What's this?
This fixes our existing UI tests so that they are up-to-date with the current designs

## Why do we want this?
Now that @ApocalypticFish can run tests on his machine, it'd be wise to have our Unit tests in great shape so he can both use them and start writing new ones!

## How is it done?
I'm basically updating the tests to consider new layouts and adding accessibility names to files. Easy peasy.

## Review hints
Review per commit and run the tests locally using VS

## :squid: Permissions
Me


Toggl.Tests.UI/Extensions/SignUpExtensions.cs
```diff
@@ -0,0 +1,28 @@
+﻿using Xamarin.UITest;
+
+namespace Toggl.Tests.UI.Extensions
+{
+    public static class SignUpExtensions
```
kornelijepetak: I am a bit confused about the whitespace between statements in this file. Some tests have no blank lines between statements, some do. What is the convention we should use?

willsb: I try grouping method calls to mimic actual user actions. I sometimes fail :P

ApocalypticFish:
Test
TappingTheStopButtonStopsTheRunningTimeEntry
Is getting stuck because it is signing up with a new email, but it doesn't agree with the terms and conditions so gets stuck.

ApocalypticFish:
Some tests are fixed, but others remain broken. Squashing this won't break anything in Develop that isn't already broken. Would like to squash this in favour of smaller branches/PR's for individual tests or batches of tests being fixed in future and so that future tests can be created from Develop.
Please squash if possible and I'll create some issues to fix the tests that aren't working and branch and create PR's for them as appropriate.

PR #3245, opened 2018-09-13
Add Start Timer Intent Siri UI ## What's this?

khoi:
It displays a view with the description of the te users just started. 

### Relationships
Tick off #3122

## Why do we want this?
Since Xamarin is having the bug where Siri doesn't respond with the response message. 

## How is it done?
Trivial

### Why not another way?
We could also make the IntentViewController as a blankview, and then it embedded other the intents view controllers in as children (refer to Apple Soupchef example).
But I couldn't find a way to add a new ViewController to the SiriExtensionUI project.

We can go with it for now and once the Xamarin toolset for Xamarin become more stable. We can comeback and improve it.

### Side effects
🙈

## Review hints
Start a time entry using Siri shortcut.

## :squid: Permissions
Go for it


Odrakir:
All it's missing now is the time. Can we show a timer running in the extension UI? I'm not sure about it.

PR #3248, opened 2018-09-13
Bar chart logic

### LINKED ISSUE #2839: Reports: Bar chart logic
This is about adding the logic for showing bar charts that group TEs by time instead of project in our reports.

screen shot 2018-04-25 at 14 10 24

## Logical Details

1. if the user selects 1 to 34 days, they’ll see that many bars, one per day
2. if the user selects 35 to 182 days, they’ll see 5-26 bars, one per week
3. if the user selects more than 182 days, they’ll see 6-12 bars, one per month (we don’t support more than 1 year per report for performance reasons, but that’s nothing new)

- When user selects 1 to 7 days, display name of the weekday and date below each bar, for example: **M 1/4**. Dates should be formatted based on the date format in Profile settings but without the year.
- When user selects more than 7 days, display only start and end dates, for example: **1/4 - 30/4**

- Data is shown in two colors – dark #06AAF5 and light blue #47C3FC. The part that is dark blue shows billable hours. Light blue shows total hours. (Billable hours will display only if the WS is on the paid plan.)
- Show billable-nonbillable legend only if the WS is on a paid plan.

## Context
Please see the iOS UI issue #1958 for more UI details.

The more of the logic we can share between the platforms the better.

Please implement this in such a way that the data is not actually fetched yet until the specific app actually implements the feature.

simonrozsival:
## What's this?
This PR adds a bar chart view model which will be used from the `ReportsViewModel` and it contains all the information for displaying the bar chart.

### Relationships
Closes #2839

## Why do we want this?
We want more nice charts!

## How is it done?
The main idea is that this VM does not contain any logic for fetching data from the API, that will stay in the `ReportsViewModel`, and it only reacts to the reports it gets from an observable passed as an argument.

### Side effects
No side effects.

## :squid: Permissions
🦑


Toggl.Foundation.MvvmCross/ViewModels/Reports/ReportsBarChartViewModel.cs
```diff
@@ -0,0 +1,114 @@
...
+using Toggl.Multivac.Models.Reports;
+using static System.Math;
+
+namespace Toggl.Foundation.MvvmCross.ViewModels.Reports
+{
+    public sealed class ReportsBarChartViewModel : MvxViewModel
+    {
+        private const int maximumLabeledNumberOfDays = 7;
+
+        private const int verticalDivisions = 5;
```
daividssilverio: Was this number arbitrary?

Toggl.Foundation.MvvmCross/ViewModels/Reports/ReportsBarChartViewModel.cs
```diff
@@ -0,0 +1,114 @@
...
+            IObservable<ITimeEntriesTotals> reports)
+        {
+            Ensure.Argument.IsNotNull(schedulerProvider, nameof(schedulerProvider));
+            Ensure.Argument.IsNotNull(preferencesDataSource, nameof(preferencesDataSource));
+            Ensure.Argument.IsNotNull(reports, nameof(reports));
+
+            IsLoading = reports.Select(report => report == null)
+                .StartWith(true)
+                .DistinctUntilChanged()
+                .AsDriver(true, schedulerProvider);
```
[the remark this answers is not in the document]
khoi: this is probably because `AsDriver` will return `default(bool)` if you don't specify it. So it's going to be `false` which is opposite of what @simonrozsival wanted.

Toggl.Foundation.MvvmCross/ViewModels/Reports/ReportsBarChartViewModel.cs
```diff
@@ -0,0 +1,114 @@
...
+
+            DateFormat = preferencesDataSource.Current
+                .Select(preferences => preferences.DateFormat)
+                .AsDriver(onErrorJustReturn: defaultDateFormat, schedulerProvider: schedulerProvider);
+
+            var finalReports = reports.Where(report => report != null);
```
khoi: where does `reports` come from? maybe we need a `shared` operator here to avoid extra work

Toggl.Foundation.MvvmCross/ViewModels/Reports/ReportsBarChartViewModel.cs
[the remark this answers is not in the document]
simonrozsival: I will make those parameters named.

Toggl.Foundation.MvvmCross/ViewModels/Reports/ReportsBarChartViewModel.cs
simonrozsival: No, comes from the specification in the issue(s) and from the way webapp uses it (always 5 vertical lines) - see the Notion docs.

Toggl.Foundation.MvvmCross/ViewModels/Reports/ReportsBarChartViewModel.cs
simonrozsival: I think I will change this a bit so it takes just an `Unit` observable which will tell it to start fetching data.

simonrozsival: ... or not. It would include a lot more complicated changes in the `ReportsViewModel`.

simonrozsival: Do you think that there would be a difference if we share the observable?

khoi: depends on where you inject `reports` from? but actually, adding `share` here probably makes sense, since you don't have to care whether `reports` is heavy lifting job or not.

Toggl.Foundation.MvvmCross/ViewModels/Reports/ReportsBarChartViewModel.cs
daividssilverio: I found it in the Toggl Business Logic > Workspaces > Reports > Report Exports. Maybe we should duplicate this in the Mobile reports docs as well?

simonrozsival:
I changed the VM a bit and started working on the UI and I think it will be easier to close this PR and open a new one with all of the changes.

PR #3253, opened 2018-09-14
Add peek and quick actions to calendar

arturs105:
## What's this?
This adds peek and ~pop~ with a delete quick action. There is no pop, because i didn't want to spend much time on this (also, no designs for the pop thingy). We can extend this in the future by adding the pop and more quick actions.

## Why do we want this?
I just wanted to learn how to implement peek and pop. I also thought this would make it easier to quickly delete entries from the calendar. Since it's already implemented, why not merge it 🤷‍♂️.

## How is it done?
This is a quick experiment - please feel free to point out any 💩 parts of the code.

## Review hints
Press down hard on a time entry in the calendar. Either use a MacBook with the new 3D touch touchpad or an iPhone with 3D touch.

## :squid: Permissions
Whenever.


willsb:
I'm afraid this can't be merged into `calendar-integration`, but I'd love to see it rebased into develop once we have the main branch merged there!

willsb:
Could you please rebase this into develop, just so it doesn't stale out? I'd love to reopen the discussion about peek and pop and this is a great use case!

willsb:
I've created an issue (#3378) for properly designing this feature so that we can peek and pop every time entry in the app. We can use this branch as a base implementation.

PR #3257, opened 2018-09-14
Fix invalid (empty) layout when date changes

juanlaube:
## What's this?
Sooooo, I was accidentally able to reproduce that bug ~from the upside down~ Jordan found where the calendar screen is just a white view and I think it is related to a day change not taken into account by the collection view layout. Unfortunately I was only able to reproduce this once so I can't confirm this actually fixes the problem.

### Relationships

Closes nothing

## How is it done?
By subscribing to the midnight observable in the layout and invalidating the layout.

### Side effects
None.

## :squid: Permissions
Whenever, anyone.


Toggl.Daneel/Views/Calendar/CalendarCollectionViewLayout.cs
```diff
@@ -62,6 +62,13 @@ public CalendarCollectionViewLayout(ITimeService timeService, ICalendarCollectio
 
             date = timeService.CurrentDateTime.Date;
 
+            timeService
+                .MidnightObservable
+                .SelectUnit()
+                .Do(_ => date = timeService.CurrentDateTime.Date)
```
arturs105: Doesn't `MidnightObservable` already return a `DateTimeOffset` that you can use?

juanlaube: 🤦‍♂️

PR #3273, opened 2018-09-18
Siri entry params

### LINKED ISSUE #3259: Siri: Start correct entry for Start entry intent.
Right now we are only passing the description through the intent. So when starting it with Siri it'll start an entry with that description but with now project or tags.

It probably makes sense to pass the entry id so the extension can download all the info before creating the new entry.

Odrakir:
## What's this?
This sends some more parameters to the intent when user starts a specific entry.

Closes #3259 

## Why do we want this?
So user can start with Siri entries that have a project or tags without loosing that information.

## How is it done?
After playing around with the sorting of the parameters I realized there's a bug that makes some parameters not work unless they are of Custom type. So in the end that's what I had to do for `projectId`, `tags` and `billable`

Also, when any `projectId` or `tag` has a negative id we ignore the donation and wait for the user to repeat that action with all properties synced.

## :squid: Permissions
Hit it!


khoi:
when users invoke the shortcut that has project or tag information, the app opens but only the description is pre-filled.

khoi:
I think you need to rebase the PR, there are changes that don't belong

Odrakir:
I just needed to push `siri-shortcuts` again.

Odrakir:
Please, check again, @khoi .

Toggl.Daneel/Startup/AppDelegate.cs
```diff
@@ -177,6 +177,21 @@ public override void PerformActionForShortcutItem(UIApplication application, UIA
             }
         }
 
+        private StartTimeEntryParameters createStartTimeEntryParameters(StartTimerIntent intent)
+        {
+            var tags = (intent.Tags.Count() == 0) ? null : intent.Tags.Select(tagid => (long)Convert.ToDouble(tagid.Identifier));
```
khoi: Invoking the intent manually with no tag will 💥

Toggl.Daneel/Services/IntentDonationService.cs
```diff
@@ -21,12 +22,24 @@ public void DonateStartTimeEntry(IWorkspace workspace, ITimeEntry timeEntry)
...
+                // If any of the tags or the project id were just created and haven't sync we ignore this action until the user repeats it
+                if (timeEntry.ProjectId < 0 || timeEntry.TagIds.Any(tagId => tagId < 0)) 
+                {
+                    return;
+                }
+
                 intent.EntryDescription = timeEntry.Description;
+                intent.ProjectId = new INObject(timeEntry.ProjectId.ToString(), timeEntry.ProjectId.ToString());
+                intent.Tags = timeEntry.TagIds.Select(tag => new INObject(tag.ToString(), tag.ToString())).ToArray();
+                intent.Billable = new INObject(timeEntry.Billable.ToString(), timeEntry.Billable.ToString());
```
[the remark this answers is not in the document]
Odrakir: 🤔 Not sure about that. The fact is this is a hack because using an Integer doesn't work (due to some bug, I'm afraid). Putting the same value in the `identifier` and `displayString` of the INObject makes it easier to decode, because any property would do.

Toggl.Daneel/Services/IntentDonationService.cs
```diff
@@ -21,12 +22,24 @@ public void DonateStartTimeEntry(IWorkspace workspace, ITimeEntry timeEntry)
...
+            if (!string.IsNullOrEmpty(timeEntry.Description))
             {
+                // If any of the tags or the project id were just created and haven't sync we ignore this action until the user repeats it
+                if (timeEntry.ProjectId < 0 || timeEntry.TagIds.Any(tagId => tagId < 0)) 
+                {
+                    return;
+                }
+
                 intent.EntryDescription = timeEntry.Description;
+                intent.ProjectId = new INObject(timeEntry.ProjectId.ToString(), timeEntry.ProjectId.ToString());
```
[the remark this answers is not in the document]
khoi: I mean when you don't set the project of the TE. The `timeEntry.ProjectId` is `null`, and calling `ToString` on it returns `""`, so I was wondering whether we should check that.

khoi:
The code looks OK and the `NSUserActivity` is working properly now. 
However, I still couldn't trigger the intent handler where the TE has a project included, it fails for me every time while all other cases work.

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
```diff
@@ -297,13 +300,24 @@ public override async Task Initialize()
...
-            else 
+            textFieldInfo = TextFieldInfo.Empty(parameter?.WorkspaceId ?? defaultWorkspace.Id);
+
+            if (initialParameters != null)
             {
-                textFieldInfo = TextFieldInfo.WithDescription(parameter?.WorkspaceId ?? defaultWorkspace.Id, initialDescription);
+                List<ISpan> spans = new List<ISpan>();
+                spans.Add(new TextSpan(initialParameters.EntryDescription));
+                if (initialParameters.ProjectId != null) {
+                    var project = await dataSource.Projects.GetById((long)initialParameters.ProjectId);
```
simonrozsival: This might throw: when we donate the intent to Siri, but the user deletes the project. You should catch and leave it empty, if it throws an exception.

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
```diff
@@ -297,13 +300,24 @@ public override async Task Initialize()
...
+                    spans.Add(new ProjectSpan(project));
+                }
+                if (initialParameters.TagIds != null) {
+                    var tags = initialParameters.TagIds.ToObservable()
+                        .SelectMany<long, IThreadSafeTag>(tagId => dataSource.Tags.GetById(tagId))
```
simonrozsival: The same problem might be with tags. I think you should create an interactor which safely fetches a tag by ID (+ the same for projects).

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
```diff
@@ -297,13 +300,24 @@ public override async Task Initialize()
...
-            {
-                textFieldInfo = TextFieldInfo.Empty(parameter?.WorkspaceId ?? defaultWorkspace.Id);
-            }
-            else 
+            textFieldInfo = TextFieldInfo.Empty(parameter?.WorkspaceId ?? defaultWorkspace.Id);
+
+            if (initialParameters != null)
             {
-                textFieldInfo = TextFieldInfo.WithDescription(parameter?.WorkspaceId ?? defaultWorkspace.Id, initialDescription);
+                List<ISpan> spans = new List<ISpan>();
```
simonrozsival: `var`

Odrakir:
@khoi There was a bug, please check now.

PR #3274, opened 2018-09-18
⏰ refresh certain TE elements when significant time change detected

### LINKED ISSUE #3194: Timer page not refreshed when significant time changes

khoi:
### Relationships

Closes #3194
Supercedes #3195

## Why do we want this?
Just like how we reload the main log every midnight, we might want to do it when significant time changes, for instance: timezone updated.

## How is it done?
Forwarding call in `AppDelegate` to `TimeService`

### Why not another way?
I used to fetch the timeone info inside `TimeService` every `tick`, but that's not very performant. Let the system handle it is probably the easiest way.

### Side effects
🤷🏻‍♂️

## Review hints
- Change the timezone when the app is running
- Check whether reload is called

## :squid: Permissions
🐙


Toggl.Foundation/TimeService.cs
```diff
@@ -52,6 +57,11 @@ public Task RunAfterDelay(TimeSpan delay, Action action)
                 .ToTask();
         }
 
+        public void SignificantTimeChanged()
```
juanlaube: `=>` this method!

Toggl.Foundation.MvvmCross/ViewModels/MainViewModel.cs
```diff
@@ -248,7 +248,12 @@ public override async Task Initialize()
                 .DisposedBy(disposeBag);
 
             timeService.MidnightObservable
-                .Subscribe(onMidnight)
+                .SelectUnit()
+                .Subscribe(onSignificantTimeChange)
+                .DisposedBy(disposeBag);
+
+            timeService.SignificantTimeChangeObservable
+                .Subscribe(onSignificantTimeChange)
```
juanlaube: `.VoidSubscribe` here and you can ignore the `Unit` param in `onSignificantTimeChange`

Toggl.Daneel/Startup/AppDelegate.cs
```diff
@@ -137,6 +140,11 @@ public override void PerformActionForShortcutItem(UIApplication application, UIA
             }
         }
 
+        public override void ApplicationSignificantTimeChange(UIApplication application)
```
juanlaube: `=>` this method!

willsb: Actually, void methods should not use EBM :P

Toggl.Daneel/ViewControllers/MainViewController.cs
```diff
@@ -210,6 +210,10 @@ public override void ViewDidLoad()
                 SendFeedbackSuccessView.Rx().AnimatedIsVisible());
             this.BindVoid(SendFeedbackSuccessView.Rx().Tap(), ViewModel.RatingViewModel.CloseFeedbackSuccessView);
 
+            ViewModel.ShouldReloadTELog
```
willsb: Properly named properties do not increase the binary size. Don't write `TE`, write `TimeEntry`

PR #3276, opened 2018-09-18
Prevent notification sounds in Oreo/Pie

### LINKED ISSUE #3275: Notification sounds should be off in Android Oreo
## 🐛 Bug Report

The idle timer and running TE notifications should not emit sound when appear.

So, here's what's going on.
We have done something wrong when doing notifications previously.
There was a breaking change ([bug](https://issuetracker.google.com/issues/65108694)) in Oreo: if Priority is specified as `Default` for a notification channel, it instead assumes a priority of `High` in API 26. Google does not seem to intend a fix 😢.

The problem for us and our users is that once the notification channel is created, it **CANNOT** be changed. Subsequent calls to change the priority are ignored. To reset this to actual default, a reinstall with the updated app (that fixed this issue) will be needed, unfortunately. 

However, there's a workaround (read below).

## Workaround

There is a workaround that can be suggested for users who'd rather do it by setting something than by reinstalling the app.

* When a notification appears, long press it
* Open the settings for that notification
* See that the priority for Toggl channel is `High`
* Disable the notification there and reenable it.
* Now the notification will show the priority of `Medium` and will not play the sound any more.

## Context

Build/Branch: `develop`
Phone OS: Only Android 8 (Oreo) and Android 9 (Pie) 
_Older versions of Android seem to be working well._

## Expected behavior

The channel created should have a priority of LOW, which actually shows on phone as `medium`, and is defined in documentation as: `Shows everywhere, but is not intrusive`. 

## To reproduce
* Make sure your volume is up.
* Have an Oreo/Pie
* Turn on notifications in the settings.
* Enjoy the sounds :)

kornelijepetak:
## What's this?

Fixes the Oreo problems with notification emitting the sound.
Please take a look at the issue to properly understand this solution.

### Relationships
Closes #3275 

## Why do we want this?
Cause the sounds are annoying for these notifications.

## How is it done?
Read the issue to understand.

### Why not another way?
Because Google thinks it's better to work on a new Pixel phone than fix their own bugs.

### Side effects
Zilch.

## Review hints
Read the issue!!!

## :squid: Permissions
Yeah, whatever.

### DIFF
--- a/Toggl.Giskard/Extensions/ActivityExtensions.cs
+++ b/Toggl.Giskard/Extensions/ActivityExtensions.cs
@@ -74,7 +74,7 @@ namespace Toggl.Giskard.Extensions
         {
             if (OreoApis.AreAvailable)
             {
-                var channel = new NotificationChannel(defaultChannelId, defaultChannelName, NotificationImportance.Default);
+                var channel = new NotificationChannel(defaultChannelId, defaultChannelName, NotificationImportance.Low);
                 channel.Description = defaultChannelDescription;
                 channel.EnableVibration(false);
                 channel.SetVibrationPattern(noNotificationVibrationPattern);


kornelijepetak:
For those who didn't read the issue in the end:

Users who already have the app installed will have to either
* Reinstall the app
* Use the workaround from #3275

daividssilverio:
The change itself it's pretty minimal. I'll run a test in the emulator.

PR #3278, opened 2018-09-18
Fix creating new tags and projects with emoji from the start time entry view

### LINKED ISSUE #3268: The start view does not allow to create projects and tags with text and emoji directly
## 🐛 Bug Report
Follow-up of #3265 

The problem is that text and emoji sequences are split into separate `TextSpan`s. The "Create project XYZ" uses only the first `TextSpan` as XYZ, so it will either be just emoji, or text.

This might be possibly a problem with Japanes, Chinese and other non-latin character sets, but I haven't tested that.

## Context
Build/Branch: `develop`
Phone OS: `iOS`

## Expected behavior
The whole sequence of normal chars and emoji should be used as the text of the project/tag name.

## To reproduce
Type in `abc #abc 👻 def` into the text field in the star time entry view. It should show you `Create tag "abc 👻 def"`, but it does not show tag suggestions at all.

## Suggestions
"Simply" collapse all consecutive `TextSpan` and `QueryTextSpan` into a single `TextSpan` (or a `QueryTextSpan` if there is a cursor positioned in any of the text spans).

simonrozsival:
## What's this?
The iOS parsing of the input text in the start time entry view is buggy. User's can't create 

### Relationships
Closes #3268 

## Why do we want this?
We want the app to be easy to use and emoji-friendly! (... and non-latin characters friendly).

## How is it done?
Sequences of `TextSpan` and `QueryTextSpan` are merged into a single `TextSpan` or `QueryTextSpan`. This way the suggestions work correctly again.

### Why not another way?
This is the simplest way - using a `foreach`. If somebody can implement this with just one LINQ statement (which will still be readable), please let me know your idea.

### Side effects
This should also fix Japanese and maybe some other languages, which use non-latin characters.

## Review hints

Try typing this into the input field: "Brew another ☕️ @☕️  Coffee breaks". The suggestion should be to create a project "☕️ Coffee breaks" (unless the user has one already), but it will be just "☕️" without this fix, because emojis and text are split into different spans and only the span containing the @ will be used for suggestions, and in this case the span contains just the @ symbol.

## :squid: Permissions
🦑


simonrozsival:
The problems with emoji/Japanes is strictly platform specific - it is splitting `NSAttributedString` into `ISpan`s, where it produces several `TextSpan`s in a row, which is fixed by collapsing them with this PR. And that is tested.

PR #3279, opened 2018-09-18
Add suggestions setting in the options screen

kornelijepetak:
## What's this?
Adds an option to show/hide suggestions from the options screen.
Until we make different suggestion providers, it will help declutter the main screen.

`needs-design` label added: Copy `Show suggestions on the main screen` may not be the best copy. Summoning @pe0ny.

## Why do we want this?

Users wanted it. I want it.

## How is it done?

The typical way.

### Side effects
I don't think so, but yell if you find something!

## :squid: Permissions
To exercise control, only me.

### DIFF
--- a/Toggl.Giskard/Adapters/MainRecyclerAdapter.cs
+++ b/Toggl.Giskard/Adapters/MainRecyclerAdapter.cs
@@ -30,6 +30,8 @@ namespace Toggl.Giskard.Adapters
 
         public SuggestionsViewModel SuggestionsViewModel { get; set; }
 
+        public bool IsShowingSuggestions { get; set; }
+
         private Subject<TimeEntryViewModel> timeEntryTappedSubject = new Subject<TimeEntryViewModel>();
         private Subject<TimeEntryViewModel> continueTimeEntrySubject = new Subject<TimeEntryViewModel>();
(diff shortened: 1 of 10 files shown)


Toggl.Giskard/Adapters/MainRecyclerAdapter.cs
```diff
@@ -51,7 +53,7 @@ public void DeleteTimeEntry(int position)
             deleteTimeEntrySubject.OnNext(deletedTimeEntry);
         }
 
-        public override int HeaderOffset => 1;
+        public override int HeaderOffset => IsShowingSuggestions ? 1 : 0;
```
daividssilverio: 🙏 Let's not change the header offset dynamically o/ This can cause bad access in the underlying collection.
Instead, let's maybe update the `MainLogSuggestionsListViewHolder` and the adapter inside of it.
It uses another strategy already for when there's no suggestion available (hidding the view, and setting the VH to 1px).

willsb:
I'm putting the `on-hold` label on this so we know we have to wait.

simonrozsival:
I suggest closing this PR and referencing it form a follow-up issue, @kornelijepetak. Let's clean our pull requests page.

kornelijepetak:
One of these days I will suggest an improvement for this to accomodate future needs. For now I am closing the PR but please do not delete the branch (for now).

PR #3280, opened 2018-09-18
Fix crash in edit time entry viewmodel (maybe)

### LINKED ISSUE #3199: NullReferenceException in EditTimeEntryViewModel.isDirty
## 💥 Crash Report
Something is null when `isDirty` is called from `closeWithConfirmation`.

https://github.com/toggl/mobileapp/blob/133c046cb7398e65ad5fdfcf356bb942a7fe02fb/Toggl.Foundation.MvvmCross/ViewModels/EditTimeEntryViewModel.cs#L56-L66

### 👥 Impact
- 165 occurrences to 138 users, 19% of crashes in 1.9
- all sorts of devices and iOS versions

### 🔗 Context
**I think this bug was introduced with 1.9.** It seems to not happen for earlier versions. Maybe that helps in finding the issue.

### 🛰 Link
https://appcenter.ms/orgs/Toggl/apps/Toggl-iOS/crashes/groups/5e9ebaf763d160e0722cdc73c43298606b31a513?period=last7Days

```cs
EditTimeEntryViewModel.get_isDirty ()
EditTimeEntryViewModel+<closeWithConfirmation>d__160.MoveNext ()
TaskAwaiter.ThrowForNonSuccess (System.Threading.Tasks.Task task)
TaskAwaiter.HandleNonSuccessAndDebuggerNotification (System.Threading.Tasks.Task task)
TaskAwaiter.ValidateEnd (System.Threading.Tasks.Task task)
ConfiguredTaskAwaitable+ConfiguredTaskAwaiter.GetResult ()
MvxAsyncCommandBase+<ExecuteConcurrentAsync>d__17.MoveNext ()
TaskAwaiter.ThrowForNonSuccess (System.Threading.Tasks.Task task)
TaskAwaiter.HandleNonSuccessAndDebuggerNotification (System.Threading.Tasks.Task task)
TaskAwaiter.ValidateEnd (System.Threading.Tasks.Task task)
ConfiguredTaskAwaitable+ConfiguredTaskAwaiter.GetResult ()
MvxAsyncCommandBase+<ExecuteAsync>d__16.MoveNext ()
TaskAwaiter.ThrowForNonSuccess (System.Threading.Tasks.Task task)
TaskAwaiter.HandleNonSuccessAndDebuggerNotification (System.Threading.Tasks.Task task)
TaskAwaiter.ValidateEnd (System.Threading.Tasks.Task task)
ConfiguredTaskAwaitable+ConfiguredTaskAwaiter.GetResult ()
MvxAsyncCommandBase+<Execute>d__14.MoveNext ()
AsyncMethodBuilderCore+<>c.<ThrowAsync>b__6_0 (System.Object state)
NSAsyncSynchronizationContextDispatcher.Apply ()
```

juanlaube:
## What's this?
An attempt to solve the crash reported in #3199. I wasn't able to reproduce the crash, but this could solve the problem.

### Relationships

Closes #3199

## How is it done?
We couldn't reproduce the crash, but I will follow @simonrozsival theory in the issue about `originalTimeEntry` being garbage collected at some point as the cause of this crash, because I don't see anything else as possible. If that's true, this should simply return true when `originalTimeEntry` is garbage collected.

### Why not another way?
I think this is the only thing we can try right now.

### Side effects
If it's true that `originalTimeEntry` is being garbage collected at some point, this will return true even if there are no changes.

## Review hints
Since we couldn't reproduce the crash, suggestions about other causes are welcome.

## :squid: Permissions
Only me.


simonrozsival:
Let's merge this and see if we stop getting those crashes. If the crashes go away, then 🎉 , otherwise we can revert this change, reopen the issue, and try again.

PR #3283, opened 2018-09-18
Fix crash in main log caused by removed sections

### LINKED ISSUE #3137: ArgumentOutOfRangeException in TimeEntriesLogViewSource
## 💥 Crash Report
Something goes wrong when updating the main table view. When `EndUpdates` is called, the state of the view source and the view are not compatible and it goes 💥.

### 👥 Impact
- 113 times for 70 users
- 35% of crashes in 1.9
- all sorts of devices and iOS versions

### 🔗 Context
Some more fallout from the rxfactor?

### 🛰 Link
https://appcenter.ms/orgs/Toggl/apps/Toggl-iOS/crashes/groups/31b1a1881ca095645fd2fd9fe059ce0c33871599?appVersion=1.9

`System.ArgumentOutOfRangeException: Index was out of range. Must be non-negative and less than the size of the collection. Parameter name: index`
```cs
TimeEntriesLogViewSource.GetViewForHeader (UIKit.UITableView tableView, System.nint section)
(wrapper managed-to-native) ObjCRuntime.Messaging.objc_msgSendSuper(intptr,intptr)
UITableView.EndUpdates ()
ReactiveTableViewBinder`2[TModel,TCell].handleCollectionChanges (System.Collections.Generic.IReadOnlyCollection`1[T] changes)
AnonymousSafeObserver`1[T].OnNext (T value)
ObserveOn`1+Context+_[TSource].OnNextPosted (System.Object value)
NSAsyncSynchronizationContextDispatcher.Apply ()
```

juanlaube:
## What's this?
Fixes a crash in the main log tableview when entire sections are removed. This can happen when from the web or another device all time entries in a section are removed, or when the date changes and the total visible days is different.

### Relationships

Closes #3137

## How is it done?
By adding another collection change to remove an entire section.

### Why not another way?
I think another possible way was to keep the empty sections, that would have caused the data source  to return 0 items for that section, but doesn't feel right.

### Side effects
None that I can think of.

## Review hints
Please test, try to reproduce the issue first then try the fix.

## :squid: Permissions
Only me.


juanlaube:
Good point! I will verify and do the changes if needed, adding the `wip` label meanwhile.

simonrozsival:
Well. This PR does not fix the problem at all :-/

Steps to reproduce:
- start with a fresh acount
- add two TEs with the same start date using the web app
- sync the mobile app
- delete both TEs in the web app
- sync the mobile app
- 💥 

The root of this problem is in `ReactiveTableViewBinder` on lines 93 - 96:
```csharp
affectedSections
    .Distinct()
    .Where(index => index < dataSource.NumberOfSections(tableView))
    .ForEach(index => dataSource.RefreshHeader(tableView, index));
```

juanlaube:
Superseded by #3287

PR #3284, opened 2018-09-18
Track selected text extension

willsb:
## What's this?
This adds Toggl to the contextual menu that opens whenever text is selected on Android.

## Why do we want this?
Because this makes the mobile app behave somewhat like the Toggl Button (merging seamlessly in the user's flow instead of forcing them to open our app) and makes it even simpler to create new time entries!

## How is it done?
I created a new URL that allows one to start arbitrary text as new time entries. I then added a new `IntentFilter` that adds Toggl to the list of apps that can handle text selections.

### Why not another way?
This is pretty straightforward and moves the heavy lifting to Foundation, making the code Xplat and allowing iOS (and even third party apps) to benefit from arbitrary time entry creation.

### Side effects
None that I am aware of, but we'd better test this!

## Review hints
Install this on an actual device and test it for yourself.

## :squid: Permissions
Me or whoever is responsible for the next android release, since this is low priority.


amulware:
No issue or designs? I see in the video posted internally that it adds the 'Track' option to the text selection menu, where is that copy taken from?

willsb:
As per a discussion I had with @amulware, I'm introducing a new section in Notion for documenting our platform specific features. You can see this feature documented [here](https://www.notion.so/Android-f8904cca9af744c7ab7465c08b1f681e).

amulware:
🦑

PR #3285, opened 2018-09-19
Add better Hebrew rendering

### LINKED ISSUE #3270: Actual Hebrew text showing wrong alignment
## 🐛 Bug Report
Time Entry descriptions in Hebrew are being pushed to the right of of the main Timer screen. 

## Context
Previously addressed through https://github.com/toggl/mobileapp/issues/2685 but still showing up for 1.3.2 Giskard user

Build/Branch: 1.3.2 Giskard
Phone OS: Android 7.1.1 N
Phone model: Sony E6633

## Expected behavior
The Time Entry description text should be aligned to the left, regardless of what direction it is in.

## Suggestions
Please view screenshot sent by user on /conversation/18538880329

It looks like the text is aligned to the right, against the time and continue button.

dshr:
## What's this?
This will 'fix' RTL language rendering and layout. 

### Relationships
Closes #3270

## Why do we want this?
To make RTL text readable

## How is it done?
By changing the `gravity` so our text will always start from the left. 

### Why not another way?
This is the easiest way so far. Ideally we'd do a separate design for RTL languages, since the whole horizontal hierarchy is flipped around, but for the time being this will do just fine!

### Side effects
_Yok_

## Review hints
Add a Hebrew KB to your device, and then create a TE with some RTL text.

## :squid: Permissions
All yours!


juanlaube:
Can you please Williamize the XML? Yes, that's a thing 😄

dshr:
> Can you please Williamize the XML? Yes, that's a thing 😄

Is there a williamisation guideline doc?

dshr:
Doesn't build tho. The Williamiser plugin broke the XML 😞

Issue #3251, opened 2018-09-13
MvxAndroidViewPresenter.ShowIntent -- Exception of type 'Java.Lang.RuntimeException' was thrown.


daividssilverio:
## 💥 Crash Report
This seems to be happening at random.
It's not hard to replicate, all you have to do is to keep creating time entries with new projects and clients.
This crash will eventually happen when you press the Play FAB.

### 👥 Impact
This has happened to 147 times, affecting 123 users so far on 1.3.2

### 🔗 Context
Found this trying to replicate #3239 
I tested on Android 6.0.

### 🛰 Link
https://appcenter.ms/orgs/Toggl/apps/Toggl-Android/crashes/groups/d2403985e56732996a05a35071e918b1c5b16a1d?appVersion=1.3.2

### 📝  Exception & Stack Trace
`MvxAndroidViewPresenter.ShowIntent (Android.Content.Intent intent, Android.OS.Bundle bundle)
Java.Lang.RuntimeException: Exception of type 'Java.Lang.RuntimeException' was thrown.`

This is what we get from AppCenter:
```cs
ExceptionDispatchInfo.Throw ()
JniEnvironment+InstanceMethods.CallNonvirtualVoidMethod (Java.Interop.JniObjectReference instance, Java.Interop.JniObjectReference type, Java.Interop.JniMethodInfo method, Java.Interop.JniArgumentValue* args)
JniPeerMembers+JniInstanceMethods.InvokeVirtualVoidMethod (System.String encodedMember, Java.Interop.IJavaPeerable self, Java.Interop.JniArgumentValue* parameters)
ContextWrapper.StartActivity (Android.Content.Intent intent, Android.OS.Bundle options)
MvxAndroidViewPresenter.ShowIntent (Android.Content.Intent intent, Android.OS.Bundle bundle)
MvxAndroidViewPresenter.ShowActivity (System.Type view, MvvmCross.Platforms.Android.Presenters.Attributes.MvxActivityPresentationAttribute attribute, MvvmCross.ViewModels.MvxViewModelRequest request)
MvxAndroidViewPresenter.<RegisterAttributeTypes>b__21_0 (System.Type view, MvvmCross.Presenters.Attributes.MvxBasePresentationAttribute attribute, MvvmCross.ViewModels.MvxViewModelRequest request)
MvxAndroidViewPresenter.Show (MvvmCross.ViewModels.MvxViewModelRequest request)
[...]

Issue #3152, opened 2018-09-04
Giskard crashes on edit screen after going to another sub screen, going home and back


daividssilverio:
## 💥 Crash Report
It looks like we are having some trouble with the mvx navigation when going to sub screens.
The app will crash if you:
1. click on an existing TE to edit
2. click the project selection box
3. check your projects, select one (or not)
4. still in the projects screen, or any other subscreen after that, press home
5. go back to the app
6. try pressing UP (<-) or the back button.
7. navigation will either not work or you will get the crash below

### 👥 Impact
Well, this affects anyone trying to create/select a project/client/task in the edit view and for some reason checking something else in another app or just by having a low-end phone.

### 🔗 Context
There was one at some point but I couldn't find it. The one about creating a project not redirecting the user back to the edit view.

### 🛰 Link
Didn't find it.

### 📝  Exception & Stack Trace
`Exception type and message`
```cs
android.runtime.JavaProxyThrowable: System.NullReferenceException: Object reference not set to an instance of an object
                         E    at Toggl.Foundation.MvvmCross.ViewModels.EditTimeEntryViewModel.get_isDirty () [0x00000] in <44bf0fd1c67d4889808313f6903411d3>:0
                         E    at Toggl.Foundation.MvvmCross.ViewModels.EditTimeEntryViewModel+<closeWithConfirmation>d__163.MoveNext () [0x0001a] in <44bf0fd1c67d48898
                            08313f6903411d3>:0
                         E  --- End of stack trace from previous location where exception was thrown ---
                         E    at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw () [0x0000c] in <b0b794627e4d420a98d0f9ff8a6f9f8a>:0
                         E    at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess (System.Threading.Tasks.Task task) [0x0003e] in <b0b794627e4d420a98d0f9
                            ff8a6f9f8a>:0
                         E    at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification (System.Threading.Tasks.Task task) [0x00028] in <b
                            0b794627e4d420a98d0f9ff8a6f9f8a>:0
                         E    at System.Runtime.CompilerServices.TaskAwaiter.ValidateEnd (System.Threading.Tasks.Task task) [0x00008] in <b0b794627e4d420a98d0f9ff8a6f9
                            f8a>:0
[...]

PR #3291, opened 2018-09-19
Safer droid dialog service

### LINKED ISSUE #3250: Unable to add window -- token android.os.BinderProxy is not valid; is your activity running?
## 💥 Crash Report
I got to reproduce this by simple trying to create a lot of time entries, one after the other and pressing back (system back, the triangle) accidentally.

### 👥 Impact
It has happened 95 times and impacted 82 of our users on 1.3.2.

### 🔗 Context
It happens you are trying to create a TE and presses back after creating a Project/Tag.
The problem seems to be related to the fact that our DialogService is giving us the wrong TopActivity when is time to create the dialog to confirm discarding a time entry.
`var activity = Mvx.Resolve<IMvxAndroidCurrentTopActivity>().Activity;`
this activity was finishing when I was testing this.

### 🛰 Link
https://appcenter.ms/orgs/Toggl/apps/Toggl-Android/crashes/groups/fe2fa681aa45490dbdbe05d330aca74e26ad494a

`Java.Lang.RuntimeException: Unable to add window -- token android.os.BinderProxy@eb1a353 is not valid; is your activity running?`
```cs
2 JniEnvironment+InstanceMethods.CallVoidMethod (Java.Interop.JniObjectReference instance, Java.Interop.JniMethodInfo method, Java.Interop.JniArgumentValue* args)
3 JniPeerMembers+JniInstanceMethods.InvokeVirtualVoidMethod (System.String encodedMember, Java.Interop.IJavaPeerable self, Java.Interop.JniArgumentValue* parameters)
4 Dialog.Show ()
5 DialogService+<>c__DisplayClass0_1.<Confirm>b__1 ()
6 Thread+RunnableImplementor.Run ()
7 IRunnableInvoker.n_Run (System.IntPtr jnienv, System.IntPtr native__this)
8 (wrapper dynamic-method) System.Object.7a77e245-1e1a-4461-9358-f394e86bcccb(intptr,intptr)
9 android.view.WindowManager$BadTokenException: Unable to add window -- token android.os.BinderProxy@5facaff is not valid; is your activity running?
10 android.view.ViewRootImpl.setView(ViewRootImpl.java:800)
11 android.view.WindowManagerGlobal.addView(WindowManagerGlobal.java:356)
12 android.view.WindowManagerImpl.addView(WindowManagerImpl.java:93)
13 android.app.Dialog.show(Dialog.java:330)
14 mono.java.lang.RunnableImplementor.n_run(Native Method)
15 mono.java.lang.RunnableImplementor.run(RunnableImplementor.java:30)
16 android.app.Activity.runOnUiThread(Activity.java:6163)
17 md5a44fea7db10133e9f6bddee3bde9d640.StartTimeEntryActivity.n_onKeyDown(Native Method)
18md5a44fea7db10133e9f6bddee3bde9d640.StartTimeEntryActivity.onKeyDown(StartTimeEntryActivity.java:65)
...
```

daividssilverio:
## What's this?
This changes try to make sure the activity used on the Giskard's DialogService is the actual current top and non finishing activity on the Mvx lifecycle.

### Relationships

Closes #3250

## Why do we want this?
Sometimes, the current top activity in `IMvxAndroidCurrentTopActivity` is not providing a safe activity to use when creating dialogs (WindowToken is invalid) (or even the wrong top activity);

This happens sometimes and will make the app crash when leaving views when you have made changes or is trying to logout.

## How is it done?
This PR adds another querying method inside `QueryableMvxLifecycleMonitorCurrentTopActivity` to find the current top activity that is not being finished.
The non finishing current activity is used in the DialogService to in the dialog builder.

### Why not another way?
A nice way to this would be changing the dialog service to be more tied to IMvxViews and somehow get hold of it inside the ViewModels doing something like:
```
await dialogService.Confirm(
                    this.view,
                    Resources.WorkspaceChangedAlertTitle,
                    Resources.WorkspaceChangedAlertMessage,
                    Resources.Ok,
                    Resources.Cancel
                )
```
and then the actual implementation of the view wouldn't have to query the current top activity/fragment/whatever, and it should be safe to open the dialog from the view.

### Side effects
If somehow, no current activity can be found on `QueryableMvxLifecycleMonitorCurrentTopActivity`, the user will get stuck in the screen.
One possible way to handle that would be just ignoring the dialog and executing the action that would happen if the user confirmed on the dialog and then showing a Toast telling the user what happened, like: "Changes discarded", "Signed out" when the worst happens.

## Review hints
Bear with me, reproducing the crash is not that easy, it seems it won't happen on every device. It does sometimes on my S8 with Android 8.0.
Try navigating a lot and executing actions that will trigger the confirmation dialog.
I did: Start TE -> Select Project -> Create Project -> Close -> Cancel -> Edit TE -> Select Project -> Create Project -> Close -> Cancel a bunch of time.

Please note that other crashes might happen because the navigation is buggy right now.
You might get #3251, please ignore that for the purposes of this review.
You might also get into #3152, so avoid leaving the app until that one is merged.

## :squid: Permissions
🦑 whenever


Toggl.Giskard/Services/DialogService.cs
```diff
@@ -3,30 +3,37 @@
...
+        private readonly long delayBeforeLookingForTopActivityAgain = 300;
+
         public IObservable<bool> Confirm(string title, string message, string confirmButtonText, string dismissButtonText)
         {
-            var activity = Mvx.Resolve<IMvxAndroidCurrentTopActivity>().Activity;
-
             return Observable.Create<bool>(observer =>
             {
-                activity.RunOnUiThread(() =>
+                void ShowDialogIfActivityIsThere(Activity activity)
```
juanlaube: lowercase pls

Toggl.Giskard/Services/DialogService.cs
```diff
@@ -43,12 +50,33 @@ public IObservable<bool> Confirm(string title, string message, string confirmBut
...
-                });
+                }
+
+                TryToRunOnTopActivity(ShowDialogIfActivityIsThere);
 
                 return Disposable.Empty;
             });
         }
 
+        private void TryToRunOnTopActivity(Action<Activity> action)
```
juanlaube: lowercase pls

daividssilverio:
@amulware @pe0ny I'll be merging this to prevent further crashes.
But the side effects might leave the user stuck in some screen where it used to crash (not always).
What do you think about the workaround to show a Toast telling the user what happened, like: "Changes discarded", "Signed out" when the worst happens.

Issue #1728, opened 2018-03-15
Find and fix bottlenecks in the start time entry suggestions


heytherewill:
A user complained of sluggishness when typing the description of a time entry 🙊 

We need to stress test the app by trying to create a time entry in a workspace with a bunch projects/clients/tasks and see how bad the situation is. 
This needs to be done in an actual device (preferably an iPhone 6/SE).

After that we need to measure performance and find possible points of improvement.

amulware:
I have this issue on an iPhone 6s. I don't even have a lot of data, but typing just lags behind a little bit and it's not super pleasant.

amulware:
I tried recording a video (posted internally), but it's hard to see it there since even the keyboard lags behind my typing. I noticed that this does not seem to be a problem in a test account with little data, but only in my real account with hundreds of time entries, all with nice descriptions and all that.

For testing, try putting a Delay(500) or so into the (realm) query that fetches the suggestions? The UI shouldn't freeze, no matter how long the query takes. If needed, we can always delay the query itself until the user stops typing too.

sofokli1:
I am not having any luck with this so I am leaving this to someone else more familiar with observables and threading.

My finding are the following:
I think the issues is in the `Prepare` func line 264 of the StartTimeEntryViewModel.
I tried adding 2 ObserveOn calls as below
```csharp
querySubject.AsObservable()
                .StartWith(textFieldInfo)
                .ObserveOn(Scheduler.Default)
                .Select(QueryInfo.ParseFieldInfo)
                .Do(onParsedQuery)
                .SelectMany(autocompleteProvider.Query)
                .Merge(queryByTypeObservable)
                .ObserveOn(SynchronizationContext.Current)
                .Subscribe(onSuggestions)
                .DisposedBy(disposeBag);
```
This crashes with the error message

The problem is that autocompleteProvider.Query loads all the projects and tags for every typed character using some interactor. This constant loading and filtering seems to be an issue when having too many data.

Other approach that I can suggest here is to have some kind of delay before performing the query. the delay resets every time the user changes the text. if no changes after the delay passes we perform the query.

Alternatively it might make sense to have some more filters when fetching data. Instead of getAll we could fetch those that were used recently (like past 2 weeks) which is a reasonable assumption.

khoi:
Adding `Delay` as @amulware suggested also causes the exception `Realm accessed from incorrect thread.` 
```
commit d18969b4115c3035835cafc15b8c940ccf8e8956
[...]

PR #3300, opened 2018-09-21
Nice database migrations

simonrozsival:
## What's this?
This PR simplifies the way we implement DB migrations and makes it easier to add new migrations in the future.

### Relationships

No directly related issue.

## Why do we want this?

Adding a migration is confusing at the moment, because we have a single method which has several if branches, some of them empty, and somebody who is not familiar with this code might be confused at first and might make a mistake, which would result in corrupting user's data if it was overlooked during code review.

## How is it done?

Migrations, which are independent, will now be separate classes with single responsibility: execute the migration from an old schema to the target schema version.

The `Database` class holds a list of migrations, which should be used, and applies them in a correct order (from lowest target schema versions to the highest).

There might be several classes targeting the same schema, but there must be no dependence between the migrations as the order of them is not guaranteed in any way.

Creating migrations will still be quite challenging and will require a lot of focus (when the schema changes, some of the old migrations might need to be replaced with different ones (e.g., when some property is renamed twice).

### Why not another way?
This is a nice and clean way which should be future-proof.

### Side effects
No side effects.

## Review hints

Try creating your own migration and change some of the `Realm*` classes and bump the schema version, then run the app and use Realm Studio (or the debugger) to see that the migration was applied correctly.

## :squid: Permissions

I will 🦑 this if we agree that it is a good thing to have.

### DIFF
--- a/Toggl.PrimeRadiant.Realm/Database.cs
+++ b/Toggl.PrimeRadiant.Realm/Database.cs
@@ -86,7 +86,20 @@ namespace Toggl.PrimeRadiant.Realm
                     migrations
                         .Where(migration => oldSchemaVersion < migration.TargetSchemaVersion)
                         .OrderBy(migration => migration.TargetSchemaVersion)
+                        .Do(ensureApplicabilityOfMigrationFrom(oldSchemaVersion))
                         .ForEach(migration => migration.PerformMigration(realmMigration.OldRealm, realmMigration.NewRealm))
             };
-    }
+
+        private Action<IMigration> ensureApplicabilityOfMigrationFrom(ulong oldSchemaVersion)
+            => migration =>
+            {
+                if (oldSchemaVersion < migration.MinimumSchemaVersionToMigrateFrom)
+                {
+                    throw new InvalidOperationException(
+                        $"The old schema is too old: {migration.GetType().FullName} " +
+                        $"can be used to migrate from schema versions {migration.MinimumSchemaVersionToMigrateFrom}-{migration.TargetSchemaVersion - 1} " +
+                        $"to schema {migration.TargetSchemaVersion}, but it was supposed to migrate from version {oldSchemaVersion}.");
+                }
+            };
+}
 }
--- a/Toggl.PrimeRadiant.Realm/Database.cs
+++ b/Toggl.PrimeRadiant.Realm/Database.cs
@@ -82,13 +82,11 @@ namespace Toggl.PrimeRadiant.Realm
             => new RealmConfiguration
             {
                 SchemaVersion = 6,
-                MigrationCallback = (migration, oldSchemaVersion) =>
-                {
+                MigrationCallback = (realmMigration, oldSchemaVersion) =>
                     migrations
-                        .Where(m => oldSchemaVersion < m.TargetSchemaVersion)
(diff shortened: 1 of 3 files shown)


amulware:
I'm not sure if I like it. What if we change tags again in the future and we need to partially replace the existing migration depending on version numbers? This also makes it very hard to see what migrations we have and what they do. Lastly, we're trying to improve startup performance, not the other way around (even if this will probably be a very tiny difference).

Also, YAGNI.

Toggl.PrimeRadiant.Realm/Database.cs
```diff
@@ -75,30 +82,24 @@ private RealmConfiguration createRealmConfiguration()
...
-                    if (oldSchemaVersion < 6)
-                    {
-                        // nothing needs explicit updating when updating from schema 4 up to 6
-                    }
+        private Action<IMigration> ensureApplicabilityOfMigrationFrom(ulong oldSchemaVersion)
+            => migration =>
+            {
+                if (oldSchemaVersion < migration.MinimumSchemaVersionToMigrateFrom)
+                {
+                    throw new InvalidOperationException(
```
amulware: So we'll make it crash for some users at runtime in a pretty untestable way? (We should have tests running all possible migrations, ideally with checking data integrity.)

Toggl.PrimeRadiant.Realm/Migrations/TagServerDeletedAt.cs
```diff
@@ -0,0 +1,23 @@
+using System.Linq;
+
+namespace Toggl.PrimeRadiant.Realm.Migrations
+{
+    public sealed class TagServerDeletedAt : IMigration
+    {
+        public ulong MinimumSchemaVersionToMigrateFrom { get; } = 0;
+
+        public ulong TargetSchemaVersion { get; } = 4;
```
amulware: `=>` please.

Toggl.PrimeRadiant.Realm/Database.cs
```diff
@@ -12,6 +14,11 @@ public sealed class Database : ITogglDatabase
     {
         private readonly RealmConfiguration realmConfiguration;
 
+        private readonly IMigration[] migrations =
```
amulware: Allocating a thing that we will hardly ever use? :/ At least make it lazy? Or better a calculated property that returns a new array only when we're actually migrating.

amulware:
I think the thing we have to optimise for with the migration stuff is (future) correctness. Not ease of writing, and maybe not even ease of reading, if it gets in the way.

simonrozsival:
I can't see any implications on startup performance. Do you to the array of migration instances? That could be lazy loaded, if we want to optimize that.

I would argue that it is easier to see what migrations we have - they now have names (name of the class).

If we need to replace some migration, we simply remove the old class and replace it with a more sophisticated version. That is the same as with the `if-if-if` "pattern", you would also have to rewrite that code.

Ad YAGNI - I just started working on my proposed solution for #1728 to see, if it would help, and adding just a few migrations makes the code quite chaotic (it gets very long).

simonrozsival:
I will close this for now and reopen/introduce again when we need to add some new migrations.

PR #3305, opened 2018-09-21
🚧 Project creating suggestion obey workspace settings

### LINKED ISSUE #2961: Only allow creation of projects and clients if workspace setting allows it
Workspaces have a setting whether project and client creation is allowed for all users, or just admins.

We should obey that setting, and check whether the workspace allows creation for all, and if not, if the user is an admin (or the other way around, whatever makes more sense in code).

Relevant documentation on permissions is under Business Logic / Global Logic / Access Levels.

We only want to show valid workspaces when creating a project/client.

If the user ONLY belongs to workspaces where they cannot create them, there are two options:
1. not show the 'create project ABC' in the start/edit views at all
2. when the user taps the same, we check and notice that there are no valid workspaces, so we show them a popup instead informing them of the same

I'm not sure which is more user-friendly, or generally better.

khoi:
## What's this?
If the users belong to a workspace where they can't create new project due to project settings, we should disable creation suggestion.

### Relationships
Closes #2961

## Why do we want this?
Law abiding citizen

## How is it done?
Adding a check to see whether the user is admin, and also check the `OnlyAdminsMayCreateProjects` setting.

### Side effects
Not that I know of.

## Review hints
- Join a workspace where the setting below is set to false.

screen shot 2018-09-21 at 18 13 45

- Check the suggestion behavior in Start and Edit TE view

## :squid: Permissions
:octopus:


Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
```diff
@@ -75,6 +76,9 @@ public bool SuggestCreation
         {
             get
             {
+                if (!defaultWorkspace.Admin && defaultWorkspace.OnlyAdminsMayCreateProjects)
```
willsb: Is this correct, though? This will also prevent the suggestion of tag creation. Is this the intended behaviour?

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
```diff
@@ -53,7 +54,7 @@ public sealed class StartTimeEntryViewModel : MvxViewModel<StartTimeEntryParamet
 
         private bool hasAnyTags;
         private bool hasAnyProjects;
-        private long defaultWorkspaceId;
+        private IThreadSafeWorkspace defaultWorkspace;
```
willsb: Are TEs always created in the default WS? 
Can't users open the start view and select a project in a different WS where they can create projects? Shouldn't we suggest project creation in such cases?

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
khoi: no it doesn't, I can still create tag fine :|

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
amulware: TEs can be created in any workspace.
We should suggest project creation unless there are no workspaces where the same is allowed for this user.
The user can then select one of the valid workspaces in the project creation view.

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
amulware: Bizlogic wise, I think all users should be allowed to create tags. Though I can see many reasons why we wouldn't want them. Please check this with the documentation and if it doesn't specify, ask product (and ping me there).

khoi: The documentation doesn't specify whether users could create tags in such case. I tried the webapp and the users can create tags fine. I'll ask the product team about the case and update here if there is something change.

khoi: I asked Syko and he said users should be able to create tags normally @amulware

Toggl.Foundation.MvvmCross/ViewModels/EditProjectViewModel.cs
```diff
@@ -106,7 +106,11 @@ public override void Prepare(string parameter)
 
         public override async Task Initialize()
         {
-            var workspace = await interactorFactory.GetDefaultWorkspace().Execute();
+            var defaultWorkspace = await interactorFactory.GetDefaultWorkspace().Execute();
+            var allWorkspaces = await interactorFactory.GetAllWorkspaces().Execute();
+            var workspace = defaultWorkspace.IsEligibleForProjectCreation()
+                ? defaultWorkspace
+                : allWorkspaces.First(ws => ws.IsEligibleForProjectCreation());
```
willsb: This will throw (and consequently crash the app) if no workspace is eligible for Project creation. Is this by design?

Toggl.Multivac/Extensions/WorkspaceExtensions.cs
```diff
@@ -0,0 +1,12 @@
+using Toggl.Multivac.Models;
+
+namespace Toggl.Multivac.Extensions
+{
+    public static class WorkspaceExtensions
+    {
+        public static bool IsEligibleForProjectCreation(this IWorkspace self)
+        {
+            return self.Admin || !self.OnlyAdminsMayCreateProjects;
```
willsb: Why not `=>`?

simonrozsival:
@khoi Why is this on-hold now?

willsb:
:doit:

PR #3299, opened 2018-09-20
Giskard main log diffutil

daividssilverio:
## What's this?
This is basically a rewrite of the update logic used in the `ReactiveSectionedRecyclerAdapter`, the base adapter of the Main log for Giskard.

Still Missing:
- [x] Move the diff calculation to a background thread
- [x] Add some queueing device for the updates

### Relationships
Closes #3295
Closes #3288

Cousin of #3137

## Why do we want this?
1. We don't like giskard 💥💥 ing when the main log gets updated.
2. DiffUtil makes the main updates nice if done properly (if the callbacks and checks are all correct).

## How is it done?
We stop relying on the `ObservableGroupedOrderedCollection` `CollectionChange`'s types and simply use android's helper class DiffUtil + DiffUtil.Callback to identify and notify the adapter the changes for us.

We basically have a copy of the data source at all times that the adapter uses for displaying the time entries and sections. Whenever there's a change, big or small, the differences between the lists (the copy and the original list) are calculated by the DiffUtil based on the criteria expressed in `HeaderOffsetAwareDiffCallback`; the list used to display the time entries is then updated and the changes are dispatched to the adapter all at once. The three lines below do everything there is to do.

```
var diffResult = DiffUtil.CalculateDiff(new HeaderOffsetAwareDiffCallback(...));
currentItems = newImmutableItems;
diffResult.DispatchUpdatesTo(this);
```

### Why not another way?
Because this is easy and reliable. Using this strategy, the adapter always have a backing list of the items it's displaying, so there's no bad access. 

### Side effects
The animations when adding/removing time entries should work just fine, but the sections still need a better way to get identified. So you might see weird updates on the sections if you try it right now.

## Review hints
Avoid reviewing from the diff page from Github, it looks funny.
There's no much point in reviewing by commit.
Please try to break it. 

## :squid: Permissions
~Not yet. We still need to move the diff calculation to a background thread and defer the updates to the main thread; A queueing method for changes also is still missing.~
🦑 whenever o/ Updates are running in the background;
Concurrent updates are handled by dropping update requests when an update is running and a single update is run at the end.

PR #3308, opened 2018-09-24
Add HashCode T4 template for variable number of type parameters

kornelijepetak:
## What's this?
We are lazy and prone to errors when adding new overloads.

### Relationships
Could be useful for #3299 

## Why do we want this?
Read the text under `What's this`.

## How is it done?
T4 to the rescue. For now I think 16 will be enough. That's a number .NET used to make `Tuple<T1, T2, ... T16>` and also `Func<T1, T2...>`.

Also, I have changed the algorithm and the numbers to something that Jon Skeet uses 😄 
Also, there is a suggestion that this number [92821 has some properties](https://stackoverflow.com/questions/1835976/what-is-a-sensible-prime-for-hashcode-calculation/2816747#2816747) making it a good number for the multiplication.

### Why not another way?
Because boxing.

### Side effects
Just a few more bytes in the build size. Like, literally.

## Review hints
Just don't get into a mistake of checking each type parameter one by one.

## :squid: Permissions
When @daividssilverio @simonrozsival @amulware approve.
Also, @willsb is welcome as well :)
And me.


willsb:
And yes, @daividssilverio, it will removed the unused ones :)

willsb:
Buttons are ridiculously hard, somehow

PR #3311, opened 2018-09-27
👩‍🏫 no running te returns a success instead of an err

khoi:
## What's this?
This modifies the Stop TE intent so that when there is no current TE, the handler returns success instead of an exception.

## Why do we want this?
- It'll be easier for the user to use the intent as an action in a whole ~workflow~ shortcut.

## How is it done?

- Update the intent definition file

### Side effects
None

## Review hints
- Try to run the intent as part of a big shortcut, make sure the shortcut doesn't fail when there is no current TE.

## :squid: Permissions
🦑


willsb:
@Odrakir y u no review dis

PR #3317, opened 2018-10-01
Fix keyboard notification lifecycle

willsb:
## What's this?
This fixes a problem introduced by #3314

### Relationships
Related to #3314

## Why do we want this?
The lifecycle is currently messed because if we push a new VC, when we return to the previous one it won't receive keyboard notifications.

## How is it done?
Only clean things when disposed

### Why not another way?
Good question, that's why I want a bunch of reviews

## :squid: Permissions
Me


Toggl.Daneel/ViewControllers/KeyboardAwareViewController.cs
```diff
@@ -7,21 +7,28 @@ namespace Toggl.Daneel.ViewControllers
...
         protected KeyboardAwareViewController(string nibName)
             : base(nibName) { }
 
         public override void ViewDidLoad()
         {
             base.ViewDidLoad();
 
-            UIKeyboard.Notifications.ObserveWillShow(KeyboardWillShow);
-            UIKeyboard.Notifications.ObserveWillHide(KeyboardWillHide);
+            willShowNotification = UIKeyboard.Notifications.ObserveWillShow(KeyboardWillShow);
```
khoi: I don't think we need to keep the references since 
```
NSNotificationCenter.DefaultCenter.RemoveObservers(this)
```
will remove everything

willsb:
Please don't merge pull requests you don't have permission to merge

khoi:
oops sorry about that

PR #3330, opened 2018-10-03
♻️ reload report if user has not looked at it for some minutes

### LINKED ISSUE #3073: Reload reports if the user has not looked at them for X minutes
Reload them when the user goes back to them only of course.

The reason is that now with the tabs, a report might stay open for days while the app is in the background, so the data gets out of date.

We could maybe also show a little text at the top 'this report was generated (on X(only if not the same day)) at Y' and have a refresh button?

khoi:
## What's this?
Reload the report 📊

### Relationships
Related to #3073

## Why do we want this?
To ensure the freshness of the report screen.

## How is it done?
The time when the report last loaded is saved to a property, and is checked when the View reappear again.

### Side effects
I don't think there is any.

## Review hints

- Navigate to report screen
- Let the app idle for 5 minutes
- Navigate back 

## :squid: Permissions
🐙


willsb:
I'm with @simonrozsival on this one. Without a decent force refresh alternative (like pull-to-refresh), this amount of time is simply not good enough.

simonrozsival:
I suggest to reload every time for now + not close the original issue and add an comment which suggests using pull-to-refresh to reports. We can then revisit the issue in the future.

Toggl.Foundation.Tests/MvvmCross/ViewModels/ReportsViewModelTests.cs
```diff
@@ -745,5 +745,33 @@ private void prepareWorkspace(bool isProEnabled)
...
+            {
+                await ViewModel.Initialize();
+                ViewModel.ViewAppeared(); // First call is skipped
+
+                for (int i = 0; i < numberOfAppearances; ++i)
+                {
+                    ViewModel.ViewAppeared();
+                }
+
+                ReportsProvider.GetProjectSummary(Arg.Any<long>(), Arg.Any<DateTimeOffset>(),
```
simonrozsival: Shouldn't you mock this before calling `ViewAppeared`?

Toggl.Foundation.Tests/MvvmCross/ViewModels/ReportsViewModelTests.cs
```diff
@@ -745,5 +745,33 @@ private void prepareWorkspace(bool isProEnabled)
...
+                        Arg.Any<DateTimeOffset>())
+                    .ReturnsForAnyArgs(Observable.Empty<ProjectSummaryReport>(SchedulerProvider.TestScheduler));
+
+                ReportsProvider.Received(numberOfAppearances).GetProjectSummary(Arg.Any<long>(), Arg.Any<DateTimeOffset>(),
```
simonrozsival: I think you should call `TestScheduler.Start();` before doing assertions 🤔

PR #3331, opened 2018-10-03
Fix headers timezones

### LINKED ISSUE #3262: Invalid date headers since Daneel 1.9
Several users have reported issues with invalid date headers since Daneel 1.9 was out. One user has confirmed the issue is present also with 1.10, hence creating the issue. 

Reporting issue based on recent example, will add other examples as they come in. 

## 🐛 Bug Report
User's timezone is set to America/Los_Angeles, their device's timezone matches. Some of their recent time entries that were recorded on September 16th as per time in LA are displayed as September 17th. 

## Context
Build/Branch: Daneel 1.9
Phone OS: iOS 11.4.1
Phone model: iPhone 7

conversations/18519489605
User ID 3570297

## Expected behavior
If user's devices timezone and profile timezone match and the user has recorded time entries on day A, time entries should be displayed in their app as being recorded on day A. 

## To reproduce
Have not been able to reproduce yet, but reporting regardless as the number of issues is high.

simonrozsival:
## What's this?
This fixes a bug where we display incorrect headers for people using timezones which are far from UTC-0.

### Relationships

Closes #3262 

## Why do we want this?

We want the app to work for everyone no matter what their race, religion, or timezone is.

## How is it done?
Consistent usage of local time, instead of the UTC time.

### Why not another way?
This is the correct way. Maybe there are other places where we use local/utc times incorrectly, but this PR deals just with bug.

### Side effects
No side effects.

## Review hints
Create a time entry for two days starting every hour. Try with different time zone settings. The headers on `develop` branch should be incorrect and on this branch correct (Today + Yesterday).

## :squid: Permissions
🦑


simonrozsival:
This PR changes both platforms so I will also need somebody from @toggl/androids to review.

Please feel free to suggest better solutions if you have some ideas :)

simonrozsival:
We could merge this into Sirius and release it in 1.5.1

willsb:
@daividssilverio Can you please take a look at this?
Also, can you ping @simonrozsival so he can cherry pick this fix into the android hotfix release branch, once we have one?

PR #3335, opened 2018-10-04
📲 add 3d touch for suggestion view

### LINKED ISSUE #1724: Open start view if 3d-touching suggestion
This is a proposal for now, but I had the idea that when one 3d-touched (i.e. with more force) a suggestion, it could open the start view with the usual 3d touch animation. That would allow the user to start the suggestion but also quickly modify it. And it would be a cool feature.

Ref @pe0ny

khoi:
## What's this?
When users force touch the suggestion view, the start view will appear with all the params prefilled. 

### Relationships
Closes #1724

## Why do we want this?
Allow the users to quickly modify some information.

## How is it done?
Using some 3d touch API.

### Why not another way?
If there is a way to make it cleaner using Mvx, tell me.

### Side effects
I can't think of any

## Review hints
- Please test it on a real device if possible.

## :squid: Permissions
Me

### DIFF
--- a/Toggl.Daneel/ViewControllers/MainViewController.cs
+++ b/Toggl.Daneel/ViewControllers/MainViewController.cs
@@ -665,7 +665,7 @@ namespace Toggl.Daneel.ViewControllers
             previewingContext.SourceRect = suggestionView.Frame;
 
             var suggestion = suggestionView.Suggestion;
-                var param = new StartTimeEntryParameters(DateTimeOffset.Now, suggestion.TaskName,
+            var param = new StartTimeEntryParameters(DateTimeOffset.Now, suggestion.TaskName,
                     suggestion.Duration, suggestion.WorkspaceId, suggestion.Description, suggestion.ProjectId,
                     suggestion.TagIds);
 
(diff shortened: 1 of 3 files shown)


Toggl.Daneel/ViewControllers/MainViewController.cs
```diff
@@ -630,5 +648,44 @@ private void reload()
...
+
+            if (suggestionView == null)
+            {
+                return null;
+            }
+
+            previewingContext.SourceRect = suggestionView.Frame;
+
+            var suggestion = suggestionView.Suggestion;
+            var param = new StartTimeEntryParameters(DateTimeOffset.Now, suggestion.TaskName,
```
simonrozsival: Please use `ITimeService` to get current time.

Toggl.Daneel/Views/Suggestions/SuggestionsView.cs
```diff
@@ -28,7 +28,7 @@ public sealed class SuggestionsView : MvxView
         private const float distanceBetweenSuggestions = 12;
 
         private readonly UILabel titleLabel = new UILabel();
-        private readonly List<SuggestionView> suggestionViews
+        public readonly List<SuggestionView> suggestionViews
```
simonrozsival: Let's not make this public. Please add a method which takes a `CGPoint` and returns the `SuggestionView` instead.

simonrozsival:
I suggest creating an issue which will reference this PR and #3253 and then close these PRs. I don't think we will be able to finish these without more discussions and designs and I definitely don't think it will be soon.

willsb:
I've added more info to #1724 and am closing this guy for the time being, since it's not going anywhere before a discussion

PR #3339, opened 2018-10-08
Add Stop TE Tracking Event

### LINKED ISSUE #3338: Track Stop TE Event in Main App
The app is currently not tracking the Stop TE Event. Let's add in a general one that could:
- Note where the stop trigger comes from (stop button, siri ...)

khoi:
## What's this?
Add a tracking event when a TE is stopped which is currently missing in the apps.

Note: Siri tracking will be done in another PR, this is for the main apps only.

### Relationships
Closes #3338

## Why do we want this?
To understand the users better.

## How is it done?
~A new `StopTimeEntryEvent` class is added.~
`TimeEntryStopOrigin` is added 

### Why not another way?
~The event is very similar to `StartTimeEntryEvent` and we could also share that event and differentiate between 2 using a flag. But I think having separated event is easier to maintain.~

Updated: only `TimeEntryStopOrigin` is used now. Having `StartTimeEntryEvent` seems like over tracking.
### Side effects
Not that I know of.

## Review hints

- There are 3 methods of stopping TEs
1. Manually
1. Deeplink
1. Siri

Let me know I'm missing something.

## :squid: Permissions
🐙


Toggl.Foundation/Analytics/StopTimeEntryEvent.cs
```diff
@@ -0,0 +1,67 @@
...
+             bool isBillable,
+             bool isRunning)
+         {
+             Origin = origin;
+             HasEmptyDescription = hasEmptyDescription;
+             HasProject = hasProject;
+             HasTask = hasTask;
+             NumberOfTags = numberOfTags;
+             IsBillable = isBillable;
+             IsRunning = isRunning;
```
simonrozsival: Why do you include so many properties? I don't think we need those and it will be only a lot of noise. Also the `IsRunning` will always be `true`. I think the only thing we really need is the origin. And if I am right, then I don't think we need a new class for the event.

PR #3345, opened 2018-10-08
Nicer tokens

simonrozsival:
## What's this?
This PR tries to make tokes we use in the creation view and in the edit view nicer. Especially the ones we use for tags (now they have no icon and they have clear background).

### Relationships

There is no issue with the design. This is my own initiative (I was stuck at an airport for too long).

## Why do we want this?
I think this makes it a bit nicer than right now.

## How is it done?
- tag tokens now have an 🏷 icon
- project tokens now have an icon (📁) instead of a dot
- both tokens now have border

### Side effects
No side effects.

## :squid: Permissions
We should not merge this without @amulware's approval.


simonrozsival:
I will close this for now and create a follow-up issue.

Issue #2914, opened 2018-08-10
Optimizations in calendar collection view layout


juanlaube:
Two optimizations that we can do in the collection view layout:

- Cache the layout attributes in a dictionary by indexPath: this avoids recalculating the attributes for each element.
- Visible supplementary views: include only the visible supplementary views in the layout attributes, right now the current time view is always included.
- Invalidate the layout only when needed: I think we the collectionview's width changes is the only case we need.

willsb:
The collection view is stable enough as is. We can work on this issue some other time, after the release

heytherewill:
Should this be included in the new calendar project as a stretch goal?

ping @juanlaube @sofokli1

juanlaube:
I think it would be nice, it's not a UI/UX improvement though.

PR #3351, opened 2018-10-08
Fix crash caused by invalidated layout

juanlaube:
## What's this?
A fix for the most common crash related to the calendar view. I couldn't reproduce the crash, however it was caused by returning a new layout attributes object for the current time before invalidating the current one.

### Relationships

Ref #2914 for more optimizations
Closes nothing.

## How is it done?
By caching the layout attributes for the current time supplementary view.

### Why not another way?
This is actually the correct way to do this and we should adopt this caching strategy for all kinds of views in the collection view, mostly for performance reasons.

### Side effects
None that I can think of.

## Review hints
Test, make sure the current time is updated every minute and scroll trying to break the layout.

## :squid: Permissions
Only me.


willsb:
Can you create a list all possible caching optimizations in #2914 and then add a reference to that issue on this PRs description? I think they are related.

juanlaube:
Done!

PR #3355, opened 2018-10-09
Retry failing API requests in integration tests for 429 and 502 responses

simonrozsival:
## What's this?
This PR will make our integration tests much more reliable. We sometimes get a random `Too Many Requests` or `Bad Gateway` errors from the staging server and our integration tests fail for this reason. This PR will make change the behavior of the API client we use in integration tests so that it repeats unnecessarily failing requests.

### Relationships
I don't think there is an issue for this.

## Why do we want this?
We want our integration tests to be reliable and to fail only when the API changes.

## How is it done?
A wrapper around the `ITogglClient` is injected into `ITogglApi` which analyzes the results of the HTTP calls and re-sends the HTTP request several times if necessary.

### Why not another way?
This is easier than hacking Xunit to repeat certain failing tests...

### Side effects
No side effects.

## Review hints
Run integration tests locally several times and also check this build in Bitrise (feel free to create your own integration test builds in Bitrise, if you have access): https://app.bitrise.io/build/1afa75ce8cbaaf8d

## :squid: Permissions
Let me 🦑 this one


Toggl.Ultrawave.Tests.Integration/Toggl.Ultrawave.Tests.Integration.xunit.runner.json
```diff
@@ -1,3 +0,0 @@
-﻿{
```
juanlaube: Is this really needed? 🤔

Toggl.Ultrawave.Tests.Integration/Properties/AssemblyInfo.cs
```diff
@@ -7,3 +8,4 @@
 [assembly: AssemblyCopyright("Copyright ©  2017")]
 [assembly: AssemblyTrademark("")]
 [assembly: AssemblyCulture("")]
+[assembly: CollectionBehavior(CollectionBehavior.CollectionPerAssembly, MaxParallelThreads = 2)]
```
khoi: more threads pls :))

Toggl.Ultrawave.Tests.Integration/Helper/RetryingApiClient.cs
```diff
@@ -0,0 +1,80 @@
...
+        }
+
+        public void Dispose()
+        {
+            internalClient.Dispose();
+        }
+
+        public async Task<IResponse> Send(IRequest request)
+        {
+            var numberOfRetries = 0;
```
khoi: we could simplify the implementation of this class by using `ConditionalRetryWithBackoffStrategy` that @daividssilverio wrote.

simonrozsival: So I looked into it and it won't work - at least it would not simplify the implementation:
- `ConditionalRetryWithBackoffStrategy` is an extension method for observables, here we work just with a `Task` - we would have to convert the task to an observable
- it retries only when an exception is thrown - we would have to throw exceptions in case the http status code is 429 or 502
- all these changes would result in even more code and much more cryptic - I believe the `do .. while` is better here - it is obvious what is going on just by looking at this class

Toggl.Ultrawave.Tests.Integration/Toggl.Ultrawave.Tests.Integration.xunit.runner.json
simonrozsival: For some reason this config file was being used in Bitrise, but not locally. The attribute works both in Bitrise and locally.

PR #3360, opened 2018-10-09
Fix relative luminance in calendar items

juanlaube:
## What's this?
Fixes the relative luminance math to select the text color in the calendar items

### Relationships

Closes nothing, but will make @simonrozsival happy 😄 

## Why do we want this?
Because users deserve to read the descriptions properly.

## How is it done?
By adjusting the math, based on [this](https://www.w3.org/WAI/GL/wiki/Relative_luminance)

### Side effects
You can read the descriptions now 😄 

## Review hints
Try it, use different project colors, see the attached screenshot as an example.

## :squid: Permissions
Only me.

simulator screen shot - iphone xs max - 2018-10-09 at 11 55 08


Toggl.Foundation.MvvmCross/Extensions/ColorExtensions.cs
```diff
@@ -64,10 +65,27 @@ public static float GetHue(this MvxColor self)
...
+            var r = rsrgb <= 0.03928 ? rsrgb * lowGammaCoeficient : adjustGamma(rsrgb);
+            var g = gsrgb <= 0.03928 ? gsrgb * lowGammaCoeficient : adjustGamma(gsrgb);
+            var b = bsrgb <= 0.03928 ? bsrgb * lowGammaCoeficient : adjustGamma(bsrgb);
+
+            var luma = r * 0.2126 + g * 0.7152 + b * 0.0722;
 
             return luma < 0.5 ? MvxColors.White : MvxColors.Black;
         }
+
+        private static double adjustGamma(double channel)
```
willsb: Move this to within the ForeGroundColor method, since it's very scope specific

PR #3364, opened 2018-10-09
Fix Siri UI Extension bundle id

Odrakir:
## What's this?
The old name was a little funky, I just changed it to be consistent with the rest,.

## Why do we want this?
We were having problems debugging on device. I created the missing prov. profile and then realized this bundle id was weird.

## Review hints
Download the new provisioning profile first.

## :squid: Permissions
Please do.


arturs105:
I'm still not able to run on a device, but since this PR already has 2 approvals, it can be merged without one from me, probably.

dshr:
@arturs105 this one is supposed to fix your problem too though. 

In my case my device wasn't registered on the dev portal, is your phone there?

PR #3343, opened 2018-10-08
Giskard 1.5.2 - Sodium Hotfix 2

daividssilverio:
## Changelog
```
- Added UNDO feature after deleting time entries by swiping
- Added notifications for idle and running time entry (enable in settings)
- Added in-app feedback form (let us know what you think!)
- Added workspace selection to Reports screen
- Added warning when user does not belong to any workspace

- Improved syncing process
- Improved application behaviour on slower networks
- Improved performance and stability
- Improved onboarding hints
- Improved the way unsyncable time entries are visualized

- Fixed application freezes when interacting with persistent notifications (running/idle)
- Prevented auto log in with previous user on reinstall
- Fixed notifications so that they no longer emit sounds
- Various other bug fixes

Hotfixes:
- Fix lollipop devices startup crash
- Fix time entries sections headers
- Fixed bug that crashed the application in main view
```

## History of changes:
Sodium - Sodium Hotfix #2
https://github.com/toggl/mobileapp/compare/sodium...sodium-hotfix-2

1.5.1 - Sodium Hotfix #2
https://github.com/toggl/mobileapp/compare/giskard-1.5.1...sodium-hotfix-2

## :squid: Permissions
@kornelijepetak or @willsb

PR #3365, opened 2018-10-09
Sodium: better rtl support

daividssilverio:
## What's this?
Cherry pick of #3285 to include slightly better support for rtl languages on the main log.

### Relationships
It's basically #3285
To be included in #3343 

Closes Nothing

## Why do we want this?
So users that use rtl languages have a better experience.

## How is it done?
Please check #3285

### Why not another way?
Actually solving the rtl will take more effort, this helps a little.
The other way would be actually going through all views, testing positioning with help from someone who actually uses rtl languages.

### Side effects
Devices with RTL languages display the description ltr on the main log.

## Review hints
Consider that the same changes have already been merged into develop.
When testing, just use Hebrew or Arabic keyboards to check how it works. Changing the system language to RTL won't go as you expect, this changes won't fix that.

## :squid: Permissions
🦑 whenever


willsb:
It makes no sense to have this in the hotfix, I'm afraid, so the changes will be instead included in 1.6

PR #3369, opened 2018-10-10
📐 Fix undo button size

### LINKED ISSUE #2998: Undo button hitbox too small
♻️hitbox button for the undo button is too small

## Context
Build/Branch: `zoranel`

## Expected behavior
Tapping "around" the Undo label should also trigger the button

## To reproduce
- Delete a TE
- Tap the very end of the snack bar, at the right of the Undo button
- Nothing happens

sofokli1:
## What's this?
Makes the undo button bigger thus easier to touch.

### Relationships
Closes #2998 

## Why do we want this?
The undo button is hard to touch thus hard to undo. We want to make this easier.

## How is it done?
Small changes to the constraints and text alignment of the button.

### Why not another way?
This is the simplest that i could think of.

### Side effects
None that I know of

## Review hints
After taking a look at the code and testing this functionally you can use `button.BackgroundColor = UIColor.Red;` to look at the button size in the UI.

## :squid: Permissions
:squid:


Toggl.Daneel/Views/SnackBar.cs
```diff
@@ -69,9 +69,11 @@ public void AddButton(String title, Action onTap)
             {
                 onTap();
             };
+            button.BackgroundColor = UIColor.Red;
```
khoi: 🔴

sofokli1: i thought i removed that line 🤦‍♂️

sofokli1: didn't save in vs before committing 🤦‍♂️

PR #3383, opened 2018-10-10
Make reports loading nicer

simonrozsival:
## What's this?
There was an annoying glitch in the reports loading state. There were two overlaying cards with a shadow displayed during loading of reports - the one with the loading state then fades out. The shadow looks weird. I don't like it. I removed this glitch it with this PR.

### Relationships

There is no issue in Github.

## Why do we want this?
It looks better now.

## How is it done?
Simplifying the view by removing some unnecessary stuff.

### Why not another way?
This is the way.

### Side effects
No side effects.

## Review hints
Compare with `develop` - two simulators side by side. The difference should be quite apparent.

## :squid: Permissions
🦑


dshr:
There's some loading strangeness happening with the labels right now. Maybe we should bind their visibility too, or at least set them to "" while loading?

loading-strangeness

simonrozsival:
I know what is wrong, thanks for pointing that out, @dshr!

simonrozsival:
@dshr Please check this now.

PR #3385, opened 2018-10-10
📐 Fix user calendar selection height in SE devices

### LINKED ISSUE #3336: Select calendar screen doesn't look good in small screens
## ⛑️ User Support
Reported by Señor Fak, he has a lot of calendars and an iPhone SE.

## About the user

App version: Daneel 1.12
Phone OS: iOS
Phone model: iPhone SE

## To Reproduce
See the attached screenshot.

img_0502

sofokli1:
## What's this?
Fixes the UI glitch of the user calendar selection when they have many calendars on SE devices

### Relationships
Closes #3336

## Why do we want this?
It looks bad on SE devices

## How is it done?
There was a constant value for the max height which i changed to a different (smaller value) for SE devices.

### Why not another way?
A better idea would be to have it relative to the height of the container instead of the static maxValue but that requires design approval before going that way. Also it makes sense to put all this static values related to the UI in a project and fix them all once and for all with dynamic/relative approach instead of the hardcoded values.

### Side effects
None

## Review hints
Just create a 4-5 calendars in the simulator and go to the selection page to see how it looks.

## :squid: Permissions
🦑


Toggl.Daneel/ViewControllers/Calendar/SelectUserCalendarsViewController.cs
```diff
@@ -15,7 +16,7 @@ public sealed partial class SelectUserCalendarsViewController
     {
         private const int heightAboveTableView = 98;
         private const int heightBelowTableView = 80;
-        private const int maxHeight = 627;
+        private int maxHeight = UIScreen.MainScreen.Bounds.Width > 320 ? 627 : 528;
```
arturs105: You could make this `readonly`.

sofokli1: done

khoi:
immutable FTW, 🚀

PR #3387, opened 2018-10-10
Add Diagnostics infrastructure

### LINKED ISSUE #3386: Add the tracking infrastructure
We'll need:

- `MeasuredOperation`: An enum that lists all operations that we are tracking
- `IStopwatch`: A simple class with a `Start` and a `Stop` methods and a reference to its underlying `Operation` 
- `IStopwatchFactory`: Interface that generates the native stopwatches that will rely on Firebase to measure the actual performance of the operations.

willsb:
## What's this?
This adds the infra we need for measuring performance using Firebase!

### Relationships
Closes #3386

## Why do we want this?
We need concrete user data before making decisions on what to improve performance-wise

## How is it done?
We are wrapping the firebase services so they are usable from Foundation. Also, it allows us to extend the tracking mechanisms with compositions in case we wanna track locally or to add behaviour (like I am doing right now by making every `IStopwatch` usable only once and by making outputting the results to the console if requested).

### Why not another way?
This is simple, easy and consistent

### Side effects
[Xamarin has a bug](https://github.com/xamarin/GooglePlayServicesComponents/issues/151) that's preventing us from tracing things on Droid. iOS is working fine tho. 

## Review hints
This is really basic, so only judge the architectural aspects of this (and disregard the droid problem for the time being).

## :squid: Permissions
Anyone


Toggl.Foundation/Diagnostics/BaseStopwatchFactory.cs
```diff
@@ -0,0 +1,80 @@
...
+
+                didStart = true;
+
+                InnerStopwatch.Start();
+            }
+
+            public virtual void Stop()
+            {
+                if (!didStart)
+                    throw new InvalidOperationException("You need to start a stopwatche before stopping it");
```
daividssilverio: stopwatche -> stopwatch

Toggl.Foundation.Tests/Diagnostics/BaseStopwatchFactoryTests.cs
```diff
@@ -0,0 +1,78 @@
...
+    public sealed class BaseStopwatchFactoryTests
+    {
+        public sealed class TheCreateMethod
+        {
+            private readonly BaseStopwatchFactory stopwatchFactory = new TestStopwatchFactory();
+
+            [Theory, LogIfTooSlow]
+            [InlineData(true)]
+            [InlineData(false)]
+            public void ReturnsStopwatchesThatCanOnlyBeStartedOnce(bool outputToConsole)
```
daividssilverio: This booleans are here just so that you can test `SingleUseStopwatch` and `ConsoleStopwatch` at the same time?

willsb: Yes, stressing all paths is crucial

PR #3388, opened 2018-10-11
🛣 Add default suggestions for Siri shortcuts

### LINKED ISSUE #3380: Siri: Add suggested shortcuts
We can add some shortcuts that don't required to be executed at least once to show in Siri's preferences (ref [doc](https://developer.apple.com/documentation/sirikit/shortcut_management/suggesting_shortcuts_to_users))

We should that with: Stop timer, start empty timer in default workspace and show different reports

khoi:
## What's this?
Introduced some default siri shortcuts action that are available when user's logged in.

### Relationships

Closes #3380

## Why do we want this?
Currently, our app requires actions (start, stop TE, show report) to be executed at least once before they're available as Siri Shortcuts.
This will add some of the actions as defaults and always there when users navigate to Settings -> Siri

## How is it done?
We set the default suggestions in MainViewModel and clear it when users log out.

### Side effects
Not that I know of

## Review hints
- Completely remove the app to get rid of all pre-donated intents.
- Install and sign in
- Go to Settings -> Siri and make sure 3 default intents are there:
1. Start empty TE
1. Stop current TE
1. Show Report

## :squid: Permissions
Me or @Odrakir


Toggl.Foundation.Tests/MvvmCross/ViewModels/MainViewModelTests.cs
```diff
@@ -1094,6 +1095,16 @@ public async void DoesNotShowTheRatingViewIfAfter24HourSnooze()
                     );
                 }
             }
+
+            public sealed class SiriShortcuts : MainViewModelTest
```
simonrozsival: No Siri in shared code! It is called Intent donation service! :)

PR #3390, opened 2018-10-11
Remove unnecessary http request

simonrozsival:
## What's this?
Avoid unnecessary HTTP requests - make tests more stable/reliable and also a bit faster.

### Relationships
Closes no issue

## Why do we want this?
My hypothesis is that sometimes it takes the server a bit longer time to process a new user (db replication delay maybe? I don't understand this enough to guess what is wrong with the sever :D) and querying the `/me` endpoint a few milliseconds after the user account was created sometimes leads to a 403 - Forbidden response from the server.

Also we use this in almost every of our unit tests (and we have > 300 of them) so this makes the integration tests run significantly faster (on my machine it was about 2min speedup).

## How is it done?
I just use the user data that is returned from the signup endpoint, we don't need to query the server again to get the same data.

### Why not another way?
This PR changes only a few lines of code, I didn't want to add any fancy extension methods for integration test boilerplate...

### Side effects
No side effects.

## Review hints
Run integration tests locally a few times + check this Bitrise build: https://app.bitrise.io/build/05ded5c7252fb653

## :squid: Permissions
🦑


simonrozsival:
It doesn't seem to be faster in Bitrise :( But still, let's not make unnecessary requests :)

Issue #2902, opened 2018-08-09
Consider ghost-ness in edit view


amulware:
The edit view will still open and show all the details, however:

I think we should not allow users to edit tags in this state (since the popup would show none, see #2900). They could be able to edit everything else though.

Alternatively:

We could allow the user to ONLY:
- edit project
- delete

Maybe @pe0ny wants to comment.

arturs105:
We should definitely indicate which properties can be edited in this state, but i think it makes more sense to do it in #2904.

PR #3396, opened 2018-10-11
Consider ghost-ness in edit view

arturs105:
## What's this?
This makes most of the properties of a ghost entry non-editable. The user can still edit the project and delete the entry.

### Relationships

Ref #2902 
The issue is not closed, because this needs to be added to Android as well.

## How is it done?
By using the old (mvx) bindings.

### Why not another way?
The edit view has not yet been rxfactored, so no reactive bindings.

## :squid: Permissions
Whenever.


Toggl.Daneel/ViewControllers/EditTimeEntryViewController.cs
```diff
@@ -236,6 +236,23 @@ public override void ViewDidLoad()
                       .To(vm => vm.IsBillableAvailable)
                       .WithConversion(visibilityConverter);
 
+            //Regarding ghost entries
+            void createUserInteractionBindingForGhostedEntries(UIView view)
```
willsb: These are hoisted and should be declared at the end of the file

arturs105: That makes more sense.

Toggl.Foundation.Tests/MvvmCross/ViewModels/EditTimeEntryViewModelTests.cs
```diff
@@ -37,19 +37,28 @@ public abstract class EditTimeEntryViewModelTest : BaseViewModelTests<EditTimeEn
...
-                    .SetWorkspaceId(11)
-                    .SetUserId(12)
-                    .SetProjectId(13)
-                    .SetTaskId(14);
+                var workspaceId = 11;
+                var userId = 12;
+                var projectId = 13;
+                var taskId = 14;
+
+                TheTimeEntry = Substitute.For<IThreadSafeTimeEntry>();
```
simonrozsival: Let's use `MockTimeEntry` - too many `Returns(...)` here, this will be a pain if we have to change this any time in the future. We should use the `Mock*` classes more often.

PR #3402, opened 2018-10-11
🚧 Make tag creation behave more like web

### LINKED ISSUE #3323: Can create the same tag multiple times when starting a new time entry
## 🐛 Bug Report
Duplicate tags can be created
## Context
Creating a new entry on a 4G connection

Build/Branch:1.12
Phone OS:12
Phone model:iPhone SE

## Expected behavior
Only a single tag with that name should be created and assigned to that time entry. Currently multiple are showing with identical names.
## To reproduce
Start a new time entry
Press the tag button or press # 
Enter a name for the tag
Press "Create tag XXXXX"
Press "Create tag XXXXX" again
This causes the entry to show "Oops Invalid tag_ids" on the entry
## Suggestions

Only press "Create tag XXXXX" once

sofokli1:
## What's this?
User could create multiple tag for the same text thus ending in a bunch of tags that are unsync-able and visible in the UI as well.

### Relationships
Closes #3323

## Why do we want this?
We want to improve how tag creation handle edge cases like creating a tag with same text but different case.

## How is it done?
- Better check in the logic that decides when to show the `+ Create tag blabla` cell.
- Call for a refresh after a tag is created so that the above cell is removed.

### Why not another way?
This seems like the right way to me but am open to ideas

### Side effects
Only good stuff

## Review hints
The issue describes how to reproduce

## :squid: Permissions
:squid:


Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
```diff
@@ -89,7 +89,10 @@ public bool SuggestCreation
                            && CurrentQuery.LengthInBytes() <= MaxProjectNameLengthInBytes;
 
                 if (IsSuggestingTags)
-                    return Suggestions.None(c => c.Any(s => s is TagSuggestion tS && tS.Name == CurrentQuery))
+                    return Suggestions.None(c => c.Any(s => 
+                               s is TagSuggestion tS 
+                               && tS.Name.ContainsIgnoringCase(CurrentQuery)
```
simonrozsival: here use `tS.Name.Equals(CurrentQuery, StringComparison.CurrentCultureIgnoreCase)` instead

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
```diff
@@ -92,8 +92,10 @@ public bool SuggestCreation
                            && shouldSuggestProjectCreation;
 
                 if (IsSuggestingTags)
-                    return Suggestions.None(c => c.Any(s => s is TagSuggestion tS && tS.Name == CurrentQuery))
-                           && CurrentQuery.LengthInBytes() <= MaxTagNameLengthInBytes;
+                    return Suggestions.None(c => c.Any(s =>
```
kornelijepetak: With such a complex return value, please wrap the if block in curly braces.

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
```diff
@@ -92,8 +92,10 @@ public bool SuggestCreation
                            && shouldSuggestProjectCreation;
 
                 if (IsSuggestingTags)
-                    return Suggestions.None(c => c.Any(s => s is TagSuggestion tS && tS.Name == CurrentQuery))
-                           && CurrentQuery.LengthInBytes() <= MaxTagNameLengthInBytes;
+                    return Suggestions.None(c => c.Any(s => 
+                               s is TagSuggestion tS
```
kornelijepetak: Please name all the lambda parameters. `c`, `s` and `tS` should be named properly.

simonrozsival: I think that should be done in a separate PR 🤔 This just changes the logic, renaming parameters would be very helpful but I would prefer to keep this PR as simple as possible.

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
simonrozsival: I disagree. Why using a block? It is just a single (nested) expression.

PR #3424, opened 2018-10-12
🌠 Add splash image

### LINKED ISSUE #3419: Add the toggl logo to the iOS splash screen
It is completely blank as of today. While this won't improve sluggishness, it will improve perceived sluggishness which is a nice bonus for something so simple

khoi:
## What's this?
This add Toggl logo as the launch image.

### Relationships
Closes #3419

## Why do we want this?
so users can look at our fabulous logo instead of the blank ◻️ screen.

## How is it done?
- Old `togglLogo` and `togglLogoBig` assets have been deleted.
- Add new vector `togglLogo` asset

### Why not another way?
This is the 🛣.

### Side effects
- Since the asset is now vectorized, instinct size is no longer available, when creating new UIImageView programmatically, we need to ensure the width and height are set.

## Review hints
- Check launch screen
- Check the logo in the navigation title for all tabs.
- Check the logo in signup screen
- Check the logo in sign in screen

## :squid: Permissions
🦑


Toggl.Daneel/Extensions/ViewExtensions.cs
```diff
@@ -33,5 +33,15 @@ public static void ConstrainToViewSides(this UIView self, UIView view)
             self.LeadingAnchor.ConstraintEqualTo(view.LeadingAnchor).Active = true;
             self.TrailingAnchor.ConstraintEqualTo(view.TrailingAnchor).Active = true;
         }
+
+        public static UIView NavigationLogoView()
```
simonrozsival: This is not an extension method, it does not belong into `ViewExtensions`. Please move it to a different helper class.

simonrozsival:
For some reason the splash screen is white for me now (yes, I am on the correct branch). Am I holding it wrong?

Also please remove all the changes which are now unrelated to the issue (we need the PDF just for the splash screen logo, not the logo with the text).

khoi:
closing this, this will be done in a more trivial PR

Issue #3434, opened 2018-10-12
Show multiple error messages when editing time entries


juanlaube:
Right now, only one error message (that comes from backend) is shown in the edit time entry view. We're introducing a new message when editing a time entry marked as ghost. This issue is to decide if we need a UI change to show multiple error messages, or we will show the most relevant only.

simonrozsival:
I think the message is OK and it would be confusing if we added a second error message. Let's leave it as it is and change only based on user's feedback.

Issue #2904, opened 2018-08-09
Visually distinguish ghost time entries in edit view


amulware:
Very related to #2902, but this one is more about showing a message like "You no longer have access to the workspace of this time entry. You can change it to a different workspace, or delete it. It will also sync correctly if you regain access to the workspace."

Maybe shorter, but that's sort of the information that might be important to give to the user.

I was first thinking we could use the existing sync error property, but that would mean actually touching all the ghost TEs in the database. That seems silly if we can just check the ghost state in the VM and set the error text for the UI based on it (and ignore any actual sync error text).

Thoughts welcome! Ping @pe0ny

simonrozsival:
I think that for now we should add a warning message which would say something like:
```
This time entry belongs to a workspace that you don't have access to anymore and it can't by synced. Either delete it or move it to a different workspace by changing the project.
```
It would be good if we could come up with a shorter copy, because this will be too long on the mobile screen of SE or even 7/8, not to mention Android phones with possibly even smaller screens.

arturs105:
I think indicating non-editable properties should be a part of this issue as well. Probably using light grey for the non-editable rows and/or text in the edit time entry view.

simonrozsival:
Yes, let's try light grey for individual properties.

PR #3435, opened 2018-10-12
Show error message when editing ghost time entries

juanlaube:
## What's this?
This shows an error message (using the existing UI for other sync issues) in the EditTimeEntryViewModel when editing a time entry that belongs to an inaccessible workspace.

### Relationships

Ref #2904 
A follow up PR will include disabling the UI elements that cannot be edited.

## How is it done?
By setting the `SyncErrorMessage` property in the view model when the time entry is marked as ghost.

### Why not another way?
This is a simple solution.

### Side effects
RIght now, if there's another sync issue and the time entry is also marked as ghost, only the ghost message will be shown. #3434 is a follow up issue to decide if we show only the most relevant error or a list of errors.

## :squid: Permissions
Anyone, whenever.


Toggl.Foundation.MvvmCross/ViewModels/EditTimeEntryViewModel.cs
```diff
@@ -603,5 +602,18 @@ private void OnProjectChanged()
         {
             hasProjectSubject.OnNext(!string.IsNullOrWhiteSpace(Project));
         }
+
+        private void setErrorMessage(IThreadSafeTimeEntry timeEntry)
```
simonrozsival: Use ternary operator and this method can be only 4 lines instead of 9! (Sorry for nitpicking... I just really don't like `if .. return`s.)

simonrozsival:
(Otherwise this PR is very nice and straightforward. I like that you decided to deal with the issue in 2 PRs instead of one!)

Toggl.Foundation.MvvmCross/ViewModels/EditTimeEntryViewModel.cs
```diff
@@ -606,5 +605,11 @@ private void OnProjectChanged()
...
+        {
+            SyncErrorMessage = timeEntry.IsGhost ? Resources.GhostTimeEntryErrorMessage : timeEntry.LastSyncErrorMessage;
+            SyncErrorMessageVisible = timeEntry.IsGhost || !string.IsNullOrEmpty(SyncErrorMessage);
```
simonrozsival: The `timeEntry.IsGhost` isn't necessary, but I actually like it - it makes it far easier to read.

PR #3448, opened 2018-10-15
🚧 Keep only negative id ghost tes

### LINKED ISSUE #3431: Hide ghost time entries from the main log
The user should not be able to see time entries he/she doesn't have access to and can't edit in the main view.

For time entries which are in sync with the server this should happen according to the previous specification anyway: we should remove them from the device anyway.

For time entries which were pulled from the server or which were created in the app, but have been pushed to the server before (and have a positive ID now), there is nothing the user can do with them - we can't update the data on the server (e.g., change the project or delete the entity entirely), because it belongs to the workspace and we don't have access to the workspace.

A solution to this cased would be to:
a) for editing - changing the project to some project in a different, accessible, workspace - create a copy of the TE in the other workspace, and delete the original TE from the device
b) for deleting - just delete the TE from the device

The problem with this solution would be that when the user would get access to the workspace back the user will pull this TE from the server again and will have a "duplicite" TE - in two different workspaces, or he will now have a TE which he had already deleted.

I think the best thing in this case is to hide the TE (despite them being unsynced or unsyncable) and keep it locally so it could be correctly restored when user gets access to the workspace back.

For time entries which exist only on user's device (they haven't been synced yet and have a negative ID), the user should be able to delete them (there is no counterpart on the server) from the device or change them to a different workspace and then work with them as with regular time entries.

What do you think about this proposal, @juanlaube and @sofokli1? Does it make sense?

sofokli1:
## What's this?
- Renames an interactor to better show what it does ( i am open to suggestions for the naming)
- Adds logic to include all TEs that are ghost but have a negative ID
- Adds a test to check the above logic

### Relationships
Closes #3431

## Why do we want this?
We want to not show ghosted data (data the user has no access to anymore) but we want to show data that was created on device and never synced.

## How is it done?
Added an extra check in the interactor that is used for showing the TEs in the main log

### Why not another way?
Since we are moving toward a `business logic should be in an interactor` approach it made sense the this way.

### Side effects
None that are unexpected

## Review hints
Check the code and the test. The tests should be ok for the functional side of things.

## :squid: Permissions
:squid:


Toggl.Foundation/Interactors/IInteractorFactory.cs
```diff
@@ -28,7 +28,7 @@ public interface IInteractorFactory
 
         IInteractor<IObservable<Unit>> DeleteTimeEntry(long id);
 
-        IInteractor<IObservable<IEnumerable<IThreadSafeTimeEntry>>> GetAllNonDeletedTimeEntries();
+        IInteractor<IObservable<IEnumerable<IThreadSafeTimeEntry>>> GetAllTimeEntriesNonDeletedAndNonAccessibleThatAreNotOwnedByWorkspace();
```
simonrozsival: 🤔 this name is very descriptive, but it is also very long. I think we could use the name to express the purpose of the data that it returns, not the actual predicate it represents: I would suggest `GetAllTimeEntriesVisibleToTheUser`

Toggl.Foundation/Interactors/TimeEntry/GetAllNonDeletedAndNonAccessibleThatAreNotOwnedByWorkspaceInteractor.cs
```diff
@@ -7,18 +7,18 @@
...
         {
             Ensure.Argument.IsNotNull(dataSource, nameof(dataSource));
             
             this.dataSource = dataSource;
         }
-        
+
         public IObservable<IEnumerable<IThreadSafeTimeEntry>> Execute()
-            => dataSource.GetAll(te => !te.IsDeleted);
+            => dataSource.GetAll(te => !te.IsDeleted && (te.IsGhost ? te.Id < 0 : true), true);
```
simonrozsival: The condition can be simplified: `te.IsGhost == false || te.Id < 0`.

Also please let's name the `true` literal - it is not obvious what it means without the knowledge of the data source methods.

simonrozsival:
```suggestion
            => dataSource.GetAll(te => !te.IsDeleted && !te.IsGhost || te.Id < 0), includeGhosts: true);
```

sofokli1: `!te.IsDeleted` will not be checked in the case of `te.Id < 0`
does that make sense?

PR #3400, opened 2018-10-11
App Diagnostics

willsb:
## What's this?
This branch will collect all the performance measurement PRs before we merge them, all at once, into develop

## Why do we want this?
This is the first phase of the performance project. We are going to first measure places with hiccups in order to get actual user data in Firebase.

## How is it done?
Check #3386 

## References

Check #3347 and #3348 for insights on what needs to/can be improved

## Review hints
Once this is ready for merging, we need to make sure that:

- [x] The Xamarin bug that caused Firebase to become useless on Droid has been worked around
- [x] Data is actually going to Firebase
- [x] The app works as expected on Debug and Release

## :squid: Permissions
Me

PR #3449, opened 2018-10-15
Sanic tapes: Open start view

### LINKED ISSUE #3422: Measure the time it takes to open the start view
When you click the start button in the main view

daividssilverio:
## What's this?
Part of the sanic project, this PR adds the tracing for the time needed to open the Start TE view.

### Relationships
Part of #3400
Closes #3422

## Why do we want this?
To measure how fast/slow our app is and take actions to improve it.

## How is it done?
~This simply adds a trace between the time the VM is initialized and the time the view appears.~
This adds a trace from the StartTE command to the point the StartTEVM view appears.

### Why not another way?
This way is simple and works enough.
~Other possible way would be track from the moment the user actually clicks the StartTE button from the main screen, start and cache the stopwatch (trace) and stop it on the Start TE VM `ViewAppeared` lifecycle. ~ <- Just did that.

### Side effects
~This way is most likely ignoring the time spent on the navigation itself (that is, the MainVM calling `navigationService.Navigate(...)` and the parsing/logic involved to find the StartTE VM.~
None.

## Review hints
None.

## :squid: Permissions
:sanic: 🦑 please


Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
```diff
@@ -296,6 +302,8 @@ public override void Prepare(StartTimeEntryParameters parameter)
 
         public override async Task Initialize()
         {
+            startTimeEntryStopwatch = stopwatchFactory.Create(MeasuredOperation.OpenStartView, true);
```
simonrozsival: I would suggest naming the `true` literal - I don't know what `stopwatchFactory.Create` does, so without looking at the method, I don't have a clue.

daividssilverio: I forgot the `true`, it was supposed to be used for testing only.

PR #3451, opened 2018-10-15
🚧Detect lose of access to workspaces

### LINKED ISSUE #3443: WS removal: track that user lost access to a workspace
Track an event every time user loses access to some workspace.

sofokli1:
## What's this?
Add a new call to the analytics service to track lose of access to workspaces and uses it in the related sync state

### Relationships
Closes #3443

## Why do we want this?
We like to track events to get more insights on how the app is used

## How is it done?
New analytic event that tracks lose of WS. Called in the related sync state.

### Why not another way?
This seemed like the right place

### Side effects
None

## Review hints
Be aware the sync state where this was added is not in use at the moment

## :squid: Permissions
:squid:


Toggl.Foundation/Sync/States/Pull/DetectLosingAccessToWorkspacesState.cs
```diff
@@ -50,5 +55,15 @@ private IObservable<Unit> markAsGhosts(IList<IThreadSafeWorkspace> workspacesToM
 
         private IObservable<IThreadSafeWorkspace> markAsGhost(IThreadSafeWorkspace workspaceToMark)
             => dataSource.Update(workspaceToMark.AsGhost());
+
+        private IObservable<IList<IThreadSafeWorkspace>> trackLoseOfWorkspaceAccessIfNeeded(IList<IThreadSafeWorkspace> workspacesNotFetched)
```
khoi: I'd change this method to
```
        private void trackLoseOfWorkspaceAccessIfNeeded(IList<IThreadSafeWorkspace> workspacesNotFetched)
        {
            if (workspacesNotFetched.Count > 0)
            {
                analyticsService.LostWorkspaceAccess.Track();
            }
        }
```

and do

```
        private IObservable<IList<IThreadSafeWorkspace>> workspacesWhichWereNotFetched(List<IWorkspace> fetchedWorkspaces)
            => allStoredWorkspaces()
                .Where(stored => fetchedWorkspaces.None(fetched => fetched.Id == stored.Id))
                .ToList()
                .Do(trackLoseOfWorkspaceAccessIfNeeded);
```

for the reason that `.Do` operator clearly state that it's doing a side effect

PR #3455, opened 2018-10-16
📊 Track Siri usage

### LINKED ISSUE #3354: Add tracking to Siri Extension
We need to track the usage of Siri Shortcuts, probably for now it'll be enough to track every execution of an intent and the responses (StopTimerSiriIntent, StopTimerSiriIntentResponse(Success/Error)...)

There are two possible approaches for this:
1) Store the events in the shared storage as they happen and send them when the app opens.
2) Send them directly from the extension:
    a - By including a light version of the analytics service in the extension
    b - By moving some the anlytics service code to the `ExtensionKit` project so it can be shared between the extension and Daneel.

I think the best approach is 2b, but I'm open to discuss it.

khoi:
## What's this?
This tracks the usage of the Siri Intent Handlers.

### Relationships
Closes #3354

## Why do we want this?
We want to know how our users are using the Siri Extension.

## How is it done?
Inside the Intent Handlers, since we don't have access to our AnalyticsService, we save the events into the `SharedStorage`, and then later on when the app opens, the main app gets the events and send it out.

### Why not another way?
Tracking in extensions are not officially supported by Apple and the analytics SDKs, so we have to save it and only send it later.

### Side effects
Since the events are not sent right away, the timestamp when the event happens is saved as an event params.

## Review hints
Review everything as a whole since early commits were a mess.  

## :squid: Permissions
🐙


Toggl.Daneel/SiriTrackableEvent.cs
```diff
@@ -0,0 +1,44 @@
+using System.Collections.Generic;
+using Foundation;
+using Toggl.Daneel.ExtensionKit.Analytics;
+using Toggl.Foundation.Analytics;
+
+namespace Toggl.Daneel
+{
+    public class SiriTrackableEvent: ITrackableEvent
```
Odrakir: Can we put this file somewhere else not in the root of the project?

Toggl.Daneel/Startup/AppDelegate.cs
```diff
@@ -8,6 +8,8 @@
 using MvvmCross.Navigation;
 using MvvmCross.Platforms.Ios.Core;
 using MvvmCross.Plugin.Color.Platforms.Ios;
+using Toggl.Daneel.ExtensionKit;
+using Toggl.Daneel.ExtensionKit.Analytics;
```
Odrakir: 🔥 These two lines

Toggl.Daneel/ViewControllers/MainViewController.cs
```diff
@@ -266,13 +267,25 @@ public override void ViewWillAppear(bool animated)
 #endif
         }
 
+        private void trackSiriEvents()
+        {
+            var events = SharedStorage.instance.PopTrackableEvents();
```
Odrakir: Can we make `PopTrackableEvents` return a `NSArray<SiriTrackingEvent>` or at least cast it to that? That way the loop could be done with a simpler `foreach`.

Toggl.Daneel.ExtensionKit/Analytics/SiriTrackingEvent.cs
```diff
@@ -0,0 +1,70 @@
...
+        }
+
+        #region INSCoding
+        [Export("initWithCoder:")]
+        public SiriTrackingEvent(NSCoder coder)
+        {
+            EventName = (NSString) coder.DecodeObject(eventNameEncodeKey);
+
+            var dict = (NSDictionary) coder.DecodeObject(parametersEncodeKey);
+            Parameters = NSDictionary<NSString, NSString>.FromObjectsAndKeys(dict.Values, dict.Keys);
```
Odrakir: I'm getting an exception here when trying to track a stop event: `System.NullReferenceException: Object reference not set to an instance of an object

Toggl.Daneel.ExtensionKit/Analytics/ITrackingEvent.cs
```diff
@@ -0,0 +1,10 @@
+using System.Collections.Generic;
+
+namespace Toggl.Daneel.ExtensionKit.Analytics
+{
+    public interface ITrackingEvent
+    {
+        string eventName { get; }
```
simonrozsival: `PascalCase` please

Toggl.Daneel.ExtensionKit/Analytics/ITrackingEvent.cs
```diff
@@ -0,0 +1,10 @@
...
+        Dictionary<string, string> parameters { get; }
```
simonrozsival: `PascalCase` please

Toggl.Daneel.ExtensionKit/Analytics/SiriTrackingEvent.cs
```diff
@@ -0,0 +1,70 @@
...
+        public SiriTrackingEvent(NSString eventName, NSDictionary<NSString, NSString> parameters, NSDate eventDate)
+        {
+            EventName = eventName;
+            Parameters = parameters;
+            EventDate = eventDate;
+        }
+
+        public static SiriTrackingEvent StartTimer(ITimeEntry te)
+        {
+
```
simonrozsival: 🔥

Toggl.Daneel.ExtensionKit/Analytics/SiriTrackingEvent.cs
```diff
@@ -0,0 +1,70 @@
...
+            var dict = new Dictionary<string, string>
+            {
+                {"HasEmptyDescription", string.IsNullOrEmpty(te.Description).ToString()},
```
simonrozsival: Consider `["key"] = value` syntax for initializing a dictionary. I think that in this case it would be easier to read.

Toggl.Daneel.ExtensionKit/Analytics/SiriTrackingEvent.cs
```diff
@@ -0,0 +1,70 @@
...
+                {"HasProject", (te.ProjectId != null).ToString()},
+                {"HasTask", (te.TaskId != null).ToString()},
+                {"NumberOfTags", te.TagIds.Count().ToString()},
+                {"IsBillable", te.Billable.ToString()}
+            };
+
+            var nativeDict = NSDictionary<NSString, NSString>
+                .FromObjectsAndKeys(dict.Values.ToArray(), dict.Keys.ToArray());
+
+            return new SiriTrackingEvent(new NSString("SiriStartTimer"), nativeDict, NSDate.Now);
```
simonrozsival: Make a string constant please.

Toggl.Daneel.ExtensionKit/Analytics/SiriTrackingEvent.cs
```diff
@@ -0,0 +1,70 @@
...
+        }
+
+        public static SiriTrackingEvent StopTimer()
+        {
+            return new SiriTrackingEvent(new NSString("SiriStopTimer"), null, NSDate.Now);
```
simonrozsival: Make a string constant please.

Toggl.Daneel/ViewControllers/MainViewController.cs
khoi: I couldn't find a trivial way to cast an `NSArray` to `NSArray<T>`. One possible way is to 
`NSArray > T[] > NSArray<T>` but that looks like a big performance hit.

Toggl.Daneel.ExtensionKit/Analytics/SiriTrackingEvent.cs
khoi: Crash fixed

Toggl.Daneel.ExtensionKit/Analytics/SiriTrackingEvent.cs
```diff
@@ -0,0 +1,90 @@
...
+    [Register("SiriTrackingEvent")]
+    public class SiriTrackingEvent: NSObject, INSCoding
+    {
+        private static readonly string eventNameEncodeKey = nameof(eventNameEncodeKey);
+        private static readonly string parametersEncodeKey = nameof(parametersEncodeKey);
+        private static readonly string eventDateEncodeKey = nameof(eventDateEncodeKey);
+
+        private static readonly string SiriStartTimerEventName = "SiriStartTimer";
+        private static readonly string SiriStopTimerEventName = "SiriStopTimer";
+        private static readonly string SiriErrorEventName = "SiriErrorEventName";
```
Odrakir: The name is better as "SiriError" or "SiriIntentError" or whatever.

Toggl.Daneel/ViewControllers/MainViewController.cs
Odrakir: Actually, now that I think about it, in my opinion `NSArray` shouldn't leave `SharedStorage`. We need it there because interop with iOS, but it's not native to C# and shouldn't be used outside of it. I think `NSArray` is `IEnumerable`, so you can probably just use that in the method signature. And then maybe use generic arrays inside `SharedStorage`.

Toggl.Daneel.ExtensionKit/SharedStorage.cs
```diff
@@ -41,6 +42,22 @@ public void SetUserId(double userId)
             userDefaults.Synchronize();
         }
 
+        public void AddSiriTrackingEvent(SiriTrackingEvent e)
+        {
+            var currentEvents = (NSMutableArray)PopTrackableEvents().MutableCopy();
```
Odrakir: No need to pop, just `getTrackableEvents()`.

amulware:
> Since the events are not sent right away, the timestamp when the event happens is saved as an event params.

Is this still the case in the current version? What is the intended use for that data?

Also, semantic question: If I use the shortcut to start 3 time entries and stop 2, will it send 5 events in total?

khoi:
> Is this still the case in the current version? What is the intended use for that data?

This is still valid, that data is used when we want to know the exact time when the events happened. 

>Also, semantic question: If I use the shortcut to start 3 time entries and stop 2, will it send 5 events in total?

Yep

amulware:
Why would we want to know the exact time of the event? We don't use that anywhere else either. We always look at aggregates over weeks/months. I suppose it doesn't really harm us, but there will be A LOT of different values for that parameter, making it definitely useless in the app center/firebase UI, and extracting raw event data and handling it ourselves is something I hope to never have to do. :D

Unless there is a case I'm not seeing, I'd vote for removing that.

Two more points:

Since this is a new event, our regular time entry creation (and stop, once we have that?) event will not count time entries created with Siri, but maybe we want to do that (with a different origin parameter value of course)?

Also, is it intended usage to not add the event to the `IAnalyticsService` interface? (ping @simonrozsival who I think wrote a lot of that).

khoi:
I'll remove the event's time param.

>(and stop, once we have that?)

The tracking for stopping TE's been added in #3339. 

>is it intended usage to not add the event to the IAnalyticsService interface?

Currently, starting and stopping TE using Siri have their own dedicated analytic events, now thinking about it, maybe we can merge it to the existing one with a new origin? 

What do you think @simonrozsival

Odrakir:
@amulware About the timestamp parameter... For every other event you have a trustable timestamp of when that event happened. For these Siri events you don't. We might not use timestamps normally, but in case you ever need to see events in a time frame...

khoi:
I've removed event's date as @amulware requested. Since this is trivial change, I believe the approvals for the PR still valid? @simonrozsival @Odrakir

PR #3457, opened 2018-10-16
Sanic tapes: Start Time Entry suggestions

### LINKED ISSUE #3405: Measure the time it takes to query suggestions in the start time entry
Let's measure:

- The time it for data to move from DB into the VM
- The time it takes for the data to be displayed in for the user

This should be, in principle, doable exclusively in the VM layer

Warning: This is only meant to measure the times. We do not want any optimizations yet.

dshr:
## What's this?
Adding traces to the suggestions load times in the Start Time entry flow.

### Relationships
Part of #3400 
Closes #3405

## Why do we want this?
We gotta know

## How is it done?
1. **The time it for data to move from DB into the VM**: Adding a trace between the DB query and the DB response callback in the VM.
2. **The time it takes for the data to be displayed in for the user:** Not quite sure how to do this without somehow passing a trace between the VM and UI layers.

### Why not another way?
This seems to be the simplest way

### Side effects
None.

## Review hints
None.

## :squid: Permissions
Still in progress, need some clarifications from @willsb.


daividssilverio:
About the android part of this.
The `StartTimeEntryRecyclerView` used to the display the list of suggestions has a `CollectionChange` observable that, based on its implementation,  could be used to stop a trace (I guess it will need a lock).

It won't make much sense to trace render times itself because the view holders will already be there, so how long does it take to the adapter to be notified of changes seems to be enough.

What do you think? If you it's sound to you, iOS has something similar?

dshr:
Good point, I'll have a look. iOS doesn't have a similar `CollectionChange`, but maybe I can do something with KVO or overriding the `ReloadData` method.. 🤔

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
```diff
@@ -658,6 +660,9 @@ private void onParsedQuery(QueryInfo parsedQuery)
 
             IsSuggestingTags = suggestsTags;
             IsSuggestingProjects = suggestsProjects;
+
+            suggestionsLoadingTrace = stopwatchProvider.Create(MeasuredOperation.StartTimeEntrySuggestionsLoadingTime);
```
simonrozsival: I am not sure this is a good way of doing this. What happens when user types a second letter before the first fetching of suggestions finishes? The original stopwatch reference will be overwritten and we will incorrect results, won't we?

Toggl.Daneel/ViewSources/StartTimeEntryTableViewSource.cs
```diff
@@ -46,6 +47,8 @@ public sealed class StartTimeEntryTableViewSource : CreateSuggestionGroupedTable
 
         public bool ShouldShowNoProjectsInfoMessage { get; set; }
 
+        public Func<Unit> TableRenderCallback { get; set; }
```
simonrozsival: What do you need the `Unit` for? This could be just an `Action` (a delegate for a `void` - method).

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
dshr: Doesn't typing stop the fetch operation? If it doesn't, what do you think about a dict like [Query: Trace], and then when we get the data from the DB we stop only the most recent one, and purge the rest?

Toggl.Daneel/ViewSources/StartTimeEntryTableViewSource.cs
dshr: Just replaced it, didn't know about `Actions` in C#, so I was looking for something like `(() -> Void)` in Swift.

Toggl.Foundation.MvvmCross/ViewModels/StartTimeEntryViewModel.cs
simonrozsival: No, it does not stop it. I would suggest something simpler - pass the stopwatch downstream in the observable in a tuple. That could work well.

Toggl.Giskard/Activities/StartTimeEntryActivity.cs
```diff
@@ -51,6 +51,11 @@ protected override void OnCreate(Bundle bundle)
                 .Select(text => text.AsImmutableSpans(editText.SelectionStart))
                 .Subscribe(async spans => await ViewModel.OnTextFieldInfoFromView(spans))
                 .DisposedBy(DisposeBag);
+
+            recyclerView.ViewTreeObserver.GlobalLayout += (sender, args) =>
+            {
+                ViewModel.StopSuggestionsRenderingStopwatch();
```
daividssilverio: I'm surprised this works (I tested it).
Please remove the listener on `OnPause`.

daividssilverio: Oh yeah, add it OnResume, otherwise it won't work after going to the background.

daividssilverio:
Little too late, but I'm "approving" this PR. I was testing it before you merged. It works fine. I've tested in with the firebase .dll and it seems to be working ok.

Just so we are aware. `StartTimeEntrySuggestionsLoadingTime` is not always being stopped, at least based on the logs.

A little sample:
```
16:09:07.557 -> StartTimeEntrySuggestionsRenderingTime 63ms
16:09:08.112 -> StartTimeEntrySuggestionsRenderingTime 30ms
16:09:09.531 -> StartTimeEntrySuggestionsLoadingTime 142ms
16:09:09.540 -> StartTimeEntrySuggestionsRenderingTime 140ms
16:09:10.957 -> StartTimeEntrySuggestionsLoadingTime 19ms
16:09:10.987 -> StartTimeEntrySuggestionsRenderingTime 46ms
16:09:12.509 -> StartTimeEntrySuggestionsLoadingTime 120ms
16:09:12.540 -> StartTimeEntrySuggestionsRenderingTime 141ms
16:09:18.032 -> StartTimeEntrySuggestionsLoadingTime 258ms
16:09:18.063 -> StartTimeEntrySuggestionsRenderingTime 273ms
16:09:18.599 -> StartTimeEntrySuggestionsLoadingTime 115ms
16:09:18.632 -> StartTimeEntrySuggestionsRenderingTime 138ms
16:09:55.048 -> StartTimeEntrySuggestionsRenderingTime 147ms
16:09:56.896 -> StartTimeEntrySuggestionsRenderingTime 512ms
16:09:58.094 -> StartTimeEntrySuggestionsRenderingTime 225ms
16:09:59.930 -> StartTimeEntrySuggestionsRenderingTime 140ms
16:10:01.464 -> StartTimeEntrySuggestionsRenderingTime 144ms
16:10:10.697 -> StartTimeEntrySuggestionsRenderingTime 152ms
16:10:12.480 -> StartTimeEntrySuggestionsRenderingTime 228ms
16:10:18.454 -> StartTimeEntrySuggestionsLoadingTime 15ms
16:10:18.649 -> StartTimeEntrySuggestionsRenderingTime 209ms
16:10:22.441 -> StartTimeEntrySuggestionsLoadingTime 119ms
16:10:22.474 -> StartTimeEntrySuggestionsRenderingTime 144ms
16:10:25.174 -> StartTimeEntrySuggestionsLoadingTime 23ms
16:10:25.227 -> StartTimeEntrySuggestionsRenderingTime 73ms
16:10:28.903 -> StartTimeEntrySuggestionsLoadingTime 121ms
16:10:28.924 -> StartTimeEntrySuggestionsRenderingTime 130ms
16:10:30.151 -> StartTimeEntrySuggestionsRenderingTime 145ms
16:10:33.030 -> StartTimeEntrySuggestionsRenderingTime 192ms
16:10:33.425 -> StartTimeEntrySuggestionsRenderingTime 112ms
16:10:33.915 -> StartTimeEntrySuggestionsRenderingTime 103ms
16:10:34.569 -> StartTimeEntrySuggestionsRenderingTime 108ms
16:10:35.046 -> StartTimeEntrySuggestionsRenderingTime 193ms
16:10:40.547 -> StartTimeEntrySuggestionsRenderingTime 141ms
16:10:41.440 -> StartTimeEntrySuggestionsRenderingTime 105ms
16:10:48.572 -> StartTimeEntrySuggestionsLoadingTime 112ms
16:10:48.598 -> StartTimeEntrySuggestionsRenderingTime 130ms
16:11:02.644 -> StartTimeEntrySuggestionsLoadingTime 194ms
16:11:02.683 -> StartTimeEntrySuggestionsRenderingTime 217ms
```

PR #3459, opened 2018-10-16
Disable UI elements when the time entry is ghost

juanlaube:
## What's this?
Disables or sets a light grey text color for all non-editable UI elements. This is only for iOS, a PR for Android will follow soon.

### Relationships

Ref #2904 

## Why do we want this?
To be clear about what can be edited in a ghost time entry.

## How is it done?
By setting the text color for all the labels/textfields and disabling the billable switch.

### Side effects
None.

## Review hints
I'm helping with an screenshot 😄 

## :squid: Permissions
Whenever, anyone

simulator screen shot - iphone xs max - 2018-10-16 at 10 24 56


juanlaube:
Now without the tags row when the TE is ghost and there are no tags. Only project is editable.
simulator screen shot - iphone xs max - 2018-10-16 at 11 25 48

Toggl.Foundation.MvvmCross/Combiners/ShowTagsValueCombiner.cs
```diff
@@ -0,0 +1,15 @@
...
+namespace Toggl.Foundation.MvvmCross.Combiners
+{
+    public sealed class ShowTagsValueCombiner : BaseTwoValuesCombiner<bool, bool>
+    {
+        public ShowTagsValueCombiner()
+        {
+        }
+
+        protected override object Combine(bool isGhost, bool hasTags)
+            => isGhost && !hasTags;
```
simonrozsival: Shouldn't this be `!isGhost || hasTags`?!

juanlaube: Either that or rename the class to `HideTagsValueCombiner` 😄

PR #3461, opened 2018-10-16
Sanic tapes: Open Settings view

### LINKED ISSUE #3410: Measure the time it takes to open the settings view
Measure the time it takes to navigate from the main view to the settings view

dshr:
## What's this?
Adding traces to measure the time it takes to open the settings view from the main view.

### Relationships

Part of #3400 
Closes #3410 

## Why do we want this?
We want to know how long this action takes.

## How is it done?
Start a trace before navigation to the Settings VM, stop it when the it's view appears.

### Why not another way?
Seems like the simplest one.

### Side effects
None.

## Review hints
None.

## :squid: Permissions
:sanic: squad only


Toggl.Foundation.MvvmCross/ViewModels/Settings/SettingsViewModel.cs
```diff
@@ -50,11 +51,13 @@ public sealed class SettingsViewModel : MvxViewModel
         private readonly IMvxNavigationService navigationService;
         private readonly IPrivateSharedStorageService privateSharedStorageService;
         private readonly IIntentDonationService intentDonationService;
+        private readonly IStopwatchProvider stopwatchProvider;
 
         private bool isSyncing;
         private bool isLoggingOut;
         private IThreadSafeUser currentUser;
         private IThreadSafePreferences currentPreferences;
+        private IStopwatch settingsStopwatch;
```
daividssilverio: Same here. I didn't give much to this before, but naming is important.
Could you make the name more specific to express what this stopwatch is tracing?

daividssilverio:
I won't request changes specifically because it's just a naming question. Up to you.

PR #3462, opened 2018-10-16
Sanic tapes: Open giskard reports

### LINKED ISSUE #3408: Measure the time it takes to open the reports view
Since android is not using the tabs, let's measure the time it takes for the view to fully load when being opened from the main view

daividssilverio:
## What's this?
This adds a trace from the moment the user clicks on reports icon on the main toolbar to the point the enter animation ends and the reports screen is visible.

### Relationships
Part of #3400 
Closes #3408 

## Why do we want this?
This transition feels kind of slow. We want to measure, try to make it better and measure again.

## How is it done?
A stopwatch is started on the main vm, and stopped on the `ReportsActivity:OnEnterAnimationComplete`. That's it.

### Why not another way?
`ViewAppeared` on the `ReportsViewModel` didn't reflect the actual moment the activity became visible.

### Side effects
None.

## Review hints
This is a very small PR, no hints.

## :squid: Permissions
:sanic: squad. 🦑 please


Toggl.Foundation.MvvmCross/ViewModels/Reports/ReportsViewModel.cs
```diff
@@ -268,6 +273,13 @@ public override void ViewAppeared()
             }
         }
 
+        public void StopTraceFromMainLog()
```
dshr: I'm not a fan of the naming here — can we perhaps name it something along the lines `StopNavigationTimeTrace`, since we're measuring the navigation time here?

Toggl.Foundation.MvvmCross/ViewModels/Reports/ReportsViewModel.cs
```diff
@@ -268,6 +273,13 @@ public override void ViewAppeared()
...
+        {
+            var stopwatchFromMainLog = stopwatchProvider.Get(MeasuredOperation.OpenReportsFromGiskard);
```
dshr: Same thing, maybe something more descriptive, like `reportsNavigationStopwatch`?

Toggl.Foundation.MvvmCross/ViewModels/Reports/ReportsViewModel.cs
daividssilverio: Also, very generic. I was looking for a specific naming pattern.
What about: `StopNavigationFromMainLogTimeTrace`?

Toggl.Foundation.MvvmCross/ViewModels/Reports/ReportsViewModel.cs
daividssilverio: Same idea, I'd like to be as specific as possible, `reportsNavigationFromMainLogStopwatch`?

PR #3464, opened 2018-10-16
Sanic tapes: Open add project view from edit te view

### LINKED ISSUE #3407: Measure the time it takes to open the add project view
Measure the navigation time from the edit time entry view to the add project view

daividssilverio:
## What's this?
This add a trace for the navigation between the edit the view and the moment the select project view appears.

### Relationships
Part of #3400
Closes #3407

## Why do we want this?
We are adding a bunch o traces to measure pain points. 

## How is it done?
A trace is started when the user clicks the Add Project area on the EditTE view and the moment the select project view appears.

### Why not another way?
This is simple enough. This measures the navigation only (to the point the target view appears).
We might want to measure the time to load projects and render as well, but not here.

### Side effects
None.

## Review hints
None.

## :squid: Permissions
:sanic: squad. 🦑 please


Toggl.Foundation.MvvmCross/ViewModels/SelectProjectViewModel.cs
```diff
@@ -33,11 +34,13 @@ public sealed class SelectProjectViewModel
         private readonly IDialogService dialogService;
         private readonly IInteractorFactory interactorFactory;
         private readonly IMvxNavigationService navigationService;
+        private readonly IStopwatchProvider stopwatchProvider;
         private readonly Subject<string> infoSubject = new Subject<string>();
 
         private long? taskId;
         private long? projectId;
         private long workspaceId;
+        private IStopwatch stopwatchFromEditView;
```
dshr: Why not `selectProjectStopwatch`?

daividssilverio: `selectProjectStopwatch` sounds too generic to me. It looks like it's the stopwatch for this VM. What I want to express is the stopwatch that was started in the editview and will be stopped in the select project vm.

PR #3472, opened 2018-10-17
Ask user to pick default workspace

### LINKED ISSUE #2551: Ask user to pick a default WS when a default one is not assigned
After pulling and persisting the user object, and the user has no default workspace set:

- if there is exactly one workspace, set that one as default, mark the user object as needing sync
- if there is more than one workspace, trigger a dialog that lets the user pick one

Platform-specific issues:
#2462
#2543

arturs105:
## What's this?
This PR adds the necessary modifications to Thy Sync to make this work. This also adds the `SelectDefaultWorkspaceViewModel`. iOS view controller and Android activity will come in separate pull requests.

### Relationships

Closes #2551

## How is it done?
1. `DetectNoDefaultWorkspaceState` detects, whether the user has a default workspace set.
2. If there is no default workspace, then `TrackNoDefaultWorkspaceState` tracks the appropriate event. Previously this state also checked, if there is a default workspace or not, but since now another state does that, i removed this part from the tracking state.
3. `TrySetDefaultWorkspaceState` sets the default workspace, but only if there is a single workspace in the data source. If there are several, then it throws `NoDefaultWorkspaceException`.

## :squid: Permissions
Whenever.


Toggl.Foundation/Sync/States/Pull/NoDefaultWorkspaceTrackingState.cs
```diff
@@ -22,15 +22,7 @@ public NoDefaultWorkspaceTrackingState(IAnalyticsService analyticsService)
 
         public IObservable<ITransition> Start(IFetchObservables fetch)
             => fetch.GetSingle<IUser>()
-                .Do(trackIfNoDefaultWorkspace)
+                .Do(analyticsService.NoDefaultWorkspace.Track)
```
simonrozsival: I don't think this state makes sense now, it adds unnecessary complexity to the syncing graph. We should move the tracking into `NoDefaultWorkspaceDetectingState`.

Toggl.Foundation/Sync/States/Pull/NoDefaultWorkspaceDetectingState.cs
```diff
@@ -0,0 +1,18 @@
...
+
+namespace Toggl.Foundation.Sync.States.Pull
+{
+    internal sealed class NoDefaultWorkspaceDetectingState : ISyncState<IFetchObservables>
+    {
+        public StateResult<IFetchObservables> Continue { get; } = new StateResult<IFetchObservables>();
+        public StateResult<IFetchObservables> NoDefaultWorkspaceDetected { get; } = new StateResult<IFetchObservables>();
+
+        public IObservable<ITransition> Start(IFetchObservables fetch)
+            => fetch.GetSingle<IUser>()
```
simonrozsival: We should not check what the server gave us, but what is in the database - our conflict resolution might have ignored the data from the server (e.g., because the user changed his workspace in the app, but it hadn't been pushed yet) and the user might have a workspace.

Toggl.Foundation/Interactors/InteractorFactory.Workspace.cs
```diff
@@ -31,5 +33,6 @@ public IInteractor<IObservable<bool>> IsBillableAvailableForWorkspace(long works
 
         public IInteractor<IObservable<Unit>> CreateDefaultWorkspace()
             => new CreateDefaultWorkspaceInteractor(idProvider, timeService, dataSource.User, dataSource.Workspaces);
+
```
simonrozsival: 🔥

Toggl.Foundation/Sync/States/Pull/TrySetDefaultWorkspaceState.cs
```diff
@@ -0,0 +1,60 @@
...
+                .Workspaces
+                .GetAll()
+                .Select(getDefaulWorkspaceIfPossible)
+                .SelectMany(setDefaultWorkspaceIfPossible)
+                .Select(_ => Continue.Transition(fetch));
+
+        private IThreadSafeWorkspace getDefaulWorkspaceIfPossible(IEnumerable<IThreadSafeWorkspace> workspaces)
+            => workspaces.Count() == 1
+                ? workspaces.First()
+                : null;
```
simonrozsival: I think it makes more sense to throw here, not in the other method.

Toggl.Foundation/Interactors/Workspace/SetDefaultWorkspaceInteractor.cs
```diff
@@ -0,0 +1,40 @@
...
+
+namespace Toggl.Foundation.Interactors
+{
+    public sealed class SetDefaultWorkspaceInteractor : IInteractor<IObservable<Unit>>
+    {
+        private readonly long workspaceId;
+        private readonly ITogglDataSource dataSource;
+
+        public SetDefaultWorkspaceInteractor(
+            ITogglDataSource dataSource,
```
simonrozsival: `ISingletonDataSource<IThreadSafeUser> dataSource` would be enough

Toggl.Foundation/Sync/States/Pull/TrySetDefaultWorkspaceState.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        private IObservable<Unit> setDefaultWorkspaceIfPossible(IThreadSafeWorkspace workspace)
+        {
+            if (workspace == null)
+                throw new NoDefaultWorkspaceException();
+
+            return dataSource
```
simonrozsival: You can use the interactor here (directly, without the interactor factory) - `new SetDefaultWorkspaceInteractor(dataSource, workspace.Id).Execute();`

Toggl.Foundation/Sync/States/Pull/TrySetDefaultWorkspaceState.cs
```diff
@@ -0,0 +1,60 @@
...
+                .User
+                .Current
+                .Select(User.Builder.FromExisting)
+                .Select(user => user
+                    .SetDefaultWorkspaceId(workspace.Id)
+                    .SetSyncStatus(SyncStatus.SyncNeeded)
```
simonrozsival: You must update `At`

Toggl.Foundation/Interactors/Workspace/SetDefaultWorkspaceInteractor.cs
```diff
@@ -0,0 +1,40 @@
...
+        }
+
+        public IObservable<Unit> Execute()
+            => dataSource
+                .User
+                .Current
+                .Select(User.Builder.FromExisting)
+                .Select(user => user
+                    .SetDefaultWorkspaceId(workspaceId)
+                    .SetSyncStatus(SyncStatus.SyncNeeded)
```
simonrozsival: You must also update `At`

Toggl.Foundation.MvvmCross/ViewModels/SelectDefaultWorkspaceViewModel.cs
```diff
@@ -0,0 +1,54 @@
...
+
+        public override async Task Initialize()
+        {
+            await base.Initialize();
+
+            await dataSource
+                .Workspaces
+                .GetAll()
+                .Select(workspaces => workspaces.Select(toSelectable))
+                .Do(Workspaces.AddRange);
```
simonrozsival: I think we should do something if the list of workspaces is empty after the initialization. We should probably navigate to no-workspace-at-all 🤔 or just throw an exception and crash?

arturs105: I'm leaning towards throwing, because the sync machine has the `NoWorkspaceDetectingState` and checking for no workspaces happens before checking for no default workspace, so in theory this view model should never be used when there are no workspaces. If this view is presented when there are no workspaces, then something has gone wrong, hence we should throw.

simonrozsival: Add some good message to the exception so we will notice it in the AppCenter if it happens too often.

Toggl.Foundation.Tests/Sync/States/Pull/TrySetDefaultWorkspaceStateTests.cs
```diff
@@ -0,0 +1,97 @@
...
+    {
+        private readonly ITogglDataSource dataSource = Substitute.For<ITogglDataSource>();
+        private readonly IFetchObservables fetch = Substitute.For<IFetchObservables>();
+        private readonly ITimeService timeService = Substitute.For<ITimeService>();
+        private readonly TrySetDefaultWorkspaceState state;
+
+        public TrySetDefaultWorkspaceStateTests()
+        {
+            state = new TrySetDefaultWorkspaceState(timeService, dataSource);
+            var user = Substitute.For<IThreadSafeUser>();
```
simonrozsival: Use `MockUser` instead.

simonrozsival: Also please set the `DefaultWorkspaceId` to `null` explicitly to simulate the real world situation.

Toggl.Foundation.Tests/Sync/States/Pull/TrySetDefaultWorkspaceStateTests.cs
```diff
@@ -0,0 +1,97 @@
...
+        public async Task SetsSyncStatusToSyncNeededOnUserObjectWhenThereIsOnlyOneWorkspace()
+        {
+            setupWorkspaces(1);
+
+            await state.Start(fetch);
+
+            await dataSource.User.Received().Update(
+                Arg.Is<IThreadSafeUser>(user => user.SyncStatus == SyncStatus.SyncNeeded));
+        }
+
```
simonrozsival: Please add a test that `At` is set to the current timestamp.

Toggl.Foundation/Sync/States/Pull/TrySetDefaultWorkspaceState.cs
```diff
@@ -0,0 +1,47 @@
...
+using Toggl.Foundation.DataSources;
+using Toggl.Foundation.Exceptions;
+using Toggl.Foundation.Interactors;
+using Toggl.Foundation.Models.Interfaces;
+using Toggl.Multivac;
+using Toggl.Multivac.Extensions;
+
+namespace Toggl.Foundation.Sync.States.Pull
+{
+    public sealed class TrySetDefaultWorkspaceState : ISyncState<IFetchObservables>
```
simonrozsival: Let's simplify this state a bit - we don't need the `IFetchObservables` as the input and we can move it so it doesn't even need to pass it to the following states.

We can easily persist all the data we pull from the server and at the very end of the `pull` process check for the default workspace. For the tracking state we had previously it was OK to have it at the beginning of the transitions, but now we would throw away all of the data pulled from the server and re-fetch them once user finishes selecting a workspace (even though not having a workspace is an edge case..).

Toggl.Foundation.Tests/Interactors/Workspace/SetDefaultWorkspaceInteractorTests.cs
```diff
@@ -0,0 +1,77 @@
...
+
+                tryingToConstructWithNull.Should().Throw<ArgumentNullException>();
+            }
+        }
+
+        public sealed class TheExecuteMethod : BaseInteractorTests
+        {
+            public TheExecuteMethod()
+            {
+                var user = Substitute.For<IThreadSafeUser>();
```
simonrozsival: Use `MockUser` please.

Toggl.Foundation/Sync/States/Pull/TrySetDefaultWorkspaceState.cs
arturs105: So, if i understood correctly:
1. Remove make this state a `ISyncState` (remove `IFetchObservables`)
2. Move this state to the end of the pull state machine?

simonrozsival: Yes. That is what I tried to say 👍

Issue #3445, opened 2018-10-15
Do not use the "ghost" word in our code


simonrozsival:
We use the term "ghost" for two cases now:
1. when we create a temporary entity instead of an archived project (see `CreateGhostProjectsState` and `DeleteNonReferencedProjectGhostsState`)
2. when user loses access to a workspace

- [x] In the first case, I suggest renaming the methods and related references in the code to "temporary entity" or explicitly "archived project placeholder" (I prefer the latter but I will leave this one on the person creating the PR)
- [x] In the second case, let's call the properties, all references in code, comments, and documentation, using the term "inaccessble": `IsInaccesssible`, "inaccessible entity", "mark as inaccessible", and so on

simonrozsival:
Also please feel free to suggest different names here in comments.

dshr:
Why not ghosts and ghouls? :trollface:

simonrozsival:
All the usages of the word 👻 are gone.

PR #3474, opened 2018-10-17
Rename IGhostable to IPotentiallyInaccessible

simonrozsival:
## What's this?
This is "just" a renaming PR. It renames the `IGhostable` interface and its `bool IsGhost` property to a name which describes its purpose much better - `IPotentiallyInaccessible` and `bool IsInaccessible`.

### Relationships
Ref #3445 - this is just the first of a series of PRs which will rename different parts of the code

## Why do we want this?
We want our code to be easier to understand to people even without deep knowledge of internal terminology (what does it mean that an entity is a ghost?)

## How is it done?
Renaming + a database migration.

### Side effects
An incorrect migration could lead to corrupting users' data. We must be careful.

## Review hints
Check the changes commit by commit.

Please check I didn't do anything stupid in the migration 🙈 

## :squid: Permissions
I will squash this myself when approved


kornelijepetak:
Somehow I find `potentially` better than `possibly`... Not a request for change, just a thought...

simonrozsival:
Or maybe `Potentially`?

Toggl.PrimeRadiant.Realm/Database.cs
```diff
@@ -98,6 +99,18 @@ private RealmConfiguration createRealmConfiguration()
                     {
                         // nothing needs explicit updating when updating from schema 4 up to 6
                     }
+
+                    if (oldSchemaVersion < 7)
+                    {
+                        var newWorkspaces = migration.NewRealm.All<RealmWorkspace>();
+                        var oldWorkspaces = migration.OldRealm.All("RealmWorkspace");
```
kornelijepetak: why not a `nameof()`?
```suggestion
                        var oldWorkspaces = migration.OldRealm.All(nameof(RealmWorkspace));
```

Toggl.PrimeRadiant.Realm/Database.cs
```diff
@@ -98,6 +99,18 @@ private RealmConfiguration createRealmConfiguration()
...
+                        for (var i = 0; i < newWorkspaces.Count(); i++)
```
kornelijepetak: Does this mean that we still have the IsGhost value in db, but we are just not using it anymore? How does Realm deal with this?

simonrozsival: Yes. Realm uses `dynamic` to return something which contains the data under the old name.

Toggl.PrimeRadiant.Realm/Database.cs
simonrozsival: The name of the class might change in the future (let’s not do that though!) and the table will still be accessible only under the old name.

Toggl.PrimeRadiant.Realm/Database.cs
```diff
@@ -98,6 +99,18 @@ private RealmConfiguration createRealmConfiguration()
                     {
                         // nothing needs explicit updating when updating from schema 4 up to 6
                     }
+
+                    if (oldSchemaVersion < 7)
+                    {
+                        var newWorkspaces = migration.NewRealm.All<RealmWorkspace>();
```
kornelijepetak: `All<RealmWorkspace>()` returns `IQueryable<T>` which means that each iteration of for loop will retrieve ALL workspaces all over again.

Add `.ToList()` to each call and also access items by index then (instead of `ElementAt`).

Toggl.PrimeRadiant.Realm/Database.cs
kornelijepetak: Don't you think it would be better to change it to specific name when that thing happens?
I am sorry to be a pain in the ass for such a simple PR...

PR #3476, opened 2018-10-17
Add state to delete synced ghost time entries

### LINKED ISSUE #2631: WS removal: remove synced TEs of lost workspace

juanlaube:
## What's this?
Adds a new state to delete all synced, inaccessible time entries. The state is not connected to the syncing graph yet.

### Relationships

Closes #2631

## Why do we want this?
To 🔥 👻 ⏳ 

## How is it done?
It's a new state, goes through all time entries suitable for deletion and burns them all 🔥

### Why not another way?
I could see a generic 🔥 state in the near future because most of them are the same with a different data source and predicate, I think we can refactor this later.

### Side effects
None.

## :squid: Permissions
Anyone, whenever.


Toggl.Foundation/Sync/States/CleanUp/DeleteInaccessibleTimeEntriesState.cs
```diff
@@ -0,0 +1,34 @@
...
+
+        public DeleteInaccessibleTimeEntriesState(ITimeEntriesSource dataSource)
+        {
+            Ensure.Argument.IsNotNull(dataSource, nameof(dataSource));
+            this.dataSource = dataSource;
+        }
+
+        public IObservable<ITransition> Start()
+            => dataSource
+                .GetAll(suitableForDeletion, true)
```
simonrozsival:
```suggestion
                .GetAll(suitableForDeletion, includeGhosts: true)
```

Toggl.Foundation.Tests/Sync/States/CleanUp/DeleteInaccessibleTimeEntriesStateTests.cs
```diff
@@ -0,0 +1,150 @@
...
+
+            await state.Start().SingleAsync();
+
+            dataSource
+                .Received()
+                .DeleteAll(Arg.Is<IEnumerable<IThreadSafeTimeEntry>>(arg => arg.SequenceEqual(syncedTimeEntries)));
+
+            dataSource
+                .DidNotReceive()
+                .DeleteAll(Arg.Is<IEnumerable<IThreadSafeTimeEntry>>(Arg => Arg.SequenceEqual(unsyncedTimeEntries)));
```
simonrozsival: `Arg => Arg.SequenceEqual(unsyncedTimeEntries)` does this really do what you want? Shouldn't it be something like `Arg => Arg.None(deletedTE => unsyncedTimeEntries.Contains(deletedTE))`?

Toggl.Foundation.Tests/Sync/States/CleanUp/DeleteInaccessibleTimeEntriesStateTests.cs
```diff
@@ -0,0 +1,150 @@
...
+        }
+
+        [Fact, LogIfTooSlow]
+        public async Task OnlyDeletesSyncedInaccessibleTimeEntries()
+        {
+            var workspaceA = getWorkspace(1000, true);
```
simonrozsival: Make reader's life easier by `isInaccessible: true` 👍

Toggl.Foundation.Tests/Sync/States/CleanUp/DeleteInaccessibleTimeEntriesStateTests.cs
```diff
@@ -0,0 +1,150 @@
...
+
+            await state.Start().SingleAsync();
+
+            dataSource
+                .Received()
+                .DeleteAll(Arg.Is<IEnumerable<IThreadSafeTimeEntry>>(arg => arg.SequenceEqual(syncedInaccessibleTimeEntries)));
+
+            dataSource
+                .DidNotReceive()
+                .DeleteAll(Arg.Is<IEnumerable<IThreadSafeTimeEntry>>(arg => arg.SequenceEqual(undeletableTimeEntries)));
```
simonrozsival: Same as above.

Toggl.Foundation.Tests/Sync/States/CleanUp/DeleteInaccessibleTimeEntriesStateTests.cs
simonrozsival: it might be better to check `unsyncedTimeEntries.Select(te => te.Id).Contains(deletedTE.Id)`, not the reference equality.

juanlaube: I think this reads better and tests what we want:
```
arg => arg.All(te => syncedTimeEntries.Contains(te)) &&
       arg.None(te => unsyncedTimeEntries.Contains(te))
```

PR #3478, opened 2018-10-17
Consider ghostness edit view

### LINKED ISSUE #2902: Consider ghost-ness in edit view

juanlaube:
## What's this?
This disable all user interaction in the edit view except for selecting the project/workspace or deleting the time entry when the time entry is inaccessible.

### Relationships

Closes #2902

## Why do we want this?
Users should only be able to select a new project or delete the time entry.

## How is it done?
Since the EditTimeEntryViewModel still uses the old mvx bindings, I added a canExecute parameter to all the commands. In the process I removed some code that was platform specific.

### Why not another way?
The other option is an rxfactor.

### Side effects
I fixed two UI problems on the way:
1. The visibility of the tags row was wrong when the time entry is inaccessible and there are no tags
2. The description text color was inverted

## Review hints
Test both apps if possible.

## :squid: Permissions
Anyone, whenever.


khoi:
I tested it on iOS and it works. 

I think the UI need some redesign though, it's very difficult to tell which options are available and which are not. Especially that when 👻-ness is something that's hidden from the user.

screen shot 2018-10-18 at 10 52 28

PR #3481, opened 2018-10-17
Sanic tapes: open reports view for the first time

### LINKED ISSUE #3409: Measure the time it takes to open the reports tab
Since iOS has the tabs approach, only the first time the reports view is opened has an actual hiccup. Let's measure this as well. It might not be doable from the shared code layer, but let's give it a try before resorting to platform specific code.

dshr:
## What's this?
Tracing the time it takes to open the Reports view for the first time round.

### Relationships
Part of #3400 
Closes #3409

## Why do we want this?
We want to know how long it takes exactly.

## How is it done?
We start a trace in the main tab bar vm, and stop it when the reports view appears. We also keep a flag that makes sure we only do this once.

### Why not another way?
This seems like the simplest one.

### Side effects
None.

## Review hints
Nope.

## :squid: Permissions
:sanic: folks only


Toggl.Daneel/ViewControllers/MainTabBarController.cs
```diff
@@ -45,5 +45,15 @@ public override void ViewDidLoad()
 
             TabBar.Translucent = UIDevice.CurrentDevice.CheckSystemVersion(11, 0);
         }
+
+        public override void ItemSelected(UITabBar tabbar, UITabBarItem item)
+        {
+            var targetViewController = ViewControllers.Single(vc => vc.TabBarItem == item);
+            if (targetViewController is UINavigationController
```
daividssilverio: You can avoid the cast like this:
```
if (targetViewController is UINavigationController uiNavigationController 
                && uiNavigationController.TopViewController is ReportsViewController)
```

Toggl.Daneel/ViewControllers/MainTabBarController.cs
```diff
@@ -45,5 +45,15 @@ public override void ViewDidLoad()
 
             TabBar.Translucent = UIDevice.CurrentDevice.CheckSystemVersion(11, 0);
         }
+
+        public override void ItemSelected(UITabBar tabbar, UITabBarItem item)
+        {
+            var targetViewController = ViewControllers.Single(vc => vc.TabBarItem == item);
```
daividssilverio: I don't know how UITabBarItem works, but do you want to use reference comparison? Is it safe to use `==` here?

Toggl.Foundation.MvvmCross/ViewModels/Reports/ReportsViewModel.cs
```diff
@@ -75,6 +75,7 @@ public sealed class ReportsViewModel : MvxViewModel<ReportPeriod>
         private DateFormat dateFormat;
         private IReadOnlyList<ChartSegment> segments = new ChartSegment[0];
         private IReadOnlyList<ChartSegment> groupedSegments = new ChartSegment[0];
+        private IStopwatch firstTimeOpenStopwatch;
```
daividssilverio: `firstTimeOpenedFromMainTabBarStopwatch` would be more clear?

Toggl.Daneel/ViewControllers/MainTabBarController.cs
dshr: Yeah, it's perfectly safe to do this in UIKit, at least as far as I know. It's supposed to be passed by reference here, but I'll double-check just in case.

Toggl.Daneel/ViewControllers/MainTabBarController.cs
dshr: Cheers! ✨

Toggl.Foundation.MvvmCross/ViewModels/Reports/ReportsViewModel.cs
dshr: Yeah, good point, changing it.

Toggl.Daneel/ViewControllers/MainTabBarController.cs
dshr: Yeah, they def are references.

Toggl.Foundation.MvvmCross/ViewModels/Reports/ReportsViewModel.cs
```diff
@@ -75,6 +75,7 @@ public sealed class ReportsViewModel : MvxViewModel<ReportPeriod>
         private DateFormat dateFormat;
         private IReadOnlyList<ChartSegment> segments = new ChartSegment[0];
         private IReadOnlyList<ChartSegment> groupedSegments = new ChartSegment[0];
+        private IStopwatch firstTimeOpenedFromMainTabBarStopwatch;
```
daividssilverio: This doesn't have to be a field, right? You are populating and nullifying it in the same method down there.

dshr: True, will change that!

PR #3487, opened 2018-10-18
Remove ghosts from Foundation

simonrozsival:
## What's this?
🔥 👻 

### Relationships

Ref #3445 

### Side effects
This PR changes the name of an analytics event. This will cause problems when analyzing data in AppCenter. I think it is OK.

## Review hints
Search for the word "ghost" in the `Foundation` project. It should not be there.

## :squid: Permissions
🦑


khoi:
you might want to update these files as well
screen shot 2018-10-18 at 19 09 23

simonrozsival:
Yes, I will create a separate PR for docs. Thanks for pointing this out!

PR #3490, opened 2018-10-18
Refactor RX Actions

simonrozsival:
## What's this?
This PR refactors the `RxAction`, `UIAction`, and `InputAction` classes and their usages.

The goal is to make the code in VMs cleaner and more readable.

### Relationships
I am not aware of any issue related to this PR. This is my own initiative.

## Why do we want this?
The code in our VMs feels weird after some rxfactorizations - too much boilerplate where it shouldn't be IMO.

## How is it done?
By adding factories which turn `Action` and `Func<Task>` into an `UIAction` or `InputAction`. These functions then contain the weird creation of observables, that was spread across the whole solution before.

### Why not another way?
I think this is a good way of doing things. Maybe this PR is too big though 🤔 Feedback is welcome.

One thing I am thinking about is to make the constructors private and add another factory `FromObservable` - this way initializing an `UIAction` or `InputAction` would look the same everywhere.

### Side effects
No side effects.

## Review hints
I recommend going commit by commit instead of looking at the whole diff.

## :squid: Permissions
🦑


khoi:
I have mixed feelings about this. 
- It does make our code look cleaner.
- What bothers me is that with the convenience provided, we'd tend to use `Task` and returns it for all ViewModel's methods. With a `Task` there is not much you can do, while with `Observable` there are a whole set of operators that we can use.

kornelijepetak:
> What bothers me is that with the convenience provided, we'd tend to use Task and returns it for all ViewModel's methods. With a Task there is not much you can do, while with Observable there are a whole set of operators that we can use.

I don't agree with this really.

`Tasks` and `async/await` is the proper C#/NET way of doing things. Many of the actions are simply a sequences of steps to take or things to do. Some of those steps are synchronous, some are asynchronous that require waiting, and some are asynchronous that are fire and forget. With RX, as I've showed recently, for even a slightly complicated scenarios, things quickly become extremely convoluted, for example: just to handle a simple conditional await. If it's really necessary, you can also convert a task into an observable by using `ToObservable`. 

In all honesty, I don't think RX was made for that purpose. Seriously. FlatMap (SelectMany) is an operator that is supposed to map the *emitted items* into something else and then flatten the results. And what we are in many cases doing is that we ignore the emitted items (or just use them as a triggers) to run some actions. I believe that is conceptually wrong. This is not debuggable and not really readable.

`Tasks` and `async/await` make the sequence of steps readable and easy to understand what's happening and when. It is also much easier to debug (debugging with breakpoints actually work well with async await). 

Converting every single thing in our codebase to an observable will not take us anywhere.

Toggl.Foundation.MvvmCross/ViewModels/Calendar/CalendarViewModel.cs
```diff
@@ -114,39 +114,34 @@ public sealed class CalendarViewModel : MvxViewModel
...
+            OnUpdateTimeEntry = InputAction<CalendarItem>.FromAsync(onUpdateTimeEntry);
 
             CalendarItems = new ObservableGroupedOrderedCollection<CalendarItem>(
                 indexKey: item => item.StartTime,
                 orderingKey: item => item.StartTime,
                 groupingKey: _ => 0);
         }
 
-        public void Init(string eventId)
+        public async void Init(string eventId)
```
willsb: No async code on Init, ever.

Toggl.Multivac/Extensions/RxAction/InputAction.cs
```diff
@@ -1,17 +1,42 @@
...
     public class InputAction<TInput> : RxAction<TInput, Unit>
     {
-        public InputAction(Func<TInput, IObservable<Unit>> workFactory) : base(workFactory)
+        private InputAction(Func<TInput, IObservable<Unit>> workFactory, IObservable<bool> enabledIf = null)
+            : base(workFactory, enabledIf)
         {
         }
 
-        public InputAction(Func<TInput, IObservable<Unit>> workFactory, IObservable<bool> enabledIf) : base(workFactory, enabledIf)
+        public static InputAction<TInput> From(Action<TInput> action)
```
willsb: I'd rename this to `FromAction`, since all the other methods are more descriptive (and I like such description)

Toggl.Multivac/Extensions/RxAction/UIAction.cs
```diff
@@ -2,25 +2,22 @@
...
+        private UIAction(Func<IObservable<Unit>> workFactory, IObservable<bool> enabledIf)
             : base(_ => workFactory(), enabledIf)
         {
         }
 
         public IObservable<Unit> Execute()
             => Execute(Unit.Default);
 
-        public static UIAction FromAction(Action action)
+        public static UIAction From(Action action, IObservable<bool> enabledIf = null)
```
willsb: Dunno why this was renamed

Toggl.Foundation.MvvmCross/ViewModels/Calendar/CalendarViewModel.cs
```diff
@@ -114,39 +114,34 @@ public sealed class CalendarViewModel : MvxViewModel
...
                 .DistinctUntilChanged();
 
-            SelectCalendars = new UIAction(() => selectUserCalendars(false), SettingsAreVisible);
+            SelectCalendars = UIAction.FromAsyncAction(() => selectUserCalendars(false), SettingsAreVisible);
 
-            OnDurationSelected = new InputAction<(DateTimeOffset StartTime, TimeSpan Duration)>(
-                tuple => onDurationSelected(tuple.StartTime, tuple.Duration)
-            );
+            OnDurationSelected = InputAction<(DateTimeOffset StartTime, TimeSpan Duration)>.FromAsyncAction(
+                tuple => onDurationSelected(tuple.StartTime, tuple.Duration));
```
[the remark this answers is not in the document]
simonrozsival: The tuple is one value and it can't be deconstructed to two separate arguments for the function. That is my best guess in this case. Tuples are ugly... :-/

Toggl.Foundation.MvvmCross/ViewModels/Calendar/CalendarViewModel.cs
```diff
@@ -114,39 +114,34 @@ public sealed class CalendarViewModel : MvxViewModel
...
+            SelectCalendars = UIAction.FromAsync(() => selectUserCalendars(false), SettingsAreVisible);
 
-            OnDurationSelected = new InputAction<(DateTimeOffset StartTime, TimeSpan Duration)>(
-                tuple => onDurationSelected(tuple.StartTime, tuple.Duration)
-            );
+            OnDurationSelected = InputAction<(DateTimeOffset StartTime, TimeSpan Duration)>.FromAsync(
+                tuple => onDurationSelected(tuple.StartTime, tuple.Duration));
 
-            OnUpdateTimeEntry = new InputAction<CalendarItem>(onUpdateTimeEntry);
+            OnUpdateTimeEntry = InputAction<CalendarItem>.FromAsync(onUpdateTimeEntry);
```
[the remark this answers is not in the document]
simonrozsival: I agree with you. The `on` prefix is weird and should not be there.

PR #3491, opened 2018-10-18
Add state to delete unneeded tags

juanlaube:
## What's this?
Adds a new generic state to delete unreferenced entities. All the deletions we need to do for a particular inaccessible entity are just referenced by only one entity (e.g: Tags depends on TimeEntry, Project depends on Client). This adds this generic sync state and the implementation for tags.

### Relationships

Closes #2633

## Why do we want this?
👻 🔥 

## How is it done?
As I said, new sync state 😄 

### Why not another way?
We could repeat the same code in 5-6 different classes, but we don't want that.

### Side effects
The only side effect is this makes everything easier for the next 👻 🔥 issues 😉 

## :squid: Permissions
Whenever, anyone, but needs @simonrozsival ✅ first.


Toggl.Foundation/Sync/States/CleanUp/DeleteEntityState.cs
```diff
@@ -0,0 +1,46 @@
...
+                .SelectMany(CommonFunctions.Identity)
+                .SelectMany(SuitableForDeletionOrNull)
+                .Where(entity => entity != null)
+                .ToList()
+                .SelectMany(dataSource.DeleteAll)
+                .Select(FinishedDeleting.Transition());
+
+        protected abstract bool Predicate(TDatabaseInterface predicate);
+
+        protected abstract IObservable<TInterface> SuitableForDeletionOrNull(TInterface entity);
```
simonrozsival: This should return `IObservable<bool>`. Then you can use `.WhereAsync(SuitableForDeletion)`.

Toggl.Foundation/Sync/States/CleanUp/DeleteEntityState.cs
```diff
@@ -0,0 +1,46 @@
...
+        where TDatabaseInterface : IDatabaseModel
+    {
+        private readonly IDataSource<TInterface, TDatabaseInterface> dataSource;
+        private readonly bool includeInaccessible;
+
+        public StateResult FinishedDeleting { get; } = new StateResult();
+
+        public DeleteEntityState(
+            IDataSource<TInterface, TDatabaseInterface> dataSource,
+            bool includeInaccessible)
```
simonrozsival: Do you think there is a use case for this to be `false`? I think it should be always `true`.

Toggl.Foundation/Sync/States/CleanUp/DeleteNonReferencedInaccessibleTagsState.cs
```diff
@@ -1,19 +1,49 @@
...
+        }
+
+        protected override bool Predicate(IDatabaseTag tag)
+            => tag.IsInaccessible;
 
-        protected override bool IsReferenced(IDatabaseTag tag, IDatabaseTimeEntry timeEntry)
+        protected override IObservable<IThreadSafeTag> SuitableForDeletionOrNull(IThreadSafeTag tag)
+            => timeEntriesDataSource.GetAll(timeEntry =>
+                    isReferenced(tag, timeEntry) &&
+                    !isSuitableForDeletion(timeEntry),
```
simonrozsival: We don't need to check this condition. Even if the TE which references it is suitable for deletion, we can't delete the tag. We simply need 0 time entries in our database to reference it. Also when everything works correctly, all the TEs which were suitable for deletion will be already deleted because this sync state will follow the one which deletes them.

Toggl.Foundation/Sync/States/CleanUp/DeleteEntityState.cs
```diff
@@ -0,0 +1,46 @@
...
+using Toggl.Multivac.Extensions;
+using Toggl.Multivac.Models;
+using Toggl.PrimeRadiant;
+using Toggl.PrimeRadiant.Models;
+
+namespace Toggl.Foundation.Sync.States.CleanUp
+{
+    public abstract class DeleteEntityState<TInterface, TDatabaseInterface>
+        : ISyncState
+        where TInterface : class, TDatabaseInterface, IThreadSafeModel
```
simonrozsival: We can say that `where TInterface : ..., IPotentiallyInaccessible`, because the specs don't say we should clean any other entities which aren't potentially inaccessible (except for workspace features, but we can deal with those in a different way). The `Predicate` abstract function is then not needed.

juanlaube: Again, I was trying to be as generic as possible, that's why I added the predicate function. I 🔥ed that and am filtering by `IsInaccessible` now in the base state.

Toggl.Foundation.Tests/Sync/States/CleanUp/DeleteNonReferencedInaccessibleTagsStateTests.cs
```diff
@@ -0,0 +1,114 @@
...
+        private readonly IDataSource<IThreadSafeTag, IDatabaseTag> tagsDataSource = Substitute.For<IDataSource<IThreadSafeTag, IDatabaseTag>>();
+
+        public DeleteNonReferencedInaccessibleTagsStateTests()
+        {
+            state = new DeleteNonReferencedInaccessibleTagsState(tagsDataSource, timeEntriesDataSource);
+        }
+
+        [Fact, LogIfTooSlow]
+        public async Task
+            DeletesUnreferencedTags()
```
simonrozsival: `\n`? 🤔

simonrozsival: Also give this a more descriptive name like: `DeleteUnreferencedTagsInInaccessibleWorkspace`

juanlaube: Done and done

PR #3505, opened 2018-10-19
Delete inaccessible tasks

### LINKED ISSUE #3362: WS removal: remove unneeded tasks from DB
When we detect the user losing access to a workspace, and after we removed synced TEs we want to remove all tasks that are no longer referred to by any time entry.

juanlaube:
## What's this?
This is a new sync state to delete inaccessible, unreferenced tasks. The state is not connected yet to the state machine.

### Relationships

Closes #3362 

## Why do we want this?
👻 🔥 

## How is it done?
Very similar to the state to delete inaccessible, unreferenced tags 😛 
For the tests, I extracted some common methods to a base class.

### Side effects
None.

## Review hints
Should be a simple review.

## :squid: Permissions


Toggl.Foundation.Tests/Sync/States/CleanUp/DeleteNonReferencedInaccessibleTagsStateTests.cs
```diff
@@ -32,26 +31,26 @@ public DeleteNonReferencedInaccessibleTagsStateTests()
...
+
+            var tag1 = GetTag(1001, accessibleWorkspace);
+            var tag2 = GetTag(1002, accessibleWorkspace);
+            var tag3 = GetTag(1003, accessibleWorkspace);
+            var tag4 = GetTag(2001, inaccessibleWorkspace);
+            var tag5 = GetTag(2002, inaccessibleWorkspace);
+            var tag6 = GetTag(2003, inaccessibleWorkspace);
+            var tag7 = GetTag(2004, inaccessibleWorkspace);
+
+            var te1 = GetTimeEntry(10001, accessibleWorkspace, SyncStatus.SyncNeeded, new[] { tag1 });
```
simonrozsival: Why did you change the sync status of the time entry?

juanlaube: 🙈 ©️ 🍝

juanlaube: Notice that the sync status for the time entries is irrelevant in this case.

simonrozsival: Yes, I see. IMO that is one more reason to add different combinations though.

PR #3506, opened 2018-10-19
Only delete synced unreferenced entities

juanlaube:
## What's this?
🙈 unreferenced entities to be deleted must be synced.

### Relationships

Closes nothing, refs #2633 

## Why do we want this?
We cannot delete unsynced entities, even if they are not referenced.

## How is it done?
Adding another condition to the predicate in `dataSource.GetAll`.

### Side effects
None.

## :squid: Permissions
Whenever, anyone.


Toggl.Foundation/Sync/States/CleanUp/DeleteInaccessibleEntityState.cs
```diff
@@ -25,13 +26,16 @@ public abstract class DeleteInaccessibleEntityState<TInterface, TDatabaseInterfa
...
+            => dataSource.GetAll(predicate, includeInaccessibleEntities: true)
                 .SelectMany(CommonFunctions.Identity)
                 .WhereAsync(SuitableForDeletion)
                 .ToList()
                 .SelectMany(dataSource.DeleteAll)
                 .Select(FinishedDeleting.Transition());
 
         protected abstract IObservable<bool> SuitableForDeletion(TInterface entity);
+
+        private bool predicate(TDatabaseInterface entity)
```
simonrozsival: `predicate` is not the best name here - what about `candidatesForDeletion` - it would read `get all candidates for deletion including inaccessible entities`.

PR #3507, opened 2018-10-19
🚧🐛Refresh UI after losing access to WS

### LINKED ISSUE #3467: Reload time entries log when user loses access to a workspace
When user loses access to a workspace during pull-sync, we must immediately update the UI to hide all the time entries which should not be visible (as defined in #3431).

I suggest simply reloading all the data. Losing access to a workspace is a rare event, we should not try to fiddle with any animations.

sofokli1:
## What's this?
It reload all the main log when detecting WS becomes inactive

### Relationships
Closes #3467

## Why do we want this?
We want to hide data the user has no more access

## How is it done?
Listen for WS Updates in main VM and call a reload to the `TimeEvtriesViewModel` to refetch the data.
Also found some bugs with tableview handling in case for reload and not individual changes of cells and sections, it took me a while to identify what was the issue due to the steps in `Review hints` that were needed to reproduce the condition but at least it is solved now.

### Why not another way?
This was the only way that made sense.

### Side effects
None

## Review hints
Have a WS with some entries in the web and synced with the app
- go to web Workspaces and create a new one
- make the above WS as default
- delete the old WS with entries
- sync the app

## :squid: Permissions
:squid:


sofokli1:
Needs PR #3513 merged into develop and updated through parent branch and this branch in order to work properly.

sofokli1:
Needs PR #3513 merged into develop and updated through parent branch and this branch in order to work properly.

Odrakir:
Hey, I'm testing this and I get a crash. These are my steps:
- Create a new workspace in web, set as default
- Go to the app, start an entry in the new workspace, stop the entry
- Go to the web, delete the workspace
- Go back to the app, sync -> CRASH

PR #3508, opened 2018-10-19
Add helpers to create mock entities

juanlaube:
## What's this?
Just a refactor of some helper methods I added to create mock entities, now those are static methods of their own class.

### Relationships

Closes nothing.

## Why do we want this?
Because this is better to read, maintain and reuse.

### Side effects
None.

## :squid: Permissions
Anyone, whenever.


Toggl.Foundation.Tests/Mocks/MockProject.cs
```diff
@@ -59,5 +59,17 @@ public sealed class MockProject : IThreadSafeProject
         public IEnumerable<IThreadSafeTask> Tasks { get; set; }
 
         public bool IsInaccessible => Workspace.IsInaccessible;
+
+        public static MockProject Create(
```
simonrozsival: I am still not sure why you don't want this  to be a constructor, but I am probably OK with it.

simonrozsival: I am still not sure why you don't want this  to be a constructor, but I am probably OK with it.

juanlaube: I don't have a strong preference about this, any particular advantage of a constructor vs a static method?

PR #3509, opened 2018-10-19
Add state to delete unneeded projects

### LINKED ISSUE #2634: WS removal: remove unneeded projects from DB
When we detect the user losing access to a workspace, and after we removed synced TEs we want to remove all projects that are no longer referred to by any TE or task.

This same state can possible reused to do the same for tags/clients and even the workspace itself?

juanlaube:
## What's this?
This is a new sync state to delete inaccessible, unreferenced projects. The state is not connected yet to the state machine.

### Relationships

Closes #2634 

## Why do we want this?
👻 🔥 

## How is it done?
Very similar to the other states to delete inaccessible, unreferenced things 😛
The only difference in this case is projects could be referenced by time entries *and* tags.

### Side effects
None.

## Review hints
Should be a simple review.

## :squid: Permissions


juanlaube:
Notice that the methods to mock data may change after #3508 is merged.

Issue #3352, opened 2018-10-08
RxFactor low hanging fruits


heytherewill:
### who dis?
As discussed during the Rome discussions, we need to start adding Rxfactors to our backlog. Not only will this improve readability and reduce complexity (as it has done in many cases) but it will also make Droid more reliable by removing the raw string bindings.

This issue can be closed once all checkboxes are marked.

### Low hanging fruits
These are simple views that can be refactored without much hassle. One ViewModel per Pull request and we'll have simple 3 file PRs that won't have any side effects! Note that in many situations, Rxfactors might simply mean making things immutable or using `RxAction`. We don't necessarily need `IObservable`s on every VM (though this is most likely the case for most).

- [x] AboutViewModel
- [x] BrowserViewModel
- [x] ForgotPasswordViewModel
- [x] NoWorkspaceViewModel
- [ ] OnboardingViewModel
- [x] OutdatedAppViewModel
- [x] SendFeedbackViewModel
- [x] SuggestionsViewModel
- [x] TermsOfServiceViewModel
- [ ] TimeEntryViewModel
- [x] TokenResetViewModel

### The Selectables
All `Selectable*ViewModel`s need to change so they only have immutable properties. This will make it much simpler to reason about them. These can be solved in bulk as long as the change set is reasonable.
- [x] SelectBeginningOfWeekViewModel
- [x] SelectClientViewModel 
- [x] SelectColorViewModel
- [x] SelectCountryViewModel 
- [ ] SelectDateFormatViewModel
- [ ] SelectDateTimeViewModel
- [ ] SelectDurationFormatViewModel
- [ ] SelectProjectViewModel
- [ ] SelectTagsViewModel
- [ ] SelectTimeViewModel
- [ ] SelectWorkspaceViewModel

heytherewill:
Superseded by specific issues

PR #3511, opened 2018-10-21
RX Refactor: Token Reset ViewModel

kornelijepetak:
## What's this?
RXFactors the TokenResetViewModel

### Relationships
Deals with part of #3352 

## Why do we want this?
To make our lives easier.

## How is it done?
By, well... er... refactoring?

### Side effects
None expected.
(Currently I have some issues with the tests)

## Review hints
Start with viewmodel, move to viewcontroller/activity, then see tests.

## :squid: Permissions
Just me. (after the tests are dealt with)


Toggl.Daneel/Extensions/Reactive/UITextFieldExtensions.cs
```diff
@@ -21,5 +23,25 @@ public static IObservable<string> Text(this IReactive<UITextField> reactive)
 
         public static Action<string> TextObserver(this IReactive<UITextField> reactive)
            => text => reactive.Base.Text = text;
+
+        public static IObservable<Unit> ShouldReturn(this IReactive<UITextField> reactive, bool shouldResignFirstResponder = true)
+        {
+            return Observable.Create<Unit>(observer =>
+            {
+                reactive.Base.ShouldReturn = textField =>
```
khoi: should we keep the traditional delegates work by 
```suggestion
                reactive.Base.ShouldReturn += textField =>
```

Toggl.Daneel/Extensions/Reactive/UITextFieldExtensions.cs
```diff
@@ -21,5 +23,25 @@ public static IObservable<string> Text(this IReactive<UITextField> reactive)
...
+                    if(shouldResignFirstResponder)
+                        textField.ResignFirstResponder();
+
+                    observer.OnNext(Unit.Default);
+                    return true;
+                };
+
+                return Disposable.Create(() =>
+                {
+                    reactive.Base.ShouldReturn = null;
```
khoi: also here I think it should only remove the `ShouldReturn` block above

Toggl.Daneel/ViewControllers/TokenResetViewController.cs
```diff
@@ -34,60 +36,24 @@ public override void ViewDidLoad()
 
             prepareViews();
 
-            var invertedBoolConverter = new BoolToConstantValueConverter<bool>(false, true);
+            this.Bind(ViewModel.Error, ErrorLabel.Rx().Text());
+            this.Bind(ViewModel.Email.SelectToString(), EmailLabel.Rx().Text());
+            this.Bind(ViewModel.Password.SelectToString(), PasswordTextField.Rx().TextObserver());
+            this.Bind(PasswordTextField.Rx().Text().Select(Password.From), ViewModel.SetPassword);
```
khoi: I think parsing text to `Password` is not the responsibility of the VC. VM can take in the text and change it how it wants

Toggl.Multivac/Extensions/ReactiveExtensions.cs
```diff
@@ -130,15 +130,24 @@ public static void DisposedBy(this IDisposable disposable, CompositeDisposable d
...
+
         public static IObservable<Unit> ToUnitObservable<T>(this Task<T> task)
             => Observable
             .FromAsync(async () => await task)
             .SelectUnit();
+
+        public static IObservable<string> SelectToString<T>(this IObservable<T> observable)
+            => observable.Select(item => item.ToString());
+
+        public static IObservable<bool> SelectInverted(this IObservable<bool> observable)
```
khoi: we have an extension for this already at line 95 same file
```
public static IObservable<bool> Invert(this IObservable<bool> observable) => observable.Select(b => !b);
```

khoi:
Tests are failing though

Toggl.Foundation.MvvmCross/ViewModels/TokenResetViewModel.cs
```diff
@@ -71,10 +80,42 @@ IAnalyticsService analyticsService
...
             this.onboardingStorage = onboardingStorage;
             this.analyticsService = analyticsService;
-
-            DoneCommand = new MvxCommand(done);
-            SignOutCommand = new MvxAsyncCommand(signout);
-            TogglePasswordVisibilityCommand = new MvxCommand(togglePasswordVisibility);
+            this.schedulerProvider = schedulerProvider;
+
+            Error = errorSubject
+                .AsDriver(schedulerProvider);
```
simonrozsival: No `onErrorJustReturn`? I would recommend adding it explicitly, not using `default(bool) == false`, which even doesn't make sense in this case 🤔

Toggl.Foundation.MvvmCross/ViewModels/TokenResetViewModel.cs
```diff
@@ -83,53 +124,65 @@ public override async Task Initialize()
...
-            IsLoading = false;
-            Error = ex is ForbiddenException ? Resources.IncorrectPassword : Resources.GenericLoginError;
+            var error = ex is ForbiddenException
+                ? Resources.IncorrectPassword
+                : Resources.GenericLoginError;
+
+            errorSubject.OnNext(error);
         }
+
+        private IObservable<bool> askToLogOut()
```
simonrozsival: 🤔 I think the name should more describe what the return value is, especially when it is used directly in a condition - I would suggest `userWantsToLogOutAnyway`, `userConfirmsLoggingOut`, or something like that.

Toggl.Foundation.MvvmCross/ViewModels/TokenResetViewModel.cs
kornelijepetak: I am not following you. `OnError` is not even called on `errorSubject` (meaning it can't do `onErrorJustReturn`).

Toggl.Foundation.MvvmCross/ViewModels/TokenResetViewModel.cs
kornelijepetak: I (dis)agree. The method name should deal more with what is being done than what is being returned. In ideal case those two would be as close as possible.

So, what I've done is introduced the result in a variable so that the flow is clear:
```
if (!needsSync)
    return;

var userConfirmedLoggingOut = await askToLogOut();
if (!userConfirmedLoggingOut)
    return;
```

kornelijepetak:
> Fixed the test

Or not... after merging develop, some tests fail again.

PR #3512, opened 2018-10-22
Implement 24 hours mode switch on Giskard

### LINKED ISSUE #3475: 12/24h time format switch is missing
## 🐛 Bug Report
There is no 12/24h time format switch on Android in the settings view. The logic is already implemented and it works well on iOS, there is no reason not to have it on Android as well.

## Context
Build/Branch: `1.5.2`
Phone OS: `Android`

## Expected behavior
Users can change the 12/24h format directly in the app.

kornelijepetak:
## What's this?
Implements UI for selecting 12/24 hours modes in Android

### Relationships
Closes #3475 

## Why do we want this?
So that users can set up the time format in the app.

## How is it done?
By introducing some UI, binding the UI to VM operations, by committing, then pushing, reviewing and merging. But I'm sure you knew that.

### Why not another way?
Wait, what other way?

### Side effects
There should not be any. And if there are, I have failed miserably.

## Review hints
It's a rather short PR, so I don't think you'll need much help.

## :squid: Permissions
Me, myself and Irene.
Wait, who dat!?


Toggl.Foundation.MvvmCross/ViewModels/Settings/SettingsViewModel.cs
```diff
@@ -98,6 +99,8 @@ public sealed class SettingsViewModel : MvxViewModel
 
         public UIAction OpenNotificationSettingsAction { get; }
 
+        public UIAction ToggleTwentyHourSettings { get; }
```
simonrozsival: `TwentyHour`, not `TwentyFourHour`? Is that intentionally? If yes, why?

Toggl.Giskard/Activities/SettingsActivity.ViewBinds.cs
```diff
@@ -12,6 +12,7 @@ public partial class SettingsActivity
         private View logoutView;
         private View feedbackView;
         private View manualModeView;
+        private View is24hoursModeView;
```
simonrozsival: This is Android code, so I don't really know your internal conventions, but `24` in the variable name seems off - don't be lazy and type  `TwentyFour` a few times 😄

Toggl.Foundation.MvvmCross/ViewModels/Settings/SettingsViewModel.cs
kornelijepetak: Of course it wasn't intentional. I am guessing the word `Hour` looked to me like `Four` so I didn't notice I'm missing a word.

Toggl.Giskard/Activities/SettingsActivity.ViewBinds.cs
kornelijepetak: 24 is used all over the code, in Giskard, in Foundation, in tests. There are 38 instances of `24hour`. Not sure what you want changed. Only things in this PR, or every instance in the codebase. Cause even if I change it here, it will still remain a lot of them in the codebase...

simonrozsival: I see, it seems to be present in almost exclusively Android code (`SelectDateTimeViewModel` is also used just on Android, isn't it?) so I suppose it is OK to keep it here, even though I personally don't like it.

Issue #3502, opened 2018-10-19
Sectioned list is not updated correctly


simonrozsival:
### 👥 Impact
This is the most frequently occurring bug in 1.12.1

### 🔗 Context
Probably related to #3287 

### 🛰 Link
- https://appcenter.ms/orgs/Toggl/apps/Toggl-iOS/crashes/groups/2703bae32546d5893b514c77b10f73ba85f1018c?appVersion=1.12.1
- https://appcenter.ms/orgs/Toggl/apps/Toggl-iOS/crashes/groups/0a8216ac8d2ef3aff7ea003741de9dfb1c7e4d9c?appVersion=1.12.1
- https://appcenter.ms/orgs/Toggl/apps/Toggl-iOS/crashes/groups/194d4943c4e526d36c924b867a1bc6d22f52dc1b?appVersion=1.12.1
- https://appcenter.ms/orgs/Toggl/apps/Toggl-iOS/crashes/groups/168ef42743e6d2ae4cc45df5b50172e2186e924d?appVersion=1.12.1
- https://appcenter.ms/orgs/Toggl/apps/Toggl-iOS/crashes/groups/68a0435bd2fa28431b9ae4ff79fe825e346867ca?appVersion=1.12.1
- https://appcenter.ms/orgs/Toggl/apps/Toggl-iOS/crashes/groups/cb53e518fe6131cf8b037d79b8a752177ddf47e9?appVersion=1.12.1
- https://appcenter.ms/orgs/Toggl/apps/Toggl-iOS/crashes/groups/2703bae32546d5893b514c77b10f73ba85f1018c?appVersion=1.12.1

### 📝  Exception & Stack Trace
```cs
ReactiveTableViewBinder`2[TModel,TCell].handleCollectionChange (Toggl.Foundation.MvvmCross.Collections.Changes.ICollectionChange change)
```

Different messages - some examples:
```
Invalid update: invalid number of sections. The number of sections contained in the table view after the update (41) must be equal to the number of sections contained in the table view before the update (43), plus or minus the number of sections inserted or deleted (1 inserted, 0 deleted).

Invalid update: invalid number of rows in section 0. The number of rows contained in an existing section after the update (11) must be equal to the number of rows contained in that section before the update (8), plus or minus the number of rows inserted or deleted from that section (0 inserted, 0 deleted) and plus or minus the number of rows moved into or out of that section (0 moved in, 0 moved out).

[...]

Issue #3501, opened 2018-10-19
Reloading the main log (on midnight, significant time change) causes the app to crash


simonrozsival:
## 💥 Crash Report
From AppCenter sometimes the 

### 👥 Impact
The issues in AppCenter aren't grouped correctly. It is hard to say what the actual impact is.

### 🔗 Context
Happens on iOS 1.12.1

### 🛰 Link
There are several similar crashes which aren't grouped correctly:
https://appcenter.ms/orgs/Toggl/apps/Toggl-iOS/crashes/groups/53a33348720d0058d9053d6cc2f233d086acea59?appVersion=1.12
https://appcenter.ms/orgs/Toggl/apps/Toggl-iOS/crashes/groups/776193aea88c128dfc5880a1528be9aee8466fc1?appVersion=1.12

... and others. Just check AppCenter.

### 📝  Exception & Stack Trace
```cs
MainViewController.reload ()
```
[...]

PR #3513, opened 2018-10-22
🚧 Fix updating of TEs in the tableView crash

sofokli1:
### Relationships
It should help with #3502 and #3501 but can't say for sure.

## Why do we want this?
Less crashes

## How is it done?
The call to modify the table view was happening before actually changing the data thus ending in a state where it was looking for items that are not available anymore. Also the `BeginUpdates()` and `EndUpdates()` are not needed when reloading the tableview in between them (app 💥).

### Why not another way?
This is the way to paradise

### Side effects
Only good things

## Review hints
Hard to test. This was cherry picked from an other PR. It works and if you see the code it makes sense. If you really want to test this, you need to go to PR #3507 cherry pick this commit there and follow the steps in that PR. (try once without cherry picking and one after cherry picking to see the difference.

## :squid: Permissions
🦑


Toggl.Daneel/Extensions/Binders/ReactiveTableViewBinder.cs
```diff
@@ -39,15 +39,21 @@ private void handleCollectionChange(ICollectionChange change)
         {
             lock (animationLock)
             {
-                tableView.BeginUpdates();
+                if (!(change is ReloadCollectionChange))
```
simonrozsival: In this case let's use `if (change is ReloadCollectionChange) { /* ... */ return; }`. These two ifs seem weird. Do you know what I mean by that? I'm not sure if I'm describing it clearly.

PR #3515, opened 2018-10-22
Use DialogService.Select for choosing beginning of week

kornelijepetak:
## What's this?
Removes all sorts of stuff related to selecting a BeginningOfWeek 

### Relationships
Deals with part of #3352 

## Why do we want this?
Because no code is the best code.

## How is it done?
By using `DialogService.Select` instead of full fledged custom UI that does the same thing.

## Review hints
Most of the things are just 🔥 . The main thing to note are the changes to the `SettingsViewModel`.

## :squid: Permissions
If you're happy and you know it.


Toggl.Foundation.MvvmCross/ViewModels/Settings/SettingsViewModel.cs
```diff
@@ -331,8 +331,19 @@ public async Task SelectDateFormat()
...
+                .Cast<BeginningOfWeek>()
+                .ToList();
+
+            var selectedIndex = beginningOfWeekCollection.IndexOf(currentUser.BeginningOfWeek);
+
+            var selectItems = beginningOfWeekCollection
+                .Select(beginningOfWeek => (beginningOfWeek.ToString(), beginningOfWeek))
+                .ToArray();
+
+            var newBeginningOfWeek = await dialogService
```
simonrozsival: If I tap on `Cancel` the beginning of week is set  to "Sunday" (`default(BeginningOfWeek) == 0`?).

Odrakir:
As discussed in Slack this is not the best solution UX wise (at least for iOS I'm sure). Maybe we should replace current modals with pushed views, but for anything else more "custom" I would like to see what a designer proposes.

kornelijepetak:
Ok, I will change just the Android version because basically, it looks identical.

kornelijepetak:
And besides, the iOS version does not have all the features (current selection for example).

kornelijepetak:
Since this way of doing things requires localization in the viewmodel, I will postpone this PR until localization is in place. Or indefinitely.

Also, because this PR does not work for iOS, the benefits are too small so I am closing it after all.

PR #3516, opened 2018-10-22
:wrench: Refactor selectcolor viewmodel

kornelijepetak:
## What's this?
Refactor SelectableColorViewModel so it's immutable.

### Relationships
Part of #3352 

## Why do we want this?
To make things safer and nicer.

### Side effects
I dunno. Don't think so.

## :squid: Permissions
Yes.


Toggl.Foundation.MvvmCross/ViewModels/SelectColorViewModel.cs
```diff
@@ -17,7 +18,7 @@ public class SelectColorViewModel : MvxViewModel<ColorParameters, MvxColor>
         private readonly IMvxNavigationService navigationService;
 
         private MvxColor defaultColor;
-        private readonly SelectableColorViewModel customColor =
+        private SelectableColorViewModel customColor =
```
willsb: I'd rename this property to make things a tad easier to reason about

PR #3520, opened 2018-10-22
Sanic tapes: Add trace for MainActivity onCreate

### LINKED ISSUE #3519: Trace MainActivity OnCreate
This is a simple trace to check the `OnCreate` method of the MainActivity.
It seems to be slow, based on systrace.

If it is so, we might wanna add a loading view there while the the data is being loaded and the main recycler adapter is creating/inflating its viewholders.

daividssilverio:
## What's this?
This is a trace for the MainActivity:OnCreate lifecycle method.
This is the time the app takes to inflate its views right before the user can actually see anything.
If this is too slow, we might wanna move some initializations to `OnResume` with some extra care and a nice loading view.

### Relationships
Part of #3400
Closes #3519

## Why do we want this?
Because the app is slow to start on low end devices. Let's measure to know how slow it is.
Improve and measure it again to see how faster it gets.

## How is it done?
It's a simple stopwatch for the OnCreate method.

### Why not another way?
I didn't want to add it to the MainViewModel because this is more related to the slow inflation and all the RecyclerView view holders inflation under under it as well... I think this doesn't have to be shared.

### Side effects
It will slow down a bit the OnCreate method. But just a tiny bit.

## Review hints
None.

## :squid: Permissions
:sanic: 🦑 it.


daividssilverio:
Tests.Unit.Split: https://app.bitrise.io/build/7ee8f84df7dab463
iOS builds: https://app.bitrise.io/build/e9da3e7bdce8463e
Android builds: https://app.bitrise.io/build/fd9729655b4b29ac

PR #3522, opened 2018-10-22
Specify user ID for bar chart  reports

### LINKED ISSUE #3521: Bar charts in Daneel reports return team-based data rather than user-based data.
## 🐛 Bug Report
Bar charts in Daneel reports return team-based data rather than user-based data. 

## Context

Build/Branch: Daneel 1.12.1
Phone OS: iOS 12.0.1
Phone model: iPhone 7

## Expected behavior

Bar charts in reports should return user based data just like the total number of hours and the pie chart do. 

## To reproduce

Open reports of a workspace where you've tracked time together with a team. Compare data with web reports. 

Reported by user in conversations/19116829644, conversation also includes screenshots for reference.

simonrozsival:
## What's this?
This PR fixes the way we fetch data from the reports API so that it only includes data for the current user and not the rest of his/her team.

### Relationships

Closes #3521 

## Why do we want this?
We want our customers to see accurate data.

## How is it done?
Using the `user_ids` filter of the endpoint.

### Why not another way?
This was recommended to me by the backend team and it requires very little changes in our code. There is probably not even a different way of doing this.

### Side effects
No side effects.

## Review hints
Check the data in the app on `develop`, on this branch, and in the web app. I suggest looking into the `Toggl` workspace for the last week - in the app store version you should see more activity than in the web app. The web app should show the same bar charts as this branch.

## :squid: Permissions
🦑 as soon as GitHub is stable


Odrakir:
Why is it that I see more time in the web than in the app (App Store version)? 🤔 

img_ec93681ae0a3-1
captura de pantalla 2018-10-23 a las 10 29 22

simonrozsival:
It seems this is a bug in the web app (see https://github.com/toggl/webapp/issues/8245).

PR #3530, opened 2018-10-23
Rx Refactor SuggestionsViewModel

khoi:
## What's this?
Rx Refactor SuggestionsViewModel

### Relationships
Part of #3352

## Why do we want this?
To have reactive code.

## How is it done?
By getting rid of Mvx binding and replace it with Rx.

### Why not another way?
One could use `ObservableGroupedOrderedCollection`. However, since the suggestion list is immutable, adopting that class is kinda overkill. so I'm going for `IObservable<Suggestion[]>`

### Side effects
None

## Review hints
Don't go commit by commit, review the whole thing at once. Follow the VM -> VC (Activity) flow

## :squid: Permissions
🐙 after getting one approval from each platform.


khoi:
WIP as the test is failing

Toggl.Foundation.Tests/MvvmCross/ViewModels/BaseViewModelTests.cs
```diff
@@ -34,7 +36,9 @@ public abstract class BaseViewModelTests<TViewModel> : BaseMvvmCrossTests
         protected IRemoteConfigService RemoteConfigService { get; } = Substitute.For<IRemoteConfigService>();
         protected IPasswordManagerService PasswordManagerService { get; } = Substitute.For<IPasswordManagerService>();
         protected IErrorHandlingService ErrorHandlingService { get; } = Substitute.For<IErrorHandlingService>();
-        protected ISuggestionProviderContainer SuggestionProviderContainer { get; } = Substitute.For<ISuggestionProviderContainer>();
+
+        protected ISuggestionProviderContainer SuggestionProviderContainer =
```
sofokli1: why remove the`{ get; }` only for this one?

Toggl.Daneel/Views/Suggestions/SuggestionsView.cs
```diff
@@ -118,14 +69,39 @@ private void prepareSuggestionViews()
...
                + index * distanceBetweenSuggestions
                + index * suggestionHeight;
+
+        private float heightForSuggestionCount(int count)
+        {
+            if (count == 0)
+            {
+                return 0;
+            }
+            return count * (suggestionHeight + distanceBetweenSuggestions) + distanceAbowTitleLabel
```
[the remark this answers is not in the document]
khoi: I copied this calculation from the previous implementation. I believe the layout is currently like this. 

For instance: count = 2

`suggestionHeight` | `distanceBetween` | `suggestionHeight` | `distanceBetween`

if we do `count - 1` * `distanceBetween` the shadow would be cut off

screen shot 2018-10-25 at 18 30 05

Toggl.Foundation.MvvmCross/ViewModels/SuggestionsViewModel.cs
```diff
@@ -16,92 +19,63 @@ namespace Toggl.Foundation.MvvmCross.ViewModels
...
-        {
-            Suggestions.Add(suggestion);
-
-            RaisePropertyChanged();
+                .ToArray()
+                .AsDriver(onErrorJustReturn: new Suggestion[0], schedulerProvider: schedulerProvider);
+
+            IsEmpty = Suggestions
+                .Select(suggestions => suggestions.Length)
+                .Select(c => c == 0)
```
Odrakir: I think this is better as just one `Select`
```suggestion
                .Select(suggestions => suggestions.Length == 0)
```

PR #3532, opened 2018-10-23
🔍Track inaccessible data before and after clean up

### LINKED ISSUE #3442: WS removal: track how many ghost entities are left in the database
During the clean-up process (before the 👻 WSs are deleted!) track how many of these entities related to some ghost WS are kept in the database:

- [x] time entries
- [x] tags
- [x] tasks
- [x] projects
- [x] clients
- [x] workspaces (after the 👻 workspaces are pruned)

We could:
1. put analytics into every clean-up state which deals with one type of an entity
2. create a purely analytics state for every entity type (let's not do that tough)
3. create one state which submits anaytics for everything but workspaces which will be executed right before we try to clean-up workspaces + a separate state which will be used after the workspaces are pruned

I think number 3 is the best solution in this case. Please comment if you see some other solutions which might be suitable.

*Important thing:* let's not track the numbers for the time entries, ..., clients if there are no ghost workspaces. On the other hand, let's always track the number of ghost workspaces in the database after their clean-up.

sofokli1:
## What's this?
Adds 2 sync states for before and after clean up to track inaccessible data

### Relationships
Closes #3442

## Why do we want this?
To know what is going on with inaccessible data

## How is it done?
By using 2 sync states

### Why not another way?
It was agreed this way

### Side effects
None

## Review hints
Just look at the code and tests

## :squid: Permissions
:squid:


Toggl.Foundation/Sync/States/CleanUp/TrackInaccessibleDataAfterCleanUpState.cs
```diff
@@ -0,0 +1,80 @@
...
+
+        private IObservable<Unit> trackDataIfNeeded(int numberOfInaccessibleWorkspaces)
+        {
+            if (numberOfInaccessibleWorkspaces == 0)
+            {
+                return Observable.Return(Unit.Default);
+            }
+
+            var tagsObservable = dataSource.Tags
+                .GetAll(tag => tag.IsInaccessible && tag.Workspace.IsInaccessible, includeInaccessibleEntities: true)
```
juanlaube: `someEntity.Workspace.IsInaccessible` is not needed here since `someEntityIsInaccessible` is the same thing. This is valid for time entries, tags, tasks, projects and clients.

Toggl.Foundation/Toggl.Foundation.csproj
```diff
@@ -13,6 +13,7 @@
   </PropertyGroup>
   <ItemGroup>
     <Compile Remove="Analytics\ComplexEvents\StartTimeEntryEvent.cs" />
+    <Compile Remove="Sync\States\CleanUp\TrackInaccessibleDataBeforeCleanUpState.cs" />
```
juanlaube: 🔥

Toggl.Foundation/Sync/States/CleanUp/TrackInaccessibleDataAfterCleanUpState.cs
```diff
@@ -0,0 +1,80 @@
...
+        public TrackInaccessibleDataAfterCleanUpState(ITogglDataSource dataSource, IAnalyticsService analyticsService)
+        {
+            Ensure.Argument.IsNotNull(dataSource, nameof(dataSource));
+            Ensure.Argument.IsNotNull(analyticsService, nameof(analyticsService));
+
+            this.dataSource = dataSource;
+            this.analyticsService = analyticsService;
+        }
+
+        public IObservable<ITransition> Start(IFetchObservables fetchObservables)
```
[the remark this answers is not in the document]
juanlaube: That's correct, we don't fetch anything, just 🔥 this param!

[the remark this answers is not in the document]
sofokli1: i first started the state as a pull state and later moved it to cleanup and forgot about this part. Will fix

Toggl.Foundation/Sync/States/CleanUp/TrackInaccessibleWorkspacesAfterCleanUpState.cs
```diff
@@ -9,12 +9,12 @@
 
 namespace Toggl.Foundation.Sync.States.CleanUp
 {
-    public sealed class TrackInaccessibleWorkspacesAfterCleanUpState
+    public sealed class TrackInaccessibleWorkspacesAfterCleanUpState : ISyncState
```
simonrozsival: Another thing I didn't notice 🙈 ... 👍

PR #3541, opened 2018-10-23
Firebase.Perf fix

willsb:
## What's this?
This is a work around for the bug in `Xamarin.Firebase.Perf`

### Relationships
This is needed while [this](https://github.com/xamarin/GooglePlayServicesComponents/issues/151) is not fixed

## Why do we want this?
We need metrics for the next release

## How is it done?
Me and @daividssilverio managed to build a working version of the binaries using the steps described in the linked issue. 

I then replaced the existing .dll with the one that works. @daividssilverio has tested and the traces are indeed getting to where we want them to be, so all is well!

### Why not another way?
It'll take too long for a new nuget package to be generated, so this is hack will do. Once the package has been published we can undo this jerry rig

### Side effects
Let's make sure we test this branch again to ensure data is going where it's supposed to go

## Review hints
Be @daividssilverio 
Test in an actual device

## :squid: Permissions
Me and @daividssilverio


daividssilverio:
It worked!

willsb:
Just checked firebase and the data is there!

PR #3546, opened 2018-10-24
Delete inaccessible running time entry

### LINKED ISSUE #3524: WS removal: hide current running time entry
When we detect that user does not have access to a workspace and there is a time entry in that workspace currently running, we must hide it.

If it is in sync, we will delete it anyway, there is no reason to stop it (it would even change it to `SyncNeeded` state), just hide it.

If it is not in sync, we should stop it and the user will then be able to see it in the log, move it to a different workspace, and then adjust the end time.

juanlaube:
## What's this?
Adds a new sync state to delete the running time entry (only if it's in sync) in any inaccessible workspace. The state is connected to the state machine.

### Relationships

Closes #3524

## Why do we want this?
🚫 👻 

## How is it done?
It's a new state that finds and delete the running time entry in an inaccessible workspace. This state runs after persisting all the time entries.

### Side effects
None.

## :squid: Permissions


Toggl.Foundation/TogglSyncManagerFactory.cs
```diff
@@ -178,7 +181,8 @@ public static class TogglSyncManager
             transitions.ConfigureTransition(updateTasksSinceDate.Finished, createProjectPlaceholders);
 
             transitions.ConfigureTransition(createProjectPlaceholders.FinishedPersisting, persistTimeEntries);
-            transitions.ConfigureTransition(persistTimeEntries.FinishedPersisting, updateTimeEntriesSinceDate);
+            transitions.ConfigureTransition(persistTimeEntries.FinishedPersisting, deleteRunningInaccessibleTImeEntry);
```
simonrozsival: I think this should happen right after we discover that a workspace is inaccessible, not after we update time entries. Persisting time entries might stop the running TE because other entry pulled from the API was running. That would make it hard to remove this TE, which was already in sync.

Toggl.Foundation/Sync/States/Pull/DeleteInaccessibleRunningTimeEntryState.cs
```diff
@@ -0,0 +1,46 @@
...
+                .SelectMany(delete)
+                .ToList()
+                .Select(_ => Continue.Transition(fetch));
+
+        private bool inaccessibleSyncedRunningTimeEntry(IDatabaseTimeEntry timeEntry)
+            => timeEntry.IsInaccessible &&
+               timeEntry.SyncStatus == SyncStatus.InSync &&
+               timeEntry.Duration == null;
+
+        private IObservable<Unit> delete(IThreadSafeTimeEntry timeEntry)
```
simonrozsival: I would suggest `deleteIfNeeded` as the name for this function.

Toggl.Foundation.Tests/Sync/States/Pull/DeleteInaccessibleRunningTimeEntryStateTests.cs
```diff
@@ -0,0 +1,137 @@
...
+
+            configureDataSource(new[] { runningTimeEntry, timeEntry1, timeEntry2 });
+
+            await state.Start(fetch).SingleAsync();
+
+            dataSource.Received().Delete(Arg.Is<long>(id => runningTimeEntry.Id == id));
+        }
+
+        [Fact, LogIfTooSlow]
+        public async Task DoesNotDeleteAnyRunningTimeEntryNotInSyncInInaccessibleWorkspaces()
```
simonrozsival: Let's turn this into a `[Theory]` with different sync statuses (sync needed, sync failed, refetching needed).

Toggl.Foundation.Tests/Sync/States/Pull/DeleteInaccessibleRunningTimeEntryStateTests.cs
```diff
@@ -0,0 +1,137 @@
...
+
+        [Fact, LogIfTooSlow]
+        public async Task DoesNotDeleteAnyRunningTimeEntryInSyncInAccessibleWorkspaces()
+        {
+            var fetch = Substitute.For<IFetchObservables>();
+
+            var start = new DateTimeOffset(2018, 10, 23, 15, 42, 00, TimeSpan.Zero);
+            var duration = (long)TimeSpan.FromMinutes(15).TotalSeconds;
+
+            var workspace = new MockWorkspace(1);
```
simonrozsival: Let's put `isInaccessible: false` explicitly here for the sake of better readability! :)

Toggl.Foundation.Tests/Sync/States/Pull/DeleteInaccessibleRunningTimeEntryStateTests.cs
```diff
@@ -0,0 +1,137 @@
...
+
+            configureDataSource(new[] { runningTimeEntry, timeEntry1, timeEntry2 });
+
+            await state.Start(fetch).SingleAsync();
+
+            dataSource.DidNotReceive().Delete(Arg.Any<long>());
+        }
+
+        [Fact, LogIfTooSlow]
+        public async Task DoesNotDeleteAnyRunningTimeEntryNotInSyncInAccessibleWorkspaces()
```
simonrozsival: This could be also a `[Theory]`

PR #3556, opened 2018-10-25
Make datepicker use correct beginning of week

### LINKED ISSUE #3554: Date picker not following user's first day of week setting
## ⛑️ User Support
"I have set first day of the week are Monday. When I change the date of a recording, the calendar starts on Sunday. Can I has a fix, please."

## About the user
None at this time.

App version: Giskard 1.5.2
Phone model: Sony G8441
Platform and OS version: 8.0.0 N_MR1

## Expected behaviour
The date picker's calendar should match what the user has set as their first day of the week for their Toggl profile.

## To Reproduce
1. Set your Toggl profile's first day of the week to Tuesday
2. Go to the app and create a new entry
3. Tap on Start time
4. Tap on date 
5. See calendar still start on Sunday

## Suggestions
This might be an opportunity to create custom date and time pickers that don't rely on native elements / run into language-related placement issues.

kornelijepetak:
## What's this?
This PR makes the giskard calendar show the correct beginning day of the week.

### Relationships
Closes #3554 

## Why do we want this?
Because we ❤️ our users.

## How is it done?
There is a property for that on the date picker.

### Why not another way?
I have not completely followed the RXfactor with this because the related viewmodels are not yet RXFactored. But I've tried to follow the idea as closely as possible so it's easier to RXFactor it in the future.

### Side effects
Should be none.

## Review hints
TogglDroidDatePicker is the starting point. Rest is just the infrastructure to enable correct change detection.

## :squid: Permissions
Anyone related to `magnesium` release.


Toggl.Foundation.MvvmCross/ViewModels/SelectTimeViewModel.cs
```diff
@@ -76,6 +77,8 @@ public TimeSpan Duration
 
         public IObservable<bool> Is24HoursModeObservable { get; private set; }
 
+        public IObservable<BeginningOfWeek> BeginningOfWeek { get; private set; }
```
daividssilverio: Why does this have to an observable? Can this change without the user going to the settings or syncing?

daividssilverio: It seems to me that this observable emits only one value. `Prepare` is not called more than once, is it?

kornelijepetak: Oh, yes. You are right. No need for an observable.

PR #3561, opened 2018-10-25
Add new state to detect gaining access to workspaces

### LINKED ISSUE #3357: Add sync state which detects new workspaces
This state should take `FetchObservables` as its input.

The only responsibility is to diff the list of workspaces it receives from the server and the list of workspaces we have in the database and identify new workspaces (see #2907 for specification of _new workspace_).

This state should have two results:
1. no new workspace was detected - continue with the `FetchObservables` as parameter
2. at least one new workspace was detected - process new workspaces with the list of the `IWorkspace`s from the server

Do not configure the transitions for this new state yet.

juanlaube:
## What's this?
Adds a new state that diffs the fetched list of workspaces with the database and has two outcomes:
1. Continue if no workspace was found
2. Transition to a different state with the list of new workspaces as a parameter.

The new state is not connected yet to the sync graph.

### Relationships

Closes #3357

## Why do we want this?
To detect new workspaces and do things in that case.

## How is it done?
It's a new sync state.

### Side effects
None.

## Review hints
Should be a simple review.

## :squid: Permissions
Anyone, whenever, but only after @simonrozsival ✅


Toggl.Foundation/Sync/States/Pull/DetectGainingAccessToWorkspacesState.cs
```diff
@@ -0,0 +1,54 @@
...
+                        : continueTransition(fetch));
+
+        private IEnumerable<IWorkspace> workspacesWhichAreNotInDatabase(
+            IEnumerable<IWorkspace> fetchedWorkspaces,
+            IEnumerable<IWorkspace> storedWorkspaces)
+            => fetchedWorkspaces
+                .Where(fetched => storedWorkspaces.None(stored => stored.Id == fetched.Id));
+
+        private IObservable<IList<IThreadSafeWorkspace>> allStoredWorkspaces()
+            => dataSource.GetAll(ws => ws.Id > 0, includeInaccessibleEntities: true)
```
simonrozsival: The predicate is confusing - I don't think we need it, it is not possible to match entities from the server to the entities with negative ID.

> A workspace is new if it was just fetched, and if we don't have a non-ghost workspace with the same ID locally.

_non-ghost_ == accessible workspace

=> I think we shouldn't include inaccessible workspaces ("ghosts") in this list. Am I correct?

Toggl.Foundation/Sync/States/Pull/DetectGainingAccessToWorkspacesState.cs
```diff
@@ -0,0 +1,54 @@
...
+        public StateResult<IEnumerable<IWorkspace>> NewWorkspacesDetected { get; } = new StateResult<IEnumerable<IWorkspace>>();
+
+        public DetectGainingAccessToWorkspacesState(IDataSource<IThreadSafeWorkspace, IDatabaseWorkspace> dataSource)
+        {
+            Ensure.Argument.IsNotNull(dataSource, nameof(dataSource));
+            this.dataSource = dataSource;
+        }
+
+        public IObservable<ITransition> Start(IFetchObservables fetch)
+            => Observable.CombineLatest(
```
simonrozsival: Invoke this as an extension method on the `fetch.GetList<IWorkspace>()` please 👍

Toggl.Foundation/Sync/States/Pull/DetectGainingAccessToWorkspacesState.cs
```diff
@@ -0,0 +1,54 @@
...
+                .Select(newWorkspaces =>
+                    newWorkspaces.Any()
+                        ? newWorkspacesDetected(newWorkspaces)
+                        : continueTransition(fetch));
+
+        private IEnumerable<IWorkspace> workspacesWhichAreNotInDatabase(
+            IEnumerable<IWorkspace> fetchedWorkspaces,
+            IEnumerable<IWorkspace> storedWorkspaces)
+            => fetchedWorkspaces
+                .Where(fetched => storedWorkspaces.None(stored => stored.Id == fetched.Id));
```
simonrozsival: This can fit on one line! Also I think you can safely rename it to `fetched` and `stored` - everybody knows this state deals only with workspaces.

Toggl.Foundation/Sync/States/Pull/DetectGainingAccessToWorkspacesState.cs
juanlaube: 🙈 that's correct!

PR #3562, opened 2018-10-25
Add no default workspace view for droid

### LINKED ISSUE #2543: Ask user to pick a default WS when a default one is not assigned
Check #2462

arturs105:
### Relationships

Closes #2543

## Review hints
Be extra attentive as it's been a while, since i've worked on Android.

## :squid: Permissions
Whenever.


Toggl.Giskard/Toggl.Giskard.csproj
```diff
@@ -474,7 +474,15 @@
     <Compile Include="Extensions\Reactive\RecyclerAdapterExtensions.cs" />
     <Compile Include="Extensions\Reactive\SwipeRefreshLayoutExtensions.cs" />
     <Compile Include="Extensions\Reactive\JavaObjectExtensions.cs" />
+<<<<<<< Updated upstream
```
daividssilverio: Forgot some git diffs here.

Toggl.Foundation.MvvmCross/ViewModels/MainViewModel.cs
```diff
@@ -375,6 +375,11 @@ public override void ViewAppearing()
             {
                 navigationService.Navigate<NoWorkspaceViewModel>();
             }
+
+            if (accessRestrictionStorage.HasNoDefaultWorkspace())
```
daividssilverio: Hey, I was testing this and it seems I've induced you to a bad approach.

If you put the app in background, and come back while the dialog is open, it will open another one. If the user selects one workspace, the other dialog will become immediately visible, leading the user to believe something went wrong, or the view isn't working. In the case of the `NoWorkspaceViewModel`, it will let the user create multiple "default workspaces"

This isn't terribly bad because we are talking about some edge case here, but would you mind taking the chance to fix this? Maybe using a simple flag to check if the navigation already happened.

Toggl.Giskard/Resources/layout/SelectDefaultWorkspaceFragment.axml
```diff
@@ -0,0 +1,38 @@
...
+        android:textSize="20sp"
+        android:letterSpacing="0"
+        android:layout_marginTop="24dp"
+        android:layout_marginBottom="12dp"
+        android:layout_marginLeft="24dp"/>
+    <TextView
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:text="@string/SetDefaultWorkspaceDescription"
+        android:background="@android:color/white"
```
daividssilverio: You don't have to set this background color.

Toggl.Giskard/Resources/layout/SelectDefaultWorkspaceFragment.axml
```diff
@@ -0,0 +1,38 @@
...
+<LinearLayout
+    xmlns:android="http://schemas.android.com/apk/res/android"
+    android:orientation="vertical"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent">
+    <TextView
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:text="@string/SetDefaultWorkspaceTitle"
+        android:background="@android:color/white"
```
daividssilverio: You don't have to set this background color.

daividssilverio:
By the way, name of the PR should've been `No Default Workspace`. `No Workspace` is already there.

arturs105:
I added 2 `bool` thingies to make sure the view does not get presented several times. This solves the issue that @daividssilverio pointed out. However this adds another potential problem:
1. User gets into no workspace or no default workspace state
2. The appropriate dialog gets presented
3. User solves the problem with the dialog and it is closed
4. User gets into the same state again without closing the app
5. This time the dialog isn't shown, since the `bool` field is already `true`

I think this could be solved by moving the `bool` fields to `AccessRestictionStorage`.

daividssilverio:
Since you are already fixing it again, I'll wait for your fix so we can :shipit:

Toggl.Foundation.MvvmCross/ViewModels/SelectDefaultWorkspaceViewModel.cs
```diff
@@ -64,6 +71,8 @@ private IObservable<Unit> selectWorkspace(SelectableWorkspaceViewModel workspace
             {
                 await interactorFactory.SetDefaultWorkspace(workspace.WorkspaceId).Execute();
                 await navigationService.Close(this);
+                accessRestrictionStorage.SetSelectDefaultWorkspaceViewShown(false);
```
daividssilverio: 🤔 Could you move the accessRestrictionStorage writes before the `navigation.Close(this)`?
It feels weird to leave it like this.
At least `accessRestrictionStorage.SetNoDefaultWorkspaceStateReached(false);` before navigation, so won't get into weird edge cases.

Toggl.Giskard/Fragments/SelectDefaultWorkspaceFragment.cs
```diff
@@ -0,0 +1,53 @@
...
+
+        public SelectDefaultWorkspaceFragment() { }
+
+        public SelectDefaultWorkspaceFragment(IntPtr javaReference, JniHandleOwnership transfer)
+            : base(javaReference, transfer) { }
+
+        public override View OnCreateView(LayoutInflater inflater, ViewGroup container, Bundle savedInstanceState)
+        {
+            base.OnCreateView(inflater, container, savedInstanceState);
+            var view = this.BindingInflate(Resource.Layout.SelectDefaultWorkspaceFragment, null);
```
daividssilverio: Oh, and you don't need to use BindingInflate, we are not using mvx anymore o/

Toggl.Foundation.MvvmCross/ViewModels/SelectDefaultWorkspaceViewModel.cs
daividssilverio: Whis GitHub didn't hide my comment 🤔

PR #3568, opened 2018-10-26
Add state to persist new workspaces

### LINKED ISSUE #3358: Add sync state which processes new workspaces
Depends on #3356 

The input of this state should be a list of workspaces (`IEnumerable<IWorkspace>`).

This state must reset the `SinceParameterStorage` before it starts persisting the workspaces (if the app crashed after it persisted some workspace, but the since parameters were not reset, some of the entities might not be pulled from the server).

All of these workspaces should be persisted into the database:
- create new records for workspaces which are not in the database yet
- overwrite ghost workspaces in the database (no need for conflict resolution)

Make sure the created/updated workspace is not marked as a ghost.

There should be only one result without any parameter which is always returned.

juanlaube:
## What's this?
Adds a new state that resets the updated since values and persists the new workspaces. The state is not connected yet in the state machine.

### Relationships

Closes #3358

## Why do we want this?
Isn't that obvious? 😄 

## How is it done?
It's a new state, either creates the new workspace if it was not in the database or replaces the old one if it was marked as inaccessible.

### Side effects
None.

## Review hints
~This is still a work in progress because the method to reset the updated since storage is not merged yet.~

## :squid: Permissions
Anyone, whenever, but only ~after this is not a `wip`~ and @simonrozsival ✅.


Toggl.Foundation/Sync/States/Pull/PersistNewWorkspacesState.cs
```diff
@@ -0,0 +1,61 @@
...
+        }
+
+        public IObservable<ITransition> Start(IEnumerable<IWorkspace> workspaces)
+            => Observable.Return(workspaces)
+                .Do(resetSinceUpdating)
+                .SelectMany(CommonFunctions.Identity)
+                .Select(toThreadSafeWorkspace)
+                .SelectMany(createOrUpdate)
+                .ToList()
+                .Select(_ => FinishedPersisting.Transition());
```
simonrozsival: There should be an extension method which takes an instance of an object which it returns - `.Select(FinishedPersisting.Transition());`

Toggl.Foundation/Sync/States/Pull/PersistNewWorkspacesState.cs
```diff
@@ -0,0 +1,61 @@
...
+
+            this.dataSource = dataSource;
+            this.sinceParameterRepository = sinceParameterRepository;
+        }
+
+        public IObservable<ITransition> Start(IEnumerable<IWorkspace> workspaces)
+            => Observable.Return(workspaces)
+                .Do(resetSinceUpdating)
+                .SelectMany(CommonFunctions.Identity)
+                .Select(toThreadSafeWorkspace)
```
simonrozsival: You can put `Workspace.Clean` here directly.

Toggl.Foundation.Tests/Sync/States/Pull/PersistNewWorkspacesStateTests.cs
```diff
@@ -0,0 +1,123 @@
...
+            {
+                new MockWorkspace { Id = 2 },
+                new MockWorkspace { Id = 3 },
+            };
+
+            var state = new PersistNewWorkspacesState(dataSource, sinceParameterRepository);
+            await state.Start(newWorkspaces);
+
+            dataSource.Received()
+                .Create(Arg.Is<IThreadSafeWorkspace>(arg => newWorkspaces.Any(ws => ws.Id == arg.Id)));
```
simonrozsival: I don't think this is correct - you should check that it received 2 calls and one of them was a call with ws 2 and the other with ws 3.

Toggl.Foundation.Tests/Sync/States/Pull/PersistNewWorkspacesStateTests.cs
```diff
@@ -0,0 +1,123 @@
...
+                new MockWorkspace { Id = 2 }
+            };
+
+            var state = new PersistNewWorkspacesState(dataSource, sinceParameterRepository);
+            await state.Start(newWorkspaces);
+
+            dataSource.Received()
+                .Update(Arg.Is<IThreadSafeWorkspace>(arg => arg.Id == 2));
+        }
+
```
simonrozsival: I would like to see another test which mixes `Create` and `Update` here.

Issue #3394, opened 2018-10-11
Create a complex test to verify that losing access to a workspace works correctly


simonrozsival:
Could this be just a Unit test? Or should it rather be a UI test (which will use the storage and the network)?

I think it should be a unit test which will mock the network and use an in-memory database mock. Feel free to have suggestions in the comments.

- [x] create test setup (new workspace, projects, tags, tasks, clients, time entries) with some unsynced and unsyncable data
- [x] initiate sync and wait for it to complete
- [x] verify that the state of the internal storage matches expectations
- [ ] ~find a way how to reliably test the process without the network and real storage~
- [x] implement correctly pushing and pulling data to the server and into the local storage

simonrozsival:
This has been finished and follow-up issues were created.

PR #3573, opened 2018-10-29
Test marking workspaces as inaccessible

simonrozsival:
## What's this?
This test is first of a series of tests for marking workspaces as inaccessible during pull syn
### Relationships

Ref #3394, depends on #3560 

## Why do we want this?
We want our sync to work reliably.

## How is it done?
This test defines a list of workspaces which are stored in the local database, but won't be fetched from the server. These should be marked as inaccessible during pull sync.

The test also checks that all entities which are related to the workspace are marked as synced.

### Why not another way?
I am confident this is a good way of testing things.

### Side effects
No side effects.

## Review hints
Run `sh build.sh --target Tests.Sync` locally.

~Only the commit https://github.com/toggl/mobileapp/pull/3573/commits/f70714d829e4acf4191ed2764277dfdb6a08b345 contains the actual code which should be in this PR. I have to figure out what is wrong with the diff and I will fix the PR later.~ Everything should be fixed now.

## :squid: Permissions
🦑


simonrozsival:
I changed the base branch to `develop`. Nothing else in this PR changed.

simonrozsival:
@juanlaube no, it won't change - this only runs `FullSync`, which is `Pull` + `Push`. Clean-up has to be triggered separately, if that is what you mean.

PR #3574, opened 2018-10-29
Add "Ios" suffix to iOS services

### LINKED ISSUE #3538: Add Ios suffix to Daneel services
### The problem
Solution wide searches yield confusing results because some service implementations have the same name on both iOS and Android

### The solution
Add the `Ios` suffix to all classes inside the `Services` folder. If any class has `Daneel` in its name, change it to Ios (since the project name has nothing to do with the platform).

arturs105:
### Relationships

Closes #3538 

## :squid: Permissions
Whenever.


kornelijepetak:
I don't like that you have remove the `why` section of this, because this is my only question 😄, and I think it's important that this PR states that.

arturs105:
I don't think it makes sense to have the "why" section for PRs that already have an issue, since one of the issue's purposes is to explain why we want the thing. If the issue fails to convince people that we need that change, then the issue should just be closed without a PR.
 
In other words:
>Why did i make this PR?

Because there's an issue for that

>Why do we want this change?

The issue describes that.


Also: i believe that you should at least skim through the issue for the PR you're reviewing to get a better understanding of the thing that's being solved.

kornelijepetak:
Indeed you are right. My bad. I was led astray by the way PR appeared in our internal slack comm. I should've been more thorough.

Thanks for the explanation. You could've also said: `Read the issue, stupid!` 😄

kornelijepetak:
As a punishment, I will now review this PR.

simonrozsival:
I'm not a fan of the issue/PR. I don't think we should adjust our code for how we work with our IDEs. It seems to me that it is good enough to search for e.g. `daneel/analytics` instead of `analyticsios` using fuzzy search.

simonrozsival:
My issue with the lack of docs was resolved. I don't really like this change, so I won't explicitly approve it. But if there are other approvals, feel free to go ahead and don't wait for further reviews from me.

PR #3489, opened 2018-10-18
Losing and gaining access to workspaces

simonrozsival:
## What's this?
This is a long-living branch for the NSync project.

While we merge most of our PRs into the `develop` branch to be always up-to-date and avoid as many merge conflicts in the future, the code won't be actually used until we set-up the sync graph correctly - and we won't merge that into develop until everything is ready. This branch is intended to gradually create the syncing graph and merge it in the end.

### Relationships
Closes #3374 
Closes #3375 

## :squid: Permissions
Do not even think about it at the moment.

PR #3576, opened 2018-10-29
Revert adding sync transitions for detecting losing access to workspace

simonrozsival:
This reverts commit 9d57917768751cf933bc06d0e95f6d58d182cab9 which accidentally got into `develop` without me noticing. It is not ready to be released to the users. Sorry for this complication.

### Relationships

Ref #3489 

## Why do we want this?
Let's not release wip code to the users.

## :squid: Permissions
@dshr should merge this PR into the release branch


juanlaube:
@dshr since this means a new build, you may want to cherry-pick 3b60ec1248a233f6853253d06d842385ddb9b585 too to avoid AppStore Connect asking if we use encryption

PR #3579, opened 2018-10-29
Fix key value storage for dates

### LINKED ISSUE #3367: Setting Eesti as first Preferred Language makes app weird
## 🐛 Bug Report
Estonian users on iOS 12 have reported the app having weird issues and even not launching when Eesti is set as their first Preferred Language through Settings > General > Language & Region.

Estonian users on Android report the app crashing when Eesti is set as the first language under Settings > System > Languages & input > Languages

## Context
Build/Branch: Daneel 1.12 & 1.11, Giskard 1.5.1
Phone OS: iOS 12, Android 8.0.0 (47.1.A.16.20) 
Phone model: Confirmed on iPhone X, iPhone 6s, Sony XPERIA XZ1 (model G8341) 

(I have not been able to reproduce this on my iPhone 8 Plus, still on iOS 11)

## Expected behavior
The app should continue behaving properly regardless of what is selected as the device's Preferred Language.

## To reproduce (on iOS)
1. Update to iOS 12, if you haven't
2. Go to Settings > General > Langauge & Region
3. Select "Eesti" and make it the first item under Preferred Langauge
4. Attempt to start the app and see it freeze/act weirdly

## To reproduce (on Android)
1. Go to Settings > System > Languages & input > Languages
2.  Add Eesti (Estonian) and make it the first language on the klist
3. Install the app 
4. Launch the app 
5. Type in Email and Password, press Log In 
6. Go to Android OS home screen 
7. Attempt to open app again, it hangs 
8. Open Android app switcher 
9. Close app
10. Attempt to open app again 
11. App is no longer logged in 

## Suggestions
Maybe the weirdness and crashing is related to how some virtual keyboard elements are rendered in Eesti?

The current workaround is to ensure that Eesti is not the first item under the Preferred Language list. 

iOS 12:
/conversation/18860146892
/conversation/18818577533

Android:
/conversation/18972554465

### LINKED ISSUE #3432: Specify explicit format for serializing and deserializing DateTimeOffset
Might be related to #3367

- [x] always use `SetDateTimeOffset` and `GetDateTimeOffset` for storing dates in the `IKeyValueStorage`
    - at least one case when we don't do it is in `FoundationExtensions.RevokeNewUserIfNeeded`
- [x] use explicit `DateTimeFormatInfo` format provider when serializing and deserializing dates to string on iOS and Android
- [x] make sure the "migration" won't cause problems

simonrozsival:
## What's this?
This PR refactors the key-value storage to always serialize and deserialize dates using the non-culture-dependent invariant format.

### Relationships

Closes #3367, closes #3432 

## Why do we want this?
See the issue

## How is it done?
This PR adds a shared code for both iOS and Android for serializing and deserializing date times (and while at it, also time spans) so that we don't have any future problems on any of the platforms.

### Side effects
There is a "migration" from the culture-specific formatting to the invariant. If the date is not "parsable" in either the invariant and culture-specific format, the date is reset (in this case - the onboarding might be shown again) but that is probably OK/minor thing, because right now these user's can't open the app at all.

## Review hints
Follow the steps in #3367 on `develop` to reproduce the issue, then switch to this branch. It should not be possible on this branch.

## :squid: Permissions
🦑


Toggl.Daneel/Services/UserDefaultsStorage.cs
```diff
@@ -1,38 +1,38 @@
 ﻿using System;
+using System.Globalization;
 using System.Linq;
+using CloudKit;
```
dshr: Why `System.Globalization` and `CloudKit`? They don't seem to be getting used anywhere.

simonrozsival: Whaaat? :D How did this get here? 🔥 🔥 🔥 ! 😆

PR #3580, opened 2018-10-29
Remove Mvx Inflator from ReactiveActivity

willsb:
## What's this?
This removes the MvxBindingInflator where it's not needed

## Why do we want this?
This inflator is slower

## How is it done?
We have a flag to check whether we want it or not

### Why not another way?
Some views might still need binding inflation

### Side effects
Some fragments might break. We need to test this.

## Review hints
Check things like the views we use to lock the user outside the app. They can potentially break.

## :squid: Permissions
Me.


daividssilverio:
Proof follows:
I've run some tests in the `MainActivity`'s `MainRecyclerAdapter` view holder's inflations.

Main log sections during startup:
Mvx:
image

Regular:
image

daividssilverio:
Main log items during startup:
Mvx:
image

Regular:
image

daividssilverio:
Main log sections after startup (scrolling):
Mvx:
image

Regular:
image

daividssilverio:
Main log items after startup (scrolling):
Mvx:
image

Regular:
image

daividssilverio:
The benefit is real, we just need to make sure dialog fragments will stay working and no `MvxBind` is used in the view that uses `UseMvxContext = false`.

daividssilverio:
Fragments that need their hosts to be have `UseMvxContext = true`:

`SelectBeginningOfWeekFragment` (host: `SettingsActivity`)
`SelectColorFragment` (host: `EditProjectActivity`)
`SelectDateFormatFragment` (host: `SettingsActivity`)
`SelectDateTimeFragment` (host: `StartTimeEntryActivity`)
`SelectTimeFragment` (host: `StartTimeEntryActivity`) -> HARD
`SelectWorkspaceFragment` (hosts: `SettingsActivity`, `EditProjectActivity`)
`TermsOfServiceFragment` (host: `SignupActivity`)

And of course, every activity that is using MvxBind directly in its layout can't have `UseMvxContext = false`

willsb:
Given that `StartTimeEntryActivity` and `EditProjectActivity` do not inherit from `ReactiveActivity`, the only views that need change are `SettingsActivity` and `SignUpActivity`

willsb:
On hold until the card from the Main activity is Rxfactored

simonrozsival:
Wow, this looks really good 👍 You guys are doing a really good job with these perf improvements 🚀 🚀 🚀

PR #3584, opened 2018-10-29
Track event when detecting new workspaces

### LINKED ISSUE #3444: Track when user gains access to a new workspace
Any time user gains back access to a new workspace, track that it happened.

Can we somehow ignore this right after login and sign-up? The user always gains access to at least one workspace (unless a user who has no workspace signs up). Please suggest a _nice_ way how to ignore these cases (for sign-ups we can check the onboarding storage for "is new user", but what about login?). Let's first discuss it and start writing the code only afterwards.

juanlaube:
## What's this?
Tracks an event when detecting access to new workspaces.

### Relationships

Closes #3444

## Why do we want this?
Analytics and stuff. We want to know how often users get access to new workspaces.

## How is it done?
Just tracking an event when new workspaces are detected. A new property in IOnboardingStorage is required to avoid tracking when the user gains access to workspaces for the first time.

### Why not another way?
We agreed to do it this way.

### Side effects
None.

## Review hints
(shrug)

## :squid: Permissions
Whenever, anyone.


Toggl.Foundation/Sync/States/Pull/DetectGainingAccessToWorkspacesState.cs
```diff
@@ -36,7 +51,18 @@ private IEnumerable<IWorkspace> workspacesWhichAreNotInDatabase(IEnumerable<IWor
...
         private ITransition newWorkspacesDetected(IEnumerable<IWorkspace> workspaces)
-            => NewWorkspacesDetected.Transition(workspaces);
+        {
+            if (onboardingStorage.GetFirstWorkspaceSynced())
+            {
+                analyticsService.GainWorkspaceAccess.Track();
+            }
+            else
+            {
+                onboardingStorage.SetFirstWorkspaceSynced(true);
```
simonrozsival: This is kind of smart - after login/signup, when the first sync runs, it will always consider it as "new" and this code will be reached. I am a bit worried that it is too smart.

The simpler solution would be to always return `Continue` if this is the first sync and skip diffing altogether (not as a performance optimization, but simply to make it easier to reason about.

simonrozsival: What do you think about it? Does it make sense or am I overthinking this? 😄

juanlaube:
Because user is persisted after workspaces, I changed the logic to decide if we should track an event or not based on if there's a user in the database. This avoids any undesired event that could be tracked on the first sync.

Toggl.Foundation/Sync/States/Pull/DetectGainingAccessToWorkspacesState.cs
```diff
@@ -1,32 +1,45 @@
...
+        private readonly IInteractor<IObservable<Unit>> trackNewWorkspaces;
+
         public StateResult<IFetchObservables> Continue { get; } = new StateResult<IFetchObservables>();
         public StateResult<IEnumerable<IWorkspace>> NewWorkspacesDetected { get; } = new StateResult<IEnumerable<IWorkspace>>();
 
-        public DetectGainingAccessToWorkspacesState(IDataSource<IThreadSafeWorkspace, IDatabaseWorkspace> dataSource)
+        public DetectGainingAccessToWorkspacesState(
+            IDataSource<IThreadSafeWorkspace, IDatabaseWorkspace> dataSource,
+            IInteractor<IObservable<Unit>> trackNewWorkspaces
+            )
```
simonrozsival: I know that there is not a consensus on where to put trailing parentheses and I understand the reasons why you would do it this way, but it is just ugly. Please put it at the end of the line 29! 😄

Toggl.Foundation/Interactors/Workspace/TrackNewWorkspacesInteractor.cs
```diff
@@ -0,0 +1,40 @@
...
+
+        public TrackNewWorkspacesInteractor(IAnalyticsService analyticsService, ITogglDataSource dataSource)
+        {
+            Ensure.Argument.IsNotNull(analyticsService, nameof(analyticsService));
+            Ensure.Argument.IsNotNull(dataSource, nameof(dataSource));
+            this.dataSource = dataSource;
+            this.analyticsService = analyticsService;
+        }
+
+        public IObservable<Unit> Execute()
```
simonrozsival: I think this observable should just return `bool` and the tracking should be in the state itself.

Toggl.Foundation/Sync/States/Pull/DetectGainingAccessToWorkspacesState.cs
```diff
@@ -1,32 +1,45 @@
...
         {
             Ensure.Argument.IsNotNull(dataSource, nameof(dataSource));
+            Ensure.Argument.IsNotNull(trackNewWorkspaces, nameof(trackNewWorkspaces));
+
             this.dataSource = dataSource;
+            this.trackNewWorkspaces = trackNewWorkspaces;
         }
 
         public IObservable<ITransition> Start(IFetchObservables fetch)
             => fetch.GetList<IWorkspace>()
```
simonrozsival: I'd say - first execute the interactor (assuming it returns `bool`) and based on the result either return `Continue` or track + return `NewWorkspacesDetected`.

PR #3585, opened 2018-10-29
RxFactor Colors selector on Android

kornelijepetak:
## What's this?
Solves the Android part of Color Selection RxFactor.

### Relationships
Android part of #3559 

## Why do we want this?
So we be more reactive. 

## How is it done?
Magic.

### Why not another way?
Instruct me.

### Side effects
Should be none.

## Review hints
Take it slowly. Shouldn't be too hard.

## :squid: Permissions
@Odrakir, me, and you.


Toggl.Giskard/Fragments/SelectColorFragment.cs
```diff
@@ -1,39 +1,84 @@
...
         public SelectColorFragment() { }
 
         public SelectColorFragment(IntPtr javaReference, JniHandleOwnership transfer)
-            : base (javaReference, transfer) { }
+            : base(javaReference, transfer) { }
 
         public override View OnCreateView(LayoutInflater inflater, ViewGroup container, Bundle savedInstanceState)
         {
             base.OnCreateView(inflater, container, savedInstanceState);
             var view = this.BindingInflate(Resource.Layout.SelectColorFragment, null);
```
daividssilverio: Please remove the binding inflate since we are not using `MvxBind` anymore for this view.

PR #3587, opened 2018-10-29
Reload main log when workspaces are created

### LINKED ISSUE #3469: Reload time entries log when user gains access to a workspace
When the user gains access to a workspace, we should show all the data we have locally which we didn't delete from the database.

Simply reload the log so all of the time entries are shown correctly.

### LINKED ISSUE #3470: Reload time entries log when user gains access to a workspace
This is an Android counterpart of #3469. The specs are exactly the same.

juanlaube:
## What's this?
Reloads the main log after a workspace is created.

### Relationships

Closes #3469, closes #3470 

## Why do we want this?
We want to show any recent data from the new workspace.

## How is it done?
By subscribing to the `Created` observable in the workspaces data source and reloading the main log.

### Why not another way?
This is the easiest way and similar to what we do when losing access to a workspace.

### Side effects
None that I can think of.

## Review hints
Since the states to detect new workspaces and persist them are not connected into the sync graph yet, testing is not easy (I will test this later and post the results in a comment).

## :squid: Permissions
Only me.


Toggl.Foundation.MvvmCross/ViewModels/MainViewModel.cs
```diff
@@ -4,6 +4,7 @@
 using System.Reactive.Disposables;
 using System.Reactive.Linq;
 using System.Reactive.Threading.Tasks;
+using System.Security.Cryptography;
```
simonrozsival: 👀

juanlaube: 😂

juanlaube:
I managed to test this and works as expected! It's not `wip` anymore.

PR #3589, opened 2018-10-30
Fix no (default) workspace view presentation

### LINKED ISSUE #3569: No (default) workspace view might not be presented a second time
`MainViewModel` has 2 bool properties to keep track if no workspace view and select default workspace view have been presented. However they are not properly reset after the view has been closed, which leads to the possibility of the following bug:
1. User gets into no workspace or no default workspace state
2. The appropriate dialog gets presented
3. User solves the problem with the dialog and it is closed
4. User gets into the same state again without closing the app
5. This time the dialog isn't shown, since the bool field is already true

The bool properties were introduced in #3562

arturs105:
## What's this?
This adds a potential fix for the issue mentioned below. **This does not yet fix the tests** as this is just a proposal. If this gets positive reviews, i will fix the tests.

### Relationships

Closes #3569 

## How is it done?
By making the view models return a value so that awaiting the `Navigate()` calls only returns when the view model is closed. After that the appropriate bool variable is reset.

## :squid: Permissions
Me


Toggl.Foundation.MvvmCross/ViewModels/MainViewModel.cs
```diff
@@ -384,18 +384,29 @@ private async Task continueMostRecentEntry()
         public override void ViewAppearing()
         {
             base.ViewAppearing();
-
+            
             IsInManualMode = userPreferences.IsManualModeEnabled;
+            handleNoWorkspaceState()
```
simonrozsival: Couldn't we move this into `Initialize`? 🤔 Would that (even) work?

arturs105: That would work, but i think the current way works better. If we moved the code to `Initialize`, then these checks would be performed only once - when presenting the `MainViewModel` for the first time, whereas now, these checks are performed whenever the `MainViewModel` is shown, i.e. when user goes back from settings, resumes app from background, switches to the main tab from tab bar etc.

PR #3593, opened 2018-10-30
🚧📐 Remove old Suggestion views before adding new ones

sofokli1:
## What's this?
The UI was adding suggestion views on top of old ones thus making it main view look strange.

### Relationships
This issue will be visible after PR #3592 is merged. Doesn't close any issue

## Why do we want this?
So it looks nice as before

## How is it done?
When suggestions are updated it removes all suggestions views before adding new ones.

### Why not another way?
🤷‍♂️

### Side effects
Only good things

## Review hints
Go to PR #3592 
- have some data that show suggestions
- pull to sync a few times and you will see the shadow of the suggestion getting stronger and stronger due to multiple suggestions views on top of each other.

## :squid: Permissions
🦑


simonrozsival:
Will there be a similar problem on Android?

sofokli1:
@simonrozsival I did a functional test on android. Visually was OK. Not sure about the code, not familiar with it.

PR #3598, opened 2018-10-30
Export app state to zipped JSON

simonrozsival:
## What's this?
This is an experiment I did while I was trying to figure why my app is not syncing with the server (the problem was between the chair and the keyboard) and I think it is an interesting concept we might want to consider.

This branch is not merge ready:
- it covers only iOS (that might be fine, Android implementation could be done in a separate PR)
- there are no unit tests for the export itself (I didn't want to make them in case we will decide to throw this away)
- we should discuss the format in which the data is exported and which data should be exported

### Relationships

Closes nothing, needs #3597 before this can be shipped to the users.

## Why do we want this?

The app won't be supported forever and we might have some bugs which will prevent apps from syncing as happened with the old apps. We never know. This feature would give the users the ability to save their data from the app (there is currently no way of doing it).

## How is it done?

I added a button into the iOS settings view which will fetch all the data from the DB, store it in an object and serialize it into a JSON. Then the data is zipped and user can share it using the app of his choice using system dialog or send it to a different device using. AirDrop.

Ultrawave is abused a bit to get JSON of the whole app state. Standard .NET classes are used for creating ZIP - no new NuGet packages are needed.

## Why not any other way?

The zipping part might not be necessary. Users don't typically have that much data in their DBs (I think I use the app a lot and I have only a few megs of data). Performance was OK on an iPhone 7 (fetching + serializing + zipping happened instantly, no noticeable lag).

### Side effects
Some pieces of information are not exported:
- sync status
- last sync error message
- soft deleted flag

Soft deleted and inaccessible entities are not included in the export.

## Review hints
Try this on your iOS device - then use AirDrop to send it to your Mac - unzip it and view it in a JSON viewer.

## :squid: Permissions
Let's not 🦑 this until it is confirmed that we want this feature and there are tests + it is not `wip` anymore.


simonrozsival:
This won't be used any time soon. I created an issue where we can discuss it and maybe consider it later.

PR #3600, opened 2018-10-31
♻️ Rxfactor SelectClientViewModel

khoi:
### Relationships
Part of #3352 

## Why do we want this?
Less Mvx, more Rx

## How is it done?
The ViewModel is now exposing
```
public ObservableGroupedOrderedCollection<SelectableClientViewModel> Clients { get; }
```
which then be used by each platform to display it in a tableview accordingly.

This property will emit events based on the input text of the search text box.

I renamed `CreateEntityViewCell` to `MvxCreateEntityViewCell` and created a new `CreateEntityViewCell` to remove Mvx related stuff.

### Why not another way?
I'm not aware of any other way.

### Side effects
I could miss something. One thign to pay attention to is

I 🔥 of `SelectClientParameters`'s `SelectedClientId`

since I've checked the design and this information is not displayed. 

## Review hints
Start from the `SelectClientViewController`'s `ViewDidLoad` to see how every thing is binded together.

## :squid: Permissions
Me.


Toggl.Daneel/ViewControllers/SelectClientViewController.cs
```diff
@@ -20,38 +23,33 @@ public override void ViewDidLoad()
...
-                      .For(v => v.SuggestCreation)
-                      .To(vm => vm.SuggestCreation);
-
-            bindingSet.Bind(source)
-                      .For(v => v.Text)
-                      .To(vm => vm.Text);
-
-            bindingSet.Apply();
+            SuggestionsTableView
+                .Rx()
```
Odrakir: Can you put the `.Rx()` in the same line as `SuggestionsTableView`. Line separation is great for each operator, but `Rx()` is not an operator.

Toggl.Daneel/ViewControllers/SelectClientViewController.cs
```diff
@@ -20,38 +23,33 @@ public override void ViewDidLoad()
...
+                .Bind(tableViewSource)
+                .DisposedBy(DisposeBag);
 
             SearchTextField.BecomeFirstResponder();
         }
 
+        public override void ViewDidDisappear(bool animated)
+        {
+            base.ViewDidDisappear(animated);
+            ViewModel.DisposeBag.Dispose();
```
Odrakir: Why this? Ideally you wouldn't need a `DisposeBag` in the view model, but if you do... shouldn't it be disposed in it without exposing this outside?

Toggl.Daneel/ViewSources/ClientTableViewSource.cs
```diff
@@ -1,76 +1,87 @@
...
 using Toggl.Daneel.Views;
+using Toggl.Daneel.Views.EntityCreation;
+using Toggl.Foundation.MvvmCross.Collections;
+using Toggl.Foundation.MvvmCross.ViewModels;
 using UIKit;
 
 namespace Toggl.Daneel.ViewSources
 {
-    public sealed class ClientTableViewSource : MvxTableViewSource
+    public sealed class ClientTableViewSource : ReactiveSectionedListTableViewSource<SelectableClientViewModel, ClientViewCell>
```
Odrakir: I think this whole table view source might be simpler if we move most of its logic to the view model. Meaning the observable collection should already have the first row if needed. The model used for the cells should be the same for that first "Creat client" cell and the rest. That way you can use the same code in the source no matter what type of cell you are dealing with. Everything else needed to differentiate them should be done in the VM.

Toggl.Foundation.MvvmCross/ViewModels/SelectClientViewModel.cs
```diff
@@ -74,38 +73,47 @@ public override async Task Initialize()
 
             allClients = await interactorFactory.GetAllClientsInWorkspace(workspaceId).Execute();
 
-            Suggestions.Add(noClient);
-            Suggestions.AddRange(allClients.Select(c => new SelectableClientViewModel(c.Name, c.Id == selectedClientId)));
+            ClientFilterText
```
Odrakir: Now that I see this... I'm thinking you don't need an Observable collection at all here, because all the elements of the table are replaced with every update. You can use a plain `IObservable<IEnumerable<MyCellViewModel>>` and reload the table view every time it changes.

That way you can probably remove this subscription and just derived that Observable from `ClientFilterText`

Take a look at #3559 to see an example, its a collection view, but the idea is the same.

Toggl.Foundation.MvvmCross/ViewModels/SelectClientViewModel.cs
```diff
@@ -74,38 +73,47 @@ public override async Task Initialize()
...
+                    if (string.IsNullOrEmpty(text))
+                    {
+                        selectableViewModels = selectableViewModels.Prepend(noClient);
+                    }
+                    return selectableViewModels;
+                })
+                .Subscribe(clients => Clients.ReplaceWith(clients))
+                .DisposedBy(DisposeBag);
+
+            ClientFilterText.Select(text =>
```
Odrakir: With my comment above, this should also be merged into the same observable, removing this subscription as well.

Toggl.Daneel/ViewControllers/SelectClientViewController.cs
khoi: ah, I just want to make sure the subscriptions are disposed when the view's dismissed. 

And according to [MvvmCross](https://www.mvvmcross.com/documentation/fundamentals/viewmodel-lifecycle#viewmodel-deactivation-activation-and-destruction), it's tricky to detect `ViewModel`'s destruction.

I tried to add it to but it's not called

```
~SelectClientViewModel() {
}
```

Toggl.Daneel/ViewControllers/SelectClientViewController.cs
```diff
@@ -20,41 +29,32 @@ public override void ViewDidLoad()
...
-                      .To(vm => vm.CreateClientCommand);
+            SuggestionsTableView.RegisterNibForCellReuse(ClientViewCell.Nib, ClientViewCell.Identifier);
+            SuggestionsTableView.RegisterNibForCellReuse(CreateClientViewCell.Nib, CreateClientViewCell.Identifier);
+            SuggestionsTableView.SeparatorStyle = UITableViewCellSeparatorStyle.None;
+            SuggestionsTableView.Source = tableViewSource;
 
-            bindingSet.Bind(source)
-                      .For(v => v.SuggestCreation)
-                      .To(vm => vm.SuggestCreation);
+            ViewModel.Clients.Subscribe(replaceClients).DisposedBy(DisposeBag);
```
Odrakir: I think this is better in three lines.
```suggestion
            ViewModel.Clients
                .Subscribe(replaceClients)
                .DisposedBy(DisposeBag);
```

Toggl.Daneel/ViewControllers/SelectClientViewController.cs
```diff
@@ -20,41 +29,32 @@ public override void ViewDidLoad()
...
 
-            bindingSet.Bind(source)
-                      .For(v => v.Text)
-                      .To(vm => vm.Text);
-
-            bindingSet.Apply();
+            CloseButton.Rx().BindAction(ViewModel.Close).DisposedBy(DisposeBag);
```
Odrakir: Same here
```suggestion
            CloseButton.Rx()
                .BindAction(ViewModel.Close)
                .DisposedBy(DisposeBag);
```

Toggl.Daneel/ViewControllers/SelectClientViewController.cs
```diff
@@ -20,41 +29,32 @@ public override void ViewDidLoad()
...
+            SearchTextField.Rx().Text().Subscribe(ViewModel.ClientFilterText).DisposedBy(DisposeBag);
```
Odrakir: And here
```suggestion
            SearchTextField.Rx().Text()
                .Subscribe(ViewModel.ClientFilterText)
                .DisposedBy(DisposeBag);
```

Toggl.Daneel/ViewControllers/SelectClientViewController.cs
```diff
@@ -20,41 +29,32 @@ public override void ViewDidLoad()
...
+            tableViewSource.ClientSelected.Subscribe(ViewModel.SelectClient.Inputs).DisposedBy(DisposeBag);
```
Odrakir: And here
```suggestion
            tableViewSource.ClientSelected
                .Subscribe(ViewModel.SelectClient.Inputs)
                .DisposedBy(DisposeBag);
```

Toggl.Giskard/Resources/layout/SelectClientCreationActivityCell.axml
```diff
@@ -0,0 +1,40 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<FrameLayout
```
kornelijepetak: This creation cell can be reused for other entities as well, so I'm not sure if we need one for each. Maybe the name should not contain the Client.

Toggl.Foundation.MvvmCross/ViewModels/SelectClientViewModel.cs
```diff
@@ -1,112 +1,121 @@
...
+
+            Clients = clientFilterText
+                .Select(text =>
+                {
+                    var trimmedText = text.Trim();
+                    var selectableViewModels = allClients
+                        .Where(c => c.Name.ContainsIgnoringCase(trimmedText))
+                        .Select(toSelectableViewModel);
+
+                    var shouldHaveNoClient = string.IsNullOrEmpty(trimmedText);
```
kornelijepetak: I think the name of this variable is confusing. It sounds as though the TE or something else should have no client. I think it should contain the information about the domain, not about what you intend to do if some condition is met. For example: `isClientFilterEmpty`.
That way, the next code looks more understandable: 
```
var suggestCreation = !isClientFilterEmpty
    && allClients.None(c => c.Name == trimmedText)
    && trimmedText.LengthInBytes() <= 
```

Toggl.Foundation.MvvmCross/ViewModels/SelectClientViewModel.cs
```diff
@@ -1,112 +1,121 @@
...
+        private ISubject<string> clientFilterText = new BehaviorSubject<string>(string.Empty);
 
-        public MvxObservableCollection<SelectableClientViewModel> Suggestions { get; }
-            = new MvxObservableCollection<SelectableClientViewModel>();
+        private readonly IInteractorFactory interactorFactory;
+        private readonly IMvxNavigationService navigationService;
+        private readonly ISchedulerProvider schedulerProvider;
+        private long workspaceId;
+        private long selectedClientId;
+        private SelectableClientViewModel noClient;
```
Odrakir: This is just a detail, but I don't see the need of having a property for this, when it's only used once. Can't we create it it when needed? It could even be a static property on `SelectableClientViewModel`

khoi: This property need to be initialized in `Prepare` so that we have access to `selectedClientId`

Odrakir: Doh!

PR #3602, opened 2018-10-31
Use native HTTP handlers instead of managed ones

### LINKED ISSUE #3588: Proxy Support
## 🐛 Bug Report

## Context

Phone OS: Android 9.0 Pie
Phone model: Essential Ph-1

## Expected behavior

## To reproduce

## Suggestions

simonrozsival:
## What's this?

This PR changes the managed HTTP handler to native ones on different platforms:
- https://docs.microsoft.com/en-us/xamarin/cross-platform/macios/http-stack
- https://docs.microsoft.com/en-us/xamarin/android/app-fundamentals/http-stack

### Relationships
Closes #3588 

## Why do we want this?
We want our app to respect the platform's network settings including the proxy settings.

That would enable us to debug our apps networking using apps like Charles Proxy.

## How is it done?
The `HttpMessageHandler` is set explicitly depending on the platform.

### Why not another way?
I tried implementing this first in #3594 but that was far from optimal.

### Side effects
Our integration tests will run with different HTTP handlers than the apps. That should not be a problem IMO.

## Review hints
- Try this on your physical device - everything should work as it did until now.
- Try verifying that it respects proxy settings - e.g. by using Charles Proxy for capturing the requests

## ~WIP~

- it works well for me on iOS 💪 
- it works well for me on Android when targeting production API 💪 
- ~I am not able to make Android against the staging API with Toggl certificate work 😢~
    - _@daividssilverio confirmed that it works well for him, it is something with my emulator_

## :squid: Permissions
Do not squash just yet.


Toggl.Daneel.SiriExtension/Helper/APIHelper.cs
```diff
@@ -24,7 +25,8 @@ public static ITogglApi GetTogglAPI()
             var version = NSBundle.MainBundle.InfoDictionary["CFBundleShortVersionString"].ToString();
             var userAgent = new UserAgent("Daneel", $"{version} SiriExtension");
             var apiConfiguration = new ApiConfiguration(environment, Credentials.WithApiToken(apiToken), userAgent);
-            return TogglApiFactory.WithConfiguration(apiConfiguration);
+            var httpHandler = new NSUrlSessionHandler(NSUrlSessionConfiguration.DefaultSessionConfiguration);
```
khoi:
```suggestion
            var httpHandler = new NSUrlSessionHandler();
```
would be the same

simonrozsival:
@khoi Can you confirm that it works with your proxy preferences? It works well with mine (Charles), but I would like somebody else to confirm it. Can you please test it?

khoi:
I've tested it on iOS using Charles Proxy for Mac. The app works when
- Proxy off
- Proxy on
- Proxy changes when app's in background

willsb:
I don't get this change, though. When you set the handler on the project, it should replace instances of the regular HttpMessageHandler with the native platform one. Do we really need to explicitly construct the native object and pass it down the chain?

simonrozsival:
@willsb we could remove those and use just the config in `*.csproj`. On the other hand the Android implementation had the native handler defined in the project config, but it was overwritten in code by the managed handler.

simonrozsival:
We should decide what to do with this PR - @willsb do you think we should have `csproj` only configuration and remove all the code? Or remove the settings from the `csproj`s and have it only explicitly in code? Or have both (as the current state of this PR)?

I think having everything defined in code is safer - nobody can override it "by accident" to explicitly set compression for iOS for example.

The settings in the `csproj` files on the other hand communicate the intent to use the platform specific http handler.

And also I don't think we should have something else in the config files and in the code (as we do have right now on Android in `develop`).

Toggl.Daneel/Startup/Setup.cs
```diff
@@ -101,6 +102,8 @@ protected override void InitializeApp(IMvxPluginManager pluginManager, IMvxAppli
             var calendarService = new CalendarServiceIos(permissionsService);
             var notificationService = new NotificationServiceIos(permissionsService, timeService);
 
+            var httpHandler = new NSUrlSessionHandler();
```
willsb: Why no `AutomaticDecompression = DecompressionMethods.GZip | DecompressionMethods.Deflate` here?

simonrozsival: The iOS class doesn't have that property: https://github.com/xamarin/xamarin-macios/blob/master/src/Foundation/NSUrlSessionHandler.cs

PR #3604, opened 2018-10-31
iOS setup improvements

willsb:
## What's this?
This improves the speed of some Setup.cs methods on iOS

## How is it done?
Since the changes are not obvious, I'm explaining the reasoning on comments. This is basically a mirror of #3571

### Side effects
:sanic:

## Review hints
Check #3571

## :squid: Permissions
Me


simonrozsival:
Do you have any (locally measured) stats or estimates of improvement as @daividssilverio showed for Android?

PR #3605, opened 2018-10-31
Add Close extension method

### LINKED ISSUE #3533: Rename our methods to close view models
### The Problem

We have too many means of closing a view model:

- The names are split between `Back` and `Close`
- We have commands, methods and actions
- Sometimes we are closing things manually instead of using the ViewModel

### The solution

- We use close since it best conveys the action of destroying a ViewModel (`Back` implies a specific mean of navigation, which might or might not be truth)
- We should use Actions, since this is the preferred way of handling View -> VM interactions
- We create an extension on `MvxViewModel` so that every concrete implementation can close itself.
- We remove all existing close code and tests. This will significantly reduce useless code.

The result is something like this:

```csharp
public static class ViewModelExtensions
{
    public static UIAction CloseAction(this MvxViewModel viewModel)
        => UIAction.FromAction(() => viewModel.NavigationService.Close(viewModel));
}
```

arturs105:
## What's this?
This PR does 2 things:
1. Adds `MvxViewModel.Close()` extension methods (one regular and the other one with a return value)
2. Refactors already RxFactored views to use these new methods. Views that still use `IMvxCommand` were not touched.

### Relationships

Closes #3533

## Review hints

Reviewing by commit might be easier.

## :squid: Permissions
🦑


Toggl.Foundation.MvvmCross/Extensions/ViewModelExtensions.cs
```diff
@@ -0,0 +1,14 @@
...
+using Toggl.Multivac.Extensions;
+
+namespace Toggl.Foundation.MvvmCross.Extensions
+{
+    public static class ViewModelExtensions
+    {
+        public static UIAction Close(this MvxViewModel viewModel)
+            => UIAction.FromAsync(() => viewModel.NavigationService.Close(viewModel));
+
+        public static UIAction Close<TResult>(this MvxViewModelResult<TResult> viewModel, TResult result)
```
willsb: Why not add something like:

```csharp
public static UIAction Close(this MvxViewModelResult<Unit> viewModel)
    => viewModel.NavigationService.Close(viewModel, Unit.Default);
```

arturs105: 1. I don't see that big of a benefit from this.
2. Wouldn't this create ambiguity, since `MvxViewModelResult<T>` inherits from `MvxViewModel`? Or would C# prefer the extension with the more concrete type?

willsb: Agreed

Issue #3374, opened 2018-10-10
Set up syncing graph for losing and gaining access to workspace


simonrozsival:
Depends on #3372 

The order of states during `pull` sync should be this:

- `FetchAllSinceState`
- `DetectLosingAccessToWorkspacesState` (#2899)
- State from #3357 
    - if there are new workspaces -> go to state #3436 and then to  #3358, from this state always loop back to `FetchAllSinceState`
    - else continue to `persistWorkspaces`

juanlaube:
Why do we need to schedule a clean up loop when there are new workspaces? 🤔 

I think it's the opposite, we need to schedule the clean up when losing access to workspaces, and for that we may need a new transition result from `DetectLosingAccessToWorkspacesState` that goes to the clean up state only if needed.

simonrozsival:
You are right.

simonrozsival:
Implementation was merged into a release branch.

PR #3607, opened 2018-10-31
Gain access to workspace transitions

### LINKED ISSUE #3359: Connect sync states with transitions for gaining access to workspaces
Depends on #3357, #3358 

The state from #3357 should be connected right after the `fetchAllSince` in the `Pull sync` part of the algorithm.

If new workspaces are NOT detected, it should continue to the `persistWorkspaces`.

Else (some new workspaces are detected) it should continue to the state from #3358.

From this state it should always continue to `fetchAllSince` to repeat the loop.

juanlaube:
## What's this?
Adds the sync transitions for detecting and persisting new workspaces.

### Relationships

Closes #3359 

## Why do we want this?
To persist the new workspaces.

## How is it done?
By connecting the states to the sync graph.

### Side effects
I had to 🔥 the `HasFinishedSyncBeforeInteractor` and move the logic into the sync state (which was easy btw) because a dependency loop. To create the SyncManager I needed the InteractorFactory, and to create this I needed the DataSource, and to create this I needed the SyncManager... and so on.

## Review hints
Test it!
1. Gain access to new workspaces
2. Pull to refresh
3. You should see new workspaces
4. Lose access to some workspace
5. Pull to refresh
6. Everything should work

## :squid: Permissions
Anyone, whenever


Toggl.Foundation/TogglSyncManagerFactory.cs
```diff
@@ -150,15 +155,17 @@ public static class TogglSyncManager
             transitions.ConfigureTransition(entryPoint, fetchAllSince);
 
             // start all the API requests first
-            transitions.ConfigureTransition(fetchAllSince.FetchStarted, detectLosingAccessToWorkspaces);
+            transitions.ConfigureTransition(fetchAllSince.FetchStarted, detectGainingAccessToWorkspaces);
```
simonrozsival: In https://github.com/toggl/mobileapp/issues/3374 I wanted to first detect losing access and afterwards gaining access. I don't really remember the reasoning at that time, I didn't write the issue with that much detail in it.

Do you think it is better to first detect gaining access and then losing access? Why?

juanlaube: I don't really think this makes a difference, the database state will be the same one way or another. The only advantage I see in doing gaining access first is we persist the new workspaces earlier.

simonrozsival: If we have tests, it should be fine :)

PR #3612, opened 2018-11-01
Add sync test for cleanup loop

### LINKED ISSUE #3581: Test that clean-up of inaccessible entities removes everything in sync
Create a sync test which
- already has some inaccessible entities in the database (no need to detect that the workspace is inaccessible in this test) - everything should be in sync
- schedule clean-up loop in `Act`
- test, that it removed all inaccessible TEs, clients, tags, projects, tasks related
- test, that the workspace itself was removed

juanlaube:
## What's this?
Adds a new sync test. This test is for the deletion of inaccessible entities that are in sync and can be deleted.

### Relationships

Closes #3581

## Why do we want this?
🚫 🐛 

### Side effects
A good one: I added a line to clear the database on the base class constructor because if the a suddenly crashes, the database contains dirty data in the next run.

## Review hints
Try running `sh build.sh --target Tests.Sync`

## :squid: Permissions
🦑


simonrozsival:
Can you please run the test in bitrise and add a link?

juanlaube:
Bitrise link: https://app.bitrise.io/build/3e4afe512b994e37

PR #3613, opened 2018-11-01
Add test to verify inaccessible entities out of sync are not deleted in a clean up

### LINKED ISSUE #3582: Test that clean-up of inacessible entities does not remove unsynced and unsyncable entities
Create a sync test which
- already has some inaccessible entities in the database (no need to detect that the workspace is inaccessible in this test) - everything should be either unsynced or unsyncable
- schedule clean-up loop in `Act`
- test, that it DID NOT remove any inaccessible TEs, clients, tags, projects, tasks related to the workspace
- test, that the workspace itself was NOT removed

juanlaube:
## What's this?
Adds a new sync test. This is to test inaccessible entities that are out of sync are not deleted.

### Relationships

Closes #3582

## Why do we want this?
🚫 🐛 

### Side effects
None.

## Review hints
Try running `sh build.sh --target Tests.Sync`

## :squid: Permissions
🦑


juanlaube:
Bitrise link: https://app.bitrise.io/build/e624d107790342b3

PR #3616, opened 2018-11-01
Add docs for syncing tests

simonrozsival:
## What's this?
Docs.

### Relationships
Ref #3560 

## Why do we want this?
So there is some information on how to deal with syncing tests.

## Review hints

Let me know if some information that should be there is missing.

## :squid: Permissions
🦑


arturs105:
Looking good, however, i think a link to this should be added in the `index.md` file

PR #3631, opened 2018-11-01
The Wheel rendering refactoring

simonrozsival:
## What's this?
This PR simplifies the rendering of the Wheel view on iOS. Until now the arc was computed using non-trivial trigonometry and the shape layers were re-created every time the layout changed, which could have been even 60 times per second.

Now Apple does the hard part and we will avoid many allocations and make the GC work less.

### Relationships
There is no issue for this PR.

## Why do we want this?
Simpler code which can be later ported to Android.

## How is it done?
It simplifies the code (a lot IMO).

### Side effects
I removed the weird tiny shadow that was at the stop cap. I don't think anybody is gonna miss it.

## :squid: Permissions
🦑


Toggl.Daneel/Views/EditDuration/WheelForegroundView.cs
```diff
@@ -144,6 +130,12 @@ public bool IsRunning
             }
         }
 
+        private CAShapeLayer fullWheel;
+        private CAShapeLayer arc;
+        private CAShapeLayer endCap;
+        private CAShapeLayer startCap;
+        private CGRect currentFrame;
```
arturs105: I don't really see the point of having `currentFrame`. It might as well be something along the lines of `bool isInitialized`, since it's value is never used.

Toggl.Daneel/Views/EditDuration/WheelForegroundView.cs
```diff
@@ -157,42 +149,55 @@ public override void AwakeFromNib()
...
-            Layer.AddSublayer(backgroundLayer);
+            CATransaction.Begin();
+            CATransaction.DisableActions = true;
+
+            // caps
+            startCap.Position = startTimePosition;
+            startCap.FillColor = foregroundColor;
+            endCap.Position = endTimePosition;
+            endCap.FillColor = foregroundColor;
+            endCap.Sublayers.Single().Hidden = IsRunning;
```
arturs105: I assume there isn't a case, where `endCap` has several sublayers.

simonrozsival: It can't have any more layers - everything is private. But I think I will create classes for the different elements.

Toggl.Daneel/Views/EditDuration/WheelForegroundView.cs
simonrozsival: The idea is to detect when the view is resized so the subviews can be scaled to the new dimensions.

simonrozsival:
Oops, I broke it today... it looks kinda cool without the bg, maybe it is a feature.. 😄

Toggl.Daneel/Views/EditDuration/Shapes/Caps/Cap.cs
```diff
@@ -0,0 +1,57 @@
...
+using MvvmCross.Plugin.Color.Platforms.Ios;
+using UIKit;
+using Math = Toggl.Multivac.Math;
+
+namespace Toggl.Daneel.Views.EditDuration.Shapes.Caps
+{
+    public abstract class Cap : CAShapeLayer
+    {
+        private readonly CGColor capColor
+            = Foundation.MvvmCross.Helper.Color.EditDuration.Wheel.Cap.ToNativeColor().CGColor;
```
arturs105: Do we really need the whole namespace? If a regular `using` creates an error, then maybe a named `using` would work?

Toggl.Daneel/Views/EditDuration/Shapes/Caps/Cap.cs
```diff
@@ -0,0 +1,57 @@
...
+
+        private readonly nfloat innerRadius = 14.05f / 128f;
+
+        public CGColor Color
+        {
+            set => FillColor = value;
+        }
+
+        protected Cap(
+            CGImage image,
```
arturs105: I don't think `image` is a good name. I assume this is the icon that's displayed on the cap right? If so then `capIcon` or something like that would be a better name.

Toggl.Daneel/Views/EditDuration/Shapes/Caps/EndCap.cs
```diff
@@ -0,0 +1,28 @@
...
+{
+    public sealed class EndCap : Cap
+    {
+        private static readonly CGImage image = UIImage.FromBundle("icEndLabel").CGImage;
+
+        // The sizes are relative to the radius of the wheel.
+        // The radius of the wheel in the design document is 128 points.
+        private static readonly nfloat imageHeight = 10f / 128f;
+
+        private static readonly nfloat imageWidth = 10f / 128f;
```
arturs105: This 128 probably should be extracted somewhere.

Toggl.Daneel/Views/EditDuration/Shapes/Caps/StartCap.cs
```diff
@@ -0,0 +1,24 @@
...
+
+namespace Toggl.Daneel.Views.EditDuration.Shapes.Caps
+{
+    public sealed class StartCap : Cap
+    {
+        private static readonly CGImage image = UIImage.FromBundle("icStartLabel").CGImage;
+
+        // The sizes are relative to the radius of the wheel.
+        // The radius of the wheel in the design document is 128 points.
+        private static readonly nfloat imageCenterHorizontalCorrection = 1.4f / 128f;
```
arturs105: I think `imageCenterHorizontalCorrection ` deserves a comment explaining it's purpose.

Toggl.Daneel/Views/EditDuration/WheelForegroundView.cs
```diff
@@ -152,46 +114,59 @@ public override void AwakeFromNib()
...
-            Layer.AddSublayer(backgroundLayer);
+            CATransaction.Begin();
+            CATransaction.DisableActions = true;
 
-            if (IsRunning == false)
-            {
-                var endCap = createCap(endTimePosition, endHandleImage, squareWidth, squareHeight, suqareCenterHorizontalOffset);
-                Layer.AddSublayer(endCap);
-            }
+            // caps
```
arturs105: Other comments in this PR are capitalised.

Toggl.Daneel/Views/EditDuration/Shapes/Caps/Cap.cs
simonrozsival: There was some kind of a conflict. I will look into it once more and try to make it a bit nicer 👍

simonrozsival: I don't think I am able to do anything about this. I suggest keeping it like this for now. The real solution would be to reorganize the colors static classes.

Toggl.Daneel/Views/EditDuration/Shapes/Caps/EndCap.cs
simonrozsival: I would keep the 128s there as they are. I think the comment describes the purpose well enough.

arturs105: It's just this value is used in a lot of other places in in `StartCap`, `BaseWheelView`, `WheelForegroundView` and `WheelBackgroundView` as well. I was thinking this value could be a public constant in `BaseWheelView`.

PR #3634, opened 2018-11-02
Fix failing sync test because we now have a clean-up loop

simonrozsival:
## What's this?
This test, which was failing, was written before we had the clean-up transitions set up. Now we remove all time entries which were marked as inaccessible which were in sync.

## Why do we want this?
We want our tests to be accurate and reliable.

## How is it done?
The entities are marked as `SyncNeeded` so they won't be removed during the full-sync process.

### Why not another way?
This is a good way. Now we also need a test for the clean-up loop.

### Side effects
No side effects.

## Review hints
Run `sh build.sh --target Tests.Sync` or look at this Bitrise build:
- :x:~https://app.bitrise.io/build/f1f8dd2ef965e6af~
- fixed build: https://app.bitrise.io/build/5091f7124b825f95

## :squid: Permissions
🦑 when unit and sync tests are passing


sofokli1:
I guess those workspaces should not even be there after sync

simonrozsival:
I have to fix the assertion, Im not sure how I missed that.

PR #3637, opened 2018-11-02
Sync tests - server push functionality upgrade

simonrozsival:
## What's this?
The push function of the `Server` helper class has at least two flaws:
- you can't set the properties of the default workspace
- you can't delete the default workspace

This PR fixes these two flaws.

## Why do we want this?
We want to have possibility to test every scenario possible.

## How is it done?
When the default workspace is missing from the list of desired workspaces, it will be deleted.

When the default workspace is in the list, it is (always) updated.

`Ultrawave` does not support these two operations, so I added them into the `WorkspaceHelper` class in the integration tests project.

### Why not another way?
Any suggestions?

### Side effects
No.

## Review hints
Run `sh build.sh --target Tests.Sync` locally or check this Bitrise build: _(link will be added later)_

## :squid: Permissions
🦑


Toggl.Ultrawave.Tests.Integration/Helper/WorkspaceHelper.cs
```diff
@@ -31,6 +32,22 @@ public static async Task<List<int>> GetAllAvailablePricingPlans(IUser user)
...
+                workspace as Workspace ?? new Workspace(workspace), SerializationReason.Post);
+            var responseJson = await makeRequest(
+                $"https://toggl.space/api/v9/workspaces/{workspace.Id}", HttpMethod.Put, user, serializedJson);
+            return serializer.Deserialize<Workspace>(responseJson);
+        }
+
+        public static async Task Delete(IUser user, long workspaceId)
+        {
+            await makeRequest(
+                $"https://toggl.space/api/v8/workspaces/{workspaceId}/leave", HttpMethod.Delete, user, null);
```
juanlaube: Is v8 correct? 🤔

PR #3651, opened 2018-11-05
Enable auto correct in start view

### LINKED ISSUE #3644: Enable auto correct in start view text field
This is probably the most important text field in the entire app, and it's missing the built-in auto-correction. I personally make tons of spelling errors because of that.

Let's see if we can enable that without breaking anything.

arturs105:
## What's this?
This enables autocorrect in the start time entry view.

### Relationships

Closes #3644

## How is it done?
I enabled autocorrect for the `UITextView` in the .xib file. XCode and VS did some other shenanigans which resulted in the other changes in the file, though they shouldn't matter much.

### Side effects
Tested this for a bit trying out different things and didn't notice any issues, but please do a thorough functional test as well.

## :squid: Permissions
🦑


arturs105:
The issue about cursor being positioned before the text is a thing on `develop` as well. I fixed it in #3661. Now looking into other problems.

arturs105:
The cursor issue was fixed in #3661 and can't really reproduce the issue that @sofokli1 pointed out, so this looks mergeable.

PR #3656, opened 2018-11-05
Fix sync graph dead ends

### LINKED ISSUE #3649: Unintentional dead ends in the push sync transitions graph
There are unintentional dead ends after the state which marks an entity as unsyncable:

screen shot 2018-11-05 at 4 29 05 pm

Instead of dead end, it should loop back into the `Push<TDatabase, TThreadsafe>` state.

simonrozsival:
## What's this?
This PR fixes incorrect transitions in the push sync transitions graph.

### Relationships
Closes #3649

## Why do we want this?
We want our sync mechanism to work seamlessly.

## How is it done?
- the `UnsyncableEntity<T>.MarkAsUnsyncable` leads into `ResetAPIDelay` state, which then goes back into the `Push<T>` state which tries to push another entity
- the `UpdateEntity<T>.EntityChanged` when pushing singletons (user object - when the default workspace changes and preferences) also goes into the `ResetAPIDelay` to repeat pushing the entity again with the new data

### Why not another way?
I can't see any other possibility.

### Side effects
No.

## Review hints
Check the graphs [before and after](https://github.com/toggl/mobileapp/files/2549678/before_and_after.zip)

## :squid: Permissions
🦑


amulware:
I checked the graph diff and it looks good.

PR #3657, opened 2018-11-05
Add a new state to handle api errors

### LINKED ISSUE #3652: Handle HTTP failures before detecting gaining and losing access to workspace
Right now we might not handle 4xx and 5xx errors correctly and sync would finish showing `Failed` message to the user.

We must check if the HTTP request did not fail and if it did, handle it the same way we already do with all the "persist" states.

I think this could be done in a generic way in a state, which would have an internal `Func<IFetchObservables, IObservable<T>>` selector and which would have transitions `Continue[FetchObservables]` and `ErrorOccured[ApiException]`.

We could then add this state before the first state which uses some of the fetch observables for the first time in the pipeline and simplify the `Persist*` states.

juanlaube:
## What's this?
Adds a new state to handle API errors when detecting gaining or losing access to workspaces and refactor all the `IPersistState`s so API error handling is done in a previous state.

### Relationships

Closes #3652

## Why do we want this?
Because error handling is important.

## How is it done?
Thx ConnectedReplay observable. Now we go through `FetchOrErrorState` the first time an observable from `IFetchObservables` is used.

### Why not another way?
We could have handled the errors individually in the detect gaining or losing access to workspaces states and avoided this refactor.

### Side effects
I don't think so.

## Review hints
All test pass (unit and sync) and I manually tested the app, but please double check the transitions in case I messed up.

## :squid: Permissions
🚫


Toggl.Foundation/Sync/States/Pull/FetchSingletonOrErrorState.cs
```diff
@@ -0,0 +1,10 @@
+using System;
+
+namespace Toggl.Foundation.Sync.States.Pull
+{
+    public sealed class FetchSingletonOrErrorState<T> : FetchOrErrorState<T>
+    {
+        protected override IObservable<T> FetchAction(IFetchObservables fetch)
+            => fetch.GetSingle<T>();
```
simonrozsival: This is really smart use of generics! 👍

simonrozsival:
The transitions seem OK to me, but please generate and attach the graph for the state before and after so we can compare it. I really like this PR overall BTW!

Toggl.Foundation/Sync/States/Pull/FetchOrErrorState.cs
```diff
@@ -0,0 +1,23 @@
...
+
+        public StateResult<ApiException> ErrorOccured { get; } = new StateResult<ApiException>();
+
+        public IObservable<ITransition> Start(IFetchObservables fetchObservables)
+            => FetchAction(fetchObservables)
+                .SingleAsync()
+                .Select(_ => Continue.Transition(fetchObservables))
+                .OnErrorReturnResult(ErrorOccured);
+
+        protected abstract IObservable<T> FetchAction(IFetchObservables fetch);
```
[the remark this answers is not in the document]
juanlaube: Naming is hard 🙈

juanlaube:
Updated sync graph: [sync-graph.svg.zip](https://github.com/toggl/mobileapp/files/2553164/sync-graph.svg.zip)

Toggl.Foundation/TogglSyncManagerFactory.cs
```diff
@@ -86,6 +87,16 @@ public static class TogglSyncManager
 
             var fetchAllSince = new FetchAllSinceState(database, api, timeService);
 
+            var fetchWorkspaces = new EnsureFetchListSucceededState<IWorkspace>();
```
simonrozsival: Can we rename these vars as well please? 😃

Toggl.Foundation/Sync/States/Pull/TryFetchInaccessibleProjectsState.cs
```diff
@@ -50,8 +48,7 @@ public IObservable<ITransition> Start(IFetchObservables fetch)
             => getProjectsWhichNeedsRefetching()
                 .SelectMany(projects => projects == null
                     ? Observable.Return(FinishedPersisting.Transition(fetch))
-                    : refetch(projects).Select(FetchNext.Transition(fetch)))
-                .OnErrorReturnResult(ErrorOccured);
```
simonrozsival: This HTTP request is not validated anywhere else, we can't remove it from here.

Toggl.Foundation/TogglSyncManagerFactory.cs
```diff
@@ -210,17 +230,15 @@ public static class TogglSyncManager
...
-            transitions.ConfigureTransition(persistUser.ErrorOccured, retryOrThrow);
-            transitions.ConfigureTransition(persistWorkspaceFeatures.ErrorOccured, retryOrThrow);
-            transitions.ConfigureTransition(persistPreferences.ErrorOccured, retryOrThrow);
-            transitions.ConfigureTransition(persistTags.ErrorOccured, retryOrThrow);
-            transitions.ConfigureTransition(persistClients.ErrorOccured, retryOrThrow);
-            transitions.ConfigureTransition(persistProjects.ErrorOccured, retryOrThrow);
-            transitions.ConfigureTransition(persistTasks.ErrorOccured, retryOrThrow);
-            transitions.ConfigureTransition(createProjectPlaceholders.ErrorOccured, retryOrThrow);
-            transitions.ConfigureTransition(persistTimeEntries.ErrorOccured, retryOrThrow);
-            transitions.ConfigureTransition(refetchInaccessibleProjects.ErrorOccured, retryOrThrow);
```
simonrozsival: 😱 I summon @amulware. This might cause an infinite syncing loop: when refetching of the project fails (e.g., the user doesn't have access to the workspace), the server returns 4xx, we loop back to `fetchAllSince` and it is guaranteed that it will fail again.

Maybe we should just silently skip the projects which we can't refetch?

simonrozsival: This PR should probably just keep things as they were and we should create an issue for this and fix it. And create a sync test for it.

juanlaube:
Updated sync graph:
[sync-graph.svg.zip](https://github.com/toggl/mobileapp/files/2553826/sync-graph.svg.zip)

juanlaube:
[sync-graph_copy_final_copy_copy_(2)_final_final.zip](https://github.com/toggl/mobileapp/files/2554223/sync-graph_copy_final_copy_copy_.2._final_final.zip)

PR #3658, opened 2018-11-05
Giskard UI: Rxfactor Reports and Reports Calendar

### LINKED ISSUE #3416: Lazy load reports calendar pages
These sometimes hiccup to show and are possibly the bottleneck that makes the reports view take 2+ seconds to open on Android. Let's fix this!

daividssilverio:
## What's this?
This is the ui part of the rxfactories of the `ReportsViewModel` and `ReportsCalendarViewModel`.
This also makes the old `ReportsCalendarFragment.cs` a simple android view to help us avoid friction when making the future reports tab (when the bottom bar navigation is done for giskard) a fragment.

### Relationships
This should not be merged before the ongoing rxfactories.
I'll update this WIP completing the bindings when they are done.
Closes #3416 (but isn't just that).

## Why do we want this?
Because our Mvx bindings are slow and we want our UI to go :sanic:

## How is it done?
This is still WIP.  But this is as far I got without the involved VM refactors.
I'll add the bindings when the VM's rxfactories are done.

### Side effects
Possible weirdness caused by making the ReportsCalendarFragment a simple view. I'll test more when the bindings start being made.

## Review hints
I've tried to make review by commit easier. Try to separate what is from Reports and what is from ReportsCalendar, the commits are a bit tangled 🙈.

## :squid: Permissions
Don't merge or 🦑 yet. This is a WIP.


daividssilverio:
Superseded by #4095

PR #3667, opened 2018-11-06
Giskard User Feedback

### LINKED ISSUE #2149: Ask users to rate us in the app
## Shared code
Already implemented on iOS in #1972. All we need is the equivalent Android stuff.

## Flow
Enjoying -> Yes -> Would you mind rating on the Play Store? -> Yes -> Direct user to the Play Store page *(If anybody starts working on this, please, let me know if it's possible to direct user directly to the rating part in the Play Store)*
Enjoying -> Yes -> Would you mind rating on the Play Store? -> No -> Close the popup
Enjoying -> No -> Would you mind giving us some feedback? -> Yes -> Open feedback form
Enjoying -> No -> Would you mind giving us some feedback? -> No -> Close the popup

## Other notes
- [ ] Display the Rating card below the suggestions (if there are any) or just at the top of the page below the nav bar if there are no suggestions.
- [ ] Padding from the Suggestions/Nav bar: 16pt.
- [ ] Padding from the Log: 16pt.
- [ ] Don’t block the UI. Meaning that user can continue his flow (starting a new entry, go to reports etc) if he wants to.
- [ ] When user hasn’t interacted with the card in a 5 min, snooze it for 24h.

### Step1
rate the app

Zeplin: https://zpl.io/a7yXemM

### Step2: v1 When user said "Yes" in the step1
rate the app v2

Zeplin: https://zpl.io/2EjXNgl

### Step2: v2 When user said "No" in the step1
rate the app v3

Zeplin: https://zpl.io/VK0XRDR

kornelijepetak:
## What's this?
Implements User feedback form in 🤖.

### Relationships
Closes #2149 

## Why do we want this?
More ⭐️⭐️ 

## How is it done?
Oh, multiple things.
The conception is the same as in iOS.

1. The dialog itself is implemented as a main log cell
1. RatingService is implemented for android 
1. RemoteConfig service takes care of fetching the data from the firebase
1. Foundation takes care of rating remote config experiment

### Side effects
There should be none, but this needs to be extensively tested.

## Review hints
Commit by commit would be helpful.
I have to think of a way to properly test this, without having to be connected to the firebase.

## :squid: Permissions
Me.

## ➡️ WIP continuation
This PR is currently WIP because I am not sure whether Firebase part is working, but I don't want this to block the remainder of the PR from being reviewed.


Toggl.Giskard/Services/RatingServiceAndroid.cs
```diff
@@ -1,12 +1,42 @@
...
+        public RatingServiceAndroid(Context context)
+        {
+            this.context = context;
+        }
+
         public void AskForRating()
         {
+            var packageName = context.PackageName;
+
+            packageName = "com.toggl.giskard";
```
kornelijepetak: Ignore this, this is for testing. Because there's no GooglePlay package named com.toggl.giskard.debug.

daividssilverio: Actually, there is a firebase package named that. I've added the firebase debug project for giskard when testing Firebase performance.

dshr:
Code-wise this looks 👍 Couldn't test it though, since I'm not hitting the required conditions, and commenting out `MainViewModel.cs:315-332` doesn't really help me get the rating view :( Let me know when you'll have a way to test it locally and I'll approve this right away!

Toggl.Giskard/Adapters/MainRecyclerAdapter.cs
```diff
@@ -129,6 +142,12 @@ protected override MainLogSectionViewHolder CreateHeaderViewHolder(ViewGroup par
             return mainLogSectionViewHolder;
         }
 
+        internal void SetupRatingViewVisibility(bool isVisible)
+        {
+            isRatingViewVisible = isVisible;
```
daividssilverio: Could you check if the view is already visible before calling `NotifyDataSetChanged`?
While debugging, it was being called twice very fast.

kornelijepetak:
> Besides that, be aware that the logic to close the rating view after 5 minutes won't work if the user leaves the main log before the time to update the log. If the puts the app in background or navigates to another screen, the rating view will stay visible.

Why do you think so? I am not following...

daividssilverio:
>Why do you think so? I am not following...

I tested it and it does happen. It's because the activity might not be there when the hint to hide the rate view gets sent.

PR #3671, opened 2018-11-06
Rxfactor the Reports Calendar VM

### LINKED ISSUE #3417: Lazy load reports calendar pages
Since it's a pretty massive collection view, lazy loading it might be what makes the view load instantly instead of hiccuping on first open like it does now.

dshr:
## What's this?
Part of yet another attempt to make the Reports page faster by RxFactoring it. This one focuses on the calendar, since it seems to be the one taking the longest!

### Relationships
Closes #3417

## Why do we want this?
Must proceed at high velocity

## How is it done?
Rx, baby

### Why not another way?
Making the calendar setup properly async will be much harder.

### Side effects
None so far, since this is WIP.

## Review hints
At this point none, just need some sanity checks.

## :squid: Permissions
When done — :sanic: team only.


Toggl.Foundation.MvvmCross/ViewModels/ReportsCalendarViewModel.cs
```diff
@@ -42,57 +47,207 @@ public sealed class ReportsCalendarViewModel : MvxViewModel
...
-        public List<ReportsCalendarBaseQuickSelectShortcut> QuickSelectShortcuts { get; private set; }
+        [Obsolete("Use QuickSelectShortcutsObservable instead")]
+        public IImmutableList<QuickSelectShortcut> QuickSelectShortcuts { get; private set; }
+        public IObservable<IImmutableList<QuickSelectShortcut>> QuickSelectShortcutsObservable { get; }
 
+        [Obsolete("Use CalendarDayTapped instead")]
         public IMvxAsyncCommand<ReportsCalendarDayViewModel> CalendarDayTappedCommand { get; }
-
-        public IMvxCommand<ReportsCalendarBaseQuickSelectShortcut> QuickSelectCommand { get; }
+        [Obsolete("Use CalendarDayTapped instead")]
```
daividssilverio: This should be `QuickSelect`, right?

Toggl.Foundation.MvvmCross/ViewModels/ReportsCalendarViewModel.cs
```diff
@@ -42,57 +47,207 @@ public sealed class ReportsCalendarViewModel : MvxViewModel
...
+                .Subscribe(dayHeaders => DayHeaders = dayHeaders)
+                .DisposedBy(oldVariablesDisposeBag);
+
+            IImmutableList<string> headers(BeginningOfWeek beginningOfWeek)
+                => Enumerable.Range(0, 7)
+                    .Select(index => dayHeaders[(index + (int)beginningOfWeek + 7) % 7])
+                    .ToImmutableList();
+
+            IImmutableList<ReportsCalendarPageViewModel> calendarPages(BeginningOfWeek beginningOfWeek)
+                => Enumerable.Range(0, MonthsToShow)
```
daividssilverio: This doesn't look right to me.
I thought we would hold a reference to the list of Months in the ViewModel and the `MonthsObservable` would emit only "CollectionChanged" events that we could use to update iOS's ViewSources and Android's adapters.

As it is, it looks like the MonthsObservable is gonna emit a whole new `List<Month>`, but only when the beginning of the week changes.

Toggl.Foundation.MvvmCross/ViewModels/ReportsCalendarViewModel.cs
dshr: Yeah, typoed a bit 😅

Toggl.Foundation.MvvmCross/ViewModels/ReportsCalendarViewModel.cs
dshr: Good point! Even better yet, we can have `Months` as something similar to `ObservableCollection`, but Rx-ey (like what we have for `ObservableGroupedOrderedCollection`), which will streamline things a bit.

daividssilverio: Yeah, I think that two simple backing lists would do the trick. One that the vm is controlling all the time (maybe attached to the beginning of the week) and other immutable list to be read by the UI when there are changes. Maybe only the second list would know about the selected/highlighted range.

Toggl.Foundation.MvvmCross/ViewModels/ReportsCalendarViewModel.cs
```diff
@@ -42,54 +47,210 @@ public sealed class ReportsCalendarViewModel : MvxViewModel
             Resources.SaturdayInitial
         };
 
-        private bool isInitialized;
         private CalendarMonth initialMonth;
         private CompositeDisposable disposableBag;
         private ReportsCalendarDayViewModel startOfSelection;
+        private CompositeDisposable calendarDisposeBag;
+        private CompositeDisposable shortcutDisposeBag;
+        private CompositeDisposable oldVariablesDisposeBag;
```
Odrakir: Why multiple dispose bags? Also... shouldn't you be disposing this somewhere?

Toggl.Foundation.MvvmCross/ViewModels/ReportsCalendarViewModel.cs
```diff
@@ -42,54 +47,210 @@ public sealed class ReportsCalendarViewModel : MvxViewModel
...
         private ReportPeriod reportPeriod = ReportPeriod.ThisWeek;
 
         public BeginningOfWeek BeginningOfWeek { get; private set; }
 
         //Properties
-        [DependsOn(nameof(CurrentPage))]
-        public CalendarMonth CurrentMonth => convertPageIndexTocalendarMonth(CurrentPage);
+        [Obsolete("Use CurrentMonthObservable instead instead")]
+        public CalendarMonth CurrentMonth { get; private set; }
+        public IObservable<CalendarMonth> CurrentMonthObservable { get; }
```
Odrakir: I think the convention is to drop the `Observable` part of the name on these. I'm guessing it's there so you can keep the old property, just take it into account.

Toggl.Foundation.MvvmCross/ViewModels/ReportsCalendarViewModel.cs
```diff
@@ -42,54 +47,210 @@ public sealed class ReportsCalendarViewModel : MvxViewModel
...
+                    .ConnectedReplay();
+
+            var highlightObservable = highlightedDateRangeSubject.AsObservable();
+
+            DayHeadersObservable = beginningOfWeekObservable
+                .Select(headers)
+                .AsDriver(schedulerProvider);
+
+            SelectedDateRangeObservable = selectedDateRangeSubject
+                .AsObservable()
```
Odrakir: I think you don't need this line.

Toggl.Foundation.MvvmCross/ViewModels/ReportsCalendarViewModel.cs
```diff
@@ -42,54 +47,210 @@ public sealed class ReportsCalendarViewModel : MvxViewModel
...
+                .AsDriver(schedulerProvider);
+
+            CurrentPageObservable = currentPageSubject
+                .StartWith(MonthsToShow - 1)
+                .AsObservable()
```
Odrakir: I think you don't need this line either.

Toggl.Foundation.MvvmCross/ViewModels/ReportsCalendarViewModel.cs
```diff
@@ -42,54 +47,210 @@ public sealed class ReportsCalendarViewModel : MvxViewModel
...
+
+            QuickSelectShortcutsObservable
+                .Subscribe(shortcuts => QuickSelectShortcuts = shortcuts)
+                .DisposedBy(oldVariablesDisposeBag);
+
+            DayHeadersObservable
+                .Subscribe(dayHeaders => DayHeaders = dayHeaders)
+                .DisposedBy(oldVariablesDisposeBag);
+
+            IImmutableList<string> headers(BeginningOfWeek beginningOfWeek)
```
Odrakir: Any reason why this mapping functions are here and not as private methods in the class?

Toggl.Foundation.MvvmCross/ViewModels/ReportsCalendarViewModel.cs
```diff
@@ -42,54 +47,210 @@ public sealed class ReportsCalendarViewModel : MvxViewModel
...
+                    .ToImmutableList();
+
+            IImmutableList<ReportsCalendarPageViewModel> calendarPages(BeginningOfWeek beginningOfWeek)
+                => Enumerable.Range(0, MonthsToShow)
+                    .Select(initialMonth.AddMonths)
+                    .Select(calendarMonth => new ReportsCalendarPageViewModel(calendarMonth, beginningOfWeek, currentDate))
+                    .Select(month => {
+                        foreach (var day in month.Days)
+                        {
+                            SelectedDateRangeObservable
```
Odrakir: I don't like this a lot: Looping and subscribing each and every day to an observable. Where do all those subscriptions get disposed? Won't those subscriptions get lost (and living forever) when you change months?

I tried to test this and see, but the calendar is empty for me... 🤔

Toggl.Foundation.MvvmCross/ViewModels/ReportsCalendarViewModel.cs
```diff
@@ -42,54 +47,210 @@ public sealed class ReportsCalendarViewModel : MvxViewModel
...
+                    QuickSelectShortcut.ForThisMonth(timeService),
+                    QuickSelectShortcut.ForLastMonth(timeService),
+                    QuickSelectShortcut.ForThisYear(timeService)
+                );
+
+                shortcuts.ForEach(shortcut =>
+                {
+                    SelectedDateRangeObservable
+                        .Subscribe(shortcut.OnDateRangeChanged)
+                        .DisposedBy(shortcutDisposeBag);
```
Odrakir: Same thing here, where does this get disposed?

Toggl.Foundation.MvvmCross/ViewModels/ReportsCalendarViewModel.cs
dshr: Forgot to add the WIP label, sorry 😅 . 

As for the calendar collection view — this is what I'm actually trying to figure out right now. It shows up the correct number of sections and items within, but no cells 🤷‍♂️

daividssilverio:
superseded by #4095

PR #3676, opened 2018-11-07
Global main scheduler

Odrakir:
## What's this?
This is a proposal on how to handle `MainThread` and `MainThread` only in an application wide scope. 

I don't have my hopes up on this being merged, I'm just wondering if we can find a way of doing something similar.

The idea is that things like what the UI thread is, are cross-cutting concerns and shouldn't need to be injected all around. Specifically with this there's not much danger being a system property that won't ever change. 

This allows for the scheduler to be switched with a test one for tests as well, I don't think we need that, though. As I said this is only for the UI thread, if something needs another scheduler, it still has to be injected.

This is sort of based on how ReactiveUI [handles the main thread](https://reactiveui.net/docs/guidelines/framework/ui-thread-and-schedulers). You can see the code [here](https://github.com/reactiveui/ReactiveUI/search?utf8=✓&q=RxApp.MainThreadScheduler&type=).

## Why do we want this?
This is mainly motivated by problems with actions not returning in the main thread, when I think they totally should (only `UIAction` and `InputAction`)

## How is it done?
There's a class called `RxApp` which has a static property called `MainScheduler`, that's it.

I'll admit it's a little ugly right now, we may find a better way of doing it. The only objective is easy access to the UI thread. As I said it's a cross-cutting concern and a non-mutable property.

### Why not another way?
The alternative to this is using factories to make actions creation easier. Maybe there's also a better approach for drivers, I can't say right now, though.

## :squid: Permissions
This is mainly for discussion, so don't.

### DIFF
--- a/Toggl.Daneel/Startup/Setup.cs
+++ b/Toggl.Daneel/Startup/Setup.cs
@@ -24,6 +24,7 @@ using MvvmCross.Platforms.Ios.Presenters;
 using System.Collections.Generic;
 using System.Net.Http;
 using System.Reflection;
+using Toggl.Multivac;
 using ColorPlugin = MvvmCross.Plugin.Color.Platforms.Ios.Plugin;
 using VisibilityPlugin = MvvmCross.Plugin.Visibility.Platforms.Ios.Plugin;
 using Toggl.Multivac.Extensions;
@@ -99,6 +100,7 @@ namespace Toggl.Daneel
             var remoteConfigService = new RemoteConfigServiceIos();
             remoteConfigService.SetupDefaults(remoteConfigDefaultsFileName);
             var schedulerProvider = new IOSSchedulerProvider();
+            RxApp.MainScheduler = schedulerProvider.MainScheduler;
             var calendarService = new CalendarServiceIos(permissionsService);
             var notificationService = new NotificationServiceIos(permissionsService, timeService);
 
--- a/Toggl.Daneel/ViewSources/TimeEntriesLogViewSource.cs
+++ b/Toggl.Daneel/ViewSources/TimeEntriesLogViewSource.cs
@@ -68,7 +68,7 @@ namespace Toggl.Daneel.ViewSources
                 deleteTableViewRowAction.BackgroundColor = Color.TimeEntriesLog.DeleteSwipeActionBackground.ToNativeColor();
             }
 
-            FirstCell = firstCellSubject.AsDriver(schedulerProvider);
+            FirstCell = firstCellSubject.AsDriver();
         }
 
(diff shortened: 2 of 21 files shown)


Odrakir:
By the way, an easy example of the problem with `UIAction` is this.

Imagine I want to know from the VC when an action is executing. I already expose the action from the VM, so I can do something like this:

```c#
ViewModel.Reset.Executing
    .Subscribe(ActivityIndicator.Rx().IsVisibleWithFade())
    .DisposedBy(DisposeBag);
```

But the action returns in another thread (non-UI) so that will cause problems (it did). So now, because I have the `SchedulerProvider` in the VM I have to expose yet another property for when the action is executing:

```c#
IsLoading = Reset.Executing
    .AsDriver(schedulerProvider);
```

simonrozsival:
My concern is: we can't use `TestScheduler` for these observables in our tests (setting `RxApp.MainScheduler` to an isntance of `TestScheduler` is not an option - we run tests in parallel.

If we don't need to use test schedulers in the case of the public VMs' observables, then it could work. But I am a bit skeptical - we test those observables, don't we?

Odrakir:
🤔 ... would use `public static Func<IScheduler> MainScheduler;` help with parallel tests?

simonrozsival:
I don't see how it could help - can you show an example how it would be used?

Odrakir:
Ok, after talking about this with @simonrozsival through Slack I'm convinced it won't work with parallel tests. It sucks, though... 😞

daividssilverio:
🤔 Actually, having a `RxApp.MainScheduler.GetInstance()` or similar might not create problems when running parallel tests, right?
It could always return the same instance when running on devices and return the new TestScheduler instances for every test.

Odrakir:
I don't know anymore, ask @simonrozsival .

simonrozsival:
@daividssilverio that can't work. Having a single instance of `TestScheduler` is what we have right now and there is no other way of passing it to the observables other than how we do it right now - directly through arguments.

Once you start using static classes/static properties, you must account for all the multiple tests, which use the same static context. They are not separate processes, they are just threads in a single process. Tests will be overriding each other's schedulers and losing access to their original scheduler. Different observables will use different schedulers. Tests will sometimes pass and sometimes fail.

I agree that the `.AsDriver(default, schedulerProvider)` is too repetitive. We could use something like

```csharp
var reactiveFactory = new ReactiveFactory(mainThreadScheduler); // this will be a service which would be injected into VMs instead of scheduler provider

var action = reactiveFactory.CreateUIAction(doSomething);
var observable = reactiveFactory.CreateDriver(nonDriverObservable, onErrorJustReturn: whatever);
```

PR #3680, opened 2018-11-07
Rxfactor ForgotPassword Android UI

daividssilverio:
## What's this?
The android ui part of the `ForgotPasswordViewModel`'s RxFactory.

### Relationships
Child of #3677

## Why do we want this?
Because we want to remove Mvx bindings, yeah.

## How is it done?
Mvx bindings are removed and the VM observation is setup on the Activity.

### Why not another way?
Not enough 🍶.

### Side effects
Everything should stay the same.

## Review hints
Review by commit is suggested.
If you are testing this in your device, replace `AndroidClientHandler` with `HttpClientHandler` in this snippet of code in `Giskard\Startup\Setup.cs`:
```
var httpHandler = new AndroidClientHandler
            {
                AutomaticDecompression = DecompressionMethods.GZip | DecompressionMethods.Deflate
            };
```

## :squid: Permissions
@Odrakir only.


Toggl.Giskard/Activities/ForgotPasswordActivity.cs
```diff
@@ -32,12 +34,45 @@ protected override void OnCreate(Bundle bundle)
...
+            ViewModel.PasswordResetSuccessful
+                .Invert()
+                .Subscribe(resetPasswordButton.Rx().IsVisible())
+                .DisposedBy(DisposeBag);
+
+            resetPasswordButton.Rx()
+                .BindAction(ViewModel.Reset)
+                .DisposedBy(DisposeBag);
+
+            closeSubject
```
Odrakir: What is all this? The view is closed from the VM.

daividssilverio: Just so closing actions are forwarded to VM.
When the user presses the `<-` button, it will call the `Close` action in the VM instead of simply killing the host activity.
I added it as a way of stopping the user from leaving the screen while the action is being executed.

Odrakir: I see, in that case I think that logic lives better in the VM (as I believe it was before), that way it'll also work for iOS. I'm pushing a commit, I hope you don't mind.

daividssilverio: Go for it. Please merge it into your branch whenever you're done.

daividssilverio: And I like your way better.

PR #3689, opened 2018-11-07
Fix sign up with google

### LINKED ISSUE #3684: Active consent to Terms of Service when signing up with Google
🙊 🙈 🙉 

When signing up with Google, we’re not actively consenting to terms of service like we do with a sign up with email. This means we're not sending `tos_accepted` and `country_id` in the `/signup?app_name=toggl_mobile` request and the API returns BadRequest (luckily, otherwise we would not be complying with GDPR 🤔).

We need to:
- Present terms of service and privacy policy modal before starting sign up with Google
- Validate we have a selected country, also before starting sign up with Google
- Send `tos_accepted` and `country_id` in the request to Toggl API

juanlaube:
## What's this?
Fixes sign up with Google. It was broken since the changes required for GDPR, we were not sending consent of terms of service and country id when signing up.

This PR introduces two changes:
1. Actively requires user to accept ToS and select a country before starting the OAuth flow with Google.
2. Sends acceptance of ToS and country id in the sign up request.

### Relationships

Closes #3684

## How is it done?
By presenting the same modal with ToS and PP we use in sign up with email and adding the missing parameters in the request.

### Side effects
None.

## Review hints
Please test this in iOS and Android. I can provide you a patch with all the required changes to make this work in debug (not adding that here because it contains secrets).

## :squid: Permissions
🦑


Toggl.Foundation.MvvmCross/ViewModels/SignupViewModel.cs
```diff
@@ -305,14 +305,33 @@ private void onCompleted()
...
-        public void GoogleSignup()
+        public async Task GoogleSignup()
         {
+            var shakeTargets = ShakeTargets.None;
+            if (!countryId.HasValue)
+            {
+                shakeTargets |= ShakeTargets.Country;
+            }
+
+            if (shakeTargets != ShakeTargets.None)
```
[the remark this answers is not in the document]
juanlaube: This is what happens when you copy & paste 🙃

Toggl.Foundation.MvvmCross/ViewModels/SignupViewModel.cs
```diff
@@ -305,14 +305,33 @@ private void onCompleted()
...
+            {
+                shakeSubject.OnNext(shakeTargets);
+                return;
+            }
+
+            if (!termsOfServiceAccepted)
+                termsOfServiceAccepted = await navigationService.Navigate<bool>(typeof(TermsOfServiceViewModel));
+
+            if (!termsOfServiceAccepted)
```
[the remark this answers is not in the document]
juanlaube: These two ifs cannot be merged though. The first one is to present the ToS modal if the user did not accept the terms before, the second is to ensure the user accepted the terms before we continue.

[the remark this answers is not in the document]
simonrozsival: I see. I overlooked that you change the variable and only then you check it again. I almost think this would deserve a separate method. This seems like something we could easily break in some refactoring.

Toggl.Foundation.MvvmCross/ViewModels/SignupViewModel.cs
```diff
@@ -305,14 +305,27 @@ private void onCompleted()
...
+                return;
+
             if (isLoadingSubject.Value) return;
 
             isLoadingSubject.OnNext(true);
+            errorMessageSubject.OnNext("");
 
             signupDisposable = loginManager
-                .SignUpWithGoogle()
+                .SignUpWithGoogle(termsOfServiceAccepted, (int)countryId.Value)
```
simonrozsival: I think we should later refactor this to use `ICountry` and not to pass the ID directly. Separate PR though :)

Toggl.Foundation.MvvmCross/ViewModels/SignupViewModel.cs
```diff
@@ -305,14 +305,27 @@ private void onCompleted()
...
+        public async Task GoogleSignup()
         {
+            if (!countryId.HasValue)
+            {
+                shakeSubject.OnNext(ShakeTargets.Country);
+                return;
+            }
+
+            if (!termsOfServiceAccepted)
+                termsOfServiceAccepted = await navigationService.Navigate<bool>(typeof(TermsOfServiceViewModel));
```
simonrozsival:
```suggestion
                termsOfServiceAccepted = await navigationService.Navigate<TermsOfServiceViewModel, bool>();
```

simonrozsival: I guess that the tests would have to be updated too if you use it :)

Toggl.Foundation.MvvmCross/ViewModels/SignupViewModel.cs
```diff
@@ -305,14 +305,27 @@ private void onCompleted()
...
+
+            if (!termsOfServiceAccepted)
+                return;
+
             if (isLoadingSubject.Value) return;
 
             isLoadingSubject.OnNext(true);
+            errorMessageSubject.OnNext("");
```
simonrozsival: I don't really like `""` for empty strings, consider `string.Empty`:

```suggestion
            errorMessageSubject.OnNext(string.Empty);
```

Toggl.Foundation.Tests/MvvmCross/ViewModels/SignupViewModelTests.cs
```diff
@@ -241,134 +241,200 @@ public async Task SetsFailedToGetCountryToFalse()
...
+                    ViewModel.GoogleSignup().Wait();
 
-                ViewModel.GoogleSignup();
+                    LoginManager.Received().SignUpWithGoogle(true, Arg.Any<int>());
+                }
 
-                NavigationService.DidNotReceive().Navigate<MainViewModel>();
-            }
+                [Fact, LogIfTooSlow]
+                public void DoesNothingWhenThePageIsCurrentlyLoading()
```
simonrozsival: `async Task` so you don't have to use the blocking `.Wait()`

simonrozsival: the same applies to the other tests

Toggl.Foundation/Login/LoginManager.cs
```diff
@@ -159,11 +159,11 @@ private IObservable<IUser> signUp(Email email, Password password, bool termsAcce
...
 
 
-        private IObservable<ITogglDataSource> signUpWithGoogle(string googleToken)
+        private IObservable<ITogglDataSource> signUpWithGoogle(string googleToken, bool termsAccepted, int countryId)
         {
             return Observable
-                .Return(googleToken)
-                .SelectMany(apiFactory.CreateApiWith(Credentials.None).User.SignUpWithGoogle)
+                .Return(apiFactory.CreateApiWith(Credentials.None))
+                .SelectMany(api => api.User.SignUpWithGoogle(googleToken, termsAccepted, countryId))
```
simonrozsival: Let's get rid of the `Observable.Return`! Just simply create `var api = apiFactory.CreateApiWith(Credentials.None);` on a separate line and then start the observable creation with `api.User.SignUpWithGoogle(...` 👍

Toggl.Foundation.MvvmCross/ViewModels/SignupViewModel.cs
```diff
@@ -305,14 +305,27 @@ private void onCompleted()
             signupDisposable = null;
         }
 
-        public void GoogleSignup()
+        public async Task GoogleSignup()
```
simonrozsival: Shouldn't this be `UIAction`? 🤔 Maybe this while class needs a bit of refactoring.

Toggl.Foundation.MvvmCross/ViewModels/SignupViewModel.cs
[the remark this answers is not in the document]
juanlaube: Moved this into a new method: `requestAcceptanceOfTermsAndConditionsIfNeeded`

Toggl.Foundation.MvvmCross/ViewModels/SignupViewModel.cs
juanlaube: Yeah, separate PR is better.

Toggl.Foundation.MvvmCross/ViewModels/SignupViewModel.cs
juanlaube: I'd prefer to do this in another PR

Toggl.Foundation/Login/LoginManager.cs
```diff
@@ -161,9 +161,10 @@ private IObservable<IUser> signUp(Email email, Password password, bool termsAcce
 
         private IObservable<ITogglDataSource> signUpWithGoogle(string googleToken, bool termsAccepted, int countryId)
         {
-            return Observable
-                .Return(apiFactory.CreateApiWith(Credentials.None))
-                .SelectMany(api => api.User.SignUpWithGoogle(googleToken, termsAccepted, countryId))
+            return apiFactory
+                .CreateApiWith(Credentials.None)
+                .User
```
simonrozsival: 🤔 please split this in two lines - creating the API client + the request. This seems like `User` is already an observable and it was confusing for me.

Toggl.Foundation.Tests/MvvmCross/ViewModels/SignupViewModelTests.cs
```diff
@@ -173,7 +173,7 @@ public async Task NavigatesToSelectCountryViewModelPassingNullIfLocationApiFaile
 
                 await ViewModel.PickCountry();
 
-                await NavigationService.Received().Navigate<SelectCountryViewModel, long?, long?>(null);
+                NavigationService.Received().Navigate<SelectCountryViewModel, long?, long?>(null);
```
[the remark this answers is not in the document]
juanlaube: Was not really needed 🤷‍♂️

PR #3694, opened 2018-11-08
Subscribe async

Odrakir:
## What's this?
Renames a specific override of `Subscribe` to `SubscribeAsync` and changes its implementation.

## Why do we want this?
If we check the old implementation of this override, we can see that the `Subscribe` method it's using internally takes an `Action<Unit>`, this, apparently can cause some problems if an exception occurs in the `onNext` Action. I've tested that and you can't catch that exception.

Also, `Subscribe` will not await for that lambda, so new values may come through (there'll be no back pressure) and that may result in uncontrolled concurrency.

To me it just seemed strange to mix Rx and async/await in that way, given that Rx is mainly synchronous. After some search, it seems like my suspicion was right. I tried to write some tests to prove this, but I've only been able to prove the unhandled exception thing. Any ideas about the rest are welcome.

There's more info in [this article](https://github.com/dotnet/reactive/issues/459).

## How is it done?
Instead one the funky use of async/await inside `Subscribe`, we'll first `Select` transforming our `Task` into an `Observable` and then `Concat` so it flattens back all events, in the same order, before subscribing.

There's another possibility described in the article, which is to use `Merge` instead of `Concat`, but that's for concurrency and in case we need that it can be done some other way, this is just about simplifying this operator.

## Also
Some of the usages of this operator could (and should) be replaced with `UIAction`, but that'll be another PR.

## :squid: Permissions
If @willsb approves.

### DIFF
--- a/Toggl.Foundation.MvvmCross/Extensions/ObservableExtensions.cs
+++ b/Toggl.Foundation.MvvmCross/Extensions/ObservableExtensions.cs
@@ -16,8 +16,11 @@ namespace Toggl.Foundation.MvvmCross.Extensions
         public static IDisposable VoidSubscribe<T>(this IObservable<T> observable, Action onNext)
             => observable.Subscribe((T _) => onNext());
 
-        public static IDisposable Subscribe(this IObservable<Unit> observable, Func<Task> onNext)
-            => observable.Subscribe(async (Unit _) => await onNext());
+        public static IDisposable SubscribeAsync(this IObservable<Unit> observable, Func<Task> onNextAsync)
+            => observable
+                .Select(_ => Observable.FromAsync(onNextAsync))
+                .Concat()
+                .Subscribe();
 
         public static IObservable<T> DeferAndThrowIfPermissionNotGranted<T>(this IObservable<bool> permissionGrantedObservable, Func<IObservable<T>> implementation)
             => Observable.DeferAsync(async cancellationToken =>
--- a/Toggl.Daneel/ViewControllers/LoginViewController.cs
+++ b/Toggl.Daneel/ViewControllers/LoginViewController.cs
@@ -98,7 +98,7 @@ namespace Toggl.Daneel.ViewControllers
 
             //Commands
             SignupCard.Rx().Tap()
-                .Subscribe(ViewModel.Signup)
+                .SubscribeAsync(ViewModel.Signup)
                 .DisposedBy(DisposeBag);
 
             LoginButton.Rx().Tap()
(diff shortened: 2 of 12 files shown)


Toggl.Foundation.MvvmCross/Extensions/ObservableExtensions.cs
```diff
@@ -16,8 +16,11 @@ public static IObservable<T> AsDriver<T>(this IObservable<T> observable, ISchedu
         public static IDisposable VoidSubscribe<T>(this IObservable<T> observable, Action onNext)
             => observable.Subscribe((T _) => onNext());
 
-        public static IDisposable Subscribe(this IObservable<Unit> observable, Func<Task> onNext)
-            => observable.Subscribe(async (Unit _) => await onNext());
+        public static IDisposable SubscribeAsync(this IObservable<Unit> observable, Func<Task> onNextAsync)
+            => observable
+                .Select(_ => Observable.FromAsync(onNextAsync))
+                .Concat()
```
simonrozsival: What does this `.Concat` do? 🤔 Wouldn't `SelectMany` work better instead of the `Select` + `Concat`?

simonrozsival:
Othewise I like that we have an explicit way subscribing with async handlers. I didn't really think about this being a problem before, so 👍 for the PR!

willsb:
I'm not so sure about this. While I like the extension method, after hearing your explanation I just wanna 🔥 the existing instances of it. Most of them seem to be methods not using the `RxAction` class, do I'd like to check this again once they're gone

Toggl.Foundation.MvvmCross/Extensions/ObservableExtensions.cs
Odrakir: I think this is probably because the code I took as an example had another operator using `merge` instead of `concat` so tasks could be concurrent.

I'm not sure if it has any other implications... 🤔 probably not. Why do you think `SelectMany` is better? Just because of clarity?

Odrakir:
I agree with @heytherewill still, I think this is something we'll use even if we switch all the code we can to `RxAction`.

But anyway, I'll try to add some tests and we can come back to it when we rxfactor some more.

Toggl.Foundation.MvvmCross/Extensions/ObservableExtensions.cs
Odrakir: Probably `selectMany` uses merge... Not that it's going to be much of a difference in our case: https://fernandocejas.com/2015/01/11/rxjava-observable-tranformation-concatmap-vs-flatmap/

heytherewill:
What's the status on this? I think we should open an issue to remove the `Subscribe(Task)` method and close this PR, but I'd like to hear what @Odrakir thinks of it

Odrakir:
I failed to write tests that prove that this fails. Not because it works fine, but because writing such tests is hard. Together with the fact that you said we should remove those calls, that's why this is on hold.

I'm okay with closing this and creating that other issue.

Odrakir:
Superseded by #3929

PR #3697, opened 2018-11-08
Show task name in start view

### LINKED ISSUE #3685: Task not shown in start time entry view
### Steps to reproduce
1. Press the green start button
2. Enter `@` to start searching projects
3. Find a project with a task
4. Expand that project
5. Select the task

The project **and the task** got added, however, the task is just not showing in the text field. If you press the green checkmark button, you'll see that the running entry has the task.

The task should probably be displayed in the project token the same way we show it in other places: `project: task`.

arturs105:
## What's this?
This PR makes tasks visible in the text field in start view the same way it's done in the rest of the app: `project: task`.

### Relationships

Closes #3685

## :squid: Permissions
🦑


Toggl.Daneel/Autocomplete/AutocompleteExtensions.cs
```diff
@@ -166,11 +167,15 @@ private static NSMutableAttributedString AsAttributedString(this TextSpan textSp
         private static NSMutableAttributedString AsAttributedString(this ProjectSpan projectSpan)
         {
             var projectColor = MvxColor.ParseHexString(projectSpan.ProjectColor).ToNativeColor();
-            var projectName = new NSAttributedString(projectSpan.ProjectName.TruncatedAt(maxTextLength), new UIStringAttributes
-            {
-                Font = tokenFont,
-                ForegroundColor = projectColor
-            });
+
+            var projectNameStringBuilder = new StringBuilder(ProjectName);
```
simonrozsival: I don't think you need a string builder for single concatenation - I would use `+=` in this case.

simonrozsival:
Otherwise it works pretty well 👍

Toggl.Daneel/Autocomplete/AutocompleteExtensions.cs
```diff
@@ -168,12 +168,12 @@ private static NSMutableAttributedString AsAttributedString(this ProjectSpan pro
         {
             var projectColor = MvxColor.ParseHexString(projectSpan.ProjectColor).ToNativeColor();
 
-            var projectNameStringBuilder = new StringBuilder(ProjectName);
+            var projectNameString = projectSpan.ProjectName;
             if (!string.IsNullOrEmpty(projectSpan.TaskName))
```
simonrozsival: This could be a ternary operator, but that is just nitpicking :)

arturs105: I though about that, but i preferred this way.

PR #3699, opened 2018-11-08
🚧 Cleanup BaseApi

### LINKED ISSUE #3741: Finish BaseAPI refactor
There's currently a [PR](https://github.com/toggl/mobileapp/pull/3699) that cleans the code in `BaseAPI`, but it can't be merged because it has some kind of deadlock in tests.

We should take a deeper look into this code and fix it once and for all.

A couple of things worth looking into:
- The way we use `Func<string, Task<T>> process` to deserialize responses could be streamlined with Observables
- The multiple overloads for very similar usages 
- Renaming `CreateListObservable`

Odrakir:
## What's this?
This is started just so I could rename `CreateObservable` to `SendRequest`, but then I also did some refactoring to simplify how the request is made.

Closes #3741 

## Why do we want this?
I think the code is a little bit clearer and also we now don't manually create an Observable, which is always delicate, mostly if resources are not freed up properly.

## How is it done?
The renaming is a no-brainer, for the refactoring I use the `ToObservable()` operator instead of manually creating an observable with `async` inside. Also, there's no need to catch and forward exceptions inside an rx operator, every `throw` will result in an error event sent.

## :squid: Permissions
If you like it, squash it.


Toggl.Ultrawave/ApiClients/BaseApi.cs
```diff
@@ -51,57 +52,43 @@ protected BaseApi(IApiClient apiClient, IJsonSerializer serializer, Credentials
...
-                catch (Exception e)
-                {
-                    observer.OnError(e);
-                    return;
-                }
-
-                observer.OnNext(data);
-                observer.OnCompleted();
-            });
+                    return throwIfRequestFailed(request, response, headerList).ToObservable()
```
simonrozsival: Suggestion:

```cs
private IObservable<T> sendRequest<T>(Endpoint endpoint, IEnumerable<HttpHeader> headers, string body, Func<string, Task<T>> process)
{
    var headerList = headers as IList<HttpHeader> ?? headers.ToList();
    return sendRequestAsync(endpoint, headersList, body, process).AsObservable();
}

private async Task<T> sendRequestAsync<T>(Endpoint endpoint, IList<HttpHeader> headers, string body, Func<string, Task<T>> process)
{
    var request = new Request(body, endpoint.Url, headers endpoint.Method);
    var response = await send(request);
    await throwIfRequestFailed(request, response, headers);
    return await processResponseData(request, response, process);
}
```

juanlaube: I like this suggestion!

sofokli1: 👍

Odrakir: I agree, in this specific case where you have to carry around some params it makes sense to fall back to async/await.

Odrakir:
I'll add it to the issue

Toggl.Ultrawave/ApiClients/BaseApi.cs
```diff
@@ -36,83 +38,80 @@ protected BaseApi(IApiClient apiClient, IJsonSerializer serializer, Credentials
...
+        protected IObservable<List<TInterface>> SendRequest<TModel, TInterface>(Endpoint endpoint,
+            HttpHeader header, string body = "")
             where TModel : class, TInterface
-            => CreateListObservable<TModel, TInterface>(endpoint, new[] { header }, body);
+            => SendRequest<TModel, TInterface>(endpoint, new[] { header }, body);
 
 
-        protected IObservable<List<TInterface>> CreateListObservable<TModel, TInterface>(Endpoint endpoint, IEnumerable<HttpHeader> headers, string body = "")
+        protected IObservable<List<TInterface>> SendRequest<TModel, TInterface>(Endpoint endpoint,
+            IEnumerable<HttpHeader> headers, string body = "")
```
simonrozsival: Unnecessary change of indentation.

Odrakir:
I agree, the comments are just an attempt to clearly identify each override. But this is not the ideal solution. I've been trying to find a nicer way to do it for a while, but I don't get along with C# generics.

In my opinion this class shouldn't do any deserialization, because that's what complicates things with all the overrides, `SendRequest` should probably just return a generic type. And it should take, somehow the method to map the results to that type.

In Swift I used a `Resource` object that encapsulates the network request and the mapping function... Just an idea: https://github.com/Odrakir/Porrua/blob/master/Porrua/Source/WebService.swift

PR #3700, opened 2018-11-08
Sanic boosters: Run suggestions query in the background

### LINKED ISSUE #3418: Improve main thread marshalling when querying for start TE suggestions
This might be a simple way of getting performance benefits in that view. Check with @simonrozsival for he has a lot of knowledge on how that works

daividssilverio:
## What's this?
This makes our suggestions queries run in a background thread.

### Relationships
Closes #3418

## Why do we want this?
Because the queries running in the main thread are making the screen laggy when typing your query.

## How is it done?
All it's done is a simple `.ObserveOn(schedulerProvider.BackgroundScheduler)` from the point we want to run in a background thread when preparing the `querySubject`.

### Why not another way?
We can improve the queries themselves at some point, but this will help reduce the laggy feel in that screen. Please suggest something if you can.

### Side effects
I couldn't reproduce my first impression that a very slow first query could affect a faster query after it.🤔
I had to update the tests to consider the TestScheduler as well.

## Review hints
Try for yourself. Run on develop then run from this branch.
Creating A LOT of time entries and projects will make you feel the difference.
@simonrozsival posted a script internally to help with that (be careful though with that though).

I've added Simon as a reviewer, but the more the merrier.

## :squid: Permissions
@daividssilverio @willsb and @dshr 🦑 please


simonrozsival:
No lag whatsoever. I don't even think we need throttling. We can consider it later when he gather some more data after the fix is released.

daividssilverio:
Throttling would help reducing the amount of db reads while the user is typing.

PR #3701, opened 2018-11-08
🆙Gaining access to workspace

### LINKED ISSUE #3395: Create a complex test to verify that gaining access to a workspace works correctly
Similar problem as #3394 - we should first try implementing that one and when we are happy with the approach, we should adapt it for this issue.

- [x] create a test setup (ghost workspace with some unsynced/unsyncable projects, tags, tasks, clients, time entries)
- [x] prepare/mock the data which will be pulled from the server
- [x] initiate sync and wait for it to complete
- [x] verify that the state of the internal storage matches expectations

sofokli1:
## What's this?
Adds some sync tests to test gaining and regaining access to a workspace

### Relationships
Closes #3395

## Why do we want this?
The use of tests should be obvious to all of us.

## How is it done?
I created a 3 tests based on the `BaseComplexSyncTest`.

### Why not another way?
This is the way.

### Side effects
none

## Review hints
There is still an issue with the `tasks`.

## :squid: Permissions
:squid: when all green

### DIFF
--- a/Toggl.Foundation.Sync.Tests/GainingAccessToWorkspace/FetchesEntitiesOlderThanSinceDate.cs
+++ b/Toggl.Foundation.Sync.Tests/GainingAccessToWorkspace/FetchesEntitiesOlderThanSinceDate.cs
@@ -128,6 +128,14 @@ namespace Toggl.Foundation.Sync.Tests.GainingAccessToWorkspace
                         ProjectId = regainedAccessProject.Id,
                         TagIds = new long[] { regainedAccessTag2.Id },
                         SyncStatus = SyncStatus.SyncNeeded }
+                },
+                sinceParameters: new Dictionary<Type, DateTimeOffset?>
+                {
+                    [typeof(IDatabaseWorkspace)] = DateTimeOffset.Now.AddDays(1),
+                    [typeof(IDatabaseClient)] = DateTimeOffset.Now.AddDays(1),
+                    [typeof(IDatabaseProject)] = DateTimeOffset.Now.AddDays(1),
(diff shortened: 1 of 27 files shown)


Toggl.Foundation.Sync.Tests/GainingAccessToWorkspace/FetchesEntitiesOlderThanSinceDate.cs
```diff
@@ -0,0 +1,170 @@
...
+using Toggl.PrimeRadiant.Models;
+
+namespace Toggl.Foundation.Sync.Tests.GainingAccessToWorkspace
+{
+    public sealed class FetchesEntitiesOlderThanSinceDate : BaseComplexSyncTest
+    {
+        protected override ServerState ArrangeServerState(ServerState initialServerState)
+            => initialServerState.With(
+                clients: new[] { new MockClient { Id = -1, WorkspaceId = -2, Name = "c1" } },
+                tags: new[] {
```
simonrozsival: No Egyptian please! 😱

simonrozsival:
`sh build.sh --target Tests.Sync`: https://app.bitrise.io/build/5f88167a4a953ef0

Toggl.Foundation.Sync.Tests/GainingAccessToWorkspace/FetchesEntitiesOlderThanSinceDate.cs
sofokli1: my main issue with the tests is the lack of `tasks` but those could be fixed at a later date.

PR #3711, opened 2018-11-09
Daneel 1.14 - Rigel

juanlaube:
## Changelog
```
- Added 3D touch shortcut to open the calendar view
- Added the task name when starting time entries

- Improved app startup time
- Improved suggestions UI when starting time entries
- Improved performance in the time entries log UI
- Improved syncing when losing and gaining access to workspaces

- Fixed a bug when user's account doesn't have a default workspace
- Fixed sign up with google
- Fixed crashes when using Estonian as the preferred language
- Fixed calendar view sometimes showing wrong day
```

## Fun fact
Rigel is generally the seventh-brightest star in the night sky and the brightest star in the constellation of Orion—though periodically it is outshone within the constellation by the variable Betelgeuse.

## History of changes
https://github.com/toggl/mobileapp/compare/daneel-1.13.1-2...rigel


ApocalypticFish:
Depending on Monday's workload testing should be completed by EOD Monday

PR #3718, opened 2018-11-14
Fail tests if sync graph has loose ends

### LINKED ISSUE #3650: Prevent unintentional dead ends in sync transitions
We should prevent our sync states which are connected to the sync graph to have results which aren't connected to any follow-up state other than those where it is the end of the sync. We already do something similar with invalid transitions.

- [x] Add a `DeadEndState` with no _public_ result
- [x] Connect the new state to all places in the graph where a dead end is intentional
- [x] Update sync graph generator to highlight this state in a similar way it does for entry points and invalid transitions
- [x] Update sync graph generator to make unintentional dead ends differently - it must be apparent that it is a bug
- [x] Add unit test which looks for any unintentional dead ends and fails if there is any

amulware:
## What's this?
This adds one(1) new test that checks that the sync graph has no loose ends.

### Relationships
Closes #3650

Depends on #3717

## Why do we want this?
So that we don't accidentally forget to hook up all sync states. This does NOT ensure that the graph is hooked up correctly, but we now have to mark dead ends that we intend to be dead ends on purpose, instead of having them be only inferred (this is done in #3717 which have to be merged first).

## How is it done?
I used some of the code from the sync graph diagram renderer to setup the sync graph in a new unit test in `Toggl.Foundation.Tests.Sync.SyncGraphTests`. Instead of actually configuring transitions, this just records all registered ones.

Then I go over all known states, use reflection to get all public state result properties, and then ensure that all those state results are hooked up to a transition. If I find one that isn't, I throw an Exception, failing the test.

### Why not another way?
This is pretty straightforward. We use the actual app's sync graph factory, which is as close as it can get.

One thing that could be done is important the mock transition configurator from the sync diagram renderer. However, I thought that duplicating this code would be easier, and more robust in case we want to change one of them in the future.

### Side effects
We cannot merge partially set up sync graphs.

## Review hints
It's pretty straightforward if you're used to reflection and how our transition configuration works.

## :squid: Permissions
Me, me, me.


Toggl.Foundation.Tests/Sync/SyncGraphTests.cs
```diff
@@ -0,0 +1,112 @@
...
+                    .ToList();
+            }
+
+            private static bool isStateResultProperty(PropertyInfo p)
+            {
+                return typeof(IStateResult).IsAssignableFrom(p.PropertyType);
+            }
+        }
+
+        private sealed class Configurator : ITransitionConfigurator
```
simonrozsival: Let's move this to a separate file and make it a regular helper class.

amulware:
I did as you asked by moving all the logic into the configurator class and testing the exposed method. The tests are verbose and repetitive and not at all exhaustive, but they cover a bunch of cases, which I think is sufficient.

If you have an idea for turning the tests into parameterized tests, that would be welcome, but I don't see an easy way.

amulware:
@simonrozsival 	Here you go: https://app.bitrise.io/build/8a9ecf41e8ce6ab1

That being said, this does not actually change any sync code.

amulware:
Why wouldn't these tests be run with the default unit tests? They are just in `Toggl.Foundation.Tests`.

simonrozsival:
`namespace Toggl.Foundation.Tests.Sync` vs `namespace Toggl.Foundation.Sync.Tests` - I confused those two.

PR #3719, opened 2018-11-14
Compress resources

heytherewill:
## What's this?
This passes ImageOptim on all of our resources and deletes the extra `bg_noise_login.png`. We just need it once, since the idea of that resource is to be stretched anyways. This will save us some good MB on each individual APK.

## :squid: Permissions
Anyone, anytime


amulware:
Can we figure out a way of always only committing compressed resources in the future?

amulware:
Another question: I remember we were going to replace all the pngs by svgs anyways. Is this still worth doing in the meantime?

khoi:
Are u using a lossy compression? 

I tried to compress `Toggl.Giskard/Resources/drawable-mdpi/timer.png` using default ImageOptim's settings (lossless) and only get 37.8% size reduction but yours is 62% 

screen shot 2018-11-19 at 10 58 00

screen shot 2018-11-19 at 10 59 19

sofokli1:
No compression for iSO or are those already compressed?

dshr:
> No compression for iSO or are those already compressed?

Yup, XCode compresses PNGs automatically. We might be able to win some MBs back by going vector, but that's a big change indeed.

juanlaube:
Vector graphics in iOS are a lie though, Xcode renders the PDFs into different PNGs under the hood.

dshr:
🤯

juanlaube:
@dshr surprise! 😄
If we want vector graphics in iOS, we can use PaintCode ($$$) to convert svg to code (not sure about Xamarin support though), that can save a lot of MBs.

dshr:
@juanlaube it does say that it can output C#, so might support Xamarin. I wonder how hard would it be to write our own SVG parser though 🤔

juanlaube:
I think @khoi is right, those compression ratios are too high. I tried to compress some assets ImageOptim and I don't get the same results.
Image quality in your assets is good though, I don't see any difference (maybe I'm blind? 🤔)

heytherewill:
@amulware 
RE Ensuring compression, I thought about it and I'm not sure that's possible :/
RE if this is worth: Yes, it is! svg replacement is gonna be messy and this is simple and clean (and we can reap benefits as of RN

@khoi and @juanlaube 
These are the settings I used:
image

The second page does not have lossy minification enabled. Maybe you guys are not stripping metadata? I think the assets look just fine, but let's double check the settings anyways

khoi:
@heytherewill I have the exact same settings, and my outputs are not as small as yours. Can you screenshot the ImageOptim app compressing `Toggl.Giskard/Resources/drawable-mdpi/timer.png`?

PR #3738, opened 2018-11-20
BaseRecyclerAdapter Performance Improvements with DiffUtils

### LINKED ISSUE #3735: BaseRecyclerAdapter Performance Improvements with DiffUtils
The `BaseRecyclerAdapter` is currently rebuilding the whole Recycler View using `NotifyDataSetChanged()` everytime `Items` is updated.

During RxFactors, this class will be used widely across the app for most of the `Select*Activity.cs`. And everytime users type in a character, the recycler view has to be reloaded to show the suggestions based on the input text. 
 
That is not the most performant behavior, introducing `DiffUtil` into `BaseRecyclerAdapter` like the one we have in `ReactiveSectionedRecyclerAdapter` will provide a noticeable performance boost, especially for users that have a large dataset.

khoi:
## What's this?
This PR improves the performance of `BaseRecyclerAdapter` by introducing `DiffUtil`.

### Relationships

Closes #3735

## Why do we want this?
Sanic

## How is it done?
Leveraging [DiffUtil]( https://developer.android.com/reference/android/support/v7/util/DiffUtil )

### Why not another way?
`ListAdapter` could also be used, but that'll require more code at the call site to implement the Diff callbacks.

### Side effects

~The animation when the data reloaded changed (the items blink)~

None

## Review hints

None beside testing it.

## :squid: Permissions
🦑


Toggl.Foundation.MvvmCross/ViewModels/Selectable/SelectableWorkspaceViewModel.cs
```diff
@@ -21,5 +22,30 @@ public SelectableWorkspaceViewModel(IThreadSafeWorkspace workspace, bool selecte
...
+        }
+
+        public override bool Equals(object obj)
+        {
+            if (ReferenceEquals(null, obj)) return false;
+            if (ReferenceEquals(this, obj)) return true;
+            return obj is SelectableWorkspaceViewModel other && Equals(other);
+        }
+
+        public override int GetHashCode()
```
simonrozsival: Use the `Toggl.Multivac.HashCode` helper.

Toggl.Giskard/Adapters/BaseRecyclerAdapter.cs
```diff
@@ -63,9 +64,36 @@ public virtual T GetItem(int viewPosition)
 
         protected virtual void SetItems(IList<T> items)
         {
+            DiffUtil.DiffResult diffResult = DiffUtil.CalculateDiff(new BaseDiffCallBack(this.items, items));
```
simonrozsival: Can you use `var` here?

Toggl.Giskard/Adapters/BaseRecyclerAdapter.cs
```diff
@@ -63,9 +64,36 @@ public virtual T GetItem(int viewPosition)
...
+            public override bool AreContentsTheSame(int oldItemPosition, int newItemPosition)
+            {
+                var oldItem = oldItems[oldItemPosition];
+                var newItem = newItems[newItemPosition];
+                return oldItem.Equals(newItem);
+            }
+
+            public override bool AreItemsTheSame(int oldItemPosition, int newItemPosition)
+            {
+                return ReferenceEquals(oldItems[oldItemPosition], newItems[newItemPosition]);
```
daividssilverio: `ReferenceEquals` is ok... but ideally, this should work compare only Ids if the items are identifiable with Ids.

Toggl.Giskard/Activities/SelectClientActivity.cs
```diff
@@ -55,6 +55,7 @@ private void setupLayoutManager(SelectClientRecyclerAdapter adapter)
             layoutManager.InitialPrefetchItemCount = 4;
             selectClientRecyclerView.SetLayoutManager(layoutManager);
             selectClientRecyclerView.SetAdapter(adapter);
+            selectClientRecyclerView.SetItemAnimator(null);
```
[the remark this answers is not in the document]
khoi: For some reason, the default animator has this weird blinking effect. setting this to `null` achieves the same behavior that we had before using DiffUtil

Toggl.Giskard/Adapters/BaseRecyclerAdapter.cs
khoi: I've introduced the `IDiffable` interface to solve this.

daividssilverio:
I'm sorry, but `notifyDataSetChanged` is the one that could be making the list blink.
Are you sure this isn't caused by having too many updates running at the same time? Are you sure this is still necessary after you've added the queueing mechanism?

I'm not comfortable with this, DiffUtil was supposed to make the list animate nicely, not the opposite.

Toggl.Giskard/Adapters/BaseRecyclerAdapter.cs
```diff
@@ -58,14 +67,75 @@ public override void OnBindViewHolder(RecyclerView.ViewHolder holder, int positi
 
         public override int ItemCount => items.Count;
 
+        public override long GetItemId(int position)
+        {
+            return items[position].Identifier.GetHashCode();
```
daividssilverio: Wouldn't it be simpler to make Identifier a long?

Toggl.Foundation.MvvmCross/ViewModels/Selectable/SelectableColorViewModel.cs
```diff
@@ -1,10 +1,12 @@
...
+﻿using System;
+using MvvmCross.UI;
+using Toggl.Foundation.MvvmCross.Interfaces;
 using Toggl.Multivac;
 
 namespace Toggl.Foundation.MvvmCross.ViewModels
 {
     [Preserve(AllMembers = true)]
-    public sealed class SelectableColorViewModel
+    public sealed class SelectableColorViewModel : IEquatable<SelectableColorViewModel>, IDiffable<SelectableColorViewModel>
```
heytherewill: Why explicitly mark this as implementing both interfaces?

Toggl.Foundation.MvvmCross/ViewModels/Selectable/SelectableWorkspaceViewModel.cs
```diff
@@ -1,11 +1,13 @@
...
+using MvvmCross.ViewModels;
 using Toggl.Foundation.Models.Interfaces;
+using Toggl.Foundation.MvvmCross.Interfaces;
 using Toggl.Multivac;
 
 namespace Toggl.Foundation.MvvmCross.ViewModels
 {
     [Preserve(AllMembers = true)]
-    public sealed class SelectableWorkspaceViewModel : MvxNotifyPropertyChanged
+    public sealed class SelectableWorkspaceViewModel : MvxNotifyPropertyChanged, IEquatable<SelectableWorkspaceViewModel>, IDiffable<SelectableWorkspaceViewModel>
```
heytherewill: Same question here

Toggl.Giskard/Extensions/Reactive/RecyclerAdapterExtensions.cs
```diff
@@ -8,6 +9,6 @@ namespace Toggl.Giskard.Extensions.Reactive
     public static class RecyclerAdapterExtensions
     {
         public static Action<IList<T>> Items<T>(this IReactive<BaseRecyclerAdapter<T>> reactive)
-            => collection => reactive.Base.Items = collection;
+            where T : IDiffable<T> => collection => reactive.Base.Items = collection;
```
simonrozsival: Put the `=>` on a new line. It is weird to see constraints and the body of the method on the same line.

Toggl.Giskard/Adapters/SimpleAdapter.cs
```diff
@@ -1,12 +1,13 @@
...
 using Android.Runtime;
 using Android.Views;
+using Toggl.Foundation.MvvmCross.Interfaces;
 using Toggl.Giskard.ViewHolders;
 using Toggl.Multivac;
 
 namespace Toggl.Giskard.Adapters
 {
-    public sealed class SimpleAdapter<T> : BaseRecyclerAdapter<T>
+    public sealed class SimpleAdapter<T> : BaseRecyclerAdapter<T> where T : IDiffable<T>
```
simonrozsival: Consider moving the constraint to a new line to make the content narrower and a bit more readable.

Toggl.Giskard/Adapters/BaseRecyclerAdapter.cs
```diff
@@ -58,14 +67,75 @@ public override void OnBindViewHolder(RecyclerView.ViewHolder holder, int positi
...
+            public override bool AreContentsTheSame(int oldItemPosition, int newItemPosition)
+            {
+                var oldItem = oldItems[oldItemPosition];
+                var newItem = newItems[newItemPosition];
+                return oldItem.Equals(newItem);
+            }
+
+            public override bool AreItemsTheSame(int oldItemPosition, int newItemPosition)
+            {
+                return string.Equals(oldItems[oldItemPosition].Identifier, newItems[newItemPosition].Identifier);
```
simonrozsival: `string.Equals`? Aren't you comparing two `long`s? Simple `==` would be better and faster (in case we are diffing "long" lists, it might be worth micro-optimizing a bit).

Toggl.Giskard/Adapters/BaseRecyclerAdapter.cs
```diff
@@ -1,23 +1,31 @@
...
 using Android.Views;
 using Toggl.Giskard.ViewHolders;
+using Java.Lang;
+using Toggl.Foundation.MvvmCross.Interfaces;
+using Handler = Android.OS.Handler;
 
 namespace Toggl.Giskard.Adapters
 {
-    public abstract class BaseRecyclerAdapter<T> : RecyclerView.Adapter
+    public abstract class BaseRecyclerAdapter<T> : RecyclerView.Adapter where T: IDiffable<T>
```
simonrozsival: I suggest a new line here 👍

Toggl.Giskard/Adapters/BaseRecyclerAdapter.cs
```diff
@@ -58,14 +67,75 @@ public override void OnBindViewHolder(RecyclerView.ViewHolder holder, int positi
...
+        }
+
         public virtual T GetItem(int viewPosition)
             => items[viewPosition];
 
-        protected virtual void SetItems(IList<T> items)
+        protected virtual void SetItems(IList<T> newItems)
+        {
+            lastestUpdates.Enqueue(newItems);
+            if (lastestUpdates.Count > 1)
```
simonrozsival: In this case the `return` seems unnecessary. It would be cleaner to invert the condition and call `updateItemsInternal(newItems)` from the if block.

Toggl.Giskard/Adapters/BaseRecyclerAdapter.cs
```diff
@@ -58,14 +67,75 @@ public override void OnBindViewHolder(RecyclerView.ViewHolder holder, int positi
...
+                return;
+            }
+            updateItemsInternal(newItems);
+        }
+
+        private void updateItemsInternal(IList<T> newItems)
+        {
+            var oldItems = items;
+            var handler = new Handler();
+            new Thread(() =>
```
simonrozsival: Creating a new thread seems weird. I don't think we should explicitly start a new thread. Let's use TPL and let the threadpool handle the management of threads. Simply replace `new Thread(() =>` with `Task.Run(() =>` and remove the `.Start()` at the end.

simonrozsival: (link to the docs: https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.run?view=netframework-4.7.2#System_Threading_Tasks_Task_Run_System_Action_)

Toggl.Giskard/Adapters/BaseRecyclerAdapter.cs
```diff
@@ -1,23 +1,31 @@
...
-    public abstract class BaseRecyclerAdapter<T> : RecyclerView.Adapter
+    public abstract class BaseRecyclerAdapter<T> : RecyclerView.Adapter where T: IDiffable<T>
     {
         private Subject<T> itemTapSubject = new Subject<T>();
         public IObservable<T> ItemTapObservable => itemTapSubject.AsObservable();
 
         public Func<T, Task> OnItemTapped { get; set; }
 
         private IList<T> items = new List<T>();
+        private ConcurrentQueue<IList<T>> lastestUpdates = new ConcurrentQueue<IList<T>>();
```
simonrozsival: Isn't this a typo? Shouldn't it be `latestUpdates` or `lastUpdates`?

Toggl.Giskard/Adapters/BaseRecyclerAdapter.cs
```diff
@@ -58,14 +67,75 @@ public override void OnBindViewHolder(RecyclerView.ViewHolder holder, int positi
...
+            new Thread(() =>
+            {
+                var diffResult = DiffUtil.CalculateDiff(new BaseDiffCallBack(oldItems, newItems));
+                handler.Post(() => dispatchUpdates(newItems, diffResult));
+            }).Start();
+        }
+
+        private void dispatchUpdates(IList<T> items, DiffUtil.DiffResult diffResult)
         {
+            lastestUpdates.TryDequeue(out IList<T> _);
```
simonrozsival: I don't understand this line. Are you trying to clear the queue? If yes, then you should remove everything, not just the item on the top of the queue. If not, why are you removing the first item no matter what it is?

Toggl.Giskard/Adapters/BaseRecyclerAdapter.cs
```diff
@@ -58,14 +67,75 @@ public override void OnBindViewHolder(RecyclerView.ViewHolder holder, int positi
...
             this.items = items;
+            diffResult.DispatchUpdatesTo(this);
+            if (lastestUpdates.Count > 0)
```
simonrozsival: I don't think you need to check if `Count > 0` - if it is empty, `TryDequeue` will return `false`.

Toggl.Giskard/Adapters/BaseRecyclerAdapter.cs
simonrozsival: Why not return the `Identifier` directly? Why do you need hash code of `long`?

Toggl.Foundation.MvvmCross/Interfaces/IDiffable.cs
```diff
@@ -0,0 +1,9 @@
+using System;
+
+namespace Toggl.Foundation.MvvmCross.Interfaces
+{
+    public interface IDiffable<T> : IEquatable<T>
+    {
+        long Identifier { get; }
+    }
+}
```
simonrozsival: `\n` 🙌 :)

Toggl.Giskard/Adapters/BaseRecyclerAdapter.cs
khoi: I changed the `Identifier` from `string` to `long` recently and missed this comparison

Toggl.Giskard/Adapters/BaseRecyclerAdapter.cs
daividssilverio: ☝️

Toggl.Giskard/Adapters/BaseRecyclerAdapter.cs
daividssilverio: We could add your suggestion to the MainRecyclerAdapter later as well, that's how I did in there.

khoi:
@simonrozsival @daividssilverio I've changed the implementation of the Queue a bit, this should be easier to understand now. Let me know if there is something unclear.

Toggl.Giskard/Adapters/BaseRecyclerAdapter.cs
```diff
@@ -69,20 +70,19 @@ public override void OnBindViewHolder(RecyclerView.ViewHolder holder, int positi
...
 
         public virtual T GetItem(int viewPosition)
             => items[viewPosition];
 
         protected virtual void SetItems(IList<T> newItems)
         {
-            lastestUpdates.Enqueue(newItems);
-            if (lastestUpdates.Count > 1)
+            latestUpdates.Enqueue(newItems);
+            if (latestUpdates.Count == 1)
```
simonrozsival: Previously the `updateItemsInternal` was called even for `Count == 0`, is it OK to ignore an empty list of updates?

Toggl.Giskard/Adapters/BaseRecyclerAdapter.cs
```diff
@@ -1,23 +1,35 @@
...
 using Android.Support.V7.Widget;
 using Android.Views;
 using Toggl.Giskard.ViewHolders;
+using Java.Lang;
+using Toggl.Foundation.MvvmCross.Interfaces;
+using Handler = Android.OS.Handler;
 
 namespace Toggl.Giskard.Adapters
 {
+
```
simonrozsival: 👀

Toggl.Giskard/Adapters/BaseRecyclerAdapter.cs
```diff
@@ -1,23 +1,35 @@
...
-        private Subject<T> itemTapSubject = new Subject<T>();
         public IObservable<T> ItemTapObservable => itemTapSubject.AsObservable();
 
         public Func<T, Task> OnItemTapped { get; set; }
 
+        private Subject<T> itemTapSubject = new Subject<T>();
+
         private IList<T> items = new List<T>();
+
+        private FixedSizedConcurrentQueue<IList<T>> updateQueue = new FixedSizedConcurrentQueue<IList<T>>(1);
```
simonrozsival: Wait. I can't see any use for a queue of length `1`. This could be simply `IList<T> nextUpdate = null` - when it is `null`, then the queue is empty, when the reference is not null, something is enqueued. I think using concurrent queue in this case is a massive overkill. Simply use a single lock whenever you manipulate with this field and all should work well and should be easily understandable.

Toggl.Giskard/Adapters/BaseRecyclerAdapter.cs
```diff
@@ -35,8 +48,8 @@ protected BaseRecyclerAdapter(IntPtr javaReference, JniHandleOwnership transfer)
 
         public override RecyclerView.ViewHolder OnCreateViewHolder(ViewGroup parent, int viewType)
         {
-            var inflater = LayoutInflater.From(parent.Context);
```
simonrozsival: 👀

khoi: not sure what happens here, probably tab vs space kinda thing

Toggl.Giskard/Adapters/BaseRecyclerAdapter.cs
```diff
@@ -58,14 +71,81 @@ public override void OnBindViewHolder(RecyclerView.ViewHolder holder, int positi
...
+        protected virtual void SetItems(IList<T> newItems)
+        {
+            lock (updateLock)
+            {
+                nextUpdate = newItems;
+            }
+            processUpdate();
+        }
+
+        private void processUpdate()
```
daividssilverio: This algorithm doesn't look correct to me.
Correct me if I'm wrong, but it looks possible to have following scenario:
```
Let's call A = [a1, a2, a3], B = [b1, b2, b3], C = []

items = A

SetItems(B) -> nextUpdate = A at moment 1.
processUpdate runs -> 
    nextUpdate = null
    oldItems = A
    newItems = B
    Task.Run -> CalculateDiff with (A, B) at moment 2. Let's say the diffUtil takes 4 time units to finish

SetItems(C) happens and nextUpdate = C at moment 3
processUpdate runs ->
    nextUpdate = null
    newItems = C
    oldItems = A
    Task.Run -> CalculateDiff with (A, C) at moment 4, Let's say the diffUtil taks 1 time unit to finish

dispatchUpdates(C, CalculateDiff(A,C).Result) happens at moment 5
    items = C
    dispatch (CalculateDiff(A,C).Result)
    ...

dispatchUpdates(B, CalculateDiff(A,B).Result) happens at moment 6
    items = B
    dispatch (CalculateDiff(A, B)) ---> BOOM, updates trying to access element changes that don't exist in C
```

khoi: you're right, I'm adding this to stress test the method and 💥

```
var diffResult = DiffUtil.CalculateDiff(new BaseDiffCallBack(oldItems, newItems));
Thread.Sleep(new Random().Next(100, 1000));
handler.Post(() => dispatchUpdates(newItems, diffResult));
```

will update it

Toggl.Foundation.MvvmCross/ViewModels/Selectable/SelectableCountryViewModel.cs
```diff
@@ -17,5 +19,23 @@ public SelectableCountryViewModel(ICountry country, bool selected)
...
+        public long Identifier => Country.Id;
+
+        public bool Equals(SelectableCountryViewModel other)
+        {
+            if (ReferenceEquals(null, other)) return false;
+            if (ReferenceEquals(this, other)) return true;
+            return Equals(Country, other.Country) && Selected == other.Selected;
+        }
+
+        public override bool Equals(object obj)
```
kornelijepetak: This `Equals` can be replaced by:

```cs
public override bool Equals(object obj) 
    => Equals(obj as SelectableCountryViewModel);
```

PR #3739, opened 2018-11-20
Improve calendar layout

### LINKED ISSUE #3621: Fix the calendar layout
Use the algorithm described [here](https://github.com/taterbase/calendar-puzzle) to calculate the overlapping items. This reduces the number of columns required to render the layout so it looks better and it’s easier to read.

juanlaube:
## What's this?
This minimizes the number columns in the calendar view so it's readable and beautiful.

### Relationships

Closes #3621

## Why do we want this?
🎨 😍 ✨ 

## How is it done?
By minimizing the number of columns. The algorithm works in two steps:
1. Group all calendar items that overlap in a cascade into buckets
2. For each bucket, put the items into columns, minimizing the number of columns*.

When inserting an item in a column, the algorithm considers that the item to insert doesn't overlap with other items in that column, otherwise, the next column will be used or a new one created.

*The items from calendar events have their own columns because it's a requirement those items are always to the left, the algorithm doesn't try to put items from time entries into columns that have items from calendar events, thus the number of columns is not exactly the minimum in some cases, but still looks good.

### Side effects
Makes @simonrozsival happy 😄

## Review hints

Please test it. The calculator is in its own class now, methods are documented and I added unit tests that cover all scenarios. 

Before/after screenshots for lazy people:

**Before**
before

**After**
after

## :squid: Permissions
⛔️ Only me.


juanlaube:
@amulware thanks for the comment, I forgot about the requirement that puts the calendar events to the left, it's fixed now!

Toggl.Multivac/Indexed.cs
```diff
@@ -0,0 +1,19 @@
+using System.Runtime.InteropServices;
```
simonrozsival: 👀

Toggl.Foundation/Calendar/CalendarLayoutCalculator.cs
```diff
@@ -0,0 +1,170 @@
+using System.Collections.Generic;
+using System.Linq;
+using Toggl.Daneel.Views.Calendar;
+using Toggl.Multivac;
+using Toggl.Multivac.Extensions;
+
+namespace Toggl.Foundation.Calendar
```
simonrozsival: I still think this should be in `Multivac` and not in `Foundation`. `Foundation` is for business logic and `Multivac` is for algorithms and helpers. This is not business logic, this is an algorithm. `Foundation` is too crowded already!

Toggl.Foundation.Tests/Calendar/CalendarLayoutCalculatorTests.cs
```diff
@@ -0,0 +1,177 @@
...
+                .And.Contain((attributes) =>
+                    attributes.StartTime == calendarItems[0].StartTime && attributes.TotalColumns == 3 && attributes.ColumnIndex == 1)
+                .And.Contain((attributes) =>
+                    attributes.StartTime == calendarItems[1].StartTime && attributes.TotalColumns == 3 && attributes.ColumnIndex == 0)
+                .And.Contain((attributes) =>
+                    attributes.StartTime == calendarItems[2].StartTime && attributes.TotalColumns == 3 && attributes.ColumnIndex == 2)
+                .And.Contain((attributes) =>
+                    attributes.StartTime == calendarItems[3].StartTime && attributes.TotalColumns == 2 && attributes.ColumnIndex == 1)
+                .And.Contain((attributes) =>
+                    attributes.StartTime == calendarItems[4].StartTime && attributes.TotalColumns == 2 && attributes.ColumnIndex == 0);
```
simonrozsival: Is this "5" or "4"? They have the same start time, so they can't be distinguished by it. This test doesn't ensure that calendar events are in the leftmost column or am I missing something?

Toggl.Foundation.Tests/Calendar/CalendarLayoutCalculatorTests.cs
```diff
@@ -0,0 +1,177 @@
...
+                .And.Contain((attributes) =>
+                    attributes.StartTime == calendarItems[2].StartTime && attributes.TotalColumns == 3 && attributes.ColumnIndex == 2)
+                .And.Contain((attributes) =>
+                    attributes.StartTime == calendarItems[3].StartTime && attributes.TotalColumns == 2 && attributes.ColumnIndex == 0)
+                .And.Contain((attributes) =>
+                    attributes.StartTime == calendarItems[4].StartTime && attributes.TotalColumns == 2 && attributes.ColumnIndex == 1);
+        }
+
+        [Fact, LogIfTooSlow]
+        public void CalendarEventsHaveTheirOwnColumnsToTheLeft()
```
simonrozsival: I would also like to see a test for overlapping calendar events.

Toggl.Foundation.Tests/Calendar/CalendarLayoutCalculatorTests.cs
juanlaube: Good catch, I added a new condition to differentiate those items. The expected column index ensures the calendar event is in the leftmost position.

Toggl.Foundation/Calendar/CalendarLayoutCalculator.cs
juanlaube: I see your point, but the layout calculator has direct dependencies on other things in Foundation, i.e: `CalendarItem`, `CalendarItemSource`, `CalendarIconKind`. In principle we can move these too because there are no other dependencies, but I think these types belong in Foundation.

juanlaube: I moved the logic to Multivac, as a side effect, I 🔥d a couple of things that were not needed 😄

Toggl.Foundation.MvvmCross/Calendar/CalendarLayoutCalculator.cs
```diff
@@ -0,0 +1,155 @@
...
+namespace Toggl.Foundation.MvvmCross.Calendar
+{
+    public sealed class CalendarLayoutCalculator
+    {
+        public IList<CalendarItemLayoutAttributes> CalculateLayoutAttributes(IList<CalendarItem> calendarItems)
+        {
+            if (calendarItems.None())
+                return new List<CalendarItemLayoutAttributes>();
+
+            var initial = new List<List<Indexed<CalendarItem>>>();
```
heytherewill: You could add a type alias to `List<List<Indexed<CalendarItem>>>` so you don't need this variable and can do:

`.Aggregate(new MoreReasonableTypeName(), groupOverlappingItems)`

Toggl.Multivac/Indexed.cs
```diff
@@ -0,0 +1,17 @@
+namespace Toggl.Multivac
```
heytherewill: We already have a `Indexed` extension method that uses Tuples instead. Is this significantly better, readability wise? I wouldn't like to have two different ways of doing the same thing lying around

Toggl.Daneel/ViewSources/CalendarCollectionViewSource.cs
```diff
@@ -254,72 +255,10 @@ private void registerCells()
...
-        {
-            var isEditing = IsEditing && editingItemIndexPath != null && index == editingItemIndexPath.Item;
-            return new CalendarCollectionViewItemLayoutAttributes(
-                calendarItem.StartTime.LocalDateTime,
-                calendarItem.Duration,
-                overlappingItemsCount,
-                positionInOverlappingGroup,
-                isEditing
-            );
+            var layoutCalculator = new CalendarLayoutCalculator();
```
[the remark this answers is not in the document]
juanlaube: No reason to, in my first refactor the calculator depended on the selected item, not that's gone 😄

Toggl.Foundation.MvvmCross/Calendar/CalendarLayoutCalculator.cs
juanlaube: Noice

Toggl.Multivac/Indexed.cs
juanlaube: I thought we didn't like tuples in general, but if we prefer tuples lets continue using that. I don't think the struct is much better, readability wise.

PR #3747, opened 2018-11-21
Giskard 1.7: Aluminium #2

daividssilverio:
### Changelog
- Added the task name when editing time entries
- Added view to collect user feedback
- Improved time entries syncing
- Improved syncing when losing and gaining access to workspaces
- Improved suggestions UI when starting time entries
- Improved performance in the time entries log UI
- Improved app startup time
- Various minor ui performance improvements

- Fixed a bug when user's account doesn't have a default workspace
- Fixed freeze on pre-nougat devices
- Fixed sign up with Google

### History of changes
https://github.com/toggl/mobileapp/compare/giskard-1.6..giskard-1.7-2

### :squid: Permissions
Merge only. Me or anyone from in my @toggl/androids in my absence and @amulware


amulware:
Released to first users. Time to go?

PR #3750, opened 2018-11-22
Toggl for iPad

dshr:
## What's this?
And assembly of .xib and code changes that make the app universal and help with rendering on iPad.

### Relationships
Closes nothing.

## Why do we want this?
Our users have been requesting this for a while now.

## How is it done?
A lot of UITraitCollection magic.

### Why not another way?
It's the only way.

### Side effects
None.

## Review hints
Still wip, will need a proper set of UI tests.

## :squid: Permissions
Only when 100% ready.


heytherewill:
Closing this for it's heavily WIP and we're not working on it immediately

PR #3752, opened 2018-11-23
🇻🇳Rxfactor SelectCountryViewModel

khoi:
## What's this?
Rxfactor of country selection screen.

### Relationships
Part of #3352

## Why do we want this?
🔫 mvx

## How is it done?
The VM exposes `IObservable<IEnumerable<SelectableCountryViewModel>>` which get binded to the table.

### Side effects
I don't think there is any.

## Review hints
None.

The implementation is similar to #3600

## :squid: Permissions
🦑


khoi:
marked as `wip` as tests are failing 🤦🏼‍♂️

Toggl.Foundation.MvvmCross/ViewModels/SelectCountryViewModel.cs
```diff
@@ -1,50 +1,50 @@
...
-using static Toggl.Multivac.Extensions.StringExtensions;
 using Toggl.Foundation.Interactors;
 
 namespace Toggl.Foundation.MvvmCross.ViewModels
 {
     [Preserve(AllMembers = true)]
     public sealed class SelectCountryViewModel : MvxViewModel<long?, long?>
     {
-        private readonly IMvxNavigationService navigationService;
+        public UIAction Close { get; }
```
heytherewill: There's a close extension, so this property is not needed

Toggl.Foundation.MvvmCross/ViewModels/SelectCountryViewModel.cs
```diff
@@ -1,50 +1,50 @@
...
-        private long? selectedCountryId;
+        public InputAction<SelectableCountryViewModel> SelectCountry { get; }
+
+        public IObservable<IEnumerable<SelectableCountryViewModel>> Countries { get; private set; }
 
-        public IMvxAsyncCommand CloseCommand { get; }
+        public InputAction<string> SetFilterText { get; }
 
-        public IMvxAsyncCommand<SelectableCountryViewModel> SelectCountryCommand { get; }
+        private ISubject<string> filterText = new BehaviorSubject<string>(string.Empty);
```
heytherewill: Make this readonly

Toggl.Foundation.MvvmCross/ViewModels/SelectCountryViewModel.cs
```diff
@@ -1,50 +1,50 @@
...
 
-        public MvxObservableCollection<SelectableCountryViewModel> Suggestions { get; }
-            = new MvxObservableCollection<SelectableCountryViewModel>();
+        private readonly IMvxNavigationService navigationService;
 
-        public string Text { get; set; } = "";
+        private long? selectedCountryId;
```
heytherewill: private fields come before props

Toggl.Foundation.MvvmCross/ViewModels/SelectCountryViewModel.cs
```diff
@@ -53,32 +53,30 @@ public override async Task Initialize()
                 allCountries.Insert(0, selectedElement);
             }
 
-            Suggestions.AddRange(
-                allCountries.Select(country => new SelectableCountryViewModel(
-                    country,
-                    country.Id == selectedCountryId)));
+            Countries = filterText
+                .Select(text =>
```
heytherewill: This would read (and perform) better like this:

```csharp
filterText
    .Select(text => text.Trim())
    .DistinctUntilChanged()
    .Select(countriesFromText)
```

Toggl.Daneel/ViewControllers/SelectCountryViewController.cs
```diff
@@ -20,29 +27,41 @@ public override void ViewDidLoad()
         {
             base.ViewDidLoad();
 
-            var source = new CountryTableViewSource(CountriesTableView);
-            CountriesTableView.Source = source;
+            CountriesTableView.SeparatorStyle = UITableViewCellSeparatorStyle.None;
```
simonrozsival: Why did you move these lines from the constructor of the view source to the view controller? 🤔 If the problem is that the constructor is called too soon (you are creating it even before the constructor of the VC is called on line 20), then just create the new instance of the view source here in `ViewDidLoad`.

Toggl.Daneel/ViewSources/CountryTableViewSource.cs
```diff
@@ -1,37 +1,34 @@
...
+            => Observable
+                .FromEventPattern<SelectableCountryViewModel>(e => OnItemTapped += e, e => OnItemTapped -= e)
+                .Select(e => e.EventArgs);
+
         private const string cellIdentifier = nameof(CountryViewCell);
+        private const int rowHeight = 48;
 
-        public CountryTableViewSource(UITableView tableView)
-            : base(tableView)
+        public CountryTableViewSource() : base(new ImmutableArray<SelectableCountryViewModel>(), cellIdentifier)
```
simonrozsival: So much information on one line - let's put `: base(...` on a separate line please! 😄

Toggl.Daneel/ViewControllers/SelectCountryViewController.cs
simonrozsival: OK

Toggl.Foundation.MvvmCross/ViewModels/SelectCountryViewModel.cs
```diff
@@ -53,32 +50,28 @@ public override async Task Initialize()
...
             }
 
-            Suggestions.AddRange(
-                allCountries.Select(country => new SelectableCountryViewModel(
-                    country,
-                    country.Id == selectedCountryId)));
+            Countries = filterText
+                .Select(text => text.Trim())
+                .DistinctUntilChanged()
+                .Select(trimmedText =>
```
heytherewill: I'd move this into a nicely named method instead of using a closure

khoi: hmm, I tried to extract this out to a local function, but it doesn't look nicer. I prefer the closure version better. How would you name the method?

PR #3826, opened 2018-11-26
Giskard: Add missing task names 

### LINKED ISSUE #3454: Assigned task not showing up on Timer + Create and Edit Entry Views
## 🐛 Bug Report
The assigned Task for an Entry does not show up on the Timer or when creating or editing an Entry.

## Context
(/conversation/18996172897)

Build/Branch: Giskard 
Phone OS: Android
Phone model:

## Expected behavior
The user should be able to see the assigned Task on the Timer, and when creating or editing the Entry.

## To reproduce
1. Create a new Entry on the Android app
2. Select a Task
3. See that the Task not appearing beside the Project's name
4. Save new Entry
5. Edit Entry
6. See that the Task not appearing beside the Project's name
7. Press X to close Edit Entry view
8. Tap on Entry's continue button on main view
9. See that the Task not appearing beside the Project's name on the Timer bar

## Suggestions
The Task already appears for the Entry while it is listed on the main view. Just carry this over to the other views mentioned here.

daividssilverio:
## What's this?
This adds task names to the running time entry card and in the start time entry view.

### Relationships
Closes #3454 
iOS version: #3685 

## Why do we want this?
So that the user can see the task names.

## How is it done?
I've just added it to the main running time entry card and updated the start time entry view to show it as well, matching iOS.

### Why not another way?
Feel free to suggest another way.

### Side effects
None.

## Review hints
Review by might be easier.

## :squid: Permissions
Anyone, 🦑 it.


Toggl.Giskard/Activities/MainActivity.cs
```diff
@@ -92,34 +94,25 @@ protected override void OnCreate(Bundle bundle)
...
 
             ViewModel.CurrentRunningTimeEntry
-                .Select(te => te?.Project?.Client != null)
-                .Subscribe(timeEntryCardClientLabel.Rx().IsVisible())
-                .DisposedBy(DisposeBag);
-
-            ViewModel.CurrentRunningTimeEntry
-                .Select(te => te?.Project?.Name ?? "")
-                .Subscribe(timeEntryCardProjectLabel.Rx().TextObserver())
+                .Select(te => createProjectClientTaskLabel(te))
```
juanlaube: For some reason this subscription is not called when `CurrentRunningTimeEntry` changes and the project/task/client label is empty.

Toggl.Giskard/Activities/MainActivity.cs
```diff
@@ -92,34 +94,25 @@ protected override void OnCreate(Bundle bundle)
...
             var projectVisibilityObservable = ViewModel.CurrentRunningTimeEntry
                 .Select(te => te?.Project != null);
 
             projectVisibilityObservable
-                .Subscribe(timeEntryCardProjectLabel.Rx().IsVisible())
-                .DisposedBy(DisposeBag);
-
-            projectVisibilityObservable
-                .Subscribe(timeEntryCardDotContainer.Rx().IsVisible())
+                .Subscribe(projectIsVisible =>
```
juanlaube: Is there a particular reason for this change? I think the previous one is better, readability wise.

daividssilverio: Yes, I didn't try to solve it other way, but there seems to be a racing condition there, the `timeEntryCardProjectClientTaskLabel` and `timeEntryCardDotContainer` were not becoming visible/invisible at the same time. Sometimes the dot would appear and not the label.
It might be related to the observable not being called when the label is empty, I'll take a look again.

Toggl.Giskard/Activities/MainActivity.cs
daividssilverio: It is being called here 🤔 Is there any other condition for it to happen (to not be called)?

Toggl.Giskard/Activities/MainActivity.cs
daividssilverio: Anyway, I couldn't replicate what I just explained, so I've reverted to how it was before.

PR #3835, opened 2018-11-27
Background Syncing

### LINKED ISSUE #2737: TimeoutException in RetryDelayService crashes app

### LINKED ISSUE #3917: Too Many Requests syncing failure UX needs changing
We have a problem that our app sometimes makes too many requests to the server in a short time span (e.g., user uses pull-sync several times in a row) and our app receives a 429 response. In this case we are supposed to wait for 60 seconds and only then continue.

With the new leaky bucket implementation (#3876) we should "hardly ever" run into this problem, but it might happen anyway (e.g., the user uses the app and the web app at the same time, which might happen when user is trying to resolve some syncing issues between the apps he thinks he/she has). There will always a possibility that this happens. The question is what to do in that case.

Right now we keep showing a grey bar with a `Syncing...` text. The app internally doesn't do anything for 60 seconds and then it tries to do whatever failed before and hopefully this time we don't get any 429. It might seem to the user that syncing got stuck. Also if the user closes the app at this moment, syncing most likely won't finish.

With a "fail fast" approach (#3879) we show a grey bar with a `Failed` text. User can then use pull-to-refresh immediately which would most likely cause yet another 429. In this case syncing does not appear to be stuck, but it might scare the user because of the copy and the most likely also failing next attempt to force sync.

Because we will be able to limit the number of cases when we run into this problem I suggest we use the "fail fast" approach and change the copy to something like a grey `Try again in a minute.` (there is very little space for a long message). We might also disable pull-to-refresh for the next 60 seconds.

Ideas and suggestions are welcome!

simonrozsival:
The background syncing implementation is now ready to be reviewed.

We especially need to review the changes in the syncing algorithm. Have a look at the [final graph](https://github.com/toggl/mobileapp/files/2691822/bg-sync-graph.svg.zip).

Also check the changes to the leaky bucket.

Read the `md` docs.

Get the `1.14.3` test build from TestFlight.

Get an Android test build from Bitrise (check the latest `Release.Android.PlayStore` build from the `simon/background-sync-test-builds` branch).

Check the data in the AppCenter, specifically events `BackgroundSyncStarted`, `BackgroundSyncFinished`, `NumberOfSyncedTimeEntriesWhenResumingTheAppFromBackground`.

This PR will make these PR(s) obsolete:

- Closes #2737 (the `RetryDelayService` was removed)
- Closes #3917 (the `TooManyRequests` error leads to a failure - "fail fast")


amulware:
Please summarise for me what testing has been done. :)

juanlaube:
The sync graph looks good to me.

simonrozsival:
Current state of testing:
- we created a build for iOS (version "1.14.3", available in TestFlight) from this code
- we created a build for Android (avialable on Bitrise) from this code
- we have apps on our phones running, we monitor AppCenter for the statistics on how sync works
- so far the data looks promising:
    - on iOS we have 9 background syncs per day per device on average so far (1 of these syncs did not finish or finishing was not reported to AppCenter)
    - on Android we have 6 background syncs per day per device on average so far (1 of these syncs did not finsih or finishing was not reported to AppCenter)
    - it's quite hard to evaluate if the app (the time entries) was up-to-date at the time when the user opened the app from this data: 
screen shot 2018-12-20 at 10 19 07 am
        - I'm not sure under which conditions can we get `0` synced time entries we should get `1` in the case when nothing was synced because of the way `since` parameters work right  now

amulware:
I'll review the graph, and check out the app center data tomorrow. Ideally, I'd like one more review from anyone else not on the project team as well. :)

Toggl.Daneel/Startup/AppDelegate.cs
```diff
@@ -310,5 +312,35 @@ private void startTimeEntryInBackground(string eventId, Action completionHandler
...
+
+        public override void PerformFetch(UIApplication application, Action<UIBackgroundFetchResult> completionHandler)
+        {
+            var interactorFactory = Mvx.Resolve<IInteractorFactory>();
+            if (interactorFactory == null) return;
+            interactorFactory
+                .RunBackgroundSync()
+                .Execute()
+                .Select(mapToNativeOutcomes)
+                .Do(completionHandler)
```
heytherewill: Why is this on a `Do` instead of being on the subscribe?

Toggl.Foundation/Interactors/Sync/RunBackgroundSyncInteractor.cs
```diff
@@ -0,0 +1,55 @@
...
+
+            this.syncManager = syncManager;
+            this.analyticsService = analyticsService;
+            this.stopwatchProvider = stopwatchProvider;
+        }
+
+        public IObservable<SyncOutcome> Execute()
+        {
+            var syncTimeStopwatch = stopwatchProvider.Create(MeasuredOperation.BackgroundSync);
+            var systemStopwatch = new Stopwatch();
```
heytherewill: I don't think this is needed at all

Toggl.Foundation/Models/SyncOutcome.cs
```diff
@@ -0,0 +1,8 @@
+namespace Toggl.Foundation.Models
+{
+    public enum SyncOutcome {
```
heytherewill: No egyptian, please

Toggl.Foundation/Services/AutomaticSyncingService.cs
```diff
@@ -0,0 +1,95 @@
...
+                .DisposedBy(disposeBag);
+
+            userAccessManager.UserLoggedOut
+                .Subscribe(_ => stop())
+                .DisposedBy(disposeBag);
+        }
+
+        private void start(ITogglDataSource dataSource)
+        {
+            stop(); // replace any previous automatic syncing (if any)
```
heytherewill: This comment is not needed

Toggl.Foundation/Interactors/Sync/RunBackgroundSyncInteractor.cs
```diff
@@ -0,0 +1,55 @@
...
+            syncTimeStopwatch.Start();
+            systemStopwatch.Start();
+            analyticsService.BackgroundSyncStarted.Track();
+            return syncManager.ForceFullSync()
+                              .LastAsync()
+                              .Select(_ => SyncOutcome.NewData)
```
heytherewill: Is there no way to know if there's actual data or not?

Toggl.Foundation/Interactors/Sync/RunBackgroundSyncInteractor.cs
```diff
@@ -0,0 +1,55 @@
...
+                              .Catch((Exception error) => syncFailed(error))
+                              .Do(_ => systemStopwatch.Stop())
```
heytherewill: Is this is even needed, I'd still make all three action happen on the same `.Do` block

Toggl.Foundation/Models/SyncOutcome.cs
```diff
@@ -0,0 +1,8 @@
...
+        NewData,
+        NoData,
+        Failed
```
heytherewill: Shouldn't we return `Failed` instead of throwing (using `.Catch`) on the sync interactor?

Toggl.Foundation/Sync/States/Pull/FetchAllSinceState.cs
```diff
@@ -1,50 +1,107 @@
...
         public IObservable<ITransition> Start() => Observable.Create<ITransition>(observer =>
         {
-            var since = database.SinceParameters;
+            if (!leakyBucket.TryClaimFreeSlots(numberOfSlots: 9, timeToFreeSlot: out var timeToNextFreeSlot))
+            {
+                observer.CompleteWith(PreventOverloadingServer.Transition(timeToNextFreeSlot));
+                return () => { };
+            }
+
+            // The dependencies between the requests are based on the intended order of processing the entities
```
heytherewill: Not sure I like such huge comment. Maybe this should be documented?

Toggl.Foundation/Sync/States/Pull/FetchAllSinceState.cs
```diff
@@ -1,50 +1,107 @@
...
+            // in the sync states. Check the transitions graph and make sure that the requests and states are aligned.
+            // We send at most 3 parallel requests to prevent overloading the server with large bursts of requests.
+
+            // first wave
```
heytherewill: I don't see how these wave comments are relevant

Toggl.Daneel/Startup/AppDelegate.cs
simonrozsival: Hmm. I guess it would make it a bit clearly that it's not just a side effect 👍

Toggl.Foundation/Interactors/Sync/RunBackgroundSyncInteractor.cs
simonrozsival: No, not at this point. Mainly because we "always" fetch at least 1 time entry because it has the same `At` value as the `?since=` parameter we use for the HTTP requests. This results in the TE being in the list of TEs we fetch on every request. We're currently discussing it with Paul and backend but we have no solution for this just yet.

Toggl.Foundation/Interactors/Sync/RunBackgroundSyncInteractor.cs
simonrozsival: You are right, we forgot to remove this.

Toggl.Foundation/Models/SyncOutcome.cs
simonrozsival: I don't understand you. We do exactly that.

amulware:
Did testing include the following?

- no default workspace case
- no workspace case
- api token reset

(We don't happen to test these automatically with our sync tests, do we?)

dshr:
Tested, all these work fine:
- [x] no default workspace case
- [x] no workspace case
- [x] api token reset

simonrozsival:
Thanks, @dshr! 🙌

Issue #3760, opened 2018-11-23
Make it possible to observe if any new data were fetched during background sync


simonrozsival:
Apple wants us to report if we fetched any new data or if there were no new data on the server. Currently we would have to listen to the data sources for update/create/delete events. Maybe it would be possible to modify the sync manager so that it is possible to monitor the progress and observe if any new data are downloaded from the server.

simonrozsival:
The implementation was merged into the feature branch.

PR #3839, opened 2018-11-27
Track starting and finishing background sync

### LINKED ISSUE #3758: Track starting and finishing background sync
The idea is to track how many times we start bg sync and how many times the sync runs untill the end without being terminated by the OS.

I think the best way would be to track an event when syncing starts and another (different) event when it finishes. Both can happen in the interactor.

We can then see the number of started bg syncs and the number of finished ones. In the ideal case they would be roughly the same. If the number of starts is higher than the number of finished ones, then we are apparently running into some issues and the OS terminates too many syncs.

dshr:
## What's this?
Adding tracking to our background sync interactor!

### Relationships
Closes #3758 and #3840 

## Why do we want this?
So we can then see the number of started bg syncs and the number of finished ones.

## How is it done?
By tracking when the sync starts and when the sync stops (if it stops).

### Why not another way?
This is as simple as it goes.

### Side effects
None.

## Review hints
Run tests, I guess? None otherwise.

## Tests
Does not impact sync in any way yet.

## :squid: Permissions
Background sync people only please


simonrozsival:
I didn't specify it in the issue but I think it would be a good idea to track the outcome as the parameter of `BackgroundSyncFinished`.

BTW keep in mind that the sync manager can return `Failed` result even when no exception is thrown and caught. It should not affect anything and I will probably poke the code a bit later with #3760 (I assume the `LastAsync` will always return `Sleep` which is not useful and it might actually hide some caught failures.

Toggl.Foundation.Tests/Interactors/Sync/RunBackgroundSyncInteractorTests.cs
```diff
@@ -45,11 +45,32 @@ public async Task ReturnsFailedIfSyncFails()
...
+            }
+
+            [Fact, LogIfTooSlow]
+            public async Task TracksIfSyncFails()
+            {
+                var exception = new Exception();
+                SyncManager.ForceFullSync().Returns(Observable.Throw<SyncState>(exception));
+                await interactor.Execute().SingleAsync();
+                AnalyticsService.BackgroundSyncStarted.Received().Track();
+                AnalyticsService.BackgroundSyncFinished.DidNotReceive().Track(Arg.Any<string>());
```
simonrozsival: Shouldn't this be `.Track(nameof(SyncOutcome.Failed))`?

Toggl.Foundation/Interactors/Sync/RunBackgroundSyncInteractor.cs
```diff
@@ -24,10 +25,18 @@ public RunBackgroundSyncInteractor(ISyncManager syncManager, IAnalyticsService a
...
                               .LastAsync()
                               .Select(_ => SyncOutcome.NewData)
-                              .Catch(Observable.Return(SyncOutcome.Failed));
+                              .Catch((Exception error) => syncFailed(error));
+                              .Track(analyticsService.BackgroundSyncFinished, nameof(SyncOutcome.NewData))
+        }
+
+        private IObservable<SyncOutcome> syncFailed(Exception error)
+        {
+            analyticsService.BackgroundSyncFailed.Track(error.GetType().ToString());
```
simonrozsival: I'm not sure what `.GetType().ToString()` actually is - it would be better to use `.GetType().FullName`. Also let's track as much data as possible here - we should track the message of the exception and maybe even the stack trace, as we did in #3842

Toggl.Foundation/Interactors/Sync/RunBackgroundSyncInteractor.cs
```diff
@@ -24,10 +25,18 @@ public RunBackgroundSyncInteractor(ISyncManager syncManager, IAnalyticsService a
 
         public IObservable<SyncOutcome> Execute()
         {
+            analyticsService.BackgroundSyncStarted.Track();
             return syncManager.ForceFullSync()
                               .LastAsync()
                               .Select(_ => SyncOutcome.NewData)
-                              .Catch(Observable.Return(SyncOutcome.Failed));
+                              .Catch((Exception error) => syncFailed(error));
```
simonrozsival: `;` 🙈

PR #3842, opened 2018-11-27
Track login error

juanlaube:
## What's this?
Tracks generic login error exceptions.

### Relationships

Closes nothing

## Why do we want this?
There's a possible bug out there in Daneel 1.14 and some users cannot login. This tracks the exception in a generic error.

## How is it done?
Tracking the exception.

### Side effects
None.

## :squid: Permissions
⛔️ Only me


Toggl.Foundation/Analytics/BaseAnalyticsService.cs
```diff
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
```
simonrozsival: 👀

Toggl.Foundation.MvvmCross/ViewModels/LoginViewModel.cs
```diff
@@ -271,6 +271,8 @@ private void onError(Exception exception)
                     errorMessageSubject.OnNext("");
                     break;
                 default:
+                    analyticsService.UnknownLoginFailure.Track("Login failed with unknown reason", exception.StackTrace);
```
simonrozsival: Somehow include the `exception.Message` and `exception.GetType().FullName` so we get as much info as possible. The event name (`UnknownLoginFailure`) already says the same thing as `"Login failed with unknown reason"`.

Toggl.Foundation.MvvmCross/ViewModels/SignupViewModel.cs
```diff
@@ -290,6 +290,8 @@ private void onError(Exception exception)
                     errorMessageSubject.OnNext(Resources.EmailIsAlreadyUsedError);
                     break;
                 default:
+                    analyticsService.UnknownLoginFailure.Track("Sign up failed with unknown reason", exception.StackTrace);
```
simonrozsival: Same as for login.

Toggl.Foundation/Analytics/BaseAnalyticsService.cs
juanlaube: 🤦‍♂️

PR #3844, opened 2018-11-28
Fix misbehaving main running timer visibility

### LINKED ISSUE #3495: Running Timer doesn't appear
## 🐛 Bug Report
We've received several reports of app users not seeing a running Timer after it is started.

## Context

Build/Branch: Giskard 1.5.2
Phone OS: 6.0 LOLLIPOP_MR1, 8.0.0 N_MR1, 8.0.0 N_MR1, 8.0.0 N_MR1, 6.0.1 LOLLIPOP_MR1, 
Phone model: Sony E5333, samsung SM-N950U1, Sony F5321, samsung SM-G960U, OnePlus ONE E1005,

## Expected behavior
When starting the Timer (on the web app, desktop app, or even on the mobile app) this should be visible when going to the main app's main screen.

## To reproduce
Unable to provide at this time; no one on support is able to reproduce.

## Suggestions
The only common factor among reports so far is that they're all on Giskard 1.5.2.

Reports below might provide some insight:

(/conversations/19088435115)
_Pulling the screen down does initiate a manual sync that grabs past entries. However, it still doesn't grab the current ongoing task that I've started on the PC desktop app. The mobile app should be showing my current ongoing task._

(/conversations/19109968332)
_I'm having major syncing issues between the new Android app and the web app. for some reason whenever I start a timer on the web app my phone is hit or miss with catching it._

(/conversation/19046710379)
_New timer entries do not display at bottom of app or in notification. Timer is still running -- if you hit the green arrow again and start a new timer, the original timer is stopped and displayed in the list of completed tasks. The new timer, however, is completely hidden and the app shows no sign that it is running._

(/conversation/19101976572)
_I had started the time entry on the app, and it disappears from the currently-running timer area (ie, the bottom of the app screen). I am not able utilize the timer due to that._

(/conversations/19101976572)
_Hi, the toggl android app is not working properly. It is not showing the timer mode sometimes, even though time entry being monitored._

(/conversations/19078387693)
_Even if I run a new timer, the ongoing timer didn't appear unless I restarted the app. Just wondering if these are two related issues._

One user has been kind enough to try a manual sync and logging out then back in:

_A manual sync does not fix the issue._
_Logging out and logging back in does appear to get the running timer to display -- hooray!_

(Many more details at /conversations/19046710379)

daividssilverio:
## What's this?
It fixes the misbehaving visibility handling of the main running timer in the MainActivity.

### Relationships
Closes #3495

## Why do we want this?
People get annoyed and confused when they start/continue time entries and don't see it in the main screen.

## How is it done?
I've added a callback to run when the card animation finishes. This seems to be reason why the card wasn't appearing in some cases. Sometimes, the stop animation callback(1) from the `radialAnimation` would run after a time entry was started(2) and set the visibility of the card to `invisible` after (2) setting the visibility to `visible`, causing the card to stay invisible, since the `IsTimeEntryRunning` observable wouldn't emit new values in case a new time entry is started/continued while one is still running.

You can even confirm this behaviour by getting is this weird state by increasing the animation duration to 1 or 2 seconds, stopping one running time entry then continuing another as fast as you can.
You'll get an invisible running time entry, confirm in the web app.
Then, stop the time entry in the web app, refresh the app, then start/continue another time entry in the app and the card should appear once more.

### Why not another way?
I did remove `IsTimeEntryRunning`'s `DistinctUntilChanged` in the viewModel, but that is more intrusive.
Ideas are welcome.

### Side effects
If this weird state happens to happen for some user, `assureRunningTimeEntryCardVisibility` will fix it, but it will take 1 second for the card to appear in these cases, instead of 0.5 seconds.

## Review hints
Try replicating, think about racing conditions, suggest stuff.
I've run into so many other weird and cryptic crashes trying to replicate this, so be aware that a simple clean, uninstall the app, install again might help you.

## :squid: Permissions
Everybody is welcome to 🦑


Toggl.Giskard/Activities/MainActivity.cs
```diff
@@ -295,11 +306,24 @@ private async void onTimeEntryCardVisibilityChanged(bool visible)
...
+                        assureRunningTimeEntryCardVisibility();
+                    })
                     .Start();
             }
         }
 
+        private void assureRunningTimeEntryCardVisibility()
+        {
+            var isCardVisible = runningEntryCardFrame.Visibility == ViewStates.Visible;
+            if (isTimeEntryRunning && !isCardVisible || !isTimeEntryRunning && isCardVisible)
```
simonrozsival: This looks like a XOR (`^`) to me 😄 But as we do not use that operator very often, maybe this explicit form is better.

amulware: How about `!=`?

Toggl.Giskard/Extensions/CircularRevealAnimation.cs
```diff
@@ -99,11 +111,21 @@ public CircularRevealAnimation SetType(Func<AnimationType> determineAnimationTyp
...
+            return this;
+        }
 
+        public CircularRevealAnimation OnAnimationCancel(Action action)
+        {
+            onAnimationCancel = action;
             return this;
         }
 
+        public CancellationTokenSource GetCancellationToken()
```
heytherewill: This name is a bit misleading; you're not getting the cancellation token, but its source instead

Toggl.Giskard/Extensions/CircularRevealAnimation.cs
```diff
@@ -129,12 +151,32 @@ public void Start()
...
+                    animator.AnimationCancel -= onCancel;
+                    onAnimationCancel?.Invoke();
+                };
+
                 animator.AnimationEnd += onEnd;
+                animator.AnimationCancel += onCancel;
 
-                if (type == AnimationType.Appear)
-                    view.Visibility = ViewStates.Visible;
+                animator.AnimationStart += (o, s) =>
```
simonrozsival: Why not keep the old code here? This callback seems weird 🤔

daividssilverio: The visibility of the view being animated to appear was getting changed to visible even if the animation didn't have a chance to start. This had to be done since the animation now can be cancelled.

PR #3856, opened 2018-11-28
Protect interactors

### LINKED ISSUE #2920: "Sequence contains no matching element" in RealmAdapter.Get(long)
## 💥 Crash Report
It crashes when fetching some object by Id. We're not doing that in a lot of places, are we? Maybe this alone is enough info to deduce the problem.

### 👥 Impact
- 21 times to 19 users in 1.8.1 so far, that's about 15% of crashes
- different phones and OS versions

### 🔗 Context
I think we were trying to give more meaningful exceptions to some of these cases, to for example resolve the generic type, so we know what is being fetched here. Maybe we have to do that here too, if we don't know how else to fix it.

### 🛰 Link
https://appcenter.ms/orgs/Toggl/apps/Toggl-iOS/crashes/groups/a83cce09bc69a9d270bb98010728a8ef3e168475?appVersion=1.8.1

`System.InvalidOperationException: Sequence contains no matching element`
```cs
RealmResultsVisitor.VisitMethodCall (System.Linq.Expressions.MethodCallExpression node)
MethodCallExpression.Accept (System.Linq.Expressions.ExpressionVisitor visitor)
ExpressionVisitor.Visit (System.Linq.Expressions.Expression node)
RealmResultsProvider.Execute (System.Linq.Expressions.Expression expression)
RealmResultsProvider.Execute[T] (System.Linq.Expressions.Expression expression)
Queryable.Single[TSource] (System.Linq.IQueryable`1[T] source, System.Linq.Expressions.Expression`1[TDelegate] predicate)
RealmAdapter`2[TRealmEntity,TModel].Get (System.Int64 id)
Repository`1+<>c__DisplayClass1_0[TModel].<GetById>b__0 ()
BaseStorage`1+<>c__DisplayClass10_0`1[TModel,T].<CreateObservable>b__0 (System.IObserver`1[T] observer)
```

Odrakir:
## What's this?
This creates a wrapper around interactors so we can catch specific exceptions and get a log of what went wrong.

Now uses a different approach, some casting is needed, but the call is a bit more clean and tests work (I think).

Usage:
```c#
interactorFactory.GetDefaultWorkspace()
    .Protect<InvalidOperationException, IThreadSafeWorkspace>("AppDelegate.startTimeEntryInBackground")
    .Execute();
```

### Relationships
All the issues in which we move datasources into interactors should be affected by this. This one should be finished and merged after those (See [refactor interactors project](https://github.com/toggl/mobileapp/projects/34))

Closes #2920 

## Why do we want this?
There's a bug we can't find.

### Why not another way?
It could be done inside every affected interactor, the current approach seems more modular... 🤔 

## :squid: Permissions
Still needs some work, a lot of interactors to protect and tests to write.


heytherewill:
>What do we want to do with the exception once we catch it? Write now is writing to the console, I assume we want to track it?
Logging to the telemetry service with some additional info that'll help us pin it down would be nice, wouldn't it?

> Do we want to throw anyway or return an Observable.Never as I'm doing right now?
I think we should either rethrow or return some default value (assuming that all call sites are aware that they can get a default value and handle it accordingly). Returning `.Never` is not a good option, IMO

> It's a bit of a pain to write all the ProtectAgainst operator where needed. Should we instead write it in other interactors? Is it worth it?
Arguably you could move the protect call to within the the Interactor factory and use [`[CallerMemberName]`](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.callermembernameattribute?view=netframework-4.7.2) to get the caller method's name

simonrozsival:
I am not sure about `Never` - we almost always await execution of interactors. With `Never` this would freeze the app, wouldn't it?

Odrakir:
`Empty` then?

Odrakir:
Ok, as it is this is the current call:

```c#
interactorFactory
    .Protect<InvalidOperationException, IThreadSafeWorkspace>(interactorFactory.GetDefaultWorkspace(), "AppDelegate.startTimeEntryInBackground")
    .Execute();
```

Alternatives: 
- Write the catch directly into the interactors we want to watch, and send the message as a parameter.
```c#
interactorFactory
    .GetDefaultWorkspace("AppDelegate.startTimeEntryInBackground")
    .Execute();
```
- Use the Decorator pattern with the InteractorFactory so we can have a ProtectedInteractorFactory (for example). This implies rewriting access methods to all interactors in the decorator (painful).
```c#
interactorFactory
    .Protected("AppDelegate.startTimeEntryInBackground")
    .GetDefaultWorkspace()
    .Execute();
```

... I can't think of any other alternative that doesn't imply many changes to the code.

amulware:
Shouldn't whoever calls the interactor decide if they want to crash or not if an error happens? If we swallow the exception silently, it might lead to all sorts of other bugs, no?

Odrakir:
That's what we do in the current approach: the caller decides wether to wrap the interaction in Protect or not. But otherwise we could still throw the exception even if we first send it to analytics.

The problem I'm finding now with the current approach is it breaks all tests where we are substituting `InteractorFactory.GetDefaultWorkspace().Execute()`... And I can't find a good way to fix this...

amulware:
I don't understand the approach then.

This is what it looks like to me:

- realm can throw an exception
- if we don't catch it we don't know where it's from, but at least the app crashes and doesn't get into a bad state
- the protection thing here catches it, so we can report where it happens better, but then it also prevents the app from crashing so it could get into a bad state

The last point bothers me.

Did I misunderstand?

Odrakir:
No, that's what it currently does, but we could also rethrow the exception after we track.

amulware:
I think that would be better. Or at least split the tracking and the protecting, so that the called can decide if the error would lead to a bad state (crashing is better, if we can't recover I think), or catch, if the error can be recovered from easily.

Odrakir:
Ok, the thing is: 
- We always want to protect the same interactors
- The current modular approach is giving problems with substitutes in tests

How about we just modify the implementation of `Execute` (and the constructor of the interactor, so it gets the message) or maybe just add a `ExecuteProtected` instead of all this juggling with the factory and all that?

PR #3857, opened 2018-11-28
Implement services needed for the android calendar

heytherewill:
## What's this?
This implements the Calendar service and the permissions service for android

## Why do we want this?
This is needed for the [suggestions project](https://github.com/toggl/mobileapp/projects/31)

## How is it done?
The calendar part is just a wrapper around the gross cursor manipulation that droid forces on us
The permission part needed some clever passing of references

### Why not another way?
These things are tricky

### Side effects
As it stands right now, you can only request permissions when in an RxFactored Activity. @arturs105 needs to tell us when they plan on requesting permissions. This needs to be very well thought, since permissions is something pretty intrusive

## Review hints
Ask For permission and then call the calendar related interactors from the MainViewModel. Everything should work fine.

Please test on real devices, not on emulators. Non-stock android tests are appreciated, since I tested this on a Pixel.

## :squid: Permissions
Only the suggestions team, but not without explicit approval from @daividssilverio


Toggl.Giskard/Properties/AndroidManifest.xml
```diff
@@ -3,5 +3,7 @@
 	<uses-sdk android:minSdkVersion="21" android:targetSdkVersion="27" />
 	<uses-permission android:name="android.permission.INTERNET" />
 	<uses-permission android:name="android.permission.VIBRATE" />
+	<uses-permission android:name="android.permission.READ_CALENDAR" />
+	<uses-permission android:name="android.permission.WRITE_CALENDAR" />
```
juanlaube: Why write? 👀

juanlaube:
I just glanced through the code, looks good except for write permission, I don't know why that's needed.
I will finish my review tomorrow and test on a Samsung device.

Toggl.Giskard/Services/CalendarServiceAndroid.cs
```diff
@@ -3,23 +3,133 @@
...
+using Android.Database;
+using Android.Content.PM;
 
 namespace Toggl.Giskard.Services
 {
     public sealed class CalendarServiceAndroid : PermissionAwareCalendarService
     {
-        public CalendarServiceAndroid(IPermissionsService permissionsService) 
+        private static readonly string eventSelection = $"({Events.InterfaceConsts.Id} = {{0}})";
+        private static readonly string dateSelection = $"(({Events.InterfaceConsts.Dtstart} >= {{0}}) AND ({Events.InterfaceConsts.Dtend} <= {{1}}))";
```
arturs105: Wouldn't these be better as methods that take the appropriate parameters and return the thing?

heytherewill: Fair enough

Toggl.Giskard/Services/PermissionsServiceAndroid.cs
```diff
@@ -1,27 +1,110 @@
...
+        private void onPermissionChanged(int requestCode, string[] permissions, Permission[] grantResults)
+        {
+            if (requestCode != calendarAuthCode)
+            {
+                calendarAuthorizationSubject?.OnNext(false);
+                calendarAuthorizationSubject?.OnCompleted();
+                calendarAuthorizationSubject = null;
+            }
+
+            var permissionWasGranted = grantResults.Any() && grantResults.First() == Permission.Granted;
```
daividssilverio: This works for now since we are only requesting permissions for the calendar, or if only request a single permission per request, but the right thing to do would be match each grantResults[x] to each permissions[x].
Just keep this in mind.

heytherewill:
@daividssilverio The way they'll consume this is up to the suggestions task force and I'll review it once they chose how to do it :)

juanlaube:
The code looks good and I can fetch the calendars, but for some reason I can't fetch the events 🤔 

I added this snippet in MainViewModel to test (after accepting the permissions):
```
var calendarService = Mvx.Resolve<ICalendarService>();
        calendarService.GetEventsForDate(DateTime.Today)
            .Subscribe(_ => Console.WriteLine(_))
            .DisposedBy(disposeBag);
```

Toggl.Giskard/Services/CalendarServiceAndroid.cs
```diff
@@ -1,25 +1,134 @@
...
+            );
+        }
+
+        private static string getEventSelectionById(string id) 
+            => $"({Events.InterfaceConsts.Id} = {id})";
 
-        protected override CalendarItem NativeGetCalendarWithId(string id)
-            => default(CalendarItem);
+        private static string getEventSelectionByDate(DateTimeOffset startDate, DateTimeOffset endDate)
+            => $"(({Events.InterfaceConsts.Dtstart} >= {startDate.ToUnixTimeSeconds()}) AND ({Events.InterfaceConsts.Dtend} <= {endDate.ToUnixTimeSeconds()}))";
```
juanlaube: [Dtstart](https://developer.android.com/reference/android/provider/CalendarContract.EventsColumns#DTSTART) and [Dtend](https://developer.android.com/reference/android/provider/CalendarContract.EventsColumns#DTEND) are in milliseconds since epoch 🤦‍♂️

Toggl.Giskard/Services/CalendarServiceAndroid.cs
```diff
@@ -1,25 +1,134 @@
...
+                yield return calendarItemFromCursor(cursor);
+            }
+        }
+
+        protected override CalendarItem NativeGetCalendarItemWithId(string id)
+        {
+            var appContext = Mvx.Resolve<IMvxAndroidGlobals>().ApplicationContext;
+
+            var selection =  getEventSelectionById(id);
+            var cursor = appContext.ContentResolver.Query(Events.ContentUri, eventsProjection, selection, null, null);
```
juanlaube: Wouldn't be better to use selectionArgs? Even if sql injection is not a problem now, at least as a reference in the future. I don't like little bobby tables 😄

juanlaube:
I tested with a Samsung device, using the Samsung calendar app instead of google calendar, works as expected (solving the milliseconds problem first).

Toggl.Giskard/Services/CalendarServiceAndroid.cs
heytherewill: I specifically tested locally with millis and, despite the docs, only seconds worked. I'll give this another spin to see what gives

PR #3858, opened 2018-11-28
Restart iOS simulator before each test

juanlaube:
## What's this?
This solves the flakiness problem when running the UI tests using the iOS simulator by restarting it before each test.

### Relationships

Closes nothing.

## Why do we want this?
We don't want flaky tests.

## How is it done?
The fix is ugly, but so far is the only thing that works. It will invoke `xcrun simctl shutdown all` and `xcrun simctl boot <device-id>` before each test to restart the simulator.

### Why not another way?
I tried a similar approach but restarting the processes inside the simulator, that didn't work.
Also I don't see a way to do this without relying on `xcrun simctl` magic.

### Side effects
Running all the tests is slower now (in fact, takes more time to shutdown and boot the simulators than running a test), but at least there are no flaky test.

## :squid: Permissions
🦑


ApocalypticFish:
Yeah, as you said it's not great but it's a start. If it allows a device to run the tests without being interrupted then it's still an improvement.

PR #3860, opened 2018-11-29
Share operator

Odrakir:
## What's this?
There are two very similar Rx operators: `ConnectedReplay` and `Share`. The only different is their functionality is the fact that the former replays all events to new subscribers and the latter doesn't.

But also, internally, `ConnectedReplay` uses `Connect` to start the subscription instead of `RefCount` which has two consequences:
- Subscription may start before we really need it (though I don't think that's a problem)
- The stream may not ever get disposed

I think they both can be implemented in a similar way and also have a similar signature so they are easier to find.

## How is it done?
I merged both operators in one called `Share(bool replay)`. Internally they both work with `RefCount()`

### Side effects
Only possible side effects would be in sync, which is the only place `ConnectedReplay` was used. I would like @amulware to take a look.

## :squid: Permissions
Let me check tests first. This is a WIP.


simonrozsival:
We use `ConnectedReplay()` to send HTTP requests ASAP (without waiting for somebody to subscribe). How do we implement this with this change?

I still don't remember how exactly `RefCount` works - what does happen when you subscribe, unsubscribe, subscribe again (in this specific order) - does the observable get created only once or will it be created twice? Creating multiple times would mean sending the same HTTP request several times, which is a no no. If it creates it only once, then all's fine.

Odrakir:
Ok, in that case I think I can close this PR 🤦‍♂️ 😄  

I thought both those things were unexpected side effects, to be honest. It's a little weird for me to start the subscription with no subscribers.

About `RefCounts`, it keeps the stream around as long as there are any subscribers, if the number of subscribers goes down to zero it'll dispose and the next subscription will recreate the Observable. Which usually is the desired effect, as you'll want to make another request at some point. But it seems like it's not for this case. Why not? Where do we expect an Observable to "be alive" after all its subscribers are gone?

simonrozsival:
We can still remove `ConnectedReplay` and move it only to the `FetchAllSinceState` class as a method. It doesn't have to be a globally accessible extension method.

Odrakir:
It's used in a couple of places outside of `FetchAllSinceState`... which makes me wonder if those were done as intended too... `SyncManager` and `TimeEntriesDataSource`.

Odrakir:
I'll close this. My assumptions were wrong and it doesn't seem necessary to move `ConnectedReplay` into a class method for now.

PR #3863, opened 2018-11-29
Daneel 1.14.2 - Rigel hotfix 2

juanlaube:
## Changelog
```
- Fixed a problem that logged out users when updating from a previous version
```

## History of changes
https://github.com/toggl/mobileapp/compare/daneel-1.14.1...rigel-hotfix-2

## Notes
`daneel-1.14.1` was cut from the wrong tag.
Should have been `daneel-1.14-2` instead of `daneel-1.14`, two commits were missing in the previous hotfix.
This branch was cut from `daneel-1.14.1` and cherry-picked those two missing commits.


simonrozsival:
I believe there is something wrong with the diff. `develop` should already contain the changes from #3723 and #3737. Does anybody have a good explanation why that happens or should we close this PR and create a new branch which would have a correct diff?

juanlaube:
@simonrozsival I added a note in the PR description regarding those two commits.

PR #3864, opened 2018-11-29
Remove migration

### LINKED ISSUE #3865: Users unable to login after updating to Daneel 1.14
## 🐛 Bug Report
Some users were logged out and unable to login again after updating from an old version to Daneel 1.14. This problem is caused by accessing a non-existent dynamic property in a database migration.

## Context
Build/Branch: Daneel 1.14 (maybe Giskard 1.7 too?)
Phone OS:
Phone model:

## Expected behavior
Users shouldn't be logged out after updating.

## To reproduce
Update from Daneel 1.8 to Daneel 1.14.

juanlaube:
## What's this?
Removes an unneeded migration that caused users to be logged out. This happened because the database migration failed when updating from a previous version without `RealmWorkspace.IsGhost` (that's <= 1.8, affecting ~800 users).

### Relationships

Closes #3865 

## Why do we want this?
We want users to be able to update without problems.

## How is it done?
Removing the migration solves the problem because `IsGhost` was not used before 1.14.

### Why not another way?
If we don't do this, users need to reinstall the app and potentially lose data.

### Side effects
None.

## :squid: Permissions
Only me.


Toggl.PrimeRadiant.Realm/Database.cs
```diff
@@ -95,21 +95,9 @@ private RealmConfiguration createRealmConfiguration()
...
                     {
-                        var newWorkspaces = migration.NewRealm.All<RealmWorkspace>().ToList();
-                        var oldWorkspaces = migration.OldRealm.All("RealmWorkspace").ToList();
-                        for (var i = 0; i < newWorkspaces.Count; i++)
-                        {
-                            var oldWorkspace = oldWorkspaces[i];
-                            var newWorkspace = newWorkspaces[i];
-                            newWorkspace.IsInaccessible = oldWorkspace.IsGhost;
-                        }
+                        // nothing needs explicit updating when updating from schema 4 up to 7
```
simonrozsival: I think we must mention that we added `IsInaccessible` in 1.14 (we don't need to mention `IsGhost` because it was unused) but Realm handles it automatically.

Toggl.Foundation/Analytics/IAnalyticsService.cs
```diff
@@ -126,9 +126,9 @@ public interface IAnalyticsService
 
         IAnalyticsEvent<int> WorkspacesInaccesibleAfterCleanUp { get; }
 
-        IAnalyticsEvent<string, string, string> UnknownLoginFailure { get; }
+        IAnalyticsEvent<string, string> UnknownLoginFailure { get; }
 
-        IAnalyticsEvent<string, string, string> UnknownSignUpFailure { get; }
+        IAnalyticsEvent<string, string> UnknownSignUpFailure { get; }
```
simonrozsival: I would prefer having two PRs for these changes. It would document the bug better if we ever run into it. We might spend some time by figuring out what did analytics have to do with it.

Issue #3759, opened 2018-11-23
Fail instead of going into a retry loop


simonrozsival:
During sync it sometimes happen that the server gives us 429 errors and blocks requests for some time. In this case we are supposed to wait for 1 minute and then try again.

On iOS we have only 30 seconds to complete the background task. There is no time for waiting.

If we run into 429, we should simply call it a failed sync and wait for the next slot the OS gives us to try again.

Implementing this will require either hacking the state machine or modifying the way we handle retries. This needs more thought and discussion before we start implementing this feature.

simonrozsival:
The PR was merged into the feature branch.

PR #3876, opened 2018-11-30
Leaky bucket

simonrozsival:
## What's this?
This PR adds an implementation of a leaky bucket algorithm for HTTP requests sending.

### Relationships
Ref #3759

## Why do we want this?
We don't want to run into 429s from the server unless a different client is used with the same API token.

## How is it done?
The algorithm is quite simple. The goal is to prevent sending more than `slotsPerWindow` requests over a `movingWindowSize` time period.

We remember the times when the last `n` requests were sent ("slots were used") and check if the next request would be over the limit.

If the difference between the first and the last one is smaller than the `movingWindowSize` then we:
- ~wait a bit if the difference is in a specific (small tolerance~ - the waiting should be done by the caller
- ~throw~ return how long the caller should wait if the difference is too big

### Why not another way?
~I like this code.~ I improved the code based on Paul's feedback. I think it is a good way to implement it.

### Side effects
No side effects right now. We don't use it yet.

## :squid: Permissions
I will 🦑 this myself


amulware:
I think the bucket and api wrapper should be two separate classes. 

The bucket could have a single method with a signature something like:
```
TimeSpan/DateTime howLongToWaitBeforeSendingRequest = bucket.TryRegisterRequestInTheNext(Timespan.FromSeconds(0.5))
```

That should allow for using a simple lock inside the bucket, and generally simpler to understand code.

IF the request is not possible, either the bucket throws, or it returns null/false and the api wrapper can handle that and throw.

simonrozsival:
I updated the implementation, I also moved it into `Foundation` (it is not related to the requests so it has nothing to do in `Ultrawave` also after a recent internal discussion I won't put this into `Multivac`).

Now the class is sync and there are no hidden delays. It is also possible to check if one of the next `n` requests which would be made at the same time at parallel (i.e., when we pull data from the server) would fail so that we can avoid sending any of these requests if the batch would fail as a whole.

I found a NuGet package which implements almost exactly the same... see https://github.com/robertmircea/RateLimiters/blob/master/src/Bert.RateLimiters/RollingWindowThrottler.cs. I think it makes sense to still keep my code because it is tailored to our needs (although it works almost exactly the same).

PR #3878, opened 2018-11-30
Login manager observables

### LINKED ISSUE #3874: Add LoginManager observables to enable/disable background fetch
To enable background fetch we need to know whether a user is signed in. As proposed by @kornelijepetak, instead of injecting anything into the LoginManager, we would expose two observables from it: `UserLoggedIn` and `UserLoggedOut`. By subscribing to them in our service we'll be able to switch bg sync on and off.

dshr:
## What's this?
A small refactor to make all logins and logouts go through the login manager. It also adds two observables so we can subscribe to auth events.

### Relationships
Closes #3874

## Why do we want this?
Clarity. So we can only log out through on place.

## How is it done?
By moving parts of the logout logic into the LoginManager.

### Why not another way?
This is better lol

### Side effects
None.

## Review hints
Check if there's any other logout calls that I might've missed.

## :squid: Permissions
I guess only me, @kornelijepetak and @heytherewill


Toggl.Foundation/Login/LoginManager.cs
```diff
@@ -93,6 +102,13 @@ public IObservable<ITogglDataSource> SignUpWithGoogle(bool termsAccepted, int co
                 .SelectMany(_ => googleService.GetAuthToken())
                 .SelectMany(authToken => signUpWithGoogle(authToken, termsAccepted, countryId));
 
+        public IObservable<Unit> Logout()
+            => database.User
+                .Single()
+                .Select(dataSourceFromUser)
```
heytherewill: Q: Have you tried keeping the instance that's created when you call `Login` or `GetIfLoggedIn`?

dshr: Hm, I think I can, and can't see a corner case where the dataSource might be `null`, but for some reason it still feels a bit off.

PR #3879, opened 2018-11-30
Fail fast

simonrozsival:
## What's this?
This PR removes the handling mechanism of 5XX and 429 errors we've had so far. Instead of exponential back off we fail fast.

There is an assumption that we will handle these failures in the UI and we wouldn't allow users to force sync for some time and block syncing after pushing for a while so we don't overload the server in case it is being under heavy load.

### Relationships
Ref #3759 

## Why do we want this?
We need fail-fast for background sync. We have limited time to sync, there is no way we can wait a minute and then try again.

In the case of UX when the app is in the background it also makes little sense to wait for more than one minute. User's typically use our app for a very short amount of time. Once BG syncing is in place we also won't have to rely on the app being open to try pushing and pulling data again. We will "periodically" try syncing in background.

## How is it done?
🔥 🔥 🔥 

### Why not another way?
We had a long discussion in Slack and I expect there will be a discussion about this in the thread of this PR as well. There are other options, but this might be the easiest and most reasonable one we've had so far.

### Side effects
Right now user can pull-to-refresh many times and bury the server in request. Notice that the base branch is `project/background-sync` and that this won't be the last PR that touches this problem and it doesn't even close the issue.

## Review hints
Check the sync graph: [no-retry-loops.svg.zip](https://github.com/toggl/mobileapp/files/2633503/no-retry-loops.svg.zip)

## Tests
Syncing tests: https://app.bitrise.io/build/798c8e87d303cacc

## :squid: Permissions
Only me.


juanlaube:
I see why we need this, but I think it would be nice to have some sort of retry mechanism when the app is in foreground. This PR goes in the right direction if we won't care about retry anymore though.
The sync graph looks good also. There are still some broken tests.

simonrozsival:
As a conclusion to all the discussions we had I think the best way would be to:
1. fail fast, no retrying
2. disable syncing for the next 60 seconds
    - with a good visual (and haptic) feedback to the user
    - no fancy automatic exponential back-off when in foreground
3. user can repeat manually after the 60s elapses...
4. ...or we will try again when next bgsync is executed

I think it would make sense to have a retry loop when the leaky bucket throws because we know that the next request can be only made in < `x` (e.g., 20) seconds or some other "short" amount of time.

juanlaube:
When you say "disable syncing", this includes also the case triggered by the user (i.e: pull to refresh)? Showing "sync failed" or "try again later" is bad I think, if we don't allow this, we need something better to show.
The rest makes sense to me.

simonrozsival:
> When you say "disable syncing", this includes also the case triggered by the user (i.e: pull to refresh)?

Yes. No sync should be triggered either automatically or manually.

> Showing "sync failed" or "try again later" is bad I think, if we don't allow this, we need something better to show.

Yes, we need to change the we we communicate syncing progress and results to the user.

Issue #1958, opened 2018-04-25
Reports: Bar chart


pe0ny:
- Bar chart is displayed between the Summary and Pie chart cards
- Show max 45 bars on the x-axis.
- When period is <= 45 days, group by days.
- When period is > 45 days, group by weeks.
- When period is > 45 weeks, group by months (for example, when user picks random date range from the calendar that is longer than 45 weeks).
- When period is <= 7 days, display name of the weekday and date below each bar, for example: **M 1/4**. Dates should be formated based on the date format in Profile settings but without the year.
- When period is > 7 days, display only start and end dates, for example: **1/4 - 30/4**
- Bar chart padding from left is 12pt and from the right 40pt (from the card edge). For example, on SE the chart width will be 236pt.
- Chart’s height is fixed: 168pt
![screen shot 2018-04-25 at 14 10 24](https://user-images.githubusercontent.com/5064445/39242356-719dd8da-4892-11e8-97e3-54bdeb35e7d3.png)
- y-axis min value is 0 and max value is the max value of selected date range group value rounded up.
- Data is shown in two colors – dark #06AAF5 and light blue #47C3FC. The part that is dark blue shows billable hours. Light blue shows total hours. (Billable hours will display only if the WS is on the paid plan.)
- Show billable-nonbillable legend at the top right corner only if the WS is on the paid plan.

### Reports with the bar chart
![toggl - ios - reports - bar chart 1x](https://user-images.githubusercontent.com/5064445/39243668-f6b28b3e-4896-11e8-9487-2a76a91edef7.png)

Zeplin: https://zpl.io/blNMRRv

### Examples of yearly, monthly and weekly charts
![toggl - ios - reports - bar charts 1x](https://user-images.githubusercontent.com/5064445/39242611-46f57178-4893-11e8-9a1a-25c2b6fc25a5.png)

Zeplin: https://zpl.io/bJNGjjn

amulware:
This. Is. Looking. Dope.

Q: what is the math behind the width of the bars and margins between them?

For the implementer: to make sure it looks even, make sure to round (down) numbers, and have all bars and margins have the same INTEGER width, not different ones, or it'll look funny/off.

amulware:
Also, more specs on the y axis legend is needed. (0h is clear, MAXh is clear as well I think, but what about others, like the 5h one in the examples?)

pe0ny:
[...]

PR #3883, opened 2018-11-30
Giskard's reports bar chart

### LINKED ISSUE #3558: Reports: Bar chart
This is the Android counterpart of https://github.com/toggl/mobileapp/issues/1958

## Logic

See issue above. It's the same as iOS.

## Design

Very similar to iOS with fixed height, and fixed paddings to sides. Take exact values, colours, etc. from design below. 

https://zpl.io/2yBEpQL
screen shot 2018-10-25 at 13 22 52

daividssilverio:
## What's this?
This is the Android counterpart of #1958

### Relationships
Closes #3558

## Why do we want this?
Because 📊 📊 📊 are cool

## How is it done?
I've written the bar chart drawing logic with the android canvas o/
Basically, all the logic was already done in the view model, I just had to draw it and somehow add it to the reports header recycler view.

## Why wip?
The drawing isn't perfect, when there's too little billable/non-billable percentage, the bars look funny.
I'm trying to fix it, but otherwise all the other aspects can be reviewed. 

### Why not another way?
Why not simple views: Canvas is faster o/ and would be messy to to actually handle adding/removing views around.
Otherwise, this view isn't rxfactored yet, so I had to use work around Mvx for the moment, it should be easy to migrate though.

### Side effects
None.

## Review hints
Try it, it's cool!

## :squid: Permissions
🦑 it, please, after the WIP label get's removed.


Toggl.Giskard/Activities/ReportsActivity.cs
```diff
@@ -46,9 +47,26 @@ protected override void OnCreate(Bundle bundle)
...
+            Observable.CombineLatest(
+                    ViewModel.StartDate,
+                    ViewModel.EndDate,
+                    ViewModel.WorkspaceHasBillableFeatureEnabled,
+                    ViewModel.BarChartViewModel.DateFormat,
+                    ViewModel.BarChartViewModel.Bars,
+                    ViewModel.BarChartViewModel.MaximumHoursPerBar,
+                    ViewModel.BarChartViewModel.HorizontalLegend,
+                    (startDate, endDate, workspaceIsBillable, dateFormat, bars, maximumHoursPerBar, horizontalLegend) => new BarChartData(startDate, endDate, workspaceIsBillable, dateFormat, bars, maximumHoursPerBar, horizontalLegend))
+                .Subscribe(updateBarChart)
```
heytherewill:
```suggestion
                .Subscribe(reportsRecyclerView.ReportsRecyclerAdapter.UpdateBarChart)
```

Toggl.Giskard/Views/BarChartView.cs
```diff
@@ -0,0 +1,188 @@
...
+using Toggl.Foundation.MvvmCross.ViewModels.Reports;
+using Toggl.Giskard.Extensions;
+using Toggl.Giskard.ViewHelpers;
+
+namespace Toggl.Giskard.Views
+{
+    [Register("toggl.giskard.views.BarChartView")]
+    public class BarChartView : View
+    {
+        private Color nonBillableColor = Color.ParseColor("#1e328fff");
```
heytherewill: Why not make this static?

Toggl.Giskard/Views/BarChartView.cs
```diff
@@ -0,0 +1,188 @@
...
+            initialize(context);
+        }
+
+        protected BarChartView(IntPtr javaReference, JniHandleOwnership transfer) : base(javaReference, transfer)
+        {
+        }
+
+        private void initialize(Context context)
+        {
+            maxWidth = 48.DpToPixels(context);
```
heytherewill: Shouldn't all these be readonly?

Toggl.Giskard/Views/BarChartView.cs
heytherewill: Why not make this readonly?

Toggl.Giskard/Views/BarChartView.cs
```diff
@@ -0,0 +1,188 @@
...
+        }
+
+        protected override void OnDraw(Canvas canvas)
+        {
+            if (!barChartData.HasValue || barsCount == 0) return;
+
+            var barWidth = Math.Min(barsWidth / barsCount, maxWidth);
+            var spaces = barsCount - 1;
+            var totalWidth = barWidth * barsCount;
+            var remainingWidth = barsWidth - totalWidth;
```
heytherewill: There's quite a lot of math being done here. I think you should instead pre-calculate this before calling Invalidate and just use the values that were pre-calculated

Toggl.Giskard/Views/BarChartView.cs
```diff
@@ -0,0 +1,188 @@
...
+                    canvas.DrawText(horizonalLabels[barIndex].Date, middleOfTheBar, barsBottom + bottomLabelMarginTop * 1.25f + textSize, othersPaint);
+                }
+
+                left += actualBarWidth + spacing;
+            }
+        }
+
+        private float defaultBarSpacingRatio = 0.2f;
+
+        private float barsWidth => Width - barsLeftMarging - barsRightMarging;
```
heytherewill: Please don't private property. This hides the actual cost of calling this (because from the caller's site it looks like a field). Use a method (or even better, cache this).

Toggl.Giskard/Views/BarChartView.cs
```diff
@@ -0,0 +1,188 @@
...
+                    var middleOfTheBar = left + (barRight - left) / 2f;
+                    canvas.DrawText(horizonalLabels[barIndex].DayOfWeek, middleOfTheBar, barsBottom + bottomLabelMarginTop * 1.25f, othersPaint);
+                    canvas.DrawText(horizonalLabels[barIndex].Date, middleOfTheBar, barsBottom + bottomLabelMarginTop * 1.25f + textSize, othersPaint);
+                }
+
+                left += actualBarWidth + spacing;
+            }
+        }
+
+        private float defaultBarSpacingRatio = 0.2f;
```
heytherewill: This should be a const (and it should be at the beginning of the file).

Toggl.Giskard/Views/BarChartView.cs
daividssilverio: I need the context to initialize them. I don't have the context to initialize them statically. Having all of these initializations in each constructor doesn't sound good as well.

Toggl.Giskard/Views/BarChartView.cs
```diff
@@ -80,47 +107,82 @@ private void initialize(Context context)
...
+        }
 
-            var barWidth = Math.Min(barsWidth / barsCount, maxWidth);
-            var spaces = barsCount - 1;
-            var totalWidth = barWidth * barsCount;
-            var remainingWidth = barsWidth - totalWidth;
-            var spacing = Math.Max(remainingWidth / barsCount, barWidth * defaultBarSpacingRatio);
+        private void updateBarChartDrawingData()
+        {
+            if (barChartData.HasValue)
```
heytherewill: Invert this comparison so you can early return instead of adding another layer of nesting on this already super complex method:

```csharp
if (barChartData == null)
{
    willDrawBarChart = false
    return;
} 
```

Toggl.Giskard/Views/BarChartView.cs
```diff
@@ -80,47 +107,82 @@ private void initialize(Context context)
...
+            {
+                willDrawBarChart = false;
+            }
+        }
 
-            var requiredWidth = totalWidth + spaces * spacing;
-            var actualBarWidth = requiredWidth > barsWidth ? barWidth * (1 - defaultBarSpacingRatio) : barWidth;
+        protected override void OnDraw(Canvas canvas)
+        {
+            if (!willDrawBarChart) return;
```
heytherewill: You don't even wanna call `base.OnDraw(canvas)` here?

Toggl.Giskard/Views/BarChartView.cs
```diff
@@ -137,15 +199,15 @@ protected override void OnDraw(Canvas canvas)
                 }
                 else
                 {
-                    var billableTop = (float) ((Height - barsTopMargin - barsBottomMargin) * (1f - bar.BillablePercent) + barsTopMargin) + 1;
+                    var billableTop = (float) (barsHeight * (1f - bar.BillablePercent) + barsTopMargin) + 1;
                     billableTop = barHasBillablePercentage && barsBottom + 1 - billableTop < 1f ? billableTop - 1 : billableTop;
```
heytherewill: This is *very* cryptic. Why don't you:

a) Encapsulate this into a method
b) Use two adequately named variables

instead of reassigning a variable?

Toggl.Giskard/Views/BarChartView.cs
```diff
@@ -137,15 +199,15 @@ protected override void OnDraw(Canvas canvas)
...
                     canvas.DrawRect(left,
                         billableTop,
                         barRight,
                         barsBottom + 1,
                         billablePaint);
 
-                    var nonBillableTop = (float) ((Height - barsTopMargin - barsBottomMargin) * (1f - bar.BillablePercent - bar.NonBillablePercent) + barsTopMargin) + 1;
+                    var nonBillableTop = (float) (barsHeight * (1f - bar.BillablePercent - bar.NonBillablePercent) + barsTopMargin) + 1;
                     nonBillableTop = barHasNonBillablePercentage && billableTop - nonBillableTop < 1f ? nonBillableTop - 1 : nonBillableTop;
```
heytherewill: Same here

Toggl.Giskard/Views/BarChartView.cs
```diff
@@ -137,15 +199,15 @@ protected override void OnDraw(Canvas canvas)
                 }
                 else
                 {
-                    var billableTop = (float) ((Height - barsTopMargin - barsBottomMargin) * (1f - bar.BillablePercent) + barsTopMargin) + 1;
+                    var billableTop = (float) (barsHeight * (1f - bar.BillablePercent) + barsTopMargin) + 1;
                     billableTop = barHasBillablePercentage && barsBottom + 1 - billableTop < 1f ? billableTop - 1 : billableTop;
                     canvas.DrawRect(left,
                         billableTop,
                         barRight,
                         barsBottom + 1,
```
heytherewill: Why the `+ 1`?

Toggl.Giskard/Views/BarChartView.cs
```diff
@@ -137,15 +199,15 @@ protected override void OnDraw(Canvas canvas)
                 }
                 else
                 {
-                    var billableTop = (float) ((Height - barsTopMargin - barsBottomMargin) * (1f - bar.BillablePercent) + barsTopMargin) + 1;
+                    var billableTop = (float) (barsHeight * (1f - bar.BillablePercent) + barsTopMargin) + 1;
```
heytherewill: There's a fair share of magic numbers going on here and I don't like them at all. Also, you're still doing a fair share of math `onDraw`

Toggl.Giskard/Views/BarChartView.cs
```diff
@@ -80,47 +107,82 @@ private void initialize(Context context)
...
+            canvas.DrawLine(0f, middleHorizontalLineY, Width, middleHorizontalLineY, othersPaint);
             canvas.DrawLine(0f, barsBottom, Width, barsBottom, othersPaint);
 
             othersPaint.Color = hoursTextColor;
-            var hoursX = Width - barsRightMarging + textLeftMargin;
-            var hoursBottomMargin = textBottomMargin * 2f;
-            canvas.DrawText("0 h", hoursX, barsBottom - hoursBottomMargin, othersPaint);
-            canvas.DrawText(halfHours, hoursX, barsHeight / 2f - hoursBottomMargin, othersPaint);
-            canvas.DrawText(maxHours, hoursX, barsTopMargin - hoursBottomMargin, othersPaint);
+            canvas.DrawText("0 h", hoursLabelsX, barsBottom - hoursBottomMargin, othersPaint);
```
heytherewill: Pls no literal strings, as we're moving away from that due to the i18n effort.

Toggl.Giskard/Views/BarChartView.cs
daividssilverio: 1. -> (1f - percentage) is remaining percentage, that isn't magic.
2. the +1 I'll rename for clarity, but it's so drawRect draws over the first horizontal line as well.
3. Now about the remaining math, it's a tradeoff, all the math left requires looping through the bars, to remove them I'd have to create some iterable every time the data changes. And OnDraw isn't running all the time.

Toggl.Giskard/Views/BarChartView.cs
daividssilverio: So drawRect draws over the bottom horizontal line and the bars don't look funny when there's little data.

Toggl.Giskard/Views/BarChartView.cs
daividssilverio: Why though?

PR #3891, opened 2018-12-03
📊 move tracking to UserAccessManager

### LINKED ISSUE #3888: Move tracking into UserAccessManager
🔑 Related to `user-access-improvements` project.

The event tracking should be as close to the point of event origin, and we are currently tracking access control things in the view models. This makes sense for some events, but events related to the users' access should be inside the `UserAccessManager`. A view model is just a model for the view. Authentication/Authorization process has nothing to do with views.

This will also work towards tests making more sense because tests related to the users' access will be closeby, grouped under tests for the `UserAccessManager`.

So, events related to the auth events (login, signup, logout) should be done inside the login manager.

#### **Precursors:**
Doing #3884 first will make the merge of this issue much easier.

khoi:
## What's this?
Move tracking to UAM instead of the VMs

### Relationships
Closes #3888

## Why do we want this?
The reasoning is stated in the issue.

## How is it done?
Inject IAnalyticService in to the UAM 

### Side effects
Not that I know of

## Review hints
None

## :squid: Permissions
🔑


Toggl.Foundation.MvvmCross/ViewModels/LoginViewModel.cs
```diff
@@ -264,15 +262,13 @@ private void onError(Exception exception)
 
             switch (exception)
             {
-                case UnauthorizedException forbidden:
+                case UnauthorizedException _:
```
kornelijepetak: Since `_` in most cases in C# does not have any semantics, I'd rather name this `exception` or `ex`.

Toggl.Foundation/UserAccessManager/UserAccessManager.cs
```diff
@@ -181,7 +186,8 @@ private IObservable<IUser> signUp(Email email, Password password, bool termsAcce
                 .CreateApiWith(Credentials.None)
                 .User
                 .SignUp(email, password, termsAccepted, countryId)
-                .Track(analyticsService.SignUp, AuthenticationMethod.EmailAndPassword);
+                .Track(analyticsService.SignUp, AuthenticationMethod.EmailAndPassword)
+                .Do(CommonFunctions.DoNothing, doSignupError);
```
kornelijepetak: I'd go with `onSignupError`

Toggl.Foundation/UserAccessManager/UserAccessManager.cs
```diff
@@ -75,6 +78,7 @@ public IObservable<ITogglDataSource> Login(Email email, Password password)
                 .SelectMany(database.User.Create)
                 .Select(dataSourceFromUser)
                 .Track(analyticsService.Login, AuthenticationMethod.EmailAndPassword)
+                .Do(CommonFunctions.DoNothing, doOnLoginError)
```
kornelijepetak: Just `onLoginError`...

PR #3893, opened 2018-12-03
Make 'none' the default trigger for rating dialog to show

### LINKED ISSUE #3892: The default settings for the user rating dialog is not correct
The correct default behaviour for the rating view trigger is `none`. 

Currently it's set to `stop`, effectively showing the rating view on TE stop even when Firebase says not to.

The defaults are selected when Firebase fails to provide the values (possible because of connectivity issues or similar.)

kornelijepetak:
## What's this?
Changes the default trigger for rating dialog to `none`.

### Relationships
Closes #3892 

## Why do we want this?
Because if Firebase fails, the users will see the rating dialog.

## How is it done?
Setting the defaults in xml.

## :squid: Permissions
Whoever.


amulware:
Not into `develop` - it's hotfix o' clock.

kornelijepetak:
Invalid PR. Branched from wrong point in history.

PR #3896, opened 2018-12-03
Introduce suggestion certainties

### LINKED ISSUE #3807: Introduce suggestion certainty/confidence/probability
A better version of #3789

When generating suggestions, the providers will need to calculate a value between 0 and 1, which indicates, how strongly the provider believes in the suggestion.

For example, calendar suggestion provider sees that there are 2 events coming up. The one that starts sooner should get a higher value (closer to one) than the other one. Similarly, most used time entry suggestion provider could find that time entry A has been used 40 times, but time entry B has been used 30 times. In this case time entry A should have a higher certainty.

A thing we have to look out is cases when different providers yield suggestions with matching probabilities. That would create ambiguity in determining which suggestion is more important. For that reason we should avoid using the value 1 and calculate the value based on some parameters that are gonna be different for each suggestion provider.

For this PR you should:
* Add the property to `Suggestion`, making sure that this property is required. In other words: there shouldn't be a way to create a `Suggestion` without that property set.
* Implement the calculation for `MostUsedTimeEntrySuggestionProvider`

arturs105:
### Relationships

Closes #3807

## :squid: Permissions
🦑


Toggl.Daneel/Toggl.Daneel.csproj
```diff
@@ -1267,8 +1267,7 @@
   <Import Project="..\packages\Xamarin.Firebase.iOS.Core.4.0.13.1\build\Xamarin.Firebase.iOS.Core.targets" Condition="Exists('..\packages\Xamarin.Firebase.iOS.Core.4.0.13.1\build\Xamarin.Firebase.iOS.Core.targets')" />
   <Import Project="..\packages\Xamarin.Firebase.iOS.InstanceID.2.0.8\build\Xamarin.Firebase.iOS.InstanceID.targets" Condition="Exists('..\packages\Xamarin.Firebase.iOS.InstanceID.2.0.8\build\Xamarin.Firebase.iOS.InstanceID.targets')" />
   <Import Project="..\packages\Xamarin.Firebase.iOS.Analytics.4.0.5.1\build\Xamarin.Firebase.iOS.Analytics.targets" Condition="Exists('..\packages\Xamarin.Firebase.iOS.Analytics.4.0.5.1\build\Xamarin.Firebase.iOS.Analytics.targets')" />
-  <Import Project="..\packages\Xamarin.Firebase.iOS.CrashReporting.2.0.0.5\build\Xamarin.Firebase.iOS.CrashReporting.targets" Condition="Exists('..\packages\Xamarin.Firebase.iOS.CrashReporting.2.0.0.5\build\Xamarin.Firebase.iOS.CrashReporting.targets')" />
```
heytherewill: 👀

arturs105: This was causing an error while building. This change fixed that error. Also: there is nothing related to CrashReporting on `develop`.

PR #3903, opened 2018-12-03
Add storage for dismissed suggestions

### LINKED ISSUE #3846: Add storage for keeping track of dismissed suggestions
We should remember which suggestions are dismissed and then filter the new suggestions based on those stored so we don't show the same dismissed suggestion several times.

This should happen in memory, since we don't want to ignore dismissed suggestions until the end of time, because a suggestion that was dismissed on the previous day, might be useful the next day.

arturs105:
### Relationships

Closes #3846 

## :squid: Permissions
🦑


Toggl.Foundation/Suggestions/DismissedSuggestionStorage.cs
```diff
@@ -0,0 +1,36 @@
...
+
+            public HashableSuggestion(Suggestion suggestion)
+            {
+                Description = suggestion.Description;
+                ProjectId = suggestion.ProjectId;
+                TaskId = suggestion.TaskId;
+            }
+
+            public override int GetHashCode()
+                => $"{Description}/{ProjectId}/{TaskId}".GetHashCode();
```
simonrozsival: Use `Toggl.Multivac.HashCode`.

Issue #3805, opened 2018-11-26
Do not run bg sync when the user is not logged in


simonrozsival:
We should cancel any further sync when the user logs out of the app.

dshr:
Closed in #3904

PR #3904, opened 2018-12-03
Naive Background sync on iOS

### LINKED ISSUE #3761: Schedule background app refresh
Use [Background App Refresh](https://developer.apple.com/documentation/uikit/core_app/managing_your_app_s_life_cycle/preparing_your_app_to_run_in_the_background/updating_your_app_with_background_app_refresh) to be able to run code in background.

We should not sync in bg before the user logs in or signs up. If the app is started and the user is already sign in, we should also start syncing in bg.

dshr:
## What's this?
This implements background sync on iOS. We schedule our refreshes every 15 minutes, and do a full sync every time. Our syncs are only scheduled if the user is logged in.

### Relationships
Closes #3761, #3805 and #3765

## Why do we want this?
For background sync.

## How is it done?
By using iOS' Background Fetch to call a custom background sync interactor.

### Why not another way?
This seems the easiest.

### Side effects
None.

## Review hints
Not really sure of a proper way to review this..

## Tests
Does not change anything in our integration tests, but we might have to.

## :squid: Permissions
Background sync people only, please.


Toggl.Daneel/Startup/AppDelegate.cs
```diff
@@ -303,5 +306,35 @@ private void startTimeEntryInBackground(string eventId, Action completionHandler
...
 
         #endregion
+
+        #region Background Sync
+
+        public override void PerformFetch(UIApplication application, Action<UIBackgroundFetchResult> completionHandler)
+        {
+            var interactorFactory = Mvx.Resolve<IInteractorFactory>();
+            if (interactorFactory == null) return;
+            CompositeDisposable disposeBag = new CompositeDisposable();
```
simonrozsival: We don't need composite disposable for a single `IDisposable` 😄

Toggl.Daneel/Startup/AppDelegate.cs
```diff
@@ -303,5 +306,35 @@ private void startTimeEntryInBackground(string eventId, Action completionHandler
...
+            interactorFactory
+                .RunBackgroundSync()
+                .Execute()
+                .Subscribe(outcome =>
+                {
+                    switch (outcome)
+                    {
+                        case Foundation.Models.SyncOutcome.NewData:
+                            completionHandler(UIBackgroundFetchResult.NewData);
```
simonrozsival: What if we used a `.Select(mapToNativeOutcomes).Do(completionHandler).Subscribe();`?

Toggl.Foundation/Services/BaseBackgroundSyncService.cs
```diff
@@ -0,0 +1,23 @@
...
+using Toggl.Foundation.Login;
+using Toggl.Multivac.Extensions;
+
+namespace Toggl.Foundation.Services
+{
+    public abstract class BaseBackgroundSyncService : IBackgroundSyncService
+    {
+        public void SetupBackgroundSync(LoginManager loginManager)
+        {
+            loginManager.UserLoggedIn
```
simonrozsival: We need to keep the disposable, because if it is garbage collected (it will be at some point), it would unsubscribe and this would stop updating.

Toggl.Foundation/Services/BaseBackgroundSyncService.cs
```diff
@@ -0,0 +1,23 @@
...
+using System.Reactive;
+using System.Reactive.Linq;
+using Toggl.Foundation.Login;
+using Toggl.Multivac.Extensions;
+
+namespace Toggl.Foundation.Services
+{
+    public abstract class BaseBackgroundSyncService : IBackgroundSyncService
+    {
+        public void SetupBackgroundSync(LoginManager loginManager)
```
simonrozsival: Could this be a constructor? Or Maybe otherwise - could the `Enable*` and `Disable*` methods be part of the interface and the subscriptions could be created elsewhere? I'm not sure now :) Maybe this is good enough.

Toggl.Foundation/Services/BaseBackgroundSyncService.cs
simonrozsival: The the whole class should be an `IDisposable` and it should dispose the internal disposables correctly.

heytherewill:
I'll wait until this is merged to submit the Android PR; It's better if I base it on a working solution

Toggl.Foundation/Services/BaseBackgroundSyncService.cs
dshr: It can be a constructor and make the IBackgroundSyncService an empty interface, at least for now. Depending on how the actual sync will work, and whether we might end up with separate SyncManagers and with extra methods to deal with that that will go in the interface.

dshr: NVM, forgot that BaseBackgroundSyncService is abstract and can't have a constructor, but we want the `Enable*` and `Disable*` methods to be abstract.

Toggl.Foundation/Services/BaseBackgroundSyncService.cs
```diff
@@ -1,23 +1,35 @@
...
 using Toggl.Foundation.Login;
 using Toggl.Multivac.Extensions;
 
 namespace Toggl.Foundation.Services
 {
-    public abstract class BaseBackgroundSyncService : IBackgroundSyncService
+    public abstract class BaseBackgroundSyncService : IBackgroundSyncService, IDisposable
     {
+        private IDisposable loggedInDisposable;
+        private IDisposable loggedOutDisposable;
```
simonrozsival: Here, on the other hand, it would make sense to use the composite disposable - we don't dispose these individually 😄

Issue #3902, opened 2018-12-03
Implement new SignupOrLoginChoiceViewModel


kornelijepetak:
🔑 Related to `user-access-improvements` project.

The main landing page that user will see if the app is started without being logged in.
From that page, user will be able to either log in, or sign up.

For now, I'll call it `SignupOrLoginChoice`, but feel free to propose a better alternative.

However it is called, we also need a view model that will facilitate the required behavior from that page.

PR #3909, opened 2018-12-04
🎨 SignupOrLoginChoice - iOS Implementation

khoi:
## What's this?
SignupOrLoginChoiceViewController

### Relationships

Part of #3902

## How is it done?
Adding new VC

### Side effects
None.

## Review hints
Nada.

## :squid: Permissions
🔑 after #3908 is merged


Toggl.Daneel/ViewControllers/UserAccess/SignupOrLoginChoiceViewController.cs
```diff
@@ -0,0 +1,35 @@
...
+        {
+        }
+
+        public override void ViewDidLoad()
+        {
+            base.ViewDidLoad();
+
+            SignUpButton.Layer.BorderWidth = (nfloat)0.5;
+            SignUpButton.Layer.BorderColor = UIColor.Black.CGColor;
+            SignUpButton.Layer.MasksToBounds = true;
```
heytherewill: Can't these be set on the .xib?

khoi: No, unfortunately, the `BorderColor` has to be set in code since it's a CGColorRef. In xib only UIColor can be set.

Odrakir: And anyway user defined runtime properties in Xibs are so hard to find. It's much better in code like it is.

PR #3914, opened 2018-12-04
Rename message property in ResponseError

### LINKED ISSUE #3877: Rename api error response message
It was agreed internally that the future format of error messages will be as follows.
```
{
  "error_code" : "FAILED_TO_DO_STUFF(format of codes is TBD)",
  "error_message" : "Stuff failed due to 2 being smaller than 5 (this message could be shown to the user if no localised message is available)",
  "parameters" : {
    "number1" : 2,
    "number2" : 5
  }
}
```

For minimal forwards compatibility, we need to ensure the message property in `ResponseError` is parsed correctly.
https://github.com/toggl/mobileapp/blob/develop/Toggl.Ultrawave/Models/ResponseError.cs#L1-L9

We could either add a Json.Net attribute to make it parse from `error_message` or if snake-case parsing works correctly, we could rename the property itself to `ErrorMessage`.

I'll leave that up to you. In either case, just make sure (test if we don't do so yet) that once we get error messages of the format above the message property will parse correctly.

We can add the other properties down the line, when they become necessary.

juanlaube:
## What's this?
Renames the `Message` property in `ResponseError` to ensure forward compatibility with the new API error format.

### Relationships

Closes #3877

## Why do we want this?
We want to ensure error messages are displayed correctly when backend introduces the new JSON formatted errors, introducing this change now ensures a minimal forward compatibility.

## How is it done?
Renaming the `Message` property to `ErrorMessage`, following the format that was agreed between all clients and backend.

### Why not another way?
There's no other way.

### Side effects
~I need to confirm if backend is already sending errors as JSON with the "message" key right now. If that's the case, we need to ensure backward compatibility with that value too! 🤔 ~
None. Backend is not sending any errors as JSON right now.

## :squid: Permissions
~🚫 Don't think about it until we can answer that concern I described in side effects.~
🦑


daividssilverio:
🤔 I guess you meant "error_message" in the Side Effects.
This looks alright. You talked about keeping backward compatibility, but we don't have to do anything about it, do we? Backend has to keep the old "message" so it doesn't break old clients, that's it, right?

Issue #3907, opened 2018-12-04
Implement new or update/refactor existing ForgotPasswordViewModel


kornelijepetak:
🔑 Related to `user-access-improvements` project.

As the title says, we need that.
The current forgot password could be useful or reused/refactored into what we need, or a new one can be created. Use your best judgement to determine which course to take.

Duties:

* [x] Providing events and (error) messages for the UI 
* [x] Taps into `UserAccessManager` to request a password reset

PR #3916, opened 2018-12-05
🔘 add new BindAction for RxAction<TInput, TElem>

khoi:
## What's this?

add a new extension that could be used to bind UIButton to any `RxAction<TInput, TElement>`

### Relationships
No issue created. I need this new extension for #3907

## Why do we want this?

Currently, `UIButton`'s `BindAction` can only bind to either

- `UIAction : RxAction<Unit, Unit>`
- `InputAction<TInput> : RxAction<TInput, Unit>`

We need a way to bind the button to an `RxAction` that produces a result other than `Unit`. For instance:

```cs
struct Result {
  bool Success
  string ErrorMessage
}

class VM {
  public RxAction<Input, Result> SomeAction
}

class VC {
  void ViewDidLoad {
      SomeButton.BindAction(VM.SomeAction, _ => Unit.Default).DisposedBy(SomeBag)
  }
}

```

## How is it done?
Add

```csharp
public static IDisposable BindAction<TInput, TElement>(this IReactive<UIButton> reactive, RxAction<TInput, TElement> action, Func<UIButton, TInput> inputConvert, ButtonEventType eventType = ButtonEventType.Tap)
```

and refactor the other two extensions to call this method.

### Why not another way?
🤔 I'm unaware of another way.

### Side effects
Nope.

## Review hints
None.

## :squid: Permissions
🐙


khoi:
> Should we add this to UIView too? Not sure if we'll ever need it... just wondering
We'll add it in when we need it, for now UIButton I think is enough

Toggl.Daneel/Extensions/Reactive/UIButtonExtensions.cs
```diff
@@ -41,27 +39,20 @@ public static Action<string> AnimatedTitle(this IReactive<UIButton> reactive)
...
+            reactive.BindAction(action, _ => Unit.Default, eventType);
 
-            return Observable.Using(
-                    () => action.Enabled.Subscribe(e => { reactive.Base.Enabled = e; }),
-                    _ => eventObservable
-                )
-                .Subscribe(action.Inputs);
-        }
+        public static IDisposable BindAction<TInput>(this IReactive<UIButton> reactive, InputAction<TInput> action,
+            Func<UIButton, TInput> convert, ButtonEventType eventType = ButtonEventType.Tap) =>
```
Odrakir: Rename `convert` to `inputTransform`.

khoi: oops, totally missed that

PR #3924, opened 2018-12-05
Ensure Mvx is initialized to avoid null refs

### LINKED ISSUE #3282: Mvx.Resolve[TService]() — Object reference not set to an instance of an object (again! 💥)
## 💥 Crash Report
Got it from AppCenter. Looks a constructor problem with the Android Logger.

### 👥 Impact
Has happened 372 times so far to 223 users on 1.3.2 over the last 90 days.

### 🔗 Context
Seems to be distantly related to https://github.com/MvvmCross/MvvmCross/issues/2342.

### 🛰 Link
https://appcenter.ms/orgs/Toggl/apps/Toggl-Android/crashes/groups/ba8a7f84f206d5023307b46e79767fc49d76ca08?appVersion=1.3.2&period=last90Days

```cs
    Mvx.Resolve[TService] ()
    MvvmCross.MvxAndroidLog..cctor () <0x7f87e45a24 + 0x0001f> in :0
```

juanlaube:
## What's this?
This is a possible fix for a crash that I couldn't reproduce. The crash in question, according to the stack trace, happens when executing a class constructor inside Mvx.
According to [this issue in the Mvx repo](https://github.com/MvvmCross/MvvmCross/issues/3115), this happens when using AOT compilation and a possible workaround is given there.

### Relationships

Ref #3282

## Why do we want this?
🚫 💥 

## How is it done?
By ensuring Mvx is initialized properly. I added this snippet in `TogglApplication.OnCreate` before calling `base`, this should be the first thing that runs in the app and initialize Mvx properly.

```
var setup = MvxAndroidSetupSingleton.EnsureSingletonAvailable(ApplicationContext);
setup.EnsureInitialized();
```

### Why not another way?
We don't have any other clue on this crash, so far this is the only thing we can try.

### Side effects
None that I'm aware of.

## :squid: Permissions
Only if @daividssilverio and @heytherewill ✅


heytherewill:
The fix is reasonable. Can you run a couple of tests using stopwatches to check if this is making our startup any slower?

juanlaube:
Initialization in `TogglApplication.OnCreate` adds 3.5 seconds to startup time., adding this extra time to all users is not acceptable for a crash that cannot be reproduced and affects only a few users.
The initialization check is now in `MainActivity.OnCreate`, which might be too late to fix the crash or not, we cannot be sure unless we stop seeing the crash in AppCenter.

daividssilverio:
@juanlaube Please note that if you don't do this on TogglApplication, you'll have to do it on every activity (or the all base activities), because the app can be restored from some other point other than the MainActivity. Unless I'm terrible wrong, @heytherewill might think differently and surely knows more about Mvx.

Issue #3932, opened 2018-12-06
Use leaky bucket in syncing algorithm to prevent 429s


simonrozsival:
We know the exact settings of rate limiting on our server, but we are not supposed to rely on them as they can be changed at any time when needed without any heads-up. We know that the safe limits are 1 req/s and it will never be more restricted than that. We can try being a bit more ambitious though, because 1 req/s (! not exactly the same as 60 req/min) is very limiting for parallel requests. 😄 

For now it would be easiest to do the same thing as the backend team:
- limit to 3 req/s
- limit to 60 req/min

I suggest we create a wrapper around the `IApiClient` which delays sending a request using a leaky bucket with 3req/s settings lets it in (this will be used throughout the whole app - even for login/signup and reports requests). This wrapper will help us prevent 429s as results of big bursts of requests.

For sending requests during syncing (both pull and push) we will use another leaky bucket which will limit us to 60s/min. This will mostly give us a way to pull data without running into 429s. We make 9 requests when we pull data and we want to make sure that none of them fails. This should also help us if the limits are changed (lowered).

simonrozsival:
This is finally done.

PR #3935, opened 2018-12-06
Use leaky bucket for pulling

simonrozsival:
## What's this?
This PR starts using leaky bucket for pulling data from the server.

### Relationships
Ref #3932

This PR contains some bits used in #3934 which targets `develop` and so the diff is a bit cluttered until the other PR isn't merged and `project/background-sync` branch isn't updated.

## Why do we want this?
We want to minimize the number of 429s we are getting.

## How is it done?
The `FetchAllSinceState` checks if it could use 9 requests without overflowing the bucket and if everything is OK, then it proceeds. Otherwise it waits for the specified time given by the leaky bucket and then it repeats. The `DelayState` is used to implement this delayed loop.

_Note 1:_ I decided not to inject the leaky bucket into the configuration itself, but instead define it in the factory. This way we will always use the same leaky bucket no matter if it runs on iOS, Android, or sync tests.

_Note 2:_ I had to make a small modification to the leaky bucket class (created in #3876). It was necessary to make it behave more like "compare exchange" to make it threadsafe.

### Side effects
No side effects.

## Tests
Syncing tests: https://app.bitrise.io/build/aa8159a5110e9025

## :squid: Permissions
🦑 🇮🇹


simonrozsival:
Now the diff is uncluttered, but I think I will make one more change to the PR and I will ask you for a second review later.

simonrozsival:
I slept on it and now I think it is good as it is, I won't make any more changes in this PR.

juanlaube:
I re-reviewed after the merge back from `project/background-sync` and this looks good!

PR #3936, opened 2018-12-06
Modified leaky bucket

simonrozsival:
## What's this?
I had to make a small modification to the leaky bucket class (created in #3876). It was necessary to make it behave more like "compare exchange" to make it threadsafe.

### Relationships
Ref #3932

## Why do we want this?
We want to avoid race conditions and incorrect behavior of the leaky bucket.

## How is it done?
Put everything under one lock + rename stuff.

### Side effects
None, it's still not used anywhere at the moment.

## :squid: Permissions
🦑


juanlaube:
This PR can be closed, right?

PR #3937, opened 2018-12-06
Rename VoidSubscribe extension

heytherewill:
## What's this?
This removes the `Void` prefix from the subscribe extension

### Relationships
This is a followup of #3929

## Why do we want this?
The void only existed because there was a `Task` counterpart. Now it's no longer needed

## How is it done?
Refactor tools®

## :squid: Permissions
Anyone, as ASAP as possible


juanlaube:
_waiting for bitrise to 🦑_

PR #3944, opened 2018-12-07
Add Android JobScheduler service

### LINKED ISSUE #3763: Create sync manager when app is started in background
Implment this in similar way as the iOS implementation: #3762

### LINKED ISSUE #3767: Schedule background job
Use [JobScheduler](https://developer.android.com/reference/android/app/job/JobScheduler) to run Android app in background.

Make sure our background sync is run with network connection.

Only schedule the job after the user logs in, signs up, or starts the app when he/she is already logged in.

### LINKED ISSUE #3768: Trigger background sync interactor
Trigger the interactor from #3756, await its result and call the callback with appropriate values.

### LINKED ISSUE #3806: Do not run bg sync when the user is not logged in
We should cancel any scheduled sync when user logs out of the app.

heytherewill:
## What's this?
This adds background sync® for Android

### Relationships
Closes #3763
Closes #3767
Closes #3768
Closes #3806

## Why do we want this?
Background sync project

## How is it done?
We are using the JobScheduler to run things while the app is not in foreground. More detailed info in the upcoming docs.

### Why not another way?
Since we support api 21, that's the most solid way

## Review hints
Add a bunch of Logs to the JobScheduler (and save the logs in the shared preferences) if you don't believe this works or wanna verify the periodicity

## :squid: Permissions
Sync bois®


Toggl.Giskard/Extensions/ActivityExtensions.cs
```diff
@@ -51,6 +52,24 @@ public static (int widthPixels, int heightPixels, bool isLargeScreen) GetMetrics
             return (displayMetrics.WidthPixels, displayMetrics.HeightPixels, isLargeScreen);
         }
 
+        public static JobInfo CreateJobInfoFor<TJobScheduler>(this Context context)
+            where TJobScheduler : JobService
+        {
+            var javaClass = Java.Lang.Class.FromType(typeof(TJobScheduler));
+            var component = new ComponentName(context, javaClass);
+
+            var builder = new JobInfo.Builder(1, component)
```
simonrozsival: What does `1` mean? Either make it a constant or a named parameter.

Toggl.Giskard/Extensions/ActivityExtensions.cs
```diff
@@ -51,6 +52,24 @@ public static (int widthPixels, int heightPixels, bool isLargeScreen) GetMetrics
...
+                // This job can run on any network, we dont care
+                .SetRequiredNetworkType(NetworkType.Any)
+                // It will happen once every hour, but not more than that
+                .SetPeriodic((long)TimeSpan.FromHours(1).TotalMilliseconds)
```
simonrozsival: We decided to go with 15 min on iOS, so I suggest we unify those constants. Maybe we could put that into some shared constant?

Toggl.Giskard/Startup/Setup.cs
```diff
@@ -118,10 +118,11 @@ protected override void InitializeApp(IMvxPluginManager pluginManager, IMvxAppli
                     .WithStopwatchProvider<FirebaseStopwatchProviderAndroid>()
                     .WithIntentDonationService(new NoopIntentDonationServiceAndroid())
                     .WithPrivateSharedStorageService(new NoopPrivateSharedStorageServiceAndroid())
-
+                    .WithBackgroundSyncService<BackgroundSyncServiceAndroid>()
                     .StartRegisteringPlatformServices()
                     .WithDialogService(dialogService)
                     .WithFeedbackService(feedbackService)
+
```
simonrozsival: 👀

Toggl.Giskard/Extensions/ActivityExtensions.cs
heytherewill: 👌

Toggl.Daneel/Services/BackgroundSyncServiceIos.cs
```diff
@@ -1,17 +1,13 @@
...
 namespace Toggl.Daneel.Services
 {
     public sealed class BackgroundSyncServiceIos : BaseBackgroundSyncService
     {
-        private const double minimumBackgroundFetchInterval = 15 * 60;
-
         public override void EnableBackgroundSync()
             => UIApplication.SharedApplication
-                .SetMinimumBackgroundFetchInterval(minimumBackgroundFetchInterval);
+                .SetMinimumBackgroundFetchInterval(MinimumBackgroundFetchIntervalInMinutes * 60);
```
daividssilverio: 60 what?

simonrozsival: [](url)
```suggestion
                .SetMinimumBackgroundFetchInterval(TimeSpan.FromMinutes(MinimumBackgroundFetchIntervalInMinutes).TotalSeconds);
```

dshr: I'm going to do the same for iOS 🙌

Toggl.Giskard/Extensions/ActivityExtensions.cs
```diff
@@ -51,6 +52,23 @@ public static (int widthPixels, int heightPixels, bool isLargeScreen) GetMetrics
...
 
+        public static JobInfo CreateBackgroundSyncJobInfo(this Context context, long periodicity)
+        {
+            var javaClass = Java.Lang.Class.FromType(typeof(BackgroundSyncJobSchedulerService));
+            var component = new ComponentName(context, javaClass);
+
+            var builder = new JobInfo.Builder(BackgroundSyncJobSchedulerService.JobId, component)
+                // This job can run on any network
+                .SetRequiredNetworkType(NetworkType.Any)
+                // It will happen once every n millis, but not more than that
```
daividssilverio: It will happen at most once within n milliseconds, but we have no control exactly when it will happen.

daividssilverio: Also, I'm not sure we have a guarantee it will run at all, the docs only say that it will run at most once during the period.

heytherewill: Yeah, clarifications about that will be added on docs. I should probably 🔥 these comments once I add docs :)

PR #3947, opened 2018-12-07
Add duration format setting

### LINKED ISSUE #3731: Add duration format setting
We should have it, but I don't see it in the app's settings.

Ref #1384

Designs in #1268

juanlaube:
## What's this?
Adds the duration format setting row in the settings activity. For some reason we had the fragment implemented but we didn't add the row in the settings 🤷‍♂️ 

### Relationships

Closes #3731

## Why do we want this?
We want user to be able to select different duration formats.

## How is it done?
New row in the settings activity layout + some bindings.

### Side effects
None.

## Review hints
Test it.

## :squid: Permissions
🦑 ❗️


Toggl.Giskard/Activities/SettingsActivity.cs
```diff
@@ -131,6 +135,10 @@ protected override void OnCreate(Bundle bundle)
                 .BindAction(ViewModel.SelectBeginningOfWeek)
                 .DisposedBy(DisposeBag);
 
+            durationFormatView.Rx()
+                .BindAction(ViewModel.SelectDurationFormat)
```
daividssilverio: Try not using `BindAction`, as it is, you can't select the date format more than once in the settings view. Nasty...

PR #3948, opened 2018-12-07
Use subjects to bind textfields

Odrakir:
## What's this?
Binds textfields to subjects instead of using RxAction

### Relationships
Doesn't close any issue.

## Why do we want this?
This started as a way to stop using `RxAction.Inputs` in the VCs, but in the end @khoi found out that using RxActions with textfields could be problematic. So instead I'm repurposing this PR to change RxAction into Subjects for textfields.

## How is it done?
It uses `BehaviorSubject` to get the textfield values. In the case of `TokenResetViewModel` there are some more changes in other for it to work.

## :squid: Permissions
Go ahead.


Odrakir:
Fair enough. Let’s do that. Although... `BindAction` does work flawlessly when the action itself is an Observable (as opposed to a `async Task`)

Toggl.Giskard/Activities/SelectClientActivity.cs
```diff
@@ -39,8 +39,8 @@ protected override void OnCreate(Bundle bundle)
                 .BindAction(ViewModel.Close)
                 .DisposedBy(DisposeBag);
 
-            filterEditText.Rx().Text()
-                .Subscribe(ViewModel.SetFilterText.Inputs)
+            filterEditText.Rx()
```
kornelijepetak: I see changes to Giskard's bindings and no extension method on Giskard.

Odrakir:
I agree with @khoi , sort of...

I'm just refactoring the call in places where we are already using an action so we don't have to access the `Inputs` property. The code was already using RxAction. I'm not trying to refactor the VM itself, which we probably should. 

I see two cases here:

- Binding an `UITextField` to some property in the VM, in which case we are better of using a `BehaviorRelay` or some other kind of `Subject`
- Binding an `UITextField` so it executes some operation on the VM (like searching for something while typing), in which case it makes more sense to use a `RxAction`

But now, after @khoi mentioned this, I took a look at these specific changes and all the actions are doing is setting values in subjects, so yes. This shouldn't be actions at all. Which doesn't mean a UITextField's text shouldn't ever be bound to an action.

khoi:
I agree that the behavior of the app doesn't change because of this PR since we're already binding the text to an action's input directly.

Refactor is due for binding those events to a Subject, or a Relay. Merging this PR would standardize these type of bindings.

> Binding an UITextField so it executes some operation on the VM (like searching for something while typing), in which case it makes more sense to use a RxAction

Binding TextField's text changed directly to a RxAction without somehow involving `Enabled` will cause 💥 sooner or later. 

In your example, I think it'd safer to do this

```
class VM {
   var TextRelay: BehaviorRelay<string>  
   var Output = TextRelay.Debounce(0.3).SelectMany(...etc)
}
class View {
   TextField.Bind(ViewModel.TextRelay).Disposeby(bag)
   ViewModel.Output.BindToSomeView().Disposeby(bag)
}
```

khoi:
> I don't know why you say that. The code you posted can be done with just one action property.

pls collaborate, cause I think it would crash if you use one Action property

Odrakir:
@khoi Is right, so I repurposed this PR to use Subjects for textfields instead of RxActions. I've updated the description.

Toggl.Daneel/Extensions/Reactive/UITextFieldExtensions.cs
```diff
@@ -3,6 +3,7 @@
 using System.Reactive.Disposables;
 using System.Reactive.Linq;
 using Toggl.Foundation.MvvmCross.Reactive;
+using Toggl.Multivac.Extensions;
```
simonrozsival: 👀

Toggl.Foundation.MvvmCross/ViewModels/SelectClientViewModel.cs
```diff
@@ -27,16 +27,12 @@ public sealed class SelectClientViewModel : MvxViewModel<SelectClientParameters,
         private long workspaceId;
         private long selectedClientId;
         private SelectableClientViewModel noClient;
-        private ISubject<string> clientFilterText = new BehaviorSubject<string>(string.Empty);
 
         public IObservable<IEnumerable<SelectableClientBaseViewModel>> Clients { get; private set; }
-
+        public readonly ISubject<string> FilterText = new BehaviorSubject<string>(string.Empty);
```
simonrozsival: Don't use public fields, instead use a property
```suggestion
        public ISubject<string> FilterText { get; } = new BehaviorSubject<string>(string.Empty);
```

Toggl.Foundation.MvvmCross/ViewModels/SelectCountryViewModel.cs
```diff
@@ -20,15 +20,11 @@ public sealed class SelectCountryViewModel : MvxViewModel<long?, long?>
...
-        private readonly ISubject<string> filterText = new BehaviorSubject<string>(string.Empty);
-
         private readonly IMvxNavigationService navigationService;
 
-        public InputAction<SelectableCountryViewModel> SelectCountry { get; }
-
         public IObservable<IEnumerable<SelectableCountryViewModel>> Countries { get; private set; }
-
-        public InputAction<string> SetFilterText { get; }
+        public readonly ISubject<string> FilterText = new BehaviorSubject<string>(string.Empty);
```
simonrozsival:
```suggestion
        public ISubject<string> FilterText { get; } = new BehaviorSubject<string>(string.Empty);
```

Toggl.Foundation.MvvmCross/ViewModels/TokenResetViewModel.cs
```diff
@@ -30,28 +29,23 @@ public sealed class TokenResetViewModel : MvxViewModel
...
         public IObservable<bool> IsPasswordMasked { get; }
-
         public IObservable<bool> HasError { get; }
         public IObservable<string> Error { get; }
-
-        public IObservable<bool> IsLoading { get; }
-
         public IObservable<bool> NextIsEnabled { get; }
 
+        public readonly BehaviorSubject<string> Password = new BehaviorSubject<string>(string.Empty);
```
simonrozsival:
```suggestion
        public BehaviorSubject<string> Password { get; } = new BehaviorSubject<string>(string.Empty);
```

simonrozsival: Also why is it explicitly a `BehaviorSubject` and not just `ISubject`? Do you use some behavior-subject specific properties somewhere? I guess in this case it's OK and it communicates that the last value is returned on subscription. I guess the `Take(1)` is the reason.

Toggl.Foundation.MvvmCross/ViewModels/TokenResetViewModel.cs
```diff
@@ -161,24 +136,21 @@ private async Task signout()
         }
 
         private IObservable<Unit> done() =>
-            Observable.Create<Unit>(observer =>
-            {
-                if (!passwordSubject.Value.IsValid)
+            Password
+                .Take(1)
```
simonrozsival: I would like `FirstAsync` better, but they do the same I guess.

Toggl.Foundation.MvvmCross/ViewModels/TokenResetViewModel.cs
```diff
@@ -161,24 +136,21 @@ private async Task signout()
...
+                .Select(Multivac.Password.From)
+                .Select(password =>
```
simonrozsival: We have a `ThrowIf(password => !password.IsValid, new InvalidOperationException())` extension method which would be perfect for this piece of code.

Toggl.Foundation.MvvmCross/ViewModels/SelectCountryViewModel.cs
```diff
@@ -57,7 +52,8 @@ public override async Task Initialize()
                 allCountries.Insert(0, selectedElement);
             }
 
-            Countries = filterText
+            Countries = FilterText
+                .Select(text => text ?? string.Empty)
                 .Select(text => text.Trim())
```
simonrozsival: You can save one line of code:
```suggestion
                 .Select(text => text?.Trim() ?? string.Empty)
```

Toggl.Foundation.MvvmCross/ViewModels/TokenResetViewModel.cs
kornelijepetak: They are not the same.
If sequence emits no elements, `Take(1)` will continue, `FirstAsync` will throw.

Odrakir: In this case `FirstAsync` makes more sense. Thanks.

PR #3951, opened 2018-12-07
Put navigation to outdated app vm back

### LINKED ISSUE #3484: App Outdated not shown
## 🐛 Bug Report
When the app is outdated, the app outdated modal is not being shown.

## Context

Build/Branch: `develop`
Phone OS: Any Android
Phone model: Any

## Expected behavior
The outdated modal view should be shown. 

## To reproduce

Hard code to make the app outdated:

In `App.cs` add `|| true` to `NavigateToFirstViewModel`

```
if (true || accessRestrictionStorage.IsApiOutdated() || accessRestrictionStorage.IsClientOutdated())
```

daividssilverio:
## What's this?
This is a very small PR that took way too long to get ready.
It basically makes the navigation to the `OutdatedAppViewModel` stay in place when it must.

### Relationships
Closes #3484
Removes a the forced navigation introduced at #3220

## Why do we want this?
Because the navigation wasn't working properly and users with outdated app could still use the app.

## How is it done?
The magic happens by removing the forced navigation to the MainViewModel in the SplashScreen. This navigation was happening after the app navigated to the `OutdatedAppViewModel`, making it without effect.
The change to the `MainActivity:Init` is so that when the application is restarted from the notification manager, it won't restart the time entry.

### Why not another way?
This seems to be fixing the problem alright and it's a small change. The proper way of fixing this would require Mvx to be 🔥 already, like making the MainActivity the actual launcher of our application, then faking a `SplashScreen` by styling the MainActivity at run time, and also handling all deep link intents in the MainActivity. Too much work right now.

### Side effects
Please help me find it. My brain is already fried after so many trial and errors.

## Review hints
It's hard to assert the impact of this PR by just looking, I suggest you run the application and see if everything is alright, specially playing with the notifications. I tested a lot, but there might still be an edge case.

I've requested @heytherewill because I bothered him a lot trying to fix this. But don't mind that.

## :squid: Permissions
anyone 🦑 it


daividssilverio:
Oh yeah, it will for sure @heytherewill o/

PR #3960, opened 2018-12-10
Add calendar suggestion provider

### LINKED ISSUE #3757: Implement calendar suggestions
Depends on #3755 
Depends on #3807 

This should be a new suggestion provider that looks a certain time period ahead. I think 1 hour should be enough, though we can discuss the time span. The provider should prioritise events that are sooner.

Now that i think about it, it might even be useful to look for events in the past as well. This would cover cases where user logs time afterwards. Please leave your thoughts below.

I don't think we will be able to determine a project from the event, so we should just use event description as time entry description.

Regarding suggestion certainty: the formula for calculating it should have the following properties:
* The closer the event is to the current time, the higher certainty
* Events in the future should have a higher score than the ones from the past, though we shouldn't radically decrease the value for past events, for example, an event 1 minute in the past still might be a good suggestion.

The calculation could look something like:
```
delta = abs(curentTimeTicks - eventStartTimeTicks)
certainty = 1 - deltaToDecimal(delta)
if (eventIsInPast)
  certainty -= some value to make sure we rate events in the past lower
```

arturs105:
## What's this?
This PR adds a suggestion provider that gets suggestions from calendar. The suggestion combining logic in `GetSuggestionsInteractor` (and it's tests) weren't touched, since that's gonna be a different PR.

### Relationships

Closes #3757

## :squid: Permissions
🦑


arturs105:
We agreed that we would rely on getting the permission from calendar onboarding. Even on droid, since it is planned for Q1 of next year.

Regarding the throwing: i had added error handling code, but i removed it to see what would happen without it and then i forgot to add it back before creating the PR. Will re-add that.

PR #3961, opened 2018-12-10
Use leaky bucket for pulling to avoid 429s

simonrozsival:
## What's this?
This PR uses a leaky bucket for every request it makes during pull sync to make just 3 requests/second (push will have a separate PR later, once this is finalized and merged).

### Relationships

Ref #3932 

## Why do we want this?
We want the app to work smoothly. 

## How is it done?
I created a wrapper (actually two wrappers) around the API clients which move some concerns away from the `FetchAllSinceState` and which delay sending HTTP requests until there is a free slot (3 requests/s). I didn't want to modify the code in Ultrawave, so I used this approach with the wrappers.

This means there is quite a few changed files.. please don't hesitate to reach out to me if you are confused by the changes during review and if it is not obvious why I moved some parts of the code around.

### Why not another way?
After I implemented all of this, I am asking myself if it is really necessary. It might be sufficient to have just the 60r/60s leaky bucket and not the 3r/1s.

### Side effects
Pull sync is now slower. The main reason is because the requests are now made in a "random" order instead of trying to preserve. It shouldn't be hard to start the pull requests in "waves".

## Tests
Syncing tests: [will be added later once we decide to finalize this PR]

## :squid: Permissions
Just me


Toggl.Foundation/Sync/Helpers/RateLimitingAwareAwareRequestSender.cs
```diff
@@ -0,0 +1,79 @@
...
+
+            this.timeService = timeService;
+            this.leakyBucket = leakyBucket;
+            this.scheduler = scheduler;
+            this.requestSender = requestSender;
+        }
+
+        public IObservable<T> Fetch<T, TApi>(TApi api)
+            where TApi : IPullingSingleApiClient<T>
+        {
```
dshr: Why not use `=>`?

simonrozsival: It wasn't very readable with the
```
    where TApi : ....
    => ...
```

PR #3966, opened 2018-12-11
Fix issue with RxAction.FromAsync

### LINKED ISSUE #3963: BindAction extension isn't working as you'd expect
## 🐛 Bug Report
In someplaces where the `BindAction` extension is being used, events are being forwarded to subscribers only once. This is causing the UI to become irresponsive after the first interaction with the user.

## Context
It seems to be affecting both android and iOS.

## Expected behavior
The events should be sent to the subscribers whenever they happen.

## To reproduce
Find a place where `BindAction` is being used, specially where the user ins't navigated to other screen after interaction.
Examples from where it was happening: 
Sign up screen -> Select Country; Sign up button
Settings screen -> Select default workspace

## Workaround & Related issues
Replacing the `BindAction` with a simple subscription does make it work.
Examples of the workaround here: #3952 #3953

Odrakir:
## What's this?
This fixes an issue introduced when removing the `async Task` override of `Subscribe`, which caused RxAction components to be deallocated after the first completion, preventing the action to be executed twice.

Closes #3963 

## Why do we want this?
We kind of want the app to work properly.

## How is it done?
As most weird and hard to solve problems this one has to do with threading. My guess is that `await` together with `toObservable` switches to some background thread resulting in some unexpected ordering of operations which causes the action components to be deallocated.

I'm not 100% sure on this. But swithching to the current SynchronizationContext inside `FromAsync` fixes all the problems.

### Why not another way?
I'm open to try other things if anyone can shed some light on it.

## :squid: Permissions
Do it. But please, review throughly and test it first.


Toggl.Multivac/Extensions/RxAction/InputAction.cs
```diff
@@ -30,7 +31,8 @@ IObservable<Unit> workFactory(TInput input)
         public static InputAction<TInput> FromAsync(Func<TInput, Task> asyncAction, IObservable<bool> enabledIf = null)
         {
             IObservable<Unit> workFactory(TInput input)
-                => asyncAction(input).ToObservable();
+                => asyncAction(input).ToObservable()
+                    .ObserveOn(SynchronizationContext.Current);;
```
khoi:
```suggestion
                    .ObserveOn(SynchronizationContext.Current);
```

kornelijepetak: `;;`

Toggl.Multivac/Extensions/RxAction/InputAction.cs
```diff
@@ -30,7 +31,8 @@ IObservable<Unit> workFactory(TInput input)
         public static InputAction<TInput> FromAsync(Func<TInput, Task> asyncAction, IObservable<bool> enabledIf = null)
         {
             IObservable<Unit> workFactory(TInput input)
-                => asyncAction(input).ToObservable();
+                => asyncAction(input).ToObservable()
+                    .ObserveOn(SynchronizationContext.Current);
```
heytherewill: This will throw when not being executed from the Main Thread. I suggest we instead use the SchedulerProvider to force main thread or throw an easier to reason about exception like, like so:

```csharp 
IObservable<Unit> workFactory(TInput input)
{
    if (SynchronizationContext.Current == null)
        throw new InvalidOperationException("InputActions need to be created from the main thread");

    return asyncAction(input)
        .ToObservable()
        .ObserveOn(SynchronizationContext.Current);
}
```

Toggl.Multivac/Extensions/RxAction/UIAction.cs
```diff
@@ -33,7 +34,8 @@ IObservable<Unit> workFactory()
         public static UIAction FromAsync(Func<Task> asyncAction, IObservable<bool> enabledIf = null)
         {
             IObservable<Unit> workFactory()
-                => asyncAction().ToObservable();
+                => asyncAction().ToObservable()
+                    .ObserveOn(SynchronizationContext.Current);
```
heytherewill: Same as above

Toggl.Multivac/Extensions/RxAction/InputAction.cs
Odrakir: Ideally I would observe on the Main thread, but then I have to inject it to every action constructor. Unless you finally realize that should be a global property.

I tried the other approach (SynchronizationContext), but the app crashes when I login because it says the constructor of RatinsViewModel is not on the main thread (SynchronizationContext.Current is null)

Odrakir:
I added the `SynchronizationContext` check to RxAction. I had to change two things:

- Login: Navigation after login in was happening in a background thread, that's not correct in my opinion, but with the new change it results in the actions being created in a background thread and thus crashing the app. I changed the navigation to happen in the main thread.

- Logout: I think this might have been introduced when we removed the `async Task` override of `Subscribe`. When logging out the wheel kept spinning but navigation to login screen never happened. I switched here to the main thread as well. But, to be honest, the mix between async/await, observables and subjects that's happening in `SettingsViewModel.tryLogout` is a huge mess. I think we should refactor that to take a more pure approach.

Other than that, this is a WIP because I have yet to test all the places where this might fail.

Toggl.Multivac/Extensions/RxAction/InputAction.cs
daividssilverio: It wouldn't hurt to have global property IMO... ISchedulerProvider could be initialized once and have static properties there for us `MainScheduler`, `DefaultScheduler`, `BackgroundScheduler`. That's how I used to have back on vanilla android (`AndroidSchedulers.IO()`, `AndroidSchedulers.Main()`, `AndroidSchedulers.Background()`)

simonrozsival: We should be very careful with static classes. I would advice using any global mutable static state (e.g., having a static global `TestScheduler`). Using parallelism is impossible with threads overwriting the global state all the time. Our unit tests run in parallel and they share global static state. Our tests won't work (but sometimes they will). But if you find a way to implement this (partial static class with pure getters with a static constructor in each project - `Daneel`, `Giskard`, tests...? is that even possible?), let's consider it.

Odrakir: The only thing we really need, and that’s all I’m asking for is access to the main thread. 

But as it’s different for each platform, it’s has to be setup at startup. So I’m not sure if that’ll cause the problems Simon mentions.

Odrakir: BTW. I already tried this before and failed: https://github.com/toggl/mobileapp/pull/3676

So if anyone else wants to give it a go...

simonrozsival: The problem I see: we would like to inject a test scheduler in tests, right? If not, then this is not a problem (but will it limit our ability to test VMs?)

Odrakir: I don't know, the main thread is the main thread. And there are things that need to happen there regardless.

The Observable sent into the RxAction can be in any thread we like, but we switch to the main thread for the transformations that happen in the action itself. I see no problem there.

Then there's the Observables the action exposes (Elements, Errors, Executing), normally those are used on the main thread as well. I'm not sure if we ever do something with them that needs some kind of scheduler for testing (like waiting or repeating things many times). But I think maybe in that case it should be the responsibility of the VM to switch to another scheduler that can be injected.

What do you think?

Odrakir: We could also add a bool in the constructor to turn this on/off. That way we could easily test RxAction itself (although I don't think there are any problems with the current tests). And also VMs that need it could turn it off, instead (but I'm not sure if it's ever necessary).

simonrozsival: If we had something like an "RX actions factory" which would replace the static factory `UIAction.From*(...)` similarly to how we have an interactor factory and this factory would inject the main thread to the constructor of the RxAction, we wouldn't have to worry about statics or ugly code. It would be just one more dependency for the VMs which would be used only in the constructor, so it doesn't require all that much boilerplate anyway :)

Odrakir: "Just one more dependency" sound like the words of an addict. I'm honestly not sure about this, your approach sounds right, but I think it might be an overkill, still I'm open to try it. Let's see what @heytherewill has to say.

daividssilverio: I didn't suggest injecting the static schedulers in tests @simonrozsival. I'm talking about UI only, that's all we need, a replacement for `SynchronizationContext.Current`. I wouldn't replace the existing ISchedulerProvider injection.

daividssilverio: What I said would involve moving the `ObserveOn` to the `BindAction` extension as well... but we already did that once with AsDriver and went back... I remember now...

Odrakir:
Superseded by #3980

Issue #3969, opened 2018-12-11
🔑🎨🤖 User Interface for User Access on Android


kornelijepetak:
🔑 Related to `user-access-improvements` project.

This issues describes the screens that need to be implemented on Android platform:

* [x] Entry screen
 * (governed by `SignupOrLoginChoiceViewModel`)
&nbsp;
* [x] Login screen
 * (governed by `LoginViewModel` #3900 )
&nbsp;
* [ ] Signup screen
 * (governed by `SignupScreenViewModel`)
&nbsp;
* [x] Forgot password screen
 * (governed by `ForgotPasswordViewModel`)

PR #3970, opened 2018-12-11
Implement signup/login screen on Giskard

kornelijepetak:
## What's this?
Implements the initial choice for the users: signup or login.
Also, renames the actions used in the VM. 

### Relationships
Part of #3969

## Why do we want this?
To minimize the mental effort.

### Side effects
None forseeable]

## Review hints
It should not be that hard.

## :squid: Permissions
🔑 people.


Toggl.Giskard/Activities/SignupOrLoginChoiceActivity.cs
```diff
@@ -0,0 +1,46 @@
...
+            this.ChangeStatusBarColor(Color.White, true);
+
+            base.OnCreate(bundle);
+            SetContentView(Resource.Layout.SignupOrLoginChoiceActivity);
+
+            OverridePendingTransition(Resource.Animation.abc_slide_in_bottom, Resource.Animation.abc_fade_out);
+
+            InitializeViews();
+
+            loginButton.Rx().BindAction(ViewModel.StartLoginFlow);
```
khoi: Add these subscriptions to a 🗑 bag

khoi:
Otherwise it's good

PR #3971, opened 2018-12-11
Show running time entry in the calendar view

### LINKED ISSUE #3625: Display running time entry in calendar view
We want to display the currently running time entry in the calendar view and allow the user to stop it. Needs design, but I’d guess fading the bottom of the cell is enough to differentiate it from another entry.

juanlaube:
## What's this?
This shows the running time in the calendar view. It's a work in progress so far because the design is not final.

### Relationships

Closes #3625 

## Why do we want this?
Users should be able to see and interact with the running time entry in the calendar view.

## How is it done?
The changes so far are more than I expected. I had to:

1. Make the duration optional in the `CalendarItem`
2. Use the current date time as the end time in a lot of places when the calendar item duration is not set
3. Disable some UI interactions like editing the end time of the running time entry, or setting the start time in the future
4. Hide the end time indicator when editing the running time entry.
5. Fade the cell bottom when it's the running time entry.

### Why not another way?
There's no other way.

### Side effects
None.

## Review hints
Please test it.

## :squid: Permissions
🦑 ❗️


Toggl.Daneel/Views/Calendar/CalendarCollectionViewEditItemHelper.cs
```diff
@@ -263,16 +268,18 @@ private void changeStartTime(CGPoint point)
 
             LastPoint = point;
             var startTime = Layout.DateAtPoint(LastPoint).ToLocalTime().RoundDownToClosestQuarter();
-            var duration = calendarItem.EndTime - startTime;
+            var duration = calendarItem.Duration.HasValue ? endTime(calendarItem) - startTime : null as TimeSpan?;
 
-            if (duration <= TimeSpan.Zero)
+            if (duration != null && duration <= TimeSpan.Zero)
+                return;
+            if (duration == null && startTime > timeService.CurrentDateTime)
```
simonrozsival: Merge the two ifs with an `||`. If you keep good indentation and parentheses, it will still be easily readable.

Toggl.Daneel/Views/Calendar/CalendarCollectionViewEditItemHelper.cs
```diff
@@ -285,6 +292,9 @@ private void changeStartTime(CGPoint point)
 
         private void changeEndTime(CGPoint point)
         {
+            if (calendarItem.Duration == null)
+                return;
+
             if (!isActive || itemIndexPath == null)
```
simonrozsival: Extend the condition here with `||` instead of another `if`. It won't be even long and it will easily fit on one line.

Toggl.Daneel/Views/Calendar/CalendarCollectionViewEditItemHelper.cs
```diff
@@ -357,5 +367,11 @@ private void onCurrentPointChanged(CGPoint? currentPoint)
                 didDragDown = false;
             }
         }
+
+        private DateTimeOffset endTime(CalendarItem calendarItem)
+            => calendarItem.EndTime.HasValue ? calendarItem.EndTime.Value.LocalDateTime : timeService.CurrentDateTime;
+
+        private TimeSpan duration(CalendarItem calendarItem)
+            => calendarItem.Duration.HasValue ? calendarItem.Duration.Value : timeService.CurrentDateTime - calendarItem.StartTime.LocalDateTime;
```
simonrozsival: This line is too complicated to understand it with just once glance, use separate lines for `?` and `:` and it will be much better! 😄

Toggl.Foundation.MvvmCross/Calendar/CalendarLayoutCalculator.cs
```diff
@@ -149,7 +158,13 @@ private CalendarItemGroups convertIntoColumns(CalendarItemGroups bucket, (Calend
...
         {
-            return new CalendarItemLayoutAttributes(calendarItem.StartTime.LocalDateTime, calendarItem.Duration, totalColumns, columnIndex);
+            return new CalendarItemLayoutAttributes(calendarItem.StartTime.LocalDateTime, duration(calendarItem), totalColumns, columnIndex);
         }
+
+        private DateTimeOffset endTime(CalendarItem calendarItem)
+            => calendarItem.EndTime.HasValue ? calendarItem.EndTime.Value.LocalDateTime : timeService.CurrentDateTime;
+
+        private TimeSpan duration(CalendarItem calendarItem)
+            => calendarItem.Duration.HasValue ? calendarItem.Duration.Value : timeService.CurrentDateTime - calendarItem.StartTime.LocalDateTime;
```
simonrozsival: 🤔 ... this could be an extension method which takes a `DateTimeOffset now` as an argument. The same for the `endTime`. You use these functions multiple times.

Toggl.Foundation.MvvmCross/ViewModels/Calendar/CalendarViewModel.cs
```diff
@@ -290,7 +290,11 @@ private async Task updateTimeEntry(CalendarItem calendarItem)
                 TagIds = timeEntry.TagIds
             };
 
-            if (timeEntry.Duration != calendarItem.Duration.TotalSeconds)
+            var duration = calendarItem.Duration.HasValue 
+                ? calendarItem.Duration.Value.TotalSeconds 
+                : (timeService.CurrentDateTime - calendarItem.StartTime.LocalDateTime).TotalSeconds;
```
simonrozsival: Another place where you could use the extension method.

PR #3979, opened 2018-12-12
Use leaky bucket when pushing

### LINKED ISSUE #3932: Use leaky bucket in syncing algorithm to prevent 429s

simonrozsival:
### Relationships

Follow up of #3977 
Closes #3932 

## Why do we want this?
We want smooth and stable sync.

## How is it done?
Similar to #3961 

### Side effects
The side effect is that there are maaany changes in this PR. I had to make changes to the leaky bucket and that led to several other changes... If you think it is a problem, let me know, I'll try to do something about it.

## Review hints
Don't review until #3977 is merged. Go over this PR commit by commit.

## Tests
Syncing tests: https://app.bitrise.io/build/4163cd7e5639f2ea - don't merge until these tests pass

## :squid: Permissions
🦑 when tests are passing


simonrozsival:
Final syncing graph: [rate-limiting-graph.svg.zip](https://github.com/toggl/mobileapp/files/2672900/rate-limiting-graph.svg.zip)

PR #3980, opened 2018-12-12
Add factory to build RxActions

### LINKED ISSUE #3963: BindAction extension isn't working as you'd expect

Odrakir:
## What's this?
This fixes an issue introduced when removing the async Task override of Subscribe, which caused RxAction components to be deallocated after the first completion, preventing the action to be executed twice.

Closes #3963

## Why do we want this?
We kind of want the app to work properly.

## How is it done?
As most weird and hard to solve problems this one has to do with threading. My guess is that await together with toObservable switches to some background thread resulting in some unexpected ordering of operations which causes the action components to be deallocated.

To solve this actions need to run on the main thread. To make it easier to create them without having to inject a scheduler all the time I created a `RxActionFactory` that's been injected to every VM that needs it.

### Why not another way?
Ideally we should find a way to access the main scheduler (and only that one) from every place in the app without having to inject anything all around, but that might not be easy due to parallel tests.

It's a bummer that we have to make our code more complicated because of tests, it should be like that. Any other ideas about this are welcome.

## :squid: Permissions
Be my guest


Toggl.Foundation.MvvmCross/ViewModels/Calendar/SelectUserCalendarsViewModel.cs
```diff
@@ -15,18 +16,23 @@ namespace Toggl.Foundation.MvvmCross.ViewModels.Calendar
...
         public UIAction Done { get; private set; }
 
         public SelectUserCalendarsViewModel(
             IUserPreferences userPreferences,
             IInteractorFactory interactorFactory,
-            IMvxNavigationService navigationService)
-            : base(userPreferences, interactorFactory)
+            IMvxNavigationService navigationService,
+            IRxActionFactory rxActionFactory
+            )
```
daividssilverio: 👀

Toggl.Foundation.MvvmCross/ViewModels/BrowserViewModel.cs
```diff
@@ -1,24 +1,33 @@
...
     {
         private readonly IMvxNavigationService navigationService;
+        private readonly IRxActionFactory rxActionFactory;
 
         public string Url { get; private set; }
 
         public string Title { get; private set; }
 
-        public BrowserViewModel(IMvxNavigationService navigationService)
+        public UIAction Close { get; }
```
daividssilverio: Please fix the android usage for all the Close UIActions you are adding.
BrowserActivity.cs:
```
private void onNavigateBack(object sender, NavigationClickEventArgs e)
        {
            ViewModel.Close().Execute();
        }
```

Toggl.Foundation.MvvmCross/ViewModels/SelectCountryViewModel.cs
```diff
@@ -26,23 +29,27 @@ public sealed class SelectCountryViewModel : MvxViewModel<long?, long?>
         public IObservable<IEnumerable<SelectableCountryViewModel>> Countries { get; private set; }
 
         public InputAction<string> SetFilterText { get; }
+        public UIAction Close { get; }
```
daividssilverio: Same for this one, please fix the usage on`SelectCountryActivity.cs`.

Toggl.Foundation.MvvmCross/ViewModels/Settings/SettingsViewModel.cs
```diff
@@ -92,6 +93,7 @@ public sealed class SettingsViewModel : MvxViewModel
         public UIAction PickDefaultWorkspace { get; }
         public UIAction SelectDurationFormat { get; }
         public UIAction SelectBeginningOfWeek { get; }
+        public UIAction Close { get; }
```
daividssilverio: Same here, fix the `Close` usage on SettingsActivity.cs

daividssilverio:
Another bug, the RxAction is getting disposed when a Dialog is cancelled by the user by clicking outside of bounds of the dialog (at least on Android). Example: try changing workspaces inside the Reports view then clicking outside of the dialog. iOS doesn't seem to be affected by it...
 
I'm fine with having this fixed in another issue if you say it's not related.

Odrakir:
Superseded by #3995

Toggl.Foundation.MvvmCross/ViewModels/BrowserViewModel.cs
```diff
@@ -1,24 +1,33 @@
...
+
+        public BrowserViewModel(IMvxNavigationService navigationService, IRxActionFactory rxActionFactory)
         {
             Ensure.Argument.IsNotNull(navigationService, nameof(navigationService));
+            Ensure.Argument.IsNotNull(rxActionFactory, nameof(rxActionFactory));
 
             this.navigationService = navigationService;
+            this.rxActionFactory = rxActionFactory;
```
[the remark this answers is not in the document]
Odrakir: How about the `navigationService`? Does the closure capture it?

Toggl.Foundation.MvvmCross/FoundationExtensions.cs
```diff
@@ -59,6 +60,7 @@ ITogglDataSource createDataSource(ITogglApi api)
 
                 Mvx.ConstructAndRegisterSingleton<IInteractorFactory, InteractorFactory>();
                 Mvx.ConstructAndRegisterSingleton<IAutocompleteProvider, AutocompleteProvider>();
+                Mvx.ConstructAndRegisterSingleton<IRxActionFactory, RxActionFactory>();
```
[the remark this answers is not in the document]
Odrakir: I didn't think twice about this, just copied the way the interactor factory is created. Why should this be different?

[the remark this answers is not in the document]
Odrakir: Oh, right, because of dependencies. I see now.

Issue #3910, opened 2018-12-04
🔑🎨🍎 User Interface for User Access on iOS


kornelijepetak:
🔑 Related to `user-access-improvements` project.

This issues describes the screens that need to be implemented on iOS platform:

* [x] Entry screen
 * (governed by `SignupOrLoginChoiceViewModel`)
&nbsp;
* [x] Login screen
 * (governed by `LoginViewModel` #3900 )
&nbsp;
* [ ] Signup screen
 * (governed by `SignupScreenViewModel`)
&nbsp;
* [x] Forgot password screen
 * (governed by `ForgotPasswordViewModel`)

PR #3983, opened 2018-12-13
🔑 ForgotPassword screen for iOS

khoi:
### Relationships
Part of #3910

## How is it done?
"Contact Us" button is added to the UI which opens the support webpage if user encounters > 2 errors.

### Side effects
None

## Review hints
None beside functionally testing it.

## :squid: Permissions
🔑


simonrozsival:
The code looks good and it works well. I just don't understand why we should show that page to users who have problems logging in. I couldn't find any information about login there. Shouldn't we just tell them the email address where they should write? I couldn't find the specs in the issue, so that's why I'm asking. Could you give me link to the specs or a discussion about this?

kornelijepetak:
@simonrozsival Current copy is not complete.
There will be a new one with two options/links:

* check KB
* write to us

First one will link to a toggl support page describing the login process (in progress with support team at the moment). The second one will open the feedback form, hoping that at least the email they entered is valid so that the feedback can be sent.

amulware:
Letting them send feedback without logging in is a big deal - we have to make sure they get to see their email when writing their message and make it clear that that is how we'll answer them.

kornelijepetak:
@amulware because of the delay in the project, only one of the two features will go out with the first release, and that one is the specialized page for troubleshooting the common login/signup problems on the support KB (which is to be finalized during the following week).

The second one requires some changes to the user feedback dialog, which you have mentioned and will have to be done separately in an issue.

PR #3987, opened 2018-12-13
Add Machine learning suggestions provider

### LINKED ISSUE #3927: Implement random forest suggestion provider
This is a new a new suggestion provider that uses methods described in #3754.

sofokli1:
## What's this?
Adds a new suggestion provider that uses Random forest to predict what the user is going to work on

### Relationships
Closes #3927

## Why do we want this?
We want more cleaver suggestions

## How is it done?
It adds the Accord Framework and uses the random forest implementation that it provides. For features we use `day of week`, `is business hours`, `is weekend`, `start time`. `Day of week` and `start time` are normalised to 0.0-1.0. It is done in 2 steps. First we predict a project and then we a predict the TE from that project. The TE is converted into a suggestion and boom Tetris for beard squad.

### Why not another way?
There are way too many other ways. This is just the start, the first step into smarting up the suggestions. In the future we could try other classifiers from the accord framework since it is relatively easy to swap them.

### Side effects
Only high ratings in the app store ↗️📈

## Review hints
Definitely take a look at the code, tests are missing and there will be changes but I do need a sanity check. Definitely some functional testing if you track your time regularly and let me know if it is getting close at all or just shooting randomly.

## Tests

Integration tests:  - do not merge this branch unless these tests are passing.

Syncing tests:  - do not merge this branch unless these tests are passing.

## :squid: Permissions
No 🧔 no 🦑


Toggl.Foundation/Suggestions/MachineLearningSuggestionProvider.cs
```diff
@@ -0,0 +1,255 @@
...
+        }
+
+        private long predictProjectID(IEnumerable<IDatabaseTimeEntry> timeEntries)
+        {
+            var tes = timeEntries.Where(te => te.ProjectId.HasValue).ToList();
+            Console.WriteLine("start -=-=-==-=-=- " + tes.Count);
+
+            double[][] inputs = prepareInputsFor(tes);
+            int[] outputs = prepareProjectOutputsFor(tes);
+            int[] uniqueOutputs = uniqueOutputsFor(outputs); // This also updates the outpus to have values same as the their index in the uniqueOutputs
```
simonrozsival: I don't understand how this works and why you do it. Can you explain it in more detail?

Toggl.Foundation/Suggestions/MachineLearningSuggestionProvider.cs
```diff
@@ -0,0 +1,255 @@
...
+        private IEnumerable<IDatabaseTimeEntry> predictUsingRandomForestClassifier(IEnumerable<IDatabaseTimeEntry> timeEntries)
+        {
+            var timeEntriesForPredictingProject = timeEntries.ToList();
+            var predictedProjectId = predictProjectID(timeEntriesForPredictingProject); //Step 1 Predict Projects
+
+            var timeEntriesFromProject = timeEntries.Where(te => te.ProjectId == predictedProjectId).ToList();
+
+            var predictedTimeEntryID = predictTimeEntryID(timeEntriesFromProject); //Step 2 Predict the TimeEntry
+
+            return timeEntries.Where(te => te.Id == predictedTimeEntryID);
```
simonrozsival: use `Any` instead of `Where` for preformance boost (it will stop once it finds a matching element).

Toggl.Foundation/Suggestions/MachineLearningSuggestionProvider.cs
```diff
@@ -0,0 +1,255 @@
...
+
+            var watch = System.Diagnostics.Stopwatch.StartNew();
+
+            var forest = teacher.Learn(trainingInputs, trainingOutputs);
+
+            watch.Stop();
+            var elapsedMs = watch.ElapsedMilliseconds;
+            Console.WriteLine("prediction time (ms): " + elapsedMs);
+
+            int[] predicted = forest.Decide(inputToPredict);
```
simonrozsival: Can't you get some level of certainty of the prediction? I think that we should set some threshold and if there isn't a good suggestion, use different suggestion provider instead of a random time entry selected by the random forest.

Toggl.Foundation/Suggestions/MachineLearningSuggestionProvider.cs
```diff
@@ -0,0 +1,255 @@
...
+
+namespace Toggl.Foundation.Suggestions
+{
+    public sealed class MachineLearningSuggestionProvider : ISuggestionProvider
+    {
+        private readonly ITogglDatabase database;
+        private readonly ITimeService timeService;
+        private readonly int maxNumberOfSuggestions;
+
+        private readonly int maxNumberOfTimeentriesForTraining = 500;
```
simonrozsival: I think there should be also a minimum number and if we don't have enough items for training, we shouldn't even try. The results would be too random. Could it be 50? 100? I'm not sure but we should try it and find a good value.

Toggl.Foundation/Suggestions/MachineLearningSuggestionProvider.cs
```diff
@@ -0,0 +1,255 @@
...
+            var inputToPredict = new double[][] { toFeatures(timeService.CurrentDateTime) };
+            var prediction = predict(inputs, outputs, inputToPredict);
+
+            return uniqueOutputs[prediction];
+        }
+
+        private long predictTimeEntryID(IEnumerable<IDatabaseTimeEntry> timeEntries)
+        {
+            var tes = timeEntries.Where(te => te.ProjectId.HasValue).ToList();
+            Console.WriteLine("start -=-=-==-=-=- " + tes.Count);
```
arturs105: 👀 There's other places, were this is left as well.

Toggl.Foundation/Suggestions/MachineLearningSuggestionProvider.cs
arturs105: Let me see, if i understand, how `uniqueOutputs` works:
1. It returns an array of time entry ids, where each element is unique
2. It modifies the passed in `outputs` so it refers to the appropriate index in the returned value. i.e. the passed in array now contains the indexes of the return value.
3. After that the array of indexes is used to find the appropriate index

P.S. There's a typo in the comment.

Toggl.Foundation/Suggestions/MachineLearningSuggestionProvider.cs
```diff
@@ -0,0 +1,255 @@
...
+                Join = forestJoin
+            };
+
+            var watch = System.Diagnostics.Stopwatch.StartNew();
+
+            var forest = teacher.Learn(trainingInputs, trainingOutputs);
+
+            watch.Stop();
+            var elapsedMs = watch.ElapsedMilliseconds;
+            Console.WriteLine("prediction time (ms): " + elapsedMs);
```
arturs105: I think it would be useful to track this performance in Firebase.

Toggl.Foundation/Suggestions/MachineLearningSuggestionProvider.cs
```diff
@@ -0,0 +1,255 @@
...
+
+            var inputToPredict = new double[][] { toFeatures(timeService.CurrentDateTime) };
+            var prediction = predict(inputs, outputs, inputToPredict);
+
+            return uniqueOutputs[prediction];
+        }
+
+        private long predictTimeEntryID(IEnumerable<IDatabaseTimeEntry> timeEntries)
+        {
+            var tes = timeEntries.Where(te => te.ProjectId.HasValue).ToList();
```
arturs105: Aren't `timeEntries` already filtered to have a project by this point?

Toggl.Foundation/Suggestions/MachineLearningSuggestionProvider.cs
sofokli1: as Arturs said. If i keep the original IDs `long` values the algo will never stop training because IDs are too large. This had makes things way simpler and faster. @simonrozsival does this answer your question?

Toggl.Foundation/Suggestions/MachineLearningSuggestionProvider.cs
sofokli1: `Any` returns a `Bool` I am trying to return a `IEnumerable`. How would it work?

Toggl.Foundation/Suggestions/MachineLearningSuggestionProvider.cs
```diff
@@ -0,0 +1,255 @@
...
+        public IObservable<Suggestion> GetSuggestions()
+        => database.TimeEntries
+                   .GetAll()
+                   .Select(filterTimeEntries)
+                   .Select(predictUsingRandomForestClassifier)
+                   .SelectMany(toSuggestions);
+
+        private IEnumerable<IDatabaseTimeEntry> filterTimeEntries(IEnumerable<IDatabaseTimeEntry> timeEntries)
+            => timeEntries.Where(te => te.ProjectId.HasValue)
+                          .Take(maxNumberOfTimeentriesForTraining);
```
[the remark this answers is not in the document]
sofokli1: Doesn't 'Take' give me the top `maxNumberOfTimeentriesForTraining` of the log? Why shuffle? I want the most recent TEs. The older the TEs the less relevant. Unless I am getting something wrong

Toggl.Foundation/Suggestions/MachineLearningSuggestionProvider.cs
sofokli1: I am going to use how well fits the training set as a score (reverse error). How does that sound?

Toggl.Daneel/Toggl.Daneel.csproj
```diff
@@ -230,6 +230,33 @@
...
+      <HintPath>..\packages\System.Buffers.4.4.0\lib\netstandard2.0\System.Buffers.dll</HintPath>
+    </Reference>
+    <Reference Include="System.CodeDom">
+      <HintPath>..\packages\System.CodeDom.4.4.0\lib\netstandard2.0\System.CodeDom.dll</HintPath>
+    </Reference>
+    <Reference Include="System.Memory">
+      <HintPath>..\packages\System.Memory.4.5.1\lib\netstandard2.0\System.Memory.dll</HintPath>
+    </Reference>
+    <Reference Include="System.Threading.Tasks.Dataflow">
+      <HintPath>..\packages\System.Threading.Tasks.Dataflow.4.8.0\lib\netstandard2.0\System.Threading.Tasks.Dataflow.dll</HintPath>
```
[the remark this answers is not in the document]
sofokli1: I don't know

Toggl.Foundation/Suggestions/MachineLearningSuggestionProvider.cs
```diff
@@ -0,0 +1,255 @@
...
+using Accord.Statistics.Kernels;
+using Accord.Statistics;
+using Accord.MachineLearning.DecisionTrees;
+using Accord.Statistics.Analysis;
+using Accord.MachineLearning.Performance;
+using Accord.MachineLearning;
+
+namespace Toggl.Foundation.Suggestions
+{
+    public sealed class MachineLearningSuggestionProvider : ISuggestionProvider
```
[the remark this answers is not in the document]
sofokli1: I left it like that on purpose because random forest is an internal implementation detail and idea was to easily change it in the future. But I can rename i do not mind

Toggl.Foundation/Suggestions/MachineLearningSuggestionProvider.cs
simonrozsival: You're right. Then maybe the function I meant was `FirstOrDefault`

Toggl.Foundation/Suggestions/MachineLearningSuggestionProvider.cs
simonrozsival: Let's try that!

Toggl.Foundation/Suggestions/MachineLearningSuggestionProvider.cs
```diff
@@ -0,0 +1,255 @@
...
+    {
+        private readonly ITogglDatabase database;
```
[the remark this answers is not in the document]
simonrozsival: Yes, let's not use database directly unless we really need it!

Toggl.Foundation/Suggestions/MachineLearningSuggestionProvider.cs
simonrozsival: I can understand it now easily, but I'm worried that we will forget how exactly this works later and have problems figuring it out when we maybe refactor this later. It would be better to put it in a comment which explains the purpose of the code.

Toggl.Foundation/Suggestions/MachineLearningSuggestionProvider.cs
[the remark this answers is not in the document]
simonrozsival: The time entries you get from the database/data source are sorted by the order in which they were written into the database. If I'm not wrong using `Take` like this will give you the oldest time entries we have in the DB, not the newest ones.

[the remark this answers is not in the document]
sofokli1: I added a for loop to prein the descriptions of the TEs after the take and those are the same as the log. most recent on top.

Toggl.Foundation/Suggestions/RandomForestSuggestionProvider.cs
```diff
@@ -0,0 +1,308 @@
...
+                     .GetAll()
+                     .Select(predictUsingRandomForestClassifier)
+                     .SelectMany(toSuggestions);
+
+        private IEnumerable<IDatabaseTimeEntry> filterTimeEntries(IEnumerable<IDatabaseTimeEntry> timeEntries)
+            => timeEntries.Where(te => te.ProjectId.HasValue)
+                          .Take(maxNumberOfTimeEntriesForTraining);
+
+        private IEnumerable<Suggestion> toSuggestions(IEnumerable<(IDatabaseTimeEntry, float)> timeEntriesAndScore)
+            => timeEntriesAndScore.Select(timeAndScore => new Suggestion(timeAndScore.Item1, timeAndScore.Item2));
```
arturs105: Rename `timeAndScore ` to `entryAndScore` or `timeEntryAndScore`.

Toggl.Foundation/Suggestions/RandomForestSuggestionProvider.cs
```diff
@@ -0,0 +1,308 @@
...
+            this.maxNumberOfSuggestions = maxNumberOfSuggestions;
+        }
+
+        public IObservable<Suggestion> GetSuggestions()
+        => dataSource.TimeEntries
+                     .GetAll()
+                     .Select(predictUsingRandomForestClassifier)
+                     .SelectMany(toSuggestions);
+
+        private IEnumerable<IDatabaseTimeEntry> filterTimeEntries(IEnumerable<IDatabaseTimeEntry> timeEntries)
```
arturs105: Unused method

Toggl.Foundation/Suggestions/RandomForestSuggestionProvider.cs
```diff
@@ -0,0 +1,308 @@
...
+            return timeEntries.Where(te => te.Id == predictedTimeEntryIDAndScore.Item1)
+                              .Select(te => (te, avgScore)); ;
+        }
+
+        private IEnumerable<(IDatabaseTimeEntry, float)> predictUsing1Step(IEnumerable<IDatabaseTimeEntry> timeEntries)
+        {
+            var timeEntriesList = timeEntries.ToList();
+            var predictedTimeEntryIDAndScore = predictTimeEntryID(timeEntriesList);
+
+            return timeEntries.Where(te => te.Id == predictedTimeEntryIDAndScore.Item1)
```
arturs105: Can there even be more than 1 entry with a certain id? 🤔 I'm thinking `SingleOrDefault` would be better here.

Toggl.Foundation/Suggestions/RandomForestSuggestionProvider.cs
```diff
@@ -0,0 +1,308 @@
...
+
+            if (timeEntries.Count() >= minNumberOfAllowedTimeEntries)
+            {
+                return predictUsing1Step(timeEntries);
+            }
+
+            return new List<(IDatabaseTimeEntry, float)>();
+        }
+
+        private IEnumerable<(IDatabaseTimeEntry, float)> predictUsing2Steps(IEnumerable<IDatabaseTimeEntry> timeEntries)
```
arturs105: I'm thinking using named tuples would be more readable, though then all the tuples in this file should be named. Up to you, if you wanna do this, since currently it's not that terrible.

Toggl.Foundation/Suggestions/RandomForestSuggestionProvider.cs
```diff
@@ -0,0 +1,308 @@
...
+            var timeEntriesList = timeEntries.ToList();
+            var predictedProjectIDAndScore = predictProjectID(timeEntriesList); //Step 1 Predict Projects
+
+            var timeEntriesFromProject = timeEntries.Where(te => te.ProjectId == predictedProjectIDAndScore.Item1).ToList();
+
+            var predictedTimeEntryIDAndScore = predictTimeEntryID(timeEntriesFromProject); //Step 2 Predict the TimeEntry
+
+            var avgScore = (predictedProjectIDAndScore.Item2 + predictedTimeEntryIDAndScore.Item2) / 2;
+
+            return timeEntries.Where(te => te.Id == predictedTimeEntryIDAndScore.Item1)
```
arturs105: Same as below

Toggl.Foundation/Suggestions/RandomForestSuggestionProvider.cs
```diff
@@ -0,0 +1,308 @@
...
+            {
+                outputs[i] = Array.IndexOf(uniqueOutputs, outputs[i]);
+            }
+
+            return uniqueOutputs;
+        }
+
+        private (long, float) predictProjectID(IEnumerable<IDatabaseTimeEntry> timeEntriesForPrediction)
+        {
+            var timeEntries = timeEntriesForPrediction.ToList();
```
arturs105: This method is only ever called with a `List`, so you can change the type for the argument and remove the extra `ToList` call.

Toggl.Foundation/Suggestions/RandomForestSuggestionProvider.cs
```diff
@@ -0,0 +1,308 @@
...
+            Console.WriteLine("start -=-=-==-=-=- " + timeEntries.Count);
+
+            double[][] inputs = featureRemovingUnchaningingValues(prepareInputsFor(timeEntries));
+            int[] outputs = prepareProjectOutputsFor(timeEntries);
+            int[] uniqueOutputs = uniqueOutputsFor(outputs); // This also updates the outpus to have values same as the their index in the uniqueOutputs
```
arturs105: The comment still has a typo.

Toggl.Foundation/Suggestions/RandomForestSuggestionProvider.cs
```diff
@@ -0,0 +1,308 @@
...
+        }
+
+        private (long, float) predictTimeEntryID(IEnumerable<IDatabaseTimeEntry> timeEntriesForPrediction)
+        {
+            var timeEntries = timeEntriesForPrediction.ToList();
+            Console.WriteLine("start -=-=-==-=-=- " + timeEntries.Count);
+
+            double[][] inputs = featureRemovingUnchaningingValues(prepareInputsFor(timeEntries));
+            int[] outputs = prepareTimeEntryOutputsFor(timeEntries);
+            int[] uniqueOutputs = uniqueOutputsFor(outputs); // This also updates the outpus to have values same as the index in the uniqueOutputs
```
arturs105: Typo in this comment as well.

arturs105:
Also: to fix Android build add `Accord.MachineLearning` to the droid project.

Toggl.Foundation.Tests/Suggestions/RandomForestSuggestionProviderTests.cs
```diff
@@ -0,0 +1,169 @@
...
+                        .GetAll()
+                        .Returns(Observable.Empty<IEnumerable<IThreadSafeTimeEntry>>());
+
+                var suggestions = await Provider.GetSuggestions().ToList();
+
+                suggestions.Should().HaveCount(0);
+            }
+
+            [Fact, LogIfTooSlow]
+            public void ReturnsOneWhenEnoughDataWithoutProjectSuggestions()
```
arturs105: This name doesn't really make sense grammatically.

arturs105: This is true for other test names as well.

Toggl.Foundation.Tests/Suggestions/RandomForestSuggestionProviderTests.cs
```diff
@@ -0,0 +1,169 @@
...
+
+                var timeEntries = getTimeEntries(120, false);
+
+                DataSource.TimeEntries
+                        .GetAll()
+                        .Returns(Observable.Return(timeEntries));
+
+                var suggestions = provider.GetSuggestions().ToList().Wait();
+
+                suggestions.Should().HaveCount(1);
```
arturs105: This (and other tests) need better AAA organisation.

Toggl.Foundation.Tests/Suggestions/RandomForestSuggestionProviderTests.cs
```diff
@@ -0,0 +1,169 @@
...
+            }
+
+            [Fact, LogIfTooSlow]
+            public void ReturnsOneWhenEnoughDataWithProjectSuggestions()
```
arturs105: This and some other tests can be made async.

Toggl.Foundation.Tests/Suggestions/RandomForestSuggestionProviderTests.cs
```diff
@@ -0,0 +1,169 @@
...
+
+                Action tryingToConstructWithEmptyParameters =
+                    () => new RandomForestSuggestionProvider(dataSource, timeService, NumberOfSuggestions);
+
+                tryingToConstructWithEmptyParameters
+                    .Should().Throw<ArgumentNullException>();
+            }
+        }
+
+        public sealed class TheGetSuggestionsMethod : RandomForestSuggestionProviderTest
```
arturs105: You should also test, that the project is set in the suggestion, when time entries contain projects. Testing that the project and description are set *correctly* would be best, but predicting the output might not be that easy with the random forest approach.

arturs105: There should also be tests for cases when there are some time entries with project and some without. For example: what happens when the minimum number of time entries is met, but only half of them have projects?

Toggl.Foundation.Tests/Suggestions/RandomForestSuggestionProviderTests.cs
```diff
@@ -0,0 +1,169 @@
...
+
+            protected RandomForestSuggestionProviderTest()
+            {
+                Provider = new RandomForestSuggestionProvider(DataSource, TimeService, NumberOfSuggestions);
+
+                TimeService.CurrentDateTime.Returns(_ => Now);
+            }
+        }
+
+        public sealed class TheConstructor : RandomForestSuggestionProviderTest
```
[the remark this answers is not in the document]
sofokli1: it only returns 0 or 1 result regardless of the `maxNumberOfSuggestions`. How would you test that? There are already tests to check if it returns 0 or 1

Toggl.Foundation.Tests/Suggestions/RandomForestSuggestionProviderTests.cs
sofokli1: will combine all in one test

Toggl.Foundation.Tests/Suggestions/RandomForestSuggestionProviderTests.cs
sofokli1: For the first one just checking that the description and project is set should be enough. If it is going to be tuff.
For the second, if we have both data with and without project. The logic will check if those with project are above the minimum and will use only those. if those with project are below the minimum it will go to the 1 step logic strait to using all the time entries (by all i mean same as the max value)

Toggl.Foundation/Suggestions/MachineLearningSuggestionProvider.cs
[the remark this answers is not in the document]
simonrozsival: Interesting. Ok then.

Toggl.Foundation/Suggestions/RandomForestSuggestionProvider.cs
```diff
@@ -79,7 +79,7 @@ private IEnumerable<(IDatabaseTimeEntry, float)> predictUsing2Steps(List<IDataba
 
             var avgScore = (predictedProjectIDAndScore.score + predictedTimeEntryIDAndScore.score) / 2;
 
-            return new[] { (timeEntries.FirstOrDefault(te => te.Id == predictedTimeEntryIDAndScore.id), avgScore) };
+            return new[] { (timeEntries.SingleOrDefault(te => te.Id == predictedTimeEntryIDAndScore.id), avgScore) };
```
simonrozsival: Why did  you change this from `First` to `Single`? I don't see the need (`Single` goes over the whole list and throws if there are > 1 items with the ID, `First` will return once it finds the item and it won't go over the rest of the list).

Toggl.Foundation/Analytics/BaseAnalyticsService.cs
```diff
@@ -200,6 +200,9 @@ protected BaseAnalyticsService()
         [AnalyticsEvent("Type", "Message")]
         public IAnalyticsEvent<string, string> UnknownSignUpFailure { get; protected set; }
 
+        [AnalyticsEvent("Type", "Duration")]
+        public IAnalyticsEvent<string, long> RandomForestExecutionTime { get; protected set; }
```
simonrozsival: Shouldn't we use Firebase for measuring performance? AppCenter can't show these values really nicely.

Toggl.Foundation/Suggestions/RandomForestSuggestionProvider.cs
```diff
@@ -0,0 +1,313 @@
...
+
+            return new[] { (timeEntries.SingleOrDefault(te => te.Id == predictedTimeEntryTuple.id), avgScore) };
+        }
+
+        private IEnumerable<(IDatabaseTimeEntry, float)> predictUsing1Step(List<IDatabaseTimeEntry> timeEntries)
+        {
+            var predictedTimeEntryTuple = predictTimeEntryID(timeEntries);
+            analyticsService.RandomForestExecutionTime.Track("TimeEntry prediction step 1of1", predictedTimeEntryTuple.duration);
+
+            return new[] { (timeEntries.SingleOrDefault(te => te.Id == predictedTimeEntryTuple.id), predictedTimeEntryTuple.score) };
```
simonrozsival: I don't understand why you changed this from `First` to `Single`. What was the reason? `Single` always traverses the whole list, unless it finds a second occurrence and then it throws. I don't think there should be duplicates so it shouldn't throw, but it doesn't make sense to go over the rest of the list and that's what `First` would do. Maybe it's a bit of early optimization, but it doesn't really make sense to use `Single` here and it's not too much work to change it back to the better version. Unless you had a reason which I can't see right now.

Toggl.Foundation/Suggestions/RandomForestSuggestionProvider.cs
```diff
@@ -0,0 +1,313 @@
...
+            var start = dateTimeOffset;
+            var dayOfWeek = (int)start.DayOfWeek;
+            var startHour = TimeZoneInfo.ConvertTime(start, TimeZoneInfo.Local).Hour;
+
+            var dayOfWeekNormalized = (double)dayOfWeek / 6;
+            var isWeekend = dayOfWeek == 0 || dayOfWeek == 6 ? 1 : 0;
+            var isBusinessHours = startHour > 8 && startHour < 18 ? 1 : 0;
+            var startHourNormalized = (double)startHour / 23;
+
+            double[] features = new double[4];
```
simonrozsival: you could simplify this to be
```csharp
var features = new[]
{
    dayOfWeekNormalized,
    isWeekend,
    isBusinessHours,
    startHourNormalized
};
```

But the current code is fine. Change it only if you like it better yourself 😄

Toggl.Foundation.Tests/Suggestions/RandomForestSuggestionProviderTests.cs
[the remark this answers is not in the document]
arturs105: I meant to test that the constructor throws if `maxNumberOfSuggestions` is invalid - lower than 1 and greater than 9.

Toggl.Foundation.Tests/Suggestions/RandomForestSuggestionProviderTests.cs
arturs105: Then we should test that the second thing you described indeed happens.

Toggl.Foundation/Analytics/BaseAnalyticsService.cs
arturs105: Yeah, i think `IStopwatchProvider` should be used (if understand it's purpose properly). You can message :sanic: team to clarify this.

Toggl.Foundation/Suggestions/RandomForestSuggestionProvider.cs
sofokli1: I was not aware of the underling implementation. will change

Toggl.Foundation/Suggestions/RandomForestSuggestionProvider.cs
sofokli1: I went with verbose way thinking that it can make a difference performance wise, but not sure anymore.

simonrozsival: I believe the compiler will emit the very same instructions.

PR #3988, opened 2018-12-14
Don't forget to start the stopwatch so that the whole bgsync doesn't explode

simonrozsival:
## What's this?
This PR fixes a bug in the background sync interactor.

### Relationships
Introduced in #3946 

## Why do we want this?
We want bgsync to work!

## How is it done?
The stopwatch needs to be started, because if it's not, it will throw when it is stopped.

## Review hints
I add this snippet into `StartTimeEntryViewModel`'s constructor:
```csharp
var interactor = new RunBackgroundSyncInteractor(dataSource.SyncManager, analyticsService, stopwatchProvider, dataSource, timeService);
interactor.Execute().Subscribe();
```

Then it's easily debuggable.

## :squid: Permissions
🦑


simonrozsival:
I missed that in the review 😅 let's 🦑  this and pretend it never happened 😄

amulware:
🦑 sees all and forgets nothing.

dshr:
@amulware that sounded proper Lovecraftian

PR #3995, opened 2018-12-14
RxActionFactory + Fixing RxAction related tests

Odrakir:
## What's this?
This is in principle the same as #3980 but instead of trying to fix tests playing with await and schedulers it reviews tests one by one to see what's wrong with them. I've just started, but turns out, so far, all tests I've seen failing should actually be failing. They were wrong from the start, but for some reason they used to pass. That's until we introduced the scheduler in `RxAction`.

A simple example in `AboutViewModelTests`. We have this test:

```c#
public async Task NavigatesToTheLicensesViewModel()
{
    await ViewModel.OpenLicensesView.Execute();
    await NavigationService.Received().Navigate<LicensesViewModel>();
}
```
That's testing this action
```c#
OpenLicensesView = rxActionFactory.FromAsync(openLicensesView);
...
private Task openLicensesView()
    => navigationService
        .Navigate<LicensesViewModel>();
```

As you can see the test waits for the execution of the action, that is, until the action completes. But the action won't complete until navigation finishes. The test is trying to prove the `Navigate` is called, but that happens at the start, there's no need for the action to finish (that's the beginning of the action, the completion is when the `Task` returned by `Navigate` finishes). 

So there's no need to `await` the execution of the action because what we are trying to prove happens immediately on execution.

With the introduction of the scheduler in RxAction this tests get stuck for ever, which I'm not sure why (does `Navigate` never end?)

Anyway, it's solved by changing the test to:

```c#
public void NavigatesToTheLicensesViewModel()
{
    ViewModel.OpenLicensesView.Execute();

    NavigationService.Received().Navigate<LicensesViewModel>();
}
```

Like this there are many, many others. I think we need to review them all.

## :squid: Permissions
Nope


daividssilverio:
All my comments from #3980 remain.

Toggl.Foundation.MvvmCross/FoundationExtensions.cs
```diff
@@ -59,6 +60,7 @@ ITogglDataSource createDataSource(ITogglApi api)
 
                 Mvx.ConstructAndRegisterSingleton<IInteractorFactory, InteractorFactory>();
                 Mvx.ConstructAndRegisterSingleton<IAutocompleteProvider, AutocompleteProvider>();
+                Mvx.ConstructAndRegisterSingleton<IRxActionFactory, RxActionFactory>();
```
juanlaube: This should be injected in `MvvmCrossFoundation` instead, the interactor factory and autocomplete provider and here only because they depend on the data source.

Toggl.Foundation/Services/RxActionFactory.cs
```diff
@@ -0,0 +1,43 @@
+using System;
+using System.Reactive;
+using System.Threading.Tasks;
+using Toggl.Multivac;
+using Toggl.Multivac.Extensions;
+
+namespace Toggl.Foundation.Services
+{
+    public class RxActionFactory : IRxActionFactory
```
juanlaube: seal pls

Toggl.Foundation.MvvmCross/FoundationExtensions.cs
juanlaube: 🤦‍♂️ Ignore pls, you already addressed this

PR #3998, opened 2018-12-17
Add more suggestion tracking

### LINKED ISSUE #3779: Add new event for tracking suggestion usage
This event is for starting a time entry from suggestion. Basically the same as `TimeEntryStarted` with `TimeEntryStartOrigin.Suggestion`.

We need a new event to be able to include the type of suggestion used as well (type of suggestion provider).

**Important:** we must keep the old event (`TimeEntryStarted`) as well. Basically we'll be logging suggestion usage twice.

arturs105:
## What's this?
This adds 2 new events for suggestions:
* An event for tracking shown suggestions
* An event for tracking a started suggestion

### Relationships

Closes #3779

## :squid: Permissions
🦑


Toggl.Foundation/Suggestions/Suggestion.cs
```diff
@@ -40,11 +40,15 @@ public sealed class Suggestion : ITimeEntryPrototype
 
         public float Certainty { get; }
 
-        internal Suggestion(IDatabaseTimeEntry timeEntry, float certainty)
+        public string ProvierType { get; }
```
heytherewill: Typo. Also, `ProviderName` is better than provider type. Even better would be to have an enum with all possible types

arturs105: How do enum values show up in analytics? 🤔

PR #4001, opened 2018-12-17
Measure if app is up to date during initial sync

### LINKED ISSUE #3760: Make it possible to observe if any new data were fetched during background sync

simonrozsival:
## What's this?
This PR will allow us to track how good our background syncing is (for time entries).

### Relationships
Closes #3760 

## Why do we want this?
We want to know if the background syncing has any effect for the users.

## How is it done?
I created a new service which now has some of the functionality which was in the `TogglDataSource` (hence so many touched files) and added tracking of the entities synced during the initial automatic sync.

The data can be cluttered because we can't reliably isolate just the data from the sync and not from user's interaction with the app. If you have ideas, please talk to me.

### Why not another way?
We want to move away from data sources + remove the circular dependency of the sync manager and the data source. That's why I didn't just put everything in the data source.

### Side effects
Many changed files...

## Review hints
Commit by commit.

## :squid: Permissions
🦑


amulware:
> We want to know if the background syncing has any effect for the users.

Do we have a baseline to distinguish users being up to date due to background sync and users who only use the mobile app?

simonrozsival:
@amulware how would we distinguish people who only use the mobile app and people who use other clients as well? We won't have such precise data and if the numbers we will collect turn unusable, we will have to add more parameters or something.

For now I think the biggest problem is still that we always fetch at lest 1 time entry (unless the user has no TEs at all).

PR #4016, opened 2018-12-17
Refactor our platform constants

### LINKED ISSUE #3824: Remove ViewModel dependepencies from UserAgent
This one is even odder because nothing should depend on it to start with!

It's only used once, in the SettingsViewModel. The data should be moved to the `IPlatformConstants` and we should fetch it using an interactor.

### LINKED ISSUE #3954: Refactor the PlatformConstants Service
This depends on #3824, which is pretty simple to solve.

The following needs to be accomplished in order for this issue to be closed:

- [x] The `Platform` enum needs to be moved the `IPlatformConstants` file
- [x] The `Platform` property of `PlatformInfo` needs to be moved into the `IPlatformConstants` file
- [x] The `PlatformInfo` class needs to be deleted
- [x] We need to rename `IPlatformConstants` to `IPlatformInfo`
- [x] We need to provide a base implementation of `IPlatformInfo` that relies on Xamarin.Essentials' [App Info](https://docs.microsoft.com/en-us/xamarin/essentials/app-information?context=xamarin%2Fxamarin-forms&tabs=android) and [Device info](https://docs.microsoft.com/en-us/xamarin/essentials/device-information?context=xamarin%2Fxamarin-forms&tabs=ios) apis.
- [x] Each client specific implementation should also be renamed accordingly, inherit from `BasePlatformInfo` and implement whichever methods we couldn't implement using Xamarin.Essentials
- [x] ViewModel dependencies need to be changed so they rely on the new service

heytherewill:
## What's this?
This refactors our service that provides device information to use Xamarin.Essentials. It also merges the two existing services that serve the same purpose into a single one.

### Relationships
Closes #3824
Closes #3954

## Why do we want this?
This is much easier to reason about

## How is it done?
It relies on Xamarin.Essentials to provide App and Device information

### Why not another way?
We were handling things on our end, but Xamarin.Essentials is a set of XPlat APIs that's maintained by Microsoft itself. It's linker safe (which means it won't make our apps bulky) and it's better than maintaining that code on our end.

## Review hints
Reviewing this commit by commit is probably significantly easier.

## :squid: Permissions
Anyone, anytime


Toggl.Foundation/TogglFoundation.cs
```diff
@@ -266,9 +257,9 @@ public Builder WithApplicationShortcutCreator<TApplicationShortcutCreator>()
                 where TApplicationShortcutCreator : IApplicationShortcutCreator, new()
                 => WithApplicationShortcutCreator(new TApplicationShortcutCreator());
 
-            public Builder WithPlatformConstants<TPlatformConstants>()
-                where TPlatformConstants : IPlatformConstants, new()
-                => WithPlatformConstants(new TPlatformConstants());
+            public Builder WithPlatformInfo<TplatformInfo>()
+                where TplatformInfo : IPlatformInfo, new()
```
simonrozsival: `TplatformInfo` ➡️  `TPlatformInfo`

Toggl.Daneel/PlatformInfoIos.cs
```diff
@@ -1,62 +1,17 @@
...
-        private static string getDeviceModel()
+        public PlatformInfoIos()
+            : base("https://support.toggl.com/toggl-timer-for-ios/", Platform.Daneel)
         {
-            try
-            {
-                //Get length of phone identifier
-                var pointerToLength = Marshal.AllocHGlobal(sizeof(int));
-                int sysctlResult;
-                sysctlResult = sysctlbyname(phoneModelProperty, IntPtr.Zero, pointerToLength, IntPtr.Zero, 0);
```
simonrozsival: 🙌 🎉

Toggl.Giskard/Startup/Setup.cs
```diff
@@ -83,10 +81,10 @@ protected override void InitializeApp(IMvxPluginManager pluginManager, IMvxAppli
             var userAgent = new UserAgent(clientName, version);
             var mailService = new MailServiceAndroid(ApplicationContext);
             var dialogService = new DialogServiceAndroid();
-            var platformConstants = new PlatformConstants();
+            var platformInfo = new XamarinEssentialsPlatformInfo("https://support.toggl.com/toggl-timer-for-android/", Platform.Giskard);
```
simonrozsival: I think it would be nicer to have a `PlatformInfoAndroid` class which would extend `XamarinEssentialsPlatformInfo` (I don't like the name BTW - it's too specific and it leaks the internal implementation) and pass the parameters to the base constructor and not here in the setup. It would also allow us to 🔒 the base class and stick to the "either abstract or sealed" rule.

PR #4020, opened 2018-12-18
Remove ObserveOn where not needed

Odrakir:
### What's this?
After adding a scheduler to `RxAction` we don't need to switch manually to the main thread anymore, so this PR removes all usages of `ObserveOn` in RxActions.

To be fair, when I thought of doing this I thought there were many more cases off this, but there's only one 🤦‍♂️ 

## How is it done?
Just :fire: that line

## :squid: Permissions
Do it.


Odrakir:
Maybe so the project team remember to do it as well?

PR #4024, opened 2018-12-18
Daneel 1.15 - Daribow

juanlaube:
## Changelog
```
- Added the running time entry to the calendar view

- Improved in the calendar view layout
- Improved time to open project selection view

- Fixed the calendar notification actions behavior
- Fixed a crash when user doesn't have a default workspace
```

## Interesting fact
A former Imperial Prefect in the Anacreon province. It's mentioned in Foundation by Anselm Haut Rodric as a possible threat to Terminus' independence from the recently formed Kingdoms.

## History of changes
https://github.com/toggl/mobileapp/compare/daneel-1.14.2..daribow

## 🦑 permissions
🚫 Only the release manager.

### DIFF
--- a/Toggl.Daneel/Info.plist
+++ b/Toggl.Daneel/Info.plist
@@ -34,7 +34,7 @@
 	<key>CFBundleName</key>
 	<string>Toggl for Devs</string>
 	<key>CFBundleShortVersionString</key>
-	<string>1.14.2</string>
+	<string>1.15</string>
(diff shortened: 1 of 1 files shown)


ApocalypticFish:
Testing is completed. Some small issues, one that occurred in 1.14.2 as well was a default entry showing up during testing on a new account.
Also noted an issue that causes freezes but was unable to replicate after multiple attempts.
Given that, I think it's safe to release.

PR #4031, opened 2018-12-19
Remove the CreateNewSyncManager method

### LINKED ISSUE #3956: Fix the need for recreation of the sync manager
In order to fix the bugs from #2769, #2816 introduce a method in the DataSource that allows one to recreate an ISyncManager. Since the DataSource shouldn't even depend on the Sync Manager to start with, we need to remove this indirection

@simonrozsival says, pun intended, that the sync manager can be fixed so that it doesn't crash and burn in this particular scenario, so I'm assigning this issue to him so he can take a look and either fix (when this issue is properly scheduled in a milestone) or at least describe in detail what needs to be done.

simonrozsival:
## What's this?
This pr _just_ removes the method. It turns out, it's not necessary to use it in the `NoWorkspaceViewModel` and that's the only place when it was used.

### Relationships

Partially reverts #2816
Closes #3956

## Why do we want this?
We want to get rid of dependency between the sync manager and the data source.

## How is it done?
🔥 🔥 🔥 

### Side effects
Removing some code without replacing it in some way seems scary to me. The issue #2769 (closed by #2816) mentioned problems because of the sync manager being unusable due to experiencing an error.

That would be a problem if we talked about an error which causes the sync manager to freeze. This operation is irreversible at the moment and it would be reasonable to just create a new sync manager.

`NoWorkspaceViewModel`, which used the `CreateNewSyncManager` method, is caused by having no workspace. But in this case the sync manager is not frozen and so there shouldn't be any problems with reusing the same SM and data source.

## Review hints
Try this in the simulator/emulator:
1. sync your app
2. delete all of your workspaces using the API (DELETE `https://toggl.space/api/v8/workspaces/{wid}/leave`)
3. sync your app
4. go to the reports and back to the main view (there is a bug in the no workspace mechanism, I will create a separate issue for that case), you should now see a popup which will ask you to create a new workspace or try again
5. A - press the create new workspace button and wait for the app to sync
B - create a workspace using the API (POST `https://toggl.space/api/v9/workspaces` body: `{"name":"ABC","initial_pricing_plan":0}`)
6. Create a new time entry (without specifying the project to use the default workspace)
7. Check that all the data is in the staging web app

## :squid: Permissions
🦑


simonrozsival:
@dshr found a bug. I already have a solution but I need to finish fixing it tomorrow (I still have to write some tests).

dshr:
@juanlaube @heytherewill you didn't get any problems like getting stuck on the "no workspaces" view by any chance?

simonrozsival:
We have to put this on hold. We have to wait until #3835 is merged because it fixes rate limiting problems the app has right now. The "Create Workspace" button puts the app in an infinite 429 retrying because... because.

juanlaube:
@dshr I noticed that once, but I tested a second time without any problem and ignored it. Do you think that's going to be a common problem?

dshr:
@juanlaube I keep getting them 429s every time

juanlaube:
The leaky bucket might solve this I guess.

simonrozsival:
It will be solved by the leaky bucket. Let's re-review this in January  when the bgsync branch is merged into develop.

heytherewill:
Removing the `on-hold` label and requesting a review from @dshr

PR #4034, opened 2018-12-19
Fix Android background service

### LINKED ISSUE #3999: App not syncing after 5 minutes of inactivity
## 🐛 Bug Report
App doesn't sync automatically when resuming after 5 minutes of inactivity.

## Context

Build/Branch: Giskard 1.7.1
Phone OS: Android
Phone model: any

## Expected behavior
App should sync automatically like Daneel.

## To reproduce
1. Be logged in, with app synced and go to background
2. Make some changes in the webapp, e.g: start a time entry
3. Wait 5 minutes and resume the app
4. 🐛 Sync should start and the changes should appear

## Comments
I double checked this only happens in Android, iOS works as expected.

juanlaube:
## What's this?
This fixes (or adds?) the detection of the app going to background and becoming foreground again. 

### Relationships

Closes #3999 

## Why do we want this?
Different parts of the app depend on this, for example, sync should happen when returning after 5 minutes in background.

## How is it done?
The new `ApplicationLifecycleObserver` is registered as application's activity lifecycle callbacks and component callbacks.
When the app goes to background, [memory is trimmed with `UiHidden` level](https://developer.android.com/reference/android/content/ComponentCallbacks2.html#TRIM_MEMORY_UI_HIDDEN), the `isInBackground` flag is set and we call `EnterBackground` in the background service.
When the app becomes foreground again, an activity is resumed, and if `isInBackground` is set, we call `EnterForeground` in background service.

### Why not another way?
Another approach is using Android Architecture Components, which includes a new `ProcessLifecycleOwner` class, but this means installing also a lot of new dependencies.

### Side effects
None.

## Review hints
Please test in real devices, put the app in background and return after 5 minutes, see if sync triggers.

## :squid: Permissions
Only after @heytherewill, @simonrozsival and @daividssilverio are ok with this.


Toggl.Giskard/Startup/ApplicationLifecycleObserver.cs
```diff
@@ -0,0 +1,74 @@
...
+        {
+        }
+
+        public void OnActivityStopped(Activity activity)
+        {
+        }
+
+        public void OnConfigurationChanged(Configuration newConfig)
+        {
+            throw new NotImplementedException();
```
daividssilverio: Throwing doesn't sound like a good idea here... could you just leave it empty?

juanlaube: oh I forgot to remove this 🙈

dshr: Do we really need this `throw`?

juanlaube: No, see previous comment ☝️ 🤦‍♂️

dshr: Nvm, missed the changes

Toggl.Giskard/Startup/Setup.cs
```diff
@@ -136,6 +138,11 @@ protected override void InitializeApp(IMvxPluginManager pluginManager, IMvxAppli
 
             ensureDataSourceInitializationIfLoggedIn();
 
+            var mvxApplication = MvxAndroidApplication.Instance;
```
simonrozsival: This block of code would deserve a separate method.

daividssilverio:
This works, as long as the application process isn't killed by the system to reclaim memory.
What it means is that the application can still be restored but `isInBackground` will be reset and the sync won't trigger.

We could add another flag there to trigger `backgroundService.EnterBackground();` when the first activity is resumed after the application is created.

You can simulate the process being killed by going into developer options > Apps > Background process limit.

juanlaube:
@daividssilverio `isInBackground` starts in `true` so whenever the app is created `EnterForeground` will be called. This should work regardless of the app being just launched or returning from a tombstoned state.

daividssilverio:
Then something else is breaking it... I'm always getting `Sync Failed` when returning to a killed process.

juanlaube:
API token maybe? The fact that it tries to sync, even if it fails, means this work tho 👀

daividssilverio:
I see why it won't trigger the sync, the `isInBackground` starts with true, but the AppResumedFromBackground will emit a value less than `minimumTimeInBackgroundForFullSync`, if you check `TogglDataSource.cs`

juanlaube:
Re the sync failures, I think it's better to follow up this in a new issue. I created #4047 for that.

juanlaube:
@simonrozsival sync fails sometimes only when the app is tombstoned by the OS (you can limit the active processes to test this), in a normal situation this is not a problem.

juanlaube:
The follow up issue has been closed. For some reason there's a problem with the self signed certificate but that only happens in debug builds, so all is good for now.

PR #4046, opened 2018-12-19
🚧 Check and remove duplicate suggestions

### LINKED ISSUE #3766: Make sure we don't have duplicate suggestions
Depends on #3764 

If 2 providers return the same suggestion, we should only one of those. Checking for equal project and description should be enough.

sofokli1:
## What's this?
Checks and removes duplicate suggestions based on description

### Relationships
Closes #3766

## Why do we want this?
No need for duplicates

## How is it done?
The code is self-explanatory and short. Hint `groupBy`

### Why not another way?
Can't think of an other way

### Side effects
none

## Review hints
It is missing tests but those can be added on a separate PR if time is pressing

## :squid: Permissions
:squid:

### DIFF
--- a/Toggl.Foundation/Interactors/Suggestions/GetSuggestionsInteractor.cs
+++ b/Toggl.Foundation/Interactors/Suggestions/GetSuggestionsInteractor.cs
@@ -43,8 +43,19 @@ namespace Toggl.Foundation.Interactors.Suggestions
             => getSuggestionProviders()
                 .Select(provider => provider.GetSuggestions())
                 .Aggregate(Observable.Concat)
+                .ToList()
+                .Select(removingDuplicates)
+                .SelectMany(s => s)
                 .Take(suggestionCount)
                 .ToList();
+
+        private IList<Suggestion> removingDuplicates(IList<Suggestion> suggestions)
+        {
+            return suggestions
+                .GroupBy(suggestion => suggestion.Description)
+                .Select(group => group.First())
+                .ToList();
+        }
 
         private IReadOnlyList<ISuggestionProvider> getSuggestionProviders()
         {


Toggl.Foundation/Interactors/Suggestions/GetSuggestionsInteractor.cs
```diff
@@ -43,9 +43,20 @@ public IObservable<IEnumerable<Suggestion>> Execute()
...
+                .ToList()
+                .Select(removingDuplicates)
+                .SelectMany(s => s)
                 .Take(suggestionCount)
                 .ToList();
 
+        private IList<Suggestion> removingDuplicates(IList<Suggestion> suggestions)
+        {
+            return suggestions
+                .GroupBy(suggestion => suggestion.Description)
```
dshr: What if two suggestions have the same description, but are for different projects/workspaces?

simonrozsival:
I have an idea - if multiple sources suggest the same time entry, maybe we should "merge" these suggestions instead of just taking one. By merging I mean increasing the certainty of the suggestion. By merging I mean at least taking the maximum of the suggestions' scores but there might be a good way to calculate a value between 0 and 1 which is greater than both of those numbers (e.g., `sqrt3(a*b)`).

I don't know if it makes any sense to do that (e.g., do you sort the suggestions by their score so that the best ones are at the top?) but maybe it's worth considering.

sofokli1:
@arturs105 What do you think about this idea 👆

Toggl.Foundation/Interactors/Suggestions/GetSuggestionsInteractor.cs
arturs105: Yeah, the issue says that we should compare description and project. For that the following (an anonymous type) would work: `GroupBy(s => new { s.Description, s.ProjectId })`.

Now that i think about it: would it be useful to include task as well? 🤔

arturs105:
I think increasing the score for matching suggestions makes sense, however i'm not sure, what is mathematically best way to do this 🤔.  Taking the highest score is probably the easiest approach, so we should definitely at least do that.

Toggl.Foundation/Interactors/Suggestions/GetSuggestionsInteractor.cs
simonrozsival: I think you need a workspace ID as well - you can have TEs with the same description (e.g., "") without any project (`null`) but in different workspaces. These would be merged into one.

arturs105: ☝️ Valid point

Also: it might be simpler to use `Distinct<TKey>(Func<Suggestion, Tkey> comparer)` 🤔

Toggl.Foundation/Interactors/Suggestions/GetSuggestionsInteractor.cs
```diff
@@ -43,9 +43,20 @@ public IObservable<IEnumerable<Suggestion>> Execute()
             => getSuggestionProviders()
                 .Select(provider => provider.GetSuggestions())
                 .Aggregate(Observable.Concat)
+                .ToList()
+                .Select(removingDuplicates)
+                .SelectMany(s => s)
```
heytherewill: Why not `SelectMany(removingDuplicates)`?

heytherewill:
Closing as agreed with @amulware  Will be reopened if needed be

PR #4050, opened 2018-12-20
Fix workspace creation

### LINKED ISSUE #3855: Create workspace button allows for multiple interractions
## 🐛 Bug Report

Clicking the "Create workspace" button on the app in an account without a workspace doesn't lock the button or prevent further interaction so with slow sync/connection it's possible to create multiple duplicates workspaces
## Context

Build/Branch:
Phone OS:
Phone model:

## Expected behavior

The button should grey out and prevent further interaction once it has been pressed and only allow interaction again if the workspace creation has an error.
## To reproduce

Have an account with no workspace (this can be done by giving admin access to your workspace to another account and having that account remove the first from the workspace) 
It took a fair while for the workspace to be created and during that time I was able to click the button again and create a second workspace.

Odrakir:
## What's this?
First it fixes a bug in the `GetDefaultWorkspaceInteractor` to prevent the app to crash when there's no workspace.

Then, when creating a new workspace, it moves sync to do it as a side effect so it won't wait for it to dismiss the view. It just creates it locally.

Closes #3855 

## How is it done?
The fix to `GetDefaultWorkspaceInteractor` is done checking the number of items instead of catching the exception. This is because the exception is `InvalidOperationException` and we are catching that one via `TrackException` to log specific errors and we don't want to hide that exception before that.

## :squid: Permissions
Do it!


Toggl.Foundation.MvvmCross/ViewModels/NoWorkspaceViewModel.cs
```diff
@@ -55,44 +51,38 @@ public sealed class NoWorkspaceViewModel : MvxViewModelResult<Unit>
...
             this.rxActionFactory = rxActionFactory;
 
-            CreateWorkspaceWithDefaultName = rxActionFactory.FromAsync(createWorkspaceWithDefaultName);
+            CreateWorkspaceWithDefaultName = rxActionFactory.FromObservable(createWorkspaceWithDefaultName);
             TryAgain = rxActionFactory.FromAsync(tryAgain);
-            IsLoading = isLoading.AsDriver(onErrorJustReturn: false, schedulerProvider: schedulerProvider);
+            IsLoading = Observable.CombineLatest(
+                CreateWorkspaceWithDefaultName.Executing,
+                TryAgain.Executing,
+                CommonFunctions.And);
```
heytherewill: Shouldn't this be `Or`?

Toggl.Foundation/Interactors/Workspace/GetDefaultWorkspaceInteractor.cs
```diff
@@ -31,6 +31,11 @@ public IObservable<IThreadSafeWorkspace> Execute()
 
         private IObservable<IThreadSafeWorkspace> chooseWorkspace()
             => dataSource.Workspaces.GetAll(workspace => !workspace.IsDeleted)
-                .Select(workspaces => workspaces.OrderBy(workspace => workspace.Id).First());
+                .Select(workspaces => workspaces.OrderBy(workspace => workspace.Id))
+                .SelectMany(workspaces =>
+                    workspaces.Count() == 0
```
heytherewill:
```suggestion
                    workspaces.Any()
```

Toggl.Foundation.MvvmCross/ViewModels/NoWorkspaceViewModel.cs
Odrakir: It should, yeah.

simonrozsival:
This will have a conflict with #4031 but I think it will be easy to merge the code "easily".

Odrakir:
> This will have a conflict with #4031 but I think it will be easy to merge the code "easily".

What you really mean is #4031 will have a conflict with this 😄

PR #4055, opened 2018-12-20
Add suggestion balancing

### LINKED ISSUE #3854: Make sure suggestion certainties are balanced
We have to make sure that all the suggestion providers are "on the same page" regarding certainties -we can't have one provider generating suggestions with 20% - 30% certainties and another one with 60% - 80% range.

arturs105:
## What's this?
This makes sure that suggestion certainties are balanced.

### Relationships

Closes #3854

## How is it done?
With the following algorithm:
1. Calculate the average score across all providers. Let's call this `a`
2. Calculate the average score of individual provider. Let's call this `b` (for one of the providers)
3. Calculate the necessary exponent `x` for a provider, where `b^x = a`
4. Raise all the scores from said provider to the power of `x`
5. Repeat for all the other providers

## :squid: Permissions
🦑


Toggl.Foundation/Interactors/Suggestions/GetSuggestionsInteractor.cs
```diff
@@ -60,5 +62,25 @@ private IReadOnlyList<ISuggestionProvider> getSuggestionProviders()
...
+
+        private IReadOnlyList<Suggestion> balancedSuggestions(IList<Suggestion> suggestions)
+        {
+            var overallAverageCertainty = suggestions.Average(suggestion => suggestion.Certainty);
+
+            var averageCertaintiesPerProvider = suggestions
+                .GroupBy(suggestion => suggestion.ProviderType)
+                .ToDictionary(
+                    keySelector: grouping => grouping.First().ProviderType,
+                    elementSelector: grouping => grouping.Average(g => g.Certainty));
```
sofokli1: use `s` instead of `g`

PR #4059, opened 2018-12-20
Droid notifications

### LINKED ISSUE #3965: Show device notification when sync starts and stops on Android
In order to make monitoring background syncs easier we can use local notifications.

dshr:
## What's this?
This adds notifications to the droid notification service.

### Relationships
Closes #3965

## Why do we want this?
To display notifications on Droid.

## How is it done?
By implementing the `NotificationServiceAndroid`.

### Why not another way?
'Tis the only way.

### Side effects
None.

## :squid: Permissions
🦑 !


Toggl.Giskard/Services/NotificationServiceAndroid.cs
```diff
@@ -1,18 +1,49 @@
...
+                    var context = Application.Context;
+                    var notificationIntent = context.PackageManager.GetLaunchIntentForPackage(context.PackageName);
+                    notificationIntent.SetPackage(null);
+                    var pendingIntent = PendingIntent.GetActivity(context, 0, notificationIntent, 0);
+                    var notificationManager = NotificationManager.FromContext(context);
+                    var notificationsBuilder = context.CreateNotificationBuilderWithDefaultChannel(notificationManager);
+
+                    notifications.Select(notification =>
+                        (
+                            notification.ScheduledTime.GetHashCode(), // this is a temp measure, might be better to change this in the future
```
juanlaube: notification has an `Id` property, might be better to use the hash for that unique id instead of the scheduled time I suppose.

Toggl.Giskard/Services/NotificationServiceAndroid.cs
```diff
@@ -1,18 +1,49 @@
...
 using Toggl.Foundation.Services;
-using Toggl.Multivac;
+using Toggl.Giskard.Extensions;
+using Toggl.Multivac.Extensions;
 
 namespace Toggl.Giskard.Services
 {
     public sealed class NotificationServiceAndroid : INotificationService
     {
+        private int notificationIdentifier = 113;
```
juanlaube: 👀

dshr: Right, leftovers from the previous implementation 😅

PR #4062, opened 2018-12-21
Forgot Password UI update

kornelijepetak:
## What's this?
Implements the new workflow for forgot password screen.

### Relationships
Implements part of #3969 

## Why do we want this?
To make it simpler for users.

## How is it done?
Not much has changed in comparison to the old one.

### Side effects
Should be none.

## Review hints
If you are testing the app functionally, know that multiple requests with the same email in short timespan will result in an error that we can't differentiate from unknown/email error, so expect a confusing message there. Until we solve the JSON ERROR API issues accross clients/BE, there's nothing we can do here.

## :squid: Permissions
🔑  people.


Toggl.Giskard/Extensions/Reactive/TextViewExtensions.cs
```diff
@@ -19,8 +19,20 @@ public static IObservable<ICharSequence> TextFormatted(this IReactive<TextView>
...
+        {
+            return text =>
+            {
+                if (!ignoreUnchanged)
+                {
+                    reactive.Base.Text = text;
+                    return;
+                }
+
+                if (reactive.Base.Text != text)
```
khoi: this could be simplified to 
```
if (reactive.Base.Text == text && ignoreUnchanged) {
   return
}
reactive.Base.Text = text;
```

Toggl.Foundation.MvvmCross/ViewModels/ForgotPasswordViewModel.cs
```diff
@@ -45,17 +47,22 @@ public sealed class ForgotPasswordViewModel : MvxViewModel<EmailParameter, Email
...
+            SuggestContactSupport = ResetPassword.Errors
                 .Skip(errorCountBeforeShowingContactSupportSuggestion)
                 .SelectValue(true)
                 .StartWith(false)
                 .AsDriver(false, schedulerProvider);
+
+            ClearErrors = Email
+                .Where(email => email.IsValid)
+                .SelectUnit()
+                .DistinctUntilChanged();
```
khoi: this line will block the stream from emitting after the first element

khoi:
I've pushed a commit to bind the new `ClearErrors` to the iOS UI

PR #4070, opened 2018-12-21
🚧 Add debug helper for suggstion

sofokli1:
## What's this?
Visual aide when for suggestions debugging

## Why do we want this?
To see from which provider the suggestions come and with what score

## How is it done?
when in debug the description of the suggestions will contain the provider and score at the beginning of the string.

### Side effects
It may interfere with tests but until than lets keep it

## Review hints
run it

## :squid: Permissions
:squid:


Toggl.Foundation/Suggestions/Suggestion.cs
```diff
@@ -81,9 +81,9 @@ internal Suggestion(CalendarItem calendarItem, long workspaceId, float certainty
             Description = calendarItem.Description;
         }
 
-        internal Suggestion(Suggestion suggestion, float certainty)
+        internal Suggestion(Suggestion suggestion, float? newCertainty = null, string newDescription = null)
```
arturs105: I think this should be a separate constructor, since we are going to need to remove this code before merging into `develop` and removing a constructor will be simpler than modifying this.

It should also be clear that the other constructor is just for debugging purposes.

sofokli1: Why not keep it in develop like this when it runs in debug?

arturs105: If we wanna have this in develop, i'd prefer if the debug information wasn't in the description of the suggestion. This causes time entries started from suggestions to have the debug info included as well, which will probably get annoying.

If we wanna have debug info in `develop`, it should be on a separate `UILabel`.

PR #4081, opened 2018-12-22
🚧💄 Refactor SuggestionsInteractor

### LINKED ISSUE #4079: Make GetSuggestionsInteractor testable
The problem is that the suggestion providers are created internally by the interactor. We should make them injectable in the constructor. IoC FTW!

sofokli1:
## What's this?
Provides the suggestionProviders in an injectable form to the `GetSuggestionInteractor`

### Relationships
Closes #4079

## Why do we want this?
This will make writing test for the `GetSuggestionInteractor` way easier since now we can mock the suggestions that it gets from the providers

## How is it done?
I added a new interactor to get the suggestionProviders list

### Why not another way?
can't think of any

### Side effects
none

## Review hints
Take a look at the code

## :squid: Permissions
:squid:


Toggl.Foundation/Interactors/Suggestions/GetSuggestionsInteractor.cs
```diff
@@ -16,60 +12,34 @@ namespace Toggl.Foundation.Interactors.Suggestions
...
+            this.suggestionProvidersInteractor = suggestionProvidersInteractor;
         }
 
         public IObservable<IEnumerable<Suggestion>> Execute()
-            => getSuggestionProviders()
-                .Select(getSuggestionsWithExceptionTracking)
-                .Aggregate(Observable.Concat)
+            => suggestionProvidersInteractor
+                .Execute()
+                .SelectMany(p => p)
```
Odrakir: This `SelectMany` is a bit weird to me because you are actually flattening the inner `IEnumerable`... 🤔 But it seems to work just fine, so my only request is using `CommonFunctions.Identity`

Toggl.Foundation/Interactors/Suggestions/GetSuggestionProvidersInteractor.cs
```diff
@@ -0,0 +1,53 @@
...
+            this.suggestionCount = suggestionCount;
+            this.stopwatchProvider = stopwatchProvider;
+            this.dataSource = dataSource;
+            this.timeService = timeService;
+            this.calendarService = calendarService;
+            this.defaultWorkspaceInteractor = defaultWorkspaceInteractor;
+        }
+
+        public IObservable<IEnumerable<ISuggestionProvider>> Execute()
+            => Observable.Return(new List<ISuggestionProvider>
```
simonrozsival: Why should this return observable? This can return `IEnumerable` directly. I can't imagine a use case where this could be beneficial. I think you should change the interactor to implement just `IInteractor<IEnumerable<ISuggestionProvider>>`. It should simplify the usage in `GetSuggestionsInteractor` on which Ricardo already commented.

PR #4092, opened 2019-01-02
🔥 IDiffable

khoi:
## What's this?
This replaces `IDiffable` interface with the existing and simpler `IIdentifiable`

Some history: The `IDiffable` was added so the `BaseRecyclerAdapter` can perform a diff using [DiffUtil](https://developer.android.com/reference/android/support/v7/util/DiffUtil) on the Selectable*VM instances by

- comparing the `Identifier` property, if they're the same, they're considered identical 
- if the `Identifier` is different, it compares the 2 instances by calling `Equals` 

Since our models almost always have some kind of unique `Id`, relying on the comparision of that property should be enough, so there is no need to enforce the classes to override `Equals` anymore.

### Relationships
No created issue.

## Why do we want this?
To simplify the implementation of `Selectable*VM` classes which ease the future simplification and rxfactor of the `Select*VM` classes.

## How is it done?
by deleting code.

### Side effects
None.

## Review hints
None.

## :squid: Permissions
🦑


Toggl.Foundation.MvvmCross/ViewModels/Selectable/SelectableClientViewModel.cs
```diff
@@ -1,59 +1,37 @@
...
-        public string Name { get; set; }
-        public bool Selected { get; set; }
+        public long Id { get; }
+        public string Name { get; }
+        public bool Selected { get; }
 
-        public SelectableClientBaseViewModel(string name, bool selected)
+        public SelectableClientBaseViewModel(long id, string name, bool selected)
         {
+            Ensure.Argument.IsNotNull(Id, nameof(Id));
```
simonrozsival: 👀 you don't need that on a non-nullable value.

Toggl.Foundation.MvvmCross/ViewModels/Selectable/SelectableColorViewModel.cs
```diff
@@ -20,22 +19,6 @@ public SelectableColorViewModel(MvxColor color, bool selected)
...
-        {
-            if (ReferenceEquals(null, obj)) return false;
-            if (ReferenceEquals(this, obj)) return true;
-            return obj is SelectableColorViewModel other && Equals(other);
-        }
-
-        public override int GetHashCode() => HashCode.From(Color, Selected);
-
-        public long Identifier => Color.GetHashCode();
+        public long Id => Color.GetHashCode();
```
simonrozsival: I get how this should work in most cases, but colors can be converted to a hex number and hex number can be a `long` which could be used with confidence that no two different colors would have the same id. But since this has been used up to now, we don't have to do that in this PR.

Toggl.Foundation.MvvmCross/ViewModels/Selectable/SelectableWorkspaceViewModel.cs
```diff
@@ -24,27 +23,6 @@ public SelectableWorkspaceViewModel(IThreadSafeWorkspace workspace, bool selecte
...
-
-        public override int GetHashCode()
-            => HashCode.From(
-                WorkspaceId,
-                WorkspaceName ?? string.Empty,
-                Selected
-            );
-
-        public long Identifier => WorkspaceId;
+        public long Id => WorkspaceId;
```
simonrozsival: We normally put properties before the constructor (even though Rider would like us to do it this way).

Toggl.Giskard/Adapters/BaseRecyclerAdapter.cs
```diff
@@ -139,7 +139,7 @@ public override bool AreContentsTheSame(int oldItemPosition, int newItemPosition
```
simonrozsival: The `AreContentsTheSame` method is the problem here, right? We used `IEquatable<T>` to check if the contents of the items is the same. I think we should not rely on the default `Equals`, because it compares the references, and will probably cause problems.

I think we should compare the IDs even here. In all of our models if the ID is the same, the contents should be the same as well (that's not true only when we replace outdated item with an updated version of that item).

khoi: `AreContentsTheSame` only get called when `AreItemsTheSame` returns `false`. so relying on the references comparison is alright here.

simonrozsival: Then it should probably always return `false`?

heytherewill: Having the same id does not imply that the content is equal, as Simon explained above. I fail to understand why this is not implemented semantically, by comparing if all relevant properties are the same.

Toggl.Giskard/Adapters/BaseRecyclerAdapter.cs
```diff
@@ -131,16 +131,10 @@ public BaseDiffCallBack(IList<T> oldItems, IList<T> newItems)
             }
 
             public override bool AreContentsTheSame(int oldItemPosition, int newItemPosition)
-            {
-                var oldItem = oldItems[oldItemPosition];
-                var newItem = newItems[newItemPosition];
-                return oldItem.Equals(newItem);
-            }
+                => oldItems[oldItemPosition].Id == newItems[newItemPosition].Id;
```
daividssilverio: :wat: This isn't good. As William said, having the same ids won't imply having the same contents.
By the way, just to be clear, this comparison could only care about data that translates to visible elements in the UI.

simonrozsival:
After the reviews from @heytherewill and @daividssilverio I think we should keep the explicit member-wise equivalence and do something simpler:
- delete the `IDiffable<T>` interface
- all the classes that used to implement `IDiffable<T>` would implement directly both `IIdentifiable` and `IEquatable<T>`
- most of the old code could stay, with the exception of renaming `.Identifier` to `.Id`

This will be a very small set of changes. I'm not sure if it's even worth doing and if it's not easier to just keep the `IDiffable<T>` interface... I still don't think it's good to have a unnecessary interface, but we will gain very little by removing it. I hoped this could be a very simple thing to remove and we won't waste too much time by doing it. The rest of the changes we talked about today with @khoi are much more important (so that we can have generic select view for simple selectable items).

khoi:
since the implementation is going to be different, I'm closing it and will submit another PR to improvise the interface

Issue #4052, opened 2018-12-20
Add total tracked time to Calendar View


amulware:
I'd like to use the Calendar more, but I find myself missing the total number of tracked hours a lot.

If there are arguments not to add that, please explain below.

simonrozsival:
I think it's a good idea to add the total time. But how should it look like? Is that up to the dev who implements this?

amulware:
My intuition says to add a thin header showing the total time, and maybe 'today' or the date or something, but that should be confirmed with a designer.

In either case, as discussed in the PR above, we should probably include the running time entry in this total as well, since it is displayed in the calendar, and since this view is similar to the 'today' timeline header on the webapp timer page.

amulware:
Actually, we could maybe just use the already existing bar at the top, and show this instead of the Toggl logo? (similar to how we show the date selection in reports)

lizaPlotnikova:
Seems reasonable. (sorry after all the pinging yesterday I still didn't look at it :P) Can you send a screenshot of what you have?

simonrozsival:
We don't display the total tracked time anywhere atm and I only wrote some code which calculates the time we should display.

simonrozsival:
Does this look OK?

![image](https://user-images.githubusercontent.com/374616/51317604-4fd03480-1a58-11e9-8585-5342fd2a4659.png)

And for running TE:
![image](https://user-images.githubusercontent.com/374616/51317652-71c9b700-1a58-11e9-90f0-1f20346b0ec3.png)

PR #4093, opened 2019-01-02
Calendar: Total tracked time today

simonrozsival:
## What's this?
This PR adds an interactor which calculates how much time the user tracked on the given day. It should be used in the UI somehow, but that's not designed yet and it will be part of a different PR.

### Relationships

Ref #4052 

## How is it done?
An interactor which observes the time tracked on the current day.

The observable tries to recalculate whenever time entries list is changed (create, update, delete), midnight is reached, or when a significant time change is detected.

## :squid: Permissions
🦑


heytherewill:
I really like this. I, however, wonder whether the duration of the running time entry should be included or not. I'd rather see it included, for this confuses me quite a lot when looking at the web app

amulware:
That may be a question for Product to address on a Toggl-wide scale - though, if we show the entry in the calendar, this is potentially a different case. Still, gotta be careful not to confuse users.

simonrozsival:
I understand how you mean it. On the other hand I think this should be implemented in the UI layer in this way:
- get the total tracked time of the stopped time entries for today
- get the currently running TE (if any)
- update the total time
    1. every second (total today + duration of the running TE) when there is a running time entry
        - stop when the TE is stopped
    2. when the today's total changes

The observable created by the interactor could have an internal state and update it based on the changes to the list of time entries (listen to the create/update/delete events), but that would be quite complex.

I even hesitated if the interactor should return an observable or not. The VM could execute the interactor whenever it detected a change in the app's state. In the end I think this is a good compromise between the flexibility it gives to the user of the interactor and complexity of the interactor itself.

amulware:
Actually, I changed my mind from above: The webapp shows a total of stopped entries above the days in the timer page list, but it shows an updating total, including the running entry, in the timeline view above - our calendar is the closest we get to that timeline right now so we should do the same. :)

simonrozsival:
The observable now ticks when there is a running time entry.

Toggl.Foundation/Interactors/TimeEntry/ObserveTimeTrackedTodayInteractor.cs
```diff
@@ -0,0 +1,63 @@
...
+                .StartWith(Unit.Default)
+                .SelectMany(_ =>
+                    calculateTimeAlreadyTrackedToday().CombineLatest(
+                        observeElapsedTimeOfCurrentlyRunningTimeEntry(),
+                        (alreadyTrackedToday, currentlyRunningTimeEntryDuration) =>
+                            alreadyTrackedToday + currentlyRunningTimeEntryDuration))
+                .DistinctUntilChanged();
+
+        private IObservable<Unit> updateIsNecessary()
+            => timeEntries.ItemsChanged().Debug("items_changed")
```
kornelijepetak: No `Debug`s in production.

simonrozsival: 🙈

Toggl.Foundation/Interactors/TimeEntry/ObserveTimeTrackedTodayInteractor.cs
```diff
@@ -0,0 +1,75 @@
...
+            this.timeEntries = timeEntries;
+        }
+
+        public IObservable<TimeSpan> Execute()
+            => whenUpdateIsNecessary()
+                .SelectMany(_ => calculateTimeTrackedToday())
+                .DistinctUntilChanged();
+
+        private IObservable<Unit> whenUpdateIsNecessary()
+            => timeEntries.ItemsChanged()
```
Odrakir: This is nitpicking, and please, feel free to ignore my comment, but to me this is clearer because it communicates better the fact that all those 3 observables are equally important:

```c#
private IObservable<Unit> whenUpdateIsNecessary()
    => Observable.Merge(
            timeEntries.ItemsChanged(),
            timeService.MidnightObservable.SelectUnit(),
            timeService.SignificantTimeChangeObservable
        )
        .StartWith(Unit.Default);
```

Toggl.Foundation/Interactors/TimeEntry/ObserveTimeTrackedTodayInteractor.cs
```diff
@@ -0,0 +1,75 @@
...
+                .Merge(timeService.SignificantTimeChangeObservable)
+                .StartWith(Unit.Default);
+
+        private IObservable<TimeSpan> calculateTimeTrackedToday()
+            => calculateTimeAlreadyTrackedToday().CombineLatest(
+                observeElapsedTimeOfRunningTimeEntryIfAny(),
+                (alreadyTrackedToday, currentlyRunningTimeEntryDuration) =>
+                    alreadyTrackedToday + currentlyRunningTimeEntryDuration);
+
+        private IObservable<TimeSpan> calculateTimeAlreadyTrackedToday()
```
Odrakir: At first I thought `SingleAsync` would fail because you'd get more than one entry that's been stopped. Then I realised you are asking not for one entry but for one `IEnumerable` of entries. 

This again might be me personally, so feel free to ignore it, but I find the mixing of `IEnumerable` and `IObservable` some of these overrides do a bit confusing.

To me it's clearer this way:

```c#
private IObservable<TimeSpan> calculateTimeAlreadyTrackedToday()
    => timeEntries.GetAll(startedTodayAndStopped)
        .Select(entries => entries.Sum(timeEntry => timeEntry.Duration ?? 0.0))
        .Select(TimeSpan.FromSeconds);
```

Toggl.Foundation/Interactors/TimeEntry/ObserveTimeTrackedTodayInteractor.cs
```diff
@@ -0,0 +1,75 @@
...
+                .SelectMany(_ => calculateTimeTrackedToday())
+                .DistinctUntilChanged();
+
+        private IObservable<Unit> whenUpdateIsNecessary()
+            => timeEntries.ItemsChanged()
+                .Merge(timeService.MidnightObservable.SelectUnit())
+                .Merge(timeService.SignificantTimeChangeObservable)
+                .StartWith(Unit.Default);
+
+        private IObservable<TimeSpan> calculateTimeTrackedToday()
```
Odrakir: Again, to better communicate equal importance of Observables:

```c#
private IObservable<TimeSpan> calculateTimeTrackedToday()
    => Observable.CombineLatest(calculateTimeAlreadyTrackedToday(), observeElapsedTimeOfRunningTimeEntryIfAny(),
        (alreadyTrackedToday, currentlyRunningTimeEntryDuration) =>
            alreadyTrackedToday + currentlyRunningTimeEntryDuration);
```

simonrozsival:
@Odrakir I changed the code according to your suggestions 👍

Issue #3957, opened 2018-12-10
Remove the SyncManager dependency from the TogglDataSource


heytherewill:
Depends on #3955 and #3956

This is quite a big one. We'll need to:

- [x] Remove the `ITogglApi` consumers other than sync manager from the DataSource
- [x] Remove the SyncManager property from the DataSource
- [ ] ~Expose Syncing as Interactors so that ViewModels can replace the calls to the SyncManager~ - this turned out to be a bad idea
- [x] Register the `ITogglDataSource` during startup, since this will now be possible for it no longer depends indirectly on the `ITogglApi`
- [x] Change all instances of `ITogglDataSource` that appear on the `IUserAccessManager` to point to `ISyncManager` instead, for it'll be the service that needs to be recreated now

PR #4094, opened 2019-01-02
Remove sync manager error handling from the TogglDataSource class

simonrozsival:
## What's this?
This PR moves sync error handling from the `TogglDataSource` to a new service.

### Relationships
Ref #3957 

## Why do we want this?
We want to get rid of the dependency between the data source and the sync manager. Error handling is the biggest obstacle.

### Why not another way?
I moved the error handling code to a new service. First I thought I would refactor the `ErrorHandlingService` we already have, but that would mean changing some stuff in the `UserAccessManager` (which uses it to handle errors when the user isn't logged in yet).

I think it makes sense to have those two services separate (for now).

### Side effects
None.

## Review hints
I suggest resetting the API token in the web app and putting a breakpoint into the `onError` method of the `SyncErrorHandlingService`.

## :squid: Permissions
🦑


daividssilverio:
I can't find where `UserAccessManager` was using the `ErrorHandlingService`. I can only find references of it in the Login and SignUp view models.

simonrozsival:
You're right, it's used directly from the VMs, not from the UAM.

PR #4095, opened 2019-01-02
Rxfactor: Reports & ReportsCalendar

### LINKED ISSUE #3831: RxFactor the ReportsCalendarViewModel
[Standard RxFactor rules apply](https://www.notion.so/List-of-Refactors-070744dc374f416eacb0381d72143734#0f425d73fe5343a38a454551e8d26a04)

When doing this one, we need to convert the `ReportsCalendarFragment` from a fragment into a View. This is needed to avoid nested fragments and to make the general flow of the app simpler.

### LINKED ISSUE #3832: RxFactor the ReportsViewModel
Depends on #3831

[Standard RxFactor rules apply](https://www.notion.so/List-of-Refactors-070744dc374f416eacb0381d72143734#0f425d73fe5343a38a454551e8d26a04)

When doing this one, we need to convert the `ReportsActivity` into a fragment. This is to prepare the app for the Bottom bar.

daividssilverio:
## What's this?
This is the rxfactory of the Reports and Reports Calendar.

### Relationships
Supersedes #3671 #3658
Closes #3832 and Closes #3831

## Why do we want this?
We want to Mvexit!

## How is it done?
I've copied most of the changes from #3671 and #3831 because both branches were too old and hard to merge, considering that the BarCharts were also added to android in the meantime.

## WIP Status
~Most of the iOS part is done.~
~Android bindings are in progress, but still buggy. It needs some fixing.~
~The Android part needs some fixing, bindings and refactoring the calendar view.~
~The day headers are still not refactored (they are static)~
~The Pie chart Segments still need to be rxfactored both on iOS and Android.~
- [x] All Unit tests still have to be updated/fixed.
- [x] ViewModels are done
- [x] iOS bindings are done
- [x] Android bindings are done

### Why not another way?
I'm up for suggestions, specially for the iOS part and the calendar days handling.

## Review hints
Should be ok to review by commit, but the changes aren't so complicated (except that there's a lot of layout changes on Android).

## :squid: Permissions
🦑 it after WIP is done and one review from at least one person from @toggl/androids and @toggl/ios-senseis .


juanlaube:
Works fine on iOS and Android, I will jump into the code now... wish me luck 😄

PR #4096, opened 2019-01-03
🗜compress image resources in Daneel

khoi:
## What's this?
Losslessly compress all the images to make the binary smaller

**Saved 5 MB out of 8.2 MB. 61.4% overall**

### Relationships
No created issue.

Similar to #3719 on Giskard

## How is it done?
I put all the pngs file to ImageOptim using the default lossless compression settings.

### Side effects
Should be none.

## :squid: Permissions
Go for it.


amulware:
How much smaller does this make the IPA/final app download/install size?

PR #4100, opened 2019-01-03
🍰 Fix CAKE_COMMIT_HASH replacing

### LINKED ISSUE #3666: Change {CAKE_COMMIT_HASH} to something else
Please change it to something else, so that the default value works for integration tests run from Rider or VS without a need for editing this thing.

simonrozsival:
## What's this?
If you forgot to manually replace the `{CAKE_COMMIT_HASH}` in `Toggl.Ultrawave.Tests.Integration/Helper/Configuration.cs` and you ran the integration tests from Rider or VS (without running the 🍰 build script), it would fail, because `UserAgent` version must be an alphanumeric string... so now the thing we replace is just `"CAKE_COMMIT_HASH"` and it won't fail anymore.

### Relationships
Closes #3666 

## Why do we want this?
We want to be able to run individual integration tests from the IDE without having to remember you have to change something somewhere somehow...

## How is it done?
the `"CAKE_COMMIT_HASH"` replacement is safe, we don't use this string anywhere else and I don't think there is a risk that somebody would use it unintentionally.

### Why not another way?
I don't really know why we have a dynamic user agent version for integration tests at all... maybe this was meant to be useful for debugging purposes? I don't think we ever used it.. If @amulware and @heytherewill agree with me, we could set it to some fixed value and remove the thing from 🍰 build script completely and save a few milliseconds on every build 😄 

### Side effects
No side effects.

## :squid: Permissions
🦑


amulware:
The idea indeed is to give backend something to bug us with for debugging or us to ask them about any particular test runs, should it ever be relevant. We indeed have never made use of this.

amulware:
I see no real reason not to keep it though.

PR #4105, opened 2019-01-03
Unify sync states naming and define naming conventions

### LINKED ISSUE #3695: Unify the naming of sync states and results
See @amulware's comment for #3489 

The naming across the syncing states and the results is not consistent. Some examples:

- We often use `Continue` as the name of the result for states which have only one possible transition, but for example `ScheduleCleanUpState.CleanUpScheduled` does not follow this "convention"
- `NoWorkspaceDetectingState` vs. `DetectNoWorkspaceState`

We should:
- [x] Identify all the inconsistencies in the naming of sync states and their results
- [x] Create some conventions for the naming and put these guidelines into the docs
- [x] Rename states and results to follow the guidelines

simonrozsival:
## What's this?
I went through all of the sync states and I unified their names and the names of (some of) their results.

I also updated the docs.

### Relationships
Closes #3695 

## Why do we want this?
We want consistency.

## How is it done?
All the syncing states start with a verb.

Where possible, the "main"/expected result of the state is called `Continue`.

### Why not another way?
I am happy with the (new) names.

### Side effects
No side effects.

## Review hints
Check the commits one by one, otherwise it will seem overwhelming and chaotic.

## Tests
Syncing tests: https://app.bitrise.io/build/fd249f3455b33381 - do not merge this branch unless these tests are passing.

## :squid: Permissions
🦑


docs/syncing/sync-states.md
```diff
@@ -1,5 +1,26 @@
...
+There are two naming conventions:
+
+1. The name of the state should start with a verb. Some examples from our codebase are:
+    - `DeleteOldTimeEntriesState`
+    - `TrackInaccessibleDataAfterCleanUpState`
+    - `DetectLosingAccessToWorkspacesState`
+    - `PersistNewWorkspacesState`
+    - `CreateEntityState`
+    - `MarkEntityAsUnsyncableState`
+2. If the state has a result which has no parameter or if the parameter is the same as the input of the state it should be called `Continue` unless there is a good reason not to.
```
amulware: Maybe add a note that otherwise the result should be named according to the action that took place, not according to what the state expects to happen after - states should not care about that.

amulware: Speaking of, 'Continue' violates that rule and gives very little information about what the state actually did it seems to me. Isn't 'Finished' or 'Done' more semantic?

Toggl.Foundation/Sync/States/Push/PushState.cs
```diff
@@ -18,7 +18,7 @@ internal sealed class PushState<TDatabaseModel, TThreadsafeModel> : IPushState<T
 
         public StateResult<TThreadsafeModel> PushEntity { get; } = new StateResult<TThreadsafeModel>();
 
-        public StateResult NothingToPush { get; } = new StateResult();
+        public StateResult Continue { get; } = new StateResult();
```
amulware: I think renamings like this lose a lot of information. The state can end with 'push entity' (which sort of prescribes what should happen next, is that what we want?) and 'continue' - continue what?

'FoundEntityToPush' and 'FoundNothingToPush' would make more sense to me, for example.

The state name is perhaps not ideal either since it doesn't actually push anything.

Toggl.Foundation/Sync/States/Push/LookForEntityToLookForChangesToPush.cs
```diff
@@ -10,20 +10,20 @@
 
 namespace Toggl.Foundation.Sync.States.Push
 {
-    internal sealed class PushState<TDatabaseModel, TThreadsafeModel> : IPushState<TThreadsafeModel>
+    internal sealed class LookForEntityToLookForChangesToPush<TDatabaseModel, TThreadsafeModel> : ILookForChangesToPush<TThreadsafeModel>
```
amulware: This must be a typo? Did you mean `LookForChangesToPush`?

simonrozsival: That's exactly what I meant 🙈

Toggl.Foundation/Sync/States/Push/LookForChangesToPushSingle.cs
```diff
@@ -8,19 +8,19 @@
 
 namespace Toggl.Foundation.Sync.States.Push
 {
-    internal sealed class PushSingleState<T> : IPushState<T>
+    internal sealed class LookForChangesToPushSingle<T> : ILookForChangesToPush<T>
```
amulware: Should this have a `State` suffix?

Toggl.Foundation/Sync/States/Push/Interfaces/ILookForChangesToPush.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
+
+namespace Toggl.Foundation.Sync.States.Push.Interfaces
+{
+    public interface ILookForChangesToPush<T> : ISyncState
```
amulware: Should this have a `State` suffix?

simonrozsival:
This is the final graph after the renaming: [sync-graph.svg.zip](https://github.com/toggl/mobileapp/files/2736277/sync-graph.svg.zip)

PR #4117, opened 2019-01-08
✨ Add calendar item snap to other TE start/end

### LINKED ISSUE #4032: Calendar: Allow snapping to other TEs
When:

- moving a TE
- adjusting a TE starting
- adjusting a TE end time

I think it should not only snap to regular intervals, but also to the start/end of other TEs (start only snaps to end, end only to start).

This allows use-cases like:

"I forgot to start my TE right after the last one, and want to move the start time to the previous TEs end time"

Other thoughts very welcome, especially if there are ways in which this could harm UX.

sofokli1:
## What's this?
Allows Calendar Entries to snap to the start and end of other Entries when dragging them.

### Relationships
Closes #4032

## Why do we want this?
To help people mach the times better (not everything happens in quarter time periods)

## How is it done?
It gets a list of all start times and a list of all end times and uses the data in there to see what is closer to the the quarter time or one of the start/end times and snaps to the closest.

### Why not another way?
This made sense to me

### Side effects
Only 5 stars reviews

## Review hints
Play with the whole entry dragging after long pressing as well as with start and end dragging. Try develop and try here. Also take a look at [this comment](https://github.com/toggl/mobileapp/issues/4032#issuecomment-452271232)

## :squid: Permissions
:squid:


juanlaube:
This is working nice now! Take care of the other suggestions and this good to go 👍

Toggl.Daneel/Views/Calendar/CalendarCollectionViewAutoScrollHelper.cs
```diff
@@ -74,6 +78,58 @@ protected void StopAutoScroll()
             isAutoScrollingDown = false;
         }
 
+        protected DateTimeOffset NewStartTimeWithDynamicDuration(CGPoint point, List<DateTimeOffset> allItemsStartAndEndTime)
+        {
+            if (!allItemsStartAndEndTime.Any())
```
simonrozsival: This code is almost identical to the one which was previously in `CalendarCollectionViewEditItemHelper.cs`. Would it be possible to reuse some of the code here to reduce the amount of code in this PR?

PR #4118, opened 2019-01-08
Rxfactor select duration format

### LINKED ISSUE #4011: RxFactor the SelectDurationFormatViewModel
Check #3352 for details

juanlaube:
## What's this?
This is an rxfactor for the select duration format view model and the necessary changes in Daneel and Giskard.

### Relationships

Closes #4011

## Why do we want this?
🔥 We want to mvexit.

## How is it done?
Rxfactoring the `SelectDurationFormatViewModel`, the view model was really simple already.

### Why not another way?
There's no other way.

### Side effects
I took the opportunity to add unit tests for the `SelectDurationFormatViewModel` we had nothing 🙊 

## :squid: Permissions
🦑 ❗️


Toggl.Foundation.MvvmCross/ViewModels/SelectDurationFormatViewModel.cs
```diff
@@ -18,22 +18,25 @@ public class SelectDurationFormatViewModel : MvxViewModel<DurationFormat, Durati
 
         public SelectableDurationFormatViewModel[] DurationFormats { get; }
```
khoi: we could make it `IReadonlyCollection<SelectableDurationFormatViewModel>` to avoid outside changes

PR #4123, opened 2019-01-09
Convert Interface Builder strings to Resources.resx

### LINKED ISSUE #3872: Create .strings files for Daneel
As it currently stands, all UI strings that do not come from Foundation are hardcoded in Daneel's .xib/.storyboard files. We got to change that.

The first step is to go in every .xib/.storyboard file and extract all of its strings into a `.strings` file which we can later be translated. [This article](https://medium.com/@mario.negro.martin/easy-xib-and-storyboard-localization-b2794c69c9db) can be of help if you're not familiar with this process.

khoi:
## What's this?
Some groundwork to make our iOS app i18n ready. All ( I hope I don't miss any ) the strings that are set in xib, storyboard files are not put inside `Resources.resx` and label init goes to `ViewDidLoad`

**Note**

There are files that contain attributed button that I haven't translated yet. I'll do it when we agree on a solution.

1. CalendarPermissionDeniedViewController.xib
2. NoWorkspaceViewController
3. OutdatedAppViewController

### Relationships
Closes #3872

## Why do we want this?
i18n l10n

### Why not another way?
We could have a string file for each IB file, but maintaing those are difficult. After discussion with @willsb, we think that having a single `Resources.resx` is the way to go.

### Side effects
Some of the unnecessary line breaks in the middle of the sentence are now removed.

## Review hints
Maybe commit by commit

## :squid: Permissions
Not yet.


Toggl.Foundation/Resources.resx
```diff
@@ -627,4 +627,12 @@ account is not set up</value>
     <data name="Next">
         <value>Next</value>
     </data>
+    <data name="AppOutdatedMessage">
+        <value>Seems you haven't updated the app
+            for quite some time...
+            Please update your app to continue.</value>
```
arturs105: This looks weird on iPhone SE (probably on other phones as well), because the newlines AND the leading spaces are included in the string. The leading spaces inside the string should definitely be removed and i think the first newline should also be removed, so the `UILabel` would break the string into lines for us. So the result would be something like:
```
<root>

    //Other strings

    <data name="AppOutdatedMessage">
         <value>Seems you haven't updated the app for quite some time...
Please update your app to continue.</value>
    </data>

    //Other strings

</root>
```

Toggl.Foundation/Resources.resx
```diff
@@ -677,4 +677,11 @@ account is not set up</value>
     <data name="DurationFormat">
         <value>Duration format</value>
     </data>
+    <data name="Projects">
+        <value>Projects</value>
+    </data>
+    <data name="EmptyProjectText">
+        <value>Your projects will show up here.
+            Use them to categorize your time.</value>
```
arturs105: More leading spaces

Toggl.Foundation/Resources.resx
```diff
@@ -684,4 +684,12 @@ account is not set up</value>
         <value>Your projects will show up here.
             Use them to categorize your time.</value>
     </data>
+    <data name="Tags">
+        <value>Tags</value>
+    </data>
+    <data name="EmptyTagText">
+        <value>Your tags will show up here.
+            Use them to add keywords
+            to your time entries.</value>
```
arturs105: Same. Also, i don't think that the second sentence should include a newline, we should rely on the `UILabel` size to break lines.

Toggl.Foundation/Resources.resx
```diff
@@ -755,4 +755,13 @@ account is not set up</value>
     <data name="IAgree">
         <value>I agree</value>
     </data>
+    <data name="TokenResetInstruction">
+        <value>Enter your password to log back in to your</value>
```
arturs105: I think there's a word missing here.

Toggl.Foundation/Resources.resx
```diff
@@ -235,7 +241,7 @@ time entries will appear here.</value>
         <value>Incorrect password. Please, try again</value>
     </data>
     <data name="GoogleLogin">
-        <value>Log in with Google</value>
+        <value>Sign in with Google</value>
```
arturs105: Not sure about this change. All Zeplin designs use `Log in`, however those are the newly redesign ones, so 🤷‍♂️

khoi: yeah, the old text is still `Sign in with Google`, and this `GoogleLogin` key is not being used anywhere.

PR #4135, opened 2019-01-09
Fix calendar suggestion certainties

arturs105:
## What's this?
Noticed a small problem in the algorithm. let's say there are 2 events (`event1` and `event2`) coming up and we calculate the delta for both of them: `delta1 = 0.5` and `delta2 = 1.4`. According to the previous implementation we would increase `delta1` to `1.5` and that would lead to `event2` having a higher certainty even though `event1` is closer to current time.

## :squid: Permissions
🦑


Toggl.Foundation/Suggestions/CalendarSuggestionProvider.cs
```diff
@@ -62,9 +62,9 @@ private float calculateCertainty(CalendarItem calendarItem)
             var now = timeService.CurrentDateTime;
             var delta = (now - calendarItem.StartTime).TotalMinutes;
 
-            //Delta has to be bigger than 1, so 1 / delta < 1
+            //Delta has to be bigger or equal to 1, so 1 / delta <= 1
             if (delta < 1)
-                delta++;
+                delta = 1;
```
sofokli1: doesn't this make all negative deltas equal to 1?
If yes, wouldn't this make something that has a delta of -300 to be 1 and this and up having a very good score?

arturs105: Yes, you're right. Good catch 👍

arturs105:
Yeah, i agree that the algorithm is flawed and we should try come up with a better one.

Issue #3622, opened 2018-11-01
Start running time entry from calendar event


juanlaube:
When the current time is in between the calendar event, we should let the user create a running time entry instead of creating the full time entry with end time (an action sheet for the UI maybe?).

amulware:
I can see users wanting to start the entry a little ahead of time as well, or if they are behind schedule, afterward.

I propose that tap+hold or a force touch should open an action sheet with the actions:

- copy to time entry (the current function)
- start at event start time (only if the same is in the past)
- start now
- cancel

Might need some better copy.

The only thing that bothers me now is: What to do about the default behaviour on tapping? Should that sometimes create a stopped entry, and sometimes a running one? By what criteria? Can we make this intuitive, and not confusing?

Or will we always need to use the action sheet? :/

lizaPlotnikova:
@juanlaube When do you need these designs by?

juanlaube:
@lizaPlotnikova this is scheduled for Daneel 1.16, deadline is Jan 16th, so a few days before that would be ideal.

juanlaube:
and by "a few days" I mean a week or so 🙂

simonrozsival:
I will start by showing the action sheet after a long press. Hopefully we will sort the rest before I finish it 😄

simonrozsival:
Oops. The action sheet is ready for reviews.

amulware:
Let's do the rest of this when we have designs. :)
This is actually in 1.17, not 1.16, so we have time.

simonrozsival:
I'm closing this until we have the designs.

amulware:
I'm re-opening this because I want it in 1.17

simonrozsival:
Ok then 😆

sofokli1:
do we have designs for this?

lizaPlotnikova:
@sofokli1 not yet :)

amulware:
Very out of date.

PR #4141, opened 2019-01-10
Calendar: show an action sheet when user long-presses calendar event

simonrozsival:
## What's this?

screen shot 2019-01-10 at 12 34 02 pm

### Relationships
Ref #3622 

## Why do we want this?
We want to have a more useful calendar view.

## How is it done?
The editing helper ignored a long press for non-editable (== calendar events) items. I added a new subject/observable which emits when there is a long press on some calendar event.

I had to modify the `IDialogService.Select` constraint to allow non-reference types. We only use it in the reports view and the callsite there is OK if we drop the constraint, so I guess it can't be a problem.

All the strings are translatable.

### Side effects
I hope I didn't break anything along the way (editing time entries). Well I am sure I didn't break anything and I couldn't find a bug in the simulator. The problem is that _sometimes_ I'm sure I didn't break anything but in fact I did... so please try playing with the calendar in the simulator/on your phone and try to break it.

## Review hints
☝️ 

## :squid: Permissions
🦑


amulware:
Please add a screenshot above. :)

Toggl.Foundation/Resources.resx
```diff
@@ -495,6 +495,18 @@ account is not set up</value>
     <data name="Calendar">
         <value>Calendar</value>
     </data>
+    <data name="CalendarWhatToDoWithCalendarEvent">
+        <value>What would you like to do with the calendar event?</value>
```
amulware: `What would you like to do with this calendar event?`

Toggl.Foundation/Resources.resx
```diff
@@ -495,6 +495,18 @@ account is not set up</value>
...
+    </data>
+    <data name="CalendarStartNow">
+        <value>Start now</value>
+    </data>
+    <data name="CalendarStartWhenTheEventStarts">
+        <value>Start at event start</value>
+    </data>
+    <data name="CalendarCopyEventToTimeEntry">
+        <value>Copy to time entry</value>
```
amulware: `Copy as time entry`

ApocalypticFish:
The copy looks good.

Issue #4146, opened 2019-01-10
Create the numeric input wheel component


daividssilverio:
This issue will take care of handling the numeric input inside the wheel, it will of course have to connect to the full wheel when ready.

Numeric input:
- draw numeric input
- numeric input handling

kornelijepetak:
Closed by #4315

Issue #4145, opened 2019-01-10
Create the inner clock for the wheel


daividssilverio:
This issue should take care of drawing the inner clock for the wheel.
Whatever works better, either canvas or svgs.

daividssilverio:
Closed by #4215

Issue #4144, opened 2019-01-10
Create the base wheel view


daividssilverio:
This issue should take care of creating the basic wheel view, including:
Basic Interface:
- get,set max start time
- get,set min start time
- get,set max end time
- get,set min end time
- get,set start time
- get,set end time
- get,set isRunning
- listeners

- draw bottom arc
- draw top arc
- draw start control
- draw stop control
- touch handling

PR #4147, opened 2019-01-10
Android Wheel Component

### LINKED ISSUE #4111: Create The Android Wheel(tm) UI Duration component
- There must be an interface to set the start and end datetimes;
- There must be a listener/observable to the start and end datetimes;
- It must be possible to drag the start indicator to increase the duration;
- It must be possible to drag the stop indicator to decrease;
- There must be a duration indicator that can be editable using the soft keyboard;
- The duration indicator, start and stop indicators must be always in sync;
- The wheel must react to the time input, updating the stop and start indicators;
- It should be possible move the start/stop times at the same time by dragging in between the stop and start indicators;
- It must be pretty and easy to connect to :)
- If the rainbow colouring can be done in a separate issue if proven difficult to implement.

Please check the iOS wheel for reference.

This task can be divided into sub tasks:

Basic Interface:
- get,set max start time
- get,set min start time
- get,set max end time
- get,set min end time
- get,set start time
- get,set end time
- get,set isRunning
- listeners

Wheel (#4144):
- draw bottom arc
- draw top arc
- draw start control
- draw stop control
- touch handling

Inner clock (#4145):
- draw static clock layer

Numeric input (#4146) :
- draw numeric input
- numeric input handling

Improved UX:
- rainbow arc colouring
- angle/minute snapping
- haptic feedback (if possible)

Validation:
- prevent negative duration
- add listeners

Child of #4030

When this issue is done, the wheel should be fully functional. Sub tasks were created to make splitting the work easier.

daividssilverio:
## What's this?
This includes the android wheel component.

### Relationships
Closes #4111
Depends on 
- [x] The base wheel #4144 
- [x] The inner clock #4145 
- [x] The numeric input component #4146 

## Why do we want this?
Because the wheel is cool.

## How is it done?
We'll try to mimic the iOS wheel as much as possible, because it looks good and easy to understand.
Android is different, so it will be different.

### Why not another way?
🤔 Canvas is cool. 

### Side effects
Coolness of the android app + 1.

## Review hints
Wait until all the sub issues are done and pieces put together.

## :squid: Permissions
After WIP is done, go for it.


kornelijepetak:
The android is not buildable because we have multiple `Color` types:
Use `using Color = Android.Graphics.Color` in the relevant files.

Toggl.Giskard/Views/EditDuration/WheelForegroundView.cs
```diff
@@ -0,0 +1,436 @@
...
+            center = new PointF(radius, radius);
+            bounds = new RectF(capWidth, capWidth, Width - capWidth, Width - capWidth);
+            endPointsRadius = radius - capWidth;
+            wheelHandleDotIndicatorDistanceToCenter = radius - capWidth / 2f;
+        }
+
+        protected override void OnDraw(Canvas canvas)
+        {
+            base.OnDraw(canvas);
+            setupDrawingDelegates();
```
heytherewill: I don't like setting these up on the `OnDraw` method

Toggl.Giskard/Views/EditDuration/WheelForegroundView.cs
```diff
@@ -0,0 +1,436 @@
...
+            calculateEndPointPositions();
```
heytherewill: Don't calculate stuff in draw please

Toggl.Giskard/Views/EditDuration/WheelForegroundView.cs
```diff
@@ -0,0 +1,436 @@
...
+            updateUIElements();
```
heytherewill: Same thing. Do all calculations before calling Invalidate

Toggl.Giskard/Views/EditDuration/WheelDurationInput.InputFilter.cs
```diff
@@ -0,0 +1,64 @@
...
+using System.Reactive.Linq;
+using System.Reactive;
+
+namespace Toggl.Giskard.Views.EditDuration
+{
+    public partial class WheelDurationInput : EditText, ITextWatcher
+    {
+        private class InputFilter : Java.Lang.Object, IInputFilter
+        {
+            private Action<int> digitHandler;
```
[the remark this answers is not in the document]
daividssilverio: I don't it's worth it to create another type just to abstract the idea of a consumer. An action already passes that idea. I've changed the name to help understanding what it does from reading it.

PR #4150, opened 2019-01-10
📐🚧 Fix delete button (Swipe) hidden under play button

### LINKED ISSUE #3714: Delete button (Swipe) hidden under play button
## Context
Need at least a full timer page of entries

Build/Branch: 1.13
Phone OS: 
Phone model: iPhone SE

## Expected behavior
The play button should either not be visible during this action or the page should scroll down further when swiped

## To reproduce
Have a full timer page of entries.
Attempt to delete the last entry by swiping
Delete button is hidden behind the play button.

img_6146

sofokli1:
## What's this?
In the main log the delete button for the last cell is always under the startTE button.

### Relationships
Closes #3714

## Why do we want this?
We want to be able to delete the last TE in the main log

## How is it done?
Added tableview bottom inset

### Why not another way?
We could add a table footer view with the same height as the desired spacing but this seemed simpler.

### Side effects
None

## Review hints
Run the app

## :squid: Permissions
:squid:


Toggl.Daneel/ViewSources/TimeEntriesLogViewSource.cs
```diff
@@ -20,7 +20,7 @@ public sealed class TimeEntriesLogViewSource : ReactiveSectionedListTableViewSou
     {
         private const int rowHeight = 64;
         private const int headerHeight = 48;
-        private const int spaceBetweenSections = 20;
+        public const int SpaceBetweenSections = 20;
```
simonrozsival: This public `const` should be moved after the `private` fields.

PR #4160, opened 2019-01-11
:memo: Corrects spelling mistake and makes minor copy changes

ApocalypticFish:
Relatively minor copy changes and spelling mistake correction.


Toggl.Foundation/Resources.resx
```diff
@@ -297,10 +297,10 @@ categorize your tracked time.</value>
         <value>Help</value>
     </data>
     <data name="WorkspaceChangedAlertTitle">
-        <value>You've changed the workspace.</value>
+        <value>You've changed workspace.</value>
```
amulware: Why? Where is this used?

Toggl.Foundation/Resources.resx
```diff
@@ -297,10 +297,10 @@ categorize your tracked time.</value>
...
     </data>
     <data name="WorkspaceChangedAlertMessage">
-        <value>Your project and time entry will be allocated under the selected workspace.</value>
+        <value>Your project and time entry will be allocated to the selected workspace.</value>
```
amulware: Why?

Toggl.Foundation/Resources.resx
```diff
@@ -319,7 +319,7 @@ categorize your tracked time.</value>
 account is not set up</value>
     </data>
     <data name="NoEmailErrorMessage">
-        <value>Go to “Settings > Mail > Add Account” to set up an email account or send us your feedback to support@toggl.com</value>
+        <value>Go to “Settings > Mail > Add Account” to set up an email account or send your feedback to support@toggl.com</value>
```
amulware: Why?

Toggl.Foundation/Resources.resx
ApocalypticFish: Under doesn't seem correct in this circumstance. Generally, things are allocated to projects, have tags assigned. It's a minor grammatical change but under doesn't seem correct here. They are attributed to that workspace not under it.

Toggl.Foundation/Resources.resx
ApocalypticFish: The workspace doesn't seem grammatically correct here, the workspace is where you're working. If you had changed between offices you would say you have changed office or have switched offices not you have changed the office as this suggests that you have made changes to the office or the workspace. It's not possible to make changes to the workspace from the app so this isn't the case.

Toggl.Foundation/Resources.resx
ApocalypticFish: We don't need to be identified again. It is implied that the feedback is sent to us. We don't need to further identify this by explaining that they are sending their feedback to "us". While technically still a correct and viable sentence, in my opinion, it seems more like a non-native English sentence than one a native speaker would use or is perhaps more professional seeming. Whether this is down to attempting to use the minimal amount of words/characters to convey a message or another reason I can't tell you.

amulware: It feels a bit impersonal to me. 'send your feedback' who knows who gets it, or if anybody is gonna read it. That being said, I don't think this sentence is used anywhere anymore, and could actually be removed. But I'll let someone else check that and figure it out.

ApocalypticFish: Okay what about "send your feedback to us at support@toggl.com" 
or "send your feedback to us support@toggl.com"

ApocalypticFish: @amulware what do you think? I can make the changes and resubmit for review?

PR #4164, opened 2019-01-11
Add RandomElement function to simplify some use cases

kornelijepetak:
## What's this?
Just a simple extension method that chooses a random element from the list.

### Relationships
Related to Edit Project screen and initial color selection. 

## Why do we want this?
Just code simplification. Nothing special.

## How is it done?
The usuals.

### Why not another way?
It's as simple as possible. And also working only on collections with known count.

### Side effects
None.

## Review hints
If you need hints, maybe this job is not for you.

## :squid: Permissions
Anyone, anywhere.


amulware:
If you need this to be threadsafe, consider using the `[ThreadStatic]` attribute.

Toggl.Multivac/Extensions/EnumerableExtensions.cs
```diff
@@ -93,5 +95,16 @@ public static TimeSpan Sum<T>(this IEnumerable<T> collection, Func<T, TimeSpan?>
         public static bool ContainsExactlyAll<T>(this IReadOnlyList<T> collection, IReadOnlyList<T> otherCollection)
             => collection.Count == otherCollection.Count
                && collection.Except(otherCollection).None();
+
+        public static T TakeAtRandom<T>(this IList<T> collection)
```
amulware: I suggest renaming to `TakeRandomElement`, or maybe just `RandomElement` to make it clear we're only taking a single item.

Toggl.Multivac/Extensions/EnumerableExtensions.cs
```diff
@@ -6,6 +6,8 @@ namespace Toggl.Multivac.Extensions
 {
     public static class EnumerableExtensions
     {
+        private static Random random = new Random();
```
amulware: Is static state the best solution here?

PR #4177, opened 2019-01-11
Disable the login button on iOS when needed

dshr:
## What's this?
Adding a disable subscription to our login button on iOS, we forgot to add it 🙈 

### Relationships
Closes nothing

## Why do we want this?
Consistency.

## How is it done?
By adding the correct subscription.

### Why not another way?
This is the only way.

### Side effects
None

## Review hints
Try a malformed email input.

## :squid: Permissions
All yours!


khoi:
we have the shaking behavior on iOS. This was why we didn't disable the button

dshr:
Closing this then, will change the test to look at iOS and Android separately.

PR #4181, opened 2019-01-11
Add Color struct

### LINKED ISSUE #4162: Implement our own Color type
This would replace `MvxColor` so we also need the `ToNativeColor` implementations on both platforms.

arturs105:
## What's this?
This adds a `Color` struct to replace `MvxColor`. In the future we should use this struct instead of `MvxColor` and instead of hex strings in view models (similar to `Password` and `Email`), but for now, this PR just adds the struct and extension methods.

### Relationships

Closes #4162

## Why do we want this?
MvExit FTW!

## :squid: Permissions
🦑


Toggl.Giskard/Startup/TogglApplication.cs
```diff
@@ -20,7 +20,7 @@ public override void OnCreate()
             Firebase.FirebaseApp.InitializeApp(this);
 #if USE_ANALYTICS
             Microsoft.AppCenter.AppCenter.Start(
-                "{TOGGL_APP_CENTER_ID_DROID}",
+                "",
```
juanlaube: 👀

PR #4192, opened 2019-01-14
Add docs with RX binding guidelines

kornelijepetak:
## What's this?
This is the proposal for a document that describes the common use cases and how to act in regards to RX binding.

Note that the documentation talks about Relays and binding extensions that are only propagating the changed event further if there actually was a change. That is why additional two files are included that should be included in the codebase as they are the prerequisites for this to work.

The docs can be read in the Markdown [preview mode here](https://github.com/toggl/mobileapp/blob/c8a2a860ba01c18083f5b436938191528e7a0426/docs/rx-binding-guidelines.md). _I know that you know how to get there, link is here to provide ease of access._

### Relationships
Every [RX refactoring issue](https://github.com/toggl/mobileapp/projects/34) and beyond.

## Why do we want this?
To create an agreement about how we're going to do things in the future after MVX is long gone. That should make the code easier to reason about (expectation-wise) and in the long run make our codebase a bit more consistent.

## How is it done?
By specifying common use cases and provide resolution for them in the form of a cookbook.

### Why not another way?
That is what you are here to say. With arguments, not 🔫s.

### Side effects
None percievable.

## :squid: Permissions
After at least 👨‍🔬👨‍🔬👨‍🔬👨‍🔬👨‍🔬 read and approve.


Toggl.Giskard/Extensions/Reactive/TextViewExtensions.cs
```diff
@@ -19,8 +19,20 @@ public static IObservable<ICharSequence> TextFormatted(this IReactive<TextView>
...
+        {
+            return text =>
+            {
+                if (!ignoreUnchanged)
+                {
+                    reactive.Base.Text = text;
+                    return;
+                }
+
+                if (reactive.Base.Text != text)
```
daividssilverio: I always get confused when I see the equality comparator used with strings. Java was annoying with this.

kornelijepetak: Yeah, don't worry. 
This is a dream. (oh, if anyone could understand this joke).

docs/rx-binding-guidelines.md
```diff
@@ -0,0 +1,198 @@
...
+
+And the corresponding view part:
+
+```csharp
+SaveButton.Rx()
+    .BindAction(ViewModel.Save)
+    .DisposedBy(disposeBag);
+```
+
+**Manual subscription**. If you don't need the bells and whistles that is the `.Errors` and `.IsExecuting` from RxAction, or your control **MUST NOT** be disabled/enabled every time an action is being executed, subscribe to the action's `.Inputs` manually.
```
Odrakir: `.Errors` and `.Executing` (not `.IsExecuting`) will still work even if you execute the action through the `.Inputs` property.

docs/rx-binding-guidelines.md
```diff
@@ -0,0 +1,198 @@
+﻿## Binding Views to ViewModels Cookbook
+
+* We have started to move away from Mvx Magic (aka ✨) in the way it deals with the bindings and started using RX bindings. In the transitional period, we used different ways to achieve the same thing. That has created lots of bugs and inconsistencies, but over the time we realized and fixed many of the internal mistakes we've made.
+* However, we still need the guidelines for how we should approach the binding dilemmas and resolve all instances the same way. That will be easier to comprehend, review and fix.
+* This document proposes the way we should approach binding and while not all of the code adheres to those guidelines, a slow switch could be made over time to be as consistent as possible.
+<br />
+
```
Odrakir: :fire:

docs/rx-binding-guidelines.md
kornelijepetak: Oh, yes. They would work, but manual subscription will not subscribe to them. But maybe the text is not clear on that. I will fix it.

docs/rx-binding-guidelines.md
```diff
@@ -0,0 +1,197 @@
...
+🛠 **How.** Expose an RX action.
+
+We are using two flavors of Rx actions. 
+* `UIAction` that simple does some work as a result of an action
+* `InputAction<T>` that does some work as a result of an action with some data of type `T` attached.
+
+Typical example for `UIAction` is a button tap, while the switch button should be used with `InputAction<bool>`.
+
+⚠ **IRxActionFactory**. Make sure to inject `IRxActionFactory` into your view model and initialize all actions by using methods from that interface. There are three methods to be used from that interface, depending on what kind of handler is handler is covering the action.
```
sofokli1: > depending on what kind of handler is handler is covering the action.

too many `handler is`?

docs/rx-binding-guidelines.md
```diff
@@ -0,0 +1,197 @@
...
+Then initialize it either from a subject, or derive from some other observable source.
+
+```csharp
+BehaviorSubject<bool> isEmptySubject { get; } 
+	= new BehaviorSubject<bool>(false);
+
+//...	
+
+{
+	IsEmpty = isEmptySubject.AsObservable();
```
khoi: I think we should put `AsDriver` as a requirement for all Observable that is going to drive a UI element. Since `AsDriver` is safer, and it provides a recovered state to the UI when there is an error.

docs/rx-binding-guidelines.md
```diff
@@ -0,0 +1,197 @@
...
+
+And the corresponding view part:
+
+```csharp
+SaveButton.Rx()
+    .BindAction(ViewModel.Save)
+    .DisposedBy(disposeBag);
+```
+
+**Manual subscription**. If you don't need the bells and whistles that is the `.Errors` and `.Executing` from RxAction, or your control **MUST NOT** be disabled/enabled every time an action is being executed, subscribe to the action's `.Inputs` manually. This will not prevent the `.Errors` and `.Executing` from working but since the view will then not be subscribed to them (unlike with `.BindAction`), they would not be disabled/enabled on every change.
```
khoi: `subscribe to the action's `.Inputs` manually.` 
this is confusing to me, we're not subscribing to `Inputs` but rather send elem to it.

docs/rx-binding-guidelines.md
```diff
@@ -0,0 +1,197 @@
...
+
+The view model part is the same as with `BindAction`, but the subscription is manual:
+
+```csharp
+SaveButton.Rx().Tap()
+    .Subscribe(ViewModel.Save.Inputs)
```
khoi: And also isn't sending items directly to `Inputs` is dangerous? Tapping SaveButton twice would cause a 💥 if there is a pending save

docs/rx-binding-guidelines.md
kornelijepetak: I like my handlers.

docs/rx-binding-guidelines.md
kornelijepetak: I was changing this sentence a few times and a wrong word remained. It was supposed to read `with` instead of `to`.

docs/rx-binding-guidelines.md
kornelijepetak: I have added another paragraph explaining the issue. Thanks for noticing this here.

docs/rx-binding-guidelines.md
```diff
@@ -31,7 +32,7 @@ toolbar.Title = ViewModel.Title;
 ```csharp
 public IObservable<bool> IsEmpty { get; private set; } 
 ```
-Then initialize it either from a subject, or derive from some other observable source.
+Then initialize it either from a subject, or derive from some other observable source. Note that in almost all cases, a driver should be exposed through the observable, so please use `AsDriver` and not merely `AsObservable`. `AsDriver` prevents execution on inappropriate thread and also handles the errors gracefully by providing the fallback value to the view in case of an error.
```
Odrakir: Maybe adding that it also shares the subscription so side effects are not triggered more than once even if it has many subscribers?

docs/rx-binding-guidelines.md
```diff
@@ -0,0 +1,199 @@
...
+
+```csharp
+NameTextView.Rx().Text()
+	.Subscribe(ViewModel.Name.Accept)
+	.DisposedBy(DisposeBag);
+```
+
+> 💡 The reason why should use `Relay` here instead of `RxAction` is as follows. 
+> * Relays are very lightweight objects without all the bells and whistles of an RxAction. Many of the RxAction features are not needed for many tasks (such as simple propagation of data changes in widgets). Having a separate error observable is very often not needed, or preventing parallel execution. 
+> * RxAction prevents multiple parallel executions by utilizing it's `Executing` property which can cause glitches/crashes in combination with some widgets. Basically, as long as a widget can freely emit events, without being able to be disabled by action's `IsEnabled` property, it will create a 💥 hazard. For some widgets, introducing the `Enabled` property could change the UX significantly so it should be avoided (i.e. Text field being disabled/enabled after each letter typed in).
```
Odrakir: The property is called `Enabled`.

Odrakir: What causes the UI element to being enabled/disabled is the `BindAction` method, not the use of `RxAction` it self.

As you said, `RxAction` does prevent multiple parallel executions, but if you do try to execute twice all that'll happen is that an error will be returned in its `Errors` property, but you'll be able to keep using the action later on.

So, using `RxAction` with textfields is fine (although you probably want to use relays instead for most cases, because you are probably binding that textfield with a subject anyway and you don't need most of what `RxAction` offers). What's not possible is using `BindAction` with text fields. That method doesn't exist and we should not create it.

docs/rx-binding-guidelines.md
```diff
@@ -0,0 +1,199 @@
...
+
+And the corresponding View part:
+
+```csharp
+SaveButton.Rx()
+    .BindAction(ViewModel.Save)
+    .DisposedBy(disposeBag);
+```
+
+**Manual subscription**. If you don't need the bells and whistles that is the `.Errors` and `.Executing` from RxAction, or your control **MUST NOT** be disabled/enabled every time an action is being executed, subscribe with the action's `.Inputs` manually. This will not prevent the `.Errors` and `.Executing` from working but since the View will then not be subscribed to them (unlike with `.BindAction`), they would not be disabled/enabled on every change.
```
Odrakir: `Errors` and `Executing` still work if you use `Inputs` instead of `BindAction`, so the first phrase is confusing. The view won't subscribe to `Errors` and `Executing` in any case (wether you use `BindAction` or not) so the second phrase is confusing.

All that changes here is wether the ui component is enabled/disabled as the `Enabled` property of the action changes. As a reference, [this](https://github.com/toggl/mobileapp/blob/e966265827977ecb3594ccb4a4d70873439822f6/Toggl.Daneel/Extensions/Reactive/UIButtonExtensions.cs#L68) is all `BindAction` does, as you can see it only uses the `Enabled` property and forwards events to the `Inputs` of the action.

docs/rx-binding-guidelines.md
```diff
@@ -0,0 +1,199 @@
...
+
+### 📘 ViewModel-To-View binding (one way binding)
+🏆 **When.** If a ViewModel exposes a value that is driven by the viewmodel itself, or some other source that is otherwise completely independent on the view.
+
+🛠 **How.** Expose the value as an observable.
+
+```csharp
+public IObservable<bool> IsEmpty { get; private set; } 
+```
+Then initialize it either from a subject, or derive from some other observable source. Note that in almost all cases, a driver should be exposed through the observable, so please use `AsDriver` and not merely `AsObservable`. `AsDriver` prevents execution on inappropriate thread and also handles the errors gracefully by providing the fallback value to the View in case of an error.
```
Odrakir: Maybe you can also mention that drivers share side effects. So even if the property has many subscribers, its side effects will only be executed once, which is not the case for plain cold observables (in other words, if you don't use drivers you could, for example, be firing network requests for every view that subscribes to a view model property)

docs/rx-binding-guidelines.md
kornelijepetak: You are correct.

PR #4193, opened 2019-01-14
🚧 Delete swiped TE when changing views

### LINKED ISSUE #3734: Swipe-deleted time entries still showing in calendar view
## 🐛 Bug Report
Entries that are swiped to delete still render in calendar view for a few seconds, which is weird.

## Context
Any iOS right now.

## Expected behavior
When I delete an entry, in any way, I'd expect it to be gone from anywhere else.

Since it's so easy to go over to the calendar screen, that expectation is broken.

## To reproduce
- have calendar view setup
- swipe-to-delete a time entry TODAY
- go to calendar quickly
- TE still shows for ~5 seconds

## Suggestions
We could fix this by deleting the entry in case the user navigates away from the main view very quickly. However, I'm not sure about what the expected behaviour is in cases like that. Does the user expect to be able to go back to the main view and still undo the deletion?

Right now, we don't delete the entry for example if the app goes into the background, so if the user force closes the app, the entry will reappear, that's maybe not ideal, or maybe it's exactly what we want?

sofokli1:
## What's this?
Completes the deleting of the toBeDeleted TE that was swiped to delete when changing views.

### Relationships
Closes #3734

## Why do we want this?
As the issue describes when swiping to delete the TE is not actually deleted until 5sec have passed thus ending up being visible in the calendar view for that amount of time.

## How is it done?
When the main view disappears it finalises the deletion thus not encountering the above issue anymore.

### Why not another way?
This is the simplest without making too many complicated changes that can backfire.

### Side effects
If the view changes, the 5sec undo logic stops immediately. Otherwise it works as usual.

## Review hints
Run the app make an entry, swipe delete it and go to calendar in less than 5sec after deleting.

## :squid: Permissions
:squid:


Toggl.Foundation.MvvmCross/ViewModels/MainViewModel.cs
```diff
@@ -352,6 +352,17 @@ private async Task continueMostRecentEntry()
             await interactorFactory.ContinueMostRecentTimeEntry().Execute();
         }
 
+        public override void ViewDisappeared()
+        {
+            base.ViewDisappeared();
+            ViewDisappearedAsync();
+        }
+
+        internal async Task ViewDisappearedAsync()
```
simonrozsival: This can be private method, we don't need to call it from anywhere else.

amulware:
I'd like @maritoggl to be aware of and sign off on this change.

Toggl.Foundation.Tests/MvvmCross/ViewModels/TimeEntriesViewModelTests.cs
```diff
@@ -311,6 +306,8 @@ public async ThreadingTask RemovesTheWholeCollectionWhenThereAreNoOtherTimeEntri
             }
         }
 
+        //public sealed class
```
simonrozsival: 👀

PR #4210, opened 2019-01-15
Add info about RxActionFactory

Odrakir:
## What's this?
This is just a small update to the `RxAction` doc to talk about the way to create actions with a scheduler through `RxActionFactory`

### Relationships
Doesn't close any issue

## Why do we want this?
Just so docs are up to date with code.

## :squid: Permissions
You can.


Odrakir:
@juanlaube Check again, please.

docs/RxAction.md
```diff
@@ -117,6 +127,18 @@ checkEmailExist.Executing
...
+
+```c#
+public UIAction Close { get; }
+...
+Close = rxActionFactory.FromAsync(close);
+...
+private Task close() => navigationService.Close(this, defaultResult);
+```
+
+A `UIAction` is a subclass of `RxAction` without inputs or outputs, meaning `RxAction<Unit, Unit>``
```
kornelijepetak: Double tick at the end messed up the formatting.

docs/RxAction.md
```diff
@@ -16,6 +16,16 @@ Regardless of the value sent as `enabledIf` the RxAction is disabled while it's
 
 Calling the `Execute` method will perform the underlying action, *regardless of whether there are any subscriptions to it*. The fact that it's returning an Observable is just there for the convenience, in case we want to be informed of the state of the execution of the action, as it'll emit the elements the action returns and its potential errors.
 
+## Threading and Action Factory
+
+As we want actions to return on the main thread we created a `RxActionFactory` service that simplifies the process of creating an action while passing the appropriate scheduler to it. ViewModels that want to use `RxActions` should include this dependency and call one of its methods to create `UIActions` or `InputActions` as needed.
+
+```c#
+StartTimeEntry = rxActionFactory.FromAsync<Suggestion>(suggestion => startTimeEntry(suggestion));
```
kornelijepetak: Maybe it was intentional for demonstration purposes, but here the type `<Suggestion>` is automatically inferred from the given delegate, and also delegate could be written as a method group instead of lambda. 

See if it makes sense to you to change this into the format it's actually gonna be used:

`StartTimeEntry = rxActionFactory.FromAsync(startTimeEntry);`

Odrakir: I think you can't remove the `<Suggestion>` because if you do the compiler will think you are trying to use the non-generic override which is used for `UIAction` and not `InputAction`.

About the lambda, you are right, but I think it's more clear this way for documenting purposes.

kornelijepetak: I've checked the interface, yes, you are right. There's a non-generic one so it can't be omitted.

PR #4211, opened 2019-01-15
Reports: last year shortcut

### LINKED ISSUE #4191: "Last year" shortcut to mobile reports
For consistency purposes in all our apps, let's add a shortcut for "Last year" to mobile reports. 

This should display data for the full last year, similarly to "Yesterday", "Last week" and "Last month". 

Please pay attention to the fact that we currently don't allow scrolling back more than one year in the calendar and any complications this may bring about.

simonrozsival:
## What's this?
This PR extends the range from which the user can select ranges from the last 25 months (the only restriction is that the range can't be longer than 365 days).

### Relationships
Closes #4191 

## How is it done?
The 25 months is necessary to be _always_ show the range in the calendar.

It's just another shortcut, the code is very straightforward.

### Why not another way?
The number of months to show so that the December of the year before the last year could be calculated and so we would always show between 13 to 25 last months but I think it would be an overkill.

There is a second option: keep the limit at 13 months so that the user can't select different ranges themselves, but keep just the shortcut. The app can get data for a range which cannot be displayed in the calendar. If @amulware or @maritoggl think this would be a better idea, please request changes to this PR.

### Side effects
The user can select very long ranges (over 365 days) "easily" now. It was already possible before (we showed 13 months) and there is a check which prevents the user from doing it so I don't see a problem in that.

## Review hints

Try this in the emulator.

## :squid: Permissions
🦑


maritoggl:
25 months seems reasonable - leaves more flexibility to the user to play around.

PR #4212, opened 2019-01-15
Start time entry with multiple tags test

### LINKED ISSUE #4180: Start TE: Create multiple tags
Sign in
Start entry
Create multiple tags
Save entry
Confirm successful sync status shown to test that this element is displaying correctly.

heytherewill:
## What's this?
Self descriptive PR

### Relationships
Closes #4180

## Why do we want this?
To prevent regressions

## How is it done?
- Open Start View
- Add a bunch of tags
- Save
- Open edit view
- Open edit tags
- Ensure all tags are there

### Why not another way?
Tags are not shown on the log view, so we need to open the start view
It's hard to check for the tag's text on the edit view, so we need to open the edit tags view

## Review hints
Run the test on both platforms

## :squid: Permissions
Anyone, anytime


daividssilverio:
Then they are flaky @juanlaube and @heytherewill.
Test failure:
```
System.Exception : Error while performing Tap(Marked("Create tag "Kale""))
  ----> System.Exception : Unable to find element. Query for Marked("Create tag "Kale"") gave no results.
```
Screenshot for reference:
image

daividssilverio:
Testing on my S8 and Nexus 5:
- Changing keyboards doesn't change the output of the test.
- Adding a space before creating a tag in the `AddTagInStartView` extension "fixes" the test

heytherewill:
@daividssilverio I made a change, could you give it another spin?

Toggl.Giskard.Tests.UI/Extensions/StartTimeEntryExtensions.cs
```diff
@@ -6,6 +6,7 @@ public static partial class StartTimeEntryExtensions
     {
         public static void TapCreateTag(this IApp app, string tagName)
         {
+            app.WaitForElement($"Create tag \"{tagName}\"");
```
daividssilverio: It seems this isn't enough to fix the test.
Waiting for the "create tag" text to disappear after the tap seems to help.
```
app.WaitForNoElement($"Create tag \"{tagName}\"");
```

PR #4214, opened 2019-01-15
UI Test project creation

### LINKED ISSUE #4167: Start TE: Create a new project
Sign in
Start time entry
Create new project
Save entry
Stop Entry
Successful sync status displayed?

heytherewill:
## What's this?
This adds a test to ensure that new Project behaviour works fine

### Relationships
Closes #4167

## Why do we want this?
Regressions are mean

## How is it done?
Very similar to #4212, but we can assert on the main view

## Review hints
Run the tests on both platforms

## :squid: Permissions
Only myself


Toggl.Daneel.Tests.UI/Extensions/StartTimeEntryExtensions.cs
```diff
@@ -0,0 +1,17 @@
+﻿using Xamarin.UITest;
+
+namespace Toggl.Tests.UI.Extensions
+{
+    public static partial class StartTimeEntryExtensions
+    {
+        public static void TapCreateTag(this IApp app, string tagName)
+        {
+            app.TapNthCellInCollection(0);
```
daividssilverio: 🤔 This doesn't look like a good idea, we have to setup the test to make sure the first item in the collection is the cell to create tags/projects.
Using the same strategy used on Android doesn't work? Why?

Toggl.Giskard.Tests.UI/Extensions/StartTimeEntryExtensions.cs
```diff
@@ -0,0 +1,17 @@
...
+namespace Toggl.Tests.UI.Extensions
+{
+    public static partial class StartTimeEntryExtensions
+    {
+        public static void TapCreateTag(this IApp app, string tagName)
+        {
+            app.Tap($"Create tag \"{tagName}\"");
+        }
+
+        public static void TapCreateProject(this IApp app, string tagName)
```
daividssilverio: `tagName` -> `projectName`

Toggl.Tests.UI/StartTimeEntryTests.cs
```diff
@@ -76,5 +77,22 @@ public void TappingDoneAfterTryingToDiscardAndEditingTheDescriptionDoesNotChange
...
+
+            var projectName = "Meme Production";
+            app.CreateProjectInStartView(projectName);
+
+            app.Tap(StartTimeEntry.DoneButton);
+            app.Tap(Main.StopTimeEntryButton);
+
+            app.PullToRefresh();
+
+            app.WaitForElement(projectName);
```
daividssilverio: on iOS:
```
System.Exception : Error while performing WaitForElement(Marked("Meme Production"), "Timed out waiting for element...", null, null, null)
  ----> System.TimeoutException : Timed out waiting for element...
```

heytherewill: Were you looking at your screen during the test fail? Because this might have been caused by the keyboard not showing up, something easily fixed by pressing Command + K

daividssilverio: Whot, the soft keyboard was enabled. It came all the way from startup through signup to the point where it should find the time entry with the "Meme Production" project label in the Main log. Keyboard was there.

Toggl.Daneel.Tests.UI/Extensions/StartTimeEntryExtensions.cs
daividssilverio: Can't guarantee the "Always" part, but I won't argue with this much, since if the layout does change, the UI Tests would have to be updated anyway. If it works, fine.

PR #4215, opened 2019-01-15
Droid background wheel: background wheel and inner clock

### LINKED ISSUE #4145: Create the inner clock for the wheel

daividssilverio:
## What's this?
This is PR adds the part of the wheel component that draws the background of the wheel that is basically static. The background is composed of the grey arc and the numbers that represent the dial on the wheel.

### Relationships
Closes #4145
Part of #4147 

## Why do we want this?
It's part of the wheel, it would be weird without it.

## How is it done?
I've added another view for the background wheel that is basically composed of a full wheel and clock dial. That's it, the logic is very similar to the ForegroundWheel, except there's no user interaction with this view.

### Why not another way?
This was very easy to do. Feel free to suggest something else.

### Side effects
None.

## Review hints
You can see the view if you put it in some layout inside the app. You have to specify the width and height of the view. Anything like:
```
<?xml version="1.0" encoding="utf-8"?>
<FrameLayout xmlns:android="http://schemas.android.com/apk/res/android"
             xmlns:app="http://schemas.android.com/apk/res-auto"
             android:layout_height="wrap_content"
             android:layout_width="match_parent">

    <toggl.giskard.views.WheelBackgroundView  
            android:layout_marginTop="40dp"
            android:layout_gravity="center_horizontal"
            android:layout_height="296dp"
            android:layout_width="296dp" />

    <toggl.giskard.views.WheelForegroundView
            android:layout_marginTop="40dp"
            android:layout_gravity="center_horizontal"
            android:layout_height="296dp"
            android:layout_width="296dp"/>
</FrameLayout>
```
The view looks like this (the background view draws the grey part of the arc and clock dial (60, 5, 10...):
image

## :squid: Permissions
Merge it please. I'd like to keep the commits' history to make reviewing #4147 easier.

### DIFF
--- a/Toggl.Giskard/Views/EditDuration/Shapes/ClockDial.cs
+++ b/Toggl.Giskard/Views/EditDuration/Shapes/ClockDial.cs
@@ -6,8 +6,9 @@ namespace Toggl.Giskard.Views.EditDuration.Shapes
 {
     public class ClockDial
     {
-        private const int minuteSegmentsPerHourMark = MinutesInAnHour / HoursOnTheClock;
         private const float angleOffsetCorrection = (float) FullCircle / 4f;
+        private const char numberPaddingChar = '0';
+        private const int digitsCount = 2;
         private readonly PointF dialCenter;
         private readonly float textRadius;
         private Rect textBounds = new Rect();
@@ -28,7 +29,7 @@ namespace Toggl.Giskard.Views.EditDuration.Shapes
 
         public void OnDraw(Canvas canvas)
         {
-            for (var minute = 5; minute <= MinutesInAnHour; minute += 5)
+            for (var minute = 0; minute < MinutesInAnHour; minute += 5)
             {
                 var angle = (float) FullCircle * (minute / (float) MinutesInAnHour) - angleOffsetCorrection;
                 drawMinuteNumber(canvas, minute, angle);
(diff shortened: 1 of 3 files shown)


Toggl.Giskard/Views/EditDuration/Shapes/ClockDial.cs
```diff
@@ -0,0 +1,52 @@
...
+            this.dialCenter = dialCenter;
+            this.textRadius = textRadius;
+            paint.TextSize = textSize;
+            paint.Color = textColor;
+            paint.SetTypeface(Typeface.Create("sans-serif", TypefaceStyle.Normal));
+        }
+
+        public void OnDraw(Canvas canvas)
+        {
+            for (var minute = 1; minute <= MinutesInAnHour; ++minute)
```
kornelijepetak: why not?

```
for(var minute = 0; minute < MinutesInAnHour; minute += 5) {
	var angle = (float)FullCircle * (minute / (float)MinutesInAnHour) - angleOffsetCorrection;
	drawMinuteNumber(canvas, minute, angle);
}  
```

kornelijepetak: It's not like the number of segments will ever change or that the number of minutes between two segments would be 7, or 11 or 12.4 in any forseeable future.

daividssilverio: Oh, that's a very good point.

PR #4216, opened 2019-01-15
Long description test

### LINKED ISSUE #4168: Start TE: Add long description (character limit test)
Sign in
Start time entry
Enter description that exceeds the 2000 character character limit.
Does the error show successfully?
Does it prevent you from saving the entry?

heytherewill:
# What's this?
This adds a test to ensure that something goes wrong when adding huge descriptions

### Relationships
Closes #4168

## How is it done?
Add a massive chunk of text and then check for errors

## Review hints
Apparently we don't have anything preventing the user from actually saving the time entry (which IMO is completely fine, since this check is super fiddly and backend is doing it anyways), but just in case we want client side validation I'm pinging @maritoggl here

For reviewers: The UI takes a long while to enter that much text on iOS (almost a full minute) and it freezes (for about 25 seconds) on Android, so bear this in mind when testing.

## :squid: Permissions
Only me


daividssilverio:
Based on what was said internally, I guess this test isn't needed.

Toggl.Tests.UI/StartTimeEntryTests.cs
```diff
@@ -38,6 +41,24 @@ public void TappingTheDoneButtonCreatesANewTimeEntryWhoseDescriptionMatchesWhatW
...
         }
 
+        [Test]
+        public void CreatingATimeEntryWithTooManyCharactersInTheDescriptionMakesTheEntryErrorWhenSyncing()
+        {
+            var description = Enumerable
+                .Range(0, 3001)
+                .Select(_ => "a")
+                .Aggregate(new StringBuilder(), (builder, character) => builder.Append(character))
+                .ToString();
```
amulware: I mean... `new String('a', 3001)`?

heytherewill: NOT FUNCTIONAL ENOUGH

heytherewill:
As discussed internally, this is no longer needed

PR #4229, opened 2019-01-17
✨ Add Adjust analytics

### LINKED ISSUE #4205: Integrate with Search Ads attribution API using Adjust
Some info can be found here: https://github.com/toggl/mobileapp/issues/3991

Coordinate with @djalex15 with regards to any needed API tokens or other secrets (and use cake to replace them as usual).

sofokli1:
## What's this?
This adds adjust to daneel for tracking signups, retention, sessions etc. for users that install the app through apple search ads.

### Relationships
Closes #4205

## Why do we want this?
We want to know how we get people and what they do when they install from payed ads

## How is it done?
- Added the Adjust sdk to daneel (they have one for ios and a separete one for android)
- Added an event to track successful signups (since the sdk is platform specific and can only be installed in iOS projects I had to create an observable to get that info to daneel to fire the adjust event)
- Added the adjust token to bitrise secret and our cake script

### Why not another way?
no other way

### Side effects
None

## Review hints
Talk to me about a ready script to run in sandbox mode and if you need an adjust account.

## :squid: Permissions
:squid:


Toggl.Daneel/ViewControllers/SignupViewController.cs
```diff
@@ -240,6 +245,14 @@ private void prepareViews()
 
         private string signupButtonTitle(bool isLoading)
             => isLoading ? "" : Resources.SignUpTitle;
+
+        private void logAdjustSignupEvent()
+        {
+            #if USE_ANALYTICS
+            var adjustEvent = ADJEvent.EventWithEventToken("b1qugc");
```
simonrozsival: What is the meaning of the token? Is this an arbitrary value? Anyway, please move it to a constant (IMO it can be a local constant, but it should have some name).

Toggl.Foundation.MvvmCross/ViewModels/SignupViewModel.cs
```diff
@@ -175,6 +178,11 @@ public enum ShakeTargets
                     (email, password, isLoading, countryName) => email.IsValid && password.IsValid && !isLoading && (countryName != Resources.SelectCountry))
                 .DistinctUntilChanged()
                 .AsDriver(this.schedulerProvider);
+
+            SuccessfulSignup = successfulSignupSubject
+                .Skip(1)
```
simonrozsival: This `Skip` feels weird. Why is it a behavior subject anyway? Don't drivers repeat the last emitted value anyway?

Toggl.Foundation.Tests/MvvmCross/ViewModels/SignupViewModelTests.cs
```diff
@@ -878,6 +924,9 @@ public void NavigatesToMainViewModel()
                 TestScheduler.Start();
                 NavigationService.Received().ForkNavigate<MainTabBarViewModel, MainViewModel>();
             }
+
+
+
```
simonrozsival: 👀

Toggl.Foundation.MvvmCross/ViewModels/SignupViewModel.cs
```diff
@@ -175,6 +178,9 @@ public enum ShakeTargets
                     (email, password, isLoading, countryName) => email.IsValid && password.IsValid && !isLoading && (countryName != Resources.SelectCountry))
                 .DistinctUntilChanged()
                 .AsDriver(this.schedulerProvider);
+
+            SuccessfulSignup = successfulSignupSubject
+                .AsObservable();
```
simonrozsival: This should stay as a driver (https://github.com/toggl/mobileapp/blob/develop/docs/rx-binding-guidelines.md).

PR #4235, opened 2019-01-17
Calendar: total tracked time UI

### LINKED ISSUE #4052: Add total tracked time to Calendar View

simonrozsival:
## What's this?
This PR adds a UI for the total tracked time. When there is a running time entry, the total is ticking.

### Relationships
Closes #4052 

✅ ~Needs #4259 to be fixed before this PR can be merged.~

## How is it done?
I just added observables to the VM and added bindings in the VC.

The "extended navbar" is faked using the IB, there is nothing really smart going on. I tried looking into how to put the views directly into the navbar, but that didn't work well.

### Why not another way?
I tried to make it very straightforward.

I could also expose the preferences and the `TimeSpan` and `DateTimeOffset` directly and combine those observables only in the VC, but I decided it makes more sense to expose only the information the view actually needs, which is the formatted string it should just present to the user.

### Side effects
The calendar itself is a bit shorter because the daily total takes a few pixels, but that's by design.

## Review hints
Try it in the emulator. Add a few TEs. Change preferences. Add a running TE. Simulate having the app open at midnight and observe the date changing.

## :squid: Permissions
🦑


simonrozsival:
Time zones are hard... the tests work locally, but they are apparently incorrect, so they need fixing.

PR #4236, opened 2019-01-17
Fix missing driver in EditDurationViewModel

simonrozsival:
## What's this?
Bugfix.

### Relationships
Leftover from #4088 

## Why do we want this?
No 💥 

## How is it done?
`AsDriver(schedulerProvider)`

## :squid: Permissions
🦑


Odrakir:
Should we write tests to prevent this in the future?

simonrozsival:
Let's discuss how to prevent this, but I would like to merge this now and tackle tests or some other kind of incorrect-thread-prevention in a separate PR and with separate issues.

Issue #4230, opened 2019-01-17
NullReferenceException in ConcurrencyAbstractionLayerImpl


juanlaube:
## 💥 Crash Report
Cannot reproduce yet and there's not enough information right now, the stacktrace is not helpful.

### 👥 Impact
- 45 times to 32 users, that's 63% of the crashes in Giskard 1.10
- Found after releasing Giskard 1.10 to 5% of active users

### 🛰 Link
https://appcenter.ms/orgs/Toggl/apps/Toggl-Android/crashes/errors/749082492u/overview

### 📝  Exception & Stack Trace
`System.NullReferenceException: Object reference not set to an instance of an object`
```cs
ConcurrencyAbstractionLayerImpl+<>c.<StartThread>b__8_0 (System.Object itemObject)
ThreadHelper.ThreadStart_Context (System.Object state)
ExecutionContext.RunInternal (System.Threading.ExecutionContext executionContext, System.Threading.ContextCallback callback, System.Object state, System.Boolean preserveSyncCtx)
ExecutionContext.Run (System.Threading.ExecutionContext executionContext, System.Threading.ContextCallback callback, System.Object state)
ThreadHelper.ThreadStart (System.Object obj)
```

heytherewill:
As per the logged [events](https://appcenter.ms/orgs/Toggl/apps/Toggl-Android/crashes/errors/749082492u/reports/2518545871749829999-35ada7d7-1074-4b96-9b8d-ed47b6504eff/events), this seems to be a bug in the background sync

amulware:
This occurs in 1.11.1 - I thought background sync was disabled?

amulware:
Also, is this actually being worked on? If not, please unassign.

juanlaube:
Ok, this is still a thing in Giskard 1.11.1, the problem is not exlusive to background sync, but something in sync in general, that's why we still see some of these even with background sync disabled.

The new events track every transition and state change in sync, I hoped to find all the crashes in the same sate but that's not a thing, however something I noticed is the last events recorded are always `PersistListState` or `EnsureFetchListSucceededState`, maybe that's useful for future research.

@amulware I will unassign myself because I'm working on the calendar, but I think this should be prioritized.

kornelijepetak:
[...]

PR #4247, opened 2019-01-17
Call JobFinished after background sync completes

juanlaube:
## What's this?
This is an attempt to fix a crash we found in Giskard 1.10 related to background sync.
After many failed attempts to debug and reproduce this, the only logic conclusion (so far) is Android is killing the app for some unknown reason. There's no call to `JobFinished` to let the OS know the job is complete, so this could explain the problem, another reason could be that the OS requests Giskard to cancel the running job, but we are not doing that.

### Relationships

Refs #4230

## Why do we want this?
The crash is bad and it's affecting our metrics.

## How is it done?
By calling `JobFinished` when background sync completes. Also tracking when the OS requests to cancel the background job (fixing this is a little bit harder, but lets see if this is the problem).

### Side effects
Best case: this solves all problems (lol, I wish).

## Review hints
Try scheduling jobs using `adb shell`. The following commands are useful:
- `cmd jobscheduler run -f packageName jobId` will force scheduling the job
- `dumpsys jobscheduler` will output all info, including the running jobs

If you try this with Giskard 1.10, you should see something like the output below. The active job will run for 10 minutes until it's killed by a timeout. With Giskard 1.10.1 the job will be active only for a few seconds.

```
Pending queue:

Active jobs:
  Slot #0: c1c0176 #u0a239/1 com.toggl.giskard/.BackgroundSyncJobSchedulerService
    Running for: +9m14s541ms, timeout at: +45s853ms
    u0a239 tag=*job*/com.toggl.giskard/.BackgroundSyncJobSchedulerService
    Source: uid=u0a239 user=0 pkg=com.toggl.giskard
    Required constraints: TIMING_DELAY DEADLINE CONNECTIVITY
    Earliest run time: -09:14
    Latest run time: 05:45
```

## :squid: Permissions
🚫 Nope


amulware:
We'll be testing this on users then? If so, let's get it out as soon as possible, since the feedback loop takes a long time already.

PR #4258, opened 2019-01-18
Use BehavioRelay for StartTime

### LINKED ISSUE #4257: Use BehaviorRelay in EditDurationViewModel
In order to follow the new [rx binding guidelines](https://github.com/toggl/mobileapp/blob/develop/docs/rx-binding-guidelines.md) the `StartTime` and `StopTime` properties, should be changed into `BehaviorSubjects`.

For each of those there's an `Observable`, a `BehaviorSubject` and an `InputAction` that all it does is send a new value to the subject. All those should be replaced by a `BehaviorSubject`.

Odrakir:
## What's this?
This tries to use a BehaviorRelay for the `StartTime` in `EditDurationViewMode`, but it fails.

### Relationships
Closes #4257 

## Why do we want this?
To follow rx-binding guidelines.

## How is it done?
I've changed the subject+observable+rxaction into a simple `BehaviorRelay` for the `StartTime` property.

### Side effects
As you can see if you execute the app this doesn't work. First the start time in the wheel is set to the wrong time. And then it behaves weirdly when dragged. For me it crashes if I drag counter-clockwise first, but also it adds much more duration than it's supposed to if I go clockwise first.

## Review hints
You can try keeping the `StartTime` observable to see where the problem is. If I do this:

```c#
StartTime = StartTimeRelay.AsDriver(schedulerProvider);
```

it works, but printing the thread before and after `AsDriver` shows the exact same ID, so I don't know what's going on here...

## :squid: Permissions
Don't!

### DIFF
--- a/Toggl.Daneel/ViewControllers/EditDurationViewController.cs
+++ b/Toggl.Daneel/ViewControllers/EditDurationViewController.cs
@@ -133,7 +133,7 @@ namespace Toggl.Daneel.ViewControllers
 
             var startTime = ViewModel.IsEditingStartTime
                     .Where(CommonFunctions.Identity)
-                    .SelectMany(_ => ViewModel.StartTime);
+                    .SelectMany(_ => ViewModel.StartTimeRelay);
 
             var stopTime = ViewModel.IsEditingStopTime
                     .Where(CommonFunctions.Identity)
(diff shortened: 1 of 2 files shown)


heytherewill:
Any updates on this?

Odrakir:
I didn't have time to look into it. But it's on my list.

heytherewill:
Is this still on your list or can it be closed?

Odrakir:
I'm going to close this for now, it's probably completely out of date.

PR #4266, opened 2019-01-18
✨ Calendar zoom

### LINKED ISSUE #4233: Pinch zooming in Calendar View
Right now I can see about ~10 hours on my iPhone 6S in the Calendar View. That's great for having an overview of the day, but any time entry smaller than ~30m gets a bit finicky, especially if I want to use the drag-to-edit features.

I'd like us to investigate how difficult it would be to allow the user to two-finger pinch to zoom in and out on the calendar.

In terms of constraints, the user could zoom out so far that they only see a single day, and maybe zoom in by a factor of 2 or 3 compared to now.

If this is very hard to do, we probably shouldn't (still, please report below how it could be done, and why it's difficult!).

If it's easy, let's at least give it a try and see how we like it. :)

sofokli1:
## What's this?
Calendar zoom

### Relationships
Closes #4233

## Why do we want this?
We would like the cal to zoom to see the small entries

## How could it be done?
The `CalendarCollectionViewLayout` has this constant called `HourHeight` and most calculations are based on that value. When this value is changed we invalidate some layouts and the magic happens

## Review hints
Take a look at the code and run it (ideally on device).

## :squid: Permissions
Yes


heytherewill:
That's amazing! I'm surprised to see how simple the solution looks 

@juanlaube Can you check this as well?

amulware:
The gif above is convincing enough to me. :D

Toggl.Daneel/ViewControllers/Calendar/CalendarViewController.cs
```diff
@@ -10,6 +10,7 @@
 using Toggl.Foundation;
 using Toggl.Foundation.MvvmCross.Extensions;
 using Toggl.Foundation.MvvmCross.ViewModels.Calendar;
+using Toggl.Multivac;
```
juanlaube: 👀

sofokli1:
calendar zoom people, feel free to close this PR if it has served its purpose

amulware:
Since people seem to like this, should we just polish it up and merge it instead?

sofokli1:
Sure, saves me the effort to look for a new issue 😄

Odrakir:
I might be nitpicking, but I think the scroll offset should change at the same time as the height changes to compensate for the movement in a way that the calendar doesn't slip under the fingers... not sure if I'm explaining my self.

sofokli1:
I totally agree. This PR was meant just as a proof of concept. All will be fixed and polished starting tomorrow. Will ask for re-review.

juanlaube:
I re-reviewed this and looks good to me :shipit:

sofokli1:
Some love from @heytherewill @Odrakir and @amulware is needed to maybe add this to the current 1.17 release

Toggl.Daneel/Views/Calendar/CalendarCollectionViewZoomHelper.cs
```diff
@@ -0,0 +1,57 @@
...
+                    layout.ScaleHourHeight(gesture.Scale, pinchCenter);
+                    gesture.Scale = 1;
+                    break;
+
+                case UIGestureRecognizerState.Ended:
+                    layout.ScaleHourHeight(gesture.Scale, pinchCenter);
+                    break;
+
+                case UIGestureRecognizerState.Cancelled:
+                case UIGestureRecognizerState.Failed:
```
heytherewill: Maybe just use default here?

Toggl.Daneel/Views/Calendar/CalendarCollectionViewLayout.cs
```diff
@@ -33,6 +37,7 @@ public sealed class CalendarCollectionViewLayout : UICollectionViewLayout
         private DateTime date;
         private readonly ITimeService timeService;
         private readonly ICalendarCollectionViewLayoutDataSource dataSource;
+        private readonly ISchedulerProvider schedulerProvider;
```
Odrakir: This is not used anywhere.

Toggl.Daneel/ViewControllers/Calendar/CalendarViewController.cs
```diff
@@ -23,18 +24,21 @@ public sealed partial class CalendarViewController : ReactiveViewController<Cale
...
         private CalendarCollectionViewCreateFromSpanHelper createFromSpanHelper;
+        private CalendarCollectionViewZoomHelper zoomHelper;
 
         private readonly UIButton settingsButton = new UIButton(new CGRect(0, 0, 40, 50));
 
         public CalendarViewController()
             : base(nameof(CalendarViewController))
         {
             timeService = Mvx.Resolve<ITimeService>();
+            schedulerProvider = Mvx.Resolve<ISchedulerProvider>();
```
Odrakir: Not sure if we should still add this kind of `Mvx` code. But anyway, as I mention in another comment the `schedulerProvider` is not used anywhere.

Odrakir:
For some reason I can't get this to work on device... 🤔

sofokli1:
@Odrakir do you mean the branch or the zoom? I tried on device and it looks much better than on the simulator.

Toggl.Daneel/Cells/Calendar/CalendarItemView.cs
```diff
@@ -32,10 +32,12 @@ public sealed partial class CalendarItemView : ReactiveCollectionViewCell<Calend
         public CGRect TopDragTouchArea => TopDragIndicator.Frame.Inset(-20, -20);
         public CGRect BottomDragTouchArea => BottomDragIndicator.Frame.Inset(-20, -20);
 
+        public CalendarCollectionViewLayout Layout;
+
         //In other words: 30 minutes or less
-        private bool shortCalendarItem => Frame.Height <= CalendarCollectionViewLayout.HourHeight / 2;
+        private bool shortCalendarItem => Frame.Height <= Layout.HourHeight / 2;
```
simonrozsival: Shouldn't this logic change as well? It seems strange that even though you zoom in, all of the <30 min entries will be shown as if they were 30 min long. 

My suggestion is to have a property `ShortItemHeight` which would be calculated as `HourHeight / 2` when the view is first displayed and it wouldn't change when the view is scaled (unlike `HourHeight` which changes).

Toggl.Daneel/Views/Calendar/CalendarCollectionViewZoomHelper.cs
simonrozsival: I think it's better to have them here explicitly. The "default" would just hide the corner as if we didn't consider them.

simonrozsival:
I will be on vacation for the next two weeks so please don't wait for my approval. Please consider my requests anyway 😄

PR #4275, opened 2019-01-21
Fix weird close icon

### LINKED ISSUE #4274: The 'x' icon on the date selection view is positioned weirdly
## 🐛 Bug Report
There is a glitch on the date selection view. The close 'x' icon is aligned all the way to the left and the right part of the icon is overlayed by something:

image

## Context

Build/Branch: `develop` and `vega` (1.16)
Phone OS: iOS

## Expected behavior
The close button looks the way it is shown in Zeplin.

simonrozsival:
## What's this?
Fix a glitch in the `StartDateTimeViewController`

### Relationships
Introduced in #4123
Closes #4274

## Why do we want this?
We want our app to look nice and we want it to be localizable correctly.

## How is it done?
The "Save" string was set as text to the incorrect button. It is supposed to be the `SaveButton` and not the `CloseButton`

## Review hints
This is what the problem was:

image

Now it looks just like it is supposed to (I won't even bother with a screenshot, you can verify it by running the simulator :D)

## :squid: Permissions
🦑


simonrozsival:
It seems it's not the only place where this happens. I just noticed:

image

I will go through the app and I will try to find all the places where this happens and try to fix those.

PR #4307, opened 2019-01-28
Track sync and app lifecycle events

juanlaube:
## What's this?
This adds some events to track sync and app lifecycle changes so we can investigate when does the crash from #4230 happens.

### Relationships

Ref #4230

## Why do we want this?
☝️ See What's this

## How is it done?
This tracks:
- When the app goes to background
- When the app returns to foreground
- Changes in the sync queue
- Which sync states are executed

### Why not another way?
🤷‍♂️ 

### Side effects
None

## :squid: Permissions
🚫 🦑


juanlaube:
@simonrozsival I'd like your opinion on this.

Toggl.Foundation/Sync/SyncManager.cs
```diff
@@ -67,6 +67,7 @@ public IObservable<SyncState> PushSync()
             lock (stateLock)
             {
                 queue.QueuePushSync();
+                analyticsService.SyncPushStarted.Track();
```
daividssilverio: 🤔 Is it worth it to put another track before `queue.QueuePushSync()` ?
Maybe `analyticsService.SyncPushWillStart`?

daividssilverio: Same for all the others related to sync

juanlaube: Maybe we could have a single event with a parameter and different values, e.g: PullWillStart, PullStarted, PushWillStart, PushStarted... 🤔
I will consider that.

simonrozsival: The queue itself doesn't do much. The only thing that is interesting is when we dequeue from it. Putting an event into `startSyncIfNeeded` and logging which sync will be next would be interesting.

simonrozsival:
To log which states were executed I have a suggestion (I do this sometimes for debugging purposes) to put the logging into `TransitionHandlerProvider`:
- change `Dictionary<IStateResult, TransitionHandler> transitionHandlers` to `Dictionary<IStateResult, (string stateName, TransitionHandler handler)> transitionHandlers`
- in `ConfigureTransition`(2x) add the state with its name (`state.GetType().GetFriendlyName()`)
- use `GetFriendlyName` extension snippet from https://stackoverflow.com/questions/4185521/c-sharp-get-generic-type-name/26429045:
```cs
public static string GetFriendlyName(this Type type)
{
    if (type.IsGenericType)
    {
        var name = type.Name.Substring(0, type.Name.IndexOf('`'));
        var types = string.Join(",", type.GetGenericArguments().Select(GetFriendlyName));
        return $"{name}<{types}>";
    }
    else
    {
        return type.Name;
    }
}
```
(it should have some tests even though it looks OK)
- in `GetTransitionHandler` log the name of the handler

Toggl.Foundation/Analytics/BaseAnalyticsService.cs
```diff
@@ -218,6 +218,24 @@ protected BaseAnalyticsService()
         [AnalyticsEvent("DelayDurationSeconds")]
         public IAnalyticsEvent<int> RateLimitingDelayDuringSyncing { get; protected set; }
 
+        [AnalyticsEvent("State")]
+        public IAnalyticsEvent<string> SyncStateDequeued { get; protected set; }
```
simonrozsival: This name is confusing. I would call it `SyncOperationStarted`. We should rename `SyncStateQueue` to `SyncOperationQueue` as well (and maybe some other things as well, it's a mess). In a different PR. I will create an issue.

Toggl.Foundation/Sync/SyncManager.cs
```diff
@@ -143,6 +144,8 @@ private void syncOperationCompleted(SyncResult result)
 
         private void processError(Exception error)
         {
+            analyticsService.SyncFailed.Track(error.GetType().FullName, error.Message, error.StackTrace);
```
simonrozsival: Can we track this as a handled exception as well? The stack trace we will get from the event will be unusable.

Toggl.Multivac/Extensions/TypeNameExtensions.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System;
+
+namespace Toggl.Multivac.Extensions
+{
+    public static class TypeNameExtensions
+    {
+        public static string GetFriendlyName(this Type type)
```
simonrozsival: This needs a `[Theory]` test! Suggested inputs:
- `string`
- `ISyncState`
- `ISyncState<string>`
- `ISyncState<string[]>`
- `ISyncState<IObservable<IEnumerable<IThreadSafeTimeEntry>>>`
- `ISyncState<IObservable<long[]>>`

You can add some more ofc :)

Toggl.Foundation.Tests/Sync/SyncManagerTests.cs
```diff
@@ -406,10 +406,10 @@ public void TellsQueueToStartCleaningUp()
             }
 
             [Fact]
-            public void TracksSyncCleanUpStarted()
+            public void TracksSyncStarted()
```
daividssilverio: 🤔 The name shouldn't have changed, right?

juanlaube: I think visual studio derped while I renamed another method. Anyway, I don't like the name of this method, gonna fix it.

Issue #4014, opened 2018-12-17
RxFactor the SelectTimeViewModel


heytherewill:
Check #3352 for details

heytherewill:
I think this too is probably not necessary because of the wheel

Pinging both @amulware and @daividssilverio

heytherewill:
If this indeed is not needed, please use this issue as a reminder to 🔥 this ViewModel and all of its strings/assets/what-not

juanlaube:
I asked @daividssilverio and The Wheel ™️ will actually replace the `SelectTimeViewModel` so there's no need to rxfactor this, however we should delete everything related to that view model and the corresponding fragment after The Wheel is ready.

simonrozsival:
I believe we should remove this from the Daneel release project then, @amulware.

kornelijepetak:
I confirm that with the integration of the Wheel-based viewmodel, this view model and all it's brothers and sisters can be 🔥d.

kornelijepetak:
I have assigned myself so that I will close the issue once this is wiped out.

PR #4308, opened 2019-01-28
Android Edit duration view

### LINKED ISSUE #4112: Create the Android EditDuration View
- Blocked by #3827;
- It should utilize the EditDurationViewModel;
- It must include the android wheel, result of #4111;
- It must have fields to edit the `start` and `stop` datetimes (they will more likely open dialogs/popups, instead of expanding inside the view like it was done in iOS)
- If the time entry is new/running, tapping `stop` field should stop the time entry;
- If the time entry is new/running, the wheel should stay in sync with the duration in progress.

Child of #4030

daividssilverio:
## What's this?
This is the android version of the iOS Edit duration view with adaptations due to platform differences.

### Relationships
Depends on #4147 and #4146
Closes #4112

## Why do we want this?
To have better UX when editing durations on Android.

## How is it done?
The view has its own Activity for now and it will show dialogs to pick start/stop times.
When #4147, it will have the Wheel as well <3

### Why not another way?
This is the other way. Have another idea? Let me know.

### Side effects
Users now navigate to another screen to edit start/stop/duration.

## Review hints
Please start reviewing as soon as possible.
Please understand that the wheel still has to be plugged and bound to the Activity.

## :squid: Permissions
Don't 🦑 it yet, wait until #4147 is merged.


daividssilverio:
🤔 SelectTimeFragment is not part of the `EditDurationViewModel` or Edit Duration view. It's part of the the old flow to edit start/end/duration on android.

Toggl.Foundation.MvvmCross/ViewModels/SelectTimeViewModel.cs
```diff
@@ -12,26 +12,19 @@
...
 using Toggl.Foundation.Interactors;
-using static Toggl.Foundation.MvvmCross.ViewModels.SelectTimeViewModel.TemporalInconsistency;
+using Toggl.Foundation.MvvmCross.Helper;
+using static Toggl.Foundation.MvvmCross.Helper.TemporalInconsistency;
 
 namespace Toggl.Foundation.MvvmCross.ViewModels
 {
     [Preserve(AllMembers = true)]
-    public sealed class SelectTimeViewModel
+    public sealed partial class SelectTimeViewModel
```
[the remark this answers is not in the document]
daividssilverio: I used Rider's extract method and it made the class Partial and the `TemporalInconsistency` was nested inside the other part of the class. Forgot to undo this change when I moved the `TemporalInconsistency` to its own class.

Toggl.Giskard/Resources/layout/EditDurationActivity.axml
```diff
@@ -0,0 +1,224 @@
...
+            local:layout_constraintRight_toLeftOf="@id/CenterGuideline"
+            local:layout_constraintLeft_toRightOf="@id/StartDotSeparator"
+            android:layout_marginTop="8dp"
+            android:layout_marginRight="8dp"
+            android:layout_marginBottom="16dp"
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content" />
+        <ImageView
+            android:id="@+id/StopIcon"
+            android:src="@drawable/ic_play"
```
juanlaube: This is the wrong asset, should be a stop icon instead of play

daividssilverio: Nice catches.

Toggl.Foundation.MvvmCross/ViewModels/EditDurationViewModel.cs
```diff
@@ -29,6 +30,7 @@ public sealed class EditDurationViewModel : MvxViewModel<EditDurationParameters,
         private DurationParameter defaultResult;
         private EditDurationEvent analyticsEvent;
 
+        private ISubject<TemporalInconsistency> temporalInconsistencies = new Subject<TemporalInconsistency>();
```
heytherewill: Are we still showing these toasts? If so, do we really need them?

Toggl.Foundation.MvvmCross/ViewModels/SelectTimeViewModel.cs
```diff
@@ -12,7 +12,8 @@
 using System.Reactive.Subjects;
 using Toggl.Foundation.DataSources;
 using Toggl.Foundation.Interactors;
-using static Toggl.Foundation.MvvmCross.ViewModels.SelectTimeViewModel.TemporalInconsistency;
+using Toggl.Foundation.MvvmCross.Helper;
```
heytherewill: I think this ViewModel (and its views) can be entirely 🔥 . Ping @kornelijepetak just in case

Toggl.Foundation.MvvmCross/ViewModels/EditDurationViewModel.cs
daividssilverio: Yeah, we can't just close and open the time/date dialogs when the user sets an invalid start/end time.
The android time picker doesn't let us set min start/end times.
The android date picker lets us do that and we are doing it already 👍. But we still need it for edge cases.

Toggl.Foundation.MvvmCross/ViewModels/SelectTimeViewModel.cs
juanlaube: This can be deleted can be done as part of this PR or to close #4014 once and for all 🤷‍♂️

kornelijepetak: This view model and the views can be deleted only after the wheel is completely done.

daividssilverio: Are you ok with leaving the SelectTimeViewModel and related files so we can proceed with the release? It looks safe to 🔥 all the `SelectTime*` stuff, unless @kornelijepetak has some plans for it.

Issue #4218, opened 2019-01-16
Implement infrastructure for grouped Time Entries


kornelijepetak:
#### Background 
Currently, we are using the following combination of view models:

* `MainViewModel` - governs the main screen
* `TimeEntriesViewModel` - governs the list of time entries on the main log.

`TimeEntriesViewModel.TimeEntries` is an `IObservable<TimeEntryViewModel>` that provides the list of time entries. 

So far, it seems to me that the easiest solution is to inherit a new class (e.g. `GroupedTimeEntryViewModel`) from the `TimeEntryViewModel` because the grouped time entry is basically a time entry template and has just a bit more information than the `TimeEntryViewModel` (and the observable should be of a single common type). Interface could be used as well, but it would basically be a copy of the `TimeEntryViewModel`. 

💡 Of course, this is just an idea. If the person solving this thinks there is a better way to solve this, this is in no way the only allowed solution. Use best judgement. Note however, that inheriting from the `TimeEntryViewModel` will make the Edit TE process much easier.

#### Requirements

🏆 This issue will be solved if:

* [x] The `TimeEntriesViewModel` exposes a collection observable that returns the items for the main log.
	* [x] The items are groupped by the matching time entry data
	* [x] The groups are sorted in a regular way (older groups further down the list)
	* [x] The groups have a way to be collapsed/expanded (which changes the collection)
* [x] The view model provides a way to navigate to the EditTimeEntryViewModel for a single item (as is currently already implemented), as well as for the group
	* [x] In case of a single TE, send TE's ID as a parameter
	* [x] In case of a group, send ID of any TE in that group (for simplicity: first)
* [x] The TEs are not grouped if user did not request that feature from settings screen
* [x] All the relevant tests for this functionality should be written.

simonrozsival:
This issue was fully implemented with a series of PRs which were merged into a release branch. The only thing that's left is sending correct parameters to the edit VM, but this part is still being developed, so it will be connected later, hopefully as part of #4219.

I'm closing this issue now manually.

PR #4309, opened 2019-01-29
Delete dead code from table view sources and recyclers

simonrozsival:
## What's this?
While I was investigating the view sources I ran into an interesting (ugly) part of code: `NestableObservableCollection`. Rider showed most of the code in gray as unused. So I started deleting code from it and all the code that handled the child collection changes. I ended up with an empty class and so I removed it as a whole.

We had these 200 lines of non-trivial code, which is useless. Let's get rid of it so we don't have to drag it with us as we refactor the app.

### Relationships

Ref #4218 

## Why do we want this?
No dead code!

## How is it done?
🔥 🔥 🔥 

### Side effects
Even though it looks scary, neither iOS nor the Android app breaks.

## Review hints
Try it yourself.

## :squid: Permissions
🦑 asap


kornelijepetak:
I somehow have a feeling that the Nestable one wasn't created without some reason, and that we lost the knowledge of the reason along the way... So I'm not sure if that's the wisest decision. I think a person that introduced this collection should check this. 🤔

simonrozsival:
We don't use it and it is extremely confusing (I mean the code which handles it). If we need it in the future, we can bring it back. But we won't. I think our tree data structure would replace it anyway.

Odrakir:
I've created `GroupedOrderedCollection` to replace `NestableObservableCollection` which is more complicated than it needs to be and depends on Mvx. I think that should be a start.

PR #4315, opened 2019-01-29
Wheel Duration Input for Android

### LINKED ISSUE #4146: Create the numeric input wheel component

kornelijepetak:
## What's this?
Implements duration input editor for use with Android Wheel.

### Relationships
Closes #4146

## Why do we want this?
🎡All the way.

## How is it done?
* `WheelDurationInput` extends `EditText` and handles all the required logic.

**Important:**
*  `android:inputType` must be `numberPassword` to provide correct keyboard.
* `local:fadedTextColor`attribute value has to be provided to mark the less relevant part of the duration during edit, such as `00:0` in `00:02`. (check designs)
* `android:textColor` attribute has to specify the default color (`fadedTextColor` applies only to the less relevant part of the duration, everything else uses the default `textColor`.

### Why not another way?
A lot of back and forth and various attempts have been done with this, and this seems to be the most reasonable way, considering some of the Android quirks.

### Side effects
There is one issue that appears (which is purely aesthetic), and for me it happens only on some platforms. When user taps in the middle of the description, the selection pin appears in the middle of the text edit, however, the real caret is at the end where it should be and all input is processed correctly. It's just that sometimes, the pin will be there until next change. I was unable to remove this kind of behaviour (at least on Android Pie on real device).

## Review hints
First functional. The rest is up to you.

## :squid: Permissions
Only @daividssilverio and @kornelijepetak


Toggl.Giskard/Views/EditDuration/WheelDurationInput.cs
```diff
@@ -0,0 +1,236 @@
...
+            editable.ClearSpans();
+
+            if (isEditing)
+            {
+                editable.SetSpan(new StyleSpan(TypefaceStyle.Bold), 0, editable.Length(), SpanTypes.InclusiveInclusive);
+                editable.SetSpan(new ForegroundColorSpan(fadedTextColor), 0, splitPoint, SpanTypes.InclusiveInclusive);
+            }
+            else
+            {
+                editable.SetSpan(new StyleSpan(TypefaceStyle.Bold), 0, splitPoint, SpanTypes.InclusiveInclusive);
```
daividssilverio: This should be using typefaces instead of setting the text to bold. In a duration string "0:15:00":
"0:15" -> `TypefaceSpan("sans-serif-medium")`
"00" -> `TypefaceSpan("sans-serif-light")`

Toggl.Giskard/Views/EditDuration/WheelDurationInput.cs
```diff
@@ -0,0 +1,236 @@
...
+
+        private void applyFormatting(string text, IEditable editable)
+        {
+            var splitPoint = getFormattingSplitPoint(text);
+
+            editable.ClearSpans();
+
+            if (isEditing)
+            {
+                editable.SetSpan(new StyleSpan(TypefaceStyle.Bold), 0, editable.Length(), SpanTypes.InclusiveInclusive);
```
daividssilverio: Same as here, it should be using `TypefaceSpan("sans-serif-medium")` instead of the `StyleSpan`

PR #4318, opened 2019-01-30
Simplify table view data sources for iOS

Odrakir:
## What's this?
This is an attempt to begin to simplify table view sources for iOS. 
- First I'm extracting the cell configuration code outside of the generic `ListTableViewSource`. This way we can use those for different types of cells easily.
- Then I do the same to `SectionedListTableViewSource` and then I go a delete `ListTableViewSource` as only the sectioned one is needed.
- Then I add some cell and header configuration delegates and some extensions to bind the table views to those sources.
- Finally I change the tables that were using those datasources. Still there are some table views that need changing though.

(As a bonus I also refactored `ListCollectionViewSource`)

### Relationships
Doesn't close any issue. But this came up while rxfactoring `StartTimeEntryViewModel` and it'll be important for the conversation about fixing table views and recycler views sources once and for all.

## Why do we want this?
Makes thing simpler when we want to use different types of cells in one table view.

For example, in `ClientTableViewSource` we see that it inherits from `ListTableViewSource` with a cell of type `ClientViewCell`. But then, when actually creating the cell it could either be that or `CreateClientViewCell`, which is confusing.

By extracting the cell configuration code to the VC (or a TV source subclass) we simplify this and make it easier to understand.

This is the way to bind a table view to a source when the cells are `BaseTableViewCell` subclasses:

```c#
var source = new SectionedListTableViewSource<SelectableCountryViewModel>(
    CountryViewCell.CellConfiguration(CountryViewCell.Identifier)
);
CountriesTableView.Source = source;

ViewModel.Countries
    .Subscribe(CountriesTableView.Rx().Items(source))
    .DisposedBy(DisposeBag);
```

And you can even get a cell selection observable:

```c#
source.Rx().ModelSelected()
    .Subscribe(ViewModel.SelectCountry.Inputs)
    .DisposedBy(DisposeBag);
```

There are more ways of doing this binding, but a doc (which everyone should read before doing anything with table views) specifying this is in the works.

### Side effects
None that I'm aware of, but further testing is appreciated.

## Review hints
Follow each commit.

## :squid: Permissions
Go!


Toggl.Daneel/ViewSources/Generic/CollectionView/ListCollectionViewSource.cs
```diff
@@ -1,32 +1,25 @@
...
 {
-    public class ListCollectionViewSource<TModel, TCell> : UICollectionViewSource
-        where TCell : BaseCollectionViewCell<TModel>
+    public class ListCollectionViewSource<TModel> : UICollectionViewSource
     {
-        private readonly string cellIdentifier;
         internal IImmutableList<TModel> items;
 
         public EventHandler<TModel> OnItemTapped { get; set; }
+        public Func<ListCollectionViewSource<TModel>, UICollectionView, NSIndexPath, TModel, UICollectionViewCell> ConfigureCell;
```
kornelijepetak: Please extract this complex delegate into a named one, along the lines of `CellConfigurator`:

```cs
public delegate UICollectionViewCell CellConfigurator<TModel>(
	ListCollectionViewSource<TModel> collection,
	NSIndexPath indexPath, 
	TModel model);
```

Also, if it's public, make it into a property:
```cs
public CellConfigurator<TModel> ConfigureCell { get; protected set; }
```

Toggl.Daneel/ViewSources/Generic/TableView/ListTableViewSource.cs
```diff
@@ -1,31 +1,26 @@
...
 {
-    public class ListTableViewSource<TModel, TCell> : UITableViewSource
-        where TCell : BaseTableViewCell<TModel>
+    public class ListTableViewSource<TModel> : UITableViewSource
     {
         protected readonly string cellIdentifier;
         protected IImmutableList<TModel> items;
 
         public EventHandler<TModel> OnItemTapped { get; set; }
+        public Func<ListTableViewSource<TModel>, UITableView, NSIndexPath, TModel, UITableViewCell> ConfigureCell;
```
kornelijepetak: Another named delegate.

Toggl.Daneel/ViewControllers/SyncFailuresViewController.cs
```diff
@@ -17,10 +17,14 @@ public override void ViewDidLoad()
 
             TableView.RowHeight = UITableView.AutomaticDimension;
             TableView.RegisterNibForCellReuse(SyncFailureCell.Nib, SyncFailureCell.Identifier);
-            TableView.Source = new ListTableViewSource<SyncFailureItem, SyncFailureCell>(
-                ViewModel.SyncFailures,
-                SyncFailureCell.Identifier
-            );
+            var tableViewSource = new ListTableViewSource<SyncFailureItem>(ViewModel.SyncFailures);
+            tableViewSource.ConfigureCell = (source, tableView, indexPath, model) =>
```
kornelijepetak: For reasons of preventing layers of abstractions when reading, please extract this cell configuration into a separate function.

Toggl.Daneel/ViewControllers/SelectDefaultWorkspaceViewController.cs
```diff
@@ -52,6 +49,14 @@ private void onWorkspaceTapped(object sender, SelectableWorkspaceViewModel works
             ViewModel.SelectWorkspace.Execute(workspace);
         }
 
+        private UITableViewCell configureCell(ListTableViewSource<SelectableWorkspaceViewModel> source,
+            UITableView tableView, NSIndexPath indexPath, SelectableWorkspaceViewModel model)
+        {
+            var cell = tableView.DequeueReusableCell(SelectDefaultWorkspaceTableViewCell.Identifier, indexPath) as SelectDefaultWorkspaceTableViewCell;
```
simonrozsival: Let's use explict casts - we don't handle the `null` case anyway on the next line.
```cs
(SelectDefaultWorkspaceTableViewCell)tableView.DequeueReusableCell(SelectDefaultWorkspaceTableViewCell.Identifier, indexPath);
```

Toggl.Daneel/ViewSources/Generic/TableView/ListTableViewSource.cs
simonrozsival: I think we should pass this in the constructor and it should be `private readonly`.

Toggl.Daneel/Cells/BaseTableViewCell.cs
```diff
@@ -26,5 +28,15 @@ protected BaseTableViewCell(IntPtr handle)
         }
 
         protected abstract void UpdateView();
+
+        public static SectionedListTableViewSource<TModel>.CellConfiguration CellConfiguration(string cellIdentifier)
```
simonrozsival: I think this should be in `SectionedListTableViewSource<TModel>`, but I'm not sure 🤔 It kind of doesn't seem right. Mainly when it's called like `SelectDefaultWorkspaceTableViewCell.CellConfiguration(SelectDefaultWorkspaceTableViewCell.Identifier)` it seems wrong.

Toggl.Daneel/ViewSources/Generic/TableView/SectionedListTableViewSource.cs
```diff
@@ -0,0 +1,60 @@
...
+{
+    public class SectionedListTableViewSource<TModel> : UITableViewSource
+    {
+        public delegate UITableViewCell CellConfiguration(SectionedListTableViewSource<TModel> source, UITableView tableView, NSIndexPath indexPath, TModel model);
+        public delegate UIView HeaderConfiguration(SectionedListTableViewSource<TModel> source, UITableView tableView, int section);
+
+        protected IImmutableList<IImmutableList<TModel>> sections;
+        private CellConfiguration configureCell;
+
+        public EventHandler<TModel> OnItemTapped { get; set; }
```
simonrozsival: Why isn't this an observable? Do we still need that somewhere?

Toggl.Daneel/ViewSources/Generic/TableView/SectionedListTableViewSource.cs
```diff
@@ -0,0 +1,60 @@
...
+
+namespace Toggl.Daneel.ViewSources
+{
+    public class SectionedListTableViewSource<TModel> : UITableViewSource
+    {
+        public delegate UITableViewCell CellConfiguration(SectionedListTableViewSource<TModel> source, UITableView tableView, NSIndexPath indexPath, TModel model);
+        public delegate UIView HeaderConfiguration(SectionedListTableViewSource<TModel> source, UITableView tableView, int section);
+
+        protected IImmutableList<IImmutableList<TModel>> sections;
+        private CellConfiguration configureCell;
```
simonrozsival: This should be `readonly`.

Toggl.Daneel/ViewSources/Generic/TableView/SectionedListTableViewSource.cs
```diff
@@ -0,0 +1,60 @@
...
+using UIKit;
+
+namespace Toggl.Daneel.ViewSources
+{
+    public class SectionedListTableViewSource<TModel> : UITableViewSource
+    {
+        public delegate UITableViewCell CellConfiguration(SectionedListTableViewSource<TModel> source, UITableView tableView, NSIndexPath indexPath, TModel model);
+        public delegate UIView HeaderConfiguration(SectionedListTableViewSource<TModel> source, UITableView tableView, int section);
+
+        protected IImmutableList<IImmutableList<TModel>> sections;
```
simonrozsival: If it is protected, then it should be a property and start with capital `S`.

Toggl.Daneel/ViewSources/Generic/TableView/SectionedListTableViewSource.cs
```diff
@@ -0,0 +1,60 @@
...
+        private CellConfiguration configureCell;
+
+        public EventHandler<TModel> OnItemTapped { get; set; }
+
+        public HeaderConfiguration ConfigureHeader { get; set; }
```
simonrozsival: Why can't this be a `readonly` field passed through and set in the constructor as well as the `configureCell`?

Toggl.Daneel/ViewSources/ColorSelectionCollectionViewSource.cs
```diff
@@ -11,17 +11,17 @@
...
+    public sealed class ColorSelectionCollectionViewSource : ListCollectionViewSource<SelectableColorViewModel>, IUICollectionViewDelegateFlowLayout
     {
         public IObservable<MvxColor> ColorSelected
             => Observable
                 .FromEventPattern<SelectableColorViewModel>(e => OnItemTapped += e, e => OnItemTapped -= e)
                 .Select(e => e.EventArgs.Color);
 
         public ColorSelectionCollectionViewSource(IObservable<IEnumerable<SelectableColorViewModel>> colors)
-            : base (new List<SelectableColorViewModel>().ToImmutableList(), ColorSelectionViewCell.Identifier)
+            : base (new List<SelectableColorViewModel>().ToImmutableList())
```
simonrozsival: Use `ImmutableList<SelectableColorViewModel>.Empty`

Toggl.Daneel/ViewSources/Generic/CollectionView/ListCollectionViewSource.cs
simonrozsival: I don't think it should be public or even protected. It should be set in the constructor same way it is done in the table view source variant.

Toggl.Daneel/Cells/BaseTableViewCell.cs
Odrakir: If interfaces could have static members the call could just be: `MyTableViewCell.CellConfiguration`... That's why it was there, but you are right, it might make more sense in the DS, at least until c#9 comes.

Toggl.Daneel/ViewSources/Generic/TableView/SectionedListTableViewSource.cs
Odrakir: I find it more complicated to create observables directly from here (I need a subject for that). So I keep this and create the Observable in an Rx extension: Toggl.Daneel/Extensions/Reactive/SectionedListTableViewSourceExtensions.cs

Toggl.Daneel/Cells/BaseTableViewCell.cs
Odrakir: Mmmmm... I'm not finding a better option for this. Any particular ideas? Is the repetition of the cell class what's bothering you?

Toggl.Daneel/ViewSources/Generic/TableView/SectionedListTableViewSource.cs
Odrakir: Basically because it's not needed most of the time and adding it to the constructor makes me either have a bunch of different constructor overrides or have some weird construction syntax on call site.

kornelijepetak: Not saying you should change it, but you can always set a default parameter to avoid multiple constructors.

Toggl.Daneel/ViewSources/Generic/TableView/BaseTableViewSource.cs
```diff
@@ -0,0 +1,44 @@
+using System;
+using System.Collections.Immutable;
+using Foundation;
+using UIKit;
+
+namespace Toggl.Daneel.ViewSources
+{
+    public class BaseTableViewSource<TModel> : UITableViewSource
```
simonrozsival: This should be `abstract`

PR #4320, opened 2019-01-30
Add Group TEs setting into the settings screen

### LINKED ISSUE #4217: Add a user settings for grouped Time Entries
🏆 User should be able to decide whether they want their time entries to be grouped or not from the Settings screen. Make sure that we save the value received from `me/preferences` endpoint and use it in the viewmodel.

kornelijepetak:
## What's this?
Adds `group similar time entries` to Settings screen.

### Relationships
Closes #4217 

## Why do we want this?
Users want customization. Also, 🎡wheels.

## How is it done?
The usuals.

### Why not another way?
Read previous point. Consistency.

### Side effects
None noticed.
I have zero ideas why xcode messed up the `.designer.cs` file. I believe it's fine.

## Review hints
Commit by commit makes most sense:
* Foundation
* iOS
* Android
* Test

## :squid: Permissions
Me and this guy @simonrozsival.

### DIFF
--- a/Toggl.Daneel/ViewControllers/Settings/SettingsViewController.cs
+++ b/Toggl.Daneel/ViewControllers/Settings/SettingsViewController.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Reactive;
 using System.Reactive.Linq;
 using MvvmCross.Platforms.Ios.Presenters.Attributes;
 using MvvmCross.Plugin.Color.Platforms.Ios;
@@ -131,8 +132,8 @@ namespace Toggl.Daneel.ViewControllers
                 .Subscribe(ViewModel.ToggleManualMode)
                 .DisposedBy(DisposeBag);
 
-            GroupSimilarTimeEntriesSwitch.Rx().Changed()
-                .Subscribe(ViewModel.ToggleTimeEntriesGrouping.Inputs)
+            GroupSimilarTimeEntriesSwitch.Rx()
+                .BindAction(ViewModel.ToggleTimeEntriesGrouping)
                 .DisposedBy(DisposeBag);
 
             BeginningOfWeekView.Rx()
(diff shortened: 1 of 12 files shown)


Toggl.Foundation.MvvmCross/ViewModels/Settings/SettingsViewModel.cs
```diff
@@ -191,6 +193,11 @@ public sealed class SettingsViewModel : MvxViewModel
                     .Select(preferences => preferences.TimeOfDayFormat.IsTwentyFourHoursFormat)
                     .DistinctUntilChanged();
 
+            IsGroupingTimeEntries =
+                dataSource.Preferences.Current
+                    .Select(preferences => preferences.CollapseTimeEntries)
+                    .DistinctUntilChanged();
```
simonrozsival: This should be a driver. I think that might explain why you were having some problems with syncing. All the others should be drivers as well.

Toggl.Daneel/ViewControllers/Settings/SettingsViewController.cs
```diff
@@ -130,6 +131,10 @@ public override void ViewDidLoad()
                 .Subscribe(ViewModel.ToggleManualMode)
                 .DisposedBy(DisposeBag);
 
+            GroupSimilarTimeEntriesSwitch.Rx().Changed()
+                .Subscribe(ViewModel.ToggleTimeEntriesGrouping.Inputs)
```
simonrozsival: `BindAction` can't be used here?

Toggl.Foundation.MvvmCross/ViewModels/Settings/SettingsViewModel.cs
kornelijepetak: I turned this into a driver. I would also rather not change the rest in this PR as it is unrelated. Also, coming with RxFactoring I think.

Toggl.Daneel/ViewControllers/Settings/SettingsViewController.cs
kornelijepetak: I introduced a `BindAction` on `UISwitch` so yes, it can 😄

PR #4322, opened 2019-01-31
Fix duration input

kornelijepetak:
Related to the 🎡.

# What's this
Fixes two things:

* `Duration` changed to property
* Made UI style adhere to designs

### Relationships
Basically, an update for #4315 

## Why do we want this?
Respecting designs and better code.

### Side effects
Nothing. 

## :squid: Permissions
@daividssilverio and me.

## Additional notes
⚠️ @daividssilverio Make sure that the xml for wherever you put this thing has:
`fontFamily="sans-serif-light"` as well as both colors properly set (`textColor` and `fadedTextColor`).


simonrozsival:
> ⚠️ @daividssilverio Make sure that the xml for wherever you put this thing has:
fontFamily="sans-serif-light" as well as both colors properly set (textColor and fadedTextColor).

Should we document this somehow? Or how would anybody know to do that, if they had to change it or use it somewhere else in the future?

kornelijepetak:
@simonrozsival It will work without that. The reason why I added this comment is just related to that specific usage in the wheel so that it respects the design on Zeplin. Without those, it would be just styled a bit differently.

But I can comment it somewhere if you think it's still needed.

PR #4333, opened 2019-02-04
🚧 Optimize the clean-up loop for cleaning inaccessible entities

### LINKED ISSUE #3675: Optimize the clean-up loop for cleaning inaccessible entities
Add a simple state which checks, if there are any inaccessible _workspaces_ in the database and have two results which should then be connected in the transitions graph like this:
- if yes, go through the states which clean inaccessible TEs, projects, clients, ...
- if no, skip them and, go to the next part of the clean-up

sofokli1:
## What's this?
Adds a new sync state in cleanup to check for inaccessible workspaces before to decide on the cleanup of inaccessible data.

### Relationships
Closes #3675

## Why do we want this?
Why try to query and delete data that do no need deleting.

## How is it done?
By adding a new clean up state.

### Why not another way?
can't think of any

### Side effects
none

## Review hints
Let me know how the code looks so that i can add some tests if ti looks good.

## Tests
Integration tests: https://app.bitrise.io/build/20cfeed69caf8d3c

Syncing tests: https://app.bitrise.io/build/904cbc8c5fa24e62

## :squid: Permissions
only by @toggl/sync-guardians


simonrozsival:
@sofokli1 please show a screenshot of the updated graph + add link to sync tests in Bitrise

sofokli1:
unknown

Toggl.Foundation/Sync/States/CleanUp/CheckForInaccessibleWorkspaces.cs
```diff
@@ -0,0 +1,39 @@
...
+            Ensure.Argument.IsNotNull(dataSource, nameof(dataSource));
+
+            this.dataSource = dataSource;
+        }
+
+        public IObservable<ITransition> Start()
+            => dataSource
+                .Workspaces
+                .GetAll(ws => ws.IsInaccessible, includeInaccessibleEntities: true)
+                .Select(data => data.Count())
```
amulware: Simplify as `workspaces => workspaces.Any()`? (with an appropriate change to the method below).

amulware:
Also, no tests?

sofokli1:
test links in the PR description

sofokli1:
@amulware and @simonrozsival some love here when possible would be nice.

Toggl.Foundation.Tests/Sync/States/CleanUp/CheckForInaccessibleWorkspacesTests.cs
```diff
@@ -0,0 +1,74 @@
...
+
+        [Theory, LogIfTooSlow]
+        [InlineData(0, SyncStatus.InSync)]
+        [InlineData(1, SyncStatus.InSync)]
+        [InlineData(3, SyncStatus.InSync)]
+        [InlineData(1, SyncStatus.SyncFailed)]
+        [InlineData(3, SyncStatus.SyncFailed)]
+        public async Task ReturnsTheCorrectStateBasedOnTheNumberOfInaccessibleWorkspaces(int numberOfInaccessibleWorkspaces, SyncStatus syncStatus)
+        {
+            List<MockWorkspace> workspaces = new List<MockWorkspace>();
```
simonrozsival: This doesn't work. You need to remove the `;` at the end of the line. The workspaces defined in the following block won't be added to the list. Also use `var` 😄 
```suggestion
            var workspaces = new List<MockWorkspace>
```

Toggl.Foundation/Sync/States/CleanUp/CheckForInaccessibleWorkspacesState.cs
```diff
@@ -0,0 +1,39 @@
...
+            Ensure.Argument.IsNotNull(dataSource, nameof(dataSource));
+
+            this.dataSource = dataSource;
+        }
+
+        public IObservable<ITransition> Start()
+            => dataSource
+                .Workspaces
+                .GetAll(ws => ws.IsInaccessible, includeInaccessibleEntities: true)
+                .Select(data => data.Count())
```
simonrozsival: `data.Any()` would be more readable and it might be even a tiny bit faster (but that's not important).

PR #4334, opened 2019-02-04
Creates docs on table view sources

Odrakir:
## What's this?
This adds some docs on table view sources so we have a reference on how this should be done from now on.

It also adds a small change to the code moving the requirement of immutability to the last moment so API usage is cleaner.

### Relationships
Doesn't close any issue.

## Why do we want this?
So everyone is aware of these and we have a future reference.

## :squid: Permissions
Yes you can.


dshr:
Can we have the immutability changes in a separate PR though?

Odrakir:
I mean, yes, it could (and probably should), but this is a small PR and I found that while writing the docs. Please, let it pass this time and I promise I'll never do it again.

dshr:
> I mean, yes, it could (and probably should), but this is a small PR and I found that while writing the docs. Please, let it pass this time and I promise I'll never do it again.

Okay, no probs then! ✨

docs/tableview-sources.md
```diff
@@ -0,0 +1,113 @@
+# iOS Table view sources
+
+This document will show how to create UITableView sources in our app so we do this in a standard way for all our table views.
+
+There are basically two types of table views in the application right now: sectioned and plain. The former have multiple sections with headers and the latter doesn't, but we treat both of those equally, taken plain table views as table views with just one section.
+
+We use the class `CollectionSection<THeader, TItem>` to create sections in our collections for table and recycler views. That means that those collections should be:
```
kornelijepetak: Maybe no need to mention recycler views since this is an article about iOS.

```suggestion
We use the class `CollectionSection<THeader, TItem>` to create sections in our collections for table views. That means that those collections should be:
```

kornelijepetak: I will probably update this docs (or create another one) when I see a bit about recyclers.

docs/tableview-sources.md
```diff
@@ -0,0 +1,113 @@
...
+
+`IEnumerable<TItem>` for plain table views or
+`IEnumerable<CollectionSection<THeader, TItem>>` for sectioned table views
+
+Then table views can be also classified as those which reload completely every time there's a change in the data and those that animate every change (or group of changes) without reloading the whole thing (`ReloadTableViewSource` and `AnimatedTableViewSource` respectively).
+
+## ReloadTableViewSource
+
+This is the class you'll have to use to create a source for your table view. You can either subclass it or use it directly. It inherits from `BaseTableViewSource`, which has some stuff that'll be common to `AnimatedTableViewSource` too.
```
kornelijepetak: `which has some stuff` sounds very vague and does not provide any informational value.

docs/tableview-sources.md
```diff
@@ -0,0 +1,113 @@
...
+`cellConfigurationDelegate` is a method that dequeues the cell, configures it and return it. Its type:
+`public delegate UITableViewCell CellConfiguration<TModel>(UITableView tableView, NSIndexPath indexPath, TModel model)`
+
+As most cells in the application behave in a very similar way (they configure themselves when their view model is set), we can make them inherit from `BaseTableViewCell` which is generic over the cell's view model and updates itself whenever that view model changes).
+
+It also adds a static method that dequeues and configures the cell automatically, so, as long as the cell is a subclass of `BaseTableViewCell`, the above example would be:
+
+```c#
+var source = new ReloadTableViewSource<SectionHeaderViewModel, MyCellViewModel>(
+    MyCustomViewCell.CellConfiguration(MyCustomViewCell.Identifier),
```
kornelijepetak: I'd rather use `BaseTableViewCell.CellConfiguration` to imply that it is the default cell creation process in play here, and not a custom one, related to the specific `MyCustomViewCell`. It's not critical, but I think it would convey what's going on a bit better. Opinions?

docs/tableview-sources.md
```diff
@@ -0,0 +1,113 @@
...
+);
+
+ViewModel.ItemsObservable
+    .Subscribe(MyTableView.Rx().Items(source))
+    .DisposedBy(DisposeBag);
+
+MyTableView.Source = source;
+```
+
+`ViewModel.ItemsObservable` has to be of type `IObservable<IEnumerable<MyCellViewModel>>` for plain table views or `IObservable<IEnumerable<CollectionSection<MyHeaderViewModel, MyCellViewModel>>>` for sectioned table views.
```
kornelijepetak: I think the bullet point formatting would make things more readable.

```suggestion
`ViewModel.ItemsObservable` has to be of type:

 * `IObservable<IEnumerable<MyCellViewModel>>` for plain table views or 
 * `IObservable<IEnumerable<CollectionSection<MyHeaderViewModel, MyCellViewModel>>>` for sectioned table views.
```

docs/tableview-sources.md
```diff
@@ -0,0 +1,113 @@
...
+[Real world example](https://github.com/toggl/mobileapp/blob/ae874b5aa87eac33a9613f6b2f389d0e97d504ac/Toggl.Daneel/ViewControllers/SelectClientViewController.cs#L41)
+
+### Events
+
+The `BaseTableViewSource` also forwards some table view events, which are transformed into Observables through reactive extensions.
+
+So we can do:
+
+```c#
+source.Rx().ModelSelected() // Sends the cell view model
```
kornelijepetak: Now that I'm reading the docs, `ModelSelected` sounds very strange...
I'd go with `SelectedItem` here. This is consistent with .NET naming in many classes in corelib and also with most of our code.

docs/tableview-sources.md
Odrakir: `CollectionSection` is part of `Foundation` and used for both platforms, I think a little background on that doesn't hurt.

docs/tableview-sources.md
Odrakir: I think it's better this way, you can always use the `new` keyword to override the static method with a custom one for the derived class.

docs/tableview-sources.md
Odrakir: I'm leaving `ItemSelected` for when you need the index path of the item that's been selected and not the view model. That is not implemented yet because we make no use of it for the moment, but we might need it in the future. 

Also the naming helps differentiate this from UIKit's `ItemSelected` for collection views or `RowSelected` for table views, which do send the index path or row index only.

Odrakir:
`UITableView` doesn't have any predefined way of dealing with sections, so no, I'm not adding anything here for now. We have to think about it before jumping to conclusions, but I think those depend more on a case to case basis.

docs/tableview-sources.md
Odrakir: I see... I'd be ok with it either way.

PR #4335, opened 2019-02-04
Refactor most table views to use the new sources 

Odrakir:
## What's this?
This unifies most table views to use the new way of creating sources described [here](https://github.com/toggl/mobileapp/blob/develop/docs/tableview-sources.md) 

### Relationships
Doesn't close any issue.

## Why do we want this?
Consistency.

## How is it done?
I searched for UITableView and refactored where I could, one by one, you can follow the commits.

All that's left are: `SelectProject`, `SelectTags` and `StartTimeEntry`, which are going through rxfactors right now, so it'll be done in those PRs. And `Main`, which is more complex and, in principle, needs to be aware of changes in the collection so it can animate (although we might try to see how it works without it)

Also, I need help with `SelectBeginningOfWeekViewModel` from some Android guy, because I changed the VM and I probably broke something in Giskard.

## :squid: Permissions
Go ahead!


simonrozsival:
I created another PR which breaks things a bit (not all that much) and I suggest that we discuss that one first before we start updating the other view sources: #4346

Toggl.Daneel/ViewSources/ReportsTableViewSource.cs
```diff
@@ -3,98 +3,85 @@
...
-        protected override void Dispose(bool disposing)
-        {
-            base.Dispose(disposing);
-
-            if (!disposing) return;
-
-            disposeBag.Dispose();
-        }
+        public override UITableViewCell GetCell(UITableView tableView, NSIndexPath indexPath)
+            => ReportsLegendViewCell.CellConfiguration(ReportsLegendViewCell.Identifier)
```
simonrozsival: The configuration should be cached in some `private readonly` field and not created every time we show a cell.

PR #4346, opened 2019-02-05
Merge ReloadTableViewSource and BaseTableViewSource

simonrozsival:
## What's this?
This is an attempt to make the refactored table view sources easier to use when we create custom subclasses of `(Reload)BaseTableViewSource`.

### Relationships
Follow-up of #4318

## Why do we want this?
I started rewriting the `TimeEntriesLogViewSource` to use `ReloadTableViewSource` and I ran into a few issues because the function passed as the cell configuration to the `BaseTableViewSource` needed to access to some instance members and the callback method had to be static so it could be passed through the constructor.

This could be solved by making the callback a property the same way as `ConfigureHeader` but that doesn't seem right at all. It felt like we were reinventing inheritance. I decided to propose this solution instead.

## How is it done?
All the configurable code, which is really handy for simple tables, is moved to a specific sealed class `ConfigurableTableViewSource` and it is used almost exactly the same way as it was used previously (see `SelectDateFormatViewController.cs:40`) with the exception that header configuration is also passed through the constructor.

For the more advanced cases, when we need a custom class to do something extra, we simply override the methods of the base class (`GetCell`, `GetViewForHeader`) and we have much more flexibility.

Once I started moving code around it turned out that we don't really need two base classes. We are fine with one base class for table view sources and one sealed for configurable TVS.

Also the naming didn't reflect how it actually works. `ReloadTableViewSource` didn't reload the table. The reloading was done in the bindings. The bindings should be called `ReloadSections` and `ReloadItems` instead (and so I renamed them).

### Why not another way?
As I described it in the first paragraph, I wanted to simplify the code. I hope I was successful. In the end I just renamed `ReloadTableViewSource` into `ConfigurableTableViewSources`, sealed it and moved some code from the base class here.

### Side effects
No side effects.

## Review hints
Definitely try this in the simulator.

## :squid: Permissions
Just me. I want to discuss this with @Odrakir and only merge this once we come to an agreement even if this PR gets some approvals.


Toggl.Daneel/Extensions/Reactive/BaseTableViewSourceExtensions.cs
```diff
@@ -13,11 +12,5 @@ public static class BaseTableViewSourceExtensions
             => Observable
                 .FromEventPattern<TModel>(e => reactive.Base.OnItemTapped += e, e => reactive.Base.OnItemTapped -= e)
                 .Select(e => e.EventArgs);
-
-        public static IObservable<CGPoint> Scrolled<THeader, TModel>(
```
Odrakir: Don't remove this please, it's not used here, but it's used when refactoring the rest of the table views.

kornelijepetak:
As the matter of fact, I will approve this, under condition that my naming suggestion is heeded to.

Toggl.Daneel/Extensions/Reactive/BaseTableViewSourceExtensions.cs
simonrozsival: OK.

simonrozsival:
> First `ConfigurableTableViewSource` is a mouthful, can we find a better name for this?

I won't mind finding a better name. My ideas: `GenericTableViewSource`, `CustomTableViewSource`, `InlineTableViewSource`, or just `TableViewSource`.

> All this results in two usable sources: `BaseTableViewSource` and `ConfigurableTableViewSource` and you'll need to use one or the other depending if you are subclassing or not... Isn't that more complex than it used to be?

It's easier because you can't do anything in between. When you subclassed `ReloadTableViewSource` you could implement headers either with the configuration or somehow else. And you could even pass some random function through the constructor override the `GetCell` so you could achieve some behavior which wasn't possible with the (static) function.

I propose that either you use the configurable TVS or you create custom class and then you do cell and header definition using inheritance. There is only one way of doing things for a specific use case.

> We might want to add some animation configuration for the animated table views (what kind of animation happens when deleting, inserting, updating or moving). The idea was to add that to the `AnimatedTableViewSource`, but if there's not one anymore... I'm not sure if there's an easy way to add that to the rx extension.

At this point I'm not sure how exactly that thing will be implemented. But I think that either it will extend `BaseTableViewSource` with an `AnimatedTableViewSource` or it will be just a static extension method (something like `ReloadSections`). I think we don't limit ourselves in any way with this change.

> For the diffing to happen we need access to the current items in the datasource, so we can compare them to the new ones. With the code like this, if we intend to do in a `AnimateSections` or `AnimateItems` rx extension methods we'll need to make the source `Sections` public, which doesn't seem great.

You have a valid point. On the other hand having a public getter for an immutable list doesn't seem problematic to me (the setter has to be private/at most protected anyway).

> Finally now we'll have 4 methods for the binding `ReloadItems`, `ReloadSections`, `AnimateItems` and `AnimateSections`. Maybe this can be simplified too.

It will depend on the usecases. If switching between animations an reload-only is only a matter of changing the name of the function you call (and maybe passing some extra configuration) and not having to create any new class, it would be awesome, but at this point I don't know how this will be used. Since both `ReloadItems` and `ReloadSections` are so simple, it should be easy to change them in the future if we need to do that for any reason.

kornelijepetak:
I just want to add that I agree with most of what Simon said. 😄

Odrakir:
Alright, you've convinced me. About animation configuration, we'll see when we need to do it, but maybe it's ok to add it in the base table view even if it's not used for reload sources.

Just change the name of `ConfigurableTableViewSource`. I like `CustomTableViewSource`... maybe.

I also was thinking on adding a static method to create a source for `BaseTableViewCells`, but that might make things harder to understand.

PR #4355, opened 2019-02-06
Time entries log reloading

simonrozsival:
## What's this?
This PR modifies the time entries view source to use table reloading instead of the old view source. This is a temporary change which will be later changed so that the changes are animated, but the new API will stay.

### Relationships
Ref #4218 

## Why do we want this?
We want to be able to implement grouping of time entries without worrying about animation issues (remember https://github.com/toggl/mobileapp/pull/1848? 😂 😢). The API for the animated table will be the same so the VM part won't be touched later on and everything will happen only in Daneel and Giskard.

## How is it done?
The VM now groups the TEs directly (only by the dates, not by similarity just yet) without the use of `ObservableGroupedOrderedCollection` and it is displayed without `ReactiveTableViewBinder` and `MutableSectionedListTableViewSource`.

Most of the changes in the code are in the VM tests.

### Why not another way?
This seems like a good idea to me. You don't agree? Please leave comments?

### Side effects
This PR breaks Giskard... it should build just fine, it just won't display any data in the main log. We need to fix that in cooperation with @kornelijepetak.

Also the app now looks weird because there are no animations in the main view. That should not be a problem because we will add this back before we release this to the users.

## Review hints
Make sure to try this in iOS simulator.

## :squid: Permissions
me or @kornelijepetak


Toggl.Foundation.MvvmCross/ViewModels/MainViewModel.cs
```diff
@@ -74,7 +76,7 @@ public sealed class MainViewModel : MvxViewModel
         public IObservable<bool> ShouldShowStoppedTimeEntryNotification { get; private set; }
         public IObservable<IThreadSafeTimeEntry> CurrentRunningTimeEntry { get; private set; }
 
-        public ObservableGroupedOrderedCollection<TimeEntryViewModel> TimeEntries => TimeEntriesViewModel.TimeEntries;
+        public IObservable<IImmutableList<CollectionSection<DaySummaryViewModel, TimeEntryViewModel>>> TimeEntries => TimeEntriesViewModel.TimeEntries;
```
simonrozsival: I think this will deserve a new class once we add the third level of nesting. Maybe that type should also handle all the grouping (it probably should). Just thinking out loud...

Toggl.Foundation.MvvmCross/ViewModels/TimeEntriesViewModel.cs
```diff
@@ -47,62 +49,44 @@ public sealed class TimeEntriesViewModel
...
+                        dataSource.Workspaces.ItemsChanged(),
+                        showUndoSubject.SelectUnit())
+                    .StartWith(Unit.Default)
+                    .SelectMany(_ =>
+                        interactorFactory.GetAllTimeEntriesVisibleToTheUser().Execute())
+                    .DistinctUntilChanged()
+                    .CombineLatest(dataSource.Preferences.Current, processTimeEntries)
+                    .AsDriver(schedulerProvider);
+
+            Empty = TimeEntries.Select(groups => groups.None()).AsDriver(schedulerProvider);
```
kornelijepetak: Operators on separate lines, please. For consistency sake.

Toggl.Foundation.MvvmCross/ViewModels/TimeEntriesViewModel.cs
```diff
@@ -47,62 +49,44 @@ public sealed class TimeEntriesViewModel
...
-            );
+            this.timeService = timeService;
+
+            TimeEntries =
+                Observable.Merge(
+                        dataSource.TimeEntries.ItemsChanged(),
+                        dataSource.Workspaces.ItemsChanged(),
+                        showUndoSubject.SelectUnit())
+                    .StartWith(Unit.Default)
+                    .SelectMany(_ =>
```
kornelijepetak: The method is not complex, please keep it on the same line.

Toggl.Foundation.MvvmCross/ViewModels/TimeEntriesViewModel.cs
```diff
@@ -47,62 +49,44 @@ public sealed class TimeEntriesViewModel
...
+                        interactorFactory.GetAllTimeEntriesVisibleToTheUser().Execute())
+                    .DistinctUntilChanged()
+                    .CombineLatest(dataSource.Preferences.Current, processTimeEntries)
+                    .AsDriver(schedulerProvider);
+
+            Empty = TimeEntries.Select(groups => groups.None()).AsDriver(schedulerProvider);
+            Count = TimeEntries.Select(log => log.Sum(day => day.Items.Count)).AsDriver(schedulerProvider);
```
kornelijepetak: Operators on the same line. For consistency sake.

Toggl.Foundation.MvvmCross/ViewModels/TimeEntriesViewModel.cs
```diff
@@ -162,65 +140,28 @@ private IObservable<TimeEntryViewModel> deleteTimeEntry(TimeEntryViewModel timeE
...
-        private void onTimeEntryAdded(IThreadSafeTimeEntry timeEntry)
-        {
-            var timeEntryViewModel = new TimeEntryViewModel(timeEntry, durationFormat);
-            TimeEntries.InsertItem(timeEntryViewModel);
-        }
+        private IImmutableList<CollectionSection<DaySummaryViewModel, TimeEntryViewModel>> processTimeEntries(
+            IEnumerable<IThreadSafeTimeEntry> timeEntries, IThreadSafePreferences preferences)
+            => timeEntries
+                .Where(isNotRunning)
+                .Where(isNotUndoablyDeleted)
```
kornelijepetak: I am so troubled by this name. 😄 Two negations, possibly three if you treat `delete` as a negative operation.

Maybe something like:
`isDeleteNotUndoable`
`isUnrecoverable`
`isPermantlyDeleted`

simonrozsival: The names you suggested don't reflect what the method does. It lets through everything that's not the time entry which is currently "deleted" by user, but can be brought back to life if the user taps the "Undo button". I will come up with a different name and try to minimize the number of negations.

Toggl.Foundation.MvvmCross/ViewModels/MainViewModel.cs
Odrakir: I insist in avoiding the creation of a generic structure for 3 levels nesting. Just add whatever `IEnumerable` you need inside that `TimeEntryViewModel` and that's it.

simonrozsival: I didn't mean a generic structure. I meant a specific class for time entries log which would contain the (fairly complex) grouping logic and which would be an alias for `IImmutableList<CollectionSection<DaySummaryViewModel, CollectionSection<TimeEntriesGroupViewModel, TimeEntryViewModel>>>`. It will be in a different PR for sure. Maybe it won't even be necessary.

Toggl.Daneel/ViewSources/TimeEntriesLogViewSource.cs
```diff
@@ -36,26 +35,15 @@ public sealed class TimeEntriesLogViewSource : ReactiveSectionedListTableViewSou
...
         public IObservable<TimeEntriesLogViewCell> FirstCell { get; }
+        public IObservable<bool> IsDragging { get; }
 
-        public TimeEntriesLogViewSource(
-            ObservableGroupedOrderedCollection<TimeEntryViewModel> collection,
-            string cellIdentifier,
-            ITimeService timeService,
-            ISchedulerProvider schedulerProvider)
-            : base(collection, cellIdentifier)
+        public TimeEntriesLogViewSource(ISchedulerProvider schedulerProvider)
```
Odrakir: I'm not sure if the `ISchedulerProvider` was added because something wasn't working, but I think we don't need it, do we? All those events are generated in the main thread to start with.

Toggl.Daneel/ViewSources/Generic/ReactiveSectionedListTableViewSource.cs
```diff
@@ -1,56 +1,23 @@
...
 
 namespace Toggl.Daneel.ViewSources
 {
-    public interface IObservableScroll
-    {
-        IObservable<CGPoint> ScrollOffset { get; }
-        IObservable<bool> IsDragging { get; }
-    }
-
     public abstract class ReactiveSectionedListTableViewSource<TModel, TCell>
```
Odrakir: Do we still need this file at all?

Toggl.Daneel/ViewSources/TimeEntriesLogViewSource.cs
```diff
@@ -160,6 +142,22 @@ public override void RowSelected(UITableView tableView, NSIndexPath indexPath)
             tableView.DeselectRow(indexPath, true);
         }
 
+        public override void DraggingStarted(UIScrollView scrollView)
+        {
+            isDraggingSubject.OnNext(true);
```
Odrakir: Does it make sense to add this to `BaseTableViewSource` and create an Rx extension as we did for `Scrolled` and `ModelSelected`?

Toggl.Foundation.MvvmCross/ViewModels/TimeEntriesViewModel.cs
```diff
@@ -47,62 +49,48 @@ public sealed class TimeEntriesViewModel
...
-            TimeEntries = new ObservableGroupedOrderedCollection<TimeEntryViewModel>(
-                indexKey: t => t.Id,
-                orderingKey: t => t.StartTime,
-                groupingKey: t => t.StartTime.LocalDateTime.Date,
-                descending: true
-            );
+            this.timeService = timeService;
+
+            TimeEntries =
+                Observable.Merge(
```
Odrakir: I feel like a lot of code in this file could (should) be in an Interactor. Kind of like `ObserveTimeTrackedTodayInteractor` works...

Toggl.Foundation.MvvmCross/ViewModels/MainViewModel.cs
Odrakir: What I mean is creating a new view model that could hold all that information.

Toggl.Daneel/ViewSources/TimeEntriesLogViewSource.cs
simonrozsival: I don't think it's needed anywhere else, so I think we shouldn't put it there just yet.

Odrakir: I'd simplify this class though.

PR #4363, opened 2019-02-08
Group TEs by similarity

simonrozsival:
## What's this?
This PR just adds a static transformation function which is not used anywhere yet. It will be used in the time entries VM to provide data for the main log table.

### Relationships
Ref #4218 

## Why do we want this?
This is core logic of this project.

## How is it done?
A static class with a static method. There is quite a lot of code and I didn't want to keep it in the VM. It also doesn't make sense to instantiate this class because it's a pure function. Keeping it static might also give us a small performance bonus.

I also added a VM for the group "header". I didn't want to unseal the `TimeEntryViewModel` class and I decided to just copy its properties. There are some interesting bits - e.g., the decision if it's continuable or if it needs syncing and so on... it is based on the discussion in slack: https://toggl.slack.com/archives/CDRNW2N1F/p1549019878090500

I didn't write any tests for the VM because it seemed trivial. Now I'm not sure if it was a good idea 🤔. Please tell me if you think if it would be beneficial to write tests for the class in the review.

### Side effects
No side effects, it's not used anywhere yet.

## Review hints
Check if I covered the important cases with unit test. ✅ should mean that I did everything correctly.

## :squid: Permissions
🦑


Toggl.Foundation.MvvmCross/ViewModels/TimeEntriesLog/TimeEntriesGroupViewModel.cs
```diff
@@ -0,0 +1,72 @@
...
+            IsBillable = representative.IsBillable;
+            TaskId = representative.TaskId;
+            TagIds = representative.TagIds;
+            HasProject = representative.HasProject;
+            HasTags = representative.HasTags;
+            HasDescription = representative.HasDescription;
+            DurationFormat = representative.DurationFormat;
+            IsInaccessible = representative.IsInaccessible;
+
+            NeedsSync = timeEntries.All(timeEntry => timeEntry.NeedsSync);
```
kornelijepetak: Is this actually been decided by product/design/lead teams?
So, only if ALL TEs require sync will the group require sync? This is strange to me.

Toggl.Foundation.MvvmCross/Transformations/TimeEntriesLogTransformation.cs
```diff
@@ -0,0 +1,80 @@
...
+                   && y != null
+                   && x.WorkspaceId == y.WorkspaceId
+                   && x.Description == y.Description
+                   && x.ProjectId == y.ProjectId
+                   && x.TaskId == y.TaskId
+                   && x.IsBillable == y.IsBillable
+                   && haveSameTags(x.TagIds, y.TagIds);
+
+            public int GetHashCode(TimeEntryViewModel timeEntry)
+                => HashCode.From(timeEntry.WorkspaceId, timeEntry.Description, timeEntry.ProjectId);
```
kornelijepetak: `GetHashCode` is used in the equality comparer to fast detect if the items differ, so that `Equals` is not necessary to call for those that differ. If hash codes match, it will call `Equals` to verify that they are actually the same. 

What you are doing here is skipping checks on tags, which means that all items with the same workspace, description and project will have the same hash code, regardless of the tags. This has a potential to increase the number of hash conflicts. That means that for items that DON'T have the same tags it would have to call `Equals` regardless (which should be prevented by `GetHashCode`). Since your `haveSameTags` uses `OrderBy` on the tag collections, it is slower than just calculating the hashcodes. I am aware that the average number of tags on a TE is rather small, so this may not be that relevant from the performance perspective (_requires empirical verification_), but I would still do it properly for consistency.

It is then better if `GetHashCode` depends on ALL of the items used in Equals, including the tags, so I'd suggest:

```cs
public int GetHashCode(TimeEntryViewModel timeEntry)
{
	var hashCode = HashCode.From(timeEntry.WorkspaceId, timeEntry.Description, timeEntry.ProjectId);
	
	foreach(var tagId in timeEntry.TagIds)
        hashCode = HashCode.From(hashCode, tagId);

	return hashCode;
}
```

Toggl.Foundation.MvvmCross/ViewModels/TimeEntriesLog/TimeEntriesGroupViewModel.cs
kornelijepetak: Ok, ignore this. It has been changed.

Toggl.Foundation.MvvmCross/Transformations/TimeEntriesLogTransformation.cs
```diff
@@ -0,0 +1,80 @@
...
+        private static CollectionSection<DaySummaryViewModel, CollectionSection<TimeEntriesGroupViewModel, TimeEntryViewModel>> transform(
+            IGrouping<DateTime, TimeEntryViewModel> dayGroup, IThreadSafePreferences preferences, DateTimeOffset currentDateTime)
+            => new CollectionSection<DaySummaryViewModel, CollectionSection<TimeEntriesGroupViewModel, TimeEntryViewModel>>(
+                new DaySummaryViewModel(
+                    DateToTitleString.Convert(dayGroup.Key, currentDateTime),
+                    totalTrackedTime(dayGroup).ToFormattedString(preferences.DurationFormat)),
+                toGroups(dayGroup, preferences));
+
+        private static TimeSpan totalTrackedTime(IEnumerable<TimeEntryViewModel> timeEntries)
+            => timeEntries.Sum(timeEntry => timeEntry.Duration ?? TimeSpan.Zero);
```
kornelijepetak: There's an ext. method to simplify this:

```suggestion
            => timeEntries.Sum(timeEntry => timeEntry.Duration);
```

Toggl.Foundation.MvvmCross/ViewModels/TimeEntriesLog/TimeEntriesGroupViewModel.cs
```diff
@@ -0,0 +1,43 @@
...
+            var sample = timeEntries.First();
+
+            GroupSize = timeEntries.Length;
+            IsExpandable = timeEntries.Length > 1;
+            Summary = new TimeEntryViewModel(
+                id: 0,
+                isBillable: sample.IsBillable,
+                description: sample.Description,
+                startTime: timeEntries.Min(timeEntry => timeEntry.StartTime),
+                duration: timeEntries.Sum(timeEntry => timeEntry.Duration ?? TimeSpan.Zero),
```
kornelijepetak: There's an ext. method to simplify this:

```suggestion
                duration: timeEntries.Sum(timeEntry => timeEntry.Duration),
```

Issue #4219, opened 2019-01-16
Update EditTimeEntryViewModel to support Grouped Time Entries


kornelijepetak:
The Edit Time entry screen currently accepts only the TE ID (`long`) and then loads the TE again from the database. Extend this parameter type to include the information about whether the edit is requested for a single item, or a group (something like `isEditingGroup`).

🏆 Update EditTimeEntryViewModel to support grouped time entries:
* [x] Parameter change
* [x] Setup logic
* [x] Deletion logic
* [x] Saving logic
* [x] Tests

The setup flow should be as follows:

* [x] Load TE specified by the ID
* [x] If the edit was requested for a single item, that's it.
* [x] If the edit was requested for a group, fetch all the TEs matching the same group criteria as the TE specified by the ID to find out the number of affected TEs
* [x] Emit that number through an observable so that view can show how many entries would be affected with the edit operation

The save flow:
* [x] If the edit was requested for a single item, update that item and close the view
* [x] If the edit was requested for a group, fetch all the TEs matching the same group criteria as the TE specified by the ID, update them according to the user's changes and then close the view.

The deletion logic:
* [x] The logic is the same as for the save, but instead of updating the TEs, delete them.

The reason why caching of grouped TE during setup is not suggested is the fact that a scheduled sync (or one started manually before opening the edit view for the group) could cause the number of affected TEs to actually change. It makes a bit more sense to update all the affected TEs at the moment of saving, and not those that would be affected at the Edit screen setup.

But I would like @amulware and @maritoggl to confirm this, or propose alternative solution.

maritoggl:
It sounds like this is about how things are handled on the background, so I don't have a comment on that. I could comment, approve or disprove if I understood how this appears to the user :)

kornelijepetak:
Closed by #4407

PR #4367, opened 2019-02-08
Grouped TEs on main log

### LINKED ISSUE #4218: Implement infrastructure for grouped Time Entries

simonrozsival:
## What's this?
This PR uses #4363 and it actually contains the grouping logic. ~It always displays the groups "expanded" (there is no way of collapsing them) and it doesn't render any headers between the groups which would contain some visual clue that the TEs are grouped (that will come in a following PR).~ This PR contains all the logic for collapsing and expanding groups of TEs afterall. Don't expect any animations though.

### Relationships
Needs #4366 to be merged first (it is branched off that branch).

Closes #4218

## Why do we want this?
This is a step before we actually start rendering the group headers and before we add the collapsing/expanding logic.

## How is it done?
I changed the `TimeEntryViewModel` (and I can assure you that it will change even more in the following PR when it will be probably merged with `TimeEntriesGroupViewModel` after all) and that lead me to a rather complicated refactoring of all the related things around the `MainVM` and `TimeEntriesViewModel`. Namely deleting and continuing of time entries which relied on data being inside of the view models (which is IMO abusing of the VMs which should only contain information for the view).

~There is temporary flattening in `UITableViewExtensions.ReloadTimeEntries`. This piece of code will be moved somewhere else in the end (this piece of code should handle the expansion and collapsing of the groups, it might be in some `Binder` class in the end, but I'm not sure about it just yet).~

Collapsing and third-level flattening is done in the VM and this code is therefore shared. Grouping by date and, if the user has the option enabled, by similarity is done by a static transformation function (it doesn't have any internal state, it doesn't need to have an instance). Flattening is done in a stateful class which keeps track of current preferences, current time, and remembers which groups are expanded and which are collapsed.

### Why not another way?
~Please don't make me rewrite this nightmare.~

### Side effects
The time entries are being "grouped" which means that they are always rendered next to each other and they are not displayed chronologically as they used to (when the grouping settings is on). It seems weird but it is necessary for the following PRs.

~This PR also most likely breaks Android even more. It should be "fine" because it's already broken in this project branch, but please bear in mind that I didn't even bother trying to work with Android code in this PR. That's for somebody else (=Kornelije) to fix later.~ Don't review Giskard stuff and don't worry about Android app, this PR will be merged into `project/grouped-time-entries`, not into `develop`. I made Giskard build and it kinda works, but the UI is not ready yet (the cells in the table don't show how many items are in a group and it's not possible to expand/collapse the groups.

## Review hints
Try this in iOS simulator and make sure it doesn't crash and that starting TEs, continuing and deleting works.

## :squid: Permissions
~Not yet. But feedback is welcome anyway so please have a look at the code.~

🦑 whenever.


kornelijepetak:
> and I can assure you that it will change even more in the following PR when it will be probably merged with `TimeEntriesGroupViewModel` after all

Please explain why.

I am troubled that we are not using proper abstractions but are conflating stuff needlessly.

I don't care if there's another similar class for the view model of the TE groups that has many of the same properties as the TE view model. It is what it is. I see no reason to use the type that in its name does not represent ALL that it is.

There's a small chance I misunderstood your intentions, so please explain it a bit further.

Toggl.Foundation.MvvmCross/Transformations/TimeEntriesLogTransformation.cs
```diff
@@ -87,9 +93,12 @@ public int GetHashCode(TimeEntryViewModel timeEntry)
                 return hashCode;
             }
 
-            private static bool haveSameTags(IReadOnlyCollection<long> a, IReadOnlyCollection<long> b)
-                => (a == null && b == null)
-                   || (a?.Count == b?.Count && a.OrderBy(id => id).SequenceEqual(b.OrderBy(id => id)));
+            private static bool haveSameTags(IEnumerable<long> a, IEnumerable<long> b)
+            {
+                var first = a as long[] ?? a.ToArray();
```
kornelijepetak: But now you no longer have a null-ref check.
If `a` or `b` gets into this function as `null`, it will fail at `a.ToArray()` or `b.ToArray()`.

Toggl.Foundation.MvvmCross/ViewModels/MainViewModel.cs
```diff
@@ -92,8 +94,7 @@ public sealed class MainViewModel : MvxViewModel
         public InputAction<bool> StartTimeEntry { get; private set; }
         public InputAction<long> SelectTimeEntry { get; private set; }
         public InputAction<TimeEntryStopOrigin> StopTimeEntry { get; private set; }
-        public InputAction<TimeEntryViewModel> DeleteTimeEntry { get; private set; }
-        public InputAction<TimeEntryViewModel> ContinueTimeEntry { get; private set; }
+        public InputAction<long> ContinueTimeEntry { get; private set; }
```
kornelijepetak: Why was `DeleteTimeEntry` deleted?

Toggl.Daneel/Extensions/Reactive/UITableViewExtensions.cs
```diff
@@ -22,6 +29,21 @@ public static class UITableViewExtensions
             });
         }
 
+        public static IObserver<TimeEntriesLog> ReloadTimeEntries(
+            this IReactive<UITableView> reactive, TimeEntriesLogViewSource dataSource)
+        {
+            return Observer.Create<TimeEntriesLog>(groupedTimeEntries =>
```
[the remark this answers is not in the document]
simonrozsival: You are probably right.

Toggl.Foundation.MvvmCross/Transformations/TimeEntriesLogTransformation.cs
simonrozsival: This wasn't supposed to be a null check. This is a conversion to an array. TagIds "cannot be null".

Toggl.Foundation.MvvmCross/ViewModels/MainViewModel.cs
simonrozsival: It's not being used. Not even in Giskard. We use the `public InputAction<long> DelayDeleteTimeEntry { get; }` action of `TimeEntriesViewModel` instead (it used to accept the time entry VM and I changed that to `long` which is one of the reasons why Android build failed).

Toggl.Foundation.MvvmCross/Transformations/TimeEntriesLogTransformation.cs
kornelijepetak: So, I understand that you are ok with this crashing if you accidentally do send some nulls inside... that makes sense. And is consistent with our fail-fast policy.

Toggl.Foundation.MvvmCross/ViewModels/MainViewModel.cs
kornelijepetak: Yes, I found it later in the related VM, but forgot to delete this comment of mine.

Toggl.Foundation.MvvmCross/Transformations/TimeEntriesLogTransformation.cs
simonrozsival: Yes.

Toggl.Daneel/Extensions/Reactive/UITableViewExtensions.cs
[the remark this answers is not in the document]
simonrozsival: I would like to do it in a separate PR though.

Toggl.Daneel/ViewControllers/MainViewController.cs
```diff
@@ -104,7 +104,7 @@ public override void ViewDidLoad()
             TimeEntriesLogTableView.Source = tableViewSource;
 
             ViewModel.TimeEntries
-                .Subscribe(TimeEntriesLogTableView.Rx().ReloadSections(tableViewSource))
+                .Subscribe(TimeEntriesLogTableView.Rx().ReloadTimeEntries(tableViewSource))
```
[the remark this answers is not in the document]
Odrakir: I completely agree with @kornelijepetak

Toggl.Daneel/Extensions/Reactive/UITableViewExtensions.cs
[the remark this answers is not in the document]
Odrakir: As I said in the other comment I agree with @kornelijepetak that having an extension for something so specific smells bad. I think any transformation should be done directly in the VM or in the VC if needed.

Toggl.Daneel/ViewSources/TimeEntriesLogViewSource.cs
```diff
@@ -21,9 +21,9 @@ public sealed class TimeEntriesLogViewSource
         private const int rowHeight = 64;
         private const int headerHeight = 48;
 
-        private readonly Subject<TimeEntryViewModel> continueTapSubject = new Subject<TimeEntryViewModel>();
-        private readonly Subject<TimeEntryViewModel> swipeToContinueSubject = new Subject<TimeEntryViewModel>();
-        private readonly Subject<TimeEntryViewModel> swipeToDeleteSubject = new Subject<TimeEntryViewModel>();
+        private readonly Subject<long> continueTapSubject = new Subject<long>();
+        private readonly Subject<long> swipeToContinueSubject = new Subject<long>();
+        private readonly Subject<long> swipeToDeleteSubject = new Subject<long>();
```
Odrakir: Why changing these? Apparently you need to fetch the `TimeEntryViewModel` in the VM again to perform the operation, so it seems like double the work.

simonrozsival: The VMs had to contain a lot of additional information which wasn't displayed anywhere (workspace ID, project ID, task ID, ...) - view models should contain data for the view, not for business logic. Now that there are two view models, one for group headers and one for individual time entries, it makes more sense to me to pass IDs (and soon arrays of IDs) around instead of the whole VMs.

Toggl.Daneel/ViewSources/TimeEntriesLogViewSource.cs
```diff
@@ -85,9 +81,48 @@ public override UITableViewCell GetCell(UITableView tableView, NSIndexPath index
         {
             var cell = (TimeEntriesLogViewCell)tableView.DequeueReusableCell(TimeEntriesLogViewCell.Identifier);
 
-            cell.Item = ModelAt(indexPath);
+            var model = ModelAt(indexPath);
+            switch (model)
+            {
+                case TimeEntriesGroupViewModel group:
```
Odrakir: Both types of cells are very similar to each other visually. I think this would be much more simpler if we have just one type of view model (`MainLogItemViewModel` or whatever)with different properties and do the configuration inside the cell when `Item` is set.

Toggl.Daneel/ViewSources/Generic/TableView/ExpandableThreeLevelTableViewSource.cs
```diff
@@ -0,0 +1,61 @@
+using System.Collections.Generic;
+using System.Linq;
+using Toggl.Foundation.MvvmCross.Collections;
+using Toggl.Foundation.MvvmCross.ViewModels;
+
+namespace Toggl.Daneel.ViewSources
+{
+    public abstract class ExpandableThreeLevelTableViewSource<THeader, TSecondLevelHeader, TModel>
```
Odrakir: We don't need this class at all. I think `TimeEntriesLogViewSource` should inherit directly from `BaseTableViewSource`. When ever a cell wants to expand / collapse, that action executes in `MainViewModel` or `TimeEntriesViewModel` and that generates a new event in the log observable.

The event is taken in the datasource and it reloads the cells with the appropriate expanding/collapsing (for now, in the future it'll animate the changes).

Of course this means that the collection we send to the table view only contains the items that's actually going to show, not the collapsed ones.

Odrakir: This also means that the table view reloads only in one point in the application: The table view extension. And also that you don't need the custom `ReloadNestedSections` extension.

Toggl.Daneel/ViewSources/TimeEntriesLogViewSource.cs
simonrozsival: We had a discussion about this with @kornelijepetak and I proposed that solution. We decided to have two separate VMs. I see the `TimeEntriesGroupViewModel` as a pain point and I agree that we can still give more thought into it.

Toggl.Daneel/ViewSources/Generic/TableView/ExpandableThreeLevelTableViewSource.cs
simonrozsival: I don't think this should happen in the shared code layer (`Foundation.MvvmCross`), because we flatten things differently on iOS and on Android, and that is why I put it into the TVS directly. You gave me a thought with this comment - I have an idea how to take this class and put it as a transformation between the VM observable and the reloading extension method.

simonrozsival:
The tests are now broken. I will fix them once I get some feedback from you on the approach I tried with the last few changes.

Toggl.Foundation.MvvmCross/ViewModels/IExpandableViewModel.cs
```diff
@@ -0,0 +1,9 @@
+namespace Toggl.Foundation.MvvmCross.ViewModels
+{
+    public interface IExpandableViewModel
```
Odrakir: Is this used anywhere?

Toggl.Foundation.MvvmCross/ViewModels/MainViewModel.cs
```diff
@@ -77,7 +77,8 @@ public sealed class MainViewModel : MvxViewModel
         public IObservable<bool> ShouldShowStoppedTimeEntryNotification { get; private set; }
         public IObservable<IThreadSafeTimeEntry> CurrentRunningTimeEntry { get; private set; }
 
-        public IObservable<IImmutableList<CollectionSection<DaySummaryViewModel, TimeEntryViewModel>>> TimeEntries => TimeEntriesViewModel.TimeEntries;
+        public IObservable<IImmutableList<CollectionSection<DaySummaryViewModel, LogItemViewModel>>> TimeEntries
```
Odrakir: Is this immutable because of Android? I mean, if this was Swift it would be immutable by default and it really does make sense for it to be immutable, I just feel that the `toImmutableList` thing pollutes the code and I'm not completely sure of the benefit. How could that list be modified anyway?

Toggl.Foundation.MvvmCross/ViewModels/MainViewModel.cs
```diff
@@ -413,11 +412,16 @@ private Task startTimeEntry(bool useDefaultMode)
             return navigate<StartTimeEntryViewModel, StartTimeEntryParameters>(parameter);
         }
 
-        private IObservable<Unit> continueTimeEntry(TimeEntryViewModel timeEntry)
+        private IObservable<Unit> continueTimeEntry(long timeEntryId)
         {
             return interactorFactory
-                .ContinueTimeEntry(timeEntry)
+                .GetTimeEntryById(timeEntryId)
                 .Execute()
```
Odrakir: For me this should be just one line: `interactorFactory.ContinueTimeEntry(timeEntry).GetTimeEntryById(timeEntryId).Execute()`, but I don't think we have guidelines on this.

Toggl.Foundation.MvvmCross/ViewModels/MainViewModel.cs
```diff
@@ -413,11 +412,16 @@ private Task startTimeEntry(bool useDefaultMode)
...
+                .Select(timeEntry => timeEntry.AsTimeEntryPrototype())
+                .SelectMany(prototype =>
+                    interactorFactory
+                        .ContinueTimeEntry(prototype)
+                        .Execute())
```
Odrakir: Same here, one line: `interactorFactory.ContinueTimeEntry(prototype).Execute()`

Toggl.Foundation.MvvmCross/ViewModels/TimeEntriesViewModel.cs
```diff
@@ -64,12 +67,22 @@ public sealed class TimeEntriesViewModel
...
 
-            var whenDeletingTimeEntry = showUndoSubject.SelectUnit().StartWith(Unit.Default);
+            DelayDeleteTimeEntry = rxActionFactory.FromAction<long>(delayDeleteTimeEntry);
+            ToggleGroupExpansion = rxActionFactory.FromAction<GroupId>(toggleGroupExpansion);
+            CancelDeleteTimeEntry = rxActionFactory.FromAction(cancelDeleteTimeEntry);
+
+            var deletingOrPressingUndo = showUndoSubject.SelectUnit().StartWith(Unit.Default);
+            var collapsingOrExpanding = ToggleGroupExpansion.Elements.StartWith(Unit.Default);
 
             TimeEntries =
```
Odrakir: Beautiful 😍

Toggl.Foundation.MvvmCross/Transformations/TimeEntriesFlattening.cs
```diff
@@ -0,0 +1,100 @@
...
+namespace Toggl.Foundation.MvvmCross.Transformations
+{
+    public static class TimeEntriesFlattening
+    {
+        public static IEnumerable<CollectionSection<DaySummaryViewModel, LogItemViewModel>> Flatten(
+            IEnumerable<CollectionSection<DateTimeOffset, IThreadSafeTimeEntry[]>> days,
+            HashSet<GroupId> expandedGroups,
+            DateTimeOffset currentDateTime,
+            IThreadSafePreferences preferences)
+            => days.Select(day =>
```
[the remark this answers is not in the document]
kornelijepetak: I simply can't make @simonrozsival stay out of nested lambdas 😄 
In general, naming could do wonders.

Toggl.Daneel/ViewControllers/MainViewController.cs
```diff
@@ -134,7 +139,7 @@ public override void ViewDidLoad()
                 .DisposedBy(DisposeBag);
 
             tableViewSource.Rx().ModelSelected()
-                .Select(te => te.Id)
+                .Select(model => model.RepresentedTimeEntriesIds.First())
```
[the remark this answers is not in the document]
simonrozsival: The `.First()` will go away once we have #4219.

I'm now thinking about what we should feed to our actions. Maybe it should be the VMs afterall even though we would just pick the ID from them, because the VC and the table needs to select the data for the VM. If we had a specific type for IDs I would be all for passing just IDs around, but since it's all `long`s, I might revert this back to using the whole VMs as the inputs for the actions.

Toggl.Foundation.MvvmCross/Transformations/TimeEntriesFlattening.cs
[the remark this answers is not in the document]
simonrozsival: I agree that it's very condensed in this case. I will revisit this.

Toggl.Foundation.MvvmCross/ViewModels/IExpandableViewModel.cs
simonrozsival: Not anymore, I forgot to 🔥 it.

Toggl.Foundation.MvvmCross/ViewModels/MainViewModel.cs
simonrozsival: We could keep this as `IEnumerable`. It's also "immutable" (until you cast it to the underlying type). I think we might get rid of it for now.

Toggl.Foundation.MvvmCross/ViewModels/TimeEntriesLog/GroupId.cs
```diff
@@ -0,0 +1,35 @@
...
+        private readonly string key;
+
+        public GroupId(IThreadSafeTimeEntry sample)
+        {
+            key = string.Join(
+                "\n",
+                sample.Start.LocalDateTime.Date,
+                sample.Description,
+                sample.WorkspaceId,
+                sample.Project?.Name,
```
kornelijepetak: If it really has to be a string (and it doesn't), please at least use the IDs here instead of names (which is something user has control of). I've pasted some text into a `create new project` name that contains a newline, and it kind of produced strange behavior on mobile AND webapp. So, please, rely on things user has no control of (and that is an ID).

Toggl.Foundation.MvvmCross/Transformations/TimeEntriesFlattening.cs
```diff
@@ -0,0 +1,100 @@
...
+using Toggl.Foundation.MvvmCross.Extensions;
+using Toggl.Foundation.MvvmCross.ViewModels.TimeEntriesLog;
+using Toggl.PrimeRadiant;
+using static Toggl.Foundation.MvvmCross.ViewModels.TimeEntriesLog.LogItemVisualizationIntent;
+
+namespace Toggl.Foundation.MvvmCross.Transformations
+{
+    public static class TimeEntriesFlattening
+    {
+        public static IEnumerable<CollectionSection<DaySummaryViewModel, LogItemViewModel>> Flatten(
```
kornelijepetak: The fact that you are carrying the IThreadSafePreferences across the whole flattening process is troublesome. Somehow, I feel the concerns are not really separated.

Toggl.Daneel/ViewControllers/MainViewController.cs
```diff
@@ -131,10 +131,12 @@ public override void ViewDidLoad()
             );
 
             continueTimeEntry
+                .Select(logItem => logItem.RepresentedTimeEntriesIds.First())
```
Odrakir: But why not move them to the VM already?

Toggl.Foundation.MvvmCross/Transformations/TimeEntriesCollapsing.cs
```diff
@@ -0,0 +1,139 @@
...
+using Toggl.Foundation.MvvmCross.Collections;
+using Toggl.Foundation.MvvmCross.Extensions;
+using Toggl.Foundation.MvvmCross.ViewModels.TimeEntriesLog;
+using Toggl.Multivac.Extensions;
+using Toggl.PrimeRadiant;
+using static Toggl.Foundation.MvvmCross.ViewModels.TimeEntriesLog.LogItemVisualizationIntent;
+
+namespace Toggl.Foundation.MvvmCross.Transformations
+{
+    internal sealed class TimeEntriesCollapsing
```
Odrakir: The name of this class is a little confusing for me...

Toggl.Foundation.MvvmCross/ViewModels/TimeEntriesViewModel.cs
```diff
@@ -115,21 +115,9 @@ public async Task FinilizeDelayDeleteTimeEntryIfNeeded()
...
         {
-            if (expandedGroups.Contains(groupId))
-            {
-                expandedGroups.Remove(groupId);
-            }
-            else
-            {
-                expandedGroups.Add(groupId);
-            }
+            collapsingStrategy.ToggleGroupExpansion(groupId);
```
Odrakir: It's strange that this action is not returning anything, but at the same time we rely on its 'Elements' property to trigger the observable chain again. I'm not saying we should change it now, but maybe it made more sense for it to be an `RxAction<GroupId, HashSet<GroupId>>` which would return the expanded groups and that way we could send that to the flattener instead of having a property for it.

But yeah... probably don't change it.

Toggl.Foundation.MvvmCross/ViewModels/TimeEntriesViewModel.cs
```diff
@@ -110,7 +111,9 @@ public async Task FinilizeDelayDeleteTimeEntryIfNeeded()
 
         private IEnumerable<CollectionSection<DateTimeOffset, IThreadSafeTimeEntry[]>> group(
             IEnumerable<IThreadSafeTimeEntry> timeEntries, IThreadSafePreferences preferences)
-            => TimeEntriesGrouping.Group(timeEntries, preferences);
+            => preferences.CollapseTimeEntries
```
Odrakir: Instead of removing preferences from TimeEntriesGrouping and moving that logic here I would probably have either:
- Use an instance instead of static methods so the preferences could be a property injected on construction
- Curry the static methods so the preferences were added before the execution

But I'm not going to enforce that change, we can maybe retake this in the future.

simonrozsival: Currying is weird in C#. The first option would be a good option but I think I'll leave it as it is for now.

Toggl.Foundation.MvvmCross/ViewModels/TimeEntriesViewModel.cs
simonrozsival: We wouldn't do anything with the value it returns. So it could be a `Unit` just to signal that it's finished and that something can react to it. I think it will be easier to keep it like this for the time being.

Toggl.Foundation.MvvmCross/Transformations/TimeEntriesCollapsing.cs
simonrozsival: I will find a better name.

Toggl.Daneel/ViewControllers/MainViewController.cs
simonrozsival: Let's do that tomorrow in a separate PR 👍

PR #4368, opened 2019-02-10
Log which tests get stuck

kornelijepetak:
## What's this?

📚 Let me tell you a story.

I've heard this the other day:

> But yeah, `sh build.sh` built the tests alright, but got stuck on some TimeService tests.

And then I thought, it's not a `TimeService` test that's the problem. I know it's the **last thing in the test log**, **BUT**, we are writing to the log **ONLY AFTER** the test has completed. Which means that if there's an info about a test, that test passed ✅.

Which leads to a simple conclusion that the tests that get stuck **ARE NOT** 🚫 output to the log making it hard to know what actually got stuck.

Until now. 😄 

Let me introduce you to a friend of mine: the new log `stuck` meta.

image

⚠ Note that in the context of this PR, the word `stuck` means running too long. So this basically does not **detect** `stuck` but those that take really too long. (At this point, it's 10 seconds and no test should take 10 seconds).

### Relationships
None.

## Why do we want this?
So that it's easier to know what's stuck.

## How is it done?
I just fire a timer (set by default at twice the critical time) that will output `STUCK` when the test took that long.

### Side effects
This introduces a small overhead to tests, but nothing too alarming. 
Alternatively, we could maybe use a flag to decide whether to run in `stuck-detection` mode or not and use it only when the tests are getting stuck. But considering the test runs on my computer (table below), I don't think it's an issue. 🤣 

**Current overhead influence**

|Without detection| With detection|
|-|-|
|38.47s|38.89s|

## Review hints
Not really much to say. Feel free to poke around.

## Tests
Wait, should I write tests for tests?

## :squid: Permissions
Me or Architecture.


Toggl.Multivac.Tests/Utilities/TestTooLongDetector.cs
```diff
@@ -0,0 +1,55 @@
+﻿using System;
+using System.Threading;
+using System.Threading.Tasks;
+
+class TestTooLongDetector
```
arturs105: Class not sealed

Toggl.Multivac.Tests/Utilities/LogIfTooSlowAttribute.cs
```diff
@@ -13,24 +15,41 @@ public sealed class LogIfTooSlowAttribute : BeforeAfterTestAttribute
...
         stopwatch.Start();
+
+        testStuckDetector = TestTooLongDetector.StartDetection(
+            reportAfter: critical * 2,
+            callback: () => reportStuckTest(methodUnderTest));
     }
 
     public override void After(MethodInfo methodUnderTest)
     {
+        testStuckDetector.CancelDetection();
```
amulware: Should the cancel (as well as the start) be inside the stopwatch? Maybe it's not a significant difference, just a thought.

Toggl.Multivac.Tests/Utilities/LogIfTooSlowAttribute.cs
```diff
@@ -13,24 +15,41 @@ public sealed class LogIfTooSlowAttribute : BeforeAfterTestAttribute
...
             getFormattedName(methodUnderTest.ReflectedType, toSentenceCase(methodUnderTest.Name)).Replace("`1", "");
 
         Console.ForegroundColor = getColorForTime(stopwatch.Elapsed);
         Console.WriteLine($"{stopwatch.Elapsed} - {formattedTestName}");
     }
 
+    private void reportStuckTest(MethodInfo methodUnderTest)
+    {
+        var formattedTestName =
+            getFormattedName(methodUnderTest.ReflectedType, toSentenceCase(methodUnderTest.Name)).Replace("`1", "");
```
amulware: Code duplication. DRY please.

Toggl.Multivac.Tests/Utilities/TestTooLongDetector.cs
```diff
@@ -0,0 +1,55 @@
...
+{
+    private CancellationTokenSource source;
+    private CancellationToken token;
+    private Action reportTooLongAction;
+    private object accessLock = new object();
```
amulware: readonly all the above?

Toggl.Multivac.Tests/Utilities/TestTooLongDetector.cs
```diff
@@ -0,0 +1,55 @@
...
+            if (token.IsCancellationRequested)
+                return;
+
+            reportTooLongAction();
+        }
+        finally
+        {
+            lock (accessLock)
+            {
+                source.Dispose();
```
amulware: Is the disposing really needed? What about the whole try block? The lock?

Toggl.Multivac.Tests/Utilities/LogIfTooSlowAttribute.cs
```diff
@@ -13,24 +15,41 @@ public sealed class LogIfTooSlowAttribute : BeforeAfterTestAttribute
...
+
+        Console.ForegroundColor = ConsoleColor.Red;
+        Console.WriteLine($"[ STUCK ] ... {formattedTestName}");
```
amulware: Can we align this with the completion log line, so that test names are easier to scan, and it would be easier to see if a test actually completes after 10s after all?

Toggl.Multivac.Tests/Utilities/TestTooLongDetector.cs
kornelijepetak: From the docs:

> The CancellationTokenSource class implements the IDisposable interface. You should be sure to call the CancellationTokenSource.Dispose method when you have finished using the cancellation token source to free any unmanaged resources it holds.

kornelijepetak: > What about the whole try block? The lock?

Not understanding your questions. What about them? You mean, whether they are needed?

kornelijepetak: Considering we need to dispose the cancellation token source, there is a racing condition here between calling `Cancel` and the test timeouting. Probably a rare occurence, but I'd rather this not waste our time debugging tests.

Since it's a 🐎, if `Dispose()` is called first, calling `IsCancellationRequested` will throw `ObjectDisposedException`.

Toggl.Multivac.Tests/Utilities/LogIfTooSlowAttribute.cs
kornelijepetak: Yes. 💄

Toggl.Multivac.Tests/Utilities/LogIfTooSlowAttribute.cs
kornelijepetak: It's not a big difference, but you are right. It should not be inside.

Toggl.Multivac.Tests/Utilities/LogIfTooSlowAttribute.cs
```diff
@@ -13,24 +16,42 @@ public sealed class LogIfTooSlowAttribute : BeforeAfterTestAttribute
...
 
+    private TestTooLongDetector testStuckDetector;
+
     public override void Before(MethodInfo methodUnderTest)
     {
         stopwatch.Start();
+
+        testStuckDetector = TestTooLongDetector.StartDetection(
+            reportAfter: critical * 2,
+            callback: () => reportStuckTest(methodUnderTest));
```
amulware: If you're pulling out the cancel, you should also pull out this call.

PR #4369, opened 2019-02-11
📐 Add Welcome View Margin

### LINKED ISSUE #4338: The Welcome back text on main view is broken
## 🐛 Bug Report
The text of the "Welcome back" screen is not showing text correctly:

image

## Context

Build/Branch: `develop` possibly `1.16`
Phone OS: iOS
Phone model: iPhone XS

## Expected behavior
The text is wrapped correctly.

## To reproduce
- Remove the app from your phone/simulator.
- Sign up
- Log out without creating any TEs
- Log back in
- There you go

## Suggestions
This is probably because we now use text from the `Resource.resx` file. There's a missing newline somewhere.

khoi:
## What's this?
Fix the broken text layout

### Relationships
Closes #4338

## How is it done?
Constraints the view to its superview's leading and trailing

### Side effects
None

## Review hints
None

## :squid: Permissions
🦑


khoi:
screen shot 2019-02-11 at 10 19 22

khoi:
the constraints in IB don't do anything as commented in the code. 

```
        //Spider is added in code, because IB doesn't allow adding subviews
        //to a UITableView and the spider needs to be a subview of the table
        //view so it reacts to pulling down to refresh
        private void prepareWelcomeBackViews()
        {
            // Welcome back view must be placed inside of the time entries
            // log table view below the spider so that it does not overlay
            // the spider at any time.
```

it could be done in a different way but I think this solution is the most trivial, I can set the IB's constraints to "Remove at build time" to clearly indicate that we're not using them. Would that work?

Odrakir:
Oh, I see. The view is being moved in code... weird... 🤔 

Then yes, removing them at build time would help.

PR #4379, opened 2019-02-11
Convert the SettingsActivity into a Fragment

### LINKED ISSUE #4240: Convert the SettingsActivity into a Fragment
This doesn't have to depend on anything else, the fragment can reside inside the current SettingsActivity until the bottom navigation isn't ready.

juanlaube:
## What's this?
This converts the SettingsActivity into a Fragment, we need this to use a bottom navigation bar.

### Relationships

Closes #4240

## Why do we want this?
⬇️ 🍫 

## How is it done?
GOTO "What's this?"

### Side effects
This breaks the toolbar title, I spent to much time trying to deal with this and eventually gave up, but I think that's ok for the moment since the future `MainTabBarActivity` will own the toolbar and can set the title when handling the fragments replacement when switching tabs.

## Review hints
Give it a try!

## :squid: Permissions
🦑


Toggl.Giskard/Fragments/SettingsFragment.cs
```diff
@@ -17,77 +18,77 @@
...
-              ScreenOrientation = ScreenOrientation.Portrait,
-              ConfigurationChanges = ConfigChanges.Orientation | ConfigChanges.ScreenSize)]
-    public sealed partial class SettingsActivity : ReactiveActivity<SettingsViewModel>
+    [MvxFragmentPresentation(
+        AddToBackStack = true,
+        EnterAnimation = Resource.Animation.abc_slide_in_right,
+        ExitAnimation = Resource.Animation.abc_slide_out_right,
+        PopEnterAnimation = Resource.Animation.abc_fade_in,
+        PopExitAnimation = Resource.Animation.abc_slide_out_right)]
+    public sealed partial class SettingsFragment : MvxFragment<SettingsViewModel>
```
daividssilverio: 🤔 We still don't have a `ReactiveFragment`? That's a bit disturbing.

Toggl.Giskard/Fragments/ReactiveFragment.cs
```diff
@@ -0,0 +1,57 @@
...
+using MvvmCross.ViewModels;
+
+namespace Toggl.Giskard.Fragments
+{
+    public class ReactiveFragment<TViewModel> : MvxFragment<TViewModel>
+        where TViewModel : class, IMvxViewModel
+    {
+        protected CompositeDisposable DisposeBag = new CompositeDisposable();
+
+        public override void OnCreate(Bundle savedInstanceState)
```
daividssilverio: Do you need to override anything here? The `MvxFragment` is already doing all that.

juanlaube: 🔥

PR #4380, opened 2019-02-12
Allow same project name if associated with different client.

### LINKED ISSUE #4237: Allow for same-name Projects to be created if assigned to a different Client
## ⛑️ User Support

User is attempting to create Projects with a same name but different Clients in the mobile app. The app doesn't currently allow this so it throws an error. They are usually using the desktop app which allows for this behavior so they wished they can do the same on the go.

## Expected behaviour

Creating Projects with the same name but a different Client should be allowed

## To Reproduce

1. Create a Project that already exists in the Workspace but add it to a different Client
2. It will show an error “Sorry, this name is already taken”.

khoi:
### Relationships
Closes #4237

## How is it done?
The `IsNameAlreadyTaken` now also taken the current `clientId` in to account.

### Side effects
~There will be changes to the create TE view, so that the app always show project suggest creation cell. See discussion in #4237~

I've updated the logic so that in the create TE view, the project suggestion cell is always shown

## Review hints

- The project creation cell should always be shown.
- Project creation should work properly with correct client associated to it

## :squid: Permissions
🦑


arturs105:
I think we should go for consistency with other apps and not make this harder than it needs to be.

Firstly, as [Paul suggests](https://github.com/toggl/mobileapp/issues/4237#issuecomment-462664350), we should suggest creation, no matter if a project with such a name already exists. This should be quite easy - just remove the 2 lines that check for existing project names.

Secondly, i think, we should display a different message, because "Sorry, this name is already taken" doesn't really explain, that the client name can be changed as well. This should probably be coordinated with the product team, though.

khoi:
I've updated the PR so that the project suggestion cell always show.

Toggl.Foundation.MvvmCross/ViewModels/EditProjectViewModel.cs
```diff
@@ -140,18 +141,43 @@ public override void ViewAppeared()
 
         private async Task setupNameAlreadyTakenError()
         {
-            var existingProjectNames = await dataSource.Projects
-                                        .GetAll(project => project.WorkspaceId == workspaceId)
-                                        .Select(projects => projects.Select(p => p.Name));
+            projectNamesWithClient.Clear();
+            var projectsInWorkspace = await dataSource.Projects
+                                        .GetAll(project => project.WorkspaceId == workspaceId);
```
arturs105: The indenting doesn't match the code formatting guidelines.

PR #4381, opened 2019-02-12
Delete groups of time entries with swipe gesture

simonrozsival:
## What's this?
This PR adds support for deleting groups of time entries.

### Relationships
Ref #4218 

## Why do we want this?
Grouped TEs Project

## How is it done?
- the deletion action takes array of ids as parameter
- if the user doesn't undo, it calls the deletion interactor repeatedly to delete all of the TEs (this could be done in a batch, but I don't think it's necessary)
- the VM exposes a modified observable which is used to show the snack bar with the information how many TEs would be deleted

### Side effects
This breaks Android. It will be fixed later as part of the Grouped TEs project before merging into `develop`.

## :squid: Permissions
🦑


Toggl.Foundation.MvvmCross/ViewModels/MainViewModel.cs
```diff
@@ -419,6 +417,7 @@ private IObservable<Unit> continueTimeEntry(long timeEntryId)
                 .SelectMany(prototype =>
                     interactorFactory.ContinueTimeEntry(prototype).Execute())
                 .Do(_ => onboardingStorage.SetTimeEntryContinued())
+                .Do(_ => onboardingStorage.SetTimeEntryContinued())
```
kornelijepetak: Merge issue.

Toggl.Foundation.MvvmCross/ViewModels/TimeEntriesViewModel.cs
```diff
@@ -93,20 +93,20 @@ public sealed class TimeEntriesViewModel
                 .Select(log => log.Sum(day => day.Items.Count))
                 .AsDriver(schedulerProvider);
 
-            ShouldShowUndo = showUndoSubject.AsObservable().AsDriver(schedulerProvider);
+            TimeEntriesPendingDeletion = timeEntriesPendingDeletionSubject.AsObservable().AsDriver(schedulerProvider);
         }
 
         public async Task FinilizeDelayDeleteTimeEntryIfNeeded()
```
kornelijepetak: I know it's not about your PR, but can you please fix this typo? 😄

Toggl.Foundation.MvvmCross/ViewModels/TimeEntriesViewModel.cs
```diff
@@ -120,48 +120,56 @@ private void toggleGroupExpansion(GroupId groupId)
...
                 .Take(1)
-                .SelectMany(deleteTimeEntry)
-                .Do(deletedTimeEntryId =>
+                .SelectMany(deleteTimeEntries)
+                .Do(deletedTimeEntries =>
                 {
-                    if (deletedTimeEntryId == timeEntryToDelete) // Hide bar if there isn't other TE trying to be deleted
-                        showUndoSubject.OnNext(false);
+                    // Hide bar if there isn't other TE trying to be deleted
+                    if (deletedTimeEntries == timeEntriesToDelete)
```
kornelijepetak: I'm confused here.

If `Observable.Return(timeEntries).Delay(...)` finishes first, the value given to `deleteTimeEntries` will be the array `timeEntries` parameter given to the `delayDeleteTimeEntries` function.

If the other one finishes first, then the `SelectValue(timeEntries)` will do the same.

That means that `deleteTimeEntries` will always get the same input.
Since `delayDeleteTimeEntries`'s returning observable ends with `SelectValue(timeEntries)`, it will always return the same array (by reference).

You are comparing `deletedTimeEntries == timeEntriesToDelete` here and those will always be the same making this always true.

simonrozsival: They won't be the same, if you call the delete method with a new array of IDs. Then `timeEntriesToDelete` will change (it's global state in the VM).

The idea behind the deleting or non-deleting didn't change, it's now just not comparing IDs of TEs, but references to arrays.

kornelijepetak: You are right. I't not about this one execution flow, it's about multiple calls. 🤦‍♂️

Issue #4241, opened 2019-01-17
Create the MainBottomBarActivity


daividssilverio:
Let's try to use the native version of the BottomNavigationView for this one. It might need some tweaking.

That's the equivalent of the `MvxTabBarViewController` for iOS and will be the view for the `MainTabBarViewModel`.

daividssilverio:
Things to remember:
- Update the animations (remove them) in the fragments that replaced the old activities. They won't be necessary as it doesn't make sense anymore.
- Update the `TogglPresenter` It now has to work with the `MainBottomBarActivity` and its fragments.
- Fix the toolbar in the `SettingsFragment`
- Take a look at the backstack configuration of the fragments. We basically don't have one for the bottom bar, pressing back on the bottom bar shouldn't go to the previous fragment, it should lead the user to the `MainFragment` if he's not there or close the application otherwise.
- You'll have to remove the existing `ForkNavigation` usages.
- Make sure that the snackbar isn't weird.

juanlaube:
Closed by #4397

PR #4390, opened 2019-02-12
Transform MainActivity into a Fragment

### LINKED ISSUE #4238: Convert the MainActivity into a Fragment
This doesn't have to depend on anything else, the fragment can reside in the current MainActivity until the main bottom navigation isn't ready.

juanlaube:
## What's this?
This transforms the main log from an activity into a fragment.

### Relationships

Closes #4238 

## Why do we want this?
⬇️ 🍫 

## How is it done?
See "What's this?"

### Why not another way?
There's no other way.

### Side effects
None.

## :squid: Permissions
🦑


daividssilverio:
The navigation is broken.
But that's due to the temporary host view model we've put in.
It should go on its own when #4241 is done.

PR #4397, opened 2019-02-13
Android bottom navigation bar

### LINKED ISSUE #4241: Create the MainBottomBarActivity

### LINKED ISSUE #4292: Add the calendar fragment to the MainBottomBarActivity
Not much to say.
Depends on #4291 and #4241

### LINKED ISSUE #4386: Delete host view models
Because we're moving from activities to fragments in Android to use the bottom navigation bar, the new fragments need to be hosted in an activity until we add the navbar. Some "host" view models are being created temporary and need to be deleted after #4241 is done.

juanlaube:
## What's this?
This adds the long awaited bottom navigation bar for Android!

### Relationships

Closes #4241 
Closes #4292 
Closes #4386 

## Why do we want this?
🎨 🤖  ⬇️ ⛵️ 🍫 

## How is it done?
With fragments and ❤️

### Why not another way?
Is there another way? 🤷‍♂️ 

### Side effects
Hopefully not, but this is a big change so I'm not 100% sure.

## Review hints
This needs a lot of manual testing!

## :squid: Permissions
🚫 Don't even think about it!


Toggl.Daneel/ViewControllers/MainTabBarController.cs
```diff
@@ -18,7 +18,8 @@ public class MainTabBarController : MvxTabBarViewController<MainTabBarViewModel>
         {
             { typeof(MainViewModel), "icTime" },
             { typeof(ReportsViewModel), "icReports" },
-            { typeof(CalendarViewModel), "icCalendar" }
+            { typeof(CalendarViewModel), "icCalendar" },
+            { typeof(SettingsViewModel), "icSettings" }
```
daividssilverio: Why Daneel?

juanlaube: Preparing for the future, doesn't hurt I guess 😄

Toggl.Giskard/Activities/MainTabBarActivity.cs
```diff
@@ -0,0 +1,124 @@
...
+
+namespace Toggl.Giskard.Activities
+{
+    [MvxActivityPresentation]
+    [Activity(Theme = "@style/AppTheme",
+              ScreenOrientation = ScreenOrientation.Portrait,
+              ConfigurationChanges = ConfigChanges.Orientation | ConfigChanges.ScreenSize)]
+    public sealed partial class MainTabBarActivity : ReactiveActivity<MainTabBarViewModel>
+    {
+        private MainFragment mainFragment;
```
daividssilverio: Holding hard references ain't good.
Could you use a WeakReference here instead?

Toggl.Giskard/Activities/MainTabBarActivity.cs
```diff
@@ -0,0 +1,124 @@
...
+
+            base.OnCreate(savedInstanceState);
+            SetContentView(Resource.Layout.MainTabBarActivity);
+            OverridePendingTransition(Resource.Animation.abc_fade_in, Resource.Animation.abc_fade_out);
+
+            InitializeViews();
+
+            var viewModels = ViewModel.Tabs.ToList();
+
+            mainFragment = new MainFragment();
```
daividssilverio: 🤔 Could you create the fragments lazily?

Toggl.Giskard/Activities/MainTabBarActivity.cs
```diff
@@ -0,0 +1,124 @@
...
+                    break;
+                default:
+                    throw new ArgumentException($"Unexpected tab with id {args.Item.ItemId} and title {args.Item.TitleFormatted.ToString()}");
+            }
+
+            SupportFragmentManager
+                .BeginTransaction()
+                .SetCustomAnimations(Resource.Animation.abc_fade_in, Resource.Animation.abc_fade_out)
+                .Replace(Resource.Id.CurrentTabFragmmentContainer, fragment)
+                .AddToBackStack(fragment.GetType().Name)
```
daividssilverio: This ain't working really well... I've tried changing pages and pressing back, it would go back to the previous page, instead of going back to the main fragment.

Also, I would rather handle the fragments switching manually without using the backstack.
`OnBackPressed` you can check the current fragment and go back to main if that's not the current fragment.

Toggl.Giskard/Activities/MainTabBarActivity.cs
daividssilverio: Maybe a HashMap Ids and WeakReference<Fragment> could work well.

Toggl.Giskard/Activities/MainTabBarActivity.cs
```diff
@@ -0,0 +1,124 @@
...
+            base.OnBackPressed();
+            var selectedItemId = findSelectedItemId();
+            navigationView.SelectedItemId = selectedItemId;
+        }
+
+        private void onTabSelected(object sender, BottomNavigationView.NavigationItemSelectedEventArgs args)
+        {
+            Fragment fragment;
+
+            switch (args.Item.ItemId)
```
daividssilverio: You could look in the map for the fragments here, create them if they don't exist or were collected.
Also, don't try to replace them if they are already in the container.

Toggl.Giskard/Activities/MainTabBarActivity.cs
```diff
@@ -0,0 +1,124 @@
...
+                .SetCustomAnimations(Resource.Animation.abc_fade_in, Resource.Animation.abc_fade_out)
+                .Add(Resource.Id.CurrentTabFragmmentContainer, mainFragment)
+                .Commit();
+
+            navigationView.NavigationItemSelected += onTabSelected;
+        }
+
+        public override void OnBackPressed()
+        {
+            base.OnBackPressed();
```
daividssilverio: Pressing back while in the main fragment doesn't close the application.

Toggl.Giskard/Activities/MainTabBarActivity.cs
```diff
@@ -0,0 +1,124 @@
...
+            if (calendarFragment.IsResumed)
+                return Resource.Id.MainTabCalendarItem;
+            if (settingsFragment.IsResumed)
+                return Resource.Id.MainTabSettinsItem;
+            return Resource.Id.MainTabTimerItem;
+        }
+
+        public void SetupRatingViewVisibility(bool isVisible)
+        {
+            mainFragment.SetupRatingViewVisibility(isVisible);
```
daividssilverio: I guess we would have to change the strategy a little bit here. Maybe set a flag for when the main fragment isn't the  current fragment and update it when it becomes the current.

Toggl.Giskard/Activities/MainTabBarActivity.cs
```diff
@@ -0,0 +1,124 @@
...
+        }
+
+        internal void ToggleReportsCalendarState(bool forceHide)
+        {
+            reportsFragment.ToggleCalendarState(forceHide);
```
daividssilverio: Same idea here.

Toggl.Giskard/Activities/MainTabBarActivity.cs
```diff
@@ -0,0 +1,124 @@
...
+                case Resource.Id.MainTabSettinsItem:
+                    fragment = settingsFragment;
+                    break;
+                default:
+                    throw new ArgumentException($"Unexpected tab with id {args.Item.ItemId} and title {args.Item.TitleFormatted.ToString()}");
+            }
+
+            SupportFragmentManager
+                .BeginTransaction()
+                .SetCustomAnimations(Resource.Animation.abc_fade_in, Resource.Animation.abc_fade_out)
```
daividssilverio: The animation is not looking good on my S8, it flashes weirdly... 🤔 Maybe leave it without animations or make it faster.

Toggl.Giskard/Fragments/SettingsFragment.cs
```diff
@@ -5,33 +5,29 @@
...
-        PopEnterAnimation = Resource.Animation.abc_fade_in,
-        PopExitAnimation = Resource.Animation.abc_slide_out_right)]
     public sealed partial class SettingsFragment : ReactiveFragment<SettingsViewModel>
     {
         public override View OnCreateView(LayoutInflater inflater, ViewGroup container, Bundle savedInstanceState)
         {
             var view = inflater.Inflate(Resource.Layout.SettingsFragment, container, false);
 
             InitializeViews(view);
+            setupToolbar();
```
daividssilverio: `OnCreateView` doesn't seem to be the best place to set the toolbar, could you try setting it in the `OnAttach` or `OnResume`.

daividssilverio: For the other fragments as well.

Toggl.Giskard/Activities/MainTabBarActivity.cs
juanlaube: :fire:

Toggl.Giskard/Fragments/SettingsFragment.cs
juanlaube: This is ok since the views are created again on each tab selection.

Toggl.Giskard/Activities/MainTabBarActivity.cs
```diff
@@ -0,0 +1,125 @@
...
+
+namespace Toggl.Giskard.Activities
+{
+    [MvxActivityPresentation]
+    [Activity(Theme = "@style/AppTheme",
+              ScreenOrientation = ScreenOrientation.Portrait,
+              ConfigurationChanges = ConfigChanges.Orientation | ConfigChanges.ScreenSize)]
+    public sealed partial class MainTabBarActivity : ReactiveActivity<MainTabBarViewModel>
+    {
+        private Dictionary<int, Fragment> fragments = new Dictionary<int, Fragment>();
```
heytherewill:
```suggestion
        private readonly Dictionary<int, Fragment> fragments = new Dictionary<int, Fragment>();
```

Toggl.Giskard/Activities/MainTabBarActivity.cs
```diff
@@ -0,0 +1,125 @@
...
+        private List<MvxViewModel> viewModels;
```
heytherewill: Why keep a local list?

Toggl.Giskard/Activities/MainTabBarActivity.cs
```diff
@@ -0,0 +1,125 @@
...
+            var fragment = getFragment(Resource.Id.MainTabTimerItem);
+            SupportFragmentManager
+                .BeginTransaction()
+                .Add(Resource.Id.CurrentTabFragmmentContainer, fragment)
+                .Commit();
+
+            navigationView.NavigationItemSelected += onTabSelected;
+        }
+
+        private Fragment getFragment(int itemId)
```
heytherewill: `getFragmentWithCache`?

Toggl.Giskard/Activities/MainTabBarActivity.cs
```diff
@@ -0,0 +1,125 @@
...
+
+            base.OnCreate(savedInstanceState);
+            SetContentView(Resource.Layout.MainTabBarActivity);
+            OverridePendingTransition(Resource.Animation.abc_fade_in, Resource.Animation.abc_fade_out);
+
+            viewModels = ViewModel.Tabs.ToList();
+            InitializeViews();
+
+            var fragment = getFragment(Resource.Id.MainTabTimerItem);
+            SupportFragmentManager
```
heytherewill: Could this be in a `showFragment` method?

Toggl.Giskard/Activities/MainTabBarActivity.cs
```diff
@@ -0,0 +1,125 @@
...
+                .BeginTransaction()
+                .Add(Resource.Id.CurrentTabFragmmentContainer, fragment)
+                .Commit();
+
+            navigationView.NavigationItemSelected += onTabSelected;
```
heytherewill: Please make a reactive extension to the `BottomNavigationView` type so this works as an observable, binding the removal of the event to disposing the dispose bag (you forgot to unsubscribe to this event, BTW)

Toggl.Giskard/Activities/MainTabBarActivity.cs
```diff
@@ -0,0 +1,125 @@
...
+            }
+            fragments[itemId] = fragment;
+            return fragment;
+        }
+
+        public override void OnBackPressed()
+        {
+            if (navigationView.SelectedItemId == Resource.Id.MainTabTimerItem)
+            {
+                base.OnBackPressed();
```
heytherewill: Early return instead of nesting

Toggl.Giskard/Activities/MainTabBarActivity.cs
juanlaube: I like `getCachedFragment` more, what do you think?

Toggl.Giskard/Activities/MainTabBarActivity.cs
juanlaube: This was needed because the `getCachedFragment` method accessed each tab by index, I made the `Tabs` property an `IList` in the view model instead of doing this.

juanlaube:
I made some changes:

* improved how we show/hide fragments when switching tabs, now each view is created only once, making things faster and keeping the state correctly, this solves some problems @daividssilverio mentioned earlier.

* all previously existing the fragments are removed now when the activity is created, this should solve the problem with "Don't keep activities" and tombstone scenarios.

Regarding the notifications not working as expected, this is working for me, @daividssilverio can you clarify what's not working exactly?
We may still need a placeholder/loading view when switching between tabs the first time, in particular reports take some time to load. This should be in a different issue though and will require some designs.

Toggl.Giskard/Activities/MainTabBarActivity.cs
heytherewill: me gusta

daividssilverio:
@juanlaube . Enable `Running timer` and `Stopped timer` notifications from the Settings page while having a time entry running. The `Stopped timer` notification will appear instead of the `Running timer`, until you go to the Timer page.

daividssilverio:
As @heytherewill suggested and you commented. Please go with the progress bar meanwhile to soften the problem. I'll create an issue for that. I've created another one to create the placeholders that will require designs #4423

daividssilverio:
Hey, using the new fragment strategy fixes the notifications problem o/
This is already looking good to me.

Toggl.Giskard/Extensions/FragmentManagerExtensions.cs
```diff
@@ -0,0 +1,16 @@
...
+using Toggl.Multivac.Extensions;
+
+namespace Toggl.Giskard.Extensions
+{
+    public static class FragmentManagerExtensions
+    {
+        public static void RemoveAllFragments(this FragmentManager fragmentManager)
+        {
+            var transaction = fragmentManager.BeginTransaction();
+            fragmentManager.Fragments.ForEach(fragment => transaction.Remove(fragment));
```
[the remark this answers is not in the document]
juanlaube: I can't because `transaction.Remove` returns another transaction! While this worked fine, it was semantically wrong though.

Toggl.Giskard/Extensions/FragmentManagerExtensions.cs
```diff
@@ -0,0 +1,15 @@
...
+using Toggl.Multivac.Extensions;
+
+namespace Toggl.Giskard.Extensions
+{
+    public static class FragmentManagerExtensions
+    {
+        public static void RemoveAllFragments(this FragmentManager fragmentManager)
+        {
+            var transaction = fragmentManager.BeginTransaction();
+            fragmentManager.Fragments.ForEach(fragment => transaction = transaction.Remove(fragment));
```
heytherewill: If you change this to a more functional way:

```csharp
fragmentManager.Fragments
    .Aggreggate(fragmentManager.BeginTransaction(), (transaction, fragment) => transaction.Remove(fragment))
    .Commit();
```

We can merge it :)

juanlaube: Why I didn't think of this first? 🙈

PR #4407, opened 2019-02-14
Refactor EditTimeEntryViewModel and implement edit of grouped items

### LINKED ISSUE #4219: Update EditTimeEntryViewModel to support Grouped Time Entries

kornelijepetak:
## What's this?
Upgrade of the `EditTimeEntryViewModel`.

* Supports editing groups
* RxFactored
* Supports wheel out of the box

⚠ It is important to read review hints.

_Also, I have noticed only after the push that I've included changes to iOS and Android dialog services as well. You can choose to ignore those, or treat them as part of the PR. If you think it's too confusing, let me know and I will remove it from the PR and leave only the change in Foundation._

### Relationships
Closes #4219 

## Why do we want this?
Because grouping.

## How is it done?
Slowly and carefully.

### Why not another way?
Suggest if you know better.

### Side effects
So far none. Except, some parts that were used only by one platform was removed to be done in the platform itself (handling focus of the description field).

## Review hints
First of all, ignore all the Giskard and Daneel stuff. This is just so the project can be built. Those parts are done in other PRs.

Go commit by commit until you reach the EditView one.
For the VM itself, I believe it's easier to follow piece by piece of the new VM because it's mostly done from scratch. Just open the file and check whether it all makes sense to you. Comparing differences between the new one and old one is not a reasonable thing to do in this case.
There are some commits after the VM, check those one by one as well.

## Tests
Tests are completely rewritten because of the large amount of changes in the view model. Please review the tests one by one instead of comparing them to the previous one.

## :squid: Permissions
Only me.


Toggl.Foundation.MvvmCross/ViewModels/EditTimeEntryViewModel.cs
```diff
@@ -272,391 +165,268 @@ public override async Task Initialize()
             stopwatchFromMainLog = stopwatchProvider.Get(MeasuredOperation.EditTimeEntryFromMainLog);
             stopwatchProvider.Remove(MeasuredOperation.EditTimeEntryFromMainLog);
 
-            var timeEntry = await interactorFactory.GetTimeEntryById(Id).Execute();
+            var timeEntries = await interactorFactory.GetMultipleTimeEntriesById(TimeEntryIds).Execute();
+            var timeEntry = timeEntries.First();
```
simonrozsival: Do we just assume that the TEs always have the same properties (except id, star time, duration)?

Toggl.Foundation.MvvmCross/ViewModels/EditTimeEntryViewModel.cs
```diff
@@ -272,391 +165,268 @@ public override async Task Initialize()
...
+            tagsSubject.OnNext(timeEntry.Tags?.ToImmutableList() ?? ImmutableList<IThreadSafeTag>.Empty);
 
-            await close();
-            return true;
-        }
+            isInaccessibleSubject.OnNext(timeEntry.IsInaccessible);
 
-        private Task close()
-            => navigationService.Close(this);
+            syncErrorMessageSubject.OnNext(
```
simonrozsival: In this case you should use `timeEntries.FirstOrDefault(te =-> te.IsInaccessible || te.LastSyncErrorMessage != null)` or something like that instead of looking just at the very first TE in the group.

Toggl.Foundation.MvvmCross/ViewModels/EditTimeEntryViewModel.cs
```diff
@@ -272,391 +165,268 @@ public override async Task Initialize()
...
+        private void save()
         {
-            hasProjectSubject.OnNext(!string.IsNullOrWhiteSpace(Project));
+            OnboardingStorage.EditedTimeEntry();
+
+            var dto = new EditTimeEntryDto
+            {
+                Id = TimeEntryIds.First(),
+                Description = Description.Value?.Trim() ?? string.Empty,
+                StartTime = startTimeSubject.Value,
```
simonrozsival: Will this set the same `StartTime` to all of the TEs in the group? Or how does it work? The same applies to `StopTime`.

Toggl.Foundation/Extensions/ThreadSafeExtensions.cs
```diff
@@ -29,5 +30,10 @@ public static string DisplayColor(this IThreadSafeProject project)
                     return project.Color ?? "";
             }
         }
+
+        public static TimeSpan? TimeSpanDuration(this IThreadSafeTimeEntry timeEntry)
```
simonrozsival: I don't like this extension method somehow. Maybe it should be just called somehow differently... but for now I think it's unnecessary.

Toggl.Foundation/Interactors/TimeEntry/DeleteMultipleTimeEntriesInteractor.cs
```diff
@@ -0,0 +1,42 @@
...
+            this.dataSource = dataSource;
+            this.timeService = timeService;
+            this.interactorFactory = interactorFactory;
+        }
+
+        public IObservable<Unit> Execute()
+            => interactorFactory.GetMultipleTimeEntriesById(ids).Execute()
+                .SelectMany(CommonFunctions.Identity)
+                .Select(te => te.Id)
+                .SelectMany(id => interactorFactory.DeleteTimeEntry(id).Execute())
```
simonrozsival: I think it would be better to use the `DeleteAll` method of the data source.

Toggl.Foundation/Interactors/TimeEntry/UpdateMultipleTimeEntriesInteractor.cs
```diff
@@ -0,0 +1,60 @@
...
+                .Select(convertToDto)
+                .SelectMany(dto => interactorFactory.UpdateTimeEntry(dto).Execute())
+                .ToList();
+
+        private EditTimeEntryDto convertToDto(IThreadSafeTimeEntry timeEntry)
+        {
+            var updatedDto = dto;
+
+            updatedDto.Id = timeEntry.Id;
+            updatedDto.StartTime = timeEntry.Start;
```
simonrozsival: I can see an answer to my previous question. But now I have a different question: how do we actually change the start time and duration of time entries when you are editing just a single time entry? Won't this keep the original value?

I think it would be much easier and clearer if the DTOs were created in the VM and passed to the interactor.

Toggl.Giskard/Services/DialogServiceAndroid.cs
```diff
@@ -102,6 +102,8 @@ public IObservable<bool> ConfirmDestructiveAction(ActionType type)
                     return Confirm(null, Resources.DiscardEditingChanges, Resources.Discard, Resources.ContinueEditing);
                 case ActionType.DeleteExistingTimeEntry:
                     return Confirm(null, Resources.DeleteThisTimeEntry, Resources.Delete, Resources.Cancel);
+                case ActionType.DeleteMultipleExistingTimeEntries:
+                    return Confirm(null, string.Format(Resources.DeleteMultipleTimeEntries, formatArguments), Resources.Delete, Resources.Cancel);
```
simonrozsival: No. This is so ugly. No. There must be another way..

Toggl.Multivac/Extensions/EnumerableExtensions.cs
```diff
@@ -93,5 +93,19 @@ public static TimeSpan Sum<T>(this IEnumerable<T> collection, Func<T, TimeSpan?>
         public static bool ContainsExactlyAll<T>(this IReadOnlyList<T> collection, IReadOnlyList<T> otherCollection)
             => collection.Count == otherCollection.Count
                && collection.Except(otherCollection).None();
+
+        public static bool SetEquals<T>(this IEnumerable<T> first, IEnumerable<T> second, IEqualityComparer<T> comparer = null)
```
simonrozsival: The `Set` at the beginning is confusing. At first I thought this is some kind of a setter. Use a different name please. `ContainsSameItems` or something more explicit?

Toggl.Daneel/Services/DialogServiceIos.cs
```diff
@@ -146,18 +150,20 @@ public IObservable<T> Select<T>(string title, IEnumerable<(string ItemName, T It
             });
         }
 
-        private (string, string) selectTextByType(ActionType type)
+        private (string Title, string ConfirmButtonText, string CancelButtonText) selectTextByType(ActionType type)
```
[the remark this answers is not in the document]
kornelijepetak: When I realized I have to do this, a bit of me died inside.
However, it's not my design choice to make this method like this in the first place, and for now I don't want to deal with restructuring our app while working on the 🔟. 

Unfortunatelly, I refuse to change this in this PR. 

But I agree, texts should not be generated here, but injected from outside.

Toggl.Foundation.MvvmCross/ViewModels/EditTimeEntryViewModel.Items.cs
```diff
@@ -0,0 +1,123 @@
...
+using Toggl.Foundation.MvvmCross.Services;
+using Toggl.Foundation.Services;
+using Toggl.Multivac;
+using Toggl.Multivac.Extensions;
+using Toggl.Multivac.Extensions.Reactive;
+using Toggl.PrimeRadiant.Settings;
+
+namespace Toggl.Foundation.MvvmCross.ViewModels
+{
+    public sealed partial class EditTimeEntryViewModel
```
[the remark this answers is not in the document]
kornelijepetak: In this specific instance I agree and I've done this only to make the review process easier because these are not important.

However, logically grouping some behavior (such as onboarding) SHOULD be done in partial classes so that merge conflicts don't become a nightmare when people are working on the same view models at the same time. It happened in the past A LOT.

Ideally, those should be separate classes to adhere more to the single-responsibility-principle. But until we start doing those things properly, partial classes do help with source control.

Other than that, partial classes are NOTHING like C++. The compiler's job is exactly the same (unlike C++).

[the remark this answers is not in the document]
kornelijepetak: I will pull it back when reviews are done.

Toggl.Foundation.MvvmCross/ViewModels/EditTimeEntryViewModel.Items.cs
```diff
@@ -0,0 +1,123 @@
...
+        private readonly IStopwatchProvider stopwatchProvider;
+        private readonly ISchedulerProvider schedulerProvider;
+        private readonly IRxActionFactory actionFactory;
+        public IOnboardingStorage OnboardingStorage { get; private set; }
+
+        // Performance
+        private IStopwatch stopwatchFromCalendar;
+        private IStopwatch stopwatchFromMainLog;
+
+        // IDs
```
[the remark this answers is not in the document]
kornelijepetak: Collateral damage.

Toggl.Foundation/Extensions/ThreadSafeExtensions.cs
kornelijepetak: The call-sites are complicated if you have to do this all over the place. Besides, DRY.
Do you have any suggestions on how to improve both?

Toggl.Foundation/Interactors/TimeEntry/DeleteMultipleTimeEntriesInteractor.cs
kornelijepetak: I will switch to that.

Toggl.Foundation/Interactors/TimeEntry/UpdateMultipleTimeEntriesInteractor.cs
kornelijepetak: Maybe. But I don't think there's much issue with this.
If the edit view is editing a single TE, then the other interactor gets used which processes it without changing anything. The edit view simply uses the first element of the array to update the Dto in that case.

But for consistency sake, I can make a `DTO[]` parameter instead of `long[]` so that both interactors are actually working with DTOs.

Toggl.Giskard/Services/DialogServiceAndroid.cs
kornelijepetak: I think the whole idea of this relying on ActionType here at this place is wrong. But like I said in another comment, I will not change it as part of this PR.

We can create an issue to improve our dialog services.

Toggl.Multivac/Extensions/EnumerableExtensions.cs
kornelijepetak: The reason for this name is the fact that `HashSet<T>` uses `SetEquals` to do whatever I'm doing here. And is also consistent with `ReferenceEquals` and `SequenceEquals`. 

For consistency's sake, I'd rather leave it as it is.

Would `UnorderedCollectionEquals` work better for you? Somehow I want to retain the `<Criterion>Equals` name.

Toggl.Giskard/Services/DialogServiceAndroid.cs
simonrozsival: Create an issue then please.

simonrozsival:
I will be on vacation for the next two weeks. Don't wait for my approval and go ahead once you have from the other reviewer(s).

Toggl.Daneel/Services/DialogServiceIos.cs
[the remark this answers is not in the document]
kornelijepetak: This will be refactored as #4419

Toggl.Foundation/Extensions/ThreadSafeExtensions.cs
kornelijepetak: Since you didn't give any suggestions on how to improve this, I refuse executing this conditional in every place it's needed.

Toggl.Giskard/Services/DialogServiceAndroid.cs
kornelijepetak: This will be refactored as #4419

Toggl.Foundation.MvvmCross/ViewModels/EditTimeEntryViewModel.cs
kornelijepetak: You seem to have answered to this question along the way, according to your comment :)

Toggl.Foundation.MvvmCross/ViewModels/EditTimeEntryViewModel.cs
kornelijepetak: We don't assume. That's how it is. They share the same properties that are used for grouping. Only ID, and times are different.

Toggl.Foundation.MvvmCross/ViewModels/EditTimeEntryViewModel.cs
```diff
@@ -248,343 +230,332 @@ public override async Task Initialize()
...
+            => duration.HasValue ? start + duration : null;
 
-        private async Task selectDuration()
+        private static string ellipsize(IThreadSafeTag tag)
         {
-            analyticsService.EditViewTapped.Track(EditViewTapSource.Duration);
-            await editDuration(true);
-        }
+            var tagLength = tag.Name.LengthInGraphemes();
+            if (tagLength <= MaxTagLength)
```
simonrozsival: Now that I look at this I see a bug :D I think it has been like this all along and we don't have to fix it here, but this might be a good opportunity to fix it.

It should be: `if(tagLength <= MaxTagLength + 3)` ... or some other correction, because the result of this method could be a string which is longer than it was before and some of the characters (1 - 2 last ones) were replaced with dots.

simonrozsival: Also it should be an extension method so that we can simplify this YUGE file.

Toggl.Foundation/Interactors/TimeEntry/DeleteMultipleTimeEntriesInteractor.cs
```diff
@@ -0,0 +1,40 @@
...
+using Toggl.Multivac;
+using Toggl.Multivac.Extensions;
+using Toggl.PrimeRadiant.Models;
+
+namespace Toggl.Foundation.Interactors
+{
+    internal sealed class DeleteMultipleTimeEntriesInteractor : IInteractor<IObservable<Unit>>
+    {
+        private readonly long[] ids;
+        private readonly ITimeService timeService;
```
simonrozsival: It seems it's not needed in the end, 🔥 it.

Toggl.Foundation/Interactors/TimeEntry/UpdateMultipleTimeEntriesInteractor.cs
```diff
@@ -0,0 +1,45 @@
...
+using Toggl.Foundation.DTOs;
+using Toggl.Foundation.Models.Interfaces;
+using Toggl.Multivac;
+using Toggl.Multivac.Extensions;
+
+namespace Toggl.Foundation.Interactors
+{
+    internal class UpdateMultipleTimeEntriesInteractor : IInteractor<IObservable<IEnumerable<IThreadSafeTimeEntry>>>
+    {
+        private readonly ITimeService timeService;
```
simonrozsival: 🔥

Toggl.Foundation/Interactors/TimeEntry/UpdateMultipleTimeEntriesInteractor.cs
```diff
@@ -0,0 +1,45 @@
...
+        private readonly ITogglDataSource dataSource;
```
simonrozsival: 🔥

Toggl.Foundation.MvvmCross/ViewModels/EditTimeEntryViewModel.cs
```diff
@@ -199,46 +125,102 @@ public TimeSpan Duration
...
-            ProjectTaskOrClientChanged = projectTaskClientSubject
-                .AsObservable();
-
-            bool canExecute()
-                => !IsInaccessible;
+            this.schedulerProvider = schedulerProvider;
+            this.actionFactory = actionFactory;
+            OnboardingStorage = onboardingStorage;
+
+            workspaceIdSubject = new BehaviorSubject<long>(0);
```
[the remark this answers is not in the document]
kornelijepetak: I don't see this as an issue. The workspace ID is correctly set from the retrieved time entry in the `Initialize` method which should happen before VC/activity/fragment gets control. If that does not happen, the app will break on other things not being prepared anyway.

But I don't mind changing it into what you suggested, to minimize the possibility of the issue.

Toggl.Foundation.MvvmCross/ViewModels/EditTimeEntryViewModel.cs
kornelijepetak: I see. I will revisit this and see if it's trivial to solve right now. Nice catch!

Toggl.Foundation.MvvmCross/ViewModels/EditTimeEntryViewModel.cs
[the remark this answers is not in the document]
kornelijepetak: Every observable that depends on that subject is already a driver so it already has a `.Replay(1)`.

Toggl.Foundation.MvvmCross/ViewModels/EditTimeEntryViewModel.cs
kornelijepetak: Wait, I don't follow you after all. The input to this method is never already ellipsized tag name. It's the real tag name which is then being converted into an ellipsized form. If you still think this is wrong, please create an issue and explain in more details. Because I'm not sure in the end what's wrong.

simonrozsival: I will create a separate issue.

PR #4414, opened 2019-02-15
Run ViewDestroy lifecycle event on modals

### LINKED ISSUE #4410: ViewModels are not getting properly destroyed
This happens in iOS, I'm guessing Android doesn't have the same problem.

If I go into `StartTimeEntryViewModel` and add a `Debug` operator...

```c#
timeService.CurrentDateTimeObservable
    .Debug("time")
    .Where(_ => isRunning)
    .Subscribe(currentTime => DisplayedTime = currentTime - StartTime)
    .DisposedBy(disposeBag);
```

I can see that when I dismiss the modal view the subscription is still alive and running. If I open the view again another subscription will be created and never disposed.

The ViewModel includes this code

```c#
public override void ViewDestroy(bool viewFinishing)
{
    base.ViewDestroy(viewFinishing);
    disposeBag?.Dispose();
}
```

But the problem is this code is never called. Because that method is called only from `DidMoveToParentViewController` and Mvx works with the assumption that `DidMoveToParentViewController` is called whenever a modal view controller is presented or dismissed, which is not true. This will only work for child view controllers (custom or through Navigation or TabBar controllers)

I think this is very serious, not only for the implications of memory consumption, but also because of the side effects of having a subscription alive when we don't really want it to be.

heytherewill:
## What's this?
This fixes MvvmCross not running the full lifecycle events for ViewModels when the view is displayed as a Modal

### Relationships
Closes #4410

## Why do we want this?
Avoiding nasty leaks

## How is it done?
We are basically intercepting the call that closes modals and calling the method ourselves

### Why not another way?
This is straightforward enough

### Side effects
This might affect some other modals (depending on whether they're [`UINavigationController`](https://github.com/MvvmCross/MvvmCross/blob/develop/MvvmCross/Platforms/Ios/Presenters/MvxIosViewPresenter.cs#L498)s or not) and make the method be called twice. It didn't happen in any of the modals I've tested, but I could have forgotten others, so please do check

Also bear in mind that, as of today, we're only using this method for disposing subscriptions, so even if it does run twice, it'll do no harm.

## :squid: Permissions
Anyone, after @Odrakir reviews


Odrakir:
Quick and dirty way to fix this. Still I think we should (someday) work on a custom solution without Mvx.

Why are you saying a `UINavigationController` modal would get called twice? Are you sure about that?

PR #4416, opened 2019-02-18
Track sync duration

### LINKED ISSUE #3716: Measure how long Sync takes
Sync can be slow for some users. We have no idea why, or even how slow it is.

We should start measuring it first and foremost. It's not as easy as adding a StopWatch around some code though since we use the complicated state machine.

This requires some thinking on how to measure and what exactly to measure. The focus should be on measuring what the user actually experiences in some sense, but also measuring so that we have an idea what we can improve.

Something perhaps important is being aware of how often we end up in retry loops since they will significantly increase how long syncing takes.

Both platforms.

dshr:
## What's this?
This PR adds some tracking to our regular sync process, along with tracking the Leaky Bucket overflows, all for the sake of understanding why our users report slow sync times.

### Relationships
Closes #3716

## Why do we want this?
Happy users

## How is it done?
By using the analytics service to track the overflows and Firebase Performance Monitoring to track our sync times.

### Why not another way?
We could've used only analytics for this and then create a funnel in Firebase, but that was a bit too complex.

### Side effects
None.

## Tests
Syncing tests: https://app.bitrise.io/build/9cff2e6f621744b6 - do not merge this branch unless these tests are passing.

## :squid: Permissions
All yours!


juanlaube:
I'm curious about how to read the `LeakyBucketOverflow` events in AppCenter, if we have 1000 of those, how do we know if that's too much? 🤔

dshr:
> I'm curious about how to read the `LeakyBucketOverflow` events in AppCenter, if we have 1000 of those, how do we know if that's too much? 🤔

We can cross-reference the amount of events with the amount of users affected, which will tell us if there's a handful of users sending lots of requests, or everyone is going slightly over.

PR #4438, opened 2019-02-21
Support Time Entries Groups in Edit TE View

### LINKED ISSUE #4221: Update Android Edit Screen to support grouped TEs
🏆 Update the edit screen on Android to support changes in the `EditTimeEntryViewModel` as defined in the #4219.

Make sure that two things are different for groups:
* [x] Remove time editing controls
* [x] Show how many entries would be affected by the edit
* [x] Update the delete label to include the number of affected TEs (something like `Delete all 12 entries in the group`)

@maritoggl, @lizaPlotnikova, please specify all the relevant copies for the changed UI elements.

kornelijepetak:
## What's this?
Implements Android part of Edit view that supports Time entry grouping.

### Relationships
Closes #4221

This branches off from #4407 because it is based on the new viewmodel. It requires #4407 to be merged first.

## How is it done?
* Create the UI from scratch to use more flat UI hierarchy (w/ constraint layout) instead of nesting views.
* Utilizing and binding to the new view model (#4407)

### Why not another way?
`:kornelije_no:`

### Side effects
Should be faster than before.

## Review hints
If you go commit by commit, make sure to skip the merges. Otherwise, should be self-explanatory.

## :squid: Permissions
🔟, which is just me.


Toggl.Giskard/ViewHolders/StringViewHolder.cs
```diff
@@ -0,0 +1,39 @@
...
+
+namespace Toggl.Giskard.ViewHolders
+{
+    /// <summary>
+    /// Represents a view holder for all cell types that have only one textual piece of UI information
+    /// </summary>
+    /// <remarks>
+    /// The inflated cell has to provide a single UI element with android:id="@+id/Text".
+    /// </remarks>
+    public class StringViewHolder : BaseRecyclerViewHolder<string>
```
heytherewill: I have created a very similar thing in the Start Time entry PR, but it is more generic and serves more purposes. Whoever merges first should change it for the other :)

kornelijepetak: I will take care of this before merging into develop.

kornelijepetak: @heytherewill said it's ok if I merge this now.

simonrozsival:
The original base branch `kornelije/edit-view-transformer` was already merged and I believe the correct base branch should be `project/grouped-time-entries`. There are too many complex conflicts thought. I will leave it for you to resolve on Monday, @kornelijepetak.

PR #4440, opened 2019-02-21
🐛Fix Calendar crash when creating TimeEntries

### LINKED ISSUE #4304: Null ref in CalendarCollectionViewSource.LayoutAttributesForItemAtIndexPath
## 💥 Crash Report
App crashes in the calendar view. I have no idea what exactly causes the problem or how to reproduce it.

### 👥 Impact
The impact is very low, the crash happened 27x for 22 users. That makes only 0.7% of users affected by this crash. It is still the most frequent crash in 1.16 at the moment.

### 🔗 Context
I haven't seen a similar crash before.

There are some similar crashes which might be related somehow:
- https://appcenter.ms/orgs/Toggl/apps/Toggl-iOS/crashes/errors/3726594500u/overview
- https://appcenter.ms/orgs/Toggl/apps/Toggl-iOS/crashes/errors/2989517920u/overview

### 🛰 Link
https://appcenter.ms/orgs/Toggl/apps/Toggl-iOS/crashes/errors/2108006625u/overview

`SIGABRT: Object reference not set to an instance of an object`
```cs
    CalendarCollectionViewSource.LayoutAttributesForItemAtIndexPath (Foundation.NSIndexPath indexPath)
    CalendarCollectionViewLayout.frameForEditingHour (Foundation.NSIndexPath indexPath)
    CalendarCollectionViewLayout.LayoutAttributesForSupplementaryView (Foundation.NSString kind, Foundation.NSIndexPath indexPath)
```

sofokli1:
## What's this?
This fixes a crash that has been happening for a while but we were not sure why in the calendar view when creating TEs with long press. I could reproduce this quite often when creating more than one TE **on device** (mine is a iphone x)

### Relationships
Closes #4304

## Why do we want this?
Go away crashes

## How is it done?
The main reason of the crash was the request from the calendar to layout the new view but the variable holding the index of the edited view was nullified when the long press gesture recogniser ended. Even though it makes sense to null the var the collection view was not done with the presenting part so a race condition was present thus not happening always. An other issue was that the long press gesture for editing a TE and the other long press gesture for creating a the were running at the same time. So the create long press gesture created a TE and the other long press gesture for edit was finding a TE at that location (the one that was just created) thus conflicting with each other. I added some code to stop touches to one of them depending on what item is at the start location of the touches.

### Why not another way?
O am open for ideas

### Side effects
for now only good things

## Review hints
Try **on device** on develop and as described at the top and then in this branch.

## :squid: Permissions
:squid:


Toggl.Daneel/Views/Calendar/CalendarCollectionViewCreateFromSpanHelper.cs
```diff
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Diagnostics;
```
juanlaube: 👀

Toggl.Daneel/Views/Calendar/CalendarCollectionViewEditItemHelper.cs
```diff
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Diagnostics;
```
juanlaube: 👀

PR #4441, opened 2019-02-21
Use older api for Lollipop support

### LINKED ISSUE #4439: Method missing during WheelDurationInput.initialize()
## 🐛 Bug Report
The app seems to be crashing during the Wheel initialization

## Context
Build/Branch: 1.11
Phone OS: Various
Phone model: Various

## Expected behavior
The method should be there 🤔 

## To reproduce
Try checking the [AppCenter Reports](https://appcenter.ms/orgs/Toggl/apps/Toggl-Android/crashes/errors/2742338112u/overview) and seeing the steps

## Suggestions
This seems to be a linker issue. Maybe just adding it to `LinkerPleaseInclude` might suffice

heytherewill:
## What's this?
This makes use of older apis to avoid crashes

### Relationships
Closes #4439

## Why do we want this?
Crashes are bad

## Review hints
Be @daividssilverio  or @kornelijepetak 

## :squid: Permissions
Me


daividssilverio:
Just to keep the history:
`SetFocusable(ViewFocusability.NotFocusable);` and
`Focusable = false;`
are not the same thing, but it turns out we don't need the former.

`SetFocusable(ViewFocusability.NotFocusable);` is the same as
```
Focusable = false;
FocusableInTouchMode = false;
```

PR #4461, opened 2019-02-25
Show calendar onboarding

### LINKED ISSUE #4458: Android Calendar Get Started: Bindings
This issue was created to complete what was missing on #4287
Now that we have bottom bar and the CalendarFragment in place, we can make all the bindings and related logic.

juanlaube:
## What's this?
This shows the calendar onboarding and binds the actions to get permissions and select the calendars.

### Relationships

Closes #4458 

## Why do we want this?
We want users to get started with the calendar, duh

## How is it done?
The view stub makes the onboarding view lazy, the rest is just bindings.

### Side effects
None.

## Review hints
Try it!

## :squid: Permissions
🦑


Toggl.Giskard/Fragments/CalendarFragment.cs
```diff
@@ -11,7 +15,42 @@ public override View OnCreateView(LayoutInflater inflater, ViewGroup container,
...
+            {
+                if (onboardingView == null)
+                {
+                    initializeOnboardingView();
+                }
+                onboardingView.Visibility = ViewStates.Visible;
+            }
+            else
+            {
+                onboardingView.Visibility = ViewStates.Gone;
```
daividssilverio: Better have another check here, the `onboardingView` will not be initialized when the first value emitted by `ViewModel.ShouldShowOnboarding` is false. It's being swallowed by the RxAction though.

PR #4467, opened 2019-02-26
🚧 Fix month name in Reports only updating when stoping scroll

### LINKED ISSUE #4395: Month name not changing while scrolling in reports calendar selection view
## 🐛 Bug Report
While scrolling from month to month in the reports calendar selection view the moth name ot top does not change until we stop scrolling

## Context
Build/Branch: `develop`
Phone OS: ios
Phone model: x

## Expected behavior
As we scroll fast between months it would be very helpful to have the month name change as we scroll since we have to stop scrolling in order to check which month we are at.

## To reproduce
Follow the steps described at the top

## Suggestions
Some slide animation with fade out for the month that is being removed and fade in and slide in place for the month coming to view would make it look nice.

sofokli1:
## What's this?
It updates the month in the reports calendar based on scroll position and not when scroll has ended

### Relationships
Closes #4395

## Why do we want this?
When scrolling fast between months it is hard to tell in which month we are.

## How is it done?
By using `Scrolled` instead if `DecelerationEnded`

### Why not another way?
I am open to ideas

### Side effects
Not that I know of

## Review hints
It works better on device since there it is easier to do the fast scrolling.

## :squid: Permissions
:squid:


arturs105:
Noticed this weird glitch, when testing on an iPhone. Happens when scrolling to the left (backwards).

ezgif-5-5a52252973b5

sofokli1:
@arturs105 Do you mean some squares in the calendar grid? if yes does it also happen in develop? If it does happen in develop then it is unrelated to this PR.

arturs105:
I checked and it does not happen on `develop`. The issue is not only the squares in the grid, but also that it scrolls by 2 months - at the very beginning of the gif you can see that it goes from February to December.

sofokli1:
Thanks, I will check it out

sofokli1:
@arturs105 please take a look again, just fixed some calculations

arturs105:
The scrolling by 2 months is indeed fixed, but there are still some glitches when scrolling - the backgrounds in the grid and the cell numbers animating. This is due to cell reuse, so we should try to find a way to disable those animations. I tried adding `UIView.AnimationsEnabled = false` and that indeed fixed the unwanted cell animations, but that also disables all animations globally, so we should find a way to disable just those unwanted animations.

sofokli1:
I split the logic into 2 parts. I use the `Scrolled` delegate just to update the month name and the `DecelerationEnded` for the rest of the logic as it was before.

PR #4430, opened 2019-02-19
Better reactive testing

heytherewill:
## What's this?
This removes all the async warnings from our test codebase

## Why do we want this?
awaitable actions are nice

## Tests
Syncing tests: Link to be added  - do not merge this branch unless these tests are passing.

## :squid: Permissions
Anyone

PR #4469, opened 2019-02-26
Better RxAction tests

Odrakir:
## What's this?
In an effort to remove warnings in tests I removed the return value from `Execute`, because we were only using it in tests anyway. The previous implementation is now called `ExecuteWithObservable`.

I then added the RxAction extension methods `ExecuteSequentally ` and `AddToSequence` that adds the needed functionality for tests only.

I also updated RxAction's docs.

### Relationships
Supersedes #4430

## :squid: Permissions
@heytherewill 

## Testing Docs
I'm adding here the testing part of the updated docs because I think this is relevant to understand the PR:

### Concatenating different executions of the same action

Sometimes we need to test how a single action can be executed multiple times in a row, this can't be done just by writing multiple `Execute` lines one after the other because the action is disabled while it's executing and this would cause an exception.

That's why we have this extension `ExecuteSequentally`:

```c#
ViewModel.SelectTag.ExecuteSequentally(selectedTags)
   .Subscribe()
```

where `selectedTags` in this case is an enumerable of all the different inputs we want to execute the action for.

In the case of `UIAction` that takes no imputs the parameter is replaced with an int for the number of times we want to execute the action.

```c#
ViewModel.EditStartTime.ExecuteSequentally(times: 2)
    .Subscribe()
```

### Concatenating different actions

Sometimes the above is not enough and we want to execute one action (once or multiple times) and then a different one on completion. That's when we use `AddToSequence`:

```c#
ViewModel.SelectTag.ExecuteSequentally(selectedTags)
    .AddToSequence(ViewModel.Save)
    .Subscribe()
```

This example will first execute `SelectTag` multiple times (once for every selectedTag) and then, when the last one completes it'll execute `Save`.

### Awaiting action execution

This is what was problematic before, we were returning an `Observable` from `Execute` but there was no real need to subscribe to that `Observable` for the action to start. That's why we were ignoring it, which resulted in some warnings during testing.

This is fixed now, `Execute` won't return anything and wether you use `Execute`, `ExecuteWithCompletion` or `ExecuteSequentally` you don't really need to await for the completion of the actions as long as you are using a `TestScheduler` because everything will happen immediately.

```c#
ViewModel.SelectTag.ExecuteSequentally(selectedTags)
    .AddToSequence(ViewModel.Save)
    .Subscribe();

TestScheduler.Start();
```

### Common mistakes

When using `ExecuteWithCompletion` or  `ExecuteSequentally` with or without `AddToSequence` don't forget to add a `Subscribe` at the end, otherwise the action will never start. This is not the case if you just use `Execute`.

If you try to observe the other properties from `RxAction`, like `Elements` or `Errors`, you have to be aware those *never complete*, even if the action does.


Toggl.Foundation.Tests/MvvmCross/ViewModels/TimeEntriesViewModelTests.cs
```diff
@@ -330,9 +330,9 @@ public async ThreadingTask ShowsTheUndoUI()
             [Fact]
             public async ThreadingTask DoesNotHideTheUndoUITooEarly()
             {
-                var observable = viewModel.DelayDeleteTimeEntry.Execute(timeEntry);
+                viewModel.DelayDeleteTimeEntry.Execute(timeEntry);
                 SchedulerProvider.TestScheduler.AdvanceBy(Constants.UndoTime.Ticks - 1);
-                await observable;
+                await ViewModel.DelayDeleteTimeEntry.Elements;;
```
heytherewill: Two ; 👀

Toggl.Foundation.Tests/Toggl.Foundation.Tests.csproj
```diff
@@ -21,6 +21,7 @@
     <ProjectReference Include="..\Toggl.Ultrawave\Toggl.Ultrawave.csproj" />
   </ItemGroup>
   <ItemGroup>
+    <PackageReference Include="System.Reactive" Version="4.1.2" />
```
heytherewill: This is not needed

Toggl.Foundation.Tests/MvvmCross/ViewModels/TimeEntriesViewModelTests.cs
kornelijepetak: Come on, help the guy with the

```suggestion
                await ViewModel.DelayDeleteTimeEntry.Elements;
```

Toggl.Foundation.Tests/TestExtensions/RxActionExtensions.cs
```diff
@@ -0,0 +1,53 @@
...
+        {
+            var observables = inputs
+                .Select(input => Observable.Defer(() => action.ExecuteWithCompletion(input)));
+
+            return Observable.Concat(
+                observables
+            );
+        }
+
+        public static IObservable<Unit> AddToSequence(this IObservable<Unit> observable, RxAction<Unit, Unit> action)
```
kornelijepetak: The naming is IMO totally off.
In methods named like this, the initial thought is that the argument will be the sequence into which to add the instance which the method is being called on.

I think safer bet would be something along the lines of promises, and that is:
```cs
ViewModel.SomeAction.ExecuteWithCompletion()
    .ThenExecute(ViewModel.ContinuationAction);
```

Btw. what if I want to execute action A twice, and then when all is completed, call action B twice?
Shouldn't the `ThenExecute` (or `AddToSequence` as it's now called) accept the count (or params[]) so that it can be executed multiple times as well. Not sure if there are any places where this is needed, but just thinking out loud.

Odrakir: > In methods named like this, the initial thought is that the argument will be the sequence into which to add the instance which the method is being called on.

:wat: ?

It's called like that because only the first one is actually being executed, the rest are only added to the observable chain and only executed when subscribed too.

And as to accept repetitions, I doubt we will need it, but if we do we can worry about it then, for now YAGNI.

kornelijepetak: > :wat: ?
I still have the issue with the naming.

But let me try to explain it a bit better than before.

When you have a function called `.WithLatestFrom(x)`, you will read the `x` argument as a continuation of the sentence: _with latest from x_. The same way when you say `y.AddToSequence(x)` it seems that `x` is the sequence to which `y` is being added and read like _add to sequence x_. While it's actually not the case. 

I agree about subscription aspect of this and it complicates the naming, yes. So `ThenExecute` would not be a good name. But I'd still rather change it into something that does not sound the other way around. In the lack of better name suggestion, I will approve this. But I am not happy about this name.

Odrakir: Oh, I see what you mean now... How about `PrependAction(x)`?

Issue #4245, opened 2019-01-17
Re-schedule calendar notifications after device Restart


daividssilverio:
This issue depends on #4244.

Android is nice to us and removes all pending intents when a device is restarted, we've got to reschedule all pending intents using the `AlarmManager` once again when the device is restarted.

Hints:
Take a look at
```
<uses-permission android:name="android.permission.WAKE_LOCK" />
<uses-permission android:name="android.permission.RECEIVE_BOOT_COMPLETED" />
```

We are most likely going to create a broadcast receiver to listen to the device boot and re-schedule pending notifications.

**The notifications will have to be stored somewhere**, or we can use `ScheduleEventNotificationsInteractor` in the broascast receiver, if possible (this would actually be more precise, in case the events changed).

juanlaube:
Closed by #4520

PR #4479, opened 2019-02-27
Calendroid: Fix notification scheduler

### LINKED ISSUE #4244: Implement/Fix Android notification service
It has to actually use the `Multivac.Notification.ScheduledTime` and notify the user only at the expected time.

This issue should take care of scheduling the notifications only. Re-scheduling notifications after the device boots will have it's own issue.

Hints: Use the `AlarmManager` to schedule a pending intent to run at the specified time and create a `BroadcastReceiver` to actually send the notifications (call `NotificationManager.Notify`)

daividssilverio:
## What's this?
This fixes the android `NotificationServiceAndroid.cs` to actually schedule notifications using the `Multivac.Notification:ScheduledTime`.

### Relationships
Closes #4244 

## Why do we want this?
To show calendar event notifications based on smart alert settings.

## How is it done?
Instead of firing the notifications on the spot, it uses the `AlarmManager` to send an intent that will be handled by the new broadcast listener that's responsible for actually firing the notifications.

### Why not another way?
This is easy enough.

### Side effects
None, this service is used by the Calendar only.

## Review hints
You can use this snippet to see it working.
It doesn't persist schedules through a device restart, that's gonna be done at #4245.

## :squid: Permissions
🦑


Toggl.Giskard/Services/NotificationServiceAndroid.cs
```diff
@@ -26,15 +30,16 @@ public IObservable<Unit> Schedule(IImmutableList<Multivac.Notification> notifica
 
                     notifications.Select(notification =>
```
heytherewill: Do you even need to select and then `foreach` at this point? This is not doing any good: it just adds makes you traverse the collection twice and adds an indirection.

Might as well fix this with a simple:
```
foreach(var notification in notifications)
{
    var notificationId = notification.Id.GetHashCode();
    var androidNofitication = notificationsBuilder.SetAutoCancel(true)
                                .SetContentIntent(pendingIntent)
                                .SetContentTitle(notification.Title)
                                .SetSmallIcon(Resource.Drawable.ic_icon_running)
                                .SetContentText(notification.Description)
                                .Build();

    scheduleNotification(notificationId, androidNotification, notification.ScheduledTime);
}
```

daividssilverio:
Oh, the code snippet:
```
var context = Application.Context;
            var notificationIntent = context.PackageManager.GetLaunchIntentForPackage(context.PackageName);
            notificationIntent.SetPackage(null);
            var pendingIntent = PendingIntent.GetActivity(context, 0, notificationIntent, 0);
            var notificationManager = NotificationManager.FromContext(context);
            var notificationsBuilder = context.CreateNotificationBuilderWithDefaultChannel(notificationManager);

            NotificationServiceAndroid.scheduleNotification(
                new Random().Next(),
                notificationsBuilder.SetAutoCancel(true)
                    .SetContentIntent(pendingIntent)
                    .SetContentTitle("Nice Titlte")
                    .SetSmallIcon(Resource.Drawable.ic_icon_running)
                    .SetContentText("Nice Content!")
                    .Build(),
                DateTimeOffset.Now + TimeSpan.FromMinutes(2)
            );
```

Toggl.Giskard/Services/NotificationServiceAndroid.cs
daividssilverio: it's still a foreach, right?

heytherewill: It is

PR #4481, opened 2019-02-27
📝 add information about the ui test apk architecture

khoi:
## What's this?
this adds a Notice section to the doc to ensure that the ui test development is setup correctly.

## :squid: Permissions
🦑


docs/running-android-ui-tests.md
```diff
@@ -17,6 +17,11 @@ Debug keystore information (MacOS):
 - Store password: `android`
 - Key password: `android`
 
+**⚠ Notice:** The debug binary by default only support ARM architecture, so you must either:
```
heytherewill: Please use the `:warning:` instead of the Unicode emoji 🙇 
```suggestion
**:warning: Notice:** The debug binary, by default, only supports the ARM architecture, so you must either:
```

PR #4485, opened 2019-02-27
Calendar settings activity

### LINKED ISSUE #4447: Calendroid Settings: Update the settings view to add calendar settings & Smart Alerts
The calendar Settings opens the fragment from #4286

The smart alerts opens the view from #4448

juanlaube:
## What's this?
This adds a new activity to select the calendars in the settings tab.

### Relationships

Closes #4447 

## Why do we want this?
📆 ⚙️ 

## How is it done?
Extracting part of the layout of the select calendars fragment and reusing that in the new activity, plus bindings and stuff like that.

### Side effects
None.

## Review hints
Try it!

## :squid: Permissions
🦑


Toggl.Foundation.MvvmCross/ViewModels/Settings/CalendarSettingsViewModel.cs
```diff
@@ -59,7 +70,33 @@ private void requestAccess()
...
+                if (!authorized)
+                {
+                    calendarListVisible = await permissionsService.RequestCalendarAuthorization();
+                }
+                else
+                {
+                    calendarListVisible = true;
+                }
+            }
+            calendarListVisibleSubject.OnNext(calendarListVisible);
```
daividssilverio: We could override `Calendars` from the `SelectUserCalendarsViewModelBase` and use `combineLatest` to re-fetch the calendars whenever the permissions are changed.

juanlaube: That way wouldn't work because the observable for the permissions doesn't emit on every permission change. It's fixed now though.

daividssilverio:
I've created another issue to handle `don't ask again' permission request results #4491

PR #4486, opened 2019-02-27
RxFactor the Edit Project view

### LINKED ISSUE #3828: RxFactor the EditProjectViewModel
[Standard RxFactor rules apply](https://www.notion.so/List-of-Refactors-070744dc374f416eacb0381d72143734#0f425d73fe5343a38a454551e8d26a04)

This view is pretty straightforward, but it's also crucial to the app's core functionality, so make sure to test your changes thoroughly and request my review!

heytherewill:
## What's this?
An RxFactor of the Edit Project view

### Relationships
Closes #3828

## Why do we want this?
RxFactors are cool and hip

## How is it done?
The same way as other RxFactors.

The cool thing here, though, is that I'm leveraging RxActions to their full potential. We had them in the past, but now I'm finally making use of the `.Elements` property!

### Side effects
Please test test and test, this view is too important

## Review hints
Commit by commit definitely helps

## :squid: Permissions
Anyone, but only when we have at least 3 approvals.


Toggl.Daneel/ViewControllers/EditProjectViewController.cs
```diff
@@ -38,55 +41,86 @@ public override void ViewDidLoad()
...
+                .Select(color => color.ToNativeColor())
+                .Subscribe(ColorCircleView.Rx().BackgroundColor())
+                .DisposedBy(DisposeBag);
+
+            ViewModel.NameIsAlreadyTaken
+                .Select(nameIsTaken => nameIsTaken ? nameAlreadyTakenHeight : 0)
+                .Subscribe(ProjectNameUsedErrorTextHeight.Rx().Constant())
+                .DisposedBy(DisposeBag);
+
+            //bindingSet.Bind(DoneButton)
```
juanlaube: 🔥

Toggl.Giskard/Resources/layout/EditProjectActivity.axml
```diff
@@ -262,8 +258,8 @@
         android:layout_width="match_parent"
         android:layout_height="wrap_content">
         <TextView
-            local:MvxBind="Click DoneCommand; TextColor CreateProjectButtonColor(SaveEnabled)"
             android:id="@+id/CreateProjectButton"
+            local:MvxBind="TextColor CreateProjectButtonColor(SaveEnabled)"
```
juanlaube: 👀 This should have an rx binding

Toggl.Foundation.MvvmCross/ViewModels/EditProjectViewModel.cs
```diff
@@ -205,72 +305,25 @@ private async Task done()
...
-                return;
-            }
-
-            var client = await interactorFactory.GetClientById(selectedClientId.Value).Execute();
-            clientId = client.Id;
-            ClientName = client.Name;
-            updateIsNameAlreadyTaken();
-        }
+            var createdProjectId = await
+                from name in Name.Take(1)
```
juanlaube: I like this query syntax 😄

Toggl.Foundation.MvvmCross/ViewModels/EditProjectViewModel.cs
```diff
@@ -1,135 +1,221 @@
...
+                .Select(w => w.Name)
+                .DistinctUntilChanged()
+                .AsDriver(schedulerProvider);
+
+            ClientName = clients
+                .Select(client => client?.Name ?? "")
+                .DistinctUntilChanged()
+                .AsDriver(schedulerProvider);
+
+            Color = Observable.Create((Func<IObserver<MvxColor>, IDisposable>)createColorObservable)
```
Odrakir: I think this has to be improved, at this point I would probably go with an external subject for `Color`, but if we had tests for all edge cases we might be able to come up with a subject-less solution.Probably using Scan to get the previous element.

My point is that code is not very easy to understand and debug. Then `Scan` solution is not easy either, but at least it just uses operators.

Toggl.Foundation.MvvmCross/ViewModels/EditProjectViewModel.cs
```diff
@@ -139,61 +225,75 @@ public override void ViewAppeared()
...
-            updateIsNameAlreadyTaken();
-        }
+        private Task close()
+            => navigationService.Close(this, null);
 
-        private void OnNameChanged()
+        private IObservable<IThreadSafeWorkspace> pickWorkspace()
         {
-            updateIsNameAlreadyTaken();
+            return from currentWorkspace in workspaces.Take(1)
```
Odrakir: Is this syntax clearer? Can we vote on it?
Also, what happens if you forget the `Take(1)`?

Toggl.Foundation.MvvmCross/ViewModels/EditProjectViewModel.cs
```diff
@@ -1,135 +1,221 @@
...
-            DoneCommand = new MvxAsyncCommand(done);
-            CloseCommand = new MvxAsyncCommand(close);
-            PickColorCommand = new MvxAsyncCommand(pickColor);
-            PickClientCommand = new MvxAsyncCommand(pickClient);
-            PickWorkspaceCommand = new MvxAsyncCommand(pickWorkspace);
-            TogglePrivateProjectCommand = new MvxCommand(togglePrivateProject);
+            Close = rxActionFactory.FromAsync(close);
+            PickColor = rxActionFactory.FromObservable<MvxColor>(pickColor);
+            PickClient = rxActionFactory.FromObservable<IThreadSafeClient>(pickClient);
+            ChangeName = rxActionFactory.FromFunction<string, string>(CommonFunctions.Identity);
```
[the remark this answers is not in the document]
heytherewill: Why add more code when this is exactly what Actions are for?

Toggl.Foundation.MvvmCross/ViewModels/EditProjectViewModel.cs
heytherewill: Yes, I find it clearer
Yes, you can vote
The same that would happen if you forget the `Take(1)` on the fluent syntax: It would never complete

Toggl.Foundation.MvvmCross/ViewModels/EditProjectViewModel.cs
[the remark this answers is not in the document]
Odrakir: Actions are not for that. As the name imply they are meant to execute an action, but right there you are just using it to pass a value to other observables without actually executing any action itself. I find it confusing.

Toggl.Foundation.MvvmCross/ViewModels/EditProjectViewModel.cs
Odrakir: You are introducing a new syntax which will make code harder to understand and debug for people not familiar with it (me included). Also I don't really see a big benefit on this over rx operators. 

I don't agree with it. But if this is where we are going there should be some kind of warning so everyone is aware and we should try to write all future code this way or maybe even refactor previous code to be like this.

Toggl.Multivac/Extensions/RxAction/InputOutputAction.cs
```diff
@@ -0,0 +1,49 @@
+﻿using System;
+using System.Reactive.Concurrency;
+using System.Reactive.Disposables;
+using System.Reactive.Linq;
+using System.Reactive.Threading.Tasks;
+using System.Threading.Tasks;
+
+namespace Toggl.Multivac.Extensions
+{
+    public sealed class InputOutputAction<TInput, TElement> : RxAction<TInput, TElement>
```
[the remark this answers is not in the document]
heytherewill: I thought `RxAction` was abstract

Toggl.Foundation.Tests/MvvmCross/ViewModels/EditProjectViewModelTests.cs
```diff
@@ -33,7 +32,14 @@ protected EditProjectViewModelTest()
             }
 
             protected override EditProjectViewModel CreateViewModel()
-                => new EditProjectViewModel(DataSource, DialogService, InteractorFactory, NavigationService, StopwatchProvider);
+                => new EditProjectViewModel(
+                    DataSource,
+                DialogService,
```
kornelijepetak: Just some indentation here.

Toggl.Foundation.Tests/MvvmCross/ViewModels/EditProjectViewModelTests.cs
```diff
@@ -338,7 +353,7 @@ public async Task SetsTheWorkspaceId()
             {
                 setupDefaultWorkspace();
                 await ViewModel.Initialize();
-                await ViewModel.DoneCommand.ExecuteAsync();
+                await ViewModel.Save.Execute(TestScheduler);
```
kornelijepetak: Is this how it works right now. I think we have at least 3 ways of writing tests at the moment, when it comes to synchronicity. 😢 We really need a guidelines document that will set up the rules about how we do those things.

How does this play with Ricardo's changes to Execute no longer returning observable, but void?

Toggl.Foundation.MvvmCross/ViewModels/EditProjectViewModel.cs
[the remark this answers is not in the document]
kornelijepetak: I agree with Ricardo here. We used to use relays for things like this one.

I feel that in general, you have added unneeded complexity to this PR and it's more confusing than needed. That's my opinion though.

Also, #4192 is something many devs agreed upon, and this either has to be respected or updated with whatever you think is better if we can as a team agree on that.

Toggl.Foundation.MvvmCross/ViewModels/EditProjectViewModel.cs
kornelijepetak: Honestly, I am against it. 

And is another example of why I think you just unnecessarily complicated the code. I can understand the hype that better understanding of monoids in the category of endofunctors can bring to one's heart, but unless the whole team is as versed in functional programming as you are, I don't think it's adding value.

I don't have problems understanding this code after I read it carefully, but it takes a lot more effort than what we used to do (and what the whole codebase is using for navigation), and that is await the results from the navigation service.

Query syntax is cool, but it's not for every case. 

2 cents.

Toggl.Foundation.MvvmCross/ViewModels/EditProjectViewModel.cs
```diff
@@ -139,61 +223,75 @@ public override void ViewAppeared()
...
+                   select newClient;
+            
+            IObservable<IThreadSafeClient> clientFromViewModel(IThreadSafeClient currentClient, IThreadSafeWorkspace currentWorkspace)
+                => navigationService
+                    .Navigate<SelectClientViewModel, SelectClientParameters, long?>(
+                        SelectClientParameters.WithIds(currentWorkspace.Id, currentClient?.Id))
+                    .ToObservable()
+                    .SelectMany(clientFromId);
+
+            IObservable<IThreadSafeClient> clientFromId(long? selectedClientId)
```
kornelijepetak: Some inline functions are lambdas, some are regular functions... I'd rather all be regular 🤷‍♂️ 
It looks more like a function, and not easily confused with assignment and that there's more code after `return`. Kind of detracts from the goal of the function, albeit from compiler perspective being correct.

Toggl.Foundation.MvvmCross/ViewModels/EditProjectViewModel.cs
kornelijepetak: The bare fact that `Take(1)` is needed for each of these shows that the syntax was NOT introduced with this scenario in mind, regardless of the fact that in it's nature it supports it.

Toggl.Foundation.MvvmCross/ViewModels/EditProjectViewModel.cs
heytherewill: We need the lambdas for captured variables

heytherewill: Unless with lambda you mean they use the EBM syntax, in which case I wouldn't mind changing

kornelijepetak: I don't know what you are asking, but I'll explain myself a bit clearly.

1. I should've used the correct term: `local functions`, not `inline`.
2. Yes, I'd rather you not use EBM syntax for `local functions` to better signify that those are functions.
3. Explicit local functions can capture variables the same way as EBM, which if I did understand your last comment you are fully aware of.
4. I read: "We need the `lambadas` for..." 😉

Toggl.Foundation.Tests/MvvmCross/ViewModels/EditProjectViewModelTests.cs
```diff
@@ -252,7 +252,7 @@ public void IsFalseWhenTheNameIsJustWhiteSpace()
             public void IsFalseWhenTheNameIsLongerThanTheThresholdInBytes()
             {
                 TestScheduler.Start();
-                ViewModel.ChangeName.Execute("This is a ridiculously big project name made solely with the purpose of testing whether or not Toggl apps UI has validation logic that prevents such a large name to be persisted or, even worse, pushed to the api, an event that might end up in crashes and whatnot");
+                ViewModel.ChangeName.Accept("This is a ridiculously big project name made solely with the purpose of testing whether or not Toggl apps UI has validation logic that prevents such a large name to be persisted or, even worse, pushed to the api, an event that might end up in crashes and whatnot");
```
kornelijepetak: Those kind of messages make me happy.

Toggl.Foundation.MvvmCross/ViewModels/EditProjectViewModel.cs
```diff
@@ -296,25 +294,24 @@ private async Task done()
...
-                    IsPrivate = isPrivate,
-                    ClientId = client?.Id,
-                    Billable = billable,
-                    WorkspaceId = workspace.Id
-                }
-                from project in interactorFactory.CreateProject(dto).Execute()
-                select project.Id;
-
-            await navigationService.Close(this, createdProjectId);
+            var name = await Name.Take(1);
```
kornelijepetak: Maybe consider changing (use your best judgement), but I find `Name.FirstAsync()` conveying the intent better than `.Take(1)`. 

`Take(1)` sounds like there's more, and you only want one. And since the underlying type is the behaviour subject always having only one value, it looks strange. Even worse, `Take(x)` implies that maybe `Take(2)` would work with multiple emissions thru the subject, but then again awaiting the observable will always return the last value and not both values (or N in case of `Take(N)`) so it kind of makes no sense. None of this confusion is present with the `FirstAsync`.

kornelijepetak: If you decide to change it, take a look whether there are other instance where `FirstAsync` could be used to clarify intent better instead of `Take(1)`. If you decide not to change it, preferrably with an explanation as to why, let me know in this comment.

Toggl.Foundation.MvvmCross/ViewModels/EditProjectViewModel.cs
```diff
@@ -139,138 +218,112 @@ public override void ViewAppeared()
...
-            updateIsNameAlreadyTaken();
-        }
-
-        private void pickRandomColor()
-        {
-            var randomColorIndex = random.Next(0, Helper.Color.DefaultProjectColors.Length);
-            Color = Helper.Color.DefaultProjectColors[randomColorIndex];
+            IObservable<CreateProjectDTO> getDto(IThreadSafeWorkspace workspace)
+                => Observable.CombineLatest(
+                    Name.FirstAsync(),
```
Odrakir: Still think some subject would make this code much easy to read. But anyway... Why so many `FirstAsync`? Couldn't a `Take(1)` at the end have the same effect?

heytherewill: @kornelijepetak Preferred `FirstAsync` over `Take(1)`. I am down to either

PR #4487, opened 2019-02-28
🧪 ui tests for starting a manual TE

### LINKED ISSUE #4347: Create manual time entry
Sign in
Start manual time entry with a long press
Complete

khoi:
## What's this?
Add a couple of UI tests to ensure starting manual TE work correctly.

### Relationships
Closes #4347

## Why do we want this?
Automation, who doesn't want it.

### Side effects
on Android, I add the key press event handler for the "Enter" event so that it can be used in UI test.

## Review hints
None beside running them.

## :squid: Permissions
🦑


Toggl.Tests.UI/StartManualTimeEntryTests.cs
```diff
@@ -0,0 +1,47 @@
...
+        [Test]
+        public void TheCreatedTimeEntryHasCorrectDefaultDurationSet()
+        {
+            app.Tap(StartTimeEntry.DoneButton);
+
+            app.WaitForElement("0:30:00");
+        }
+
+        [Test]
+        public void TheDurationIsSetCorrectly()
```
arturs105: For some reason this test always fails on Android

khoi: @arturs105 can you show me the error log, and also what simulator are u using?

arturs105: Can't get it to fail anymore, so nevermind. I guess it was just a bad day for VS, or maybe i had a different build cached.

PR #4497, opened 2019-03-01
Refactor Rating view by using observables instead of hints

### LINKED ISSUE #2882: Refactor showing/hiding the Rating view using observables instead of hints
We had a problem with the presentation hints for showing and hiding the rating view in 1.8 (issue #2829 fixed by #2880). The main view is already rxfactored, so we can remove the hints and use an boolean observable for visibility of the rating dialog and subscribe to this observable instead of using hints.

sofokli1:
## What's this?
Refactors the Rating view by using observables instead of hints

### Relationships
Closes #2882

## Why do we want this?
we want to remove MVX

## How is it done?
remove all signs of `hints` related to the rating view and use observable to show hide the view

### Why not another way?
Open to new ideas

### Side effects
none

## Review hints
none

## :squid: Permissions
:squid:


Toggl.Foundation.MvvmCross/ViewModels/MainViewModel.cs
```diff
@@ -266,10 +267,15 @@ public override async Task Initialize()
...
             }
 
-            ratingViewExperiment
-                .RatingViewShouldBeVisible
-                .Subscribe(presentRatingViewIfNeeded)
-                .DisposedBy(disposeBag);
+            ShouldShowRatingView = Observable.Merge(
+                    ratingViewExperiment.RatingViewShouldBeVisible,
+                    RatingViewModel.HideRatingView.Select(_ => false),
+                    hideRatingView.AsObservable().Select(_ => false)
```
heytherewill: We have a `SelectValue` operator which will make this nicer

Toggl.Daneel/ViewControllers/MainViewController.cs
```diff
@@ -238,6 +238,16 @@ public override void ViewDidLoad()
...
 
+            ViewModel.ShouldShowRatingView
+                .Where(visible => visible)
+                .Subscribe(_ => showRatingView())
+                .DisposedBy(disposeBag);
+
+            ViewModel.ShouldShowRatingView
+                .Where(visible => !visible)
+                .Subscribe(_ => hideRatingView())
+                .DisposedBy(disposeBag);
```
heytherewill: Why two subscriptions instead of just one that checks the `visible` argument and calls the appropriate method?

sofokli1: got inspired from line 393

```
            ViewModel.IsTimeEntryRunning
                .Where(visible => visible)
                .Subscribe(_ => showTimeEntryCard())
                .DisposedBy(disposeBag);

            ViewModel.IsTimeEntryRunning
                .Where(visible => !visible)
                .Subscribe(_ => hideTimeEntryCard())
                .DisposedBy(disposeBag);
```

PR #4515, opened 2019-03-01
Remove the dependency on the shortcut creator

### LINKED ISSUE #3955: Make services subscribe to login/logout events
Right now, we are keeping references of the `ITogglDataSource` and of the `IApplicationShortcutCreator` lying around in the `UserAccessManager`. We should change the direction of this dependency, making these two services subscribe to the `UserLoggedIn` and `UserLoggedOut` methods, as appropriate.

heytherewill:
## What's this?
This removes a needless dependency  from the AccessManager

### Relationships
Closes #3955

## Why do we want this?
The access manager needs to be free so it can instantiate as lightly as possible

### Side effects
None, the dependency wasn't even relevant. The LogoutInteractor was already doing all the work

## :squid: Permissions
Anyone, anytime


sofokli1:
I have the same question as @khoi

heytherewill:
Bear in mind that this subscription will be changed later in the DI refactor, so if it looks a bit hacky is because it indeed is

PR #4517, opened 2019-03-02
Remove Reports Provider

heytherewill:
## What's this?
This refactors the Reports Provider in an interactor, to further remove the ITogglDataSource's dependencies on the API

### Relationships
Ref #3957

## Why do we want this?
Check the issue

## How is it done?
This is mostly trivial. However, since the interactors are short lived and we need a simple memory cache to avoid hitting the API too many times when the user has projects that do not exist locally, I created a simple class that keeps the cache alive after the interactor itself has been collected.

### Why not another way?
We might consider having a global cache that can store all sorts of entities. That's something for the future, though

### Side effects
None, as per tests

## Review hints
Reviewing by commit will prove to be better
Also, using some diff tool to diff the ReportsProvider and the interactor will help you visualize the differences

## :squid: Permissions
Anyone


Toggl.Foundation/Interactors/Reports/GetProjectSummaryInteractor.cs
```diff
@@ -0,0 +1,196 @@
...
+            => projectsApi.Search(workspaceId, projectIds)
+                .Do(persistInMemoryCache);
+
+        private void persistInMemoryCache(List<IProject> apiProjects)
+        {
+            memoryCache.PersistInCache(apiProjects);
+        }
+
+        private IObservable<long[]> projectsIdsNotInDatabase(long[] projectIds)
+        {
```
sofokli1: why not just:
```
=> projectIds
        .Select(id => tryGetProjectFromDatabase(id))
        ...
```

Toggl.Foundation/Interactors/Reports/GetProjectSummaryInteractor.cs
```diff
@@ -0,0 +1,196 @@
...
+                    ? memoryProjectsObservable
+                    : memoryProjectsObservable
+                        .Merge(searchApi(workspaceId, notInMemoryIds))
+                        .SelectMany(list => list)
+                        .ToList();
+            });
+
+        private IObservable<List<IProject>> searchApi(long workspaceId, long[] projectIds)
+            => projectsApi.Search(workspaceId, projectIds)
+                .Do(persistInMemoryCache);
```
sofokli1: you could go for `.Do(memoryCache.PersistInCache);` and also completely remove the `persistInMemoryCache` func since it is just one line

Toggl.Foundation/Interactors/Reports/GetProjectSummaryInteractor.cs
```diff
@@ -0,0 +1,196 @@
...
+
+                var notInMemoryIds = projectsInMemory.SelectAllLeft().ToArray();
+                var memoryProjectsObservable =
+                    Observable.Return(projectsInMemory.SelectAllRight().ToList());
+
+                return notInMemoryIds.Length == 0
+                    ? memoryProjectsObservable
+                    : memoryProjectsObservable
+                        .Merge(searchApi(workspaceId, notInMemoryIds))
+                        .SelectMany(list => list)
```
sofokli1: `CommonFunctions.Identity` maybe?

Toggl.Foundation/Interactors/Reports/GetProjectSummaryInteractor.cs
```diff
@@ -0,0 +1,196 @@
...
+            var percentage = totalSeconds == 0 ? 0 : (summary.TrackedSeconds / (float)totalSeconds) * 100;
+            var billableSeconds = summary.BillableSeconds ?? 0;
+
+            return project == null
+                ? new ChartSegment(Resources.NoProject, null, percentage, summary.TrackedSeconds, billableSeconds, Helper.Color.NoProject)
+                : new ChartSegment(project.Name, client?.Name, percentage, summary.TrackedSeconds, billableSeconds, project.Color);
+        }
+
+        private IObservable<IList<IProject>> searchProjects(long workspaceId, long[] projectIds) =>
+            Observable.DeferAsync(async cancellationToken =>
```
sofokli1: what is the purpose of the `cancelationToken` and where is it used?

Toggl.Foundation/Interactors/Reports/GetProjectSummaryInteractor.cs
```diff
@@ -0,0 +1,196 @@
...
+
+                var notInDatabaseIds = projectsInDatabase.SelectAllLeft().ToArray();
+                var databaseProjectsObservable =
+                    Observable.Return(projectsInDatabase.SelectAllRight().ToList());
+
+                return notInDatabaseIds.Length == 0
+                    ? databaseProjectsObservable
+                    : databaseProjectsObservable
+                        .Merge(searchMemoryAndApi(workspaceId, notInDatabaseIds))
+                        .SelectMany(list => list)
```
sofokli1: `CommonFunctions.Identity` maybe?

Toggl.Foundation/Interactors/Reports/GetProjectSummaryInteractor.cs
```diff
@@ -0,0 +1,196 @@
...
+                .SelectNonNulls()
+                .ToArray();
+
+            return projectsIdsNotInDatabase(summaryProjectIds);
+        }
+
+        private IObservable<IClient> findClient(IProject project)
+            => project != null && project.ClientId.HasValue
+                ? clientsRepository.GetAll()
+                    .SelectMany(clients => clients)
```
sofokli1: `CommonFunctions.Identity` maybe?

sofokli1:
Do you think the `ReportsMemoryCache.cs` needs tests or is it ok since it is quite small and simple?

heytherewill:
I think it's simple enough and it's covered by the interactor tests

Toggl.Foundation/Interactors/Reports/GetProjectSummaryInteractor.cs
heytherewill: The DeferAsync method provides a cancellationToken, but its usage is optional

PR #4520, opened 2019-03-04
Schedule calendar notifications on device restart

### LINKED ISSUE #4245: Re-schedule calendar notifications after device Restart

juanlaube:
## What's this?
This schedules the calendar notifications when the device is restarted.

### Relationships

Closes #4245 

## Why do we want this?
Because we want users to receive notifications, even if they restart their devices.

## How is it done?
It's a broadcast receiver that listens for `BOOT_COMPLETED` and starts a `JobIntentService` that triggers the interactor to schedule notifications.

### Why not another way?
There's no other way.

### Side effects
None.

## Review hints
I left some logs for debugging and testing purposes, you can restart a device (or the emulator) and see the logs with `adb logcat | grep asdf`

## :squid: Permissions
🚫 Don't, I will remove all the logs after someone tested this.


Toggl.Giskard/BroadcastReceivers/BootCompletedBroadcastReceiver.cs
```diff
@@ -0,0 +1,24 @@
...
+using Android.Support.V4.App;
+using Android.Util;
+using Toggl.Giskard.Services;
+
+namespace Toggl.Giskard.BroadcastReceivers
+{
+    [BroadcastReceiver(
+        Exported = true,
+        Permission = "android.permission.BIND_JOB_SERVICE",
+        Name = "com.toggl.giskardBootCompletedBroadcastReceiver")]
```
heytherewill:
```suggestion
        Name = "toggl.giskard.BootCompletedBroadcastReceiver")]
```

daividssilverio:
Also, please make `ScheduleEventNotificationsService` call `StopSelf` in case of an unhandled exceptions. We don't the app to keep crashing in case something fails horribly and the job intent service is retried.

It's alright not to call it if handle all possible exceptions though.

Toggl.Giskard/Services/ScheduleEventNotificationsService.cs
```diff
@@ -0,0 +1,50 @@
...
+
+namespace Toggl.Giskard.Services
+{
+    [Service(
+        Exported = true,
+        Permission = "android.permission.BIND_JOB_SERVICE", 
+        Name = "com.toggl.giskard.ScheduleEventNotificationsService")]
+    public class ScheduleEventNotificationsService : JobIntentService
+    {
+        private IDisposable disposable;
```
daividssilverio: dispose this on `OnDestroy`.

daividssilverio: and set it to null there.

Toggl.Giskard/Services/ScheduleEventNotificationsService.cs
```diff
@@ -0,0 +1,50 @@
...
+
+        protected override void OnHandleWork(Intent intent)
+        {
+            var setup = MvxAndroidSetupSingleton.EnsureSingletonAvailable(ApplicationContext);
+            setup.EnsureInitialized();
+
+            if (!Mvx.TryResolve<IInteractorFactory>(out var interactorFactory))
+                return;
+
+            if (disposable != null)
```
daividssilverio: you could just use disposable?.Dispose() here as well.

Toggl.Giskard/Services/ScheduleEventNotificationsService.cs
juanlaube: Done

PR #4522, opened 2019-03-04
Change integration tests for reports totals

simonrozsival:
## What's this?
Fixes integration tests.

### Relationships
https://github.com/toggl/backend/issues/7492

## Why do we want this?
We want integration tests to pass.

## How is it done?
Our integration test checked if the server returned data grouped by weeks, days, or months. It's not necessary actually and so I relaxed what the test tests: server never returns more than 34 groups which we show as bars in our app. I think this test actually makes more sense than the previous one.

### Why not another way?
It could be a property test... but it's probably not a good idea for an integrationt test.

### Side effects
We test something different, but I think that in the end it's better now.

## Review hints
Run integration tests yourself or check the integration tests link.

## Tests
Integration tests: https://app.bitrise.io/build/da2651f42e077c80 - do not merge this branch unless these tests are passing.

## :squid: Permissions
🦑


amulware:
I don't like so much that we're not testing for the resolution this way. :/

simonrozsival:
My reasoning is: we're not in control of the resolution so we should just take whatever the server gives us.

amulware:
Very well. Not happy about Backend dictating our UX though. But that's a point for @maritoggl

Issue #4389, opened 2019-02-12
Create the header row for the calendar


daividssilverio:
It should have two states:

When the calendar permissions weren't given:
https://app.zeplin.io/project/5a0d3cd27a75add1a8924f6e/screen/5c61d5586f0f43465eb218dc

Displaying a button to restart the calendar permissions request.

When the calendar permissions were given:
https://app.zeplin.io/project/5a0d3cd27a75add1a8924f6e/screen/5c61d558a981283c0b3f6493

It should contain the current day, time entries (with summed times) and the number of events

daividssilverio:
Closed by #4525

daividssilverio:
🤦‍♂️ The issue isn't closed yet... #4525 handles the layout only...

juanlaube:
Closed by #4525 and #4540

PR #4525, opened 2019-03-05
Calendar header layout

juanlaube:
## What's this?
This implements the calendar header layout. It's just the layout, the views are not bound because that part is blocked by another PR.

### Relationships

Refs #4389 

## Why do we want this?
🎨 

## How is it done?
📐 

### Side effects
None.

## :squid: Permissions
🦑


Toggl.Giskard/Resources/layout/CalendarHeaderView.axml
```diff
@@ -0,0 +1,123 @@
...
+            android:layout_marginTop="2dp"
+            android:layout_marginBottom="12dp"
+            android:textSize="12sp"
+            android:fontFamily="sans-serif"
+            android:textStyle="normal"
+            android:textColor="@android:color/white"
+            tools:text="Mon"/>
+    </LinearLayout>
+    
+    <FrameLayout
```
daividssilverio: Please take a look at these two containers, this frame layout and the linear layout above.
They seem to be missing some bottom margin. When there's text set to the `HeaderTimeEntriesTextView`, it's to close to the bottom of this header view.

Toggl.Giskard/Resources/layout/CalendarHeaderView.axml
```diff
@@ -0,0 +1,123 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<LinearLayout
+    xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:tools="http://schemas.android.com/tools"
+    android:layout_width="match_parent"
+    android:layout_height="wrap_content"
```
daividssilverio: I think it's safe for this view to have a fixed height of `64 dp` instead of `wrap_content` to match the designs.

Toggl.Giskard/Resources/layout/CalendarHeaderView.axml
```diff
@@ -0,0 +1,134 @@
...
+            android:id="@+id/Day"
+            android:layout_width="24dp"
+            android:layout_height="24dp"
+            android:layout_marginTop="8dp"
+            android:padding="3dp"
+            android:textSize="15sp"
+            android:fontFamily="sans-serif"
+            android:textStyle="bold"
+            android:textColor="@android:color/white"
+            android:background="@drawable/calendarheaderdaybackground"
```
daividssilverio: Use the actual name of the file here, it should've been CalendarHeaderDayBackground.

Toggl.Giskard/Resources/layout/CalendarHeaderView.axml
```diff
@@ -0,0 +1,134 @@
...
+        android:layout_height="36dp"
+        android:layout_marginRight="8dp"
+        android:layout_gravity="center_vertical|right"
+        android:paddingLeft="10dp"
+        android:paddingRight="10dp"
+        android:textSize="14sp"
+        android:fontFamily="sans-serif"
+        android:textStyle="normal"
+        android:textColor="@android:color/white"
+        android:background="@drawable/bluebutton"
```
daividssilverio: Same here, `BlueButton`

Toggl.Giskard/Resources/layout/CalendarHeaderView.axml
daividssilverio: This background is looking odd when the text view is set with a text with 2 digits, the numbers are not centred in the background.
My suggestion is to first remove the fixed height and width and the padding. You shouldn't use those with text views, because we are already specifying the text size.
Then you can set `minHeight` and `minWidth` to say, `28dp`, and set the `TextView` `gravity` to `center`. That should do the job.

PR #4528, opened 2019-03-06
📊 Track Siri actions under regular events

### LINKED ISSUE #4107: Track Siri actions under regular events
We currently have separate events for tracking Siri Start and Stop events. I see no reasons why these should exist. Instead, we should track these under the existing Start/Stop events to make the data easier to evaluate.

khoi:
## What's this?
Remove separated Start and Stop tracking events for Siri, and use the existing ones instead. 

### Relationships
Closes #4107

## Why do we want this?
To make the data easier to evaluate

## How is it done?
- When an intents are invoked, they're saved as `SiriTrackingEvent` to the iOS native shared storage.
- When the app opens, we'll fetch those objects out and convert them to `ITrackableEvent` counterparts using the `ToTrackableEvent` extension method.
- These then are fed to the MainViewModel for tracking.

### Why not another way?
I couldn't find a clearer way to do this.

### Side effects
None.

## Review hints
- Donate multiple intents in the app.
- Trigger them
- Set debug point in `MainViewController`'s `trackSiriEvents` and make sure event are converted properly.

## :squid: Permissions
🎙 ppl


Toggl.Daneel.ExtensionKit/Analytics/SiriTrackingEvent.cs
```diff
@@ -5,65 +5,82 @@
...
+        StopTimer = 2,
+        Error = 3
+    }
+
+    public struct SiriTrackingErrorEventEncodingKeys
+    {
+        public static string Message = "Message";
+    }
+
+    public struct SiriTrackingStartTimerEventEncodingKeys
```
Odrakir: Can we have something just a little less verbose, like: `SiriStartTimerEventKeys`? Same for the rest.

Toggl.Daneel.ExtensionKit/Analytics/SiriTrackingEvent.cs
```diff
@@ -5,65 +5,82 @@
...
+        private static readonly string eventTypeEncodeKey = nameof(eventTypeEncodeKey);
         private static readonly string parametersEncodeKey = nameof(parametersEncodeKey);
 
-        private static readonly string SiriStartTimerEventName = "SiriStartTimer";
-        private static readonly string SiriStopTimerEventName = "SiriStopTimer";
-        private static readonly string SiriIntentErrorEventName = "SiriIntentError";
-
-        public readonly NSString EventName;
+        public readonly SiriTrackingEventType EventType;
         public readonly NSDictionary<NSString, NSString> Parameters;
```
Odrakir: The thing that bothers me a little now, is that I think this dictionary should be a `Dictionary<string, string>` (C#'s) and then we should do the conversion when encoding/decoding, and not when tracking. What do you think?

kornelijepetak: I agree with Ricardo. Adapt to platform only at the first and the last touching point and work with regular .NET stuff in between.

Odrakir:
Fix tests and this is ready to go.

PR #4534, opened 2019-03-06
iPad: Onboarding screens

### LINKED ISSUE #4500: iPad version: Fix onboarding graphic aspect ratio to 3:4 on all iPads
Bear in mind that this has to work on vertical split screen, so please use `UIDevice.Current.UserInterfaceIdiom` to check that we're running on an iPad.

dshr:
## What's this?
A set of AL and code changes that make the onboarding screens behave nicely on iPad.

### Relationships
Closes #4500 

## Why do we want this?
iPad version y'all

## How is it done?
We define a separate set of constrains for iPads, that make the preview graphic 3:4 on all screens, among some other changes. Then, in order to catch split-screen modes, we override the `TraitCollection` property of the `OnboardingViewController` to always return wRhR on iPads, so our split-screen modes don't render the iPhone version of the screen.

### Why not another way?
Seems like the simplest way to me 🤷‍♂️ 

### Side effects
None

## Review hints
Build and run, trying to read .xib files is hard.

## :squid: Permissions
iPad crew only


Toggl.Daneel/ViewControllers/OnboardingViewController.cs
```diff
@@ -24,6 +25,23 @@ public OnboardingViewController()
         {
         }
 
+        public override UITraitCollection TraitCollection
```
sofokli1: same as the login and signup PRs.
Why override and force regular?

dshr: We do this in so we don't break mobile, since our narrow split-screen views differ from it.

sofokli1: apple does not like that https://developer.apple.com/documentation/uikit/uitraitenvironment/1623514-traitcollection

If we are going for a design based on TraitCollection we should talk about possible minor UI changes in order to do it the proper way on all platforms. Also how will the ipad behave in 1:3 if it will always be `Regular`? Same goes for future proofing, if someone else wants to make a change and expects Compact on iPad and changes the override things will break.

dshr: True, this might not be the best implementation route. As for 1:3 behaviour stuck to wRhR — we're overriding it only for a given ViewController, so it's fine on a case by case basis. Still, your point re future-proofing is valid.

dshr: I'll talk to product/design re this and try and come up with a solution.

sofokli1:
why only the first screen has 2 lines of text?
can you make that into one as well or is that by design?
Simulator Screen Shot - iPad Pro (11-inch) - 2019-03-12 at 13 57 21

dshr:
It's in the designs, but maybe a single line might look better. I'll give it a look.

dshr:
Managed to get it to work with some elbow grease! @sofokli1, give it a look please.

PR #4542, opened 2019-03-07
🔍 Track Siri start and stop events when manually invoked

### LINKED ISSUE #4530: Siri - Track Intents when invoked manually
We're currently not count Siri events : start timer, stop timer when users invoke the intents manually (by tapping on the suggestions, Spotlight search result). 

Let's track that by setting the Origin of the event to `Siri`

khoi:
## What's this?
When users invoke a donated intent by pressing on the Spotlight suggestion, we want to track the origin of the action. This PR allows us to do that.

### Relationships
Closes #4530

## How is it done?

- Create a dedicated deeplink for stopping TE from Siri.
- Create a new CreateTimeEntryInteractor that allows Origin to be passed in

### Side effects
Should be none.

## Review hints
None.

## :squid: Permissions
🎙


khoi:
@Odrakir I've merged the new interactor from develop, this should be ready for review

PR #4549, opened 2019-03-07
iPad: Login View

### LINKED ISSUE #4501: iPad version: Login
Check form width and its behaviour with the KB both hidden and extended.

dshr:
## What's this?
A set of AL and code changes that make the login view work nicely on iPads.

### Relationships
Closes #4501

## Why do we want this?
iPad version needs this.

## How is it done?
I've changed the login view constraints to work the same way on both tablets and iPhones, so we'd require a minimum amount of size class changes. The only place where we had to check what device we're running on is when we're changing the form's top constraint in order to make enough space for the keyboard.

### Why not another way?
This one might not be the simplest, but will require less upkeep down the line.

### Side effects
None.

## Review hints
Run the app 🤷‍♂️ 

## :squid: Permissions
iPad crew only


Toggl.Daneel/ViewControllers/LoginViewController.cs
```diff
@@ -181,6 +184,9 @@ public override void ViewDidLayoutSubviews()
             if (View.Frame.Height > iPhoneSeScreenHeight && !keyboardIsOpen)
                 TopConstraint.Constant = topConstraintForBiggerScreens;
 
+            if (UIDevice.CurrentDevice.UserInterfaceIdiom == UIUserInterfaceIdiom.Pad && !keyboardIsOpen)
```
heytherewill: I think an `UIDevice.CurrentDevice.IsIpad()` extension would be nice

sofokli1: we should do it in the end as a small refactoring issue, since there are other branches doing the same

Toggl.Daneel/ViewControllers/LoginViewController.cs
```diff
@@ -40,6 +40,23 @@ public LoginViewController(IntPtr handle) : base(handle)
         {
         }
 
+        public override UITraitCollection TraitCollection
```
sofokli1: why override this? Why force `Regular` size class?
Also apple doesn't like it (as mentioned in the Important part of the doc) https://developer.apple.com/documentation/uikit/uitraitenvironment/1623514-traitcollection

PR #4554, opened 2019-03-07
Modal (dialog) presentation support for iPad

### LINKED ISSUE #4499: iPad version: Modals
- Max width
- Get height from the presented VC preferred height, or use default if not set
- KB behaviour
- Ensure it works as intended on mobile vs narrow split-screen options

For more in-depth details, see https://www.notion.so/Popover-specs-e9f66fb413414155b01fe10a9ac36387

sofokli1:
## What's this?
What the tittle says

### Relationships
Closes #4499

## Why do we want this?
We want modal views to conform to our ipad design rules

## How is it done?
By adding some logic to `ModalPresentationController` and `ModalDialogPresentationController`

## :squid: Permissions
:squid:


sofokli1:
@dshr Regarding 1. The jump is due to the keyboard showing. Since it is showing it triggers the with keyboard behaviour. Try having the full size virtual keyboard enabled and let me know how it looks.
2. will fix

dshr:
Can we animate the jump though? So at least it animates up with the KB?

sofokli1:
@dshr Take a look again when possible

PR #4556, opened 2019-03-07
Set initial tab item after activity restarts

### LINKED ISSUE #4489: Calendar bug: The tab sometimes is out of sync
It happens sometimes that when you leave the application and get back to it, the bottom navigation bar will have a tab selected, but the main fragment is the current fragment.

juanlaube:
## What's this?
After the main tab bar activity is restarted, whatever the reason, the selected tab item didn't match the fragment. This solves the problem by always setting it to the main fragment.

### Relationships

Closes #4489 

## Why do we want this?
🚫 🐛 

## How is it done?
By setting the selected item to the main tab item.

### Why not another way?
The other way could be setting the fragment instead of the tab bar item, but this is easier and safer.

### Side effects
None.

## Review hints
Try it, an easy way to force the app to restart is disabling the calendar permissions from the AppInfo.

## :squid: Permissions
🦑


juanlaube:
@daividssilverio what's the issue with those steps?

daividssilverio:
> @daividssilverio what's the issue with those steps?

It's still unsynced.

PR #4563, opened 2019-03-08
iPad: Signup view

### LINKED ISSUE #4502: iPad version: Sign Up
Check form width and its behaviour with the KB both hidden and extended.

dshr:
## What's this?
A set of AL changes to make the signup page work on iPad.

### Relationships
Closes #4502

## Why do we want this?
iPad version.

## How is it done?
Same as the login page #4549 

### Why not another way?
Seems simple enough.

### Side effects
None.

## Review hints
Run on iPad and on iPhone.

## :squid: Permissions
iPad crew only


Toggl.Daneel/ViewControllers/SignupViewController.cs
```diff
@@ -33,10 +33,30 @@ public sealed partial class SignupViewController : ReactiveViewController<Signup
...
         private const int emailTopConstraintWithKeyboard = 12;
 
+        private const int tabletFormOffset = 246;
+        private const int tabletLandscapeKeyboardOffset = 80;
+
         public SignupViewController() : base(nameof(SignupViewController))
         {
         }
 
+        public override UITraitCollection TraitCollection
```
sofokli1: same as the login PR
why override and forece `Regular` this?

PR #4576, opened 2019-03-08
📝 Update Siri text copies

### LINKED ISSUE #3382: Siri: Review UI and response copies
We should list all texts used as Siri' responses and in the UI extension, including buttons and intent types (the `Do` may not be the best option) and then someone should review those texts and see if they can be improved or are correct.

The thing is I wrote them my self as temporal texts while development, so we should probably read them again.

### LINKED ISSUE #4538: Siri: Fix message and UI for entries with empty description
The message says "You've been doing for { TIME }", so it's missing what I've been doing.

The UI shows just the time with nothing to the side, we should probably put some kind of placeholder?

img_44e6c6b3432b-1

khoi:
## What's this?
Change Siri responses text.

### Relationships
Closes #3382 
Closes #4538

### Side effects
None

## Review hints
None beside playing with it.

## :squid: Permissions
🎙


Toggl.Daneel/Services/IntentDonationServiceIos.cs
```diff
@@ -82,24 +82,31 @@ public void DonateShowReport(ReportPeriod period)
             {
                 case ReportPeriod.Today:
                     intent.Period = ShowReportPeriodReportPeriod.Today;
+                    intent.SuggestedInvocationPhrase = $"Show ${Resources.Today.ToLower()}'s time";
```
Odrakir: The `$` shouldn't be part of the phrase.

Odrakir:
Oh! Something else I forgot to mention: The message when trying to stop a timer and there's no entry running should be an error now that we have fixed the custom response bug.

Toggl.Daneel/Services/IntentDonationServiceIos.cs
khoi:
```suggestion
                    intent.SuggestedInvocationPhrase = $"Show {Resources.Today.ToLower()}'s time";
```

PR #4579, opened 2019-03-08
Clarify a release testing step

arturs105:
## What's this?
I was testing the Daneel 1.19 and couldn't get the character limit to show in the start view and after an investigation I found out that it was removed a while ago. This PR explains that test a bit more so hopefully no one has to repeat my investigation in the future.

## :squid: Permissions
🦑


.github/ISSUE_TEMPLATE/d-manual-test-template.md
```diff
@@ -31,7 +31,7 @@ about: A manual testing checklist that can be assigned to a release to track the
 - [ ] Add new project
 - [ ] Add new client
 - [ ] Add new tag
-- [ ] Add long description and confirm character limit error shows
+- [ ] Add long description and confirm character limit error shows in the edit view after sync
```
sofokli1: Maybe also specify the limit so that one can actually reach the limit and see the message?

PR #4587, opened 2019-03-11
Siri: Setup relevant shortcuts

### LINKED ISSUE #3381: Siri: Include relevant shortcuts
Apparently shortcuts won't be suggested in the Siri Watch face unless they are saved as `Relevant Shortcuts`.

We should do that including the appropriate information for it to be suggested. Probably `INDateRelevanceProvider`, not sure if `INDailyRoutineRelevanceProvider`, but sadly no `INLocationRelevanceProvider` as we don't have the user's location.

Reference: https://developer.apple.com/documentation/sirikit/relevant_shortcuts

Odrakir:
## What's this?
Adds Relevant Shortcuts so they appear on the Siri Watch Face.

### Relationships
Closes #3381 

## Why do we want this?
This is the only way to make shortcuts appear in the Siri watch face.

## How is it done?
After a lot of reading I decided this:
- User generated shortcuts, meaning things that the user actually does in the app and we want him to be able to repeat, won't have a relevance provider, we will just donate them whenever the user executes that action in the app. Hopefully Siri will learn of user habits and propose those shortcuts when appropriate.
- Default shortcuts (start and stop empty entries, basically) do include relevance providers, but as there's no way for us to know when will the user use this shortcuts I decided to just pick some common usages for now. We can see how well this works and decide on changes during testing. Basically for now start will happen when user is at work, school or gym and stop when user is at home.

## :squid: Permissions
@khoi


Toggl.Daneel/Services/IntentDonationServiceIos.cs
```diff
@@ -134,43 +161,59 @@ public void ClearAll()
             INVoiceShortcutCenter.SharedCenter.SetShortcutSuggestions(new INShortcut[0]);
```
khoi: you probably want to add
```
INRelevantShortcutStore.DefaultStore.SetRelevantShortcuts(new INRelevantShortcut[0]);
```

Toggl.Daneel/Services/IntentDonationServiceIos.cs
```diff
@@ -120,7 +144,10 @@ public void DonateShowReport()
             var intent = new ShowReportIntent();
             intent.SuggestedInvocationPhrase = Resources.ShowReportsInvocationPhrase;
             var interaction = new INInteraction(intent, null);
-            interaction.DonateInteraction(onCompletion);
+            interaction.DonateInteraction(trackError);
+
+            var shortcut = createRelevantShortcut(intent);
```
khoi: report shouldn't be added as relevant shortcut

Toggl.Daneel/Services/IntentDonationServiceIos.cs
```diff
@@ -54,15 +54,11 @@ public void DonateStartTimeEntry(IWorkspace workspace, ITimeEntry timeEntry)
                 intent.ProjectId = new INObject(timeEntry.ProjectId.ToString(), timeEntry.ProjectId.ToString());
                 intent.Tags = timeEntry.TagIds.Select(tag => new INObject(tag.ToString(), tag.ToString())).ToArray();
                 intent.Billable = new INObject(timeEntry.Billable.ToString(), timeEntry.Billable.ToString());
-<<<<<<< HEAD
-                intent.SuggestedInvocationPhrase = timeEntry.Description;
+                intent.SuggestedInvocationPhrase = $"Track {timeEntry.Description}";
```
khoi: this is redundant since it's been added into the main feature branch, but it's fine.

khoi: oh, this is your merge commit, I see

PR #4592, opened 2019-03-11
Fix identity comparison for diffing

### LINKED ISSUE #4578: Consider replacing `LogItemViewModel.Identity` with a function
Our IDs use hash codes and there could potentially be collisions which would lead to crashes or undefined behavior (depending on platform).

The idea is to remove the `long Identity { get; }` from the `IDiffable` and replace it with `bool RepresentsSameItem(IDiffable other)` which would do whatever is necessary to determine whether it is the same item or not (even if the contents is different/updated).

There is a potential problem with the custom diffing algorithm which internally uses dictionaries indexed by the `long Identity`, but it should be easy (and still performant) to replace the key with anything implementing `IEquatable<T>`.

simonrozsival:
## What's this?
This PR replaces mandatory `long`s used for identifying rows with a generic `TKey : IEquatable<TKey>` type.

### Relationships
Closes #4578 

## Why do we want this?
We don't want to match entities incorrectly.

## How is it done?
The code is made generic and I introduce 3 key types for the 3 different main log elements - day sections (days), groups, individual time entries.

### Why not another way?
Generics FTW

### Side effects
Some parts of the code become a bit ugly :-/ Namely https://github.com/toggl/mobileapp/compare/project/grouped-time-entries...simon/identity?expand=1#diff-21c857bd05d2e1f894fe7739a403b98eR111 and https://github.com/toggl/mobileapp/compare/project/grouped-time-entries...simon/identity?expand=1#diff-631cb9415489154e03e5d203ac6a7f52R22.

The Android part should be refactored soon so I don't care about that too much. If you have an idea how to tame the crazy generic definition in the iOS binding, please tell me about it.

## Review hints
Commit by commit should make more sense than going through the file changes.

Please test in the iOS simulator and Android emulator - and confirm, that it works as expected in your review.

## :squid: Permissions
🦑 once 🍎 and 🤖 are confirmed to work well


kornelijepetak:
I will not be able to take a look at this right now, but you can alias the types into something more readable.

Note however to do this inside the namespace declaration, not with all other `using`s, because otherwise you will have to fully qualify all of them.

Odrakir:
Man the lack of self referencing type constraints in C# generics is killing me.

PR #4606, opened 2019-03-11
Fix android calendar settings

### LINKED ISSUE #4596: The two columns layout won't always be in sync with settings
To replicate:

1. Accept the calendar permissions at least once.
2. Link at least one calendar.
3. Close the app (optional).
4. Remove the calendar permissions from the app in the system settings.
5. Re open the app.
6. Go to the settings page.
7. Click calendar calendars button.
8. Enable some calendars.
You should see the previously selected calendars already selected (not good) in the calendars list.
9. Press done.
Go to the calendars page. You shouldn't see the two columns layout yet, even though you have calendars enabled. (That's the main bug). If you press the `Link calendars` button, it will sync and the two columns will appear immediately.

The source of this problem is probably the same as #4557.

### LINKED ISSUE #4604: Calendar list is always visible on calendar settings
Steps to replicate: Follow steps on #4596 and you should see it.

### LINKED ISSUE #4605: Previously selected calendars are still selected when permissions are changed
Steps to replicate: follow steps on #4596 and you should see it.

daividssilverio:
## What's this?
This fixes two bugs I found while trying to fix #4596

### Relationships
Closes #4605 and Closes #4604 

## Why do we want this?
1. For #4605, it doesn't make sense to hold reference to the previously selected calendars when the calendar permissions change outside of the context of the app.
2. For #4604, the `<merge>... </merge>` usage made the `CalendarSettingsActivity` to not actually be able to find the view with id `CalendarsContainer`, because the root element of the `SelectUserCalendarsBase.axml` was't there. The null pointer exception was being swallowed and the container visibility wasn't being updated, that's it.

## How is it done?
The layout was fixed with another layout strategy, making sure that it still works and the effect of the fix from #4580 was kept.

### Why not another way?
I'm sure there's bunch of other ways, this one works.

### Side effects
```
if (!PermissionGranted)
            {
                UserPreferences.SetEnabledCalendars();
            }
```
this ☝️piece of code sets previously selected calendars to null when permissions are revoked (this won't happen at runtime, both android and iOS reset the apps when permissions are revoked directly from the system settings)

## Review hints
Try to replicated it following the steps on #4596.

## :squid: Permissions
🦑 it


Toggl.Foundation.Tests/MvvmCross/ViewModels/CalendarSettingsViewModelTests.cs
```diff
@@ -4,6 +4,7 @@
 using System.Reactive;
 using System.Reactive.Linq;
 using System.Threading.Tasks;
+using Castle.Core.Internal;
```
juanlaube: 👀

Toggl.Giskard/Resources/layout/SelectUserCalendarsBase.axml
```diff
@@ -1,5 +1,9 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
-<merge
+<LinearLayout
```
juanlaube: This breaks the layout for `SelectUserCalendarsFragment`. we should keep the merge and wrap the include in `CalendarSettingsActivity.axml` with a FrameLayout and an id.

juanlaube: Ignore this comment, everything work as expected

PR #4609, opened 2019-03-12
Add tracking related to Grouped Time Entries

### LINKED ISSUE #4403: Collect analytics data for grouped time entries
Some ideas of what to track:
- how many users have grouping enabled
- how many users enabled grouping using the app
- how many users disabled grouping using the app
- what is the percentage of time entries which belong to some group in the log (of all of the "2 months" of time entries)

Please suggest more ideas in comments and feel free to point out which of my ideas don't seem worth tracking.

We should have a finalized list of metrics we want to collect before we start implementing this. I think this issue should be replaced with a new one a the final list.

kornelijepetak:
## What's this?
The title says it all.

### Relationships
Closes #4403 

## Why do we want this?
To help users help us help them help us all.

## How is it done?
By introducing two new events and expanding on the existing one.

### Why not another way?
The only alternative I'm thinking about is to avoid tuples in subjects/observables and introduce named types. Do you think the code would be more readble? Please post your comments.

### Side effects
➕ More info in analytics consoles.
❓ I have deprecated one of the enum values (`Continue`) which will no longer appear in analytics with new events. Instead, it is separated into several possible continuation ways so that we can get a more detailed idea about it. Unfortunatelly, we'll have to do some manual math to get to some more meaningful numbers. But we can also just look at the data retrieved after this change so 🤷‍♂️.

## Review hints
Easier to do commit-by-commit.

## :squid: Permissions
🔟 people.


kornelijepetak:
Yes, unfortunatelly, it's not gonna be any nicer with lots of methods that just call another with various arguments 🤷‍♂️ 

Cross-cutting concerns are always ugly. 😢

simonrozsival:
We should re-visit how we track data. Ideally we could do that with just some attributes... and without any unit tests.

PR #4619, opened 2019-03-12
Add check for sync conflict

### LINKED ISSUE #3384: Siri: Review rules to avoid sync errors
This was mentioned in the specification but we never got around to do it properly.

We should store the last edit time of the latest entry (that's the weirdly name `at` property) in the shared storage. And compare that to the `at` property of the latest entry in the server.

- If local at time <= server at time: Everything is in sync. Do action.
- If local at time > server at time: There's an unsynced entry, we can continue the shortcut in the app, but not online directly. Not sure if we can do that or we just have to force the user to open the app and try again.

There are many more edge cases, but I'm not sure if it's worth it to handle them or if it's worth it to do it now. Like, what happens if there's a running entry in the server but not in the app?

Odrakir:
## What's this?
This adds some checks in the confirmation phase of the intent handling. The idea is to see if the data in the app is outdated and if so force the user to open the app before continuing.

### Relationships
Closes #3384 

## Why do we want this?
In theory this won't happen too often, because it requires using the app offline, then updating the server some other way and then using Siri (which requieres internet), but still, it's good practice checking for this errors.

## :squid: Permissions
🎤


Toggl.Daneel.SiriExtension/StopTimerIntentHandler.cs
```diff
@@ -33,42 +34,66 @@ public override void ConfirmStopTimer(StopTimerIntent intent, Action<StopTimerIn
...
+                .Select(getRunningTimeEntry)
+                .Subscribe(
+                runningTE =>
+                {
+                    runningEntry = runningTE;
+                    completion(new StopTimerIntentResponse(StopTimerIntentResponseCode.Ready, null));
+                },
+                exception =>
+                {
+                    SharedStorage.instance.AddSiriTrackingEvent(SiriTrackingEvent.Error(exception.Message));
```
amulware: This'll let us know how often this type of error happens?

Odrakir: Yesssir. Through the message.

Toggl.Daneel.SiriExtension/StartTimerIntentHandler.cs
```diff
@@ -27,7 +28,22 @@ public override void ConfirmStartTimer(StartTimerIntent intent, Action<StartTime
                 return;
             }
 
-            completion(new StartTimerIntentResponse(StartTimerIntentResponseCode.Ready, null));        }
+            var lastUpdated = SharedStorage.instance.GetLastUpdateDate();
+            _ = togglAPI.TimeEntries.GetAllSince(lastUpdated)
```
juanlaube: This disposable is collected after it goes out of scope, right? How is this working? 🤔 
Also please don't use `_`

Toggl.Daneel.SiriExtension/StopTimerIntentHandler.cs
```diff
@@ -33,42 +34,66 @@ public override void ConfirmStopTimer(StopTimerIntent intent, Action<StopTimerIn
                 return;
             }
 
-            completion(new StopTimerIntentResponse(StopTimerIntentResponseCode.Ready, null));
+            var lastUpdated = SharedStorage.instance.GetLastUpdateDate();
+            _ = togglAPI.TimeEntries.GetAll()
```
juanlaube: Same question here 👀

PR #4620, opened 2019-03-12
Implement handling of UI elements for Inaccessible Workspaces

### LINKED ISSUE #4570: Disable editing inaccessible time entries
Inaccessible time entries should not be editable. Rxfactoring broke that.

kornelijepetak:
## What's this?
It just prevents editing of time entries belonging to inaccessible workspaces.
Well, all except project (because changing the project to another that belongs to an accessible workspace will reenable editing).

### Relationships
Closes #4570 

## Why do we want this?
So that there's no mess.

## How is it done?
Just some RX and UI handling

### Why not another way?
Come on!

### Side effects
None.

## :squid: Permissions
🔟


Toggl.Daneel/ViewControllers/EditTimeEntryViewController.cs
```diff
@@ -310,18 +310,13 @@ private void adjustUIForInaccessibleTimeEntry(bool isInaccessible)
 
             DescriptionTextView.TextColor = textColor;
 
-            foreach (var label in getLabelsToChangeColorWhenEditingInaccessibleEntry())
+            new UILabel[]
```
heytherewill: Not even sure you need to instantiate this array. Since it is not used anywhere, adding `.TextColor = textColor` to each line would literally make it smaller

PR #4625, opened 2019-03-12
Calendroid: Fix calendar columns update

### LINKED ISSUE #4596: The two columns layout won't always be in sync with settings

daividssilverio:
## What's this?
This fixes one edge case: When the user revokes the calendar permissions and tries to access the calendar page after enabling the calendar integration from the app settings, as described here #4596.

### Relationships
Closes #4596

## Why do we want this?
To have the calendar view columns match the actual state of the calendar integration (enabled/disabled).

## How is it done?
I stole the idea from #4615, but basically we are checking for the calendar permissions when the view appears, that's it.

### Why not another way?
We could make an effort at some point to actually change how we handle the permissions, maybe check for permissions every time the app comes to foreground and update shared preferences as necessary. We could do that on the `ReactiveActivity` and register which permissions we would like to check on a activity basis. But that's a lot more work and needs more thought. Just pinging @heytherewill so we can maybe touch this again later.

### Side effects
None.

## Review hints
Give it a try. Follow the steps from the issue.

## :squid: Permissions
🦑 it, don't wait for me.


Toggl.Foundation.MvvmCross/ViewModels/Calendar/CalendarViewModel.cs
```diff
@@ -231,6 +235,22 @@ public async override Task Initialize()
                 .DisposedBy(disposeBag);
         }
 
+        public override void ViewAppeared()
+        {
+            base.ViewAppeared();
+            checkCalendarPermissions();
```
juanlaube: I don't like that we repeated this pattern in multiple view models, we should change this in the future.

PR #4633, opened 2019-03-13
Make sure keyboard does not hide the "password reset successful" message

### LINKED ISSUE #4571: Keyboard hides the "password reset successful" message
## 🐛 Bug Report
The keyboard isn't dismissed after pressing "Get password reset link", which leads to the success message not being visible. This only happens on an iPhone, on simulator it works fine.

## Context
Build/Branch: latest `develop` (tested on Daneel 1.19 from Testflight)
Phone OS: iOS 12
Phone model: iPhone X

## Expected behavior
The keyboard should be immediately dismissed after pressing the "Get password reset link".

## To reproduce
Do this on an iPhone, since this does not happen on a simulator
1. Go to the password reset view
2. Enter a valid e-mail
3. Press "Get password reset link"

arturs105:
## What's this?
This fixes the bug described in the issue.

### Relationships

Closes #4571

## How is it done?
`ResignFirstResponder()`

## Review hints
Test this out on an iPhone and see if the bug is fixed. Using an iPhone is important, because this does not happen on a simulator.

## :squid: Permissions
🦑


Toggl.Daneel/ViewControllers/ForgotPasswordViewController.cs
```diff
@@ -93,6 +89,8 @@ public override void ViewDidLoad()
             ResetPasswordButton.Rx()
                 .BindAction(ViewModel.Reset)
                 .DisposedBy(DisposeBag);
+
+            ResetPasswordButton.TouchUpInside += resetPasswordButtonTapped;
```
heytherewill: Please make this reactive

Toggl.Daneel/ViewControllers/ForgotPasswordViewController.cs
```diff
@@ -178,5 +180,8 @@ private void prepareBackbutton()
             backButton.Frame = frame;
             backButton.HorizontalAlignment = UIControlContentHorizontalAlignment.Left;
         }
+
+        private void resetPasswordButtonTapped(Unit _)
```
heytherewill: There is an extension that allows for making this a `void -> void`

PR #4643, opened 2019-03-13
GTE: reload headers of sections which were somehow changed

### LINKED ISSUE #4642: Diffing: section headers aren't diffed
Related to #4564 and #4614 

We don't use the `UpdatedSections` property of `Diffing<>.Changeset` and we can't correctly display updated headers (the dates don't change automatically, we reload the whole table "on midnight" and when the user changes timezones - significant time change; the duration sums don't update when) a new TE is created in an existing section.

Another problem is that the `changeset.FinalSections` doesn't contain the correct section header. This might most likely not be an issue if we used the `UpdatedSections`. I captured this problem in the debugger:

image

### LINKED ISSUE #4719: Time entry log headers are incorrect
I believe this is related to #4655 

image

Steps to reproduce:
- open the app and let it sync
- create a new TE which will be in a new section on top of the log (e.g., select a start data which is after the date of the first section)
- the date of the first and second sections will both show the date and duration of the first section

### LINKED ISSUE #4724: App sometimes crashes when swipe-deleting TEs
I "stress tested" the app by adding and continuing and swipe deleting TEs and after one swipe-to-delete the app froze for a few seconds and then it showed me this exception which was thrown at this spot:

image

Apparently we update the underlying data structure before we diff and update the UI which still tries to display some of the sections and rows.

simonrozsival:
## What's this?
This is as straightforward solution to a problem with diffing section headers (we don't diff section headers).

### Relationships
Closes #4642, closes #4719, closes #4724 

This is an alternative solution to #4720. The original PR #4655 seemed like a better solution, because it required fewer code changes, but it didn't work properly and with the changes which are needed to make it work I don't think it is well encapsulated and that it might easily break if somebody touches that code. This solution makes the header changes explicit and easy to understand.

## Why do we want this?
We want to reliably show correct data to our users.

## How is it done?
Update section headers which changed and should be re-rendered.

### Why not another way?
I tried changing the diffing algorithm itself (setting the "updated sections" ids of the changeset but I couldn't make it work correctly). We can create an issue and look into it later.

This is what I tried:

```cs
foreach (var difference in changeset)
{
    reactive.Base.BeginUpdates();
    dataSource.SetSections(difference.FinalSections);
    reactive.Base.performChangesetUpdates(difference);
    reactive.Base.EndUpdates();
}

// Make sure the headers are up-to-date
var chagnedSections = changeset.SelectMany(difference =>
    difference.DeletedItems.Select(item => item.sectionIndex)
        .Concat(difference.InsertedItems.Select(item => item.sectionIndex))
        .Concat(difference.MovedItems.SelectMany(item =>
            new[] {item.Item1.sectionIndex, item.Item2.sectionIndex})))
        .Distinct();

foreach (var i in chagnedSections)
{
    if (reactive.Base.GetHeaderView(i) is BaseTableHeaderFooterView<THeader> headerView)
    {
        headerView.Item = finalSections[i].Header;
    }
}
```

This creates a problem: when you move an item from one section to a new section which will appear _before_ the old section (and therefore shift it’s index). The `changedSections` enumerable will contain the old index of the section and the header of the day won't be reloaded (and the duration will be incorrect).

### Side effects
The code is not perfect as I couldn't fit it into `Diffing`.

## Review hints
Try this in the simulator.

## :squid: Permissions
🦑


simonrozsival:
Superseded by #4655

Toggl.Foundation.MvvmCross/ViewModels/TimeEntriesLog/DaySummaryViewModel.cs
```diff
@@ -19,5 +19,22 @@ public DaySummaryViewModel(DateTime day, string title, string totalTrackedTime)
...
         }
+
+        public bool Equals(DaySummaryViewModel other)
+        {
+            if (ReferenceEquals(null, other)) return false;
+            if (ReferenceEquals(this, other)) return true;
+            return string.Equals(Title, other.Title) && string.Equals(TotalTrackedTime, other.TotalTrackedTime) && Equals(Identity, other.Identity);
+        }
+
+        public override bool Equals(object obj)
```
kornelijepetak: This can be simplified to:

```suggestion
        public override bool Equals(object obj) => (obj as DaySummaryViewModel);
```

simonrozsival: yeah, I can do that. I used the code generated by Rider because of laziness :)

simonrozsival:
Superseded by #4754

PR #4654, opened 2019-03-14
GTE: Open change date view instead of wheel view when tapping the Start date

### LINKED ISSUE #4645: Edit time entry: the wheel view is shown instead of "select date" view
This was lost in the rxfactoring - when the user taps on the start date, he/she should be able to open the small modal for changing dates, but instead we show the wheel view. Now the wheel view is shown for everything - start, end, duration, and start date 😄 apparently this is a bug.

simonrozsival:
## What's this?
Bring back functionality that was lost during rxfactoring of the edit view.

### Relationships
Closes #4645

## Why do we want this?
🔥 🐛 

## How is it done?
I basically copied code from the old VM code and rxfactored it.

### Side effects
There's a difference in behavior between iOS and Android - Android doesn't have this filed for selecting dates. It used to be this way so let's not change it in this project.

## Review hints
Run it in the iOS simulator.

## :squid: Permissions
🦑


sofokli1:
is the bounce normal?
Mar-14-2019 14-03-49

simonrozsival:
@sofokli1 yes, that bounce happens if you have the sw keyboard disabled in the simulator. It would be great to fix that, but I suppose nobody ever complaned about that.

Issue #4614, opened 2019-03-12
Daily total in timer view does not update after adding more than one time entry 


maritoggl:
## 🐛 Bug Report
Daily total in main timer view does not update after adding new time entries after the first one. 

## Context
While testing the grouped time entries feature on iOS, noticed that when I added a first time entry of the day, the daily total duration was valid. After adding a second, third etc time entry, the daily total did not update and stayed as the total of the first time entry that was created. 

Tested also with 1.19 and couldn't reproduce the same issue, so should be specific to 1.20 only I think. 

Build/Branch: 1.20
Phone OS: 12.1.4
Phone model: iPhone 7

## Expected behavior
Daily total updates when adding new time entries to the day.

## To reproduce
1. Create a time entry
2. Check the daily total - all should be good and matching
3. Create another time entry to the same day
4. Check the daily total - it still shows as the total of the first time entry that was created

simonrozsival:
This is related to #4564. The durations are calculate correctly, but the headers are not reloaded by the diffing algorithm.

simonrozsival:
Fix was merged into the project branch.

PR #4655, opened 2019-03-14
Make section headers observable

### LINKED ISSUE #4642: Diffing: section headers aren't diffed

Odrakir:
## What's this?
This changes the way headers are populated in the main view, because we need them to update when items in the section change.

### Relationships
Closes #4642 

## Why do we want this?
So the total duration in the headers updates when it has to.

## How is it done?
I create an observable for each heather from the main TimeEntries observable.

## :squid: Permissions
Do it


Toggl.Daneel/Cells/TimeEntriesLog/TimeEntriesLogHeaderView.cs
```diff
@@ -31,6 +34,12 @@ public override void AwakeFromNib()
             DurationLabel.Font = DurationLabel.Font.GetMonospacedDigitFont();
         }
 
+        public override void PrepareForReuse()
+        {
+            base.PrepareForReuse();
+            DisposeBag = new CompositeDisposable();
```
simonrozsival: Shouldn't we call `Dispose` on the old dispose bag before throwing it away? Wouldn't this possibly introduce a memory leak? 🤔 I'm not sure.

Odrakir: You re correct, that was the Swift developer in me.

heytherewill: You could use a SerialDisposable which does exactly that: Disposes the old disposable when setting the new one. IMO it would convey the meaning a lot better

simonrozsival:
This PR also closes #4614

PR #4656, opened 2019-03-14
Make sure bar charts are initially shown

### LINKED ISSUE #4613: Report bar charts missing on app stratup
## ⛑️ User Support

A user reported that the bar charts are missing on the Report page in the app. We tested this on our end and found that if you change the date range or click anywhere else that's clickable and go back to reports the bar charts will appear. 

## About the user

App version: Daneel 1.18
Phone OS: iOS 12.1.4
Phone model: iPhone 6s

## Expected behaviour
The bar chart should be there on app startup when it's first viewed without having to navigate away from it or change the date range

## To Reproduce
1. Close out the app completely if you can see the bar chart at the moment, if you can't see it skip this step,
2. Go to check out the Repor - the bar chart should be missing
3. Change the date range or click on the calendar or navigate away and go back to the report page and the bar chart should appear.

NOTE: if you close out the app and re-start it the bar chart should be gone again.

arturs105:
## What's this?
This fixes the bug described in the issue.

### Relationships

Closes #4613

## How is it done?
The bug was caused due to the observable emitting items before the view was able to subscribe, so when the view subscribed, it didn't get any data. To fix this i used `Replay(1)` so every subscriber gets the latest value upon subscription.

### Why not another way?
🤷‍♂️ This seems to work.

## Review hints
Run the app and see if the bug is gone.

## :squid: Permissions
🦑


sofokli1:
what I mean is, can it update without going to the grey state?

arturs105:
Something can definitely be done in that regard, but it isn't related to this issue or PR, since that happens on `develop` as well. Feel free to create an issue for that.

PR #4659, opened 2019-03-14
Show start time entry view when tapping on an empty space

### LINKED ISSUE #4647: The calendar view shouldn't create a time entry
## 🐛 Bug Report
When tapping on an empty space in the calendar view, it creates a time entry in the database and opens the edit time entry view. Instead it should open the start time entry view and only create the time entry when it's saved.

## Context

Build/Branch: `project/calendroid`
Phone OS: Android
Phone model:

## Expected behavior
It should open the start time entry view and only create the time entry when it's saved, the new time entry should be visible immediately in the calendar view.

## To reproduce
Tap on an empty space in the calendar view.

juanlaube:
## What's this?
This shows the start time entry view instead of creating a time entry and then openning the edit view when tapping on an empty space in the calendar view.

### Relationships

Closes #4647 

## Why do we want this?
We want to create the time entries only when the user saves them, not when tapping on the calendar view.

## How is it done?
Navigating to the start time entry view with some params.

### Why not another way?
It's the only way.

### Side effects
None.

## Review hints
Try it.

## :squid: Permissions
🦑


Toggl.Foundation.MvvmCross/ViewModels/Calendar/CalendarViewModel.cs
```diff
@@ -78,6 +79,8 @@ public sealed class CalendarViewModel : MvxViewModel
 
         public InputAction<(DateTimeOffset, TimeSpan)> OnDurationSelected { get; }
 
+        public InputAction<DateTimeOffset> OnCreateTimeEntryAtOffset { get; }
```
heytherewill: This is not an Action name. This is named like a callback

juanlaube: True, I will remove the `On` prefix.

PR #4677, opened 2019-03-15
GTE: Make main-log footers non-sticky

### LINKED ISSUE #4660: Weird white bar in main log
When I was debugging the fade view and I had my TEs' descriotions highlighted with red background color, I noticed that there is a white floating line close to the bottom of the screen. It might be a white opaque footer (introduced in a PR solving #4224) or it's some other element.

image

Investigate and eliminate.

simonrozsival:
## What's this?
This PR does quite a lot for a few changed lines... It removes the default sticky header (which can't be made non-sticky while keeping the header sticky) and replaces it with a custom "section separator" which is the last row and it scrolls with the section as you would expect.

### Relationships
Closes #4660 

## Why do we want this?
We want to have non-sticky footers to hide whatever is under them.

## How is it done?
An extra row + a comment which explains the situation.

### Why not another way?
I tried several approaches and it was all ugly hacks. This one is the most explicit and I think it might not be necessarily easy to break.

### Side effects
The TEs log table view source becomes more complex.

## Review hints
Try it in the simulator.

## :squid: Permissions
🦑


simonrozsival:
Superseded by #4681

PR #4680, opened 2019-03-15
Fix project and tag creation cells in start view

### LINKED ISSUE #4673: Project creation cell in start time entry view shows wrong name
## 🐛 Bug Report
When creating a project in start time entry view, the project name displayed in the creation cell is wrong. There are 2 things that shouldn't happen:
* The project name includes the `@` symbol
* The project name includes any text that appears before the `@` symbol.

## Context
Phone OS: iOS 12
Phone model: iPhone 8 simulator

Maybe happens on android as well.

Checked the released Daneel 1.19 as well and that version does not contain this bug. I suspect this bug was caused by the RxFactor of that view.

## Expected behavior
The creation cell should only show the project name

## To reproduce
1. Enter start time entry view
2. Enter text something like: `entry description @new project`
3. The creation cell will show: `Create project "entry description @new project"`, but it should show `Create project "new project"`

arturs105:
## What's this?
This fixes the bug described in the issue, and a bonus bug: the same problem, but with tags.

### Relationships

Closes #4673

## Review hints
Run the app and see that the bug is fixed.

❗️ Important ❗️ 
I wasn't able to test this on android, since VS and the emulator weren't cooperating, so, whoever reviews this: test this out on android as well.

## :squid: Permissions
🦑


arturs105:
Since i'm on vacation from 18.03 till 22.03, i hereby give permission to anyone from mobile team to make any requested changes to this PR.

simonrozsival:
I'm not @toggl/androids but I tried it in the simulator/emulator on both iOS and Android and it works well:

before ... after
image

Let's merge this PR.

PR #4687, opened 2019-03-15
Fix calendar events background colour on Android Api Pre-23

### LINKED ISSUE #4667: Calendar events are all white on Android 21-22
How to replicate:

Try the app on android 21-22.
Both calendar events and time entries are all white.

Screenshot for reference.
Screen Shot 2019-03-14 at 6 07 33 PM

daividssilverio:
## What's this?
This fixes how we set the events background colour, such that it works on Android api < 23.

### Relationships
Closes #4667 

## Why do we want this?
Because it was all white on android 21 and 22. Apparently setting the background tint on those API levels doesn't work as expected. I've changed to change the background drawable colour directly instead.

## How is it done?
Check the code, I've explained ☝️  as well.

### Why not another way?
Because this works for both api-pre-23 and api-23+.

### Side effects
None.

## Review hints
Give it a try, on an emulator with api level 21 or 22.

## :squid: Permissions
🦑 it

### DIFF
--- a/Toggl.Giskard/ViewHolders/CalendarEntryViewHolder.cs
+++ b/Toggl.Giskard/ViewHolders/CalendarEntryViewHolder.cs
@@ -1,5 +1,6 @@
 using System;
 using Android.Graphics;
+using Android.Graphics.Drawables;
 using Android.Runtime;
 using Android.Util;
 using Android.Views;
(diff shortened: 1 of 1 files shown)


daividssilverio:
Superseded by #4689

PR #4697, opened 2019-03-18
Fix non expandable groups

### LINKED ISSUE #4695: Groups with tags can't be expanded
The groups that have tags can't be expanded because we are comparing group equality by comparing tags the wrong way.

simonrozsival:
## What's this?
It wasn't possible to expand groups which had some tags because of incorrect equality check of group IDs (we used reference equality for `long[]`).

### Relationships
Closes #4695 

## Why do we want this?
We want good UX.

## How is it done?
Use `HashSet<long>` and `SetEquals` instead of `long[]` to fix comparison. Plus tests.

## :squid: Permissions
🦑


Toggl.Foundation.Tests/MvvmCross/ViewModels/TimeEntriesLog/GroupIdTests.cs
```diff
@@ -0,0 +1,116 @@
...
+
+namespace Toggl.Foundation.Tests.MvvmCross.ViewModels
+{
+    public sealed class GroupIdTests
+    {
+        public sealed class TheEqualsMethod
+        {
+            [Theory]
+            [MemberData(nameof(MatchingPairs))]
+            public void Equal(IThreadSafeTimeEntry timeEntryA, IThreadSafeTimeEntry timeEntryB)
```
kornelijepetak: Maybe the names of the test should follow our usual conventions?

```suggestion
            public void CorrectlyCalculatesEquality(IThreadSafeTimeEntry timeEntryA, IThreadSafeTimeEntry timeEntryB)
```

Toggl.Foundation.Tests/MvvmCross/ViewModels/TimeEntriesLog/GroupIdTests.cs
```diff
@@ -0,0 +1,116 @@
...
+            {
+                var groupA = new GroupId(timeEntryA);
+                var groupB = new GroupId(timeEntryB);
+
+                groupA.Equals(groupB).Should().BeTrue();
+            }
+
+            [Theory]
+            [MemberData(nameof(NonMatchingPairs))]
+            public void DoNotEqual(IThreadSafeTimeEntry timeEntryA, IThreadSafeTimeEntry timeEntryB)
```
kornelijepetak: The naming?

PR #4710, opened 2019-03-18
Fix notification unscheduling and double schedules

### LINKED ISSUE #4652: Smart alerts don't respect selected calendars
## 🐛 Bug Report
I received one smart alert notification from an event that didn't make part of the calendars selected to be integrated with the app through the app settings.

## Context
Found it while doing tests for #4601

Build/Branch: `project/calendroid`
Phone OS: Android
Phone model: S8

## Expected behavior
Only events from the calendars selected in the calendar settings should receive smart alerts at the times set in the calendar smart alerts settings.

## To reproduce
Make sure you have at least 2 calendars in your system (`A`, `B`).
1. Create two events, `a` and `b`. (`a` in the `A` calendar, `b` in the `B` calendar)
2. Enable the calendar integration in the app, selecting both calendars
3. Make sure you see the events in the calendar
4. Set your calendar smart alerts settings such that you can receive the events in reasonable time (or find some adb magic, up to you).
5. Now go to the calendar settings, de-select the `B` calendar.
6. Go back to the settings page.
7. Don't go back to the calendar page.
8. Close the app, leave it be, just wait for time of your smart alerts
9. You will receive the notifications for both events `a` and `b`.

## Suggestions
It seems this is just a matter of updating the scheduled events correctly when the calendars integrated with the app changes. We have to unschedule previously scheduled events from calendars that are not integrated anymore.

daividssilverio:
## What's this?
This fixes the pending intent unscheduling in the `NotificationServiceAndroid`.

### Relationships
Closes #4652 

## Why do we want this?
Because we actually want to cancel notifications pending intents when changing smart alerts settings and integrated calendars.

## How is it done?
We now save enough state to recover information about the pending intents created when the smart alerts are scheduled, basically all we id is the notification id.

We create the PendingIntent's Intent differently, passing an Uri with the notificationId, such that the `PendingIntent.GetBroadcast(..., ..., scheduledNotificationIntent, PendingIntentFlags.NoCreate)` can be used to retrieve existing PendingIntents and the `AlarmManager` is able to cancel them.

I've also added another check to cancel existing pending intents during the schedule phase, because the `PendingIntentFlags.CancelCurrent` wasn't working as expected. I believe this is due to some race condition and the `INotificationService` being called too many times too fast.

More details on how this works: check https://developer.android.com/reference/android/app/PendingIntent

### Why not another way?
In theory:
`var pendingIntent = PendingIntent.GetBroadcast(Application.Context, notificationId, scheduledNotificationIntent, PendingIntentFlags.CancelCurrent);` should be enough to cancel existing matching events (with matching a `PendingIntent`). Thing is, as described [Here](https://developer.android.com/reference/android/content/Intent.html#filterEquals), the `PendingIntent` uses the `filterEquals` to check for matching intents to take some action based on the `PendingIntentFlags` used; and our intents didn't have anything that them unique.

From the docs:

>public boolean filterEquals (Intent other):
>Determine if two intents are the same for the purposes of intent resolution (filtering). That is, if their action, data, type, class, and categories are the same. This does not compare any extra data included in the intents.

### Side effects
None really.

## Review hints
Not easy to see this happening actually. 
But you can check the scheduled alarms intents to `SmartAlertCalendarEventBroadcastReceiver` in the system using (remove `wc -l` if actually want see the lines):
```
adb shell dumpsys alarm | grep SmartAlertCalendarEventBroadcastReceiver | wc -l
```
Check that on `project/calendroid` as you change enabled calendars/smart alert settings. You should see the number increasing as you play with the settings, until the apps crashes when you reach the system limit.

Then check the behaviour on this branch.

## :squid: Permissions
🦑


Toggl.Giskard/Services/NotificationServiceAndroid.cs
```diff
@@ -46,18 +62,64 @@ public IObservable<Unit> UnscheduleAllNotifications()
...
+        {
+            var oldIntent = PendingIntent.GetBroadcast(Application.Context, notificationId, scheduledNotificationIntent, PendingIntentFlags.NoCreate);
+
+            if (oldIntent == null) return;
+
+            alarmManager.Cancel(oldIntent);
+        }
+        
+        private static Uri getSmartAlertIdentifierUri(int notificationId)
+            => Uri.Parse($"smartTogglAlerts://{notificationId}");
```
juanlaube: Not all notifications will be smart alerts, use something different please.

PR #4713, opened 2019-03-18
Fix broken signup on android < 24

### LINKED ISSUE #4712: Android users with devices on API < 24 can't sign up
## 🐛 Bug Report
I was making tests for the Android calendar and noticed I couldn't sign up.
I debugged and noticed that we are using a feature that's only available on android api >= 24.

## Context
Build/Branch: develop
Phone OS: Android
Phone model: Android Api level < 24
Introduced by #4411 

## Expected behavior
Users with android devices with api levels < 24 should also be able to sign up.

## To reproduce
Try to sign up on an emulator with API Level < 24.

## Suggestions
It seems we are just using the wrong class: `android.icu.util.TimeZone` instead of `java.util.TimeZone`. ICU seems to be an alternative to `java.util`, so it shouldn't matter.

daividssilverio:
## What's this?
This is a very simple fix for #4712

### Relationships
Closes #4712

## Why do we want this?
So users with android APIs < 24 can also sign up from the app.

## How is it done?
It just changes one package usage.

### Why not another way?
That's the simplest way.

### Side effects
The APIs should be the same. Since @khoi wrote it, I've requested his review just to make sure.

## Review hints
Test on the current develop, then try here (with an android device with API < 24).

## :squid: Permissions
🦑 it


daividssilverio:
`Android.Icu.Util.TimeZone` is only available on Android Levels >= 24.
`Java.Util.TimeZone` is is available from Android API Level 1.

PR #4721, opened 2019-03-19
GTE: Measure diffing performance

simonrozsival:
## What's this?
Add performance measurement for iOS diffing.

### Relationships
We don't have an issue for this.

## Why do we want this?
We ran into performance issues and this might give us more ideas about where are our bottlenecks. I think the diffing algorithm is actually pretty fast and I hope this will give us a reason to rule out why updating and deleting many things in batch takes so long...

## How is it done?
I moved the extension into a separate class which then grabs the stopwatch provider and measures perf of every diff.

### Why not another way?
I could just grab the stopwatch provider in the extension method from the static container, but I would like to make that easy to update after Mvxit is finished and we have just our custom DI container which is hopefully not static (@heytherewill could chime in and tell us how we're gonna access the services without exposing them as VM properties... hopefully this will be possible).

### Side effects
A bit of (nice?) refactoring.

## Review hints
Run in the iOS simulator.

## :squid: Permissions
🦑

### DIFF
--- a/Toggl.Daneel/Binding/AnimatedTableViewReloadingBinder.cs
+++ b/Toggl.Daneel/Binding/AnimatedTableViewReloadingBinder.cs
@@ -1,0 +1,110 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reactive;
+using Foundation;
+using Toggl.Daneel.ViewSources;
+using Toggl.Foundation.MvvmCross.Collections;
+using Toggl.Foundation.MvvmCross.Collections.Diffing;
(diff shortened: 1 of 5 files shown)


Odrakir:
I don't like the fact that now reload and animated table views have different APIs. And all this is done to measure performance, meaning, it's not something completely necessary to make the code work.

I would personally try to find another way to do this that would fit better with the current implementation.

Also, if there's any performance issue in the diffing due to calling it multiple times in a short amount of time, it can be solved easily adding a `Debounce` operator to the Observable collection.

simonrozsival:
Superseded by #4737

Issue #4722, opened 2019-03-19
Pull-to-refresh causes an irritating glitch


simonrozsival:
When you pull-to-refresh, we sync the data. As a result of how we calculate and store the `since` date (please refer to the syncing docs in this repo if you want to learn more about this) the TE which was updated LAST is always pulled and processed by the data source.

This has in fact an effect of reloading and re-diffing the whole main log even if there were no actual changes to the tree. The last edited TE is hidden and shown with an animation. It used to flash a tiny bit with the old implementation of reloading and I thought this effect would disappear with the diffing, but apparently it did not, it is now even more pronounced than on `develop`. At least this is what happens on iOS, it may or may not be similar on Android. Please try it somebody and leave a comment (and if it doesn't happen on Android, please remove the `android` label from this issue).

We must get rid of it. I hate it.

The ideal solution would be to avoid fetching this unnecessary entity from the server, but that is problematic and it isn't int the scope of the GTE project. We need a UI-only solution which will block log reloading when nothing actually changes.

simonrozsival:
steps to reproduce:

    create a TE in the app
    sync the app
    set a breakpoint on Diffing.cs:92 (return result;)
    pull-to-refresh
    you’ll see two changesets, the first one contains one change which deletes the newly created TE, the other one re-inserts the item at the same spot
    notice how the UI jumps a bit because the row hides and reappears really fast, the whole UI jumps a bit
    every single pull-to-refresh will do the same although the collection doesn’t change at all

@Odrakir suggested that there might be something wrong with identifiers.

simonrozsival:
A fix was merged into the project branch.

PR #4736, opened 2019-03-20
GTE. Refactor flattening

Odrakir:
## What's this?
Fixes some kind of issue when flattening the time entries which caused it to emit them multiple times in a row. And this then caused problems when updating the table view.

### Relationships
It might close some bugs, but I'm not sure. @simonrozsival should check.

## How is it done?
I'm not 100% on why this was failing, but definitely had something to do with the `SelectMany`. This way is much simpler.

Not part of this PR, but throttling the `TimeEntries` observable still might be a good idea to prevent bugs.

## :squid: Permissions
@simonrozsival


Toggl.Foundation.MvvmCross/Transformations/TimeEntriesGroupsFlattening.cs
```diff
@@ -23,30 +23,22 @@ internal sealed class TimeEntriesGroupsFlattening
...
         private readonly HashSet<GroupId> expandedGroups;
-        private readonly IObservable<IThreadSafePreferences> preferencesObservable;
 
         private DurationFormat durationFormat;
 
         public TimeEntriesGroupsFlattening(
-            ITimeService timeService,
-            IObservable<IThreadSafePreferences> preferencesObservable)
+            ITimeService timeService
+            )
```
simonrozsival: This looks ugly now. It could fit on one line.

Toggl.Foundation.MvvmCross/ViewModels/TimeEntriesViewModel.cs
```diff
@@ -72,19 +72,19 @@ public sealed class TimeEntriesViewModel
...
 
-            TimeEntries =
-                interactorFactory.ObserveAllTimeEntriesVisibleToTheUser().Execute()
-                    .Select(timeEntries => timeEntries.Where(isNotRunning))
-                    .ReemitWhen(deletingOrPressingUndo)
-                    .Select(timeEntries => timeEntries.Where(isNotDeleted))
-                    .Select(group)
-                    .ReemitWhen(collapsingOrExpanding)
-                    .SelectMany(groupsFlatteningStrategy.Flatten)
+            var groupedTimeEntries = interactorFactory.ObserveAllTimeEntriesVisibleToTheUser().Execute()
```
simonrozsival: I'm not sure if the name is not confusing. They are grouped, but by days, not by similarity, which is what we refer to as "grouped time entries". Maybe this could be called just `visibleTimeEntries` to avoid a confusion in the future.

simonrozsival:
This PR will close #4722

PR #4746, opened 2019-03-20
Fix threading problem when fetching calendar items

### LINKED ISSUE #4730: Investigate occurrences of time entries updates not being propagated to the calendar
This is most likely caused by some thread marshalling problem and exceptions being swallowed. 

Not easy to track. One case was documented here: https://github.com/toggl/mobileapp/issues/4599#issuecomment-474357310

juanlaube:
## What's this?
Somehow, sometimes, when executing the interactor to fetch the calendar items, there is a threading exception in Realm. This is a possible solution to that problem.

### Relationships

Closes #4730 

## Why do we want this?
🚫 🐛 

## How is it done?
~Subscribing on the background scheduler seems to do the trick.~
Converting the results from the data source to a list.

### Why not another way?
~Maybe the `SubscribeOn` should be in the interactor and not in its usage? 🤔~
🤷‍♂️ 

### Side effects
None.

## :squid: Permissions
🚫


juanlaube:
@Odrakir since you're the Rx expert, can you take a look at this?

amulware:
@daividssilverio For why the `SubscribeOn` doesn't do anything in that long chain, see my comments in the issue. You're already locked into the thread of `ItemsChanged` at that point.

amulware:
In general, should we maybe review if we should have `SubscribeOn` in all interactors that access realm? Or maybe just all of them in general?

amulware:
I'm not entirely sure though why the `ObserveOn` didn't do anything. Or why it broke in the first place...

simonrozsival:
The `IEnumerable` you emit by the observable is not evaluated immediatelly - you don't access Realm until somebody starts enumerating it. Only at that point you access it from some thread, which might be different from the thread on which the enumerable was created (by the Realm library). If you force enumeration by using `ToList()` on the enumerable, it should solve it. I think this might be our problem all across the app. Lazy evaluation does not play nice with thread consistency.

juanlaube:
@simonrozsival that makes sense so I did a quick test and works fine. I think it's a better solution than switching threads, I'm gonna update the PR with that.

daividssilverio:
Thanks @amulware @simonrozsival . Your comments make sense, and yeah Paul, we can't guarantee the same the background scheduler will provide the same background thread on `ObserveOn` and `SubscribeOn`, it should be just some background thread (unless there's something smart under the hood going on).

I'm gonna try to break it now. But the problem should be solved by the forced enumeration.

PR #4752, opened 2019-03-21
📖 update Siri Continue Timer Intent title text

khoi:
### Relationships
No created issue.

### Side effects
None

## :squid: Permissions
@Odrakir


Odrakir:
🤔 There's a "Continue Tracking" already down there in that same file... I think I maybe forgot to delete the "Continue timer" Can you check which one is getting used?

khoi:
both are used, all I did was re-check the English localization and the file get updated. 

Screen Shot 2019-03-21 at 16 18 23

PR #4754, opened 2019-03-21
Ricardo/fix header update

### LINKED ISSUE #4719: Time entry log headers are incorrect

Odrakir:
## What's this?
This modifies the diffing a little bit so it updates section headers as well. Then it removes the old way of doing it via an Observable.

### Relationships
Closes #4719 

## Why do we want this?
With this, headers should update correctly and we keep all the logic in the diffing.

## How is it done?
The diffing algo wasn't ready to update section headers and also it's not ready to have the header dependent on the items in that section (which is what happens to us with the duration). That's why some modifications need to be done in order to get the correct section header when doing the diffing.

### Why not another way?
I think this is cleaner and more performant than cycling through all the sections every time there's any change to the source.

## :squid: Permissions
@simonrozsival


Toggl.Foundation.MvvmCross/Collections/Diffing/Diffing.cs
```diff
@@ -382,9 +397,16 @@ private List<Changeset> generateDeleteSectionsDeletedItemsAndUpdatedItems()
                 return new List<Changeset>();
             }
 
+            updatedSections = deletedItems
+                .Concat(updatedItems)
+                .Select(item => item.sectionIndex)
+                .Concat(deletedSections)
+                .ToList();
```
simonrozsival: We should add `.Distinct()` to remove duplicates.

Toggl.Foundation.MvvmCross/Collections/Diffing/Diffing.cs
```diff
@@ -543,8 +567,14 @@ private IEnumerable<Changeset> generateInsertAndMovedItems()
                 return new List<Changeset>();
             }
 
+            updatedSections = insertedItems
+                .Select(item => item.sectionIndex)
+                .Concat(movedItems.SelectMany(movedItem => new[] { movedItem.Item1.sectionIndex, movedItem.Item2.sectionIndex }))
```
simonrozsival: We should add `.Distinct()` to remove duplicates.

simonrozsival:
This would also close #4724

PR #4759, opened 2019-03-21
Throttle Time Entries

### LINKED ISSUE #4758: The group size does not always update when swipe-deleting time entries of a group
It seems that the size of the group is not considered when diffing and so the group header row is not updated correctly when some of the TEs are deleted and the number does not correspond to the correct number of TEs in the group.

Odrakir:
## What's this?
It throttles the `TimeEntries` property so updates can't happen too fast, which can make the table view update weirdly

### Relationships
Closes #4758

### Why not another way?
This prevents bugs and glitches in the table view, but in any case we should try to find the source of the problem and see why those updates where happening so fast.

## :squid: Permissions
@simonrozsival


Toggl.Foundation.MvvmCross/ViewModels/MainViewModel.cs
```diff
@@ -81,7 +81,7 @@ public sealed class MainViewModel : MvxViewModel
         public IObservable<bool> ShouldShowRatingView { get; private set; }
 
         public IObservable<IEnumerable<AnimatableSectionModel<DaySummaryViewModel, LogItemViewModel, IMainLogKey>>> TimeEntries
-            => TimeEntriesViewModel.TimeEntries.ObserveOn(schedulerProvider.MainScheduler);
```
simonrozsival: Let's just move the duration to a constant.

PR #4760, opened 2019-03-21
💥 Fix crash in settings when no avatar picture is available

### LINKED ISSUE #4553: App crashes when navigating to the settings page
It happened all of the time for @maritoggl in the last adhoc test build.

Might be worth it to open the settings with the last build using different devices and see what gives.

When it happens: Sometimes, when you don't have a profile pic set up.

kornelijepetak:
## What's this?
Settings screen no longer crashes if the user has no avatar image set.

### Relationships
Closes #4553

## Why do we want this?
So that there's no explosions.

## How is it done?
Do not be `null`, please!

## :squid: Permissions
Oh, by all means.


Odrakir:
Shouldn't that be done in the view model?

PR #4764, opened 2019-03-22
GTE: Fix delayed deletion flickering

### LINKED ISSUE #4758: The group size does not always update when swipe-deleting time entries of a group

simonrozsival:
## What's this?
This PR solves the cause of the problem described in #4758.

### Relationships

Closes #4758

## Why do we want this?

We want nice UI and UX.

## How is it done?
The main idea of this PR is to **remember IDs of all the deleted TEs** until the available time for undoing runs out or user presses the undo button.

The effect of this is that we don't emit time entries which have the already been deleted, but are still cached in the observable chain (combine latest/reemit when) and haven't managed to be propagated from the data source yet.

I also extracted the delayed observable from the VM and put it into an `DelayedInteractorExecution` class. This is the main reason this PR is a draft. I think it makes it much easier to reason about what's going on in the VM and it could be tested much more thoroughly, but maybe it is overengineered. I want to share this PR with you before I spend a lot of time updating and adding tests.

### Side effects
I'm not aware of any.

## Review hints
Try this in simulator/emulator.

## :squid: Permissions
🚫

### DIFF
--- a/Toggl.Foundation.MvvmCross/ViewModels/TimeEntriesViewModel.cs
+++ b/Toggl.Foundation.MvvmCross/ViewModels/TimeEntriesViewModel.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Reactive;
 using System.Reactive.Linq;
 using System.Reactive.Subjects;
 using System.Threading.Tasks;
@@ -16,26 +17,24 @@ using Toggl.Foundation.MvvmCross.Transformations;
 using Toggl.Foundation.MvvmCross.ViewModels.TimeEntriesLog;
 using Toggl.Foundation.MvvmCross.ViewModels.TimeEntriesLog.Identity;
 using Toggl.Foundation.Services;
-using Toggl.Foundation.Sync;
 using Toggl.Multivac;
 using Toggl.Multivac.Extensions;
 
 namespace Toggl.Foundation.MvvmCross.ViewModels
 {
-
     [Preserve(AllMembers = true)]
     public sealed class TimeEntriesViewModel
     {
-        private readonly ISyncManager syncManager;
         private readonly IInteractorFactory interactorFactory;
         private readonly IAnalyticsService analyticsService;
         private readonly ISchedulerProvider schedulerProvider;
 
         private readonly TimeEntriesGroupsFlattening groupsFlatteningStrategy;
 
-        private Subject<int?> timeEntriesPendingDeletionSubject = new Subject<int?>();
-        private IDisposable delayedDeletionDisposable;
-        private long[] timeEntriesToDelete;
+        private readonly Subject<int?> timeEntriesPendingDeletionSubject = new Subject<int?>();
+        private readonly HashSet<long> timeEntriesToDelete = new HashSet<long>();
+
+        private DelayedInteractorExecution<IObservable<Unit>> delayedDeletion;
 
         public IObservable<IEnumerable<AnimatableSectionModel<DaySummaryViewModel, LogItemViewModel, IMainLogKey>>> TimeEntries { get; }
         public IObservable<bool> Empty { get; }
@@ -48,7 +47,6 @@ namespace Toggl.Foundation.MvvmCross.ViewModels
 
         public TimeEntriesViewModel(
             ITogglDataSource dataSource,
-            ISyncManager syncManager,
             IInteractorFactory interactorFactory,
             IAnalyticsService analyticsService,
             ISchedulerProvider schedulerProvider,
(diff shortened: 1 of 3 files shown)


Toggl.Foundation.MvvmCross/ViewModels/TimeEntriesViewModel.cs
```diff
@@ -100,15 +96,13 @@ public sealed class TimeEntriesViewModel
 
         public async Task FinalizeDelayDeleteTimeEntryIfNeeded()
         {
-            if (timeEntriesToDelete == null)
+            if (delayedDeletion != null)
```
kornelijepetak: Why did you change the early return? Please don't. We are using this as a convention.

Toggl.Foundation/Interactors/DelayedExecutionInteractor.cs
```diff
@@ -0,0 +1,60 @@
...
+            {
+                throw new InvalidOperationException("Interactor can't be executed repeatedly.");
+            }
+
+            subscription = Observable.Return(Unit.Default)
+                .Delay(delay, scheduler)
+                .Select(_ => interactor.Execute())
+                .Subscribe(result);
+
+            return result.AsObservable();
```
kornelijepetak: There's no reason to turn this to AsObservable. Subject<T> implements that interface as is.

 ```suggestion
             return result;
```

File: Toggl.Foundation/Interactors/DelayedExecutionInteractor.cs [hunk 3]
```diff
@@ -0,0 +1,60 @@
...
+
+        public IObservable<T> Execute()
+        {
+            var result = new Subject<T>();
+            if (subscription != null)
+            {
+                throw new InvalidOperationException("Interactor can't be executed repeatedly.");
+            }
+
+            subscription = Observable.Return(Unit.Default)
```
kornelijepetak: Can be simplified to:

```cs
subscription = Observable.Delay(interactor.Execute(), delay, scheduler).Subscribe(result);
```

Toggl.Foundation/Interactors/DelayedExecutionInteractor.cs
```diff
@@ -0,0 +1,60 @@
...
+                .Delay(delay, scheduler)
+                .Select(_ => interactor.Execute())
+                .Subscribe(result);
+
+            return result.AsObservable();
+        }
+
+        public T ExecuteImmediately()
+        {
+            Cancel();
```
kornelijepetak: I'm confused a bit.
`ExecuteImmediatelly` will call `Cancel` in attempt to prevent delayed execution. However, if `Cancel` is called, it will throw if `subscription` is `null`. And `subscription` will be `null` until `Execute` is called.

Transitively that means that calling `ExecuteImmediately` can only be called after `Execute` in order not to crash. This does not sound right to me from the API perspective.
If a class has two methods `Execute` and `ExecuteImmediately`, it sounds like I should choose whether I want a delayed version or immediate version. But there's a hidden dependency there.

I am not sure why `Cancel` would ever throw. I would rather that function just dispose of the sub, and set the variable to `null` and make it idempotent so that one can call Cancel as many time they want. 

Am I missing something?

Toggl.Foundation/Interactors/DelayedExecutionInteractor.cs
simonrozsival: No, this would call the `Execute` method **before** the delay passes. This would only be safe in cases when the interactor creates an observable which is cold and will only be executed once subscribed to. Since this is a general interactor wrapper, it's not an option. I want to delay execution, not delay returning the result.

Toggl.Foundation/Interactors/DelayedExecutionInteractor.cs
simonrozsival: You have a point. I think `Cancel` doesn't have to throw.

Odrakir:
I find the code a little hard to understand, so the first thing I did is test it, and I see `TimeEntries` still emit as much as it used to, so I'm not sure what's the point of this.

simonrozsival:
It emits the data multiple times, but the data is correct (the same data is emitted multiple times in a row, I'm not able to fix that in this PR).

Let's say you have a log with time entries `[A + B + C]` (it doesn't matter if they  are grouped or not).
- when you delete B, the log observable emits `[A + C]`
- when you then delete C before the undo period is over, without this fix, the log observable emits `[A + B]` and then imediatelly just `[A]` (which is the correct value)
- with this fix, it would not emit the extra and incorrect `[A + B]`

I think it's easier to look at #4766 and see the logic in there.

simonrozsival:
We should still combine this with the throttling to get better performance.

Toggl.Foundation/Interactors/DelayedExecutionInteractor.cs
kornelijepetak: No, you are completely right.

simonrozsival:
We decided to go with #4764 which doesn't add additional complexity.

PR #4773, opened 2019-03-22
Fix several Wheel issues

### LINKED ISSUE #4524: Duration input is being reset when editing running time entry
## 🐛 Bug Report
While editing a running time entry, its duration is being reset while editing.

## Context
Build/Branch: 1.11.1
Phone OS: Android
Phone model: All, _systemic anomaly_.

## Expected behavior
* Editing duration for a running time entry should change StartTime accordingly
* While the duration is being edited by using the input field, no changes to that input should be happening.

## To reproduce
1. Open app
2. Make sure manual mode is off
3. Tap on play button to create new running TE
4. Tap on duration field on new TE screen
5. Tap on duration, try to enter something
6. Duration field is resetting

## Suggestions
I've noticed while reproducing that it resets it in a timely manner, and my intuition says that it's CurrentTime observable emitting events that end up resetting the input. Check this theory first, might be easy to fix.

### LINKED ISSUE #4541: Setting the Start or Stop Time for a TE uses the seconds of the current time
Setting the Start and Stop Times in Manual mode uses the current time's seconds, potentially resulting in Time Entries that are shorter than their apparent length. For example, the wheel may show 9:00 - 9:05, but the new TE's duration ends up being 00:04:47.

## Context

Build/Branch: Giskard 1.11.1
Phone OS: Android 9
Phone model: ONEPLUS A6003_22_190123

## Expected behavior
The new TE created should have a duration that matches the set Start and Stop Time.

## To reproduce
1. Open a computer clock that shows you seconds. Open the app and get it ready.
2. Create a new Time Entry under manual mode when the it’s :00 seconds on the computer clock. Take your time with the next steps.
3. Tap on the Duration of the new TE view (this opens the wheel)
4. Set the Start Time to 9:00AM
5. Set the Stop Time to 9:05AM
6. Tap Save (this closes the wheel)
7. See the Duration now show 4:XX, where XX is the number of seconds that has elapsed between steps 2 and 5 above

## Suggestions
It looks like this is unintended behavior. Per @amulware "changing start/stop time manually should always set the seconds component to 0".

[case](https://app.intercom.io/a/apps/ayixs927/inbox/inbox/2100764/conversations/21089583473)

### LINKED ISSUE #4770: In wheel view, Stop Timer should only stop the timer, not show time selecting UI
On the wheel view, when editing times/duration for the running time entry, there's a button saying "Stop Timer".

Currently:

|Android|iOS|
|-|-|
|1. Stops the Timer|1. Stops the Timer|
|2. Opens modal time picker for Stop Time|2. Displays barrel for Stop Time|

In both cases, 2. is intrusive and should be removed.

`Stop Timer` should just stop the Time Entry. Period.

Decided by @maritoggl and @amulware after a short [discussion](https://toggl.slack.com/archives/C02APMYTM/p1553258728259100).

kornelijepetak:
## What's this?
Fixes several related issues with the Wheel view.

* Seconds are shown correctly in duration field for running time entry
* Editing duration for the time entry is no longer reset every second
* Stop Timer button will now simply stop the time entry, not open dialogs
* Touching start/stop time manually no longer preserves the seconds
* Remove redundant code
* Smaller refactorings along the way

### Relationships
Closes #4524 
Closes #4770 
Closes #4541 

## Why do we want this?
To fix bugs and have more stable code

### Side effects
Should be none. But please do test it on both platforms.

## Review hints
If you review this commit by commit and take the commit msg into account, it should be a breeze.

## :squid: Permissions
Me. After at least one iOS and one Android dev reviews this.


Toggl.Giskard/ViewHelpers/TimePickerListener.cs
```diff
@@ -17,7 +17,7 @@ public TimePickerListener(DateTimeOffset currentTime, Action<DateTimeOffset> onT
 
         public void OnTimeSet(TimePicker view, int hourOfDay, int minute)
         {
-            var pickedTime = new DateTimeOffset(currentTime.Year, currentTime.Month, currentTime.Day, hourOfDay, minute, currentTime.Minute, currentTime.Millisecond, currentTime.Offset);
+            var pickedTime = new DateTimeOffset(currentTime.Year, currentTime.Month, currentTime.Day, hourOfDay, minute, 0, currentTime.Millisecond, currentTime.Offset);
```
simonrozsival: This is very confusing. If we set seconds to `0`, shouldn't we also set the milliseconds to `0`? Or maybe we should keep `currentTime.Seconds` as well (the minutes were obviously incorrect)?

Toggl.Giskard/Activities/EditDurationActivity.cs
```diff
@@ -243,14 +225,23 @@ public override bool OnOptionsItemSelected(IMenuItem item)
...
-           toolbar.Title = Foundation.Resources.StartAndStopTime;
+            toolbar.Title = Foundation.Resources.StartAndStopTime;
 
             SetSupportActionBar(toolbar);
 
             SupportActionBar.SetDisplayHomeAsUpEnabled(true);
             SupportActionBar.SetDisplayShowHomeEnabled(true);
         }
 
+        private void updateStopTimeUiVisibility(bool isRunning)
```
simonrozsival: `UI` is normally used with all capital letters even in C#, is this an Android convention?

kornelijepetak: Of course, this was my mistake.

Toggl.Giskard/ViewHelpers/TimePickerListener.cs
kornelijepetak: Yes totally.

PR #4778, opened 2019-03-25
Only show project creation error when users hit Save

### LINKED ISSUE #4401: Show project creation error when users hit "Create"
Since the users are now allowed to create projects with the same name if they're associated with different clients (#4380).

The error message showing when users enter the project creation view is confusing (see video). 

We'll change the behavior so that message only show when users hit "Create"

name

khoi:
## What's this?
as the title says

### Relationships
Closes #4401

## How is it done?
By updating the Save UIAction 

### Why not another way?
Since now button is the one that drives the error, it's easier to link the error to the action instead of having separated IObservable

### Side effects
Should be none.

## Review hints
Commit by commit could help.

- Edit a te
- Enter an existing client name
- Click "Create client Name" cell
- Error should not be shown.
- Hit save, the error should be shown.

## :squid: Permissions
🦑


khoi:
https://khoi-screenshots.sgp1.digitaloceanspaces.com/Screen-Recording-2019-03-26-at-5.30.22-PM.mov

@maritoggl Should we clear the error message when the name, client pair becomes valid? See the attached clip.

maritoggl:
Yes

khoi:
PR has been updated so that the error is cleared out when users update the inputs: project name, client

amulware:
Blocking this until "Sorry, such project already exists" has been checked by a native. That sounds like extremely awkward grammar to me.

khoi:
We'll go for "Sorry, this project already exists" for now

PR #4786, opened 2019-03-25
Fix crash when updating date range in reports calendar

### LINKED ISSUE #4779: Crashing when I click on 23 in reports
## 🐛 Bug Report
Clear app cache and using a slow connection

## Context
There is no data on that day but when I attempt to retrieve it and give the app several seconds it then crashes. I've replicated this 5 times now.

Build/Branch:1.12
Phone OS: Android 8.0
Phone model: HTC U Ultra

## Expected behavior
To have the blank reports page displayed

## To reproduce
Clear the app cache
Use slow connection or mobile data
Log into an existing account
Tap reports
Tap the date range selection
Click on the day 2 days prior
Click back to the reports screen
After several seconds the app crashes
image

juanlaube:
## What's this?
Fixes a threading problem when updating the selected date range.
I tried to reproduce this in every possible way, even on slower networks as the issue mentions, but still I couldn't. However, based on the stack trace posted in the issue, this seems to be the fix.

### Relationships

Closes #4779 

## Why do we want this?
🚫 💥 

## How is it done?
Based on the stack trace, the update should happen in the main thread.

### Why not another way?
🤷‍♂️ 

### Side effects
None.

## Review hints
Please try it.

## :squid: Permissions
⚠️ Only the release manager, a.k.a. @amulware ⚠️


amulware:
Feel free to squash, if someone else approves this.

PR #4790, opened 2019-03-26
Fix invalid order of labels

### LINKED ISSUE #4789: Inverted positions of duration and date labels
## 🐛 Bug Report
The positions of the date and summary duration are inverted.

image

## Context
Build/Branch: released application
Phone OS: Android 9
Phone model: OPO5	

## Expected behavior
The date label should be on the left, and the summary duration on the right.

## To reproduce
Open 1.12 and just look at the screen.

## Suggestions
This could have been introduced by some bad merging probably.

kornelijepetak:
## What's this?
This is correct natural order.
The date label should be on the left, the duration on the right.

### Relationships
Closes #4789 

## Why do we want this?
Because it just makes more sense.

## How is it done?
By undoing an error somebody has done 😄 

### Why not another way?
This is the only way that makes sense.

### Side effects
It will now work correctly.

## Review hints
By careful examination of the code.

## :squid: Permissions
@daividssilverio

### DIFF
--- a/Toggl.Giskard/ViewHolders/MainLogSectionViewHolder.cs
+++ b/Toggl.Giskard/ViewHolders/MainLogSectionViewHolder.cs
@@ -35,8 +35,8 @@ namespace Toggl.Giskard.ViewHolders
 
         protected override void UpdateView()
         {
-            mainLogHeaderTitle.Text = Item.DurationText;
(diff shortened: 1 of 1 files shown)


amulware:
This should be on the 1.12 release branch.

kornelijepetak:
Based incorrectly.

PR #4792, opened 2019-03-26
Dependency Injection Refactor

### LINKED ISSUE #3958: Create our own dependency containers
Depends on #3957
For a base implementation, check [this branch](https://github.com/toggl/mobileapp/tree/ioc-refactor)

We need to:

- [x] Turn `Foundation` and `FoundationMvvmCross` into dependency containers
  - [x] These should expose `Lazy<T>` services, as discussed in the proposal of this refactor
- [x] Create our own `IMvxViewModelLocator` instead of relying on MvvmCross' implementation of it

heytherewill:
## What's this?
This removes our dependency on the Mvx IoC container

### Relationships
Closes #3958

## Why do we want this?
Mvexit

## How is it done?
I am creating my own dependency container
It uses `Lazy<T>` internally so we only initialize objects at the last possible point
We then create our own ViewModel locator that does not use reflection, unlike the existing Mvx one

### Why not another way?
There still are a couple of references to `Mvx.Resolve`, but these are deeply tied to the presentation layer and will be removed when the time to Mvexit comes

### Side effects
This needs a heckton of testing because it touches all sorts of lifecycles, so please stress test it.

## Review hints
Reviewing by commit will help, please review it like that!

## :squid: Permissions
Only myself


Toggl.Daneel/Startup/IosDependencyContainer.cs
```diff
@@ -0,0 +1,135 @@
...
+        private const ApiEnvironment environment =
+#if USE_PRODUCTION_API
+            ApiEnvironment.Production;
+#else
+            ApiEnvironment.Staging;
+#endif
+        
+        private readonly Lazy<SettingsStorage> settingsStorage;
+
+        public IMvxNavigationService ForkingNavigationService { get; internal set; }
```
daividssilverio: `ForkingNavigationService` this name is a leftover from when android didn't have the bottom navigation bar.
just `NavigationService` should be enough.

Toggl.Foundation/UserAccessManager/UserAccessManager.cs
```diff
@@ -42,6 +42,7 @@ public sealed class UserAccessManager : IUserAccessManager
             this.database = database;
             this.apiFactory = apiFactory;
             this.googleService = googleService;
+            this.privateSharedStorageService = privateSharedStorageService;
```
daividssilverio: I better stop reviewing then.

daividssilverio:
Sad thing we can't remove `Mvx.Resolve<IMvxAndroidCurrentTopActivity>().Activity` just yet.

daividssilverio:
Also, I wonder if `Lazy` is the right way to go 🤔 
I mean, it works, but it will keep stuff on memory, right? We have a singleton for Android and iOS right?

It would be nice to able to say what's is a singleton and what should be created on every invocation.

Toggl.Daneel/Startup/Setup.cs
```diff
@@ -10,141 +8,44 @@
...
 
-#if USE_PRODUCTION_API
-        private const ApiEnvironment environment = ApiEnvironment.Production;
-#else
-        private const ApiEnvironment environment = ApiEnvironment.Staging;
-#endif
+        protected override IMvxApplication CreateApp()
+        {
+            var version = NSBundle.MainBundle.InfoDictionary["CFBundleShortVersionString"].ToString();
+            IosDependencyContainer.Instance = new IosDependencyContainer(Presenter as TogglPresenter, version);
```
Odrakir: Isn't this a Singleton? Is this the correct way of creating it in C#? Shouldn't Instance have a private setter?

Toggl.Foundation.MvvmCross/ViewModels/LoginViewModel.cs
```diff
@@ -249,12 +258,10 @@ private async Task startPasswordManager()
             Login();
         }
 
-        private async void onInteractorFactory(ISyncManager syncManager)
+        private async void onInteractorFactory(Unit syncManager)
```
Odrakir: `Unit syncManager`?

Odrakir: Also... why is this method called `onInteractorFactory`?

Toggl.Foundation.MvvmCross/ViewModels/SignupViewModel.cs
```diff
@@ -295,14 +295,12 @@ private async Task signup()
                 .Subscribe(onInteractorFactory, onError, onCompleted);
         }
 
-        private async void onInteractorFactory(ISyncManager syncManager)
+        private async void onInteractorFactory(Unit syncManager)
```
Odrakir: Same questions as in `LoginViewModel`

Toggl.Foundation/DependencyContainer.cs
```diff
@@ -0,0 +1,245 @@
...
+using Toggl.Foundation.Shortcuts;
+using Toggl.Foundation.Suggestions;
+using Toggl.Foundation.Sync;
+using Toggl.Multivac;
+using Toggl.Multivac.Extensions;
+using Toggl.PrimeRadiant;
+using Toggl.Ultrawave;
+using Toggl.Ultrawave.Network;
+using Toggl.PrimeRadiant.Settings;
+using Toggl.Foundation.Autocomplete;
```
Odrakir: 🔥

Toggl.Foundation/DependencyContainer.cs
```diff
@@ -0,0 +1,245 @@
...
+        public IInteractorFactory InteractorFactory => interactorFactory.Value;
+
+        public IApiFactory ApiFactory => apiFactory.Value;
+        public ITogglDatabase Database => database.Value;
+        public ITimeService TimeService => timeService.Value;
+        public IPlatformInfo PlatformInfo => platformInfo.Value;
+        public ITogglDataSource DataSource => dataSource.Value;
+        public IGoogleService GoogleService => googleService.Value;
+        public IRatingService RatingService => ratingService.Value;
+        public ICalendarService CalendarService => calendarService.Value;
```
Odrakir: `GoogleService`, `CalendarService`, `ShortcutCreator` and `NotificationService` properties are not used. I guess that's because those services are only accessed through interactors. Can we maybe remove the properties?

Odrakir:
Also, it crashes for me after login, stack trace [here](https://gist.github.com/Odrakir/8739feb3eb917c2666a395b934c5ff2d).

Toggl.Daneel/Startup/Setup.cs
heytherewill: The instance depends on external things (like in this case, the Presenter) so in order to create it in the appropriate lifecycle moment we need a public getter and setter

Toggl.Foundation.MvvmCross/ViewModels/LoginViewModel.cs
heytherewill: Leftovers

Toggl.Daneel/Startup/Setup.cs
simonrozsival: Can the setter check that the instance hasn't been assigned yet and if it has, throw an exception? If it's locked, it might even be threadsafe.

Toggl.Foundation.MvvmCross/ViewModelLocator.cs
```diff
@@ -0,0 +1,415 @@
...
+            var viewModel = findViewModel(viewModelType) as IMvxViewModel<TParameter>;
+
+            RunViewModelLifecycle(viewModel, param, parameterValues, savedState);
+
+            return viewModel;
+        }
+
+        private IMvxViewModel findViewModel(Type viewModelType)
+        {
+            if (viewModelType == typeof(BrowserViewModel))
```
simonrozsival: I know this won't be a bottleneck, but couldn't we turn this into a `Dictionary<Type, Func<IMvxViewModel>>` or something like.. could we use switch expression once we use C#8?

Toggl.Foundation.MvvmCross/ViewModelLocator.tt
```diff
@@ -0,0 +1,112 @@
+﻿<#@ template debug="false" hostspecific="True" language="C#" #>
```
simonrozsival: Can we remove this template file once we have a stable first version of the locator and from then keep it up-to-date manually?

Toggl.Foundation.MvvmCross/ViewModels/LoginViewModel.cs
```diff
@@ -81,16 +82,24 @@ public enum ShakeTargets
...
-            Ensure.Argument.IsNotNull(analyticsService, nameof(analyticsService));
-            Ensure.Argument.IsNotNull(onboardingStorage, nameof(onboardingStorage));
-            Ensure.Argument.IsNotNull(navigationService, nameof(navigationService));
-            Ensure.Argument.IsNotNull(passwordManagerService, nameof(passwordManagerService));
-            Ensure.Argument.IsNotNull(errorHandlingService, nameof(errorHandlingService));
-            Ensure.Argument.IsNotNull(lastTimeUsageStorage, nameof(lastTimeUsageStorage));
-            Ensure.Argument.IsNotNull(timeService, nameof(timeService));
-            Ensure.Argument.IsNotNull(schedulerProvider, nameof(schedulerProvider));
-            Ensure.Argument.IsNotNull(rxActionFactory, nameof(rxActionFactory));
+            try
```
simonrozsival: I assume this is just WIP?

Toggl.Foundation.MvvmCross/ViewModelLocator.cs
```diff
@@ -0,0 +1,415 @@
...
+        public override IMvxViewModel<TParameter> Load<TParameter>(Type viewModelType, TParameter param, IMvxBundle parameterValues, IMvxBundle savedState)
+        {
+            var viewModel = findViewModel(viewModelType) as IMvxViewModel<TParameter>;
+
+            RunViewModelLifecycle(viewModel, param, parameterValues, savedState);
+
+            return viewModel;
+        }
+
+        private IMvxViewModel findViewModel(Type viewModelType)
```
simonrozsival: This needs a "constructor test" for every view model to test that the locator works well (with the assumption of correctly initialized DI container).

Toggl.Daneel/Startup/Setup.cs
kornelijepetak: We don't need a public setter here.

For some safety, I'd rather use some kind of an `IosDependencyContainer.Initialize(Presenter as TogglPresenter, version)`. And then inside 

* initialize calls the constructor (which should be private) to initialize the singleton instance
* Does the check whether the instance is already initialized to prevent double inits (what @simonrozsival said)

Toggl.Foundation.MvvmCross/UIDependencyContainer.cs
```diff
@@ -0,0 +1,56 @@
...
+using Toggl.PrimeRadiant.Settings;
+using Toggl.Foundation.MvvmCross.Services;
+using Toggl.Foundation.Services;
+using Toggl.Ultrawave;
+using Toggl.Ultrawave.Network;
+using MvvmCross.Navigation;
+
+namespace Toggl.Foundation.MvvmCross
+{
+    public abstract class UiDependencyContainer : DependencyContainer
```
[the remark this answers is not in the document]
heytherewill: If you have a better name suggestion, I am all ears

Toggl.Foundation.MvvmCross/ViewModelLocator.cs
heytherewill: When we migrate to C# 8 we can tots do it

Toggl.Foundation.MvvmCross/ViewModelLocator.tt
heytherewill: I think this is stable enough and we can 🔥 it before even merging

Toggl.Foundation.MvvmCross/ViewModelLocator.tt
```diff
@@ -0,0 +1,120 @@
...
+        }
+
+        private IMvxViewModel findViewModel(Type viewModelType)
+        {
+<#
+    const string ifLine = "            if (viewModelType == typeof({0}))";
+    const string returnLine = "                return new {0}({1});";
+    
+    Func<string, bool> fileIsValid = fileName => 
+        !(fileName.Contains(".DS_Store")
```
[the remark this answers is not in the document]
simonrozsival: 🔥 , this will be 🔥ed

Toggl.Foundation/UserAccessManager/UserAccessManager.cs
```diff
@@ -58,142 +54,130 @@ public IObservable<ISyncManager> Login(Email email, Password password)
 
             var credentials = Credentials.WithPassword(email, password);
 
-            return database
+            return database.Value
```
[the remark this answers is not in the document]
heytherewill: That sounds like a refactor for latter on. I do not feel like introducing a `Lazy<T>` type that is as efficient/safe as the BCL one

heytherewill:
@daividssilverio Lazy is the right way to go because there is no reason to create our dependencies all at once (which is what would happen without lazy loading because the interactor factory is an early dependency)
> I mean, it works, but it will keep stuff on memory, right? 
In the current system (the MvvmCross IoC) everything is already kept in memory, so I do not get your concern.

@Odrakir At the moment, we have a completely different system to register dependencies. There is no comparison that can be done. Implementing it without lazy would basically means reimplementing everything twice.

heytherewill:
@simonrozsival I added the test. It is currently being skipped, but this will be fixed when we remove MvvmCross :)

daividssilverio:
@heytherewill , `Lazy` is alright.
>Lazy is the right way to go because there is no reason to create our dependencies all at once (which is what would happen without lazy loading because the interactor factory is an early dependency)

What I wanted to say is that we might not want to always hold stuff in memory after they are done being used. Of course we shouldn't create everything at once.

It might not make much of a difference because most of our view models have similar dependencies.

Toggl.Foundation.MvvmCross/UIDependencyContainer.cs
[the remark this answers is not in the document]
arturs105: I'd also prefer `UIDependencyContainer ` over `UiDependencyContainer `, since we already have `UIAction`, `UILabelExtensions` etc.

Toggl.Daneel/Startup/AppDelegate.cs
```diff
@@ -279,14 +280,21 @@ private void observeAndStoreLastUpdateDate()
...
+
                 interactorFactory.ObserveTimeEntriesChanges().Execute().StartWith(default(Unit))
                     .SelectMany(interactorFactory.GetAllTimeEntriesVisibleToTheUser().Execute())
                     .Select(timeEntries => timeEntries.OrderBy(te => te.At).Last().At)
                     .Subscribe(privateSharedStorage.SaveLastUpdateDate)
                     .DisposedBy(lastUpdateDateDisposable);
             }
+            catch (Exception)
+            {
+                // Ignore errors when logged out
```
amulware: Is that the only errors that are ignored here? A catch-all like that seems pretty dangerous.

Toggl.Daneel/Startup/IosDependencyContainer.cs
```diff
@@ -0,0 +1,143 @@
...
+using Toggl.Ultrawave.Network;
+
+namespace Toggl.Daneel
+{
+    public sealed class IosDependencyContainer : UiDependencyContainer
+    {
+        private const int numberOfSuggestions = 3;
+        private const string clientName = "Daneel";
+        private const string remoteConfigDefaultsFileName = "RemoteConfigDefaults";
+        private const ApiEnvironment environment =
```
amulware: Should these four things really be hardcoded here?

Toggl.Daneel/Startup/IosDependencyContainer.cs
```diff
@@ -0,0 +1,143 @@
...
+using Toggl.Multivac;
+using Toggl.PrimeRadiant;
+using Toggl.PrimeRadiant.Realm;
+using Toggl.PrimeRadiant.Settings;
+using Toggl.Ultrawave;
+using Toggl.Ultrawave.Network;
+
+namespace Toggl.Daneel
+{
+    public sealed class IosDependencyContainer : UiDependencyContainer
```
amulware: I like having this class much much more than configuring all the dependencies in the startup of the app directly.

Toggl.Daneel/Startup/IosDependencyContainer.cs
```diff
@@ -0,0 +1,143 @@
...
+        public IMvxNavigationService MvxNavigationService { get; internal set; }
+
+        public TogglPresenter ViewPresenter { get; }
+
+        public static IosDependencyContainer Instance { get; private set; }
+
+        public static void Initialize(TogglPresenter viewPresenter, string version)
+        {
+            if (Instance != null)
+                return;
```
amulware: Maybe throw instead?

Toggl.Daneel/Startup/IosDependencyContainer.cs
```diff
@@ -0,0 +1,143 @@
...
+
+        protected override IApplicationShortcutCreator CreateShortcutCreator()
+            => new ApplicationShortcutCreator();
+
+        protected override IStopwatchProvider CreateStopwatchProvider()
+            => new FirebaseStopwatchProviderIos();
+
+        protected override ISuggestionProviderContainer CreateSuggestionProviderContainer()
+            => new SuggestionProviderContainer(
+                new MostUsedTimeEntrySuggestionProvider(Database, TimeService, numberOfSuggestions)
```
amulware: This seems like something that should be in shared code, shouldn't it? 🤔

amulware: There might be other things as well, where iOS and Android use the same thing, but this one stood out to me as definitely not being iOS specific.

Toggl.Daneel/ViewControllers/Calendar/CalendarViewController.cs
```diff
@@ -36,7 +36,7 @@ public sealed partial class CalendarViewController : ReactiveViewController<Cale
         public CalendarViewController()
             : base(nameof(CalendarViewController))
         {
-            timeService = Mvx.Resolve<ITimeService>();
+            timeService = IosDependencyContainer.Instance.TimeService;
```
amulware: It bothers me a little that these are always prefixed with 'ios' - why not create an Initialise on the base class which can be called to initialise a property on that?

Toggl.Foundation.MvvmCross/ViewModelLocator.cs
```diff
@@ -0,0 +1,417 @@
...
+        public override IMvxViewModel Load(Type viewModelType, IMvxBundle parameterValues, IMvxBundle savedState)
+        {
+            var viewModel = findViewModel(viewModelType);
+
+            RunViewModelLifecycle(viewModel, parameterValues, savedState);
+
+            return viewModel;
+        }
+
+        public override IMvxViewModel<TParameter> Load<TParameter>(Type viewModelType, TParameter param, IMvxBundle parameterValues, IMvxBundle savedState)
```
amulware: Can't we make these Load methods generic in the view model type? That would allow us to make `findViewModel` generic as well, which would make it compile to constant time (it should remove all the obviously false type checks then). Generally, generic seems nicer than passing types around as well.

Toggl.Foundation.MvvmCross/ViewModelLocator.cs
```diff
@@ -0,0 +1,417 @@
...
+        {
+            var viewModel = findViewModel(viewModelType) as IMvxViewModel<TParameter>;
```
amulware: I suspect this can't be null, so it should use an explicit cast.

Toggl.Foundation/DependencyContainer.cs
```diff
@@ -0,0 +1,240 @@
...
+        private readonly Lazy<ITimeService> timeService;
+        private readonly Lazy<IPlatformInfo> platformInfo;
+        private readonly Lazy<ITogglDataSource> dataSource;
+        private readonly Lazy<IGoogleService> googleService;
+        private readonly Lazy<IRatingService> ratingService;
+        private readonly Lazy<ICalendarService> calendarService;
+        private readonly Lazy<ILicenseProvider> licenseProvider;
+        private readonly Lazy<IUserPreferences> userPreferences;
+        private readonly Lazy<IRxActionFactory> rxActionFactory;
+        private readonly Lazy<IAnalyticsService> analyticsService;
```
amulware: Isn't it important to initialise our analytics services as soon as possible make sure we track all crashes? We've had issues with AppCenter not recording android crashes in the past because it was initialised too late. Similarly, we want accurate performance measurements with Firebase.

Toggl.Foundation/DependencyContainer.cs
```diff
@@ -0,0 +1,240 @@
...
+                TimeService,
+                AnalyticsService,
+                LastTimeUsageStorage,
+                SchedulerProvider.DefaultScheduler,
+                StopwatchProvider,
+                AutomaticSyncingService
+            );
+            SyncErrorHandlingService.HandleErrorsOf(syncManager);
+
+            syncManager.ForceFullSync().GetAwaiter().GetResult();
```
amulware: This heavy piece of code inside a dependency container? That really does not seem right.

Toggl.Multivac/Extensions/LazyExtensions.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System;
+
+namespace Toggl.Multivac.Extensions
+{
+    public static class LazyExtensions
+    {
+        public static Lazy<TResult> Select<T, TResult>(this Lazy<T> lazy, Func<T, TResult> transform)
+            => new Lazy<TResult>(() => transform(lazy.Value));
```
amulware: I see what you did there.

Toggl.Daneel/Startup/IosDependencyContainer.cs
heytherewill: This behaviour is similar to the existing one (`SetupSingleton.EnsureInitialized()`), so I would rather keep it

Toggl.Daneel/ViewControllers/Calendar/CalendarViewController.cs
heytherewill: Mainly because we have some presentation-related platform-specific services that are needed in a bunch of places on the platform code. This makes exposing the more specific containers tricky

Toggl.Foundation.MvvmCross/ViewModelLocator.cs
heytherewill: We can during Mvexit. Here, we are constrained to what MvvmCross gives us, which is very type unsafe in a way

Toggl.Foundation/DependencyContainer.cs
heytherewill: The `IAnalyticsService` interface is merely used for tracking events. The crashing system is independent and platform specific. That one is initialized per platform on startup

Toggl.Foundation/DependencyContainer.cs
heytherewill: It does not, I agree. It however enforces the correctness of the program. I do not see any other extension point before we Mvexit. I am open to suggestions, however

Toggl.Multivac/Extensions/LazyExtensions.cs
heytherewill: I appreciate

Toggl.Daneel/Startup/IosDependencyContainer.cs
heytherewill: Agreed

Toggl.Daneel/Startup/IosDependencyContainer.cs
heytherewill: They are all related to the creation of dependencies and are used nowhere else, so I think they should, yeah

Toggl.Daneel/Startup/AppDelegate.cs
heytherewill: Since we are not awaiting the `IObservable`, the errors that it would emit are being ignored already, so container exceptions are the only one in play here

Toggl.Daneel/Startup/IosDependencyContainer.cs
amulware: Maybe change the method name to `EnsureInitialised()` then? That implies that it can be called multiple times without failure.

Toggl.Daneel/Startup/IosDependencyContainer.cs
amulware: We agreed in a call that the number of suggestions should be up to the UI directly, where the suggestions service is used. @heytherewill is creating another issue for that refactor.

The remove config thing should be in the appropriate service directly most likely, and the other things should be in an injected config object, with the version number which is already injected.

Toggl.Daneel/ViewControllers/Calendar/CalendarViewController.cs
amulware: I understand the argument. The prefix naming still bothers me, but that's something for the future.

Toggl.Foundation/DependencyContainer.cs
amulware: We only want to force wait on sync on login, not on cold start for example.

amulware:
@simonrozsival @kornelijepetak Please keep an eye on the changes around how full sync is being triggered during login and otherwise (around the creation of a new sync manager) to make sure we don't change behaviour. We never wanna wait on full sync, except for the login case I believe. (not sure about the re-login for the reset-api-token thing? I don't think so though?)

Toggl.Foundation/DependencyContainer.cs
juanlaube: I think this is just fine, the problem was not the analytics service initialized late, but the appcenter configuration happening after the crash, afaik it's the appcenter sdk that sends crash reports.

juanlaube:
I'm testing this on Android, my first impression is the app is faster. I will post if I can find bugs.

Toggl.Foundation/UserAccessManager/UserAccessManager.cs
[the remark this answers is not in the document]
kornelijepetak: Ok. But when I said our own, I meant literally this.

```cs
class TogglLazy<T> : Lazy<T> 
{
	public TogglLazy(Func<T> creation) : base(creation)	{ }
	public static implicit operator T(TogglLazy<T> lazy) 
		=> lazy.Value;
}
```
with a proper name, of course.

Thes relies on efficient/safe impl of `Lazy<T>` with an additional method call during construction (which in a sense of efficiency and our use cases is nothing.)

Toggl.Foundation.MvvmCross/Toggl.Foundation.MvvmCross.csproj
```diff
@@ -22,7 +22,6 @@
     <PackageReference Include="MvvmCross" Version="6.1.2" />
   </ItemGroup>
   <ItemGroup>
-    <Folder Include="ViewModels\Settings\" />
-    <Folder Include="ViewModels\TimeEntriesLog\" />
+    <Service Include="{508349b6-6b84-4df5-91f0-309beebad82d}" />
```
daividssilverio: What is this?

Toggl.Foundation.MvvmCross/ViewModelLocator.cs
```diff
@@ -0,0 +1,417 @@
...
+                    dependencyContainer.NavigationService,
+                    dependencyContainer.RxActionFactory);
+
+            if (viewModelType == typeof(SelectDateFormatViewModel))
+                return new SelectDateFormatViewModel(
+                    dependencyContainer.NavigationService,
+                    dependencyContainer.RxActionFactory);
+
+            if (viewModelType == typeof(SelectDateTimeViewModel))
+                return new SelectDateTimeViewModel(
```
daividssilverio: This constructor is now missing the `RxActtionFactory`

Toggl.Giskard/Startup/Setup.cs
```diff
@@ -175,19 +102,6 @@ protected override void PerformBootstrapActions()
             // calls their Run method. We can skip it since we don't have such classes.
         }
 
-        void ensureDataSourceInitializationIfLoggedIn()
```
daividssilverio: it's good to see this gone

PR #4796, opened 2019-03-26
Fix crash in adapter when list of items is null

### LINKED ISSUE #4785: Occasional crashes after starting a time entry immediately after signup.
## 🐛 Bug Report
App crashes immediately after signup

## Context
I had signed in to the app so this was occurring after I logged out and signed up for a new account. Doesn't seem to occur with a fresh install but I can't rule it out as haven't tested that many times.
This has happened at least 4 or 5 times so far.

Build/Branch:1.12
Phone OS:8.0
Phone model:HTC U Ultra

## Expected behavior
Being signed in after signing up or creating a new time entry after that.

## To reproduce
Sign out of an account
Sign up with a new one
(crash may occur)
Start a new entry
(crash may occur)

## Suggestions
There should be some crash reports and I sent feedback through from the crashes.

juanlaube:
## What's this?
This solves a crash in the start time entry view.
It's another crash that I couldn't reproduce but the solution is based on the stacktrace. In this case, `newItems` is the only thing that could be a null reference so breaking the iterator earlier solves the problem.

### Relationships

Closes #4785 

## Why do we want this?
🚫 💥 

## How is it done?
If `newItems` is null, just return early.

### Why not another way?
🤷‍♂️ 

### Side effects
None

## Review hints
Could anything else in `flattenItems` be null? 🤔 

## :squid: Permissions
⚠️ Only with permission from the release manager ⚠️


amulware:
I think this won't do it. The crash occurs INSIDE `MoveNext` which happens in the foreach statement after calling `newItems.Count` and `newItems.GetEnumerator()` which both appear to succeed.

amulware:
Except it's the MoveNext() of ToList() so my comment is actually irrelevant.

juanlaube:
I rewrote `flattenItems` as a regular loop instead of using lazy enumerables.

Toggl.Giskard/Adapters/BaseSectionedRecyclerAdapter.cs
```diff
@@ -125,21 +125,28 @@ public void SetItems(IList<CollectionSection<TSection, TItem>> newItems)
...
         }
 
-        private IEnumerable<Either<TSection, TItem>> flattenItems(IList<CollectionSection<TSection, TItem>> newItems)
+        private IList<Either<TSection, TItem>> flattenItems(IList<CollectionSection<TSection, TItem>> newItems)
         {
-            var hasMultipleSections = newItems.Count > 1;
+            var flattenedItems = new List<Either<TSection, TItem>>();
 
-            foreach (var section in newItems)
+            if (newItems != null)
```
daividssilverio: Invert condition and return early to reduce nesting.

Toggl.Giskard/Adapters/BaseSectionedRecyclerAdapter.cs
```diff
@@ -125,21 +125,28 @@ public void SetItems(IList<CollectionSection<TSection, TItem>> newItems)
...
+                foreach (var section in newItems)
                 {
-                    yield return Either<TSection, TItem>.WithRight(item);
+                    var shouldIncludeHeader = hasMultipleSections && !(section.Header?.Equals(default(TSection)) ?? true);
+                    if (shouldIncludeHeader)
+                        flattenedItems.Add(Either<TSection, TItem>.WithLeft(section.Header));
+
+                    foreach (var item in section.Items)
+                    {
+                        flattenedItems.Add(Either<TSection, TItem>.WithRight(item));
```
daividssilverio: It would be nice to be able to use `AddRange` instead. But it seems it doesn't work with `Either`.

daividssilverio: Oh, Rider for the help:
```
flattenedItems.AddRange(section.Items.Select(Either<TSection, TItem>.WithRight));
```

PR #4800, opened 2019-03-26
Fix missing project label on start te vm

### LINKED ISSUE #4797: Start time entry time entry suggestions don't display the right project color
## 🐛 Bug Report
The start time entry isn't showing the color of the projects in the StartTimeEntry view.

image

## Context
Build/Branch: release/giskard-1.12-argon
Phone OS: Android
Phone model: Tested on Lollipop & Pie.

## Expected behaviour
We should be seeing the projects there.

## To reproduce
1. Create a time entry with a project.
2. Start the create time entry view again and type something that would suggest the previously created time entry.
4. You should see time entry suggestion, like in the screenshot example, having the space for the project description, but all white.

## Suggestions
It's mostly likely something wrong with the view binding or the

daividssilverio:
## What's this?
This fixes the logic that sets the visibility of the project and task labels suggestions in the start time entry view.

### Relationships
Closes #4797

## Why do we want this?
Take a look at the screenshots in #4797, we don't want that.

## How is it done?
It just makes sure that the task label visibility is changed based on the presence of a task in the time entry suggestion, instead of the project label.

### Why not another way?
That's the right way.

### Side effects
None.

## Review hints
Try to replicate it on the base branch.
Try again on this branch.

## :squid: Permissions
⚠️ Only with permission from the release manager ⚠️


amulware:
Go ahead.

PR #4802, opened 2019-03-26
Clear swipe backgrounds when clearing views on MainRecyclerViewTouchCallback clear callback

### LINKED ISSUE #4433: Deleted and Continue bars get stuck on the screen.
## 🐛 Bug Report
Continued 5+ entries and Deleted 5+ entries in quick succession. Haven't discovered the lower limit that this requires yet but the Continue and Delete bars gets stuck on the screen under some UI elements.
image

I made you a video but it's MP4 so here's a link.
https://toggl.slack.com/archives/C02APMYTM/p1550608375956600

## Context
Continued and deleted several entries on an old workspace in succession (all entries are in the one workspace)

Build/Branch: 1.11
Phone OS: 8.0.0
Phone model: HTC U Ultra

## Expected behavior
These should disappear completely once done

## To reproduce
Continue and delete lots of things quickly.

## Suggestions
https://youtu.be/xhy7dXWjpAA

daividssilverio:
## What's this?
This fixes the delete and continue backgrounds that get stuck when you go loco in the main log view.

### Relationships
Closes #4433

## Why do we want this?
Because we want our ui to look good and work well no matter how it's used. (Actually, deleting a bunch o time entries very fast doesn't look like a very weird use case to me).

## How is it done?
The view holder callback to clear the swipe backgrounds is called when the `MainRecyclerViewTouchCallback` callback to clear the viewholder when the swiping ends.

### Why not another way?
🤔 I don't know. Please give your suggestion.

### Side effects
None.

## Review hints
Try to replicate the bug on this branch. You can't be sure it works just based on the change.

## :squid: Permissions
🦑 it

### DIFF
--- a/Toggl.Giskard/ViewHelpers/MainRecyclerViewTouchCallback.cs
+++ b/Toggl.Giskard/ViewHelpers/MainRecyclerViewTouchCallback.cs
@@ -89,6 +89,7 @@ namespace Toggl.Giskard.ViewHelpers
         {
             if (viewHolder is MainLogCellViewHolder logViewHolder)
             {
+                logViewHolder.HideSwipeBackgrounds();
                 DefaultUIUtil.ClearView(logViewHolder.MainLogContentView);
             }
             else


daividssilverio:
Closing this as it doesn't fix the problem yet. I'll try again later if no one does.

juanlaube:
This doesn't solve the problem I'm afraid, I can still reproduce it on a device 😢

PR #4810, opened 2019-03-27
Fix invisible project label in Start TE suggestions

### LINKED ISSUE #4797: Start time entry time entry suggestions don't display the right project color

kornelijepetak:
## What's this?
Fixes the invisible project label.

### Relationships
Closes #4797 

## Why do we want this?
So that user can see the project of TE suggestions.

## How is it done?
It was an honest mistake. The wrong label was being hidden. Now, it's fixed.

### Why not another way?
There is no better/other way.

## :squid: Permissions
Anyone.


daividssilverio:
🤔 I did this yesterday here: https://github.com/toggl/mobileapp/pull/4800

daividssilverio:
Although you seem to be doing more than just fixing the project label. Please close #4800 if you think this is the complete solution and add more details of what it is fixing with your other changes.

kornelijepetak:
@daividssilverio I wasn't doing more, I just failed to correctly base the PR. It should've been the giskard 1.12, not `develop`.

In any case, closing in favor of #4800

PR #4824, opened 2019-03-28
Fix missing FAB icon after visibility changes

### LINKED ISSUE #4801: + and play icon not appearing after switching to manual mode
## 🐛 Bug Report
Use the app for some time maybe several hours, create a few entries, access reports then switch manual mode on and the + icon doesn't display, switch manual mode off and the play icon doesn't show. Just the green circle is displayed

## Context
Used the app for some time previously and left it open

Build/Branch:1.12
Phone OS:8.0
Phone model:HTC U Ultra

## Expected behavior
Play button and + icon should show

## To reproduce
~Use the app for a little while then switch Manual mode on~
Updated (and consistent) [repro steps](https://github.com/toggl/mobileapp/issues/4801#issuecomment-477172042)

## Suggestions
image

kornelijepetak:
## What's this?
Fix for a missing image on the FAB after changing visibility.

### Relationships
Closes #4801 

## Why do we want this?
To be nice and peachy.

## How is it done?
I've researched a bit more into the actual problem and seems that the problem is that the scale of a matrix is not set correctly in SDK code.

### Why not another way?
Among about 12 things I've tried, this one is the only one that actually worked.

## Review hints
If you have a better name for the `*Safe` method, please let me know.

## :squid: Permissions
Anyone who wants to see 1.12 released asap. But only after functional test. 😛


amulware:
Go ahead with the squash once approved.

PR #4847, opened 2019-03-29
Remove billable flag when the workspace is not paid

### LINKED ISSUE #4845: The billable flag is not unset when switching to a project in a free workspace
Steps to reproduce:
- create a time entry in a paid workspace
- set the time entry as billable
- change the project to a project in a free workspace
- the billable flag `$` is still visible in the main log, there's no problem with syncing

Expected behavior:
- the billable flag is unset
- there's no `$` visible anymore

simonrozsival:
## What's this?
This PR fixes a bug.

### Relationships
Closes #4845 

## Why do we want this?
It makes sense and product team confirmed that this is a bug.

## How is it done?
It's a simple few lines of code.

### Why not another way?
This seemed like the easiest way to do it to me.

### Side effects
It doesn't work... 😆 I think there is a bug in the API endpoint for updating TEs, I'll have to investigate this further next week:
image

1. there is a TE in a premium workspace, which has the billable flag set
2. I change the workspace of the TE to a different one, which "is free", and the billable flag is set to `false` (project is set to `null`)
3. the update succeeds, the workspace ID changes, project ID changes, billable flag stays set to `true` even though the workspace does not allow it

## Review hints
The billable flag is reset when the time entry syncs. You can try this offline and see, that our app's state updates correctly. We can merge this, and it will work effectively once BE fixes this bug (if it is a bug, maybe I'm holding it wrong, and I'll be able to fix it in the next week, but it doesn't seem to be the case).

## :squid: Permissions
Let me do it.

### DIFF
--- a/Toggl.Foundation.MvvmCross/ViewModels/EditTimeEntryViewModel.cs
+++ b/Toggl.Foundation.MvvmCross/ViewModels/EditTimeEntryViewModel.cs
@@ -159,16 +159,24 @@ namespace Toggl.Foundation.MvvmCross.ViewModels
             ProjectClientTask = projectClientTaskSubject
                 .AsDriver(ProjectClientTaskInfo.Empty, schedulerProvider);
 
-            IsBillableAvailable = workspaceIdSubject
-                .Where(id => id.HasValue)
-                .SelectMany(workspaceId => interactorFactory.IsBillableAvailableForWorkspace(workspaceId.Value).Execute())
-                .DistinctUntilChanged()
-                .AsDriver(false, schedulerProvider);
-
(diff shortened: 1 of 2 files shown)


khoi:
@maritoggl should we do this in the release? or knowingly shipping this bug until we sort this out properly with BE?

simonrozsival:
Let's not include this in 1.20. After reading the messages from the backend team, I am worried that there's nothing we can really do about this. The "bug" in the API is probably not a bug, but it is just a result of undefined/not-well-defined behavior of transferring TEs between workspaces. I'm afraid this would just introduce a visual glitch. We might want to re-think the way we implement moving TEs between workspaces (delete from the old one, create new with same data in the new one).

PR #4892, opened 2019-04-08
Fix Android builds on new stack

heytherewill:
## What's this?
This removes `armeabi` from our builds and fixes the ambiguous references to HashCode

## Why do we want this?
Removing `armeabi`: This is no longer a valid configuration and Visual Studio is enforcing it. Bitrise updated their stack, and now we need to comply in order for the apps to build. For more info read [here](https://stackoverflow.com/questions/28926101/is-it-safe-to-support-only-armeabi-v7a-for-android-4-and-above)
HashCode: New class introduced in `System`, so we need to make it unambiguous.

## How is it done?
Removing it from the `.csproj` and fixing references, respectively

### Side effects
We will no longer ship the armeabi binaries, but I am afraid there isn't much we can do about it, since Google does not support it anyways.

## :squid: Permissions
Anyone so we can build things again


daividssilverio:
Manually checked for armeabi devices on the supported list from our GooglePlay listing. We are fine.

PR #4903, opened 2019-04-10
Fix get multiple entities by ids

### LINKED ISSUE #4895: Some TE details not saved when TE created via calendar view
## 🐛 Bug Report
When creating a time entry using the calendar view, time entry details (description, project) do not get saved. TE is created, but has no description or a project attached to it. 

## Context

Build/Branch: Daneel 1.20
Phone OS: iOS 12.2
Phone model: iPhone 7

## Expected behavior
TE details that I specify upon creating a time entry using the calendar view get saved at first attempt. 

## To reproduce
Open the calendar view on an iphone, 
long press the view to create a new time entry, 
type in a description, 
perhaps select a project as well, 
save.

TE gets saved with no description and project.

juanlaube:
## What's this?
This solves a problem in the get multiple by id interactor. It wasn't considering the original id of the entity.

### Relationships

Closes #4895 

## Why do we want this?
🚫 🐛 

## How is it done?
Adding a `GetByIds` method to `IRepository` and solving the query with expression trees in `RealmAdapter` because `Contains` is not supported by Realm.

### Why not another way?
Is there another way? 🤔 

### Side effects
None.

## Review hints
Try it.

## :squid: Permissions
🦑


Toggl.PrimeRadiant.Realm/Repository.cs
```diff
@@ -14,28 +16,55 @@ public Repository(IRealmAdapter<TModel> adapter)
...
+        }
+
         private static long getId<TRealmEntity>(TRealmEntity entity)
             where TRealmEntity : RealmObject, IIdentifiable, TModel
             => entity.Id;
+
+        private static Expression contains<T>(Expression nullableProperty, T[] ids)
+        {
+            var equalsList = ids.Select(id => Expression.Equal(nullableProperty, Expression.Convert(Expression.Constant(id), typeof(T))));
+            return equalsList.Aggregate(Expression.Constant(false) as Expression, (partial, expr) => Expression.OrElse(expr, partial));
```
juanlaube: For reviewers:
`Expression.Empty()` looks nicer but doesn't work 😢

juanlaube:
@kornelijepetak I just realized this needs to go to Giskard 1.13 also because this affects editing grouped time entries, pinging you since you're the release manager.

Toggl.PrimeRadiant.Realm/Repository.cs
```diff
@@ -14,28 +16,55 @@ public Repository(IRealmAdapter<TModel> adapter)
...
         private static Expression<Func<TRealmEntity, bool>> matchById<TRealmEntity>(long id)
             where TRealmEntity : RealmObject, IIdentifiable, IModifiableId, TModel, IUpdatesFrom<TModel>
             => x => x.Id == id || x.OriginalId == id;
 
+        private static Expression<Func<TRealmEntity, bool>> matchByIds<TRealmEntity>(long[] ids)
+            where TRealmEntity : RealmObject, IIdentifiable, IModifiableId, TModel, IUpdatesFrom<TModel>
+        {
+            var nullableIds = ids.Select(id => (long?)id).ToArray();
+
+            var entityParam = Expression.Parameter(typeof(TRealmEntity), "entity");
```
daividssilverio: could you make this "entity" string a constant? What does it represent? A column?

Toggl.PrimeRadiant.Realm/Repository.cs
```diff
@@ -14,28 +16,55 @@ public Repository(IRealmAdapter<TModel> adapter)
...
+            var idProperty = Expression.Property(entityParam, typeof(TRealmEntity), "Id");
```
daividssilverio: Same here.

Toggl.PrimeRadiant.Realm/Repository.cs
kornelijepetak: It's a bit more complicated than that. In this particular case, I think that making a constant to represent this constant would be overkill. This is like saying `I want a parameter name to some function to be named 'entity'` in _compiled_ code. Except it's not compiled.

Toggl.PrimeRadiant.Realm/SingleObjectStorage.cs
```diff
@@ -57,7 +57,9 @@ public IObservable<TModel> Single()
             where TRealmEntity : RealmObject, TModel, IUpdatesFrom<TModel>
             => new SingleObjectStorage<TModel>(new RealmAdapter<TRealmEntity, TModel>(
                 getRealmInstance,
-                convertToRealm, _ => __ => true,
+                convertToRealm,
+                _ => __ => true,
```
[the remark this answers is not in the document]
simonrozsival: `AlwaysTrue<T>(T _) => true;` and `AlwaysTrue<T1,  T2>(T1 _, T2 __) => true;` would be awesome.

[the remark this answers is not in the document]
simonrozsival: (except I'm not sure if that would even compile :D)

kornelijepetak:
@juanlaube I'm confused as I've tried to

* Go offline
* Create a new TE
* Create a new project ABC (which now should not have .Id, but only OriginalId)
* Create a new project DEF
* Create a new TE
* Edit a TE
* Change description
* Change the project to one that has a database ID
* Change the project to one ABC
* Save

I can see everything saved correctly 😱

@simonrozsival would you know why this would work on Android?

Toggl.PrimeRadiant.Realm/Repository.cs
juanlaube: I agree with @kornelijepetak, that would be overkill since that is just the name of the parameter in the lambda expression and it's not reused.

Toggl.PrimeRadiant.Realm/SingleObjectStorage.cs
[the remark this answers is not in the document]
juanlaube: I will try :)

Toggl.PrimeRadiant.Realm/Repository.cs
daividssilverio: I agree with you guys, I didn't realize what the statement was doing when I saw it. Thanks for clarifying.

Toggl.PrimeRadiant.Realm/SingleObjectStorage.cs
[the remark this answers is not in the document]
juanlaube: The generic function doesn't work I'm afraid 😞

juanlaube:
@kornelijepetak the Android build problem was the upgrade to VS2019, updating the branch solved it.
@khoi you will need to cherry-pick only the fix and the VS2019 update fix, make sure nothing else goes into the hotfix release branch.

khoi:
I'll do it after @kornelijepetak's approval

Toggl.PrimeRadiant.Realm/SingleObjectStorage.cs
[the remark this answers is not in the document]
kornelijepetak: Oh, wait. You can't do what I wanted. This is an Expression, not a delegate. My bad. Ignore my request for this kind of function. At least not in this use case.

simonrozsival:
The problem actually doesn't happen when you are offline, because both the original ID and the Id itself are set to the same (negative) local Id.

The problem doesn't even happen on the main log, even when you are online, because the row always has an up-to-date information about the Id of the row (it is reloaded after sync).

The problem is with iOS calendar (maybe also the Android one, I haven’t tested that one), where you can create a TE, it is shown as a colored rectangle, which remembers the local, negative, Id of the TE. The app than has to sync, changing the Id to a positive Id from the server, but the calendar items don’t reload. When you tap on the calendar item, it tries to open the Edit view with the negative Id.

The “get by id” method can deal with id/original Id, but our new “get multiple by id” could not.

If the Android calendar reloads after sync, the bug doesn’t affect it.

Does this answer your question sufficiently?

kornelijepetak:
This issue seems to not affect the calendar on Android at all. I have labeled this as `iOs`.

khoi:
- rebased on top of `release/daneel-1.20.1-ifni-hotfix`

PR #4913, opened 2019-04-12
1.20.1 hotfix

khoi:
## Changelog

```
Fixes a problem when creating time entries in the Calendar View
```

## :squid: Permissions
me


amulware:
'Fixes a problem when creating time entries in Calendar'

- 'a' instead of 'the', since the problem was not mentioned before
- 'time entries', plural
- no full stop

PR #4915, opened 2019-04-12
Fix crash in the wheel view when endTime < startTime

### LINKED ISSUE #4875: ArgumentOutOfRangeException in PingPongClamp
## 💥 Crash Report
Something goes wrong with index calculations when getting the colour for the wheel on Android (does not happen on iOS).

Here is the code calculating the index:

https://github.com/toggl/mobileapp/blob/8ace74985a26eb976ff0d3a25d74b91da55f74ac/Toggl.Giskard/Views/EditDuration/WheelForegroundView.cs#L58-L62

Looks to me like it's not accounting for start time > end time (which is strange, I grant you...).

Could point at an underlying issue, or maybe this simply happens in cases where someone manages to move a start time into the future (should be easily doable with wrong time settings).

### 👥 Impact
Not a very common issue, but one that should be easy to fix.

Last occurred on 1.12 (current release as of writing).

### 🛰 Link
https://appcenter.ms/orgs/Toggl/apps/Toggl-Android/crashes/errors/3311916819u/overview

`System.ArgumentOutOfRangeException: Specified argument was out of the range of valid values. Parameter name: The clamped number a non-negative integer, -17690556 given.`
```cs
Math.PingPongClamp (System.Int32 number, System.Int32 length)
ArrayExtensions.GetPingPongIndexedItem[T] (T[] array, System.Int32 index)
WheelForegroundView.get_backgroundColor ()
WheelForegroundView.setupDrawingDelegates ()
WheelForegroundView.OnLayout (System.Boolean changed, System.Int32 left, System.Int32 top, System.Int32 right, System.Int32 bottom)
View.n_OnLayout_ZIIII (System.IntPtr jnienv, System.IntPtr native__this, System.Boolean changed, System.Int32 left, System.Int32 top, System.Int32 right, System.Int32 bottom)
(wrapper dynamic-method) System.Object.68(intptr,intptr,bool,int,int,int,int)
```

juanlaube:
## What's this?
Fixes a crash in the wheel color selection. In some scenario that I couldn't reproduce, endTime < startTime and numberOfFullLoops was < 0, which is used as the index of the color.

### Relationships

Closes #4875 

## Why do we want this?
🚫 💥 

## How is it done?
`Math.Max` ftw!

### Why not another way?
🤷‍♂️ 

### Side effects
The bug was only for Android, but I used some defensive-programming-fu and did the same in iOS too.

## Review hints
Would be nice if you can reproduce the crash first.

## :squid: Permissions
🦑


Toggl.Droid/Views/EditDuration/WheelForegroundView.cs
```diff
@@ -55,7 +55,7 @@ public class WheelForegroundView : View
         private WheelUpdateType updateType;
         private double editBothAtOnceStartTimeAngleOffset;
 
-        private int numberOfFullLoops => (int) ((EndTime - StartTime).TotalMinutes / MinutesInAnHour);
+        private int numberOfFullLoops => Math.Max(0, (int)((EndTime - StartTime).TotalMinutes / MinutesInAnHour));
```
heytherewill: I wonder if that shouldn't be in some shared layer. Maybe not for this PR, but for the future

juanlaube: I changed this a bit, but I your comment is still valid, most of the wheel math could be in a shared layer I think.

Toggl.Droid/Views/EditDuration/WheelForegroundView.cs
```diff
@@ -55,7 +55,7 @@ public class WheelForegroundView : View
         private WheelUpdateType updateType;
         private double editBothAtOnceStartTimeAngleOffset;
 
-        private int numberOfFullLoops => (int) ((EndTime - StartTime).TotalMinutes / MinutesInAnHour);
+        private int numberOfFullLoops => (int)((EndTime - StartTime).TotalMinutes / MinutesInAnHour);
```
heytherewill: This change is not really needed

PR #4917, opened 2019-04-12
Make iOS builds sanic again!

juanlaube:
## What's this?
Make iOS builds sanic again!

## Why do we want this?
We want fast iOS builds for commits.

## How is it done?
This has three parts:
1. a new configuration in the solution that builds Daneel only in debug for the iPhoneSimulator platform, it's named `Debug.iOS.Fast`
2. a new task in cake that builds in the previous configuration and platform, named `Build.Debug.iOS`
3. a new workflow in bitrise that builds the previous cake task, named `Tests.AppBuilds.iOS.Fast`

### Why not another way?
We can still tweak a lot of things, but this cuts the builds to 6 minutes (from about 40!) and that's a start.

### Side effects
None?

## Review hints
Successful build: https://app.bitrise.io/build/e6afc870fabf75e1
Broken build on purpose: https://app.bitrise.io/build/793272988ebd1684

## :squid: Permissions
🚫 Needs @amulware approval and we need to integrate this into the GitHub checks first.


juanlaube:
The irony when you make the iOS build faster, but the GitHub check is still the slow build 😄

amulware:
Why is this faster?

juanlaube:
Some clarification: we don't need a new GitHub check, once this is approved I will delete the workflow I added for testing on bitrise and replace the cake task in the current workflow used for the GitHub check.

juanlaube:
@amulware debug builds are much faster. There are no optimizations, no llvm code generation, no code signing... basically this is just code compilation.

amulware:
I suspected that those things were the cause of these slow build times, but I am surprised at how much of a difference they make. A whole new era of productivity? :D

amulware:
Oh, one thing though: Do we still retain all the safety of the previous check? It seems to me like this faster build will miss at least _some_ issues we would catch otherwise?

heytherewill:
This is by no means less safe. Compilation-wise, both builds behave the same

amulware:
Yet we might not detect issues with signing or other weird framework glitches. That being said, that might actually be a positive, since our code changes are unlikely to introduce these. They are more likely caused by external factors and we should be able to catch them with nightly builds, and even more so with UI tests, once automated.

juanlaube:
Ok, I missed a lot of the conversation while I was having lunch 🌯 
I agree with William, compilation-wise, this is the same, and we still have ad hoc builds running the full compilation package 📦

Issue #4187, opened 2019-01-14
Remove Toggl.Foundation.Helper.Color


arturs105:
Depends on #4181
Depends on #4186

This issue addresses the `Toggl.Foundation.Helper.Color` class.

It has a few public members that we can move around and then remove the file:
* `IsValidHexColor`. I think it makes sense that this would be a static method on `Multivac.Color`.
* `DefaultProjectColors`, `NoProject` and `ClientNameColor` should be moved to `Colors.cs` file from #4186

heytherewill:
I am a bit against this one. I agree that it should be named Colors, but I don't know if it should live in Toggl.Foundation when color is clearly a UI related concept.

PR #4920, opened 2019-04-12
Remove references to MvxColor

### LINKED ISSUE #4186: Remove MvxColor definitions
Once #4181 is merged, we can start moving away from `MvxColor` as a part of MvExit.

This issue is for the long file, that contains the color definitions - `Toggl.Foundation.MvvmCross.Helper.Color`.

We now have the option to move all the defined colors in the `Multiac.Color` struct, but i think the color definitions should remain in a separate file, so the color struct doesn't get too bloated and is easier to read.

Furthermore, i think we should rename the file with definitions to plural `Colors`, because currently we have 3 files named `Color.cs` and file searching is confusing because of that. There will be a separate issue to remove the third `Color.cs`, so in the end we will only have `Color.cs` and `Colors.cs`.

heytherewill:
## What's this?
This completely replaces `MvxColor` with `Shared.Color`

### Relationships
Closes #4186
Ref #4187

## Why do we want this?
MVExit. This touches too many files and we want it to be in develop ASAP.

## How is it done?
We are removing the references to MvxColor. Then we add the necessary code to make our Color struct behave like MvvmCross'. Then we delete the Color plugin.

### Why not another way?
No other way tbh.

### Side effects
Functional test to be sure, but there should be none.

## Review hints
Per commit.

## :squid: Permissions
Only myself.


juanlaube:
When the hex string is null or empty, the color should be black to keep the same behavior as mvvmcross for now.

juanlaube:
I functionally tested both iOS and Android

Issue #4905, opened 2019-04-11
Modal doesn't resize when width changes on iPad


khoi:
## 🐛 Bug Report
When a modal (popup) is presented, and the width of the application changed, the modal doesn't resize itself.

## Context
Build/Branch: `project/ipad`
Phone OS: iOS
Phone model: iPad

## Expected behavior
- it should be resized properly

## To reproduce
- Signup
- "Review the Terms" modal appears.
- Split the ipad screen

khoi:
![IMG_0916](https://user-images.githubusercontent.com/6994441/55939485-0f9fdd00-5c68-11e9-9db3-f5b85e6669e9.PNG)

sofokli1:
@dshr I think this only happens when presenting using `ModalDialogPresentationController.cs`

PR #4927, opened 2019-04-15
iPad: Fix modal dialog positioning

### LINKED ISSUE #4924: iPad: Modal positioning breaks after going from 2/3 to fullscreen with overlay
## 🐛 Bug Report
1. Have the 2/3 split
2. Start a new TE
3. Open the Change Start Date modal
4. Change the 1/3 app to overlay
5. The Change Start Date modal is positioned weirdly.

## Context
Build/Branch: project/ipad
Phone OS: 12.1
Phone model: iPad Mini

## Expected behavior
It should be centered.

dshr:
## What's this?
A small one-liner that fixes problems with our modal dialogs.

### Relationships
Closes #4924 and #4905 

## Why do we want this?
Consistency

## How is it done?
By mimicking the behaviour of our regular modals.

### Why not another way?
This seems to be the simplest way.

### Side effects
None.

## Review hints
Open a modal dialog (like the select TE start date), and try changing the size class (e.g. 2/3 split to fullscreen with overlay)

## :squid: Permissions
iPad crew, I choose you!


Toggl.Daneel/Presentation/Transition/ModalDialogPresentationController.cs
```diff
@@ -1,4 +1,4 @@
-﻿using CoreGraphics;
+﻿    using CoreGraphics;
```
sofokli1: 👀

PR #4931, opened 2019-04-15
iPad: Fix onboarding crash

### LINKED ISSUE #4929: iPad: Crash on Onboarding view
## 🐛 Bug Report
Starting the app in split screen when not logged in causes a crash.

## Context
Build/Branch: project/ipad
Phone OS: iOS
Phone model: iPad Pro

## Expected behavior
Should run fine

## To reproduce
Log out, re-launch app in split-screen mode.

dshr:
## What's this?
A small change to fix a crash during onboarding.

### Relationships
Closes #4929

## Why do we want this?
💥 🔫 

## How is it done?
By adding some defensive programming :(

### Why not another way?
This seems to be the simplest option.

### Side effects
None.

## :squid: Permissions
iPad crew only


Toggl.Daneel/ViewControllers/OnboardingViewController.cs
```diff
@@ -146,8 +146,11 @@ public override void ViewDidLayoutSubviews()
         public override void ViewWillTransitionToSize(CGSize toSize, IUIViewControllerTransitionCoordinator coordinator)
         {
             base.ViewWillTransitionToSize(toSize, coordinator);
-            var scrollPoint = new CGPoint(toSize.Width * PageControl.CurrentPage, 0);
-            ScrollView.SetContentOffset(scrollPoint, true);
+            if (PageControl != null)
```
heytherewill: [Prefer early returning over nesting](https://github.com/toggl/mobileapp/blob/develop/docs/style-guide.md#L179)

sofokli1:
How can i reproduce the crash?

dshr:
@sofokli1 log out, start a different app fullscreen, then start Toggl in a 1/3 split.

PR #4932, opened 2019-04-15
iPad: Fix bug that caused action dismiss action sheets to break

### LINKED ISSUE #4906: Dismiss action sheet disappears on iPad
## 🐛 Bug Report
The native action sheet disappears if there is a modal displaying in the app. 

## Context
Build/Branch: `project/ipad`
Phone OS: iOS
Phone model: iPad
 
## Expected behavior
Native action sheet always on top

## To reproduce
- make changes to an entry
- tap outside the view to show the native action sheet
- tap outside again and the action sheet disappears

dshr:
## What's this?
A small fix to an action sheet bug

### Relationships
Closes #4906 

## Why do we want this?
We don't want our action sheets broken, do we?

## How is it done?
By adding the properly needed .Cancel action to the action sheet.

### Why not another way?
This is the correct Apple-approved way

## Review hints
Try and replicate the behaviour from the issue

## :squid: Permissions
iPad crew, just you

### DIFF
--- a/Toggl.Daneel/Services/DialogServiceIos.cs
+++ b/Toggl.Daneel/Services/DialogServiceIos.cs
@@ -71,11 +71,7 @@ namespace Toggl.Daneel.Services
                     preferredStyle: UIAlertControllerStyle.ActionSheet
                 );
 
-                var actionStyleForCancel = UIDevice.CurrentDevice.UserInterfaceIdiom == UIUserInterfaceIdiom.Pad
-                    ? UIAlertActionStyle.Default
-                    : UIAlertActionStyle.Cancel;
-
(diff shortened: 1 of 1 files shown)


khoi:
>This is the correct Apple-approved way

Can you link to the Apple doc here? Having an extra "cancel action" for the iPad is something we have never seen before

dshr:
It's not exactly an "extra" action per se, more like the required action. Like, according to https://developer.apple.com/documentation/uikit/uialertactionstyle we have 3 action styles: destructive, default and cancel. Destructive shows a red button, default shows a blue button and the cancel action is only executed when the actionsheet is dismissed. In our case, there was no cancel action provided.

PR #4941, opened 2019-04-16
iPad: Better reports sizing

### LINKED ISSUE #4935: iPad: Report view weird size changes
## 🐛 Bug Report
The report view's width changes weirdly when selecting "Yesterday"

## Context
Build/Branch: project/ipad
Phone OS: 12.1
Phone model: iPad

## Expected behavior
It should be the same for all dates.

## To reproduce
Select "yesterday".

dshr:
## What's this?
A set of AL and code changes to fix our reports view conundrum.

### Relationships
Closes #4935 

## Why do we want this?
✨ 

## How is it done?
By setting the max width programmatically instead of AL.

### Why not another way?
Seems the simplest tbh.

### Side effects
None.

## :squid: Permissions
iPad crew only


dshr:
@sofokli1 here are the screencaps you asked for:
Before:
shrunk
After:
unshrunk

PR #4943, opened 2019-04-17
💥 Fix crash with armeabi v7

kornelijepetak:
## What's this?
This will probably fix problems with armeabi v7 builds (debugs and releases)

### Relationships
The problem (and the workaround) is described here:

xamarin/xamarin-android/issues/2966
xamarin/xamarin-android/issues/2920
xamarin/xamarin-android/issues/2960

Refs: #4944 

## Why do we want this?
No 💥

## How is it done?
By preventing 64bit architectures run 32bit version of the app.

### Why not another way?
When `Xamarin.Android` is updated with a fix, this change will no longer be necessary.

### Side effects
Not that we know of.

## Review hints
This applies the fix from the mentioned `xamarin-android` repo links, so feel free to read about it yoursef.

## :squid: Permissions
Me, @heytherewill

### DIFF
--- a/Toggl.Droid/Toggl.Droid.csproj
+++ b/Toggl.Droid/Toggl.Droid.csproj
@@ -20,6 +20,7 @@
     <EnableProguard>true</EnableProguard>
     <JavaMaximumHeapSize>2G</JavaMaximumHeapSize>
     <AndroidUseSharedRuntime>false</AndroidUseSharedRuntime>
+    <AndroidAotAdditionalArguments>llvmopts="-data-layout='e-p:32:32-n32-S64' -O2 -disable-tail-calls"</AndroidAotAdditionalArguments>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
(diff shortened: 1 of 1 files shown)


kornelijepetak:
Fixing this as a hotfix instead #4946

brendanzagaeski:
Hopefully I'm replying in time to help catch my mistake.

Another user and I were re-examining this problem today, and we discovered something I should have realized sooner. The workaround of adjusting `$(AndroidAotAdditionalArguments)` is only valid if the project is targeting *no architectures other than armeabi-v7a*.  It is important *not* to have those adjustments applied to the arm64-v8a architecture. Otherwise the original crash symptoms will happen again, but this time on the 64-bit architecture.

So the approach in this pull request will unfortunately cause trouble on the arm64-v8a architecture. 

At least in theory, there is a different a way to work around the issue by customizing the whole `_BuildApkEmbed` target (see <https://github.com/xamarin/xamarin-android/issues/2966#issuecomment-484281295> for details), but that's at best inelegant, so if possible it might be best to wait for a servicing update of Xamarin.Android that includes the fix.

PR #4965, opened 2019-04-22
Track everything we control in the droid handler scheduler

daividssilverio:
## What's this?
This is an attempt to find more about one of our major source of crashes in the droid app.

### Relationships
It could help #4230 

## Why do we want this?
Based on the stacktrace we currently have:
```
System.NullReferenceException: Object reference not set to an instance of an object
  at System.Reactive.Concurrency.ConcurrencyAbstractionLayerImpl+<>c.<StartThread>b__8_0 (System.Object itemObject) <0x6f4db11018 + 0x00014> in <12cae72f24674302a4c964a96bed8f31>:0
  at System.Threading.ThreadHelper.ThreadStart_Context (System.Object state) <0x6f4f189a9c + 0x000a7> in <f377425eac5e4515ba2e6acd0fac10bf>:0
  at System.Threading.ExecutionContext.RunInternal (System.Threading.ExecutionContext executionContext, System.Threading.ContextCallback callback, System.Object state, System.Boolean preserveSyncCtx) <0x6f4f187c60 + 0x0019b> in <f377425eac5e4515ba2e6acd0fac10bf>:0
  at System.Threading.ExecutionContext.Run (System.Threading.ExecutionContext executionContext, System.Threading.ContextCallback callback, System.Object state) <0x6f4f187bec + 0x0002f> in <f377425eac5e4515ba2e6acd0fac10bf>:0
  at System.Threading.ThreadHelper.ThreadStart (System.Object obj) <0x6f4f189bb8 + 0x00047> in <f377425eac5e4515ba2e6acd0fac10bf>:0
```

I've made a guess and thought that the problem could be sourced or around the implementation of our HandlerScheduler for droid. More specifically, I think it could around our Rx bindings related to the main thread, since part of the package in the stacktrace points to `System.Reactive.Concurrency` and our SchedulerProvider only has custom implementations for the `MainScheduler`.

## How is it done?
I've added extra `try-catch`es around some of the statements in the `Schedule` implementations, more specifically, before the `post` of the handler and inside the post lambda before the `action` is invoked.

### Why not another way?
I've tried to make it less horrible, but couldn't. Feel free to suggest something.

### Side effects
There shouldn't be any.

## Review hints
Consider that this was done because we don't know for sure how fix our problem on #4230. This PR is only trying to help figuring out a solution by collecting more data.

## :squid: Permissions
🦑

## Note:
If this gets approved and merged, please create a follow-up to review the data and revert the changes done here after the crash is fixed.


daividssilverio:
After further investigation and fiddling with the Scheduler providers and I stumped over one exception that looked similar but not at all the same as the current tracked exception. All I did was, using the `TrackedSchedulerWrapper`, set the parameters of the `action` in the `Schedule` methods to null (at random) and checked the resulting exception's stack traces.

With that experiment, I think it's worth to track the other schedulers as well. I've added a wrapper that spies over the schedule methods, tracking any exceptions and forwarding them afterwards.

Toggl.Droid/Startup/AndroidSchedulerProvider.cs
```diff
@@ -10,11 +12,69 @@ public sealed class AndroidSchedulerProvider : ISchedulerProvider
...
+        public sealed class TrackedSchedulerWrapper : IScheduler
+        {
+            private readonly string analyticsEventName;
+            private readonly IAnalyticsService analyticsService;
+            private readonly IScheduler innerScheduler;
+
+            public TrackedSchedulerWrapper(IScheduler innerScheduler, string name, IAnalyticsService analyticsService)
+            {
+                this.innerScheduler = innerScheduler;
+                this.analyticsEventName = $"DROID_HANDLER_{name}_SCHEDULER_ERROR";
```
kornelijepetak: Our event names are `PascalCase`, not yelling-with-underscores-case.

kornelijepetak: Also, I'm not really sure how you inject the `{name}` here?
The only place you are using this as the first parameter of the `analyticsService.HandledException.Track` which seems to be 

```
[AnalyticsEvent("ExceptionType", "ExceptionMessage")]
public IAnalyticsEvent<string, string> HandledException { get; protected set; }
```

so the first parameter should be the `exception type` and in your case it's event name. And the second parameter should be the `exception message` and in your case it's a name of the failing method.

I'm confused what you are doing? If you are using existing methods by abusing just the count of the parameters to send two pieces of information, please don't. Rather create a new event type.

Toggl.Droid/Startup/AndroidSchedulerProvider.cs
```diff
@@ -10,11 +12,69 @@ public sealed class AndroidSchedulerProvider : ISchedulerProvider
...
 
-        public AndroidSchedulerProvider()
+        public AndroidSchedulerProvider(IAnalyticsService analyticsService)
         {
-            MainScheduler = new HandlerScheduler(new Handler(Looper.MainLooper), Looper.MainLooper.Thread.Id);
-            DefaultScheduler = Scheduler.Default;
-            BackgroundScheduler = NewThreadScheduler.Default;
+            MainScheduler = new HandlerScheduler(new Handler(Looper.MainLooper), Looper.MainLooper.Thread.Id, analyticsService);
+            DefaultScheduler = new TrackedSchedulerWrapper(Scheduler.Default, "DEFAULT", analyticsService);
+            BackgroundScheduler = new TrackedSchedulerWrapper(NewThreadScheduler.Default, "BACKGROUND", analyticsService);
```
juanlaube:
```suggestion
            BackgroundScheduler = new TrackedSchedulerWrapper(NewThreadScheduler.Default, "Background", analyticsService);
```

Toggl.Droid/Startup/AndroidSchedulerProvider.cs
```diff
@@ -10,11 +12,69 @@ public sealed class AndroidSchedulerProvider : ISchedulerProvider
...
         public IScheduler BackgroundScheduler { get; }
 
-        public AndroidSchedulerProvider()
+        public AndroidSchedulerProvider(IAnalyticsService analyticsService)
         {
-            MainScheduler = new HandlerScheduler(new Handler(Looper.MainLooper), Looper.MainLooper.Thread.Id);
-            DefaultScheduler = Scheduler.Default;
-            BackgroundScheduler = NewThreadScheduler.Default;
+            MainScheduler = new HandlerScheduler(new Handler(Looper.MainLooper), Looper.MainLooper.Thread.Id, analyticsService);
+            DefaultScheduler = new TrackedSchedulerWrapper(Scheduler.Default, "DEFAULT", analyticsService);
```
juanlaube:
```suggestion
            DefaultScheduler = new TrackedSchedulerWrapper(Scheduler.Default, "Default", analyticsService);
```

Toggl.Droid/Startup/AndroidSchedulerProvider.cs
juanlaube:
```suggestion
                this.analyticsEventName = $"DroidHandler{name}SchedulerError";
```

Toggl.Droid/Startup/HandlerScheduler.cs
```diff
@@ -2,37 +2,70 @@
...
 using System.Reactive.Disposables;
 using Android.OS;
+using Toggl.Core.Analytics;
+using Toggl.Droid.Extensions;
 
 namespace Toggl.Droid
 {
     public sealed class HandlerScheduler : IScheduler
     {
+        private const string analyticsEventName = "DROID_HANDLER_SCHEDULER_ERROR";
```
juanlaube:
```suggestion
        private const string analyticsEventName = "DroidHandlerSchedulerError";
```

Toggl.Droid/Startup/AndroidSchedulerProvider.cs
```diff
@@ -10,11 +12,69 @@ public sealed class AndroidSchedulerProvider : ISchedulerProvider
...
+
+            public IDisposable Schedule<TState>(TState state, Func<IScheduler, TState, IDisposable> action)
+            {
+                try
+                {
+                    return innerScheduler.Schedule(state, action);
+                }
+                catch (Exception exception)
+                {
+                    analyticsService.HandledException.Track(analyticsEventName, "Schedule:1");
```
juanlaube: I would add another custom event and track the exception name and stack trace as event parameters, not just here, but in every event you added.

kornelijepetak: @juanlaube If you follow what the following line `(:46)` does, you will see that the crash reporting will trace the whole exception (including the stack trace).

Toggl.Droid/Startup/AndroidSchedulerProvider.cs
daividssilverio: > Also, I'm not really sure how you inject the {name} here?

From the constructor. I build the `analyticsEventName` by formatting the "... {name} ..." string.

> The only place you are using this as the first parameter of the analyticsService.HandledException.Track which seems to be ...

I'm not using the name directly on `HandledException.Track`, I'm using the `analyticsEventName` I've built in the constructor.

> so the first parameter should be the exception type and in your case it's event name. And the second parameter should be the exception message and in your case it's a name of the failing method.

What you said doesn't match the source code. `HandledException` takes two strings; it doesn't have to be event type and the exception message. I've used it this way because I'm already tracking the exception and it's stacktrace.

```
analyticsService.HandledException.Track(analyticsEventName, "Schedule:2");
analyticsService.Track(exception, exception.Message);
```

>I'm confused what you are doing? If you are using existing methods by abusing just the count of the parameters to send two pieces of information, please don't. Rather create a new event type.

I see your point. I'd rather not go round creating events just for debug tracking because I'm always weary of method counting, but since both of you @juanlaube and @kornelijepetak think it's better to create another custom event, I'll do it.

Toggl.Droid/Startup/AndroidSchedulerProvider.cs
daividssilverio: I'll just do that.

PR #4966, opened 2019-04-23
Improve project and tag creation cell unit tests

### LINKED ISSUE #4846: Add unit test(s) for project and tag creation cells
We need tests for the behaviour that was fixed in #4844 and #4680, so we don't accidentally break it a third time.

arturs105:
## What's this?
Turns out we already had tests that check that project and tag creation cells are included. I just improved the assertions a bit to:
* make sure that the entity name in the creation cell is correct
* make sure that the creation cell is the first item of the section

### Relationships

Closes #4846

## Why do we want this?
Because we had to fix a single bug two times (mentioned in the issue). These tests now fail, when that bug is present.

## :squid: Permissions
🦑


Toggl.Core.Tests/UI/ViewModels/StartTimeEntryViewModelTests.cs
```diff
@@ -538,6 +547,16 @@ public async Task TracksTagSelection()
 
                     AnalyticsService.StartEntrySelectTag.Received().Track(ProjectTagSuggestionSource.TextField);
                 }
+
+                [Fact, LogIfTooSlow]
+                public async Task IncludesTagCreationCellWhenThereIsNoMatchingTag()
+                {
```
kornelijepetak: What up with these tests?
I mean, the test pass, but should they? :)

arturs105: 🤦‍♂️ I started this PR by adding empty test cases with the intention of implementing them later. Then i discovered that these cases already had tests and i just updated those, forgetting about the empty tests that i added.

Toggl.Core.Tests/UI/ViewModels/StartTimeEntryViewModelTests.cs
```diff
@@ -318,8 +319,13 @@ private string createLongString(int length)
 
             private bool hasCreateSuggestions(SectionModel<string, AutocompleteSuggestion> collection)
```
kornelijepetak: Hey, since you've changed the behavior of `hasCreateSuggestions`, can you please rename it so that it's name actually says what the function returns, such as
`isFirstItemCreateSuggestion`.

PR #4967, opened 2019-04-23
Fix bug and add diagnostics

### LINKED ISSUE #4926: NullReferenceException in EditTimeEntryActivity.setupViews
## 💥 Crash Report
This supposedly happens to a significant number of users on 1.13, when they open the Edit View.

### 👥 Impact
Current rollout: 5% (halted)
By now, occured `65` times to `43` users. 

### 🔗 Context
In 1.13, the Edit View has been rxfactored. This error has not been seen on any previous build (the excepting method hasn't existed before 1.13).

### 🛰 Link
[AppCenter details.](https://appcenter.ms/orgs/Toggl/apps/Toggl-Android/crashes/errors/1586665989u/overview)

Unfortunatelly, this is all there is:

```cs
EditTimeEntryActivity.setupViews ()
EditTimeEntryActivity.OnCreate (Android.OS.Bundle bundle)
Activity.n_OnCreate_Landroid_os_Bundle_ (System.IntPtr jnienv, System.IntPtr native__this, System.IntPtr native_savedInstanceState)
```

kornelijepetak:
## What's this?
This could probably help with failing Edit View (setupViews bug #4926).
Additionally, it another tracking event to help see what's going on.

### Relationships
Closes #4926

## Why do we want this?
Because we want users to have a stable app, not 💥 

## How is it done?
We have a strong belief that the problem is that the value of the time entries' IDs are not reset when the application is resumed after tombstoning. So, we are now saving the values we need into a bundle.

### Why not another way?
If you know better, you do it! 😄 

### Side effects
Hopefully none, but needs testing.

## Review hints
Shouldn't be hard.

## :squid: Permissions
Only me and @daividssilverio .


Toggl.Core.UI/ViewModels/EditTimeEntryViewModel.cs
```diff
@@ -238,6 +239,25 @@ private IEnumerable<long> tagIds
             Delete = actionFactory.FromAsync(delete);
         }
 
+        protected override void ReloadFromBundle(IMvxBundle state)
+        {
+            base.ReloadFromBundle(state);
+
+            var ids = state.Data[nameof(TimeEntryIds)];
```
juanlaube: I think this would be more readable by doing `state.Data.TryGetValue(nameof(TimeEntryIds), out var ids) { ...` 🤔

Toggl.Core.UI/ViewModels/EditTimeEntryViewModel.cs
```diff
@@ -238,6 +239,25 @@ private IEnumerable<long> tagIds
...
+
+            if (ids == null)
+                return;
+
+            TimeEntryIds = ids.Split(',').Select(long.Parse).ToArray();
+        }
+
+        protected override void SaveStateToBundle(IMvxBundle bundle)
+        {
+            base.SaveStateToBundle(bundle);
```
juanlaube: Do you think tracking this and reload would be useful somehow? If this fix the bug I would like to know at least 🙂

Toggl.Core.UI/ViewModels/EditTimeEntryViewModel.cs
kornelijepetak: Your code is syntactically incorrect... 🤔 
What did you want to write?

But I never find TryGetValue more readable than `[]` syntax with a null check.

Toggl.Core.UI/ViewModels/EditTimeEntryViewModel.cs
kornelijepetak: What do you mean. Tracking what exactly?

You mean tracking this regardless of whether it crashes?

kornelijepetak:
Closed as just a WIP of #4968

PR #4979, opened 2019-04-24
🐛 Fix hidden Create Project item when filtering

### LINKED ISSUE #4822: Project Creation Cell behaves weirdly on Android
when user has many projects, the creation cell is shown at the very top, out of the scroll view's bound, which requires the users to manually scroll up to see it.

See the video and pay attention to the right scroll bar
https://khoi-screenshots.sgp1.digitaloceanspaces.com/Untitled.mov

## Context
Giskard latest `develop`

kornelijepetak:
## What's this?
Improves the experience of filtering projects so that the `Create Project XYZ` label is visible when filter changes.

### Relationships
Closes #4822 

## Why do we want this?
It makes more sense than the current behavior.

## How is it done?
Scrolling to top in recycler view after the items are updated.

### Why not another way?
I know not of other ways.

## Review hints
It's rather simple. You need no help there.

## :squid: Permissions
@toggl/androids


kornelijepetak:
@juanlaube In [this comment](https://github.com/toggl/mobileapp/issues/4822#issuecomment-486990846), there's an explanation for this conundrum. For now, we are going with this solution until we switch to plan B and extract this thing out of the table.

heytherewill:
I really don't like this. Is there no point of extension in the RecyclerView itself? Do you really need to add an observable in shared code?

kornelijepetak:
During my vacations, I hereby grant permission to all @toggl/androids to continue with this PR as they see fit.

PR #4985, opened 2019-04-25
Improve swipe actions threshold

### LINKED ISSUE #4805: Make Android swipe to delete/continue actions more robust
Android swipe to delete/continue time entry actions are responsive even to the slightest swipes, making people accidentally delete or continue their time entries while they simply wanted to scroll their log up or down. 

The same feature on iOS is nice. Slight swipe makes the continue/delete button appear, after which the button can be tapped to proceed with the action. In order to continue/delete an entry purely by swiping, the swipe gesture needs to be made over at least half the screen. 

Let's make the swipe to continue/delete action a bit more robust for Android as well in order to prevent users from accidentally losing their data by unwantedly deleting it.

juanlaube:
## What's this?
This is an attempt to improve the swipe actions in the main log so users do not delete or continue time entries accidentally.

### Relationships

Closes #4805 

## Why do we want this?
We don't want angry users yelling at us because they deleted a time entry by accident.

## How is it done?
Overriding some methods in the item callback helper.
I changed the parameters until it was difficult to delete a time entry "by accident":
- `GetSwipeVelocityThreshold` controls the maximum velocity for the swipe action.
- `GetSwipeEscapeVelocity` controls the minimum velocity for the gesture to be considered a swipe action.

### Why not another way?
🤷‍♂️ 

### Side effects
None

## Review hints
Please do a functional testing, if you still think some values need to be improved let me know.

## :squid: Permissions
🚫


heytherewill:
I think this should feel as heavy as Inbox once felt (or if we wanna play safe, Gmail)

kornelijepetak:
My review can be dismissed in my absence if at least one more dev confirms that this solution works ok for them as well.

juanlaube:
@kornelijepetak I came to the initial values by experimenting with it but also trying to not change the default value too much.

juanlaube:
I considered your suggestions and adjusted the parameters even more.
Also I introduced a new parameter, `GetSwipeThreshold` controls the percentage that should be opened to trigger the action according to the docs, however based on my tests, it just takes into account the position of the last touch.

daividssilverio:
There's something else we could add to help avoiding accidental swipes: We could change the `GetSwipeDirs` to not let the user swipe when the RecyclerView is not idle. We'd just have to check the RecyclerView.ScrollState to do that.

PR #4989, opened 2019-04-25
Fix freezes on the wheel caused by poor memory allocation

### LINKED ISSUE #4823: App crashes/freezes when editing duration in start view
## 🐛 Bug Report
The app becomes unresponsive, when wheel view is opened from the start time entry view.

## Context

Phone OS: Android 8.0
Phone model: Nexus 5X emulator

## Expected behavior
The app is responsive and does not crash.

## To reproduce
1. Start a time entry
2. Press on the duration label in the start view
3. The wheel view opens and becomes unresponsive after about a second
4. Android shows the app not responding dialog
5. 💥 

## Suggestions
This actually doesn't seem like a crash, since i don't get any exception breakpoints while debugging, it just stops working.

daividssilverio:
## What's this?
This fixes freezes on the wheel caused by poor memory allocation.
Thing is, as @kornelijepetak was investigating and commented here https://github.com/toggl/mobileapp/issues/4823#issuecomment-486228070, I also investigated in that front and found that the Wheel was creating too many instances of `Android.Graphics.PointF` because we were simply instantiating them a lot; specially when accessing the EditDuration view on a running time entry (Either new or existing). We were also updating the wheel way more than it was needed.

### Relationships
Closes #4823

## Why do we want this?
Because the wheel is cool and our users deserve to use it crash-less/freeze-less.

## How is it done?
It replaces usages of `toPointF` that created new instances of `PointF` all the time with a combination of single instances and methods with `ref` parameters.
Also, it reduced the number of calls to the setters of `StartTime` and `EndTime`, reducing even more the number of updates in the wheel, instantiations, work on the main thread, all that.

### Why not another way?
This way required less changes. Feel free to request changes.

### Side effects
None that my eyes caught.
I've made an unrelated change because the app wouldn't stop crashing because of the usage of `Mvx.Resolve` in the `/Toggl.Droid/Activities/EditTimeEntryActivity.cs`, which shouldn't work anymore since we are now using the new `DependencyContainer`.

## Review hints
Try to replicate the crash/freeze on another branch; Try here and let me know if it didn't work for you. 

## :squid: Permissions
@juanlaube only 🦑


Toggl.Droid/Activities/EditTimeEntryActivity.cs
```diff
@@ -63,8 +58,9 @@ protected override void OnCreate(Bundle bundle)
             }
             catch (Exception exception)
             {
-                analyticsService.Track(exception, $"{nameof(setupViews)} failed to execute.");
-                analyticsService.DebugEditViewInitialSetup.Track(hasViewModel, hasTimeEntries, timeEntriesCount, rehydrationCount);
+                analyticsService = AndroidDependencyContainer.Instance.AnalyticsService;
+                analyticsService?.Track(exception, $"{nameof(setupViews)} failed to execute.");
```
kornelijepetak: I don't think you should use the `?` here.
First of all, if `AndroidDependencyContainer.Instance.AnalyticsService` is not able to provide the service, the app should really not continue. Core container items should always be available (maybe some platform things don't have to, but these ones we really REQUIRE).
So, if it is `null`, I'd rather the app crash at that point than just silently _not track anything_. 

Because then we'll have some problems which would be swallowed, the app would continue and no events would be reported to firebase. The end result: we have problems we don't know about.

Toggl.Droid/Extensions/MotionEventExtensions.cs
```diff
@@ -7,5 +7,10 @@ public static class MotionEventExtensions
     {
         public static PointF ToPointF(this MotionEvent motionEvent)
             => new PointF(motionEvent.GetX(), motionEvent.GetY());
+
+        public static void ToPointF(this MotionEvent motionEvent, ref PointF outPointF)
+        {
+            outPointF?.Set(motionEvent.GetX(), motionEvent.GetY());
```
kornelijepetak: Another case of undesired behavior of `?`.
If `outPointF` is `null`, something is terribly wrong.

Toggl.Droid/Extensions/PointExtensions.cs
```diff
@@ -5,8 +5,10 @@ namespace Toggl.Droid.Extensions
 {
     public static class PointExtensions
     {
-        public static PointF ToPointF(this TogglPoint point)
-            => new PointF((float)point.X, (float)point.Y);
+        public static void ToPointF(this TogglPoint point, ref PointF outPointF)
+        {
+            outPointF?.Set((float)point.X, (float)point.Y);
```
kornelijepetak: Same about `?`.

Toggl.Droid/Views/EditDuration/Shapes/Dot.cs
```diff
@@ -39,7 +39,7 @@ public void Update(double startTimeAngle, double endTimeAngle)
             var diffAngle = endAngle - startAngle + (endAngle < startAngle ? FullCircle : 0);
             var diffInDegrees = (float) Math.ToDegrees(diffAngle);
             hidden = diffInDegrees < visibilityThresholdInDegrees;
-            position = PointOnCircumference(pivotCenter, startAngle + diffAngle / 2f, distanceToPivot).ToPointF();
+            PointOnCircumference(pivotCenter, startAngle + diffAngle / 2f, distanceToPivot).ToPointF(ref position);
```
kornelijepetak: It is not necessary to use `ref` parameter here.
Two reasons:
1) You are not really changing the variable itself in `ToPoint`, so you don't need a reference to a variable
2) `PointF` is a `class`, not a `struct`, so it is passed by reference.

That being said, `ToPoint` is an invalid name because this is not creating a new point. `ToSomething` name is by convention in .NET used for conversions from A to B, such as `.ToList()` etc.

Since your new overload of `ToPoint` is actually just updating the given point with new coordinates, please rename both methods to `Update` and ditch the `ref`. But then the `Update` would mean that you are updating the left side with the argument, which is a wrong lead, so I would propose a switch in the arguments (for that new overload only).

With all my comments taken into account, I suggest it like this:
```cs
public static void Update(this PointF updatingPoint, TogglPoint point)
{
    updatingPoint.Set((float)point.X, (float)point.Y);
}
```

and then

```cs
position.Update(PointOnCircumference(pivotCenter, startAngle + diffAngle / 2f, distanceToPivot));
```

Similarly, do the same with motion event
```cs
public static void Update(this PointF updatingPoint, MotionEvent motionEvent)
{
    updatingPoint.Set(motionEvent.GetX(), motionEvent.GetY());
}
```

Alternatively, it could be `UpdateWith` or `UpdateFrom` to mark it even further, but I don't think it's strictly necessary.

Toggl.Droid/Activities/EditTimeEntryActivity.cs
kornelijepetak: Just so you know, I've fixed the same thing in `develop` because you have branched from it from a bad commit.

It's at #4990, for reference.

kornelijepetak:
Just so you know, I've fixed the analytics MVX bug as #4990 in `develop`.

This release branch was branched from develop from a bad commit, so I had to fix it there. If you'd rather revert this change for this PR and then do a 🍒, please do.
If not, just make sure you remove the `?` and it will just disappear during merge back to develop because the changes should be the same.

This code all together should be 🔥 in 1.15.

Toggl.Droid/Activities/EditTimeEntryActivity.cs
juanlaube: I agree with @kornelijepetak, if analyticsService is null, I want the app to crash immediately because somethings is terribly wrong.

kornelijepetak:
I hereby give permission to @juanlaube to dismiss my review if he can confirm that my requests were heeded to :)

Toggl.Droid/Activities/EditTimeEntryActivity.cs
daividssilverio: I get your point guys, but the app will crash anyway, being the analyticsService null or not.
The exception is being re-thrown after tracking. I just didn't want to cause a crash because of our IoC not working properly. I got a crash with the AnalyticsService being null during debug, but fixing it would take more time, since I don't know how it happened.

Toggl.Droid/Extensions/MotionEventExtensions.cs
daividssilverio: Ok. I'll fix this from the usages.
But be aware that we are mixing mono with the droid "jvm", `PointF` is managed by droid. And we've already seen stuff being null where it shouldn't.

Toggl.Droid/Views/EditDuration/Shapes/Dot.cs
daividssilverio: Very good points. I'll take your suggestions right away.

Toggl.Droid/Activities/EditTimeEntryActivity.cs
kornelijepetak: The point of exception is to break at the point of the problem, not down the line. Analytics service not being available is an exceptional case of program flow. We should throw as soon as we encounter such a case.

PR #4993, opened 2019-04-26
🐛 Fix crashing in main log

### LINKED ISSUE #4992: NullReferenceException in MainTabBarActivity.getCachedFragment
## 💥 Crash Report
Main log crash, most likely when returning after being tombstoned.

### 👥 Impact
* `17 times` to `16 users` (1.13.4)
* `131 times` to `112 users` (1.12)
* Introduced when switching to bottom navbar.

### 🛰 Link
https://appcenter.ms/orgs/Toggl/apps/Toggl-Android/crashes/errors/1534276311u/overview

```cs
MainTabBarActivity.getCachedFragment (System.Int32 itemId)
MainTabBarActivity.showInitialFragment ()
MainTabBarActivity.OnCreate (Android.OS.Bundle savedInstanceState)
Activity.n_OnCreate_Landroid_os_Bundle_ (System.IntPtr jnienv, System.IntPtr native__this, System.IntPtr native_savedInstanceState)
```

kornelijepetak:
## What's this?
I believe this fixes the crash in main log after returning from a tombstoned app.

### Relationships
Closes #4992

## Why do we want this?
No 💥

## How is it done?
Well, I believe the problem is that by the moment `MainTabBarActivity` is trying to fetch `ViewModel.Tabs` in `getCachedFragment`, the `Tabs` property is still `null` as it is assigned asynchronously in Initialize.

I have moved it to the constructor so that as long as a ViewModel exists (and its constructor has executed), the `Tabs` will be available as well, causing no crashes in the activity.

### Why not another way?
I think this will be enough.

## :squid: Permissions
Anyone who believes in this PR.


Toggl.Core.UI/ViewModels/MainTabBarViewModel.cs
```diff
@@ -152,14 +152,14 @@ public sealed class MainTabBarViewModel : ViewModel
                 rxActionFactory,
                 permissionsService,
                 schedulerProvider);
+
+            Tabs = getViewModels().ToList();
```
heytherewill: While at it, remove the setter from the `Tabs` property

kornelijepetak: Done.

PR #4995, opened 2019-04-26
Fix reports "this week" label

### LINKED ISSUE #4873: Clocked Hour Chart doesn't reload first day of week settings
## 🐛 Bug Report
When the first day of the week setting updated, the bar chart doesn't get updated. It's only updated when users relaunch the app.

## Context

Build/Branch: `develop`
Phone OS: 12.2
Phone model: iOS

## Expected behavior
The bar chart should reflect the new setting change.

## Suggestions
This could be happening on Android as well, I haven't checked.

arturs105:
## What's this?
The current period label in the reports navigation bar had a problem, where the "this week" label would only show up with Monday selected as beginning of week. Check my comment in the issue below for more info.

### Relationships

Closes #4873 

## Why do we want this?
Gotta respect all options for beginning of week (even the wrong ones).

## :squid: Permissions
🦑


arturs105:
If you test this out on Android, you might find that it does not seem to be working. However, that is a separate problem and there's an issue for that already: #4996.

Everything should work fine on iOS.

Toggl.Shared/Extensions/DateTimeOffsetExtensions.cs
```diff
@@ -68,5 +68,11 @@ public static DateTimeOffset RoundUpToClosestQuarter(this DateTimeOffset time)
                 return new DateTimeOffset(time.Year, time.Month, time.Day, time.Hour, minute, 0, time.Offset);
             }
         }
+
+        public static DateTimeOffset BeginningOfWeek(this DateTimeOffset time, BeginningOfWeek beginningOfWeek)
+        {
+            int diff = (7 + ((BeginningOfWeek)time.DayOfWeek - beginningOfWeek)) % 7;
```
daividssilverio: Could add a UnitTest for this?

daividssilverio: This method*

PR #4997, opened 2019-04-26
:satellite: Track low memory warnings

heytherewill:
## What's this?
This is used to track low memory warnings.

## Why do we want this?
We have been having weird android crashes due to heavy memory usage. This can help us diagnose what's causing memory issues on real-world usage of the app so we can act proactively instead of reactively

## How is it done?
Both platforms have a way to notify the app that we are using too much memory. Ideally, we would be able to free resources or do something to free memory. We are, instead, simply tracking that memory is low so we can have a better idea on how to avoid this scenario altogether.

If you think the event could have a better name, feel free to point it out.

### Side effects
None

## :squid: Permissions
Anyone, after both @toggl/androids and @toggl/ios-senseis have approved


daividssilverio:
I like the idea, and it would be great if we could correlate this events with `CurrentPage`, sync events, `AppSentToBackground` and  `AppDidEnterForeground` events to help understand how the user got to that state of low memory.

heytherewill:
We get this for free in AppCenter! Whenever we see a spike in Low memory events we can easily recreate the steps the user took in-app to reproduce the memory issues :)

Toggl.Droid/Startup/TogglApplication.cs
```diff
@@ -25,5 +26,15 @@ public override void OnCreate()
                 typeof(Microsoft.AppCenter.Analytics.Analytics));
 #endif
         }
+
+        public override void OnLowMemory()
+        {
+            base.OnLowMemory();
```
daividssilverio: Docs say we should use `OnTrimMemory(int)` instead. `OnLowMemory` seems to be there to support older android versions.

https://developer.android.com/reference/android/app/Application.html#onLowMemory()

dshr:
I'm getting some weird behaviour on iOS — triggering a low memory warning always closes the app when `AppDelegate::ReceiveMemoryWarning` is overridden. Is there a chance that iOS is expecting us to drastically lower memory usage, and, when it sees it stay the same, decides to kill the app?

heytherewill:
Oh yeah, when you override methods that are optional in Obj-C, calling base throws [You_Should_Not_Call_base_In_This_Method](https://github.com/xamarin/xamarin-macios/blob/master/src/Foundation/ModelNotImplementedException.cs#L30). Removing the call to base fixes it, please try again :)

daividssilverio:
After some research and testing I found out that these callbacks on Android (both onTrimMemory and on onLowMemory) aren't really useful for diagnosing memory allocation problems.

We can surely use it to release resources when the system asks for it, but during the lifetime of the application, it doesn't make much of a difference.

We could build our own monitoring library, but I'm not sure if it's worth it or not.

Other option is to follow steps https://developer.android.com/studio/profile/memory-profiler case-by-case based on support reports and whatever we see that's sluggish or crashing.

I'm removing my previous request for changes and I this can be merged without harm, but it's not really useful for droid at the moment.

simonrozsival:
BTW guys, is it possible that Android allows much less memory for background jobs? Could the background syncing errors we were having be caused by Android freeing some parts of memory we rely on?

daividssilverio:
Answering your question @simonrozsival: 

I never heard about anything of the sort for android. Researching a bit further into the framework, it could be the case that whatever is started at the `BackgroundSyncJobSchedulerService` isn't actually running in the background.

The framework wouldn't reclaim memory we are currently holding references without notice, if it does need to do so, it would go through the callbacks introduced in this PR.

What could be happening is that we are losing a network connection, `OnStopJob` is being called, the device could start sleeping and we are still trying to run the background sync, which of course could misbehave.

PR #4998, opened 2019-04-26
:construction: Add basic presenter infra

### LINKED ISSUE #4852: Create basic presenter infrastructure
Our navigation service will be as dumb as it gets. It will simply be a thin layer that can be injected into ViewModels (similar to what the MvvmCross service is, but since our implementation will be opinionated towards Toggl and its needs, we will not need a lot of bloat).

The real work will happen inside the presenters, which will be platform specific classes that handle all the quirks of displaying Views to users. This issue is meant to create the barebones of the presenter infra, which are:

- [x] Create a `NavigationInfo<T>` class, which will contain the payload used to pass information between ViewModels
- [x] Create an `IPresenter` interface, which will be injected in our navigation service
  - [x] It will simply have a `Task<TOutput> Present<TInput, TOutput>(NavigationInfo<TInput> navigationInfo)` method
- [x] Create abstract classes for iOS and android that implement `IPresenter`. Each of these classes will be responsible for presenting a specific type of View (like what we currently do with attributes)
  - [x] The classes must contain as much basic information needed to display views as possible (e.g.: `Context` property on Android and a `Window` property on iOS. We can base this on MvvmCross)
  - [x] The classes must also have an abstract method `bool CanPresent(NavigationInfo<T> navigationInfo)`. Since we are gonna have a bunch of presenters, the composite presenter needs to know to which presenter if should defer creation.
- [x] Create a composite for each platform that takes a bunch of `BasePresenter`s and combines then so that each navigation request gets forwarded to the correct place

heytherewill:
## What's this?
This adds the basic presenter infra that will be needed further down the line

### Relationships
Closes #4852

## Why do we want this?
Good ole MVExit

## How is it done?
I opted to have the composite presenter in a shared layer. It just felt right.
iOS really needs a hold of both Window and AppDelegate
Android doesn't need anything atm

### Why not another way?
Opinions welcome

### Side effects
Literally none; this code is not executed yet

## :squid: Permissions
Myself


Toggl.Core.UI/Navigation/IPresenter.cs
```diff
@@ -0,0 +1,36 @@
+﻿using System;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Toggl.Core.UI.Navigation
+{
+    public interface IPresenter
+    {
+        bool CanPresent<TInput>(NavigationInfo<TInput> navigationInfo);
```
daividssilverio: Just curious, what's this for?

heytherewill: Different presenters can show different types of ViewModels. We first ask if the presenter can handle a request before actually making it handle it.

Issue #4359, opened 2019-02-07
Siri. Create entries with description from clipboard


Odrakir:
In order to have a workflow for users to dictate the entry descriptions to Siri we need first to add an intent that will create an entry from whatever is in the clipboard.

After that's done we just make a custom workflow that fires up dictation, takes whatever the user said and copies it to the clipboard and then runs that intent.

PR #5001, opened 2019-04-29
🐛 fix TE with project not getting donated

khoi:
## What's this?
Fix a bug where TE with project associated not getting donated.

### Relationships

Introduced by #4359

## To reproduce

- Start a TE with a synced project attached
- The donation doesn't go through because of

```cs
if (!(timeEntry.ProjectId is long projectId))	
    return;
}
```

## Expected behavior

- The intent donation should work properly.

## Review hints
None.

## :squid: Permissions
🎙


Odrakir:
I'm testing this... That check was there so entries for new projects are not donated, because we don't have the real (server) project ID yet, so we ignore those donations. Entries with project that were already synced before work as expected.

As far as I can tell the behaviour is correct and this doesn't need fixing. Am I missing something?

khoi:
oops, the code was

```
if (!(timeEntry.ProjectId is long projectId)) {
   return
}
```

in cases where projectId is `null`, the function exits where it shouldn't be. 

This PR reverts the checking logic and add logic for building the `StartTimerIntent`'s `ProjectId`

Odrakir:
I see, so this is a fix for entries without project. Understood.

Toggl.iOS/Services/IntentDonationServiceIos.cs
```diff
@@ -58,20 +58,22 @@ public async Task DonateStartTimeEntry(IWorkspace workspace, ITimeEntry timeEntr
...
                 {
                     return;
                 }
 
-                if (projectId < 0 || timeEntry.TagIds.Any(tagId => tagId < 0))
+                if (timeEntry.ProjectId is long projectId)
                 {
-                    return;
+                    var project = await dataSource.Projects.GetById(projectId).FirstAsync();
+                    startTimerIntent.ProjectId = startTimerWithClipboardIntent.ProjectId = new INObject(projectId.ToString(), project.Name);
```
Odrakir: I didn't realise about this before, but I'm not a big fan of multiple assignments in the same line. This is much harder to read than two lines.

Toggl.iOS/Services/IntentDonationServiceIos.cs
```diff
@@ -58,20 +58,22 @@ public async Task DonateStartTimeEntry(IWorkspace workspace, ITimeEntry timeEntr
...
+                }
+                else
+                {
+                    startTimerIntent.ProjectId = startTimerWithClipboardIntent.ProjectId = new INObject(timeEntry.ProjectId.ToString(), String.Empty);
```
Odrakir: Isn't `timeEntry.ProjectId` null here? Do we need this line at all?

Issue #4862, opened 2019-03-29
Implement our own navigation service


heytherewill:
Depends on #4851
Depends on #4863

This one is surprisingly straight forward, since the heavy labor is all divided among the various presenters. This should simply be constructed using a presenter, create a thin presentation object and pass it so the presenter can do its magic. It should also abstract away the handling of `Unit` in ViewModels that do not receive/return data.

heytherewill:
Closed by #5011

PR #5002, opened 2019-04-29
:milky_way: Implement the new Navigation service

heytherewill:
## What's this?
This implements the new navigation service

### Relationships
Ref #4862 (I will only close it in another PR, when we actually use this code)

## How is it done?
I am basically delegating the display of the VM to the Presenter while storing a reference to a TaskCompletionSource in the VM itself, this way the method will only finish when the VM is closed. Simple and effective.

## Review hints
You can check the branch `custom-navigation` to see what this code will look like in use, but for now it's simple enough since it isn't used anywhere

## :squid: Permissions
Only myself


daividssilverio:
Oops, I think your target branch might be wrong, was it supposed to be develop?

PR #5006, opened 2019-04-30
Siri Tapping the response UI opens edit view

### LINKED ISSUE #4559: Siri: Tapping the response UI should open edit view
When starting or stopping time entries through Siri, tapping the response UI opens the app.

Ideally we would open the edit view for that entry. This probably needs syncing first, so it might not be to easy, and maybe it takes too long for the user to wait for it to show up. Still we should try to see if it makes sense.

khoi:
### Relationships
Closes #4559

## How is it done?
- When the (Start|Stop)TimerHandler execute succesfully, the TE returned from the server will JSON serialized and saved to SharedStorage
- When the app opens, the JSON is parsed back to a `TE` object, and the first time the Sync's status switch from `Syncing` -> `Synced`, the Edit view for that TE will be opened.

### Why not another way?
It's been suggested that the TE could be saved to Realm directly so we don't have to wait for syncing to be completed. But I couldn't figure out how to do that at the moment.

### Side effects
Should be none since this is new behavior.

## Review hints
None beside functional testing.

**Question**: Do you think that we should have a time out when waiting for the app to be "synced"? Let's say that after 5 seconds from openning, if the app doesn't sync, we don't display the Edit view at all.

## :squid: Permissions
🎙


Odrakir:
I'm not sure if we should merge this. In my opinion the delay makes it very confusing for the user. Let's freeze it for now and discuss if we can use this or if there's a way to implement it how it was intended in the first place.

I'm attaching a GIF for reference, this is a good scenario, but sync can take longer. Also... Why is the syncing indicator not showing?

edit_te

heytherewill:
Any news on this?

amulware:
If we still want this, please redo.

Issue #4871, opened 2019-03-29
Figure out and implement a new TopActivity service


heytherewill:
The solution can probably be largely dependent on the existing implementation that we are using over MvvmCross' implementation. We need to think this through, though, since it is a very important part of the application.

Also, we need to scan the app for classes that inherit from `MvxTask` and make sure that the new system will allow those classes to work.

daividssilverio:
To completely remove the need for the TopActivity, we'll need to complete the following:
- [x] Replace the usages of IDialogService with the IView implementations on Droid
- [ ] Replace the usages of IDialogService with the IView implementations on iOS
- [ ] Replace the usages of IPermissionsService with the IView implementations on Droid
- [ ] Replace the usages of IPermissionsService with the IView implementations on iOS
- [ ] Replace the `TogglPresenter` mechanism to close Views.

heytherewill:
Further, we need to:

- [ ] Get rid of the Google Service on iOS
- [ ] Get rid of the Google Service on Android
- [ ] Remove the unused service for good: `IDialogService`, `IPermissionsService` and `IGoogleService`

PR #5007, opened 2019-04-30
:sparkles: Introduce attachable concept to new ViewModels and replace start replacing IDialogService on Droid

daividssilverio:
## What's this?
This is part of the Mvexit project, this PR has part of the changes necessary to remove the usage of `TopActivity` on droid. It also introduces changes that will change how we'll work with code that touches platform specific (ui related) code.

### Relationships
Closes nothing, part of #4871

## Why do we want this?
To remove the need to keep hard references of views lying around somewhere.
The attach/detach (if called properly) is what actually replaces the need of the `TopActivity` we currently have on Droid.

## How is it done?
Check the changes, this PR does four things
- Introduces the new `IView` and `IViewModel` combo.
- Makes the ReactiveActivity implement the new `IView` and call the `IViewModel` Attach and Detach callbacks `onCreateView` and `onDestroyView` (temporary?)
- Introduces an extension that looks for a View in the view models, if finds it, uses it instead of the IDialogService injected from the constructor
- Makes the current implementation of the IDialogService for droid throw if any method is called.

### Why not another way?

### Side effects
iOS will have to be implemented as well.

## Review hints
Test it, there shouldn't be any visual changes.

## :squid: Permissions
@heytherewill 🦑 it


daividssilverio:
>I like this. When implementing iOS, however, please remove all references to the DialogService and get rid of any extension methods :)

That's the idea!

PR #5010, opened 2019-05-01
Siri: Refactor intent donation service

Odrakir:
## What's this?
This refactors the intent donation service so it's iOS only. It implies it can't be used from Core anymore, so there are some changes that needed to be done.

Also we were donating intents from interactors most of the time, but those sometimes get called not on response to a user action, I removed that too.

Lastly I'm removing some unneeded code in `IntentDonationService` because we don't really need to donate the clipboard shortcut every time the user starts an entry.

### Relationships
Doesn't close any issue.

## Why do we want this?
We need it so now the service can be changed to return iOS native objects, which are needed to make the shortcuts listing easier (or possible at all).

## Review hints
I tried to make it as tidy as possible, just follow each commit.

## :squid: Permissions
@khoi or me


Toggl.iOS/Services/IntentDonationServiceIos.cs
```diff
@@ -51,9 +50,7 @@ public async Task DonateStartTimeEntry(IWorkspace workspace, ITimeEntry timeEntr
             var relevantShortcuts = new List<INRelevantShortcut>();
 
             var startTimerIntent = new StartTimerIntent();
-            var startTimerWithClipboardIntent = new StartTimerFromClipboardIntent();
```
khoi: the code related to start timer with clipboard is now missing

Odrakir: Yes, I found this while refactoring that class. Why are we donating that one every time the user starts an entry? I think it's enough having it as a default shortcut. Am I missing something?

khoi: I had to do it since we want tags, projects to be attached to the intent

PR #5011, opened 2019-05-01
Unify ViewModel lifecycles

### LINKED ISSUE #4862: Implement our own navigation service

### LINKED ISSUE #4863: Unify ViewModel Lifecycle
Depends on #4850

Right now, this is a mess. We have `Init`, `Prepare`, `Initialize` and `Start`. Our custom ViewModel type should have a single method:
`public void Initialize(TInput input)`

It has a void return type to force async work to be done using Observables.

heytherewill:
## What's this?
This unifies the ViewModel lifecycles from the many methods that MvvmCross calls to a single unified `Initialize(TInput input)` method

### Relationships
Closes #4862
Closes #4863
Ref #4871

## How is it done?
I extracted all used view lifecycle methods into the IViewModel interface and then made our base viewmodel implement them. This PR is big because to do that I had to completely remove MvvmCross from the startup.

While it might sound like a lot, that is not the case. I am literally mimicking the MvvmCross way of starting up the apps; I just removed the Mvx named classes. The new startup will come in a later PR.

### Why not another way?
We would have to make this swap one time or the other; better sooner than later

### Side effects
The app is effectively broken at the moment. While everything builds, once the app loads it tries to present an Activity/ViewController, but those were not implemented, so the app simply throws. Bear in mind that this is temporary and this PR is to be merged before the presenters because then we can test them properly.

## Review hints
The commit history is pretty clean and you can easily review commit by commit. Tests are broken, but I will fix them as reviews come.

## :squid: Permissions
Only myself


Toggl.iOS/Startup/AppDelegate.Analytics.cs
```diff
@@ -0,0 +1,32 @@
...
+                typeof(Microsoft.AppCenter.Analytics.Analytics));
+            Firebase.Core.App.Configure();
+            Google.SignIn.SignIn.SharedInstance.ClientID =
+                Firebase.Core.App.DefaultInstance.Options.ClientId;
+            Adjust.AppDidLaunch(ADJConfig.ConfigWithAppToken("{TOGGL_ADJUST_APP_TOKEN}", AdjustConfig.EnvironmentProduction));
+            #endif
+        }
+
+        #if USE_ANALYTICS
+        public override bool OpenUrl(UIApplication app, NSUrl url, NSDictionary options)
```
juanlaube: This method is not exactly part of analytics (the conditional is misleading, in fact it's doing google login stuff only).
I think the `OpenUrl` method should stay in the main file or at least be on `AppDelegate.UI`.

Toggl.Core.UI/ViewModels/SelectCountryViewModel.cs
```diff
@@ -38,10 +36,8 @@ public SelectCountryViewModel(INavigationService navigationService, IRxActionFac
             Close = rxActionFactory.FromAsync(close);
         }
 
-        public override async Task Initialize()
+        public override async Task Initialize(long? selectedCountryId)
```
daividssilverio: calling `return base.Initialize(param)` is going to be optional?

Toggl.iOS/Startup/AppDelegate.Analytics.cs
heytherewill: Agreed

Toggl.Core.UI/ViewModels/SelectCountryViewModel.cs
heytherewill: No

Toggl.Core.Tests/UI/ViewModels/ForgotPasswordViewModelTests.cs
```diff
@@ -360,12 +358,11 @@ public void ClosesTheViewModelReturningTheEmail(NonEmptyString emailString)
...
-                    .Received()
-                    .Close(
-                        ViewModel,
-                        Arg.Is<EmailParameter>(
-                            parameter => parameter.Email.Equals(email)));
+
+                var result = ViewModel.ReturnedValue().GetAwaiter().GetResult();
+                result.Email.Should().BeEquivalentTo(email);
+
+                ViewModel.CloseCompletionSource = new TaskCompletionSource<EmailParameter>();
```
daividssilverio: What is this doing here? A little hack to make property tests work?

heytherewill: Yes. Not a hack though, it's just the way things are; a VM is not really meant to be used twice. We should revise our Property Tests to ensure they use one VM per iteration, but that's beside the scope of this PR.

PR #5021, opened 2019-05-02
Implement root presenter

### LINKED ISSUE #4856: Create a presenter for RootViewControllers
Depends on #4852

We need to create an IosPresenter that is capable of displaying ViewModels that are associated with Root View Controllers

juanlaube:
## What's this?
Implements the root presenter

### Relationships

Closes #4856 

## Why do we want this?
🇬🇧 ➡️ 🚪 

## How is it done?
It's a simple presenter that can handle the root view models, instantiate each view controller and set the root view controller on the window.

### Why not another way?
I don't think there's another way without using reflection 🤔

### Side effects
None, this is not used.

## Review hints
This won't compile because the mvexit branch doesn't have `ViewModel` = `ViewModel<Unit, Unit>` 
yet, that means `IPresenter` cannot handle non-generic view models right now.

## :squid: Permissions
🚫


Toggl.iOS/Presentation/RootPresenter.cs
```diff
@@ -0,0 +1,69 @@
...
+
+namespace Toggl.iOS.Presentation
+{
+    public class RootPresenter : IosPresenter
+    {
+        public RootPresenter(UIWindow window, AppDelegate appDelegate) : base(window, appDelegate)
+        {
+        }
+
+        protected override HashSet<Type> AcceptedViewModels => new HashSet<Type>
```
heytherewill: Please make this a property with a getter instead of recreating the HashSet whenever the property is called

Toggl.iOS/ViewControllers/LoginViewController.cs
```diff
@@ -35,6 +35,11 @@ public sealed partial class LoginViewController : ReactiveViewController<LoginVi
         private const int tabletFormOffset = 246;
         private const int tabletLandscapeKeyboardOffset = 90;
 
+        public LoginViewController()
```
heytherewill: This VC uses a storyboard, so it should be loaded using a storyboard instead

juanlaube: Done!

Toggl.iOS/Presentation/RootPresenter.cs
juanlaube: Done!

Toggl.iOS/Presentation/RootPresenter.cs
```diff
@@ -39,7 +39,7 @@ public RootPresenter(UIWindow window, AppDelegate appDelegate) : base(window, ap
                     rootViewController = onboardingViewController;
                     break;
                 case LoginViewModel loginViewModel:
-                    var loginViewController = new LoginViewController();
+                    var loginViewController = LoginViewController.NewInstance();
```
heytherewill: This is beautiful

Toggl.iOS/Presentation/RootPresenter.cs
```diff
@@ -22,6 +18,12 @@ public RootPresenter(UIWindow window, AppDelegate appDelegate) : base(window, ap
             typeof(OutdatedAppViewModel),
         };
 
+        protected override HashSet<Type> AcceptedViewModels => acceptedViewModels;
```
heytherewill: That is not what I meant D:

`protected override HashSet<Type> AcceptedViewModels { get; } = new HashSet<Type>` does the trick without the need for a backing field

heytherewill:
Also, if you check the `TogglPresenter`, I think we were using animations for changing the Root VC, can you verify?

Toggl.iOS/Presentation/RootPresenter.cs
juanlaube: 🙈

juanlaube:
You're right, there are some animations involved, I will use that here too.

heytherewill:
Please do a mergeback from `MVExit` and let's see this working :)

juanlaube:
Putting this on hold because we need to implement `IView` first.

Toggl.iOS/Toggl.iOS.csproj
```diff
@@ -10,6 +10,7 @@
     <RootNamespace>Toggl.iOS</RootNamespace>
     <AssemblyName>Toggl.iOS</AssemblyName>
     <IPhoneResourcePrefix>Resources</IPhoneResourcePrefix>
+    <LangVersion>latest</LangVersion>
```
daividssilverio: Shouldn't we use specific versions? `latest` could bring bugs and unexpected behaviours without us realizing it.

Toggl.iOS/Presentation/RootPresenter.cs
juanlaube: Thanks :)

Toggl.iOS/Toggl.iOS.csproj
juanlaube: Makes sense

PR #5023, opened 2019-05-03
 Fix un-representable date crash in the wheel view

### LINKED ISSUE #5014: Un-representable DateTime in WheelForegroundView.updateEditedTime
## 💥 Crash Report
Seems like you can get the wheel view to crash by adding or subtracting some time span form some date time, which results in an unrepresentable date.

My first guess is that you might get this crash if you set the start date of a time entry to be `DateTimeOffset.MinValue` (jan 1 2001) and then try to set the start time before that using the wheel. Alternatively you could try going higher than `DateTimeOffset.MaxValue`, though why would someone do this?

Also: the crashes happened on April 28 - May 1, so those dates might have something special, though doesn't seem like it. Maybe the jump from April to May causes this crash?

### 👥 Impact
This crash has affected the most users for 1.21, and it seems like each user has experienced the crash only once.
25 users affected in the last 7 days.
25 crashes in the last 7 days

### 🛰 Link
https://appcenter.ms/orgs/Toggl/apps/Toggl-iOS/crashes/errors/3161697801u/overview

`SIGABRT: The added or subtracted value results in an un-representable DateTime. Parameter name: t`
```cs
DateTime.op_Addition (System.DateTime d, System.TimeSpan t)
DateTimeOffset.op_Addition (System.DateTimeOffset dateTimeOffset, System.TimeSpan timeSpan)
WheelForegroundView.updateEditedTime (System.TimeSpan diff)
WheelForegroundView.touchesMoved (CoreGraphics.CGPoint position)
WheelForegroundView.handleTouch (UIKit.UIPanGestureRecognizer recognizer)
UIPanGestureRecognizer+Callback.Activated (UIKit.UIPanGestureRecognizer sender)
```

arturs105:
## What's this?
This fixes the crash described in the issue.

### Relationships

Closes #5014

## How is it done?
I went to fix the exceptions that happen, when the date ends up being out of bounds.

## Review hints
To test this out, you'll need a time entry that's either close to the min date or the max date. Use this piece of code to create an entry that is close to the minimum date:
```cs
var defaultWorkspace = await interactorFactory.GetDefaultWorkspace().Execute();
var tenMinutes = TimeSpan.FromMinutes(10);
var timeEntryPrototype = tenMinutes.AsTimeEntryPrototype(
    DateTimeOffset.MinValue + TimeSpan.FromHours(24),
    defaultWorkspace.Id
);

await interactorFactory.CreateTimeEntry(timeEntryPrototype, TimeEntryStartOrigin.Manual).Execute();
```
After you have this time entry, just try edit the start time to be less than the minimum value. You can't.

## :squid: Permissions
🦑


Toggl.Daneel/Views/EditDuration/WheelForegroundView.cs
```diff
@@ -359,6 +359,27 @@ private void updateEditedTime(TimeSpan diff)
...
+            }
+            catch (ArgumentOutOfRangeException)
+            {
+                //ArgumentOutOfRangeException gets thrown when adding or subtracting results in a DateTime that is out of bounds.
+                //When that happens, we need to clamp the DateTime to either max or min.
+                var deltaToMax = DateTimeOffset.MaxValue - currentDateTimeOffset;
+                var deltaToMin = currentDateTimeOffset - DateTimeOffset.MinValue;
+
+                if (deltaToMin < deltaToMax)
+                    return DateTimeOffset.MinValue;
```
simonrozsival: I don't think these are good fallback values. Maybe it would be easier to specify the fallback value explicitly as a parameter of the method? Or just return the original value unchanged? Having to deal with start or end dates set to min/max date time offset values will be a pain.

Also the cause of the problem is probably here:

`Toggl.Shared/Math.cs`:
```cs
public static double AngleBetween(Point a, Point b)
    => System.Math.Atan2(a.Y - b.Y, a.X - b.X);
```

Maybe in some cases it is `0/0` and atan2 returns infinity or negative infinity and we use the double for calculations which later causes unrepresentable dates? This is probably worth investing. It would also fix the problem in Android, because it has to be there as well.

dshr: `System.Math.Atan2(0, 0)` always returns 0, so this might not be the case :(

simonrozsival: Hmm, you're correct. Then I have no idea what causes unrepresentable times... I still think that min/max value is not a valid value in this case.

arturs105: 🤔 We could change the min/max to be the min/max of Toggl business logic. I think max year was 2030 and the min was somewhere between 2000 and 2010.

dshr: This still feels smelly to me too. I just noticed that Japan has switched eras recently, around May 1st, so maybe it might be bc of that https://docs.microsoft.com/en-us/dotnet/api/system.datetimeoffset?view=netframework-4.8 ? Ping @daividssilverio

arturs105: After thinking a bit about it and researching weird calendars that countries use, i think our best bet would be what Simon suggested: returning the original value in the case of exception.

Also: it might be useful to add a bit of analytics to gather more info, like, what is the `DateTimeOffset` and `TimeSpan` combo that causes this exception.

PR #5024, opened 2019-05-03
Implement IView for iOS

juanlaube:
## What's this?
This implements the `IView` interface for `ReactiveViewController` and `ReactiveTableViewController`.

### Relationships

Ref #4871 

## Why do we want this?
🇬🇧 ➡️ 🚪 mvexit!

## How is it done?
Implementing the interface + extracting already existing pieces of code to extensions and helpers.

### Why not another way?
This is simple.

### Side effects
Nothing.

## :squid: Permissions
🦑


Toggl.iOS/ViewControllers/Reactive/ReactiveViewController.cs
```diff
@@ -0,0 +1,87 @@
...
+            ViewModel?.ViewDisappearing();
+        }
+
+        public override void ViewDidDisappear(bool animated)
+        {
+            base.ViewDidDisappear(animated);
+            ViewModel?.ViewDisappeared();
+        }
+
+        public Task Close()
```
simonrozsival: This could be an extension method for `UIViewController`. This way the code would be shared between the table VC and plan VC. But I'll leave it up to you. No big deal.

heytherewill: Extension methods, sadly, cannot be used as a way to comply with an interface :/

simonrozsival: Oh, that's an interface method... ofc.

heytherewill: Only now have I seen that this method is huge! Yeah, that makes sense 😄

juanlaube:
@daividssilverio I think we should do that in another PR, what do you think @heytherewill?

heytherewill:
Should be done in another PR, as instructed in Slack

Toggl.iOS/ViewControllers/Reactive/ReactiveTableViewController.cs
```diff
@@ -53,26 +52,7 @@ public override void ViewDidDisappear(bool animated)
...
-                }
-            }
-
-            // Is this a modal?
-            else if (PresentingViewController != null)
-            {
-                DismissViewController(true, null);
-            }
-
+            Close();
```
heytherewill: If this is an extension method, you should call is as `this.Close()`. As it stands, this will call itself.

PR #5025, opened 2019-05-03
Add the new droid Activity Presenter

### LINKED ISSUE #4853: Create a presenter for Activities
Depends on #4852 
Depends on #4871

We need to create an AndroidPresenter that is capable of displaying ViewModels that are associated with Activities

daividssilverio:
## What's this?
This is the implementation of the new android presenter for activities.

### Relationships
Closes #4853
Doesn't not depend on #4871

## Why do we want this?
We have to be able to present activities, that's the basic presentation mechanism for android.

## How is it done?
That's basically the same as @heytherewill poc from the custom navigation branch.

### Why not another way?
That works pretty well. Feel free to suggest changes.

### Side effects
One step closer to removing Mvx!

## Review hints
This PR is quite small, go for it.
DialogFragments are not presented yet, don't expect it to work.

## :squid: Permissions
🦑 it, @heytherewill please have a word, since you made the Poc.


Toggl.Droid/Extensions/PermissionRequesterComponentExtensions.cs
```diff
@@ -40,7 +40,7 @@ public static IObservable<bool> ProcessCalendarAuthorizationRequest(this IPermis
                     return permissionRequester.CalendarAuthorizationSubject.AsObservable();
                 
                 permissionRequester.CalendarAuthorizationSubject = new Subject<bool>();
-                permissionRequester.RequestPermissions(new[] { Manifest.Permission.ReadCalendar, Manifest.Permission.WriteCalendar }, calendarAuthCode);
```
heytherewill: Is this change needed?

Toggl.Droid/Presentation/ActivityPresenter.cs
```diff
@@ -0,0 +1,98 @@
...
+        public TViewModel GetCachedViewModel<TViewModel>()
+        where TViewModel : IViewModel
+        {
+            var cachedViewModel = (TViewModel)viewModelCache[typeof(TViewModel)];
+            viewModelCache[typeof(TViewModel)] = null;
+            return cachedViewModel;
+        }
+
+        private static ActivityFlags clearBackStackFlags()
+            => ActivityFlags.ClearTop | ActivityFlags.ClearTask | ActivityFlags.NewTask;
```
heytherewill: No need to make this a method that gets called multiple times. While I'm almost sure this will get inlined, let's just inline it ourselves as a field.

Toggl.Droid/Presentation/ActivityPresenter.cs
```diff
@@ -0,0 +1,98 @@
...
+        
+        struct ActivityPresenterInfo
```
heytherewill:
```suggestion
        private struct ActivityPresenterInfo
```

Toggl.Droid/Presentation/ActivityPresenter.cs
```diff
@@ -0,0 +1,98 @@
...
+        {
+            var viewModelType = viewModel.GetType();
+            
+            if (presentableActivitiesInfos.TryGetValue(viewModelType, out var presentableInfo))
+            {
+                viewModelCache[viewModelType] = viewModel;
+                var intent = new Intent(Application.Context, presentableInfo.ActivityType).AddFlags(presentableInfo.Flags);
+                Application.Context.StartActivity(intent);
+            }
+            else
```
heytherewill: No need for an else branch; early return instead. Also, I would rather early return from the failure branch and continue on the same indentation level as the method

Toggl.Droid/Toggl.Droid.csproj
```diff
@@ -20,6 +20,7 @@
     <EnableProguard>true</EnableProguard>
     <JavaMaximumHeapSize>2G</JavaMaximumHeapSize>
     <AndroidUseSharedRuntime>false</AndroidUseSharedRuntime>
+      <LangVersion>latest</LangVersion>
```
heytherewill: You asked about this on the iOS PR: Why latest instead of a specific one? Also, this is one level of indenting too deep

Toggl.Droid/Extensions/PermissionRequesterComponentExtensions.cs
daividssilverio: Yeah, we don't need this permission, we don't declare it on the `AndroidManifest`, we are just requesting it to see it being rejected automatically.

daividssilverio:
as discussed internally, we are going with:
- Injecting the `INavigationService` as a private field in the base `ViewModel` (every vm will have to have it injected in its constructor)
- Implementing a new `Navigate` method in the base `ViewModel` that will be used for navigation in every VM, that won't accept null as a parameter to execute navigations
- Having the `App` and `ErrorHandlingService` use the raw navigation service, passing null as a parameter to make its navigations

Issue #4857, opened 2019-03-29
Create a presenter for child view controllers


heytherewill:
Depends on #4852

We need to create an IosPresenter that is capable of displaying ViewModels that are associated with child View Controllers

Issue #4855, opened 2019-03-29
Create a presenter for modal dialog fragments


heytherewill:
Depends on #4852

We need to create an AndroidPresenter that is capable of displaying ViewModels that are associated with modal dialog fragments

daividssilverio:
Closed by #5050

PR #5034, opened 2019-05-06
Add source view parameter to navigation and presenters

daividssilverio:
## What's this?
This adds source views as parameters to be passed to navigation and presentation handlers.

### Relationships
Closes Nothing.
Part of #4855 and maybe #4857 

## Why do we want this?
Basically we need a reference to an `Activity` or `Fragment` to present `DialogFragments`.
We need the view because we need the `FragmentManager`.

## How is it done?
A view parameter is added to the navigation and presentation `navigate` and `present` methods.
Simple as that.

### Why not another way?
Because this is the other way. Holding references isn't good for droid.

### Side effects
None.

## Review hints
None.

## :squid: Permissions
🦑 it mvexit bois


daividssilverio:
@heytherewill you rather have that and have to pass `null` when we explicitly don't have a view?

That's the case on `ErrorHandlingService` and `App`.

daividssilverio:
as discussed internally, we are going with:

Injecting the INavigationService as a private field in the base ViewModel (every vm will have to have it injected in its constructor)
Implementing a new Navigate method in the base ViewModel that will be used for navigation in every VM, that won't accept null as a parameter to execute navigations
Having the App and ErrorHandlingService use the raw navigation service, passing null as a parameter to make its navigations

daividssilverio:
No more optionals btw! I learned my lesson.

PR #5046, opened 2019-05-07
Fix the tabs presentation in iOS

### LINKED ISSUE #4860: Create a presenter for Tab Item ViewControllers
Depends on #4852

We need to create an IosPresenter that is capable of displaying ViewModels that are associated with Tab Item ViewControllers

juanlaube:
## What's this?
As the title says, this fixes the tabs presentation in iOS. We don't need a tab presenter, just loading the view models and view controllers is enough.

### Relationships

Closes #4860 

## Why do we want this?
🇬🇧 ➡️ 🚪 

## How is it done?
Creating the view controllers for each view model and setting them as tabs in a UITabBarController.

### Why not another way?
I think we could have a ViewControllerLoader in the future, but right now this is enough.

### Side effects
Nothing.

## Review hints
A few things are still broken:
1. The settings won't open because we need other presenters
1. The reports tab opens in a white screen because the reports calendar is broken, go to another tab and back to the reports tab, it should work
1. The calendar onboarding cannot be dismissed yet because we need other presenters
1. The calendar settings won't open because we need other presenters

## :squid: Permissions
🦑


Toggl.iOS/ViewControllers/MainTabBarController.cs
```diff
@@ -23,27 +23,47 @@ public class MainTabBarController : UITabBarController
...
                 var item = new UITabBarItem();
                 item.Title = "";
-                item.Image = UIImage.FromBundle(imageNameForType[viewModel.GetType()]);
-                screen.TabBarItem = item;
-                controller.PushViewController(screen, true);
-                return controller;
+                item.Image = UIImage.FromBundle(imageNameForType[childViewModel.GetType()]);
+                childViewController.TabBarItem = item;
+                navigationController.PushViewController(childViewController, true);
+                return navigationController;
```
heytherewill: Instead of calling push with animations (which won't be seen), couldn't you just:

`return new UINavigationController(childViewController)`?

Toggl.iOS/ViewControllers/Calendar/CalendarViewController.cs
```diff
@@ -1,22 +1,18 @@
 using System;
 using System.Reactive.Linq;
 using CoreGraphics;
-using MvvmCross;
```
heytherewill: Could you revert this commit? I am doing the same thing in #5037 and I'd rather avoid unnecessary clashes :)

juanlaube: Can dooo!

Toggl.iOS/ViewControllers/MainTabBarController.cs
juanlaube: Good point, consider it done :)

PR #5048, opened 2019-05-07
:world_map: Add navigation presenter for iOS

### LINKED ISSUE #4857: Create a presenter for child view controllers

juanlaube:
## What's this?
Adds the navigation presenter for iOS to push view controllers on a navigation stack. This kind of presentation is what used to be a ChildPresentation in MvvmCross.

### Relationships

Closes #4857

## Why do we want this?
🇬🇧 ➡️ 🚪 

## How is it done?
It's a new presenter. One important detail is the new `Task Cancel()` method in `IViewModel` this is because pressing the back button in the navigation controller needs to cancel the `CloseCompletionSource`.

### Why not another way?
I thought about subclassing `UINavigationController` instead of overriding `DidMoveToParentViewController` to handle the close action, but went with this because it's exactly how MvvmCross handles that.

### Side effects
None.

## :squid: Permissions
🦑


Toggl.Core.UI/ViewModels/Base/ViewModel.cs
```diff
@@ -18,6 +18,12 @@ public async Task Finish(TOutput output)
             CloseCompletionSource.SetResult(output);
         }
 
+        public async Task Cancel()
+        {
+            await View.Close();
+            CloseCompletionSource.SetCanceled();
```
heytherewill: So elegant

Toggl.Core.UI/ViewModels/Base/ViewModel.cs
```diff
@@ -46,6 +52,10 @@ public virtual void ViewDisappeared()
 
         public virtual void ViewDestroyed()
         {
+            if (typeof(TOutput) == typeof(Unit))
```
heytherewill: I don't like this. The VM shouldn't cancel itself, specially not in a lifecycle method.

Toggl.iOS/Presentation/IosPresenter.cs
```diff
@@ -36,5 +36,13 @@ public IosPresenter(UIWindow window, AppDelegate appDelegate)
 
             return tcs.Task;
         }
+
+        protected UIViewController FindPresentedViewController()
+            => FindPresentedViewController(Window.RootViewController);
+
+        private UIViewController FindPresentedViewController(UIViewController current)
```
heytherewill:
```suggestion
        private UIViewController findPresentedViewController(UIViewController current)
```

Toggl.iOS/Presentation/IosPresenter.cs
```diff
@@ -36,5 +36,13 @@ public IosPresenter(UIWindow window, AppDelegate appDelegate)
...
+            => current.PresentedViewController == null || current.PresentedViewController.IsBeingDismissed
+                ? current
+                : FindPresentedViewController(current.PresentedViewController);
```
heytherewill:
```suggestion
                : findPresentedViewController(current.PresentedViewController);
```

Toggl.iOS/ViewControllers/Reactive/ReactiveViewController.cs
```diff
@@ -51,6 +51,16 @@ public override void ViewDidDisappear(bool animated)
             ViewModel?.ViewDisappeared();
         }
 
+        public override void DidMoveToParentViewController(UIViewController parent)
+        {
+            base.DidMoveToParentViewController(parent);
+
+            if (parent == null)
```
heytherewill: America, please explain

Toggl.Core.UI/ViewModels/Base/ViewModel.cs
juanlaube: :sogood:

Toggl.iOS/ViewControllers/Reactive/ReactiveTableViewController.cs
```diff
@@ -51,6 +51,20 @@ public override void ViewDidDisappear(bool animated)
             ViewModel?.ViewDisappeared();
         }
 
+        public override void DidMoveToParentViewController(UIViewController parent)
+        {
+            base.DidMoveToParentViewController(parent);
+
+            // When this is removed from the parent controller (e.g: a pop action in a navigation controller)
+            // then the view is destroyed and the view model needs to be cancelled
+            if (parent == null)
```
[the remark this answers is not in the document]
juanlaube: Fixed

PR #5051, opened 2019-05-08
Fix unrepresentable date crash in edit time entry view

### LINKED ISSUE #5015: Un-representable DateTime in EditTimeEntryViewController
### 👥 Impact
11 reports from 11 users in the last 7 days

### 🔗 Context
Might be related to #5014 

### 🛰 Link
https://appcenter.ms/orgs/Toggl/apps/Toggl-iOS/crashes/errors/4146125805u/overview

`Exception type and message`
```cs
DateTime.op_Addition (System.DateTime d, System.TimeSpan t)
TimeZoneInfo.ConvertTime (System.DateTimeOffset dateTimeOffset, System.TimeZoneInfo destinationTimeZone)
DateTimeToFormattedString.getDateTimeOffsetInCorrectTimeZone (System.DateTimeOffset value, System.TimeZoneInfo timeZone)
DateTimeToFormattedString.Convert (System.DateTimeOffset date, System.String format, System.TimeZoneInfo timeZoneInfo)
EditTimeEntryViewController+<>c.<ViewDidLoad>b__8_10 (System.DateTimeOffset stopTime, Toggl.Core.Models.Interfaces.IThreadSafePreferences preferences)
WithLatestFrom`3+_+FirstObserver[TFirst,TSecond,TResult].OnNext (TFirst value)
Sink`1[TTarget].ForwardOnError (System.Exception error)
WithLatestFrom`3+_+FirstObserver[TFirst,TSecond,TResult].OnNext (TFirst value)
Sink`1[TTarget].ForwardOnNext (TTarget value)
Select`2+Selector+_[TSource,TResult].OnNext (TSource value)
Sink`1[TTarget].ForwardOnNext (TTarget value)
Where`1+Predicate+_[TSource].OnNext (TSource value)
Sink`1[TTarget].ForwardOnNext (TTarget value)
ObserveOnObserverNew`1[T].DrainStep (System.Collections.Concurrent.ConcurrentQueue`1[T] q, System.Boolean delayError)
ObserveOnObserverNew`1[T].DrainShortRunning (System.Reactive.Concurrency.IScheduler recursiveScheduler)
<.cctor>b__20_1 (System.Reactive.Concurrency.IScheduler scheduler, System.Reactive.ObserveOnObserverNew`1[T] self)
NSRunloopScheduler+<>c__DisplayClass2_0`1[TState].<Schedule>b__0 ()
DispatchQueue.static_dispatcher_to_managed (System.IntPtr context)
(wrapper native-to-managed) CoreFoundation.DispatchQueue.static_dispatcher_to_managed(intptr)
```

arturs105:
## What's this?
This is a bit different than the other unrepresentable date crash. This one is not a result of adding a `TimeSpan` to a `DateTimeOffset`, but a result of calling `TimeZoneInfo.ConvertTime`.

### Relationships
Closes #5015

## Why do we want this?
💥➡️🚫💥 

## How is it done?
In the case of encountering unrepresentable date exception, just return the same `DateTimeOffset` that was passed in. This will probably lead to incorrect data being displayed in the app, but it won't crash. I also added some analytics, so we can find out, which `DateTimeOffset` and `TimeZoneInfo` combinations cause the crash.

### Why not another way?
Can't really come up with anything else, since we have no idea, how, why and what data causes these exceptions.

The docs for `TimeZoneInfo.ConvertTime` say that it throws this exception when "The value of the `dateTime` parameter represents an invalid time", which means that we are passing in the unrepresentable `DateTimeOffset` and that date time might be created by #5014.

### Side effects
The app will show the unrepresentable date time instead of crashing.

## :squid: Permissions
🦑


Toggl.Core.UI/Transformations/DateTimeToFormattedString.cs
```diff
@@ -1,11 +1,16 @@
...
 namespace Toggl.Core.UI.Transformations
 {
     public class DateTimeToFormattedString
     {
-        public static string Convert(DateTimeOffset date, string format, TimeZoneInfo timeZoneInfo = null)
+        public static string Convert(
+            DateTimeOffset date,
+            string format,
+            TimeZoneInfo timeZoneInfo = null,
+            IAnalyticsService analyticsService = null)
```
kornelijepetak: You don't seem to be using the `analyticsService` in this method. Why the parameter then?

arturs105: 🤦‍♂ it was meant to be passed to `getDateTimeOffsetInCorrectTimeZone`

amulware:
I definitely do not like the idea of catching exceptions if that leads to bad state. Can't we re-throw an exception with the additional information to help us debug the problem?

Also, why are we using analytics events instead of the app center error api?

Toggl.Core.UI/Transformations/DateTimeToFormattedString.cs
```diff
@@ -1,21 +1,42 @@
...
+            TimeZoneInfo timeZone,
+            IAnalyticsService analyticsService = null)
+        {
+            try
+            {
+                return value == default(DateTimeOffset) ? value : TimeZoneInfo.ConvertTime(value, timeZone);
+            }
+            catch (ArgumentOutOfRangeException)
+            {
+                analyticsService?.UnrepresentableDateErrorWhenConvertingDateToFormattedString?.Track(
```
kornelijepetak: I've only noticed this right now.
I'd also rather have the app crash if `analyticsService` is null. If we don't have core functionality, the app should not work at all. I'd ditch the `?.` operator.

If we go with what you wrote here, what will happen is that an error will happen, and we will not even know it.

PR #5056, opened 2019-05-08
Implement modal dialog presenter for iOS

### LINKED ISSUE #4858: Create a presenter for modal dialog ViewControllers
Depends on #4852

We need to create an IosPresenter that is capable of displaying ViewModels that are associated with modal dialog ViewControllers

juanlaube:
## What's this?
Modal dialogs are back! 🎉 

### Relationships

Closes #4858 

## Why do we want this?
🇬🇧 ➡️ 🚪 

## How is it done?
It's a new presenter, handles what used to be `ModalDialogPresentation`.

### Why not another way?
This is the only way.

### Side effects
Only good side effects:
- Sign up works again
- The calendar works again

## :squid: Permissions
🦑


Toggl.iOS/Presentation/ModalDialogPresenter.cs
```diff
@@ -0,0 +1,91 @@
...
+using Toggl.Core.UI.ViewModels.Calendar;
+using Toggl.Core.UI.Views;
+using Toggl.iOS.Presentation.Transition;
+using Toggl.iOS.ViewControllers;
+using Toggl.iOS.ViewControllers.Calendar;
+using UIKit;
+
+namespace Toggl.iOS.Presentation
+{
+    public class ModalDialogPresenter : IosPresenter
```
simonrozsival: `sealed`

Toggl.iOS/Presentation/ModalDialogPresenter.cs
```diff
@@ -0,0 +1,91 @@
...
+
+        public ModalDialogPresenter(UIWindow window, AppDelegate appDelegate) : base(window, appDelegate)
+        {
+        }
+
+        protected override void PresentOnMainThread<TInput, TOutput>(ViewModel<TInput, TOutput> viewModel, IView sourceView)
+        {
+            UIViewController viewController = null;
+
+            switch (viewModel)
```
simonrozsival: What if we used a `Dictionary<Type, Func<ViewModel<TInput, TOutput>, UIViewController>` instead of this switch?

Toggl.iOS/Presentation/ModalDialogPresenter.cs
```diff
@@ -0,0 +1,91 @@
...
+            {
+                case CalendarPermissionDeniedViewModel calendarPermissionDeniedViewModel:
+                    var calendarPermissionDeniedViewController = new CalendarPermissionDeniedViewController();
```
simonrozsival: I know this is not something introduced in this PR or something that should be discussed here... but I'll mention it here anyway because this is the first time I review a PR related to the refactoring. If we're creating controllers explicitly on iOS, why don't we pass the VM to the constructor of the VC? Is it a good idea to have a mutable VM property in the VC?

Toggl.iOS/Presentation/ModalDialogPresenter.cs
heytherewill: Why keep references of functions living in memory forever when we can have a simple switch?

Toggl.iOS/Presentation/ModalDialogPresenter.cs
heytherewill: I see no reason why not, Simon. Good point. I will create an issue to mass refactor it later, but let's keep it as is atm

juanlaube: Good points, and I was thinking of creating a ViewControllerLocator after I finish with all the presenters, since we have `switch`es like this in every presenter already.

Issue #5026, opened 2019-05-06
Define lightweight pull workflow for background syncing


simonrozsival:
In order to make our app syncs fast when a PN is received or when the app syncs in the background, we should define a simpler pull workflow.

My idea is to pull only time entries using the `?since` parameter and then create placeholder entities for any of the referenced entities which we don't know: workspaces, projects, tasks, clients, tags. This means just 1 HTTP request per PN sync or background sync. These entities will then be replaced with the real entities once we do a proper full sync.

This issue should be thought through more (e.g., is it a good idea to create placeholders for workspaces? what is the alternative? ~will we need a scary migration or just a minor one?~ we don't need any migration, the `SyncStatus` will just be set to another valid enum value `RefetchingNeeded`) and it should be split into several smaller issues.

simonrozsival:
When we fetch a time entry, which belongs to a workspace which we don't have in the database, either we:
- ignore the time entry (even if it is a running time entry)
- create a placeholder for the workspace, when the workspace is later refetched, we should still trigger "new workspace detected" code to make sure the state of our app is correct

I think we should create placeholders even for workspaces. Maybe it even isn't such a big deal 😄

simonrozsival:
The specifications are written in Notion document: https://www.notion.so/Sync-Light-Specification-e9ff6a28db3b496cbd35a3a269340d8a

PR #5079, opened 2019-05-10
Sync light

simonrozsival:
## What's this?
This is a WIP part of the PN project.

### Relationships
Ref #5026 

## Why do we want this?
This should be used to discuss how we want to handle the "lightweight sync" of just TEs with placeholders for the missing referenced entities.

## How is it done?
Modification of the syncing algorithm + syncing tests.

### Why not another way?
This is how I imagine it right now. Let's discuss it and maybe change.

### Side effects
The unit tests are broken at the moment. That will be dealt with later.

## Review hints
This won't be merged just yet. Please consider what is written in the issue and give me feedback for this approach.

## Tests
Syncing tests: ~https://app.bitrise.io/build/10f7b5ee19dc30b8~ ~https://app.bitrise.io/build/71b6ba4fd8e8e204~ https://app.bitrise.io/build/73fa1ccb9f3a0e26 - do not merge this branch unless these tests are passing.

## :squid: Permissions
⛔️ 🦑


amulware:
I'd like us to measure the performance of whatever you implement here compared to our regular sync, so we can later discuss whether we can replace app startup sync with light sync (or maybe at least prepend it).

simonrozsival:
I will create an issue for measuring the performance.

I don't think that it'll be a good idea to use this for the initial sync. We still need to discuss how to display the items which have placeholder clients/tasks/tags/projects/workspaces. Showing `(unsynced client)` will look weird...

Maybe we should then fetch all the placeholder entities by IDs through the reports API or something? I know BE guys didn't like us doing it this way for archived projects but maybe it's not that bad? What do you think?

amulware:
Showing (unsynced client) once in a while might be a worthwhile tradeoff for significantly faster initial sync for all users all the time.

Toggl.Core/Sync/States/Pull/CreateMissingReferencedEntitiesPlaceholdersState.cs
```diff
@@ -0,0 +1,174 @@
...
+                    .SelectMany(timeEntry =>
+                        timeEntry.TagIds?.Select(tagId => (tag: tagId, workspace: timeEntry.WorkspaceId))
+                            ?? new(long, long)[0])
+                    .Distinct(tuple => tuple.tag)
+                    .WhereAsync(tuple => hasUnknownTag(tuple.tag))
+                    .Select(tuple => createTagPlaceholder(tuple.tag, tuple.workspace))
+                    .Count()
+                    .Track(analyticsService.TagPlaceholdersCreated);
+
+            return createWorkspacePlaceholders
```
simonrozsival: I'm not sure if it's a good thing that this is just a single state or if it would be better if this were several consecutive states which would each create one placeholder type. We will not reuse this in any way probably (although the original state for creating placeholders for projects could stay untouched and we would not affect the Full sync workflow (this code will now "slow down" normal syncing a bit (1ms maybe?)).

amulware:
With regards to fetching only unsynced entities, it's worth looking into, can you write up a proposal or with options?

Toggl.Core/Sync/States/Pull/CreateMissingReferencedEntitiesPlaceholdersState.cs
```diff
@@ -0,0 +1,174 @@
...
+
+        private IObservable<IThreadSafeProject> createProjectPlaceholder(ITimeEntry timeEntry)
+        {
+            if (!timeEntry.ProjectId.HasValue)
+            {
+                throw new InvalidOperationException();
+            }
+
+            var placeholder = Project.Builder.Create(timeEntry.ProjectId.Value)
+                .SetName(Resources.InaccessibleProject)
```
simonrozsival: The copy now says `(inaccessible project)` which might not be true now... In case of "Full sync" this will probably create placeholders just for archived projects, which are inaccessible. On the other hand, in the case of "Sync Light" it will create placeholders for projects, which haven't been synced yet and it should say something different.

simonrozsival:
> With regards to fetching only unsynced entities, it's worth looking into, can you write up a proposal or with options?

I'll create a specific issue for that.

simonrozsival:
This draft will now be closed and it will be replaced with a "subproject" PR.

PR #5080, opened 2019-05-10
Unify ViewModel life cycle calls in iOS

### LINKED ISSUE #4864: Make ReactiveViewController call the appropriate VM lifecycle methods
Depends on #4863 

We need to define what methods of UIVIewController map to the ViewModel lifecycle methods we decide to stick to when working on #4863. Then we need to make `ReactiveViewController` call those methods.

juanlaube:
## What's this?
This unifies the view model life cycle calls from ViewControllers.

### Relationships

Closes #4864 

## Why do we want this?
We want the view model life cycles to be called in the same order, regardless of the view and platform.

## How is it done?
For the root view controllers, this is done in the presenter and I had no other choice than pattern-matching on the view controller type (if you have a better idea, let me know).
For the rest of the view controllers, it's all about calling the same methods in order.

### Why not another way?
There's no other way

### Side effects
Nothing

## Review hints
Please do a functional test, add some breakpoints in the view model life cycle and make sure everything is called in the same order.

## :squid: Permissions
🦑


Toggl.iOS/Presentation/RootPresenter.cs
```diff
@@ -69,16 +70,56 @@ public RootPresenter(UIWindow window, AppDelegate appDelegate) : base(window, ap
...
+                () => detachOldRootViewController(oldRootViewController)
             );
         }
 
         private UINavigationController wrapInNavigationController(UIViewController viewController)
             => new UINavigationController(viewController);
+
+        private void detachOldRootViewController(UIViewController viewController)
+        {
+            switch (viewController)
```
heytherewill: Couldn't you:

```csharp
var viewControllerToDetach = viewController is UINavigationController navigationController
    ? navigationController.ViewControllers.First()
    : viewController;
```

juanlaube: Oh that's nice, much better!

PR #5081, opened 2019-05-10
Don't allow spaces in any of the email text fields

### LINKED ISSUE #5076: Remove spaces from password reset field when autofill is used
When you use autofill with one of the e-mail addresses saved on your phone, iOS automatically inserts a space after the e-mail, making the e-mail address invalid. The user has to manually delete the trailing space to reset the password. We could do it automatically.

arturs105:
## What's this?
This fixes the problem described in the issue and more. Now none of the e-mail input fields now accept whitespace characters.

### Relationships

Closes #5076

## How is it done?
By creating a subclass of `LoginTextField` that does not allow whitespace.

## :squid: Permissions
🦑


Toggl.Daneel/Views/LoginTextField.cs
```diff
@@ -12,7 +12,7 @@
 namespace Toggl.Daneel.Views
 {
     [Register(nameof(LoginTextField))]
-    public sealed class LoginTextField : UITextField
+    public class LoginTextField : UITextField
```
simonrozsival: I'm not sure I like this. No, I'm sure I don't like this. Do we have any other use for the `LoginTextField` other than the `EmailTextField`? I don't think so. I think we should stick to "either sealed or abstract" as much as possible.

arturs105: `LoginTextField` is also used for password fields in login and signup views.

simonrozsival: I see. OK, let's not complicate things too much, I'll approve this now.

PR #5105, opened 2019-05-13
Siri: Edit intent texts with description

### LINKED ISSUE #4641: Siri: Review texts for watch
We can have specific texts for the Apple Watch and though some of them or short enough to be the same as the ones in the phone, there are others that get cropped. We should list them, review them and find alternatives.

Odrakir:
## What's this?
At first this was meant to make different versions of the shorcut texts to fit the Apple Watch by using `NSStringVariableWidthRuleType`, but in the end this is not really necessary as the only texts that can have problems are user's, so there's no solution for that.

What I did do was using the entry description as subtitle for intents with it, instead of using always the workspace. That way we remove it from the title making it a little bit easier to fit in the Apple Watch.

### Relationships
Closes #4641 

## Review hints
This is just a simple change to the intent definition file.

## :squid: Permissions
@khoi and me


Toggl.iOS.IntentStaticLib/Toggl.iOS.IntentStaticLib.xcodeproj/project.pbxproj
```diff
@@ -7,15 +7,15 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
-		8C23AA902282DB51004B97E6 /* StartTimerFromClipboardIntent.m in Sources */ = {isa = PBXBuildFile; fileRef = 8C23AA852282DB4F004B97E6 /* StartTimerFromClipboardIntent.m */; };
-		8C23AA922282DB51004B97E6 /* StartTimerIntent.m in Sources */ = {isa = PBXBuildFile; fileRef = 8C23AA882282DB50004B97E6 /* StartTimerIntent.m */; };
-		8C23AA932282DB51004B97E6 /* ShowReportIntent.m in Sources */ = {isa = PBXBuildFile; fileRef = 8C23AA892282DB50004B97E6 /* ShowReportIntent.m */; };
```
khoi: I don't think we need to change IntentStaticLib

Odrakir: You are probably right. But it doesn't hurt, does it? To revert this change back I'd have to "push -f" and it seems unnecessary...

khoi: a 
```bash
git checkout project/siri-newfeatures -- Toggl.iOS.IntentStaticLib/Toggl.iOS.IntentStaticLib.xcodeproj/project.pbxproj
```
should do

Issue #4867, opened 2019-03-29
Implement Url Navigation for Android


heytherewill:
Depends on #4862 

We need to implement Url based navigation. Since we are redoing whole thing, we can make our deep links behave like proper deep links, using [TaskStackBuilder](https://developer.android.com/reference/android/support/v4/app/TaskStackBuilder) instead of  start multiple activities, for example.

Bear in mind that once the MvvmCross dependency is gone and we figure out what to do with the Droid startup, we will be able to make different activities handle different urls, so let us keep this system flexible

PR #5106, opened 2019-05-13
Add url-based navigation

### LINKED ISSUE #4868: Implement Url Navigation for iOS
Depends on #4862 

We need to implement Url based navigation. Since we are redoing the whole thing, we can make take this opportunity investigate whether implementing [`x-callback-url`](http://x-callback-url.com/) is worth or not.

juanlaube:
## What's this?
Adds url-based navigation.

### Relationships

Closes #4868
Refs #4867

## How is it done?
Adding a new UrlHandler service that can handle urls, change navigation and presentation.

### Why not another way?
This is still WIP, but seems to be the best way.

### Side effects
Nothing so far.

## :squid: Permissions
⛔️


Toggl.iOS/Startup/AppDelegate.Background.cs
```diff
@@ -28,13 +28,11 @@ public override void OnActivated(UIApplication application)
 
         public override void WillEnterForeground(UIApplication application)
         {
-            base.WillEnterForeground(application);
```
heytherewill: This should be in a 🐛 PR to make these easier to track :)

Toggl.Core.UI/Services/UrlHandler.cs
```diff
@@ -68,10 +68,12 @@ public async Task<bool> Handle(Uri uri)
 
                 case "/reports":
                     // e.g: toggl://tracker/reports?workspaceId=1&startDate="2019-04-18T09:35:47Z"&endDate="2019-04-18T09:45:47Z"
+                    viewPresenter.ChangePresentation(PresentationChange.ShowReports);
                     return true;
 
                 case "/calendar":
                     // e.g: toggl://tracker/calendar?eventId=1
+                    viewPresenter.ChangePresentation(PresentationChange.ShowCalendar);
```
heytherewill: This should check if the event id has been passed. If it has, it should try creating a new time entry from that calendar id

Toggl.Core.UI/Services/UrlHandler.cs
```diff
@@ -68,10 +68,12 @@ public async Task<bool> Handle(Uri uri)
 
                 case "/reports":
                     // e.g: toggl://tracker/reports?workspaceId=1&startDate="2019-04-18T09:35:47Z"&endDate="2019-04-18T09:45:47Z"
+                    viewPresenter.ChangePresentation(PresentationChange.ShowReports);
```
heytherewill: This should take the actual URL parameters in consideration

daividssilverio:
Looking nice so far. It seems like the droid part will fit quite well (unless we need to handle deeplinks to  dialog fragments, in which case it will be a bit harder, but still possible).

Toggl.Core.UI/Services/UrlHandler.cs
```diff
@@ -0,0 +1,88 @@
...
+                    // e.g: toggl://tracker/calendar?eventId=1
+                    viewPresenter.ChangePresentation(PresentationChange.ShowCalendar);
+                    return true;
+
+                default:
+                    return false;
+            }
+        }
+
+        public Task navigate<TViewModel, TNavigationInput>(TNavigationInput payload)
```
daividssilverio: Use PascalCase for public methods.

juanlaube: It should be a private method :)

Toggl.Core.UI/Services/UrlHandler.cs
```diff
@@ -0,0 +1,365 @@
...
+            var workspaceId = getLong(args, "workspaceId");
+            var startTime = getDateTimeOffset(args, "startTime");
+            var stopTime = getDateTimeOffset(args, "stopTime");
+            var description = getString(args, "description");
+            var projectId = getLong(args, "projectId");
+            var taskId = getLong(args, "taskId");
+            var tags = getLongs(args, "tags");
+            var isBillable = getBool(args, "billable") ?? false;
+
+            if (!timeEntryId.HasValue && !workspaceId.HasValue && !startTime.HasValue)
```
daividssilverio: Can we update time entries without Ids?
Shouldn't this be using `||`s instead?

Toggl.Core.UI/Services/UrlHandler.cs
```diff
@@ -0,0 +1,365 @@
...
+            return true;
+        }
+
+        private bool handleTimeEntryEdit(Dictionary<string, string> args)
+        {
+            // e.g: toggl://tracker/timeEntry/edit?timeEntryId=1
+            var timeEntryId = getLong(args, "timeEntryId");
+            if (timeEntryId.HasValue)
+            {
+                navigate<EditTimeEntryViewModel, long[]>(new[] { timeEntryId.Value });
```
daividssilverio: shouldn't this be waited?

Toggl.Core.UI/Services/UrlHandler.cs
```diff
@@ -0,0 +1,365 @@
...
+
+            return true;
+        }
+
+        #endregion
+
+        private Task navigate<TViewModel, TNavigationInput>(TNavigationInput payload)
+            where TViewModel : ViewModel<TNavigationInput, Unit>
+            => navigationService.Navigate<TViewModel, TNavigationInput, Unit>(payload, null);
+
```
daividssilverio: 🔥 extra line

daividssilverio:
Also @juanlaube, there's a broken reference on `Toggl.Droid/Helper/PersistentNotificationsHelper.cs:L18`

```
 private static Uri togglMainNavigationUri = Uri.Parse(ApplicationUrls.Main.Open);
```
`ApplicationUrls.Main` isn't defined.

Toggl.Core.UI/Services/UrlHandler.cs
```diff
@@ -57,37 +66,46 @@ public async Task<bool> Handle(Uri uri)
...
 
-        private bool handleTimeEntryStop(Dictionary<string, string> args)
+        private async Task<bool> handleTimeEntryStop(Dictionary<string, string> args)
         {
-            // e.g: toggl://tracker/timeEntry/stop?stopTime="2019-04-18T09:45:47Z"&duration=600&source=Siri
-            var stopTime = getDateTimeOffset(args, "stopTime");
-            var duration = getTimeSpan(args, "duration");
-            var source = getString(args, "source");
+            // e.g: toggl://tracker/timeEntry/stop?stopTime="2019-04-18T09:45:47Z"&source=Siri
+            var stopTime = getDateTimeOffset(args, "stopTime") ?? DateTimeOffset.Now;
```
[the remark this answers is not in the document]
heytherewill: I think I'd rather see something in the lines of extension methods:

`args.GetValueAsDateTimeOffset("stopTime") ?? DateTime.Now`

What do you think?

Toggl.Core.UI/Services/UrlHandler.cs
```diff
@@ -57,37 +66,46 @@ public async Task<bool> Handle(Uri uri)
...
+            var source = getStopOrigin(args, "source");
```
heytherewill: I don't like that these values are all hard-coded strings. Please refactor to make them constants, this way we can change the protocol without worrying much about breaking everything

Toggl.Core.UI/Services/UrlHandler.cs
```diff
@@ -96,24 +114,67 @@ private bool handleTimeEntryCreate(Dictionary<string, string> args)
...
-            var duration = getTimeSpan(args, "duration");
             var description = getString(args, "description");
             var projectId = getLong(args, "projectId");
             var taskId = getLong(args, "taskId");
             var tags = getLongs(args, "tags");
-            var billable = getBool(args, "billable");
+            var isBillable = getBool(args, "billable") ?? false;
+
+            if (!timeEntryId.HasValue && !workspaceId.HasValue && !startTime.HasValue)
+            {
```
heytherewill: No need for curly braces when returning or throwing

Toggl.Core.UI/Services/UrlHandler.cs
heytherewill: He's right. Though we should default the workspace id to default workspace and the start time to now

Toggl.Core.UI/Services/UrlHandler.cs
```diff
@@ -96,24 +114,67 @@ private bool handleTimeEntryCreate(Dictionary<string, string> args)
...
+                return false;
+            }
+
+            var editTimeEntryDto = new EditTimeEntryDto
+            {
+                Id = timeEntryId.Value,
+                WorkspaceId = workspaceId.Value,
+                StartTime = startTime.Value,
```
heytherewill: How are you making sure that you are not updating parameters that were not supplied?

Toggl.Core.UI/Services/UrlHandler.cs
```diff
@@ -240,5 +310,30 @@ private string getString(Dictionary<string, string> dictionary, string key)
...
+            }
+
+            return TimeEntryStartOrigin.Deeplink;
+        }
+
+        private TimeEntryStopOrigin getStopOrigin(Dictionary<string, string> dictionary, string key)
+        {
+
+            if (dictionary.TryGetValue(key, out var sourceString) && TimeEntryStopOrigin.TryParse<TimeEntryStopOrigin>(sourceString, out var source))
+            {
```
heytherewill: No need for braces when returning

Toggl.Core.UI/Services/UrlHandler.cs
```diff
@@ -240,5 +310,30 @@ private string getString(Dictionary<string, string> dictionary, string key)
...
+                return source;
+            }
+
+            return TimeEntryStartOrigin.Deeplink;
+        }
+
+        private TimeEntryStopOrigin getStopOrigin(Dictionary<string, string> dictionary, string key)
+        {
+
+            if (dictionary.TryGetValue(key, out var sourceString) && TimeEntryStopOrigin.TryParse<TimeEntryStopOrigin>(sourceString, out var source))
```
heytherewill: This is a pretty confusing line of code. I'd rather see it as a nicely named extension method:

`dictionary.TryParseValueAsEnum<TimeEntryStopOrigin>(key, out var source)`

Toggl.Core/Extensions/TimeEntryExtensions.cs
```diff
@@ -69,44 +72,44 @@ public static ITimeEntryPrototype AsTimeEntryPrototype(this IThreadSafeTimeEntry
                 timeEntry.Task?.Id,
                 timeEntry.TagIds.ToArray(),
                 timeEntry.Billable);
+    }
 
-        private sealed class TimeEntryPrototype : ITimeEntryPrototype
-        {
-            public long WorkspaceId { get; }
+    public sealed class TimeEntryPrototype : ITimeEntryPrototype
```
heytherewill: I liked when this was private. Why not use the provided extension methods or create another one? Exposing this class is super leaky

Toggl.Core.UI/Services/UrlHandler.cs
```diff
@@ -207,10 +207,24 @@ private bool handleReports(Dictionary<string, string> args)
...
 
-        private bool handleCalendar(Dictionary<string, string> args)
+        private async Task<bool> handleCalendar(Dictionary<string, string> args)
         {
             // e.g: toggl://tracker/calendar?eventId=1
-            viewPresenter.ChangePresentation(PresentationChange.ShowCalendar);
+            var calendarItemId = getString(args, "eventId");
+
+            if (calendarItemId == null)
+            {
```
heytherewill: No need for braces when returning

Toggl.Core.UI/Services/UrlHandler.cs
juanlaube: Nope, it's fire and forget.

Toggl.Core.UI/Services/UrlHandler.cs
```diff
@@ -0,0 +1,365 @@
...
+            };
+
+            await interactorFactory.UpdateTimeEntry(editTimeEntryDto).Execute();
+
+            return true;
+        }
+
+        private bool handleTimeEntryNew(Dictionary<string, string> args)
+        {
+            // e.g: toggl://tracker/timeEntry/new?workspaceId=1&startTime="2019-04-18T09:35:47Z"&stopTime="2019-04-18T09:45:47Z"&duration=600&description="Toggl"&projectId=1&taskId=1&tags=[]&billable=true
```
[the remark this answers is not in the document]
juanlaube: Ouch, forgot about this, thanks!

Toggl.Core.UI/ViewModels/Reports/ReportsViewModel.cs
```diff
@@ -417,5 +417,37 @@ private async Task selectWorkspace()
...
+
+        public async Task ChangePresentation(ShowReportsPresentationChange presentationChange)
+        {
+            var workspaceId = presentationChange.WorkspaceId;
+            var startDate = presentationChange.StartDate;
+            var endDate = presentationChange.EndDate;
+
+            if (startDate.HasValue && endDate.HasValue)
+            {
+                IThreadSafeWorkspace workspace;
```
heytherewill:
```csharp
var getWorkspaceInteractor = workspaceId.HasValue 
    ? interactorFactory.GetWorkspaceById(workspaceId.Value)
    : interactorFactory.GetDefaultWorkspace();

var workspace = getWorkspaceInteractor.Execute();
```

Toggl.Core.UI/ViewModels/MainViewModel.cs
```diff
@@ -158,10 +157,8 @@ public sealed class MainViewModel : ViewModel
             ratingViewExperiment = new RatingViewExperiment(timeService, dataSource, onboardingStorage, remoteConfigService);
         }
 
-        public void Init(string action, string description)
+        public void Init(string description)
```
heytherewill: This method never gets called and can be safely 🔥

Toggl.Core.UI/Services/UrlHandler.cs
juanlaube: Only the parameters in the URL are updated now.

Toggl.Core/Extensions/TimeEntryExtensions.cs
juanlaube: It's private again.

Toggl.Core.UI/ViewModels/Reports/ReportsViewModel.cs
```diff
@@ -428,5 +428,33 @@ private async Task selectWorkspace()
 
         private float percentageOf(List<ChartSegment> list)
             => list.Sum(segment => segment.Percentage);
+
+        public async Task ChangePresentation(ShowReportsPresentationChange presentationChange)
```
heytherewill: I think this could be better named. A presentation change is, inherently, a UI change. This could be a method called `ChangeRange` (and I think we probably already have ways of changing a range).

Toggl.Core/Analytics/TimeEntryStartOrigin.cs
```diff
@@ -64,6 +64,11 @@ public enum TimeEntryStartOrigin
         /// <summary>
         /// Starting a time entry via Siri
         /// </summary>
-        Siri
+        Siri,
+
+        /// <summary>
+        /// Starting a time entry via an url
+        /// </summary>
+        Deeplink,
```
heytherewill: <3

Toggl.Droid/Presentation/AndroidPresenter.cs
```diff
@@ -28,5 +28,7 @@ public abstract class AndroidPresenter : IPresenter
 
             return tcs.Task;
         }
+
+        public abstract bool ChangePresentation(IPresentationChange presentationChange);
```
heytherewill: I don't think this should be abstract. Most presenters won't handle this. Just make this virtual and return false; then consumers can opt-in if they need to handle presentation changes

Toggl.iOS/Presentation/ModalCardPresenter.cs
```diff
@@ -44,5 +45,8 @@ public ModalCardPresenter(UIWindow window, AppDelegate appDelegate) : base(windo
             UIViewController topmostViewController = FindPresentedViewController();
             topmostViewController.PresentViewController(viewController, true, null);
         }
+
+        public override bool ChangePresentation(IPresentationChange presentationChange)
+            => false;
```
heytherewill: Similar to what I said about droid

Toggl.Core.UI/ViewModels/Reports/ReportsViewModel.cs
juanlaube: Better and DRY incoming!

Issue #4866, opened 2019-03-29
Make ReactiveFragment call the appropriate VM lifecycle methods


heytherewill:
Depends on #4863 

We need to define what methods of Fragment map to the ViewModel lifecycle methods we decide to stick to when working on #4863. Then we need to make `ReactiveFragment` call those methods.

daividssilverio:
About the `ReactiveDialogFragment`, the lifecycle is is also alright.
It's almost identical to the Activity, except by the `AttachView`/`DetachView` calls, which on the DialogFragment happen on `OnViewCreated` and `OnDestroyView` respectively.

heytherewill:
Keeping this issue open and removing Daivid's assignment because we still need to deal with some corner cases, but that's for later.

daividssilverio:
I've double checked the lifecycles of our `TabFragments` and `DialogFragments`, alongside with `Activities` as well.

Found nothing critical related to the lifecycle itself. I say we ready to close this issue.

On the other hand, I found some leakage happening in the `Activities` and the cursors opened in the calendar service. I'll open other issues for that, as they were there before and not introduced by mvexit.

Feel free to close this issue at your discretion.

daividssilverio:
Also ref #5230 #5083, PRs that handled fragments details.

heytherewill:
Given the above, I consider this closed

Issue #4865, opened 2019-03-29
Make ReactiveActivity call the appropriate VM lifecycle methods


heytherewill:
Depends on #4863
Depends on #4871

We need to define what methods of `Activity` map to the ViewModel lifecycle methods we decide to stick to when working on #4863. Then we need to make `ReactiveActivity` call those methods.

daividssilverio:
The Activity lifecycle is alright.
We could use different strategies, but for our purposes and how everything is setup right now, it makes sense to have:

`OnCreate` -> `ViewModel.Attach` (Activity reference is there)
`OnDestroy` -> `ViewModel.Detach` (Activity is going to be lost)

`OnStart` -> `ViewModel.ViewAppearing`
`OnResume` -> `ViewModel.ViewAppeared` (view is visible)
`OnPause` -> `ViewModel.ViewDisappearing` (view can be visible or invisible, might disappear)
`OnStop` -> `ViewModel.ViewDisappeared` (view is invisible)

Right now we are not using the lifecycle to release resources when we are not using them. But we can do it if necessary later.

heytherewill:
Fair enough

PR #5107, opened 2019-05-13
Cache view models & change vm injection on droid Activities & DialogFragments

### LINKED ISSUE #5061: Make each IViews ViewModel property readonly
This means receiving the ViewModel as part of the constructor and then passing it back to base. Since we have manual control over this, there's no reason not to.

Since activities are the oddball here, they won't receive it during construction but rather just continue doing what they are doing right now.

daividssilverio:
## What's this?
This PR introduces a more aggressive cache for VMs for `Activities` and `DialogFragments` and change where the VMs are injected to accommodate droid's lifecycles.

### Relationships
Closes #5061
Make changes again to lifecycles Ref #4865 #4866 

## Why do we want this?
To improve the behaviour of the app when the application is restored after activities are killed by the system to free resources.

## How is it done?
The android framework needs empty constructors for both `Activities` and `Fragments`, it uses them to recreate the components after they are destroyed to free resources and need to be recreated.

The dialog fragments in special needed to have the bindings setup changed to be executed after calling `base.OnViewCreated` to make sure the VM is available for use. The dialogs work for the common usage, but they will make the app crash if a user leaves the app with the dialog open and come back much later or after doing some resource intensive activity (recording a video for exemple), depending on the device.

We still need to add proper `Activity`/`Fragment` recreation.

### Why not another way?
I tried to use the android architecture `ViewModel` component, but the use case isn't quite the same. We could use it someday when we start supporting tablets and foldable phones (if ever).

### Side effects
If the view models are not properly closed, the cache could end up causing leaks.

## Review hints
Review by commit.
Please do some functional testing.

## :squid: Permissions
🦑 mvexit bois


heytherewill:
Code is looking good. I will functional test this later today

Toggl.Droid/Activities/MainTabBarActivity.cs
```diff
@@ -41,6 +44,30 @@ protected override void OnCreate(Bundle savedInstanceState)
                 .DisposedBy(DisposeBag);
         }
 
+        private void restoreFragmentsViewModels()
+        {
+            SupportFragmentManager
+                .Fragments
+                .ForEach(frag =>
+                {
+                    switch (frag)
```
kornelijepetak: I will not force you to change this, but this javascript-looking code is breaking levels of abstractions 😛

daividssilverio: Please, give your suggestion.
I thought about changing the public `Tabs` to use a `Dictionary`, make it private and then have a method to get the VM, but it felt unnecessary.

heytherewill: I think using `ViewModel.Tabs.OfType<MainViewModel>().Single()` is already a good start over accessing indexes

kornelijepetak: @daividssilverio When I said 'I will not force you' that meant that I don't put obstacles to approvals/merge. 

Your suggestion is better, but I was more concerning about levels of abstractions. What I wanted is to put the factory switch into a method. But it's not a heavy case this one, so that's why I did not hard-request-changes.

PR #5129, opened 2019-05-15
🐛 Fix not all tasks showing as suggestions in start time entry view

### LINKED ISSUE #5059: Autocomplete does not return all matching time entries
## 🐛 Bug Report
Autocomplete seems to ignore the task field of time entries, thus returning an incomplete list of previously created time entries. 

## Context

Seems to be the case on both iOS and Android.

## Expected behavior
When creating a new time entry and writing anything in the description field, autocomplete returns examples of all previously created time entries that are matching to the text I have typed. 

## To reproduce
Create a time entry and assign it to a task. 
Create a time entry with an identical description and assign it to a different task (in the same project).
Create a new time entry and start typing the same description as before - only one of the previously created time entries appears, although previous two time entries belonged to different tasks and were thus different. 

Example screenshots attached as well. 

IMG_3076
IMG_3077

kornelijepetak:
## What's this?
All searchable tasks are now visible when substring is matching.

### Relationships
Closes #5059

## Why do we want this?
Because it's the expected behaviour.

## How is it done?
We were incorrectly comparing TEs.

### Why not another way?
The trivial one.

## :squid: Permissions
Me.


Toggl.Core.UI/ViewModels/StartTimeEntryViewModel.cs
```diff
@@ -252,7 +253,8 @@ public override async Task Initialize()
                         // Intentionally left blank
                     }
                 }
-                if (initialParameters.TagIds != null) {
+                if (initialParameters.TagIds != null)
```
heytherewill: Egyptian brackets are the source of all bugs

kornelijepetak: IKR. Who knows what else in our app is broken because of the egyptian brackets? I just had to fix them.

PR #5130, opened 2019-05-15
Fix 1password service on iOS

### LINKED ISSUE #5126: Fix the OnePassword extension on iOS
Since this is not even being used on Droid anyways, we might use this opportunity to refactor it as a platform specific thing, integrated with the LoginViewController

heytherewill:
## What's this?
This fixes the commented out code from `OnePasswordService.cs` on iOS

### Relationships
Closes #5126

## Why do we want this?
1password is useful

## How is it done?
Since this service is only ever needed in iOS and we were doing many a stretch to make it cross-platform, we simply accepted our shortcomings and made it work on the iOS layer only

### Why not another way?
We could do it by fixing the existing services, but we were thinking about removing this for a long time. TBH, since iOS now supports native password managers, we could possibly remove this service completely, but that's not up to me.

### Side effects
I'd like someone to test this on a device and check if everything works fine but in principle none.

## :squid: Permissions
Me


heytherewill:
I have started internal discussions as to whether we should completely remove this or not. Putting it on hold for the time being

simonrozsival:
If we're gonna remove this btn from the app, then we must finally finish the native password manager integration (we should do that anyway).

PR #5132, opened 2019-05-15
Add iOS startup docs

### LINKED ISSUE #4870: Figure out and document the new startup
We need to figure out how the app is going to boot and load the first ViewModel now that we no longer use MvvmCross

juanlaube:
## What's this?
📝 Docs!

### Relationships

Closes #4870 

## Why do we want this?
📝 Docs, we don't want to write, but we want to have them.

## How is it done?
With ❤️ 

### Why not another way?
Because 💔 is 😞 

### Side effects
None

## :squid: Permissions
🦑


docs/app-lifecycle/startup-ios.md
```diff
@@ -0,0 +1,37 @@
+# The iOS Startup
+
+This document explains the iOS startup and also how the ViewModels are injected on the different `ViewController` instances.
+
+
+## Startup
+
+As with any other iOS application, `AppDelegate` is the class responsible for controling the app lifecycle. For mantainabilty, this class is `partial` and splitted in several files.
```
heytherewill:
```suggestion
As with any other iOS application, `AppDelegate` is the class responsible for controlling the app lifecycle. For maintainability, this class is `partial` and split into several files.
```

docs/app-lifecycle/startup-ios.md
```diff
@@ -0,0 +1,37 @@
...
+
+The app startup happens on `AppDelegate.FinishedLaunching`. The initialization follow these steps:
```
heytherewill:
```suggestion
The app startup happens on `AppDelegate.FinishedLaunching`. The initialization follows these steps:
```

docs/app-lifecycle/startup-ios.md
```diff
@@ -0,0 +1,37 @@
...
+1. Create a window and make it key and visible.
+1. Create the different view presenters, these all depend on the the window and app delegate.
```
heytherewill:
```suggestion
1. Creating the view presenters. These all depend on `Window` and `AppDelegate`.
```

docs/app-lifecycle/startup-ios.md
```diff
@@ -0,0 +1,37 @@
...
+
+This document explains the iOS startup and also how the ViewModels are injected on the different `ViewController` instances.
+
+
+## Startup
+
+As with any other iOS application, `AppDelegate` is the class responsible for controling the app lifecycle. For mantainabilty, this class is `partial` and splitted in several files.
+
+The app startup happens on `AppDelegate.FinishedLaunching`. The initialization follow these steps:
+1. Create a window and make it key and visible.
```
heytherewill:
```suggestion
1. Creating a window and make it key and visible.
```

docs/app-lifecycle/startup-ios.md
```diff
@@ -0,0 +1,37 @@
...
+1. Create the different view presenters, these all depend on the the window and app delegate.
+1. Calling `IosDependencyContainer.Instance.EnsureInitialized` injects the view presenter and ensures our dependency container is ready to be used.
```
heytherewill:
```suggestion
1. Calling `IosDependencyContainer.Instance.EnsureInitialized`, which injects the view presenter and ensures our dependency container is ready to be used.
```

docs/app-lifecycle/startup-ios.md
```diff
@@ -0,0 +1,37 @@
...
+1. Create the app.
```
heytherewill:
```suggestion
1. Creating the app.
```

docs/app-lifecycle/startup-ios.md
```diff
@@ -0,0 +1,37 @@
...
+1. Calling `app.Start()` will navigate to the first viewmodel.
```
heytherewill:
```suggestion
1. Calling `app.Start()`, which will navigate to the first ViewModel.
```

docs/app-lifecycle/startup-ios.md
```diff
@@ -0,0 +1,37 @@
...
+
+## URL-based navigation
+
+`AppDelegate.OpenUrl` is the entry point for all url-based navigations, from inside or outside the app. Because this method is part of the app lifecycle and it's called after `AppDelegate.FinishedLaunching`, we don't need to ensure the app is initialized.
+All urls with the `toggl` scheme are forwarded to the `UrlHandler`.
+
+
+## ReactiveViewController
+
+`ReactiveViewController`, ahd it's twin `ReactiveTableViewController`, are our implementation of `IView`. All our view controllers are subclasses of these two parent classes (with a few exceptions).
```
heytherewill:
```suggestion
`ReactiveViewController`, and its twin `ReactiveTableViewController`, are our implementation of `IView`. All our view controllers are subclasses of these two parent classes (with a few exceptions).
```

docs/app-lifecycle/startup-ios.md
```diff
@@ -0,0 +1,37 @@
...
+The `ReactiveViewController` parent class is responsible for binding their own lifecycle with the view model lifecycle.
```
heytherewill: The usage of the word parent here is a bit confusing. Base would be clearer

docs/app-lifecycle/startup-ios.md
```diff
@@ -0,0 +1,37 @@
...
+
+The ViewModel property in `ReactiveViewController` is not readonly, but should be used like that (there's an exception that doesn't allow us to have an actual readonly property).
```
heytherewill:
```suggestion
The `ViewModel` property in `ReactiveViewController` is not read-only but should be used like that (there's an exception that doesn't allow us to have an actual read-only property).
```

juanlaube:
@simonrozsival I added more information about all the entry points.

PR #5133, opened 2019-05-15
Add documentation for url navigation

juanlaube:
## What's this?
Documents how deeplink and url navigation works, and all the urls we support

### Relationships

Refs #4868 

## Why do we want this?
📝 

## How is it done?
📝 

## :squid: Permissions
🦑


docs/url-navigation.md
```diff
@@ -0,0 +1,191 @@
...
+1. billable: `bool`, optional
+
+Examples:
+1. toggl://tracker/timeEntry/update?timeEntryId=1workspaceId=1&startTime="2019-04-18T09:35:47Z"&stopTime="2019-04-18T09:45:47Z"&description="Toggl"&projectId=1&taskId=1&tags=[]&billable=true
+
+
+*New time entry*
+
+Description: navigates to `StartTimeEntryViewModel` passing the given parameters
+Url: toggl://tracker/timeEntry/new?workspaceId=1&startTime="2019-04-18T09:35:47Z"&stopTime="2019-04-18T09:45:47Z"&duration=600&description="Toggl"&projectId=1&tags=[]
```
simonrozsival: This is a bit confusing to me - the URLs don't have a verb like HTTP requests, so it's hard to distinguish which action actually does something (like `start`) and which just navigates to some view. I know this PR is just about the documentation and so this isn't a good place where to discuss this but while we're at it - was there or is there still some discussion about the URLs? I'd like to join it.

docs/url-navigation.md
```diff
@@ -0,0 +1,191 @@
...
+1. Use static methods in those classes to build the actual url
+
+
+## Supported urls
+
+### Time entries
+
+*Start time entry*
+
+Description: starts a new time entry
```
simonrozsival: Will the app then be at the main view or will it go to the edit view for the new TE or what else happens? I think it should be documented what will be the final state of the app after navigating to the URL (this applies to all the other URLs as well).

docs/url-navigation.md
juanlaube: Talk with us in `#project-mvexit`! We can still change this 🙂

PR #5134, opened 2019-05-15
Fix reports loading

### LINKED ISSUE #5112: Reports are not loading
When you open the reports tab, no information appears. Seems like something is erroring out.

daividssilverio:
## What's this?
This is a dirty fix for the reports on droid (it seems it wasn't happening on iOS).

### Relationships
Closes #5112 

## Why do we want this?
We want to see some reports o/

## How is it done?
By checking if the start and end dates are valid.

### Why not another way?
This is a dirty fix, the right thing to do would be actually refactoring the `ReportsViewModel` and `ReportsCalendarViewModel` to untangle them and fix the initialization logic on both view models.

### Side effects
None, this just prevents unexpected behaviour.

## Review hints
Go for it, there's only one commit.

## :squid: Permissions
🦑 it mvexit bois


simonrozsival:
I don't really understand how this fixes the situation (does the observable error out when the dates are `default`?) 🤔 Could we maybe instead initialize `startDate` and `endDate` to be some valid dates, if that's what causes the problem? 

If you say that there is a proper solution, but we won't do it now (that's understandable), please create an issue and link it to this PR so we don't forget about this.

daividssilverio:
@simonrozsival that works as well :)
If we move the initialization that is happening inside the `CalendarReportsViewModel`, and is being propagated to the `ReportsViewModel`, to the `ReportsViewModel` itself, the problem should be fixed.

heytherewill:
Superseded by 5145

PR #5135, opened 2019-05-16
Re-add the droid url nav

### LINKED ISSUE #4867: Implement Url Navigation for Android

daividssilverio:
## What's this?
This re-adds the droid url & deeplink handling.

### Relationships
Closes #4867 

## Why do we want this?
To support url navigation & deeplinks.

## How is it done?
It uses the already existing `UrlHandler` from `Toggl.Core.UI`, delegating the common methods to it and using the `AndroidUrlHandler` to handle urls that need special care on droid.

Why draft? What still needs to be done?
- [x] ApplicationUrls.TimeEntry.New.Path
- [x] ApplicationUrls.TimeEntry.Edit.Path
- [x] ApplicationUrls.Reports.Path
- [x] ApplicationUrls.Calendar.Path 
- ~Update `App.cs` to navigate properly for the cases handled by the `Toggl.Core.UI.Services.UrlHandler`, currently it doesn't work properly~

### Why not another way?
Please make your suggestions.
I thought about doing it in two other ways:
1.  Changing `IUrlHandler.Handle`'s return type to an `enum` that would indicate wether or not the url handing resulted in navigation to some specific view.
2. Or, the other option would be always navigate to the `MainTabBar` when handling urls no matter what, and forward the `intent` with the url data to the MainTabBar.
This way would involve changing the `IUrlHandler.Handle`, `ChangePresentation` and related methods to receive an extra view parameter.

## Review hints
Please start reviewing.
Review by commit.

## :squid: Permissions
🦑 after the Draft is done


Toggl.Droid/Presentation/AndroidUrlHandler.cs
```diff
@@ -0,0 +1,140 @@
...
+using Toggl.Core.UI.Parameters;
+using Toggl.Core.UI.Services;
+using Toggl.Core.UI.ViewModels;
+using Toggl.Shared;
+using Toggl.Shared.Extensions;
+using Toggl.Shared.Extensions.RxAction;
+
+namespace Toggl.Droid.Presentation
+{
+    public class AndroidUrlHandler : IUrlHandler
```
juanlaube: I don't like this class, there's a lot of repeated and untested code and that makes me sad.
Maybe `UrlHandler` should be abstract and let `AndroidUrlHandler` and `IosUrlHandler` do things different?

daividssilverio: But the only the droid part that needs special handling. And that isn't easily tested.

Would you have all the `UrlHandler` now private methods made `virtual` and overridden only on Android?

Effectively, what is testable would stay testable, and what's not would stay untestable.

Toggl.Droid/Startup/ExternalIntentsHandlerActivity.cs
```diff
@@ -0,0 +1,131 @@
...
+            throw new InvalidOperationException("You shouldn't be doing this from a splash screen");
+        }
+
+        public IObservable<string> GetGoogleToken()
+        {
+            throw new InvalidOperationException("You shouldn't be doing this from a splash screen");
+        }
+
+        public Task Close()
+        {
```
heytherewill: No braces for throwing

Toggl.Droid/Startup/SplashScreen.cs
```diff
@@ -94,5 +71,50 @@ private void createApplicationLifecycleObserver(IBackgroundService backgroundSer
             //mvxApplication.RegisterActivityLifecycleCallbacks(appLifecycleObserver);
             //mvxApplication.RegisterComponentCallbacks(appLifecycleObserver);
         }
-}
+
+        public IObservable<bool> Confirm(string title, string message, string confirmButtonText, string dismissButtonText)
```
heytherewill: Why are you implementing `IView` here?

daividssilverio: To call the navigation service `Navigate` using the activity reference.

heytherewill: I'd rather not use the navigation service at all. Just grab a hold of the VMLoader and start the activity manual for :sanic:

Toggl.Droid/Presentation/AndroidUrlHandler.cs
```diff
@@ -0,0 +1,195 @@
...
+                default:
+                    return await urlHandler.Handle(uri)
+                        .ContinueWith(_ => AndroidStartupHelper.StartMainTabBarActivity(activity))
+                        .ContinueWith(_ => true);
+            }
+        }
+
+        private bool handleTimeEntryNew(Dictionary<string, string> args, Activity activity)
+        {
+            // e.g: toggl://tracker/timeEntry/new?startTime="2019-04-18T09:35:47Z"&stopTime="2019-04-18T09:45:47Z"&duration=600&description="Toggl"&projectId=1&tags=[]
```
juanlaube: I'd make `UrlHandler` an open class and have virtual methods that take all the parsed arguments to minimize the repeated code.

Toggl.iOS/Startup/AppDelegate.cs
```diff
@@ -50,7 +50,16 @@ public override bool FinishedLaunching(UIApplication application, NSDictionary l
             IosDependencyContainer.EnsureInitialized(compositePresenter, environment, Platform.Daneel, version);
             var app = new App<OnboardingViewModel, Unit>(IosDependencyContainer.Instance);
 
-            app.Start();
+            var userHasAccess = app
+                .Initialize()
+                .CheckIfUserHasFullAppAccess().GetAwaiter().GetResult();
```
juanlaube: The name `CheckIfUserHasFullAppAccess` feels wrong because it has a navigation side effect.

daividssilverio:
Closed as this PR is introducing more trouble than actually bringing benefits.
We'll follow the approach that's already there with the single `SplashScreen` when the #5180 is done.
We might need to revisit this at some point if performance is a problem.

PR #5140, opened 2019-05-17
🐛 Parse workspaceId as long

khoi:
## What's this?
a bug-typo fix

### Relationships
Closes nothing.

## :squid: Permissions
🎙


Toggl.iOS/Startup/AppDelegate.cs
```diff
@@ -202,10 +202,10 @@ public override void PerformActionForShortcutItem(UIApplication application, UIA
                     var tabbarVC = (MainTabBarController)UIApplication.SharedApplication.KeyWindow.RootViewController;
                     var reportViewModel = (ReportsViewModel)tabbarVC.ViewModel.Tabs.Single(viewModel => viewModel is ReportsViewModel);
 
-                    int? parseLong (string val)
+                    long? parseLong (string val)
```
kornelijepetak: Would you mind converting this into

```cs
long? parseLong(string val) => long.TryParse(val, out long i) ? i : (long?)null;
```

PR #5141, opened 2019-05-17
🐛 Fix tasks not expanding

### LINKED ISSUE #4982: Tasks Not Selectable in TE Edit Screen
## ⛑️ User Support
Just what the title says: when creating a time entry or editing one, selecting a project works, but attempting to select a task nested in the project does not. 

NOTE: cannot be replicated on Moto g6; see user information below for specific use-case

## About the user

App version: Giskard 1.12
Phone OS: 6.0.1
Phone model: Redmi 4A

## Expected behaviour
Selecting a task from the task list should immediately attach the task to the time entry. 

## To Reproduce
1. Start a time entry
2. From the edit screen, attempt to select a task dropdown from a project that has tasks 
3. Either task list doesn't drop down, or task list inside of dropdown doesn't allow for task selection 

## Suggestions

Intercom [Case Link](https://app.intercom.io/a/apps/ayixs927/inbox/inbox/conversation/21753852670#part_id=comment-21753852670-2975203191)

Slack [convo](https://toggl.slack.com/archives/C029WU7AY/p1556131540017600) with @juanlaube  being able to replicate

kornelijepetak:
## What's this?
Fixes a bug that prevented the tasks from being expanded in some cases.
This was actually a bug that was probably happening on iOS too as it originates in Toggl.Core, not in platform code.

### Relationships
Closes #4982 

## Why do we want this?
So that tasks can be expanded as they need to.

## How is it done?
The equality calculations are wrong for the project suggestions. I have changed it to check only for ProjectID as that is the all that's needed and is unique.

## :squid: Permissions
Me.


Toggl.Core.UI/ViewModels/StartTimeEntryViewModel.cs
```diff
@@ -658,7 +658,7 @@ private IEnumerable<AutocompleteSuggestion> includeTasksIfExpanded(AutocompleteS
         private IList<SectionModel<string, AutocompleteSuggestion>> addStaticElements(IEnumerable<SectionModel<string, AutocompleteSuggestion>> sections)
         {
             var suggestions = sections.SelectMany(section => section.Items);
-            IEnumerable<SectionModel<string, AutocompleteSuggestion>> collections = sections;
+            var collections = sections;
```
simonrozsival: 🔥? This doesn't seem to be used anywhere anyway.

Toggl.Core.UI/ViewModels/StartTimeEntryViewModel.cs
```diff
@@ -79,8 +79,8 @@ public sealed class StartTimeEntryViewModel : ViewModelWithInput<StartTimeEntryP
 
         private string currentQuery;
 
-        private BehaviorSubject<HashSet<ProjectSuggestion>> expandedProjects =
-            new BehaviorSubject<HashSet<ProjectSuggestion>>(new HashSet<ProjectSuggestion>());
+        private BehaviorSubject<HashSet<long>> expandedProjects =
```
simonrozsival: Shouldn't we rather fix `AutocompleteSuggestion.Equals`?

kornelijepetak: Well, I was avoiding this for fear of breaking more than I fix, but as the matter of fact, you are right. I want to fix this properly as it was supposed to be done in the first place. 😄

juanlaube: I agree with Simon.

simonrozsival: It won't hurt to compare IDs. We don't expect changes of IDs while the user is on this screen. Please create an issue to fix the `Equals` function of the `AutocompleteSuggestion` (it has to be `virtual`).

kornelijepetak: Opened #5175 for this purpose. Resolving this conversation.

PR #5144, opened 2019-05-17
Fix bug when dismissing view controllers

### LINKED ISSUE #5142: Start and edit time entry views cannot be closed
## 🐛 Bug Report
Start and edit time entry views cannot be closed after a confirmation dialog is presented.

## Context

Build/Branch: project/mvexit
Phone OS: iOS
Phone model:

## Expected behavior
After a confirmation dialog is presented, I'd expect to be able to close the view if I want.

## To reproduce
1. Open the start or edit time entry views.
1. Make some change (e.g: set the project).
1. Tap the close button so the confirmation dialog is shown.
1. :bug: Selecting `Discard` doesn't close the view and the close button cannot be pressed again.
1. :bug: Selecting `Continue Editing` freezes the app.

juanlaube:
## What's this?
This fixes a bug when dismissing view controllers. We should call the ViewModel lifecycle methods _only_ when the the view controller is being dismissed and not for other view controllers.
I noticed this bug after an alert view controller (used for confirmation dialogs) is presented.

### Relationships

Closes #5142 

## Why do we want this?
🚫 🐛 

## How is it done?
Using `IsBeingDismissed`.

### Side effects
None.

## :squid: Permissions
🦑


Toggl.iOS/ViewControllers/Reactive/ReactiveViewController.cs
```diff
@@ -82,8 +82,11 @@ public override void DismissViewController(bool animated, Action completionHandl
         {
             base.DismissViewController(animated, completionHandler);
 
-            ViewModel?.DetachView();
-            ViewModel?.ViewDestroyed();
+            if (IsBeingDismissed)
```
heytherewill: Why not early return?

juanlaube: Sure, that's better 😄

PR #5148, opened 2019-05-19
🐛 Fix reports calendar update when beginning of week changes

### LINKED ISSUE #4996: Reports calendar does not update when beginning of week is changed
## 🐛 Bug Report
The dates in the calendar for selecting time period for reports do not update after changing the beginning of week setting. The column headers (day of week initials) do update though.

## Context

Phone OS: Android 8.0
Phone model: Nexus 5X emulator

## Expected behavior
The dates in the calendar should get offset after changing the beginning of week setting.

## To reproduce
1. Go to reports view
2. Open the reports calendar
3. Verify that the dates and headers match. For example, April 26 is shown as a friday
3. Go to settings
4. Change beginning of week to something else
5. Go back to reports
6. The calendar headers are updated, however the dates in the calendar are not, which leads to dates being displayed as the wrong day of week. For example, the same April 26 is now shown as a Monday.

kornelijepetak:
## What's this?
Ok, this one solves the issue of calendar reports not updating its dates when beginning of week option is changed.

Additionally, this also fixes the problem with invalid disposal of items that are being added and removed to the recycler.

⚠ Read `How is it done?` carefully before reviewing.

👥🔍 **Reviewers.**

I have added @heytherewill and @daividssilverio as reviewers since the original code is theirs and I want to make sure I haven't missed something important. If they'd rather not review let them unassign themselves as reviewers.

### Relationships
Closes #4996 

## How is it done?
🐛 **Fix.**

The core of the problem is that the invalid adapter was being notified about changes. The `ReportsCalendarPagerAdapter` is a pager adapter that deals with pages (in this cases, a page represents a month). For each page (month), another recyclerview and corresponding adapter are created inside `InstantiateItem()`. The type of that adapter is `ReportsCalendarRecyclerAdapter` and that one deals with items inside a month (in this case: days).

The subscription that was active on the Beginning of Week option was updating (via `NotifyDataSetChanged` call) the outer pager. However, this pager is only assigned to take care of pages, so if number/order of months doesn't get changed (and in this case it doesn't), there is no change and nothing gets updated.

So, when this happens, instead of calling `NotifyDataSetChanged`, all adapters belonging to recycler views in pages that are still in memory have to be updated. That's why I'm caching the recycler views when they are created. And also releasing them when the item is supposed to be destroyed.

♻ **Disposal.**

Another thing that has to be noted is that when adapters were being created in `InstantiateItem`, some subscriptions were being added to the dispose bag. However, as you swipe the months, some items get destroyed, some get instantiated again, and disposal used to happen only when the outer (the report calendar) adapter is being disposed. That is wrong. When no longer needed (in `DestroyItem`), the subscriptions no longer used should be disposed. Now they are.

### Side effects
Should be none.

## Review hints
Commit by commit.

## :squid: Permissions
Me.


Toggl.Droid/Adapters/ReportsCalendarPagerAdapter.cs
```diff
@@ -85,14 +121,16 @@ public void UpdateMonths(List<ReportsCalendarPageViewModel> newMonths)
...
         }
 
         protected override void Dispose(bool disposing)
         {
             base.Dispose(disposing);
             if (!disposing) return;
 
-            disposeBag.Dispose();
+            disposableBags.Values.ForEach(bag => bag.Dispose());
+            disposableBags.Clear();
```
daividssilverio: Could you also clear any leftover view references in the `recyclerViews` dictionary? It's never nice to hold on to view references.

kornelijepetak: I have cleared the references now, but, strictly speaking, this is not needed.

Since, the reports are a fragment inside our nav system, it's gonna be in the memory for the duration of the app, and the pager is holding only references to those pages around the displayed one.

And even if the system decides to scrap the fragment, would it not communicate that intent to the views so they can properly destroy themselves, which in turn would make the pager adapter call `destroyItem` on all objects, effectively removing also the references to the views as well?

daividssilverio: You are right, but I'm just being cautious. Normally, that wouldn't be necessary. But we are mixing c# and java apis here, I wouldn't be surprised if something weird happened.

PR #5160, opened 2019-05-20
Fetch just time entries

### LINKED ISSUE #5152: SyncLight: Sync state which fetches only time entries
We must create a state which fetches TEs, but none of the other entities types.

The state should be based on `FetchAllSince`. It should respect the leaky buckets we have and it should fetch the TEs based on the "since date storage" the same way we currently fetch TEs (do not use sync date older than 2 months).

The state should return a result with `FetchObservables` as parameter, so it can be processed by the already existing states for TEs.

simonrozsival:
## What's this?
This PR adds a sync state which fetches time entries and nothing else.

### Relationships
Closes #5152 

## Why do we want this?
We want to fetch just the time entries.

## How is it done?
This is a modification of already existing `FetchAllSinceState`. All the other observables in `FetchObservables` throw when subscribed to. This is because I want to reuse `PersistListState`.

### Why not another way?
This is a simple and safe way to do it.

### Side effects
A bit of code duplicity, but within reason I believe.

## :squid: Permissions
:squid:


Toggl.Core/Sync/States/PullTimeEntries/FetchJustTimeEntriesSinceState.cs
```diff
@@ -0,0 +1,90 @@
...
+using Toggl.Networking;
+using Toggl.Shared.Extensions;
+using Toggl.Shared.Models;
+using Toggl.Storage;
+
+namespace Toggl.Core.Sync.States.PullTimeEntries
+{
+    internal sealed class FetchJustTimeEntriesSinceState : ISyncState
+    {
+        private const int sinceDateLimitMonths = 2;
```
juanlaube: Maybe these should be in a constants file and used in `FetchAllSinceState` too, unless we want these to be different for another reason 🤔

simonrozsival: Yeah. I will move that somewhere else.

PR #5162, opened 2019-05-20
Fix scrolling issues on iPad

### LINKED ISSUE #5005: Scrolling issues on iPad
## 🐛 Bug Report
On the iPad, the suggestions cover up the scroll indicator, making it look weird.

It's also not possible to scroll the app by touching the margins.

## Context

Build/Branch: Daneel/1.21
Tablet OS: iOS 12.2
Tablet model: iPad 8,5

## Expected behavior
The scroll bars shouldn't be covered by any elements on the app's main screen. It should also be possible to scroll the main screen even if you tap and drag the margins to either side of the main content column.

## To reproduce
1. Run Daneel 1.21 on an iPad
2. Track time long enough for suggestions to appear under "Working on these?"
3. See the scroll bar is cut in two when you initially start scrolling. It’s fine after you go past the suggestions.
4. See that you can’t scroll by touching the margins.

Reproduced on a 2017 iPad Pro.

## Suggestions
Video available at https://photos.app.goo.gl/GqsrPVVL7QKMYtVf8

dshr:
## What's this?
A set of code and AL changes that fix the scrolling behaviour on iPad.

### Relationships
Closes #5005

## Why do we want this?
Better scrolling

## How is it done?
Before this PR we would set a max width to the table view, thus constraining its built-in scroll view, making the "margins" unresponsive. This PR changes that to give a max width to the contents of the cells themselves, thus having the scroll view span full width.

### Why not another way?
The other option is wrapping our whole table view in a separate scroll view, that will handle the scrolling, but that option is quite messy and bug-prone.

### Side effects
Tapping on the "margins" counts as selecting a cell, so it can trigger the Edit TE screen. Also, swipe-to-delete/continue also starts from the edge of the screen, and not from the edge of the table view now.

## Review hints
Run app, scroll the main log, select a TE from the centre, select a TE from the paddings, check swipe-to-delete/continue actions.

## :squid: Permissions
All yours!

### DIFF
--- a/Toggl.iOS/Cells/TimeEntriesLog/TimeEntriesLogHeaderView.cs
+++ b/Toggl.iOS/Cells/TimeEntriesLog/TimeEntriesLogHeaderView.cs
@@ -10,6 +10,7 @@ namespace Toggl.iOS.Views
 {
     public partial class TimeEntriesLogHeaderView : BaseTableHeaderFooterView<DaySummaryViewModel>
     {
+        private const double maxWidth = 834;
         public static readonly string Identifier = "timeEntryLogHeaderCell";
 
         public static readonly NSString Key = new NSString(nameof(TimeEntriesLogHeaderView));
@@ -37,6 +38,13 @@ namespace Toggl.iOS.Views
         {
             DateLabel.Text = Item.Title;
             DurationLabel.Text = Item.TotalTrackedTime;
+        }
+
+        public override void LayoutSubviews()
+        {
+            if (Superview != null)
+                ContentWidthConstraint.Constant = (nfloat)Math.Min(Superview.Bounds.Width, maxWidth);
+            base.LayoutSubviews();
         }
     }
 }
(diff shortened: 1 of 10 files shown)


sofokli1:
I think the issues should be split into 2 different issues/PRs
a) one for the scroll indicator (that could be fixed with just a change in the scroll indicator offset)
b) one for the margins not scrolling the tableView. 

a) is a problem for the main log only
b) can be argued that needs to be implemented in other views as well. Your approach is an option but the side effects seem to out weight the benefits of this approach. The delete and continue buttons in landscape are pretty far away from the cell itself and the taping on the margin (that brings up the edit view) is a bit strange UX wise.

Also regarding a) I think it is a designers decision to choose where the scroll indicator should go in general since now this brings inconsistencies with other views.

Regarding b) we could have a chat about different approaches that could be reused in other views too.

dshr:
Closing this since we're going to solve it by moving the scroll indicator.

Issue #5153, opened 2019-05-20
SyncLight: Generic state for creating placeholders


simonrozsival:
Once we have a list of TEs from the server, we must create placeholders for unknown dependencies.

The process is always the same:
1. filter the referenced entities IDs and select only those which are not already in the database
2. create a placeholder object based on the TE data (e.g., the workspace ID) and the ID of the entity
3. create records for all of the placeholder objects in the database

If we can create a generic state, which can be reused for all of the 4 types of entities which can have placeholders, it would be awesome. On the other hand, we must consider if this does not add too much complexity and if the generic approach makes sense.

simonrozsival:
I'm closing this implementation was merged into the project branch.

PR #5163, opened 2019-05-20
SyncLight: Create placeholders state

simonrozsival:
## What's this?
This PR adds basics for placeholder creating. There is the generic state and a factory which creates placeholders for projects. More placeholder types will be added in a follow-up PRs.

This state will replace the `CreateArchivedProjectPlaceholdersState` class in the future. It stays for now.

### Relationships
Ref #5153 

## Why do we want this?
Otherwise SyncLight won't work.

## :squid: Permissions
🦑


Toggl.Core/Sync/States/PullTimeEntries/CreatePlaceholdersState.cs
```diff
@@ -0,0 +1,66 @@
...
+
+namespace Toggl.Core.Sync.States.PullTimeEntries
+{
+    public sealed class CreatePlaceholdersState<TThreadSafe, TDatabase>
+        where TDatabase : IDatabaseModel, IIdentifiable
+        where TThreadSafe : TDatabase, IThreadSafeModel
+    {
+        private readonly IDataSource<TThreadSafe, TDatabase> dataSource;
+        private readonly IAnalyticsEvent<int> analyticsEvent;
+        private readonly Func<ITimeEntry, long[]> dependencyIdsSelector;
```
simonrozsival: This is because one TE can have multiple dependencies of the same time - tags.

PR #5171, opened 2019-05-20
Add restriction checking

heytherewill:
## What's this?
This adds restriction checking to the testing template

## Why do we want this?
Those views are very seldom used, but they need to be working for every release because they might be needed.

## How is it done?
I simply added it to the template. Now people need to test it before shipping

### Side effects
Currently, there's no easy way to receive a 410 or a 418 from the API. I'm putting this on hold until we figure this out.

## :squid: Permissions
Anyone, anytime.


heytherewill:
Re #5172

heytherewill:
@simonrozsival Now that #5488 is merged, these views can be easily tested :)

.github/ISSUE_TEMPLATE/d-manual-test-template.md
```diff
@@ -188,6 +188,14 @@ about: A manual testing checklist that can be assigned to a release to track the
 - [ ] Check that hint to swipe continue entry is shown
 - [ ] Check that hint to swipe delete is shown
 
+## Testing app restriction UI
+
+- [ ] Check that when the api is outdated (api returned 410 Gone) the outdated view appears
```
simonrozsival: The wording does not make it easy to understand how to force the API to be outdated. Please mention the new error triggering view and that it's OK to just tap the appropriate button.

kornelijepetak:
Resuming control in OP's absence in order to carry the PR to fruition.

kornelijepetak:
As far as I can read what's being suggested in this PR, it deals with UI for those screens, not with the way to cause such errors. The screen implemented in #5488 introduces a way to open the screens for each of the mentioned errors. However, in most of the cases it does not actually do that as a reaction to an API error, but directly instead, by navigating to the appropriate view.

Although, it also allows actual changes to the data so that the system will indeed behave as it should in the case of outdated app/api errors:
* `outdatedAppPermanentlyErrorTriggered` 
* `outdatedApiPermanentlyErrorTriggered`

That being said, I'm suggesting a change in the text so that it's obvious that it's not the API that is being tested, but UI instead.

kornelijepetak:
I've modified the text so that it's obvious that the screens are being tested and that in the case of permanent-error screens, the app needs to be updated/reinstalled.

Testing whether the app will react correctly to 410 🚪 and 428 🍵 errors should be implemented in our API. i.e. we should make a mechanism to inform the API to simulate the response. This PR and (#5488) handle only the simulation of symptoms, not the causes.

PR #5179, opened 2019-05-21
Schedule full sync after sync light

### LINKED ISSUE #5156: SyncLight: schedule full sync if we have placeholders
We need a sync state which checks if we have any placeholders in the database and if it is the case, we must store this information in the local storage of the app and whenever the app enters foreground, we should run full sync even though the app stayed in the background for less than 5 minutes.

simonrozsival:
## What's this?
This PR marks the time when we detect that any placeholders exist. This sync state is intended to be placed at the end of the sync workflow of "Sync Light" after some placeholders could have been created.

The way this PR is implemented, we will "schedule" full sync even if the placeholders weren't actually created during the last sync, but if they were created some time ago and they haven't been resolved yet. It's not a big deal. The only thing that this PR impacts is that the app will trigger Full Sync whenever the app is opened after this time was marked. If the app is opened after more than 5 minutes, the app would sync anyway.

### Relationships
Closes #5156 

## Why do we want this?
Keep the app as much up-to-date as possible while keeping the number of HTTP requests low.

## How is it done?
Sync state + interactor

### Why not another way?
This is a simple solution which is also decoupled from the actual creation of placeholders (it's not inside of #5163).

### Side effects
No side effects.

## :squid: Permissions
🦑


Toggl.Core/Sync/States/PullTimeEntries/ScheduleFollowUpSyncIfPlaceholdersWereCreatedState.cs
```diff
@@ -0,0 +1,44 @@
+using System;
+using System.Reactive.Linq;
+using Toggl.Core.Interactors;
+using Toggl.Shared;
+using Toggl.Storage.Settings;
+
+namespace Toggl.Core.Sync.States.PullTimeEntries
+{
+    public sealed class ScheduleFollowUpSyncIfPlaceholdersWereCreatedState : ISyncState
```
juanlaube: The name is a bit confusing because it doesn't schedule anything, maybe `DetectPlaceholdersWereCreatedState` or something similar is better.

simonrozsival: Good point.

PR #5184, opened 2019-05-21
Use native startup on platforms

heytherewill:
## What's this?
We are no longer relying on the navigation service for the first navigation when the user is logged in.

### Relationships
Ref #4867

## :squid: Permissions
Anyone, anytime


Toggl.iOS/Startup/AppDelegate.cs
```diff
@@ -49,8 +49,13 @@ public override bool FinishedLaunching(UIApplication application, NSDictionary l
 
             IosDependencyContainer.EnsureInitialized(compositePresenter, environment, Platform.Daneel, version);
             var app = new App<OnboardingViewModel, Unit>(IosDependencyContainer.Instance);
-
-            app.Start();
+            var hasFullAccess = app.NavigateIfUserDoesNotHaveFullAccess();
+            if (hasFullAccess)
+            {
+                var loader = new ViewModelLoader(IosDependencyContainer.Instance);
```
juanlaube: 💡 Why don't we put the vm loader in the DI container? I actually was thinking of doing this in the deeplink refactor for example, it's better than creating new loaders everywhere.

Issue #5180, opened 2019-05-21
Refactor UrlHandler


juanlaube:
## 🚧 Refactor Proposal
We will decouple the url parsing from the handling.

## Platforms affected:
- [x] 🤖 Android
- [x] 🍎 iOS
- [x] 🖥️ Shared layer

## Technical aspects
Remove `UrlHandler` and introduce `IUrlParser` with just one method, `Parse(Uri uri)` that returns a `DeeplinkParameters` (it's a `OneOfMany`) struct that can be matched on the platform side.
Because some urls are handled identically on all platforms, we will have extensions methods on this `DeeplinkParameters` to handle the actions.

PR #5185, opened 2019-05-21
Deeplink refactor

juanlaube:
## What's this?
This decouples the deeplink parsing from actually handling each url.

### Relationships

Refs #5180, doesn't close because needs some 🤖 bits.

## Why do we want this?
Deeplinking is completely different on iOS and Android, this moves the handling to the platform code and we have extension methods for everything we can share.

## :squid: Permissions
Only mvexit bois.


Toggl.iOS/Startup/AppDelegate.Deeplink.cs
```diff
@@ -0,0 +1,94 @@
...
+        private async Task showNewTimeEntry(DeeplinkParameters.TimeEntryParameters timeEntry, ITimeService timeService, INavigationService navigationService)
+        {
+            var workspaceId = timeEntry.WorkspaceId;
+            var startTime = timeEntry.StartTime ?? timeService.CurrentDateTime;
+            var duration = timeEntry.Duration;
+            var description = timeEntry.Description;
+            var projectId = timeEntry.ProjectId;
+            var tags = timeEntry.TagIds;
+
+            var startTimeEntryParameters = new StartTimeEntryParameters(
```
daividssilverio: Can we add another extension method to convert the
`DeeplinkParameters.TimeEntryParameters` into `StartTimeEntryParameters`?

juanlaube: Done!

Toggl.Core.UI/Parameters/DeeplinkParameters.cs
```diff
@@ -0,0 +1,129 @@
...
+                    matchEditTimeEntry(editTimeEntryParameters);
+                    return;
+                case DeeplinkShowReportsParameters showReportsParameters:
+                    matchShowReports(showReportsParameters);
+                    return;
+                case DeeplinkShowCalendarParameters showCalendarParameters:
+                    matchShowCalendar(showCalendarParameters);
+                    return;
+                default:
+                    matchOther(this);
```
daividssilverio: Won't this crash by default? The default is null. The iOS implementation isn't passing any value for this.

Toggl.Core.UI/Parameters/DeeplinkParameters.cs
```diff
@@ -120,9 +120,11 @@ public static DeeplinkParameters WithCalendar(string eventId)
...
                     matchShowCalendar(showCalendarParameters);
                     return;
-                default:
-                    matchOther(this);
-                    return;
+            }
+
+            if (matchOther != null)
+            {
+                matchOther(this);
```
heytherewill: `matchOther?.Invoke(this)`

PR #5186, opened 2019-05-21
Make app access check faster

### LINKED ISSUE #5183: Make faster LoggedIn checks
We can improve the fresh startup times by simply using the key-value storage instead of realm in the `App.Start`.

daividssilverio:
## What's this?
This makes the authentication check a bit faster.

### Relationships
Closes #5183 

## Why do we want this?
Because realm is slow in this case. 
Checking the private storage in the startup speeds up about 600+ ms on droid.

https://github.com/toggl/mobileapp/blob/ec61ab3b557047aa2b8eae9a47a6fa990fe96321/Toggl.Core.UI/App.cs#L41
If you put a Stopwatch there ☝️ , it takes about 400 ms on a Samsung S8

https://github.com/toggl/mobileapp/blob/ec61ab3b557047aa2b8eae9a47a6fa990fe96321/Toggl.Core.UI/App.cs#L47
Takes about 300 ms on a Samsung S8 as well.

Using the private storage, the cases above takes about 15 ms and 90 ms respectively.

## How is it done?
This simply makes use of the existing `IPrivateSharedStorageService` logic to also check in the private storage for the user API token to prevent unauthenticated access.

### Why not another way?
I could add another class just for it, but it made sense to simply implement the what was `Noop` on droid and take profit.

## :squid: Permissions
🦑 if you think it's worth it. But I've based this branch on #5184 , so that has to be merged before this branch.


heytherewill:
Please don't merge before I review :)

PR #5191, opened 2019-05-22
Fix Reports flicker

### LINKED ISSUE #4879: Stop Reports from flickering when opening tab
When I open the reports tab on Daneel the report for the first time, it remains empty at first and then the report fades in. When I navigate somewhere else and back it displayed flickers and seems to reload.

Several problems with that:

- Most importantly, it should definitely not flicker. It should always smoothly animate (fade) between states.
- Secondly, we have an empty/loading view which should be showing when it's loading the report instead of an empty screen
- Thirdly, it should not reload the same report when I navigate back and forth.

On the last point, I propose to only reload if the view has not been seen in at least five minutes (this state can be kept in the view so that if it gets cleaned up for whatever reason it'll load again next time you open it).

I thought this was already part of the original designs, but it was either not communicated properly, or it was implemented wrongly.

dshr:
## What's this?
A tiny refactor to prevent the reports view from flickering every time we open it.

### Relationships
Closes #4879 

## Why do we want this?
Better experience

## How is it done?
By creating a new subject for `ViewAppeared`, which is then throttled and used to trigger the reload action.

### Why not another way?
Seems simple enough, and rx-ey too.

### Side effects
No obvious way to manually refresh the reports page (can add a pull-to-refresh if needed)

## Review hints
Just test if the flickers.

## :squid: Permissions
Go fer it

### DIFF
--- a/Toggl.Core.UI/ViewModels/Reports/ReportsViewModel.cs
+++ b/Toggl.Core.UI/ViewModels/Reports/ReportsViewModel.cs
@@ -214,11 +214,8 @@ namespace Toggl.Core.UI.ViewModels.Reports
 
             viewAppearedSubject
                 .AsObservable()
-                .Throttle(new TimeSpan(0, 5, 0))
-                .Subscribe((_ =>
-                {
-                    reportSubject.OnNext(Unit.Default);
-                }))
+                .Throttle(TimeSpan.FromMinutes(5))
+                .Subscribe(_ => reloadReportsWhenViewAppears())
                 .DisposedBy(disposeBag);
 
             reportSubject
@@ -309,6 +306,11 @@ namespace Toggl.Core.UI.ViewModels.Reports
             isLoading.OnNext(false);
 
             trackReportsEvent(true);
+        }
+
+        private void reloadReportsWhenViewAppears()
+        {
+            reportSubject.OnNext(Unit.Default);
         }
 
         private void onError(Exception ex)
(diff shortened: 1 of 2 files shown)


Toggl.Core.UI/ViewModels/Reports/ReportsViewModel.cs
```diff
@@ -211,6 +212,15 @@ public override async Task Initialize()
                 .Subscribe(changeDateRange)
                 .DisposedBy(disposeBag);
 
+            viewAppearedSubject
+                .AsObservable()
+                .Throttle(new TimeSpan(0, 5, 0))
```
simonrozsival: Using `TimeSpan.FromMinutes(5)` would be much more understandable here.

dshr:
> We will reload the data at most once every 5 minutes? If I understand it correctly, then I don't think this is a good solution.

Yup, that's the idea, as per a suggestion in the original issue: "On the last point, I propose to only reload if the view has not been seen in at least five minutes (this state can be kept in the view so that if it gets cleaned up for whatever reason it'll load again next time you open it)."

simonrozsival:
I think the real problem with the flicker is:
- when you open the tab, the old state is shown
- we hide the old state as soon as we start loading the new data
- the new data is shown and it possibly shows the same that that was shown before (if nothing changed in the meantime, which we can't guarantee, something can always change, a time entry can be stopped...)

I think the proper solution would be to always show the empty/loading state when you open the app instead of the old state. There should be animations which fade from the loading state to the data states. If these animations are broken, they should be fixed as well.

arturs105:
The flicker is indeed fixed, however the loading animations are not shown. I tried applying the fix from my comment in the issue, but it brought back the flicker.

dshr:
@simonrozsival triggering a reload every time we open a tab feels quite off UX-wise, since we normally expect them to already have all the info required. Maybe we can trigger the reload in the background when the user leaves the tab?

sofokli1:
Is it possible to not show the empty state while reloading, just keep showing what was already loaded?

dshr:
Closing this since it's going to get fixed in MVExit

PR #5192, opened 2019-05-22
Re-re-add droid url nav 

### LINKED ISSUE #4867: Implement Url Navigation for Android

daividssilverio:
## What's this?
This re-adds the droid url & deeplink handling.

### Relationships
Closes #4867 
Supersedes #5135
Based off the branch from #5185 

## Why do we want this?
To support url navigation & deeplinks.

## How is it done?
It uses the url parser and handles the initial navigation from the SplashScreen.

Why draft? What still needs to be done?
- [x] ApplicationUrls.TimeEntry.New.Path
- [x] ApplicationUrls.TimeEntry.Edit.Path
- [x] ApplicationUrls.Reports.Path
- [x] ApplicationUrls.Calendar.Path 

### Why not another way?
The stuff from #5135 wasn't working well. I'm trying to keep it simple. It might not be :sanic: but it should work.

## Review hints
Please start reviewing.
Review by commit.

## :squid: Permissions
🦑 after the Draft is done & #5185 is merged


Toggl.Droid/Startup/SplashScreen.DeepLink.cs
```diff
@@ -0,0 +1,130 @@
...
+                te => RunAndStartRootActivity(te.Stop(interactorFactory, timeService)),
+                te => RunAndStartRootActivity(te.Create(interactorFactory)),
+                te => RunAndStartRootActivity(te.Update(interactorFactory, timeService)),
+                showNewTimeEntry,
+                showEditTimeEntry,
+                showReports,
+                showCalendar);
+        }
+
+        private void RunAndStartRootActivity(Task task)
```
heytherewill: private methods get a lowercase name

heytherewill: Are you even using that Task? I mean, I get that the code gets executed but that's a weird design choice

Toggl.Core.UI/ViewModels/Settings/SettingsViewModel.cs
```diff
@@ -491,7 +491,7 @@ private async Task selectBeginningOfWeek()
 
         private async Task checkCalendarPermissions()
         {
-            var authorized = await permissionsChecker.CalendarPermissionGranted;
+            var authorized = permissionsChecker.CalendarPermissionGranted.FirstAsync().GetAwaiter().GetResult();
```
heytherewill: No me gusta

Toggl.Droid/Startup/SplashScreen.DeepLink.cs
daividssilverio: I'm not using the task, but the other options I see is to wait for it using `ContinueWith` or run the login in this method after inline after each `te.Action`.

I don't want to wait for the action finish to start the main tab bar activity and finish the splash.

Suggestions?

Toggl.Core.UI/ViewModels/Settings/SettingsViewModel.cs
daividssilverio: No me gusta as well, couldn't make it work otherwise.

The scenario is super cryptic, but it wasn't working for me without it.
1. Start the app once through the launcher (you can debug to see it, or add a few logs, it would execute just fine with the old implementation for the first time)
2. Press home
3. Click on any deeplink -> it freezes and statement `await permissionsChecker.CalendarPermissionGranted` never returns.

What's the difference on 3 you might ask: The dependency container has already been initialized.

heytherewill: Approving for :sanic: but opening an issue.

Issue #5136, opened 2019-05-16
Display placeholder bars instead of "(unsynced project/task)"


simonrozsival:
Instead of this:
![image](https://user-images.githubusercontent.com/374616/57846989-b4688980-77d5-11e9-91d4-7360c8316923.png)

We will show gray bars take from the "empty main log" placeholders in the place of a placeholder project and placeholder task
- [x] in the main log cell
- [x] in the currently running time entry card

![iOS](https://user-images.githubusercontent.com/374616/57846898-8420eb00-77d5-11e9-92c6-cd2acd5cd875.png)
![Android](https://user-images.githubusercontent.com/374616/59102225-c6000580-892b-11e9-9d3a-15acc729e285.png)

simonrozsival:
Screenshot from the android design:
![image](https://user-images.githubusercontent.com/374616/59102225-c6000580-892b-11e9-9d3a-15acc729e285.png)

simonrozsival:
The final decision (confirmed by @maritoggl): We will display the placeholder bars in the main log and for the currently running time entry. Everywhere else (e.g., the edit view), we will simply show the name which says "(unsynced project/task/workspace/tag)". The color of the placeholders should be always gray.

daividssilverio:
I'm working on the droid bits.

PR #5193, opened 2019-05-23
SyncLight transitions

### LINKED ISSUE #5155: SyncLight: Set up transitions correctly
The order of creating placeholders matters.

The order should be (roughly) as follows:

1. Fetch just time entries
2. Ensure the HTTP request succeeded
3. Create placeholders for workspace
4. Create placeholders for projects
5. Create placeholders for tasks
6. Create placeholders for tags
7. Persist time entries
8. Remember to run full sync as soon as the app enters foreground if there are any placeholders in the database

simonrozsival:
## What's this?
This is the final functional piece of SyncLight. This PR adds placeholder builders and sets up transitions.

After this PR, the sync tests from #5157 should finally pass.

### Relationships
Closes #5155 

## Why do we want this?
We want SyncLight.

## How is it done?
It was necessary to add placeholder creation for tasks and update the `CreatePlaceholdersStateFactory`. Then it's just a matter of setting up transitions to match the specs from Notion. The syncing tests had to be fixed as well, because there were errors in them.

### Why not another way?
I could split this into 2 small PRs, but it would require 2 reviews, which would make it more work for reviewers and I wanted to avoid that.

### Side effects
The placeholders which we create aren't perfect. The strings used as the names will be moved to `Resources.resx` in a separate PR. We also still have to sort out #5136 so this will be definitely revisited in the future.

## Tests
✅ Syncing tests: https://app.bitrise.io/build/96514b95a4e52d23 (commit `8154ffc`) - do not merge this branch unless these tests are passing.

## :squid: Permissions
🦑 if sync tests pass.


simonrozsival:
This PR is now waiting for #5194

simonrozsival:
The transitions look like this: 
image

Full sync graph can be downloaded from here: [sync-graph.svg.zip](https://github.com/toggl/mobileapp/files/3216378/sync-graph.svg.zip)

simonrozsival:
I noticed I forgot to use the state which logs the analytics events from #5156

Toggl.Core/Sync/StateMachineOrchestrator.cs
```diff
@@ -45,6 +45,9 @@ public void Start(SyncState state)
                 case SyncState.CleanUp:
                     startSync(SyncState.CleanUp, entryPoints.StartCleanUp);
                     break;
+                case SyncState.PullTimeEntries:
+                    startSync(SyncState.CleanUp, entryPoints.StartPullTimeEntries);
```
sofokli1: shouldn't the newState param be `PullTimeEntries` instead of `CleanUp`

simonrozsival: Oh, good catch.

PR #5194, opened 2019-05-23
SyncLight: placeholders

simonrozsival:
## What's this?
Factories for placeholders. It is needed for #5193. I created this PR as a separate PR after all to make it easier to review.

### Relationships
Ref #5155 

## Why do we want this?
We need to create placeholder entities for workspaces, projects, tasks, and tags.

## :squid: Permissions
🦑


Toggl.Core/Models/Workspace.cs
```diff
@@ -47,6 +49,14 @@ public Builder SetAt(DateTimeOffset at)
             }
         }
 
+        public static Workspace CreatePlaceholder(long id, ITimeEntry timeEntry)
```
amulware: I observe that the time entry isn't used.

Toggl.Core/Models/Tag.cs
```diff
@@ -77,6 +79,15 @@ private void ensureValidity()
             }
         }
 
+        public static Tag CreatePlaceholder(long id, ITimeEntry timeEntry)
```
amulware: Not sure I like that a time entry is passed in here, it's not clear why it would need that, and it doesn't really.

Toggl.Core/Models/Workspace.cs
simonrozsival: It isn't, but it's something like a "static interface" - it will be used like this: https://github.com/toggl/mobileapp/pull/5193/files#diff-033a28cd4e9b9c917c1ed9d5e16a1d4cR31

I can change it and add a lambda, which adapts the signature to the simplified signature, but I myself think that it's OK for now to leave it like this. If you're not happy with this explanation, please request changes 😄

Toggl.Core/Models/Tag.cs
simonrozsival: It's used on line 86.

amulware: Yes, but the time entry is not actually the important data. This looks to me like the function is written to be convenient for who's calling it, which doesn't seem right. The tag doesn't care about time entries and shouldn't even know what a time entry is.

Toggl.Core/Models/Workspace.cs
amulware: Mostly see my other comment on this.

Toggl.Core/Models/Tag.cs
simonrozsival: I get what you mean. I think I'll just pass the workspace ID directly. That seems cleaner.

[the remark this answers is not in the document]
simonrozsival: 🤔 there's no particular reason. It should be `internal`. I'll change it.

PR #5201, opened 2019-05-23
📐 Tweak some UI to look normal on large font/display settings

### LINKED ISSUE #5108: Reports labels are overlapping on smaller screens
On smaller screen sizes with larger text/font settings, the date labels are overlapping. They should instead be adjusted depending on the size of the screen and available real estate.

image

### LINKED ISSUE #5109: Adjust Start TE screen UI to be nice to smaller screens and large text size
The image below shows two different issues with smaller phones.

1. (RED LINE) Long text is wrapping and is cut off (the cell size is smaller than the height of two rows). The text should not wrap anyway. Turn the wrapping off.

2. (YELLOW BLOCK) The duration label is cut off. Let this label expand as much as it needs, position the remaining elements to the right of it, regardless of its width.

image

kornelijepetak:
## What's this?
Improves some 🤖 UI on larger font/display sizes.
Most of the changes have no effect at all for normal font/display sizes.

* Reports screen: header now resizes as needed
* Reports screen: Legend below the barchart now displays scaled labels
* Start TE screen: The duration below now resizes with content so it's not cut off
* Start TE screen: Height of the row of 56dp will be used if more is not needed (most of the user's settings scenarios). On very large display/font sizes, the row will resize accordingly so that no text is cut off.

### Relationships
Closes #5108 
Closes #5109

## Why do we want this?
So that even users with large fonts/display settings can reasonably use Toggl.

## How is it done?
Some of the issues were solved by setting the current values as minimum, and let the content resize itself accordingly to accustom larger UI elements on large font/display size.

### Why not another way?
I know of no other ways.

### Side effects
Should be none, but please do thoroughly functionally test both of the mentioned screens.

## Review hints
Commit by commit.

# :squid: Permissions
Me.


Toggl.Droid/Views/BarChartView.cs
```diff
@@ -206,20 +209,27 @@ protected override void OnDraw(Canvas canvas)
...
                     var middleOfTheBar = left + (barRight - left) / 2f;
-                    canvas.DrawText(horizontalLabels[barIndex].DayOfWeek, middleOfTheBar, dayLabelsY, othersPaint);
-                    canvas.DrawText(horizontalLabels[barIndex].Date, middleOfTheBar, dayLabelsY + textSize, othersPaint);
+                    var dayOfWeekText = horizontalLabels[barIndex].DayOfWeek;
+                    othersPaint.TextSize = originalTextSize;
+                    canvas.DrawText(dayOfWeekText, middleOfTheBar, dayLabelsY, othersPaint);
+
+                    var dateText = horizontalLabels[barIndex].Date;
+                    setTextSizeFromWidth(dateText, othersPaint,  othersPaint.TextSize, actualBarWidth);
+                    var bounds = new Rect();
```
daividssilverio: Please create a field for this `Rect` and re-use it. Creating new instances of `Rect`, `Point` onDraw indiscriminately can make the app consume more memory than it should and trigger gc more often, making the app freeze and jank.

Toggl.Droid/Views/BarChartView.cs
```diff
@@ -246,5 +256,24 @@ private float calculateNonBillableTop(float billableTop, float nonBillableBarHei
...
                 : nonBillableTop;
         }
+
+        private static void setTextSizeFromWidth(string text, Paint paint, float originalTextSize, float maxWidth)
+        {
+            // 48f is enough for reasonable resolution (older phones could have issues with memory)
+            const float sampleTextSize = 48f;
+
+            paint.TextSize = originalTextSize;
+            var bounds = new Rect();
```
daividssilverio: Same thing here.

Toggl.Droid/Views/BarChartView.cs
```diff
@@ -246,5 +256,24 @@ private float calculateNonBillableTop(float billableTop, float nonBillableBarHei
                 ? nonBillableTop - minHeightForBarsWithPercentages
                 : nonBillableTop;
         }
+
+        private static void setTextSizeFromWidth(string text, Paint paint, float originalTextSize, float maxWidth)
+        {
+            // 48f is enough for reasonable resolution (older phones could have issues with memory)
+            const float sampleTextSize = 48f;
```
daividssilverio: What unit of measurement is this? 48 pixels?

Toggl.Droid/Views/BarChartView.cs
```diff
@@ -246,5 +256,24 @@ private float calculateNonBillableTop(float billableTop, float nonBillableBarHei
...
+
+            paint.TextSize = originalTextSize;
+            var bounds = new Rect();
+            paint.GetTextBounds(text, 0, text.Length, bounds);
+
+            if (bounds.Width() > maxWidth)
+            {
+                paint.TextSize = sampleTextSize;
+                bounds = new Rect();
```
daividssilverio: Same thing here, reuse your rects.

Toggl.Droid/Views/BarChartView.cs
kornelijepetak: Yes, pixel units.

This SO question is from 2011 so it may not be relevant on today's phones, but I'd rather not risk it.

https://stackoverflow.com/questions/6253528/font-size-too-large-to-fit-in-cache

daividssilverio: Good damn, if that still happens we are in trouble.

kornelijepetak:
@daividssilverio It's not really intended, but, honestly, I'd rather not mess up with this as I've already lost enough time on this the last time. The point is that I couldn't set the `scaleType` in such a way so that if the image should be resized, aligns to the bottom of the card as it should. It either zooms in (cutting off the image) or it scales it and moves it upwards.

Since this is happening only on the largest of text/display size, I'd leave this as is.

PR #5215, opened 2019-05-27
Fix UI tests for iOS ✨

arturs105:
## What's this?
Make UI tests great again. 🎊 They are now working on iOS again.

## How is it done?
By updating `NUnit` and `Xamarin.UITest` packages to the latest versions. The `IgnoreAttribute` in the latest `NUnit` required a `string reason` in the constructor, so i added that as well.

### Why not another way?
This works

## Review hints
Check to see if you can get iOS UI tests to ruin

## :squid: Permissions
🦑


heytherewill:
Two questions:

- Is it possible to configure the app to actually build the app and run it instead of relying on a hardcoded path?
- Can we get rid of the hack that forces the simulator to be reset now that we are using NUnit 3?

juanlaube:
> Can we get rid of the hack that forces the simulator to be reset now that we are using NUnit 3?
I tried, we can't 😞

arturs105:
> Is it possible to configure the app to actually build the app and run it instead of relying on a hardcoded path?

Don't think so

juanlaube:
I don't know why I didn't realize before, but I think we can do that.

```
var app = ConfigureApp.iOS
                .InstalledApp("com.toggl.daneel")
                .StartApp();
```

PR #5225, opened 2019-05-27
Project: Push Notifications

simonrozsival:
## What's this?
This is the PN project PR: https://github.com/toggl/mobileapp/projects/60

## Why do we want this?
We want our apps to be synced with the server all the time.

## :squid: Permissions
🚫


Toggl.iOS/Services/PushNotificationsTokenServiceIos.cs
```diff
@@ -0,0 +1,29 @@
...
+{
+    public sealed class PushNotificationsTokenServiceIos : IPushNotificationsTokenService
+    {
+        public PushNotificationsToken? Token => getToken();
+
+        public void InvalidateCurrentToken()
+        {
+            InstanceId.SharedInstance.DeleteId(error =>
+            {
+                Console.WriteLine("Error deleting instance");
```
heytherewill: 👀

Toggl.iOS/Startup/AppDelegate.Notifications.cs
```diff
@@ -0,0 +1,66 @@
+using System;
```
heytherewill: Shouldn't this whole file be called `.PushNotifications` instead?

Toggl.iOS/Startup/AppDelegate.cs
```diff
@@ -8,23 +8,31 @@
 using Toggl.Networking;
 using UIKit;
 using UserNotifications;
+using Firebase.CloudMessaging;
 
 namespace Toggl.iOS
 {
     [Register(nameof(AppDelegate))]
-    public sealed partial class AppDelegate : UIApplicationDelegate, IUNUserNotificationCenterDelegate
+    public sealed partial class AppDelegate : UIApplicationDelegate, IUNUserNotificationCenterDelegate, IMessagingDelegate
```
heytherewill: Shouldn't the `IMessagingDelegate` be put in the file that implements it?

Toggl.Core/Interactors/PushNotifications/SubscribeToPushNotificationsInteractor.cs
```diff
@@ -0,0 +1,72 @@
...
+                return Observable.Return(Unit.Default);
+
+            var previouslyRegisteredToken = keyValueStorage.GetString(PreviouslyRegisteredTokenKey);
+            var dateOfRegisteringTheToken = keyValueStorage.GetDateTimeOffset(DateOfRegisteringPreviousTokenKey);
+
+            if (!string.IsNullOrEmpty(previouslyRegisteredToken)
+                && previouslyRegisteredToken == token.ToString()
+                && dateOfRegisteringTheToken.HasValue
+                && !shouldReSubmitToken(dateOfRegisteringTheToken.Value))
+            {
```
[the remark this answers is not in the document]
simonrozsival: I disagree. In the case of a condition spread over several lines, it's harder to spot where the condition ends and where the body of the if is. I'd like to keep it like this.

simonrozsival:
There are now two (minor) conflicts between this branch and `develop`. They will be resolved with #5344

PR #5231, opened 2019-05-28
Make sure ReportsCalendarVM has appeared before ReportsVM

### LINKED ISSUE #5197: Reports: bar chart empty state
## 🐛 Bug Report
The bar chart in reports doesn't show the empty state when there's no data, instead of showing a bar with 0s clocked, the view is empty (only the chart legend is visible).

## Context

Build/Branch: project/mvexit
Phone OS: Android
Phone model:

## Expected behavior
Should show the chart like in `develop`

## To reproduce
Load a report for a date range without time entries.

daividssilverio:
## What's this?
This simply makes sure the `ReportsCalendarViewModel`'s `ViewAppeared` callback is called before the `ReportsViewModel`'s `ViewAppeared`.

### Relationships
Closes #5197

## Why do we want this?
Oh boy... `ReportsViewModel` and `ReportsCalendarViewModel` need to be refactored.
The thing is, when `ReportsViewModel`:`reportSubject` `OnNext` is called on `ViewAppeared` for the first time, `ReportsViewModel`:`startDate` and `endDate` are still not initialized, because the child `ReportsCalendarViewModel`'s `ViewAppeared` is called after the `ReportsViewModel`'s`ViewAppeared`... but only on Android, because iOS wasn't even calling `ViewAppeared` and was relying on `SelectInitialShortcut` instead.

## How is it done?
This PR simply checks if the `startDate` was initialized during `ReportsViewModel`:`ViewAppeared`, calling the `ReportsCalendarViewModel`:`ViewAppeared` in case it didn't.

### Why not another way?
Refactoring this view is kind of out of the scope for Mvexit.
But... we could stop relying on the views' lifecycle to initialized anything when we do refactor this vms combo.

### Side effects
There shouldn't be any.

## Review hints
:trollface: review by commit

## :squid: Permissions
🦑 after some careful thought


Toggl.Core.UI/ViewModels/Reports/ReportsViewModel.cs
```diff
@@ -232,7 +230,10 @@ public override void ViewAppeared()
 
             intentDonationService.DonateShowReport();
 
-            reportSubject.OnNext(Unit.Default);
+            if (startDate == default(DateTimeOffset))
```
simonrozsival: Let's give this a name - like `bool isShownForTheFirstTime() => startDate == default(DateTimeOffset);`. It could be a private member function, a local function, or just a variable, but let's make it self-documenting. Now it's like "WTF? Why do we call the lifecycle method only sometimes?".

PR #5232, opened 2019-05-28
🐛 fix crash when user doesn't have access to a project

### LINKED ISSUE #5220: Shortcuts screen crashes when changing account
## Context

Build/Branch: `project/siri-newfeatures`
Phone OS: iOS
Phone model: Any

## Expected behavior
The app functions properly

## To reproduce
- Create a shortcut with a project associated with it.
- Log out
- Login to another account that doesn't have access to the project
- Settings > Shortcuts 
- 💥

## Suggestions
It seems like we're not removing the created shortcuts when users log out.

khoi:
## What's this?
As the title says

### Relationships
Closes #5220

## How is it done?
The code is self-explanatory

### Side effects
None.

## Review hints
None.

## :squid: Permissions
🎙


Toggl.Core.UI/ViewModels/Settings/Siri/SiriShortcutsViewModel.cs
```diff
@@ -42,6 +42,6 @@ public sealed class SiriShortcutsViewModel : ViewModel
             navigationService.Navigate<SiriShortcutsCustomTimeEntryViewModel>();
 
         public IObservable<IThreadSafeProject> GetProject(long projectId)
-            => interactorFactory.GetProjectById(projectId).Execute();
+            => interactorFactory.GetProjectById(projectId).Execute().Catch(Observable.Empty<IThreadSafeProject>());
```
kornelijepetak: I'm maybe missing some more context, but isn't it bad to just swallow all the errors and act in that case as if there are no projects with that id?

Odrakir:
@khoi I think this is not the correct approach. It's probably better to just check the intent's workspace agains all the user's workspaces. If it's not there, then we don't show the shortcut.

Let me do a PR.

PR #5235, opened 2019-05-28
Track the creation of unrepresentable date times

arturs105:
## What's this?
In daneel 1.21 and 1.21.1 we've been experiencing crashes due to weird date times (Jan 1 0001). Backend reported that the minimum start time of a time entry is Jan 8 0001, which wouldn't cause any crashes. So now we know that we can't get "bad" date times externally. That means that Jan 1 0001 dates are created internally in our app either by users doing that manually (unlikely) or by a bug in our code.

The best way to find out how and where these date times come from, is to track the creation of those and the obvious place to do that is the gateway to the local time entry storage - the `TimeEntriesDataSource`. 

## How is it done?
By checking the start and end times of a time entry, whenever it is created, updated or batch-updated. The start time and end time of each time entry is checked. However, since we want to track the start and end times of time entries _becoming_ out of range, we also need to check the value before the update.

## :squid: Permissions
🦑


arturs105:
Why do you think that? Does it say anywhere that the stack traces are trimmed? Because this is the same way App Center gets stack traces for any exception - from `Exception.StackTrace` property, which is `string`. I had to override it to make it assignable, since it was read-only.

[Here's](https://appcenter.ms/orgs/Toggl/apps/Toggl-iOS/crashes/errors/481846876/reports/2518432619559279999-52fa86ab-bd3b-4f3a-84f0-415f94395e96/events) an exception that has a 555 character stack trace.

Issue #5033, opened 2019-05-06
Implement API: Subscribe and Unsubscribe from push services


simonrozsival:
The server needs to know token of the device which we obtained from #5030 and #5031 in order to send any PNs to it.

We must wait for the backend team to create the endpoint specification before we can start working on this issue.

simonrozsival:
Implementation was merged into the project branch.

PR #5238, opened 2019-05-28
Push Notifications: API integration mock

simonrozsival:
## What's this?
This PR adds the API token subscription/unsubscription requests mocks. It will allow us to implement our side of the app before the BE team implements the endpoints.

### Relationships

Ref #5033 

## Why do we want this?
:sanic:

## How is it done?
The endpoints return `Unit` observable for now and they don't make any actual HTTP requests.

### Side effects
None.

## :squid: Permissions
🦑


Toggl.Networking/ApiClients/PushServicesApi.cs
```diff
@@ -0,0 +1,34 @@
...
+            Endpoints endPoints,
+            IApiClient apiClient,
+            IJsonSerializer serializer,
+            Credentials credentials)
+            : base(apiClient, serializer, credentials, endPoints.LoggedIn)
+        {
+            this.endPoints = endPoints.PushServices;
+        }
+
+        public IObservable<Unit> Subscribe(PushNotificationsToken token)
```
heytherewill: Shouldn't this be a boolean to indicate success/failure?

Toggl.Shared/PushNotificationsToken.cs
```diff
@@ -0,0 +1,17 @@
+namespace Toggl.Shared
+{
+    public struct PushNotificationsToken
+    {
+        private readonly string token;
+
+        public PushNotificationsToken(string token)
+        {
+            Ensure.Argument.IsNotNullOrEmpty(token, nameof(token));
```
heytherewill: Do we know what the token looks like? Some validation here would be nice

simonrozsival: The token is whatever the library gives us, it's not user input, I don't want to have any validation here.

Toggl.Networking/ApiClients/PushServicesApi.cs
simonrozsival: The observable will throw if the HTTP request fails.

PR #5239, opened 2019-05-28
🔎 Siri: Filter shortcuts by workspace

### LINKED ISSUE #5220: Shortcuts screen crashes when changing account

Odrakir:
## What's this?
In order to not show shortcuts from other users which could crash the app, we filter them by workspace.

### Relationships
Closes #5220 Supersedes #5232 

## How is it done?
I just get the user workspaces and filter each shortcut workspace (if any) agains them.

## :squid: Permissions
Go!


Odrakir:
@simonrozsival Pointed out two users could be in the same workspace but not have access to the same project, so I'm doing a second filtering for projects.

I'm also simplifying the filtering itself, as it can be done with the shortcut parameters.

PR #5240, opened 2019-05-28
Don't await on stop interactor to avoid app freezing

### LINKED ISSUE #5221: App freezes after resuming from background
## 🐛 Bug Report
Reported by @amulware. When resuming from background after a few minutes (> 10 minutes) and pressing the stop button, sometimes the app freezes and nothing responds after.

## Context

Build/Branch: `develop`
Phone OS: iOS
Phone model: Any

## Expected behavior
The app shouldn't freeze

## To reproduce
1. Open `AutomaticSyncingService.cs`
1. Edit `MinimumTimeInBackgroundForFullSync` to `TimeSpan.FromSeconds(5)`
1. Start a time entry 
1. Go to background
1. Resume after 10 seconds
1. Press the stop button
1. :bug: Sometimes the app completely freezes

juanlaube:
## What's this?
This is a possible fix to the app freezing problem when pressing the stop button.

### Relationships

Closes #5221 

## Why do we want this?
🚫 ⛄️ 

## How is it done?
Instead of `await`ing, just returning `IObservable<Unit>`.

### Why not another way?
🤷‍♂ 

### Side effects
❓dunno❓ 

## Review hints
I need 👀 from experts on Rx and async execution in C# (god I miss grand central dispatch 😄)

## :squid: Permissions
🦑


heytherewill:
I'm afraid this does not fix the problem :/

heytherewill:
Closing as this doesn't really fix the issue

PR #5242, opened 2019-05-28
Make TaskCompletionSource readonly

heytherewill:
## Why do we want this?
A bit more safety

## How is it done?
`{ get; }`

## :squid: Permissions
Anyone, anytime


Toggl.Core.UI/ViewModels/Base/ViewModel.cs
```diff
@@ -12,12 +12,14 @@ public abstract class ViewModel<TInput, TOutput> : IViewModel
 
         public IView View { get; set; }
 
-        public TaskCompletionSource<TOutput> CloseCompletionSource { get; set; }
```
simonrozsival: I actually think that this should be `private readonly`. There should be a public getter just for the `CloseCompletionSource.Task`.

PR #5244, opened 2019-05-28
Do a fresh start when re-opening the app after the application process was killed

### LINKED ISSUE #5243: Crash on MainTabBarActivity.getTabViewModel
## 💥 Crash Report
Crash on Toggl.Droid.Activities.MainTabBarActivity.getTabViewModel[TTabViewModel] ()

### 👥 Impact
It has affected 3 users (including myself) in the internal testing track for Giskard 2.0.

### 🔗 Context
Introduced by #5192 to help restoring the fragments in the `MainTabBarActivity`

### 🛰 Link
https://appcenter.ms/orgs/Toggl/apps/Toggl-Android/crashes/errors/737112481u/overview

`Exception type and message`
```cs
System.NullReferenceException: Object reference not set to an instance of an object
  at Toggl.Droid.Activities.MainTabBarActivity.getTabViewModel[TTabViewModel] () <0x79fa1981c4 + 0x00018> in <2e47f2e1e27d4650b34889ebaf17304e>:0
  at Toggl.Droid.Activities.MainTabBarActivity.restoreFragmentsViewModels () <0x79fa197c18 + 0x00127> in <2e47f2e1e27d4650b34889ebaf17304e>:0
  at Toggl.Droid.Activities.MainTabBarActivity.OnCreate (Android.OS.Bundle savedInstanceState) <0x79fa197834 + 0x00067> in <2e47f2e1e27d4650b34889ebaf17304e>:0
  at Android.App.Activity.n_OnCreate_Landroid_os_Bundle_ (System.IntPtr jnienv, System.IntPtr native__this, System.IntPtr native_savedInstanceState) <0x79f8f94aac + 0x0005b> in <cafeed20decf42bb938bc89c15083ee2>:0
  at (wrapper dynamic-method) System.Object.10(intptr,intptr,intptr)
```

daividssilverio:
## What's this?
This is a way to gracefully redirect the users to the `SplashScreen` when the application state was lost and would be recreated but ours ViewModels wouldn't be ready for it.

### Relationships
Closes #5243

## Why do we want this?
This is one droid edge case. It happens when the application is recreated by the system. When this happens our `ViewModelCache` isn't worth anything as the `DependencyContainer` will be re-instantiated from scratch.

## How is it done?
It simply checks for the presence of the `ViewModel` in all `ReactiveActivities`, redirecting the user to the `SplashScreen` if the `ViewModel` isn't there, effectively giving the user a fresh start.

The `SplashScreen` theme is now used for all activities and the actual activity theme is set right before calling `base.OnCreate`. That removes the white screen that would appear when the user re-opens the application after it was killed by the system.

### Why not another way?
Because it's kind of out of scope to implement a proper application restoration mechanism at this point in the mvexit project.

But yeah, ideally, we need a restoration mechanism that works when the application process is killed no matter how deep the application state is (e.g `SelectClientViewModel`).

### Side effects
The application won't be restored to the latest activity it was left a user might have left. But... that's better than crashing or leaving the application unresponsive.
We could track how often this happens to help decide how fast we implement actual restoration.

## Review hints
Review by commit.
You can test the behaviour by enabling `Don't keep activities` and changing the `Background process limit` to 1 in the developer options.

## :squid: Permissions
🦑 after careful thought


Toggl.Droid/Activities/AboutActivity.cs
```diff
@@ -8,14 +8,20 @@
 
 namespace Toggl.Droid.Activities
 {
-    [Activity(Theme = "@style/AppTheme",
+    [Activity(Theme = "@style/Theme.Splash",
```
juanlaube: Can we leave the activity theme here and set the splash theme only on `BailOutToSplashScreen`?

Toggl.Droid/Activities/AboutActivity.cs
```diff
@@ -8,14 +8,20 @@
...
         ScreenOrientation = ScreenOrientation.Portrait,
         ConfigurationChanges = ConfigChanges.Orientation | ConfigChanges.ScreenSize)]
     public sealed partial class AboutActivity : ReactiveActivity<AboutViewModel>
     {
         protected override void OnCreate(Bundle bundle)
         {
+            SetTheme(Resource.Style.AppTheme);
             base.OnCreate(bundle);
+            if (ViewModelWasNotCached())
```
juanlaube: Why not move this to the base class?

Toggl.Droid/Activities/AboutActivity.cs
```diff
@@ -8,14 +8,20 @@
...
 {
-    [Activity(Theme = "@style/AppTheme",
+    [Activity(Theme = "@style/Theme.Splash",
         ScreenOrientation = ScreenOrientation.Portrait,
         ConfigurationChanges = ConfigChanges.Orientation | ConfigChanges.ScreenSize)]
     public sealed partial class AboutActivity : ReactiveActivity<AboutViewModel>
     {
         protected override void OnCreate(Bundle bundle)
         {
+            SetTheme(Resource.Style.AppTheme);
```
juanlaube: Related to my previous comment, instead of doing this, why not doing `SetTheme(Resource.Style.Splash)` when the view model is not cached?

Toggl.Droid/Activities/AboutActivity.cs
daividssilverio: No. Setting the theme here is the same as setting it in the `AndroidManifest.xml`
Effectively, setting the theme during `OnCreate` is too late.

The way it is, the activity is loaded with the theme from the manifest, changing it during `onCreate`. Without it, when the system is re-creating the activity, a white screen would appear until we've set the `Theme.Splash` manually.

Toggl.Droid/Activities/AboutActivity.cs
daividssilverio: Because returning from within `base.OnCreate(bundle)` would not stop the flow of execution.
That is, if:
```
if (ViewModelWasNotCached())
            {
                BailOutToSplashScreen();
                return;
            }
```
was inside the base class, the return would not stop whatever comes after calling `base.OnCreate` to happen. And `Finish` doesn't not guarantee that the flow of execution will be halted.

That is, this:
```
MyActivity : BaseActivity {
  OnCreate(bundle) {
    base.OnCreate(bundle);
    ViewModel.DoSomething(); <--- would still run.
  }
}
```
`ViewModel.DoSomething()` would still run, even if the `BailOutToSplashScreen` was called.

Toggl.Droid/Activities/AboutActivity.cs
daividssilverio: Because it wouldn't render the same effect, setting the theme here is too late. A white screen would appear when re-creating an activity.

juanlaube:
Looks ok to me, considering we can't do much about it.

heytherewill:
Let's do this

daividssilverio:
I'm adding an issue to revisit this: #5307

PR #5245, opened 2019-05-29
🐛 Safeguarding against the DateTime value

khoi:
## What's this?
For some mysterious reason, on Mari's phone the value of `SharedStorage`'s `GetLastUpdateDate` is set to default(DateTimeOffSet). This value is then used in the Intent handlers, and passed as param to the `GetAllSince` with `?since=0`. 

The server rejects request that asks for > 3 months of TEs. And this leads to all Siri intent fails.

## How is it done?
This is more of an attempt to test it.

### Why not another way?
I'd be grateful if u could figure out why this happens in the first place.

## Review hints
not really.

## :squid: Permissions
Me.


khoi:
bad idea

PR #5254, opened 2019-05-29
🎙Remove ability to create projects and tags from shortcut creation

Odrakir:
## What's this?
Projects and Tags need to be synced before we can use them with Siri. This means that the custom shortcut creation shouldn't be allowed to create new tags and projects.

This PR removes that.

### Relationships
Doesn't close any issue.

## How is it done?
I just added a bool parameter to disable creation from shortcut settings.

## :squid: Permissions
🎙


khoi:
code looks good, tests need to be updated though

PR #5258, opened 2019-05-29
Prevent deadlock on the main thread

### LINKED ISSUE #5221: App freezes after resuming from background

juanlaube:
## What's this?
This fixes the app freeze problem when resuming the app from background and pressing the stop button or continuing a time entry.

### Relationships

Closes #5221 

## Why do we want this?
🚫 🐛 

## How is it done?
`InvokeOnMainThread` is synchronous and waits for completion, leaving us with potential deadlocks. We should use `DispatchQueue.MainQueue.DispatchAsync` when we need to move an operation to the main thread.
There are still some cases where `InvokeOnMainThread` may be needed, but I'm sure those are only one or two.

In this particular case:
- When there are time entry changes (e.g: stop or continue a time entry), `ObserveAllTimeEntriesVisibleToTheUserInteractor` emits on the main thread.
- The shortcuts creator and some on boarding steps, subscribe to time entry changes on the same thread, and also dispatch synchronous, blocking operations on the main thread using `InvokeOnMainThread`, causing a deadlock.
- By deferring that operation on the main thread with `DispatchAsync`, we ensure that it is executed after the observer finishes.

### Why not another way?
I'm sure there are other ways to do this, probably `ObserveAllTimeEntriesVisibleToTheUserInteractor` shouldn't emit on the main thread, and the subscriber shouldn't use the main thread either, but changing that may introduce more errors.
When doing UI changes on iOS, `DispatchAsync` is the way to go and this is safe enough for now.

### Side effects
None.

## :squid: Permissions
🚫 Only me


juanlaube:
Ad hoc build: https://app.bitrise.io/build/eec683c82ac6c867

juanlaube:
There should be a follow-up issue to check all the usages of `InvokeOnMainThread`.

amulware:
Please create said issue. :D

juanlaube:
Follow up issue to review all usages of `InvokeOnMainThread` is #5263.

PR #5259, opened 2019-05-30
Reposition scrolling indicator in iPad landscape mode

### LINKED ISSUE #5005: Scrolling issues on iPad

dshr:
## What's this?
A small set of AL and code changes that fix scrolling issues on iPad

### Relationships
Closes #5005 

## Why do we want this?
Better looks

## How is it done?
By repositioning the scrolling indicator a bit to the right when in iPad landscape mode.

### Why not another way?
Seems simple enough, and was what we agreed with @maritoggl on.

### Side effects
None.

## Review hints
Scroll around!

## :squid: Permissions
All yours!

### DIFF
--- a/Toggl.iOS/Cells/TimeEntriesLog/TimeEntriesLogViewCell.xib
+++ b/Toggl.iOS/Cells/TimeEntriesLog/TimeEntriesLogViewCell.xib
@@ -4,13 +4,14 @@
         <adaptation id="fullscreen"/>
     </device>
     <dependencies>
+        <deployment identifier="iOS"/>
         <plugIn identifier="com.apple.InterfaceBuilder.IBCocoaTouchPlugin" version="14490.49"/>
         <capability name="documents saved in the Xcode 8 format" minToolsVersion="8.0"/>
     </dependencies>
     <objects>
         <placeholder placeholderIdentifier="IBFilesOwner" id="-1" userLabel="File's Owner"/>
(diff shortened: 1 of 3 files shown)


Odrakir:
If I start with a full landscape app and then resize it to have compact width, the scroll doesn't show up. I think you need to call `updateTableViewScrollbarPosition` from `ViewDidLayoutSubviews` instead of `DidRotate`.

*But*... I'm thinking (and I've just tried it and it seems to work). Why not just make the `SuggestionsView` background color `Clear`?

dshr:
Closing this due to @Odrakir's suggestion being simpler.

PR #5260, opened 2019-05-30
Mvexit: Get google token refactoring

simonrozsival:
## What's this?
I had several objections to `IView` extending `IGoogleTokenProvider` - all of the views were able to access google tokens, even though it was needed just in the login and sign up views.

I submit this draft as an example of how I imagine the implementation. If you think this is a viable option, I might change this to a regular PR.

## How is it done?
The `LoginViewModel` and `SignUpViewModel` require the `IView` to be also `IGoogleTokenProvider` at runtime. I believe this is a good trade-off because it would be very easy to catch during testing.

The tests required only minor adjustments.

### Why not another way?
There would be an option to make a generic `IViewModel<TView>` and the `LoginViewModel` and `SignUpViewModel` could have explicit constraints for `TView : IView, IGoogleTokenProvider`, but it would make the whole thing much more complicated.

## :squid: Permissions
This is not meant to be merged. It has to be judged by the Mvexit project team.


juanlaube:
I like the idea, it's a trade-off I can live with as long as we don't abuse of the idea for other things.

simonrozsival:
The build now fails because of the renaming in `f066360`. I have no idea why and what else to change  so that it builds and the file name can be changed... but that does not have any impact on the rest of the PR and the idea behind.

PR #5280, opened 2019-05-31
iOS push notifications setup

### LINKED ISSUE #5043: Sort out iOS configuration
We need to set the `Entitlements.plist` and `GoogleServices-Info.plist` files correctly and any additional configuration which might be necessary to be able to work with push notifications in our iOS app.

Link to docs: https://firebase.google.com/docs/cloud-messaging/ios/client

simonrozsival:
## What's this?
This PR adds some settings and boilerplate to the iOS project so that we can now receive notifications in the iOS app (in the dev and production environments).

### Relationships
Closes #5043 

## Why do we want this?
We want our PNs to work on iOS.

## How is it done?
Settings + 🍰 script + some entry points in `AppDelegate`.

### Side effects
It'll be a bit of a pain to merge this into Mvexit style `AppDelegate`, but we'll create a separate file with the partial class once it's merged (unless the PN project is faster).

## Review hints
If you want to check if you can really send a (silent) notification to your device, call me and I'll explain the steps to send it using Postman or curl.

## :squid: Permissions
🦑


build.cake
```diff
@@ -367,7 +370,7 @@ private TemporaryFileTransformation GetIosEntitlementsConfigurationTransformatio
     {
         Path = path,
         Original = file,
-        Temporary = file.Replace(groupIdToReplace, groupId)
+        Temporary = file.Replace(groupIdToReplace, groupId).Replace(defaultApsEnvironment, apsEnvironment)
```
juanlaube: I'm not sure this is needed. Xcode automatically changes the entitlement based on the provisioning profile when building, but I don't know how that works with Xamarin.

Toggl.iOS/Startup/AppDelegate.cs
```diff
@@ -46,30 +47,46 @@ public sealed class AppDelegate : MvxApplicationDelegate<Setup, App<OnboardingVi
...
-            UNUserNotificationCenter.Current.Delegate = this;
-
             #if ENABLE_TEST_CLOUD
             Xamarin.Calabash.Start();
             #endif
 
             return true;
         }
 
+        public override void RegisteredForRemoteNotifications(UIApplication application, NSData deviceToken)
```
juanlaube: I think all these new methods should be in `AppDelegate.Notification.cs`.

juanlaube: There's another callback that might be useful, used when registering for remote notifications fail and we get an NSError with the reason.

simonrozsival:
@juanlaube I would suggest keeping the 🍰 transformation there, we can remove it later if it proves unnecessary.

Issue #5292, opened 2019-06-03
Implement service: Destroy FCM token


simonrozsival:
Add Android implementation of invalidating/destroying of the FCM token of #5283 

It should be something like (see https://docs.microsoft.com/en-us/xamarin/android/data-cloud/google-messaging/remote-notifications-with-fcm?tabs=macos#disconnecting-from-fcm):
```cs
FirebaseInstanceId.Instance.DeleteInstanceId();
```

daividssilverio:
Closed by #5303 (at least for now, reopen if necessary)

PR #5303, opened 2019-06-04
Implement IPushNotificationsTokenService for droid

### LINKED ISSUE #5031: Implement service: Access FCM token
Implement the platform specific part of #5283.

https://firebase.google.com/docs/cloud-messaging/android/client
Xamarin: https://docs.microsoft.com/en-us/xamarin/android/data-cloud/google-messaging/remote-notifications-with-fcm - it might be a bit outdated, check also https://github.com/xamarin/GooglePlayServicesComponents/tree/master/samples/com.google.firebase/firebase-messaging/FirebaseMessagingQuickstart

daividssilverio:
## What's this?
This implements the core of what's needed in droid to receive firebase notifications.

### Relationships
Closes #5031 

## Why do we want this?
To be able to send our fcm tokens to the backend and later receive notifications.

## How is it done?
This is a bunch of boilerplate. It's always the same, but here it's for Xamarin, and using an old api.

### Why not another way?
Because the `Xamarin.Firebase.Messaging` and related nuget packages are not updated or not working.

## Review hints
Check the docs linked on #5031.
Review by commit.

## Tests
You can test by adding logs into `TogglFirebaseMessagingService`:`OnMessageReceived`. And using Firebase's http apis.

## :squid: Permissions
PN team


simonrozsival:
I believe this will also close #5292

Toggl.Droid/Services/TogglFirebaseIIDService.cs
```diff
@@ -0,0 +1,31 @@
...
+            ApiEnvironment.Production;
+        #else
+            ApiEnvironment.Staging;
+        #endif
+        
+        public override void OnTokenRefresh()
+        {
+            var applicationContext = Application.Context;
+            var packageInfo = applicationContext.PackageManager.GetPackageInfo(applicationContext.PackageName, 0);
+            AndroidDependencyContainer.EnsureInitialized(environment, Platform.Giskard, packageInfo.VersionName);
```
simonrozsival: I'd like it more if the `AndroidDependencyContainer.EnsureInitialized` detected the `envirionment` itself and we could get rid of lines `12:17`. Let's not make any changes to the DI container in this project though 😄

Toggl.Droid/Services/PushNotificationsTokenServiceAndroid.cs
```diff
@@ -1,15 +1,33 @@
...
-            throw new System.NotImplementedException();
+            var dependencyContainer = AndroidDependencyContainer.Instance;
+            var keyValueStorage = dependencyContainer.KeyValueStorage;
+            keyValueStorage.Remove(PushNotificationStorageKey);
+            FirebaseInstanceId.Instance.DeleteInstanceId();
+        }
+
+        private PushNotificationsToken? getToken()
+        {
+            var dependencyContainer = AndroidDependencyContainer.Instance;
```
simonrozsival: ~Couldn't we just use `FirebaseInstanceId.Instance.Token`?~ I checked the docs and it's deprecated, so I guess this is the base way as `FirebaseInstanceId.Instance.GetInstanceId()` returns just a `Task` and not the `Task<IInstanceIdResult>` it should return (and which would contain the current token) :-/ I'll create an issue in the Microsoft GH repo, but I don't have any hopes.

Toggl.Droid/Services/TogglFirebaseIIDService.cs
daividssilverio: It would make sense to do so, right?

Toggl.Droid/Services/PushNotificationsTokenServiceAndroid.cs
daividssilverio: The new APIs work differently, we have a task to fire and we can add a listener and wait for completion. But that's on the new apis.

Toggl.Droid/Services/TogglFirebaseIIDService.cs
simonrozsival: Yep.

daividssilverio:
@simonrozsival I guess it does close #5292, but it does do any extra cleanup to avoid unwanted push notifications.

Toggl.Droid/Services/TogglFirebaseIIDService.cs
heytherewill: Do so in develop and then mergeback :)

Also, in general, I think we could remove all parameters from `EnsureInitialized`, make it receive only a context, and then moving this code that gets the package name to EnsureInitialized

Toggl.Droid/Properties/AndroidManifest.xml
```diff
@@ -1,10 +1,24 @@
...
+            android:name="com.google.firebase.iid.FirebaseInstanceIdInternalReceiver"
+            android:exported="false"/>
+        <receiver
+            android:name="com.google.firebase.iid.FirebaseInstanceIdReceiver"
+            android:exported="true"
+            android:permission="com.google.android.c2dm.permission.SEND">
+            <intent-filter>
+                <action android:name="com.google.android.c2dm.intent.RECEIVE"/>
+                <action android:name="com.google.android.c2dm.intent.REGISTRATION"/>
+                <category android:name="${applicationId}"/>
```
heytherewill: Do we need a cake replace or is this a thing?

daividssilverio: We don't need a cake replace. This is handled by the GooglePlayServices plugin.
It reads from the `google-services.json` and does its magic.

PR #5306, opened 2019-06-04
Allow non-descriptive TEs in suggestions

kornelijepetak:
## What's this?
TEs that had a project but no description were previously discarded from suggestion candidates. However, if a non-descriptive TE has a project, it can be allowed as a suggestion. 

This PR actually does this in one line, but also fixes test suite. Previously it was hard/impossible to set up some stuff needed for testing this particular use case. So, I moved to mocks for simplicity.

> Also, one test had an error in it: `Enumerable.Range(20, 0)` returned no elements when some were expected.

### Relationships
Closes #4977

## Why do we want this?
So that TEs without description don't feel left out.

## How is it done?
Relaxing the condition.

### Why not another way?
I've simplified (albeit not shorten) the code.

### Side effects
Should be none.

## :squid: Permissions
🔮 guys.


Toggl.Core.Tests/Interactors/Suggestions/MostUsedTimeEntrySuggestionProviderTests.cs
```diff
@@ -119,26 +124,29 @@ public async Task SortsTheSuggestionsByUsage()
...
-                                       .SetAt(Now)
-                                       .SetStart(Now)
-                                       .SetDescription("");
-                var emptyTimeEntries = Enumerable.Range(20, 0)
-                    .Select(_ => builder.Build());
-                var timeEntries = new List<IThreadSafeTimeEntry>(emptyTimeEntries);
-                timeEntries.AddRange(getTimeEntries(1, 2, 3, 4, 5));
+                var timeEntries = getCustomTimeEntries(10, (te, index) =>
+                {
+                    te.Project = index % 2 == 0 ? te.Project : null;
```
sofokli1: `index % 2 == 0` could be cached but it looks cleaner your way

kornelijepetak: Were it not just a test, I'd actually consider it. But it's just 10 time entries in a single test. 😄

PR #5310, opened 2019-06-04
Add generic sync interactor

### LINKED ISSUE #5285: Add interactor: Run sync after receiving a push notification in foreground
The interactor has only one responsibility at the moment: call `ISyncManager.ForceFullSync`.

We will add metrics as well later in #5066.

### LINKED ISSUE #5291: Add interactor: Run sync after receiving a push notification in background
The interactor has only one responsibility at the moment: call `ISyncManager.PullTimeEntries()`.

We will add metrics as well later in #5066.

daividssilverio:
## What's this?
This is a simple interactor that will be used to trigger a "light sync" when push notifications are received when the apps are in background.

### Relationships
Closes #5291 
It looks like this closes #5285 now as well

## Why do we want this?
To keep the apps updated even when the user isn't looking.

## How is it done?
Pretty straight forward, it's just an interactor that call `syncManager.PullTimeEntries`.

### Why not another way?
This seems simple enough.

## Review hints
Review by commit.

## :squid: Permissions
PN team 🦑 it


Toggl.Core/Interactors/Sync/RunLightSyncInteractor.cs
```diff
@@ -0,0 +1,27 @@
+using System;
+using System.Reactive.Linq;
+using Toggl.Core.Models;
+using Toggl.Core.Sync;
+using Toggl.Shared;
+
+namespace Toggl.Core.Interactors
+{
+    public class RunLightSyncInteractor : IInteractor<IObservable<SyncOutcome>>
```
simonrozsival: Let's not leak how it does it, let's call it by when it's indented to be used. I would prefer something like `RunPushNotificationInitiatedSyncInBackgroundInteractor`... ugh, it's very long, but I guess we will at least not use it somewhere by accident (especially when the specific metrics are in place). What do you think? Does it make any sense?

Odrakir: `RunBackgroundSyncInteractor` ?

simonrozsival: We can make that interactor generic/abstract/universal and instantiate it with different analytics events/performance stopwatches and with a different sync manager method it should trigger. The interactor factory would instantiate different instances based on where it is supposed to be used.

daividssilverio: I've added a generic interactor using composition. I'm passing the sync manager action, `MeasuredOperation` and analytic events as parameters to build the interactor. Please take a look.

daividssilverio: The `RunBackgroundSyncInteractor` is still there though, I haven't touched it.

Toggl.Core.Tests/Interactors/Sync/RunSyncInteractorTests.cs
```diff
@@ -0,0 +1,162 @@
...
+                await interactor.Execute().SingleAsync();
+
+                SyncStartedAnalyticsEvent.Received().Track();
+                SyncFinishedAnalyticsEvent.Received().Track(nameof(SyncOutcome.NewData));
+                SyncFailedAnalyticsEvent.DidNotReceive().Track(Arg.Any<string>(), Arg.Any<string>(), Arg.Any<string>());
+            }
+        }
+
+        [Serializable]
+        public struct NamedFuncHolder<T, T2>
```
daividssilverio: This is a helper to make the tests look pretty.

simonrozsival: Good idea!

Issue #5284, opened 2019-05-31
Add interactor: Subscribe to push notifications


simonrozsival:
This is necessary for #5030, #5031.

1. Obtain FCM token through a platform specific service (#5283).
2. API request (#5238).

simonrozsival:
There is an open question: what to do when the API request fails? We should probably either repeat the request immediatelly (e.g., the same way as we implement it for signup - delayed retry) or next time the app is opened.

First implement a naive solution of the interactor (which ignores failure of the HTTP request), but don't close this issue. Please suggest then a solution to this issue and we'll discuss it here and decide on the next steps.

daividssilverio:
We can check if the app has successfully registered an fcm token with our servers every time the app is resumed.

On droid, the usual strategy requires us to save the token anyways.

daividssilverio:
🤔 We could call the `SubscribeToPushNotificationsInteractor` (#5316) everytime the app tries to sync or comes to foreground. That would check if the token stored internally has already been registered with the api, and register it if not, doing nothing otherwise.

simonrozsival:
The implementation was merged into project branch.

daividssilverio:
@simonrozsival  But I thought we'd close this issue after we took care of:
> First implement a naive solution of the interactor (which ignores failure of the HTTP request), but don't close this issue. Please suggest then a solution to this issue and we'll discuss it here and decide on the next steps.

PR #5316, opened 2019-06-04
Add SubscribeToPushNotificationsInteractor

daividssilverio:
## What's this?
This adds an interactor to call register the apps in the backend to receive push notifications

### Relationships
Ref #5284 

## Why do we want this?
To be able to receive push notifications :)

## How is it done?
This interactor is quite dumb, it just calls the push services api to register the token.
It won't do anything if the token has already been registered, or the token is invalid/null.
It also does nothing when the api call fails, at least for now.
It stores the fcm token in the key value storage if the api succeeds to register it.

### Why not another way?

### Side effects
It stores the fcm token in the key value storage if the api succeeds to register it.

## Review hints
Review by commit. Check the tests.

## :squid: Permissions
PN team, 🦑 it


Toggl.Core/Interactors/PushNotifications/SubscribeToPushNotificationsInteractor.cs
```diff
@@ -0,0 +1,51 @@
...
+        }
+
+        public IObservable<Unit> Execute()
+        {
+            var token = pushNotificationsTokenService.Token;
+            if (!token.HasValue || token.Value == default(PushNotificationsToken))
+                return Observable.Return(Unit.Default);
+
+            var previouslyRegisteredToken = keyValueStorage.GetString(PreviouslyRegisteredTokenKey);
+            if (!string.IsNullOrEmpty(previouslyRegisteredToken) && new PushNotificationsToken(previouslyRegisteredToken) == token.Value)
```
simonrozsival:
```suggestion
            if (!string.IsNullOrEmpty(previouslyRegisteredToken) && previouslyRegisteredToken == (string)token.Value)
```

Toggl.Shared/PushNotificationsToken.cs
```diff
@@ -13,5 +13,15 @@ public PushNotificationsToken(string token)
 
         public static explicit operator string(PushNotificationsToken pushNotificationsToken)
             => pushNotificationsToken.token;
+
+        public static bool operator ==(PushNotificationsToken one, PushNotificationsToken other)
```
simonrozsival: We don't need this for now (if you accept my suggestion above).

Toggl.Core/Interactors/PushNotifications/SubscribeToPushNotificationsInteractor.cs
```diff
@@ -0,0 +1,51 @@
...
+using Toggl.Networking;
+using Toggl.Networking.ApiClients;
+using Toggl.Shared;
+using Toggl.Storage.Settings;
+
+namespace Toggl.Core.Interactors.PushNotifications
+{
+    public class SubscribeToPushNotificationsInteractor : IInteractor<IObservable<Unit>>
+    {
+        public const string PreviouslyRegisteredTokenKey = "PreviouslyRegisteredTokenKey";
```
simonrozsival: We must get rid of accessing the key directly from `Core`. Let's keep it like this for now and I'll create an issue which will add a storage service which will have concrete methods for accessing the tokens.

simonrozsival: Or maybe it's not necessary? Will it be used only from here? And it's public just because of unit tests?

daividssilverio: Basically, I've made it public to help with the tests.
But @Odrakir added an interface in #5317 which we could reuse. (Toggl.Core/Interactors/PushNotifications/PushNotificationTokenKeys.cs)
```
namespace Toggl.Core.Interactors
{
    public class PushNotificationTokenKeys
    {
        public static string TokenKey = "PushNotificationTokenKey";
    }
}
```

simonrozsival: If we need it to be public, then let's keep it in the `PushNotificationTokenKeys` class.

Issue #5298, opened 2019-06-03
Execute interactor: After logout


simonrozsival:
Execute the #5293 interactor from the `LogoutInteractor`.

PR #5317, opened 2019-06-05
Push token service implementation

### LINKED ISSUE #5030: Implement service: Access FCM token
Implement platform specific code for #5283 

Link to the docs: https://firebase.google.com/docs/cloud-messaging/ios/client, https://github.com/xamarin/GoogleApisForiOSComponents/blob/master/Firebase.CloudMessaging/component/GettingStarted.md#access-the-registration-token

Odrakir:
## What's this?
This implements the push token service for iOS, but also changes Android's implementation so the common code is in interactors.

### Relationships
Closes #5030 

## Why do we want this?
My idea is to share as much code as possible through interactors and live the services for platform specific code only. That's why I moved the access to the key/value storage.

I'm not sure if this would work on Android due to bugs/deprecations, but in any case I think that should be fixed in its service.

I'm not 100% happy with `StoreNewTokenInteractor` having an optional token parameter. It's optional because Android doesn't get this value in `OnTokenRefresh`, so in the case we send null the Interactors gets the current value.

Please discuss.

## :squid: Permissions
PNs members


Toggl.Core/Interactors/PushNotifications/PushNotificationTokenKeys.cs
```diff
@@ -0,0 +1,7 @@
+namespace Toggl.Core.Interactors
+{
+    public class PushNotificationTokenKeys
+    {
+        public static string TokenKey = "PushNotificationTokenKey";
```
daividssilverio: Actually, we could name this `NewPushNotificationTokenKey`.
We can then store a token that was registered in backend with the key `PushNotificationTokenKey`, that would be used in https://github.com/toggl/mobileapp/pull/5316/files#diff-907610234afc5f597fa7fd9a32506571

Odrakir: 🤔 So... just to clarify: We are storing previous and new tokens in the key/value storage. The previous one is to see if we need to update the server and the new is only in case we can't retrieve it from Firebase SDK, correct?

daividssilverio: > The previous one is to see if we need to update the server
correct.

> the new is only in case we can't retrieve it from Firebase SDK
now that you said it... maybe we don't need to store the new one 🤔
If we can rely on `FirebaseInstanceId.Instance.Token`... we shouldn't need it...

Odrakir: I assumed that works for iOS, but I understood from your comments (or Simon's, I can't remember) it isn't like that for iOS. Am I mistaken?

In the case it works then yes, we only need to store one token, and it actually can be just a hash.

simonrozsival: Let's try a hash of one token.

daividssilverio: Maybe it was me, either way, checking the docs, it seems ok (not ideal) to use `FirebaseInstanceId.Instance.Token`. It will return null when the app is has not yet been registered with Firebase, otherwise, that token should be valid.

Odrakir: Ok, update to this PR coming.

daividssilverio:
This is also closing #5298 , right?

Odrakir:
Yes, these were supposed to be different PRs into the main branch, but Simon merged the other one into this one. I'll close both issues.

PR #5330, opened 2019-06-06
Change check for user logged in

Odrakir:
## What's this?
This changes how we check for user logged in when the app gets a push notification.

### Relationships
It doesn't close any issue, it fixes the implementation of #5326 

## :squid: Permissions
@simonrozsival


Toggl.iOS/Startup/AppDelegate.Notifications.cs
```diff
@@ -25,14 +25,20 @@ public void DidReceiveRegistrationToken(Messaging messaging, string fcmToken)
 
         public override void DidReceiveRemoteNotification(UIApplication application, NSDictionary userInfo, Action<UIBackgroundFetchResult> completionHandler)
         {
-            var interactorFactory = IosDependencyContainer.Instance.InteractorFactory;
+            var dependencyContainer = IosDependencyContainer.Instance;
+            var interactorFactory = dependencyContainer.InteractorFactory;
+
+            if (!dependencyContainer.UserAccessManager.CheckIfLoggedIn())
```
daividssilverio: Good call. I'll update my PR with this change as well.

PR #5331, opened 2019-06-06
Remove duplicate strings

simonrozsival:
## What's this?
We have some duplicate strings in `Resource.resx` again...

## Why do we want this?
Rider shows it as an error and it annoys me.

## How is it done?
🔥 

### Side effects
None.

## :squid: Permissions
🦑


Toggl.Shared/Resources.resx
```diff
@@ -937,9 +931,6 @@ Use them to add keywords to your time entries.</value>
   </data>
     <data name="ShowReportsInvocationPhrase" xml:space="preserve">
     <value>Show my time</value>
-  </data>
-    <data name="ContinueTimerInvocationPhrase" xml:space="preserve">
```
Odrakir: Is `xml:space="preserve"` necessary?

simonrozsival: It doesn't have any effect.

simonrozsival: (in this case)

Issue #4955, opened 2019-04-19
Syncing issue after switching between connections


VeljkoTN:
## ⛑️ User Support

A user reported that if they are connected to a WiFi network while tracking time and switch to their carrier network at one point the app will have trouble syncing and manual sync will not help. 

Both Nina and I were able to reproduce this with version 1.20, you can find the steps below. 

## About the user

App version: Daneel/1.20
Phone OS: iOS 12.2
Phone model: iPhone X

## Expected behaviour
The app should be able to sync after users switch between networks. 

## To Reproduce

1. Start a timer while connected to a WiFi 
2. While the timer is running turn off WiFi to allow your phone to connect to your 3g/4g network
3. Stop the timer and the app will try syncing without success and you should end up with something like this:

![image](https://user-images.githubusercontent.com/36152180/56420517-865b5c80-629e-11e9-8154-146e2c75d84d.png)

NOTE: restarting the app helps, but the issue can be reproduced again after that.

lise-toggl:
This has also been reported by a user with S10+ ([case](https://app.intercom.io/a/apps/ayixs927/inbox/inbox/3155115/conversations/21886885017))

dshr:
Seems to be related to #5150 — we get the same connection timeouts as there.

simonrozsival:
I don't think it has something to do with the issue that @dshr linked. If there is a pending HTTP request which fails (because the connection to the server fails when user switches from wifi to cellular or the other way round or when he switches base stations while on the move), and it should display "Offline" (as is shown in the screenshot). But if the user pulls to refresh, it should all work. If not, and the only way to recover is restarting the app, then this is something we should look into.

@VeljkoTN can you check if the app is irrecoverably stuck after this happens?

VeljkoTN:
@simonrozsival yes it is, triggering a sync manually doesn't work, only restarting the app helps.

dshr:
After initially looking into the case I have discovered that we don't get repeated "offline mode" on the latest `develop` commit anymore, but any sync started after a connection change takes a quite while to actually return the "offline mode detected" message. This way, following the steps from the issue (start a TE, switch to TE, stop TE) would first result in an unsynced TE, then the first manual sync would return an offline mode message, and the next one would sync just fine. Still, this is quite bad UX.

[...]

Issue #3687, opened 2018-11-07
Misbehaving AndroidClientHandler


daividssilverio:
## 🐛 Bug Report
After #3602, Giskard Https requests stopped working in debug mode.
AndroidClientHandler doesn't seem to trust our certificate.

## Context
#3602 was made to help improve our integration tests. We kind of want it.

## Expected behavior
We should be able to run the app in debug mode.

## To reproduce
Run the app with `AndroidClientHandler` as the `HttpMessageHandler` for the `ApiFactory`.

## Suggestions
Testing can still be done if you replace the `AndroidClientHandler` with `HttpClientHandler`.

daividssilverio:
ref #3686

heytherewill:
🤖

PR #5332, opened 2019-06-06
iOS: use native HTTP handler

simonrozsival:
## What's this?
This PR adds native `NSUrlSessionHandler` for handling HTTP requests.

### Relationships
Ref #3588 (does not close it, because this PR affects only iOS and the user reported the problem on Android). It was originally implemented in #3602 in November, but the PR was reverted because of problems on Android. It might be also related to #4955 and some other syncing issues related to VPNs.

## Why do we want this?
We want our HTTP requests to be reliable. Let the OS handle it.

This will also allow us to debug HTTP communication on iOS - we will be able to intercept all of the communication between the API server and the app and check all of the HTTP request. This 

## How is it done?
Explicit dependency injection.

The changes in the iOS csproj files do not have any effect - they are overwritten with the explicit specification of the `HttpMessageHandler`. I changed them to avoid any confusion.

### Why not another way?
I could have leave the default implementation in `TogglApiFactory` and let the integration and syncing tests use it. I decided that it would be weird, because it would mean that there is code which is necessary only for tests in the production code. I also didn't want to make it optional in the constructor of `TogglApi`, because passing `null`s around is confusing and error prone.

### Side effects
None.

## Review hints
Try logging and and try creating a TE in the app and check that it syncs to the server.

## Tests
Integration tests: https://app.bitrise.io/build/d7253bd053794c33 - do not merge this branch unless these tests are passing.

Syncing tests: https://app.bitrise.io/build/f4a8c833bea571b2 - do not merge this branch unless these tests are passing.

## :squid: Permissions
I'll merge this if we can verify that it doesn't break Android, iOS, and Siri.

### DIFF
--- a/Toggl.iOS.SiriExtension/Toggl.iOS.SiriExtension.csproj
+++ b/Toggl.iOS.SiriExtension/Toggl.iOS.SiriExtension.csproj
@@ -25,7 +25,7 @@
     <IOSDebuggerPort>29481</IOSDebuggerPort>
     <MtouchLink>None</MtouchLink>
     <MtouchArch>x86_64</MtouchArch>
-    <MtouchHttpClientHandler>HttpClientHandler</MtouchHttpClientHandler>
+    <MtouchHttpClientHandler>NSUrlSessionHandler</MtouchHttpClientHandler>
     <DeviceSpecificBuild>false</DeviceSpecificBuild>
     <MtouchVerbosity>
     </MtouchVerbosity>
@@ -49,7 +49,7 @@
     <IOSDebuggerPort>17594</IOSDebuggerPort>
     <MtouchLink>SdkOnly</MtouchLink>
     <MtouchArch>ARM64</MtouchArch>
-    <MtouchHttpClientHandler>HttpClientHandler</MtouchHttpClientHandler>
+    <MtouchHttpClientHandler>NSUrlSessionHandler</MtouchHttpClientHandler>
     <MtouchVerbosity>
     </MtouchVerbosity>
   </PropertyGroup>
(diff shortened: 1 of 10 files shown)


Toggl.iOS.SiriExtension/Toggl.iOS.SiriExtension.csproj
```diff
@@ -80,7 +80,7 @@
     <CodesignEntitlements>Entitlements.plist</CodesignEntitlements>
     <MtouchLink>SdkOnly</MtouchLink>
     <MtouchArch>ARM64</MtouchArch>
-    <MtouchHttpClientHandler>HttpClientHandler</MtouchHttpClientHandler>
+    <MtouchHttpClientHandler>NSUrlSessionHandler</MtouchHttpClientHandler>
```
heytherewill: Shouldn't changing this be enough to change the underlying implementation, making the rest of the code changes irrelevant? 🤔 

Also, do we remember what the issue with Droid was?

simonrozsival: It is enough. On the other hand, there is no reason to keep the `HttpClientHandler` in the csproj. It's just confusing to have two places configuring the same thing with different settings. Unless you remember what the precedence is, you must check the docs.

The android issue was #3687 - the staging certificate didn't work with it and it was easier to revert the PR than figure out how to fix the problem with the certificate back then.

dshr: According to the [docs](https://docs.microsoft.com/en-us/xamarin/cross-platform/macios/http-stack?context=xamarin/ios) setting this will make all `HttpClient()` calls return a `NSUrlSessionHandler` one, so we shouldn't really need the explicit setting.

simonrozsival: Maybe we could get rid of the explicit setting of the `HttpMessageHandler` and let .NET decide? It seems that `Gzip | Deflate` is the default anyway.

heytherewill: I'd rather let .net do it using its own configs, Simon. Please change that :)

As for Android, you can have multiple HttpClientHandler entries in the CSProj: A normal one for debug and the OkHttp one for staging.

simonrozsival:
I'm closing this and I will replace it with a new PR.

PR #5333, opened 2019-06-06
Sync on droid pushes

### LINKED ISSUE #5282: Initiate sync when a PN is received
- [x] If we receive a push notification and the user is not logged in, we ignore it

When the app is:
- [x] in foreground `=>` execute interactor from #5285 
- [x] in background `=>` execute interactor from #5291

Call any necessary completion callbacks or implement any other platform specific requirements.

daividssilverio:
## What's this?
This PR adds the calls to the right interactors to sync the app when a push notification is received.

### Relationships
Closes #5282

## Why do we want this?
Because we want our users to have their time entries updated.

## How is it done?
This PR does two things:
1. It adds better detection of foreground/background on droid.
2. It calls the right interactors to sync the app based on the foreground/background state.

### Why not another way?
Because this is simple enough.

### Side effects
The app is cooler?

## Review hints
Review by commit.

## :squid: Permissions
PN team, 🦑 it


Toggl.Droid/Toggl.Droid.csproj
```diff
@@ -1187,4 +1191,11 @@
...
   <Import Project="..\packages\Xamarin.Android.Support.Design.28.0.0.1\build\monoandroid90\Xamarin.Android.Support.Design.targets" Condition="Exists('..\packages\Xamarin.Android.Support.Design.28.0.0.1\build\monoandroid90\Xamarin.Android.Support.Design.targets')" />
   <Import Project="..\packages\Xamarin.Firebase.Messaging.60.1142.1\build\MonoAndroid80\Xamarin.Firebase.Messaging.targets" Condition="Exists('..\packages\Xamarin.Firebase.Messaging.60.1142.1\build\MonoAndroid80\Xamarin.Firebase.Messaging.targets')" />
+  <Import Project="..\packages\Xamarin.Android.Arch.Lifecycle.Extensions.1.1.1.1\build\monoandroid90\Xamarin.Android.Arch.Lifecycle.Extensions.targets" Condition="Exists('..\packages\Xamarin.Android.Arch.Lifecycle.Extensions.1.1.1.1\build\monoandroid90\Xamarin.Android.Arch.Lifecycle.Extensions.targets')" />
+  <Target Name="EnsureNuGetPackageBuildImports" BeforeTargets="PrepareForBuild">
```
heytherewill: This is not needed

Toggl.Droid/Services/TogglFirebaseMessagingService.cs
```diff
@@ -1,14 +1,46 @@
...
 using Firebase.Messaging;
+using Toggl.Shared.Extensions;
 
 namespace Toggl.Droid.Services
 {
     [Service]
     [IntentFilter(new[] { "com.google.firebase.MESSAGING_EVENT" })]
     public class TogglFirebaseMessagingService : FirebaseMessagingService
     {
+        private CompositeDisposable disposeBag = new CompositeDisposable();
```
heytherewill: I'd rather this be a plain IDisposable rather than a bag that only holds one element

Toggl.Droid/Startup/ApplicationComponentCallbacksObserver.cs
```diff
@@ -0,0 +1,31 @@
...
+        {
+        }
+
+        public void OnLowMemory()
+        {
+        }
+
+        public void OnTrimMemory(TrimMemory level)
+        {
+            switch (level)
```
heytherewill: At this point, why not override the call directly on Application and check for nulls?

daividssilverio: Might as well.

daividssilverio:
@simonrozsival While testing this I've come across a very noticeable slow sync (full syncs).
I've git bisected it and found that the problem seems to have been originated from https://github.com/toggl/mobileapp/pull/5328 (78e9c6f3a0a25ad8e7b6f94c303c97b15a953b83).

So, I'm safely merging this. It didn't introduce any slowdowns.

PR #5335, opened 2019-06-07
Android: use native HTTP handler

### LINKED ISSUE #3588: Proxy Support

simonrozsival:
## What's this?
This is an extension of #5332. It adds native HTTP handler (`Xamarin.Android.Net.AndroidClientHandler`) to the `Toggl.Droid` project.

### Relationships
Closes #3588 (the user reported the problem on Android). It was originally implemented in #3602 in November, but the PR was reverted because of problems on Android, but that can be easily overcome by using the managed HTTP handler for non-production builds. It might be also related to #4955 and some other syncing issues related to VPNs.

## Why do we want this?
Better integration with the OS - respect proxies and stuff.

## How is it done?
I used `#if USE_PRODUCTION_API` to select the native handler only for production builds. If somebody has an idea how to make the native handler work with our staging SSL certificate, I'm all ears and I will use that instead.

### Side effects
The app might behave a bit differently when testing in the simulator and in ad-hoc builds than in production.

## Review hints
Try this with `USE_PRODUCTION_API` on and off.

## :squid: Permissions
Not yet, this is just a draft. I have no idea if you guys want this on Android or not. Also  this PR depends on #5332 and it should probably be rebased and merged when it's merged into `develop`.


simonrozsival:
I'm closing this and I will replace it with a new PR.

PR #5338, opened 2019-06-07
🚧Remove duplicate suggestions

### LINKED ISSUE #4975: Suggestions: Design system to merge suggestion providers
~~This may be a bit tricky. We want a system that uses some kind of scoring system for the sorting of suggestions that come from suggestion providers. There already was an effort for this in the previous project but it may need some rethinking.~~

~~Reference issue from the past project #3807 #3854~~

This will be simplified to use 1 suggestion from the machine learning approach, 1 or more from the frequency based provide (the current logic of the suggestion) and if the calendar event based suggestion provider is used we may add 1 from that provider as well. for a total of up to 3.

Make sure there is some logic that removes possible duplicate TEs and replaces the duplicate with an other one from the frequency based provide.

sofokli1:
## What's this?
Removes duplicate suggestions

### Relationships
Closes #4975

## Why do we want this?
We do not want duplicate suggestions

## :squid: Permissions
only 🔮 pals


Toggl.Core/Interactors/Suggestions/GetSuggestionsInteractor.cs
```diff
@@ -8,56 +9,41 @@
...
-            return new List<ISuggestionProvider>
-            {
-                new RandomForestSuggestionProvider(stopwatchProvider, dataSource, timeService, suggestionCount),
-                new CalendarSuggestionProvider(timeService, calendarService, defaultWorkspaceInteractor),
-                new MostUsedTimeEntrySuggestionProvider(timeService, dataSource, suggestionCount)
-            };
-        }
+        private IList<Suggestion> removingDuplicates(IList<Suggestion> suggestions)
+            => suggestions
+                .GroupBy(s => new { s.Description, s.ProjectId })
```
kornelijepetak: Are we not gonna be using client and task into account for determining the duplicity?

Toggl.Core.Tests/Interactors/Suggestions/GetSuggestionsInteractorTests.cs
```diff
@@ -46,14 +41,120 @@ public void ThrowsIfTheCountIsOutOfRange(int count)
...
+                for (int i = 0; i < numberOfProviders; i++)
+                {
+                    var suggestions = new List<Suggestion>();
+
+                    for (int j = 0; j < numberofSuggestionsPerProvider; j++)
+                    {
+                        suggestions.Append(makeSuggestion($"{i}{j}", i * 1000 + j, SuggestionProviderType.MostUsedTimeEntries));
+                    }
+
+                    providers.Append(makeProvider(suggestions));
```
kornelijepetak: `Append` is an extension method on `IEnumerable<T>` and creates a new enumerable, it does not modify the original. Use `.Add` instead.

Toggl.Core/Interactors/Suggestions/GetSuggestionsInteractor.cs
sofokli1: No I think Description and Project are enough unless you thing we will have an issue. In that case do elaborate.

Do you think we need workspace check? Should we have providers provide only based on default workspace? @amulware could also weigh in on this question

amulware: If the project id is the same, then the client and workspace MUST be the same. The task might be different though. If the task is shown in the suggestion and is selected for the time entry when started, we should match by that as well.

amulware: Oh, if the project is not set, the workspace might actually be different. Maybe worth checking too? Though, arguably, that'll result in two suggestions that look the same but do different things - Do confer with @maritoggl about that case perhaps.

PR #5343, opened 2019-06-07
Monitor token changes

### LINKED ISSUE #5294: Monitor FCM token changes
When the APN/FCM token changes, we must send the new token to the server so that we keep receiving the push notifications.

If a new token is received _and the user is logged in_, execute the interactor from #5284.

See https://github.com/xamarin/GoogleApisForiOSComponents/blob/master/Firebase.CloudMessaging/component/GettingStarted.md#monitor-token-generation.

Odrakir:
## What's this?
Monitors token changes in order to subscribe to push notifications. It also does some cleanup.

### Relationships
Closes #5294 

## How is it done?
I deleted the `StoreNewTokenInteractor`, which was useless. And renamed the `InvalidateCurrentTokenInteractor` one to `UnsubscribeFromPushNotificationsInteractor`.

Then I just check if the user is logged in and call `SubscribeToPushNotificationsInteractor` if that's the case.

## :squid: Permissions
Go PNs!


Toggl.iOS/Startup/AppDelegate.Notifications.cs
```diff
@@ -20,7 +21,17 @@ public override void FailedToRegisterForRemoteNotifications(UIApplication applic
...
+            var dependencyContainer = IosDependencyContainer.Instance;
+            var interactorFactory = dependencyContainer.InteractorFactory;
+
+            if (!dependencyContainer.UserAccessManager.CheckIfLoggedIn())
+            {
+                return;
+            }
+
+            interactorFactory.SubscribeToPushNotifications().Execute()
+                .Take(1)
```
daividssilverio: What's this for in this case? Why is it necessary?

PR #5346, opened 2019-06-08
Decompose UserAccessManager.CheckIfLoggedIn responsibilities 

### LINKED ISSUE #5345: Fix UserAccessManager CheckIfLoggedIn responsibility and usages
I found out that our `UserAccessManager:CheckIfLoggedIn` has a nasty side-effect: it re-creates the sync manager.

Thing is, the observables generated and already bound in our views were using the old `syncManager` (which stays functional). When a push arrives, the call to `CheckIfLoggedIn` is made, effectively recreating the `api`, `dataSource`, `syncManager` and `interactorFactory` in the dependency container.

This can cause some problems whenever we check if the user is logged in using that method.

I suggest that we separate the responsibilities in the `UserAccessManager`, adding another method to actually create/recreate the `api` and related fields in the dependency container; and making `CheckIfLoggedIn` just a query method, without drastic side-effects.

daividssilverio:
## What's this?
This refactors the `CheckIfLoggedIn` from the `UserAccessManager`, removing most of its side effects and creates another method to execute said side-effects.

### Relationships
Closes #5345 

## Why do we want this?
To query if the user is logged-in without spreading side-effects.
I found this problem when implementing the push notifications for android. When checking if the user was logged in to execute the sync after a push, I was ending up with a new instance of the `syncManager`, which was different to what was bound to the view.
This PR separates the concerns of checking if the user is logged in and creating the initial credentials when the app is first started.

## How is it done?
`CheckIfLoggedIn` now queries for the user api token presence, while maintaining the cache in the private storage to speed up startups.
A new method is added to `UserAccessManager`, `LoginWithSavedCredentials` that executes the old side effects that were generated the `CheckIfLoggedIn`.

### Why not another way?
Feel free to make suggestions. I'm making this a draft until everybody is happy or solves the problem otherwise.

### Side effects
On droid, the startup has two steps, half in the `TogglApplication`, half in the `SplashScreen`.
This is necessary because a push notification or a service can be started that might need to check if the user is authenticated or not. Now the api and syncManager are created in the `TogglApplication`, so there's no need to execute that portion of the code in the `SplashScreen`.

iOS effectively doesn't change much.

## Review hints
Review by commit.
Make Suggestions and feel free to create another PR to supersede this one if you think this is a bad idea.

## :squid: Permissions
🦑 after draft and the tests are fixed.


Toggl.Core/UserAccessManager/UserAccessManager.cs
```diff
@@ -98,19 +99,33 @@ public bool CheckIfLoggedIn()
         {
             if (privateSharedStorageService.Value.HasUserDataStored())
             {
```
heytherewill: No need for braces when returning or throwing

daividssilverio:
> To me, it's confusing that there's no change around the places where CheckIfLoggedIn. Am I missing something? 🤔

@simonrozsival 
- `CheckIfLoggedIn` was being used inside `AppStart`'s `GetAccessLevel`.
- I've made `CheckIfLoggedIn` responsible for querying the presence of the ApiToken.
- Then, a separate call (`LoginWithSavedCredentials`) to run the old side effects of `CheckIfLoggedIn`, that would create a new `Api` instance, sync manager and related instances

- On droid, I'm doing something similar to what happens in the `AppDelegate`, but in the `TogglApplication`. 
- The `SplashScreen` on droid now only queries the presence of the api token.
- The other usages of `CheckIfLoggedIn` always happen after the application has been started inside droid specific services (native services, not the service pattern).

PR #5349, opened 2019-06-10
Use native network clients

### LINKED ISSUE #3588: Proxy Support

### LINKED ISSUE #4955: Syncing issue after switching between connections

simonrozsival:
## What's this?
This PR lets the OS pick the  HTTP handler on its own and it does not force the managed http client. 

### Relationships
Closes #3588, supersedes #5332, supersedes #5335, might resolve #4955 but that has to be yet determined.

## How is it done?
The `HttpClient` selects the HTTP handler based on the environment/runtime and configuration it runs in. On iOS, it uses `NSUrlSessionHandler` and on Android it uses `Xamarin.Android.Net.AndroidClientHandler`. The only exception will be the debug build of Android, where we will prefer the managed version, because of #3687.

### Side effects
Only positive side effects.

## Review hints
We must test this both on iOS in simulator and on a device and on Android in emulator and on a device.

## Tests
Integration tests: https://app.bitrise.io/build/5fb2d7e399721fa4 - do not merge this branch unless these tests are passing

## :squid: Permissions
me or @heytherewill


simonrozsival:
I had problems inspecting the HTTP calls in Charles. I was able to do that with #5332. Can somebody else try this on iOS?

simonrozsival:
I'm quite certain now that this PR doesn't work and I would like to close it in favor of #5332 again.

simonrozsival:
This doesn't solve neither the UX issue, nor the problem with VPN. Apparently, we must use explicit DI to force usage of native HTTP handler. I'm closing this PR.

PR #5350, opened 2019-06-10
Use only synced TEs for suggestions

### LINKED ISSUE #5339: Suggestions: Do not use un-synced data in suggestion providers
- check and perform necessary changes so that un-synced data are not used for suggestion

kornelijepetak:
## What's this?
Ignores unsynced TEs from suggestions in the following providers:

* Most Used
* Random Forest

The calendar suggestions are not coming from the TEs at all, so it's not applicable to them.

### Relationships
Closes #5339 

## Why do we want this?
May create some inconsistencies. Not really anything heavy, but still.

## How is it done?
Just filtering the TEs.

## :squid: Permissions
🔮 krew.


Toggl.Core/Suggestions/RandomForestSuggestionProvider.cs
```diff
@@ -38,17 +38,17 @@ public sealed class RandomForestSuggestionProvider : ISuggestionProvider
...
+                .SelectMany(toSuggestions);
 
         private IEnumerable<Suggestion> toSuggestions(IEnumerable<IDatabaseTimeEntry> timeEntries)
             => timeEntries.Select(timeEntry => new Suggestion(timeEntry, SuggestionProviderType.RandomForest));
 
         private IEnumerable<IDatabaseTimeEntry> predictUsingRandomForestClassifier(IEnumerable<IDatabaseTimeEntry> timeEntriesForPrediction)
         {
             var timeEntries = timeEntriesForPrediction
-                .Where(te => te.ProjectId.HasValue)
+                .Where(te => te.ProjectId.HasValue && te.SyncStatus == Storage.SyncStatus.InSync)
```
sofokli1: I think this check should also be in line 60

sofokli1: Actually shouldn't this happen in `GetSuggestion()` so it happens in one place for both cases? (by cases I mean `predictUsing2Steps` and `predictUsing1Step `)

kornelijepetak: You are right.

Toggl.Core/Suggestions/RandomForestSuggestionProvider.cs
```diff
@@ -38,9 +39,10 @@ public sealed class RandomForestSuggestionProvider : ISuggestionProvider
 
         public IObservable<Suggestion> GetSuggestions()
             => dataSource.TimeEntries
-                     .GetAll()
-                     .Select(predictUsingRandomForestClassifier)
-                     .SelectMany(toSuggestions);
+                .GetAll()
+                .Select(timeEntries => timeEntries.Where(te => te.SyncStatus == SyncStatus.InSync))
```
sofokli1: forgive my lack of knowledge here. does this first load all time entries to memory and then filters those or is the `where` check done in the database?

kornelijepetak: The mere fact of calling `GetAll` does not itself load things into memory, but it returns an `IEnumerable` as opposed to `IQueryable` so that at whatever point in the future this enumerable gets enumerated, it **will work on ALL TEs in memory**. It will not modify the way data is fetched from the database. (Un)fortunatelly, our current datasources/repositories do not leak IQueryables outside so it's not possible to help Realm be faster by LINQ queries after `.GetAll()`.

PR #5365, opened 2019-06-11
Display project & task placeholders

daividssilverio:
## What's this?
This PR adds some spans to time entries texts in the main log, running time entry and edit time entry screen.

### Relationships
Ref #5136 (Droid bits)

## Why do we want this?
This could be better than displaying `(unsynced project)` and `(unsynced task)`.
Maybe not?

## How is it done?
It adds another span that serves the purpose of the placeholder.
It replaces the `(unsynced X)` text with the drawable placeholder.

### Why not another way?
Because adding extra views to the layouts would be too much work and would make them even more complicated than they already are. 

### Side effects
This shouldn't have any.

## Review hints
You can force:
```
public bool ProjectIsPlaceholder
            => ProjectName == Resources.ProjectPlaceholder ||
               ProjectName == Resources.InaccessibleProject;

         public bool TaskIsPlaceholder
            => TaskName == Resources.TaskPlaceholder;
```
to return `true` in `Toggl.Core.UI/ViewModels/TimeEntriesLog/LogItemViewModel.cs`.
Likewise on `Toggl.Droid/Extensions/TimeEntryExtensions.cs`.

## :squid: Permissions
🦑 after the Draft. We are likely to change the drawable and the logic to transform the placeholder in the text (the code calculate `transY` in `Toggl.Droid/Views/VerticallyCenteredImageSpan.cs:Draw` is very cryptic at the moment)


daividssilverio:
It looks like this at the moment:

Main Log:
image

Running time entry:
image

Edit Screen:
image

Toggl.Core.UI/ViewModels/TimeEntriesLog/LogItemViewModel.cs
```diff
@@ -41,6 +41,13 @@ public sealed class LogItemViewModel : IDiffable<IMainLogKey>, IEquatable<LogIte
         public bool BelongsToGroup =>
             VisualizationIntent == LogItemVisualizationIntent.GroupItem;
 
+        public bool ProjectIsPlaceholder
+            => ProjectName == Resources.ProjectPlaceholder ||
```
simonrozsival: This is not a reliable way of checking if it is or isn't a placeholder. We must check for `SyncStatus.RefreshingNeeded`.

Toggl.Droid/Extensions/TimeEntryExtensions.cs
```diff
@@ -1,29 +1,39 @@
...
+                spannableString.Append(" ");
+            } else 
+                spannableString.Append(project, new ForegroundColorSpan(Color.ParseColor(projectColor)), SpanTypes.ExclusiveInclusive);
 
             if (!string.IsNullOrEmpty(task))
             {
-                spannableString.Append($": {task}");
+                if (isTaskPlaceholder(task))
+                {
+                    spannableString.Append(" : ");
```
simonrozsival: :fire: the `:`

Toggl.Core.UI/ViewModels/TimeEntriesLog/LogItemViewModel.cs
daividssilverio: Is that enough? Can I check only that?

daividssilverio:
It looks like this with the latest updates:

image

image

PR #5370, opened 2019-06-12
Add analytics to push initiated syncs

### LINKED ISSUE #5066: Analytics: Track success and performance of processing PNs
- [x] Track what was the state of the app at the time when the notification was tracked (foreground, background)
- [x] Track how long sync triggered by push notifications take
- [x] Track how many TEs were pulled during PN initiated sync
- [x] Track how many placehoders were created
- [x] Track that PN initiated sync started
- [x] Track that PN initiated sync finished
- [x] Track when PN initiated sync fails

Inspiration: [RunBackgroundSyncInteractor](https://github.com/toggl/mobileapp/blob/develop/Toggl.Core/Interactors/Sync/RunBackgroundSyncInteractor.cs)

Odrakir:
## What's this?
This adds analytics to the interactor that fires sync for push notifications.

Part of the sync graph with the new state.
image

### Relationships
Closes #5066 

## How is it done?
I had to create some `ITrackableEvent` instances, because `IAnalyticsEvent` is not possible to create them and pass them as params that way.

## :squid: Permissions
Don't. I have to fix tests.


Toggl.Core/Interactors/Sync/RunSyncInteractor.cs
```diff
@@ -12,55 +12,54 @@ public class RunSyncInteractor : IInteractor<IObservable<SyncOutcome>>
...
+
+            var syncTimeStopwatch = stopwatchProvider.Create(meassuredOperation);
             syncTimeStopwatch.Start();
-            syncStartedAnalyticsEvent?.Track();
-            return syncAction(syncManager)
-                .LastAsync()
+
+            analyticsService.PushNotificationSyncStarted.Track(sourceState.ToString());
+
+            var syncAction = sourceState == PushNotificationSyncSourceState.Foreground
```
simonrozsival: I wasn't sure about this at first but it makes sense and it simplifies the interactor, so let's uses it.

Toggl.Core/TogglSyncManagerFactory.cs
```diff
@@ -359,6 +360,8 @@ public static class TogglSyncManager
             transitions.ConfigureTransition(persistTimeEntries.Done, updateTimeEntriesSinceDate);
             transitions.ConfigureTransition(updateTimeEntriesSinceDate.Done, detect);
 
+            transitions.ConfigureTransition(fetchTimeEntries.Done, timeEntriesAnalytics);
```
simonrozsival:
```suggestion
            transitions.ConfigureTransition(detect.Done, timeEntriesAnalytics);
```

Toggl.Core/TogglSyncManagerFactory.cs
```diff
@@ -359,6 +360,8 @@ public static class TogglSyncManager
...
+
             transitions.ConfigureTransition(detect.Done, new DeadEndState());
```
simonrozsival:
```suggestion
            transitions.ConfigureTransition(timeEntriesAnalytics.Done, new DeadEndState());
```

Toggl.Core/TogglSyncManagerFactory.cs
Odrakir: 🤔 So this is kind of a sequence of operators? I see, but I can't use this line as `detect.Done` is not returning a `StateResult<IFetchObservables>`.

Man... Wouldn't all this be much simpler with plain old Rx?

simonrozsival: Yeah, it's "if this finishes with this result, continue running this state".

It should be moved between `updateTimeEntriesSinceDate.Done` and `detect` instead.

PR #5373, opened 2019-06-12
🔮 Project: Suggestion improvements

sofokli1:
## What's this?
This are improvements to suggestions like:
- make android suggestions look like iOS
- add support for suggestions without description
- time based suggestions (machine learning)
- calendar event based suggestions
- better tracking for suggestion usage

## :squid: Permissions
Only 🔮


amulware:
Since we're ending the end of the project, already leaving this here: Based on the testing of recent builds, and with the 10m calculation removed, I'm good to go with merging this. 👌

Do get a formal review from each platform though. :)

Toggl.Core/Analytics/CalendarSuggestionContinuedEvent.cs
```diff
@@ -0,0 +1,40 @@
...
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Toggl.Core.Analytics
+{
+    public class CalendarSuggestionContinuedEvent : ITrackableEvent
+    {
+        public string EventName => "CalendarSuggestionContinued";
+
+        private string offsetCategory;
```
heytherewill:
```suggestion
        private readonly string offsetCategory;
```

Toggl.Core/Analytics/SuggestionsPresentedEvent.cs
```diff
@@ -0,0 +1,51 @@
...
+{
+    public class SuggestionsPresentedEvent : ITrackableEvent
+    {
+        public const string SuggestionsCountName = "SuggestionsCount";
+        public const string CalendarProviderStateName = "CalendarProviderState";
+        public const string DistinctWorkspaceCountName = "DistinctWorkspaceCount";
+        
+        public string EventName => "SuggestionsPresented";
+
+        private Dictionary<string, string> parameters;
```
heytherewill:
```suggestion
        private readonly Dictionary<string, string> parameters;
```

Toggl.Core/Interactors/InteractorFactory.Suggestions.cs
```diff
@@ -0,0 +1,16 @@
...
+using Toggl.Core.Suggestions;
+
+namespace Toggl.Core.Interactors
+{
+    public sealed partial class InteractorFactory : IInteractorFactory
+    {
+        public IInteractor<IObservable<IEnumerable<Suggestion>>> GetSuggestions(int count)
+            => new GetSuggestionsInteractor(count, this);
+
+        public IInteractor<IObservable<IReadOnlyList<ISuggestionProvider>>> GetSuggestionProviders(int count)
```
heytherewill: I don't think this needs to be an interactor at all. It's only ever used internally and the contents of the interactor are basically creating a list. I'd rather create this list in `GetSuggestionsInteractor`

Toggl.Core/Interactors/Suggestions/GetSuggestionsInteractor.cs
```diff
@@ -0,0 +1,44 @@
...
+            Ensure.Argument.IsNotNull(interactorFactory, nameof(interactorFactory));
+
+            this.getsuggestionProvidersInteractor = interactorFactory.GetSuggestionProviders(suggestionCount);
+            this.suggestionCount = suggestionCount;
+        }
+
+        public IObservable<IEnumerable<Suggestion>> Execute()
+            => getsuggestionProvidersInteractor
+                .Execute()
+                .SelectMany(CommonFunctions.Identity)
```
heytherewill: Replacing this whole call with a simple `SelectMany` will make things much simpler to read as well

Toggl.Core/Interactors/Suggestions/GetSuggestionsInteractor.cs
```diff
@@ -0,0 +1,44 @@
...
+                .Select(provider => provider.GetSuggestions())
+                .SelectMany(CommonFunctions.Identity)
```
heytherewill:
```suggestion
                .Flatten()
```

Toggl.Core/Interactors/Suggestions/GetSuggestionsInteractor.cs
```diff
@@ -0,0 +1,44 @@
...
+                .ToList()
+                .SelectMany(removingDuplicates)
+                .Take(suggestionCount)
+                .ToList();
+
+        private IList<Suggestion> removingDuplicates(IList<Suggestion> suggestions)
+            => suggestions
+                .GroupBy(s => new { s.Description, s.ProjectId, s.TaskId, s.WorkspaceId })
```
heytherewill: `.Distinct()` + a private equality comparer class would make this much nicer to wrap the head around

Toggl.Core/Suggestions/RandomForestSuggestionProvider.cs
```diff
@@ -0,0 +1,239 @@
...
+        private IEnumerable<IDatabaseTimeEntry> predictUsingRandomForestClassifier(IEnumerable<IDatabaseTimeEntry> timeEntriesForPrediction)
+        {
+            var timeEntries = timeEntriesForPrediction
+                .Where(te => te.ProjectId.HasValue)
+                .Take(maxNumberOfTimeEntriesForTraining)
+                .ToList();
+
+            if (timeEntries.Count >= minNumberOfTimeEntriesForTraining)
+            {
+                return predictUsing2Steps(timeEntries);
```
heytherewill: No need for braces when returning

Toggl.Core/Suggestions/RandomForestSuggestionProvider.cs
```diff
@@ -0,0 +1,239 @@
...
+            }
+
+            timeEntries = timeEntriesForPrediction
+                .Take(maxNumberOfTimeEntriesForTraining)
+                .ToList();
+
+            if (timeEntries.Count >= minNumberOfTimeEntriesForTraining)
+            {
```
heytherewill: No need for braces when returning

Toggl.iOS/Views/Suggestions/SuggestionView.cs
```diff
@@ -117,5 +122,13 @@ private void hideProjectTaskClient()
                 = ClientLabel.Hidden
                 = true;
         }
+
+        [Conditional("ADHOC")]
+        [Conditional("DEBUG")]
+        private void prefixWithProviderNameInDebug()
```
heytherewill: This looks like a leftover

Toggl.Core/Interactors/InteractorFactory.Suggestions.cs
kornelijepetak: If I remember correctly, we opted to do this because otherwise we are not able to mock test this list and make sure the suggestion mechanisms get the correct providers.

Toggl.iOS/Views/Suggestions/SuggestionView.cs
kornelijepetak: While I dislike debug code being there, it is not compiled into production, so at least that part is not an issue.

However, whenever we wanna check something up with the providers, it's useful to have this visible there, and it's a pain to introduce this every time again when we need to check things. 

The only thing I'd change her is remove this in ADHOC. It should be only in `DEBUG` so that our internal testers don't see this thing.

Check #5503 as well before making a final decision.

Toggl.Core/Interactors/InteractorFactory.Suggestions.cs
sofokli1: ☝️

Toggl.Core/Interactors/Suggestions/GetSuggestionsInteractor.cs
sofokli1: not sure how to do this, can you be a bit more specific?

Toggl.Core/Interactors/InteractorFactory.Suggestions.cs
heytherewill: Okay, that's reasonable

Toggl.Droid/Toggl.Droid.csproj
```diff
@@ -1179,4 +1197,11 @@
...
   <Import Project="..\packages\Xamarin.Android.Support.v7.RecyclerView.28.0.0.1\build\monoandroid90\Xamarin.Android.Support.v7.RecyclerView.targets" Condition="Exists('..\packages\Xamarin.Android.Support.v7.RecyclerView.28.0.0.1\build\monoandroid90\Xamarin.Android.Support.v7.RecyclerView.targets')" />
   <Import Project="..\packages\Xamarin.Android.Support.Design.28.0.0.1\build\monoandroid90\Xamarin.Android.Support.Design.targets" Condition="Exists('..\packages\Xamarin.Android.Support.Design.28.0.0.1\build\monoandroid90\Xamarin.Android.Support.Design.targets')" />
+  <Import Project="..\packages\Accord.3.8.0\build\Accord.targets" Condition="Exists('..\packages\Accord.3.8.0\build\Accord.targets')" />
+  <Target Name="EnsureNuGetPackageBuildImports" BeforeTargets="PrepareForBuild">
+    <PropertyGroup>
+      <ErrorText>This project references NuGet package(s) that are missing on this computer. Enable NuGet Package Restore to download them.  For more information, see http://go.microsoft.com/fwlink/?LinkID=322105.The missing file is {0}.</ErrorText>
```
juanlaube: What does this mean? Is this expected? 👀

Toggl.iOS/Views/Suggestions/SuggestionsView.cs
```diff
@@ -78,7 +81,7 @@ public void OnSuggestions(Suggestion[] suggestions)
                     SuggestionTapped.OnNext(suggestionView.Suggestion);
                 }));
             }
-            heightConstraint.Constant = heightForSuggestionCount(suggestions.Length);
+            heightConstraint.Constant = heightForSuggestionCount(suggestionCount);
```
juanlaube: The last suggestion is cut a few points on the bottom and hidden behind the main log entries, but gets fixed after scrolling a bit. The problem seems to be related to this constraint.

Toggl.Droid/ViewHolders/MainLogSuggestionItemViewHolder.cs
```diff
@@ -10,58 +12,54 @@ namespace Toggl.Droid.ViewHolders
...
-            timeEntriesLogCellClientLabel = ItemView.FindViewById<TextView>(Resource.Id.TimeEntriesLogCellClientLabel);
-            timeEntriesLogCellContinueImage = ItemView.FindViewById<ImageView>(Resource.Id.TimeEntriesLogCellContinueImage);
+            descriptionLabel = ItemView.FindViewById<TextView>(Resource.Id.DescriptionLabel);
+            projectLabel = ItemView.FindViewById<TextView>(Resource.Id.ProjectLabel);
+            clientLabel = ItemView.FindViewById<TextView>(Resource.Id.ClientLabel);
+
+            var horizontalMargin = 16.DpToPixels(ItemView.Context);
+            var marginLayoutParams = ItemView.LayoutParameters as ViewGroup.MarginLayoutParams;
+            var newLayoutParams = marginLayoutParams.WithMargins(horizontalMargin, null, horizontalMargin, null);
+            newLayoutParams.Width = ViewGroup.LayoutParams.MatchParent;
```
daividssilverio: Why was this done here and not in the xml itself?

Toggl.Droid/Toggl.Droid.csproj
kornelijepetak: I am not sure. It was automatically added by VS when Accord (dependency for ML algorithm) was added as a nuget package. Since this is injecting .targets during build, I don't think it's completely safe to remove it. It might be ok if it's already built once, but it may struggle on clean install. _At least I think so._
Maybe @heytherewill knows more about how `MSBuild's .targets` files work... For now, I'd rather be safe than sorry.

Toggl.iOS/Views/Suggestions/SuggestionsView.cs
kornelijepetak: @sofokli1 can you please see about this?

Toggl.Droid/ViewHolders/MainLogSuggestionItemViewHolder.cs
kornelijepetak: History. 🤷‍♂ 
Updating it now.

Issue #5363, opened 2019-06-11
Add Firebase remote configuration flag for Push Notifications


simonrozsival:
- [x] Enable/disable sending sending the PN token to the server (put the configuration into the sync state)
- [x] Enable/disable reacting to push notifications (add to platform specific entry points)

simonrozsival:
The implementation was merged into project branch.

PR #5377, opened 2019-06-12
Use remote when deciding to handle push notifications

daividssilverio:
## What's this?
This PR adds remote configuration for push notifications and use it to decide wether or not to handle new push notifications based on remote configuration.

### Relationships
Part of #5363
> Enable/disable reacting to push notifications (add to platform specific entry points)

## Why do we want this?
For 2 things:
- To disable push notification's handling if something goes terribly wrong.
- To enable push notification's handling only we are comfortable with it.

## How is it done?
It's very similar to what was done for the rating view remote config.
It uses the remote config service in the PN's entry points in both platforms.

### Why not another way?
Feel free to suggest something else.

### Side effects
Nothing.

## Review hints
You can use firebase remote config conditionals to test the remote configurations in debug builds (at least on droid).
You can also change the default values on `Toggl.Droid/Resources/xml/RemoteConfigDefaults.xml` on droid and `Toggl.iOS/RemoteConfigDefaults.plist` on iOS.
The default expiration time of the remote config is 12 hours.
You can change this value by calling `FirebaseRemoteConfig.Instance.Fetch` with an extra parameter. 

## :squid: Permissions
🦑 PN team


Toggl.Droid/Services/TogglFirebaseMessagingService.cs
```diff
@@ -24,7 +27,12 @@ public override void OnMessageReceived(RemoteMessage message)
                 ? interactorFactory.RunPushNotificationInitiatedSyncInForeground()
                 : interactorFactory.RunPushNotificationInitiatedSyncInBackground();
 
-            syncDisposable = syncInteractor.Execute()
+            var shouldHandlePushNotifications = dependencyContainer.RemoteConfigService.ShouldHandlePushNotifications(); 
+
+            syncDisposable = shouldHandlePushNotifications
+                .SelectMany(willHandlePushNotification => willHandlePushNotification
+                    ? syncInteractor.Execute().Select(_ => Unit.Default)
```
simonrozsival: `.SelectUnit()`

Toggl.Droid/Services/RemoteConfigServiceAndroid.cs
```diff
@@ -13,8 +13,10 @@ namespace Toggl.Droid.Services
 {
     public class RemoteConfigServiceAndroid : IRemoteConfigService
     {
-        private const string delayParameter = "day_count";
-        private const string triggerParameter = "criterion";
+        private const string ratingViewDelayParameter = "day_count";
```
simonrozsival: Could we make these constants shared between iOS and Android? Or should these be two different constants so that we can disable it only for one platform?

daividssilverio: We can share them. Will update.
And no, we shouldn't create different keys for droid and iOS, we can disable pushes on just one platform by adding conditionals in Firebase.

PR #5387, opened 2019-06-13
Only reload reports after significant time

juanlaube:
## What's this?
This avoids reloading the report when the user navigates from reports to another view and returns immediately.

### Relationships

Refs #4879 

## Why do we want this?
To avoid reloading

## How is it done?
Calculating the time since the user left the reports and returned, we only reload when that timespan is more than 5 seconds as suggested in the issue.

### Why not another way?
This is simple and avoids using subjects.

### Side effects
None.

## :squid: Permissions
🦑


Toggl.Core.UI/ViewModels/Reports/ReportsViewModel.cs
```diff
@@ -250,6 +257,9 @@ public void StopNavigationFromMainLogStopwatch()
         private bool viewAppearedForTheFirstTime()
             => startDate == default(DateTimeOffset);
 
+        private bool shouldReloadOnViewAppeared()
+            => timeService.CurrentDateTime - viewDissapearedAtTime >= TimeSpan.FromSeconds(5);
```
simonrozsival: Move `TimeSpan.FromSeconds(5)` into a `private readonly`.

PR #5389, opened 2019-06-13
Split ViewModel loading and Initialization

heytherewill:
## What's this?
This decouples the ViewModel loader from calling initialize on our views

### Relationships
This is a leftover from MVExit that we've only got time to refactor now

## Why do we want this?
This is needed for future refactors that will make us go :sanic:

## How is it done?
Loading has nothing to do with calling initialize, so we split these two things.

### Side effects
Please test that things work on your device, but it's all looking good so far.

## Review hints
Please test on actual devices rather than simulators

## :squid: Permissions
Any of the MVExit bois


Toggl.iOS/Startup/AppDelegate.cs
```diff
@@ -103,9 +103,11 @@ private void navigateAccordingToAccessLevel(AccessLevel accessLevel)
                     navigationService.Navigate<TokenResetViewModel>(null);
                     return;
                 case AccessLevel.LoggedIn:
-                    var viewModel = IosDependencyContainer.Instance.ViewModelLoader
-                    .Load<Unit, Unit>(typeof(MainTabBarViewModel), Unit.Default).GetAwaiter().GetResult();
+                    var viewModel = IosDependencyContainer.Instance
+                        .ViewModelLoader
+                        .Load<MainTabBarViewModel>();
                     Window.RootViewController = ViewControllerLocator.GetViewController(viewModel);
+                    viewModel.Initialize();
```
juanlaube: We should call Initialize before setting the root view controller to avoid possible problems of using uninitialized stuff from the view controller.

Toggl.Droid/Startup/SplashScreen.cs
```diff
@@ -64,8 +65,10 @@ protected override void OnCreate(Bundle savedInstanceState)
                 return;
             }
             
-            clearAllViewModelsAndSetupRootViewModel(dependencyContainer.ViewModelCache, dependencyContainer.ViewModelLoader); 
-            
+            var initializeTask = clearAllViewModelsAndSetupRootViewModel(
```
simonrozsival: do we need this variable?

heytherewill: Yeah, we are waiting on it before clearing the stack and what not

heytherewill: Though... Yeah, we might not 🤔

Toggl.Core.UI/Navigation/NavigationService.cs
```diff
@@ -29,8 +29,12 @@ public sealed class NavigationService : INavigationService
         public async Task<TOutput> Navigate<TViewModel, TInput, TOutput>(TInput payload, IView sourceView = null)
             where TViewModel : ViewModel<TInput, TOutput>
         {
-            var viewModel = await viewModelLocator.Load<TInput, TOutput>(typeof(TViewModel), payload);
-            await presenter.Present(viewModel, sourceView);
+            var viewModel = viewModelLocator.Load<TViewModel>();
+
+            await Task.WhenAll(
```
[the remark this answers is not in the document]
heytherewill: It is. The presenter needs to marshall and does a bunch of other things before actually showing the view. Testing yielded no weirdness, so all VMs are behaving

PR #5411, opened 2019-06-14
Handle remote config push notifications subscription

### LINKED ISSUE #5363: Add Firebase remote configuration flag for Push Notifications

daividssilverio:
## What's this?
This PR makes the push notification's subscription logic consider remote configuration.

### Relationships
Closes #5363
> Enable/disable sending sending the PN token to the server (put the configuration into the sync state)

## Why do we want this?
To be able to enable/disable push notifications subscriptions from firebase remote config.
We might want to use this if something goes wrong when handling push notifications in the mobile apps or backend.

## How is it done?
On android & iOS:
- It checks the remote config service before subscribing to receive push notifications.

During sync:
- On `SyncPushNotificationsTokenState`, it subscribes/unsubscribes (only when necessary) based on the remote configuration.

When un-Subscribing, it clears the firebase instance id and saved `PreviouslyRegisteredTokenKey` even if it fails to unsubscribe with backend.

It will not try to unsubscribe again if the `PreviouslyRegisteredTokenKey` is already null or different from what is received from `FirebaseInstanceId.Instance.Token`.
Why? To avoid wasting resources when calling `FirebaseInstanceId.Instance.DeleteInstanceId()`, which needs to be run outside of the main thread (🙈 droid was crashing...), and will trigger re-creation of the firebase instance id (and fcm token).

### Why not another way?
This makes sense to me. Feel free to make suggestions.

### Side effects
It seems that it made sync slower again... I need @simonrozsival or someone from @toggl/sync-guardians to take a look and try to understand what happened. It looks good to me.

## Review hints
Please do a functional test, not only code review.
Please test iOS unsubscription, droid was crashing, iOS could be as well.

## :squid: Permissions
🦑 after the tests are fixed sync looks alright.


Toggl.Core/Interactors/PushNotifications/UnsubscribeFromPushNotificationsInteractor.cs
```diff
@@ -33,17 +33,23 @@ ITogglApi togglApi
...
                 return pushServicesApi.Unsubscribe(currentToken.Value)
-                    .Catch<Unit, Exception>(_ => Observable.Return(Unit.Default));
+                    .Catch<Unit, Exception>(_ => Observable.Return(Unit.Default))
+                    .SelectMany(clearTokenReferences());
             }
 
             return Observable.Return(Unit.Default);
         }
+
+        private IObservable<Unit> clearTokenReferences()
```
simonrozsival: This can be void and be used from “Do”

daividssilverio: Good point.

daividssilverio:
> Why don’t we put the logic into the interactor? It seems that we have the same code in 3 places instead of one.

Sounds like a good plan, what about putting it into `SubscribeToPushNotificationsInteractor`?
`SubscribeToPushNotificationsInteractor` would use `UnsubscribeFromPushNotificationsInteractor` internally though.

simonrozsival:
I don't observe anything wrong with sync. Can you give me steps to reproduce it? Maybe a screen recording showing the problem? Console output (showing which sync state was entered at what time)?

daividssilverio:
Maybe it was fluke? I'll run again today and give you updates.

simonrozsival:
We have to go with Remote Config and we have to figure out what the problem on Android is also because we rely on RC in other places of the app and we can't ignore this problem anyway. I can confirm that there's no problem on iOS and that it works smoothly and it doesn't make it any longer (even after fresh install on my phone, when there's no cached value). Can you try this on `develop` and check if we didn't introduce the problem with the chagnes to the Xamarin.Firebase NuGet packages?

daividssilverio:
> We have to go with Remote Config and we have to figure out what the problem on Android is also because we rely on RC in other places of the app and we can't ignore this problem anyway. 

Ok.

> I can confirm that there's no problem on iOS and that it works smoothly and it doesn't make it any longer (even after fresh install on my phone, when there's no cached value). 

Of course iOS works... 

> Can you try this on develop and check if we didn't introduce the problem with the changes to the Xamarin.Firebase NuGet packages?

Yes, I'll keep investigating.

daividssilverio:
> Why don’t we put the logic into the interactor? It seems that we have the same code in 3 places instead of one.

@simonrozsival, I checked again and it doesn't make much sense to put the code in the interactor.

In the `SyncPushNotificationsTokenState`, it subscribes/unsubscribes based on the remote config.
In the `AppDelegate:DidReceiveRegistrationToken` and `TogglFirebaseIIDService:OnTokenRefresh`, it subscribes or does nothing, it doesn't unsubscribe.

PR #5412, opened 2019-06-16
🚧 Add userCalendar filtering and sort by offset ABS to calendar suggestions provider

### LINKED ISSUE #5382: Suggestions: Make sure calendar events from non selected calendars are not shown
Check and fix if needed that calendar events from calendars that are not selected in the calendar view are not shown as suggestions

### LINKED ISSUE #5413: Suggestions: Decide on  how many calendar based events are we going to show
At the moment we are showing all we can.
First we show calendar machine learning (1 or 0)
Then we show calendar events (0 - 3 based on the above and availability of calendar events) 
In the end we show suggestions based on frequency (0 - 3 based on the above and availability of enough time entries)

Max total is 3.

The question is should we constrain the calendar events to the 1 max, since it can completely remove from the UI the frequency based once?

@maritoggl what do you think?

sofokli1:
## What's this?
Add userCalendar filtering based on user calendar selections in calendar view or settings.
Also add sorting of results baed on ABS value of the offset from `now`
Shows at most 1 calendar suggestion

### Relationships
Closes #5382 
Closes #5413 

## Why do we want this?
We do not want users to see suggestions from calendar that they do not use in the app.

## How is it done?
Fetch the user calendars check which are selected by the user and use that to filter calendar events that we get from the calendar service

## :squid: Permissions
🔮 only


Toggl.Core/Suggestions/CalendarSuggestionProvider.cs
```diff
@@ -55,5 +74,16 @@ private Suggestion suggestionFromEvent(CalendarItem calendarItem, long workspace
...
+        private TimeSpan absOffset(CalendarItem item)
+        {
+            var currentTime = timeService.CurrentDateTime;
+            var startTime = item.StartTime;
+            var offset = currentTime - startTime;
+            return new TimeSpan(Math.Abs(offset.Ticks));
+        }
+
+        private IEnumerable<CalendarItem> orderByOffset(IEnumerable<CalendarItem> calendarItems)
+            => calendarItems.OrderBy(item => absOffset(item));
```
kornelijepetak:
```suggestion
            => calendarItems.OrderBy(absOffset);
```

Toggl.Core/Suggestions/CalendarSuggestionProvider.cs
```diff
@@ -55,5 +74,16 @@ private Suggestion suggestionFromEvent(CalendarItem calendarItem, long workspace
 
         private bool eventHasDescription(CalendarItem calendarItem)
             => !string.IsNullOrEmpty(calendarItem.Description);
+
+        private TimeSpan absOffset(CalendarItem item)
+        {
+            var currentTime = timeService.CurrentDateTime;
+            var startTime = item.StartTime;
+            var offset = currentTime - startTime;
+            return new TimeSpan(Math.Abs(offset.Ticks));
```
kornelijepetak: Oh, there's no need to do it like this.
`TimeSpan` has a `Duration` method.

So, you can do this instead:
`return (currentTime - startTime).Duration();`

Toggl.Core/Suggestions/CalendarSuggestionProvider.cs
```diff
@@ -41,12 +47,25 @@ public IObservable<Suggestion> GetSuggestions()
...
+            var selectedUserCalendars = interactorFactory
+                .GetUserCalendars()
+                .Execute()
+                .Select(calendars => calendars.Where(c => c.IsSelected));
+
+            var eventsFromSelectedUserCalendars = eventsObservable.WithLatestFrom(selectedUserCalendars,
+                    (calendarItem, calendars) =>
+                        (calendarIteam: calendarItem, userCalendarIds: calendars.Select(c => c.Id)))
+                .Where(tuple => tuple.userCalendarIds.Contains(tuple.calendarIteam.CalendarId))
+                .Select(tuple => tuple.calendarIteam);
```
kornelijepetak: typo `Iteam`.

Toggl.Core/Suggestions/CalendarSuggestionProvider.cs
```diff
@@ -41,12 +47,25 @@ public IObservable<Suggestion> GetSuggestions()
...
+
+            return interactorFactory.GetDefaultWorkspace().Execute()
                 .CombineLatest(
-                    eventsObservable,
+                    eventsFromSelectedUserCalendars,
                     (workspace, calendarItem) => suggestionFromEvent(calendarItem, workspace.Id))
+                .Take(1)
```
kornelijepetak: Can you please move this literal into a constant, such as `maxCalendarSuggestionCount` or something like that?

kornelijepetak:
Otherwise, it works. I no longer see the event suggested when I turn off a calendar in the settings.

Toggl.Core/Suggestions/CalendarSuggestionProvider.cs
```diff
@@ -18,6 +18,8 @@ namespace Toggl.Core.Suggestions
 {
     public sealed class CalendarSuggestionProvider : ISuggestionProvider
     {
+        private int maxCalendarSuggestionCount = 1;
```
kornelijepetak: `readonly` or `const`.

PR #5421, opened 2019-06-17
Update tag names in release checklist

juanlaube:
## What's this?
Update tag names in the release checklist to use `android` and `ios` instead of `giskard` and `daneel`.

## Why do we want this?
We want clear and up to date docs.

## How is it done?
📝 

## :squid: Permissions
🦑


heytherewill:
Please only merge when bitrise has been updated :)

PR #5428, opened 2019-06-17
Revert Use remote config to handle push notifications

daividssilverio:
## What's this?
This reverts commit 3009130c447f09f4fdea50466040543cbf02f51b. (#5377)

### Relationships
Closes nothing

## Why do we want this?
Because Firebase remote config is misbehaving and taking 60 seconds to fetch remote configuration during the initial sync.
It'll take a similar amount of time to fetch the values when the firebase cache is invalidated.

## How is it done?
This just reverts the commit generated by #5377 

### Why not another way?
I tried updating our `Xamarin.Firebase` libraries... but they are messed up.
When methods are not present (e.g. `FirebasePerformance.Instance.NewTrace`), it adds too many libraries and it 💥 our method count... which makes us have to enable MultiDex, which of course isn't working.

## :squid: Permissions
@simonrozsival only.
If this is merged, please close #5411 as well.


simonrozsival:
We'll have to ship this with Remote Config, so I'm afraid we can't get rid of it :( I'm closing this PR now.

PR #5433, opened 2019-06-18
Allow app to be installed in an SD card

### LINKED ISSUE #5431: Allow the app to be installed in the SD card
We don't have `android:installLocation` in our manifest and that's why the app can only be installed in the internal storage. Should be easy to fix.

arturs105:
## What's this?
This PR makes it possible to install our app in the external storage.

### Relationships

Closes #5431

## Why do we want this?
Why not? Also: got complaints from a user about this.

## How is it done?
With `android:installLocation="auto"`

### Why not another way?
'Tis the only hway.

## Review hints
❗️ **Must** be tested on an actual device with an SD card. Once tested, post a comment saying that you were able to move the app to the SD card. ❗️ 

## :squid: Permissions
🦑, but only after the app has been successfully installed in external storage ☝️


Toggl.Droid/Properties/AndroidManifest.xml
```diff
@@ -1,5 +1,11 @@
-<?xml version="1.0" encoding="utf-8"?>
-<manifest xmlns:android="http://schemas.android.com/apk/res/android" android:versionCode="987654321" android:versionName="2.0.1" package="com.toggl.giskard.debug">
+﻿<?xml version="1.0" encoding="utf-8"?>
+<manifest
+    xmlns:android="http://schemas.android.com/apk/res/android"
+    android:versionCode="987654321"
+    android:versionName="2.0.1"
```
amulware: I'd approve the PR for this change alone - makes it much easier to quickly find and change the version number for releases! :D

arturs105: You mean splitting the properties into separate lines? Let's hope our build/release flows can handle newlines and don't fall back to using spaces 😄

amulware: They will be fine. The replace the "987654321" string only.

amulware: They will be fine. They replace the "987654321" string only.

PR #5441, opened 2019-06-18
Modify test template

Odrakir:
## What's this?
I just want to modify the test template to remove tests of stuff that's not working (and should be fixed) or tests that are a bit of an overkill.

I'm also adding some tests for Siri.

### Relationships
Doesn't close any issue.

## Why do we want this?
To make manual testing better and less painful.


.github/ISSUE_TEMPLATE/d-manual-test-template.md
```diff
@@ -174,12 +177,8 @@ about: A manual testing checklist that can be assigned to a release to track the
 - [ ] Does it show up correctly on web?
 - [ ] Set timezone to JP and start and stop an entry (does it show up correctly?)
 - [ ] Does it show up correctly on web?
-- [ ] Set timezone to AU and start and stop an entry (does it show up correctly?)
```
heytherewill: Aren't these countries in this list because they are the ones that use Toggl the most? What was the reasoning for removing some?

Odrakir: Mainly that it is a pain to test all those countries, I'm not sure if there's a reason why this should work for some and fail for others...

heytherewill: Historically, we've had bugs that only happened in Estonian (the irony). I agree with you that this is maybe too much, I just wondered why you chose to remove the ones you did

Odrakir: Did those bugs really only happened in Estonia or that's the only place where we happened to test and catch those bugs?

kornelijepetak:
Resuming control in OP's absence in order to carry the PR to fruition.

kornelijepetak:
Regarding the discussion between @Odrakir and @heytherewill, I think all zones should stay. I believe it's a valid thing to test as many zones as possible. But it's also important to notice that if a release does not include any changes related to anything that could be related to timezones and such, I think it doesn't have to be tested as thoroughly as when there are some changes that touch time-related code.

I'd like your opinions on that.

1. Should we remove some zones as @Odrakir proposed
2. Should we leave things as they are now, and let RM decide whether to test this thoroughly considering the content and the nature of a particular release
3. Should we change it to say something like: `[ ] Test whether the app works correctly regardless of the zone` in order to leave the decision in the hands of RM, having the particular release in mind. (This is basically like 2. but with less explicit time zones).

PR #5445, opened 2019-06-18
Unify the way we close ViewModels

### LINKED ISSUE #5241: Unify the way we close ViewModels
## 🚧 Refactor Proposal
Following what was being done by MvvmCross, we currently have more than one way of closing our VMs (Finish method, Cancel method and a couple of `UIAction Close`). Once MVExit is merged, we'll have full control over the base ViewModel and then we can add a unified way of closing our VMs

## Platforms affected:

- [X] 🤖 Android
- [X] 🍎 iOS
- [X] 🖥️ Shared layer

## Technical aspects
IMO we should add a `UIAction Close` in the BaseViewModel that attempts to close the VM using the default value. The default value is protected so implementation can provide it. This will allow us to kill Finish, Cancel and the custom UIActions because now VMs can provide default values other than `default(TOutput)`

heytherewill:
## What's this?
This unifies the way we close ViewModels

### Relationships
Closes #5241 

## Why do we want this?
We currently have many ways of closing a VM. Now we only have one

### Why not another way?
I'm open to suggestions

## Review hints
Go over it commit-by-commit and please do make a functional test.

## :squid: Permissions
Only myself


Toggl.iOS/Presentation/RootPresenter.cs
```diff
@@ -115,7 +115,8 @@ private void detachOldRootViewController(UIViewController viewController)
         private void detachViewModel<TViewModel>(TViewModel viewModel)
             where TViewModel : IViewModel
         {
-            viewModel?.Cancel();
+            viewModel?.DetachView();
```
juanlaube: You are detaching the view twice 👀

Toggl.iOS/ViewControllers/Reactive/ReactiveNavigationController.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System.Linq;
+using Toggl.Core.UI.Views;
+using UIKit;
+
+namespace Toggl.iOS.ViewControllers
+{
+    public class ReactiveNavigationController : UINavigationController
```
juanlaube: You must use this class instead of `UINavigationController` in `ViewControllerLocator.wrapInNavigationController`

Toggl.Core.UI/ViewModels/Base/IViewModel.cs
```diff
@@ -16,6 +15,6 @@ public interface IViewModel
         void ViewDisappeared();
         void ViewDestroyed();
 
-        Task Cancel();
+        void CloseWithDefaultResult();
```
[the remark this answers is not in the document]
heytherewill: Because that is dangerous, especially for ViewModels that return Unit. I'd rather be explicit

PR #5452, opened 2019-06-19
Split manual testing template into automated and manual tests

arturs105:
## What's this?
I went over all the manual tests and cross-referenced them with UI tests to find, which manual tests are already automated. Then i split all the manual tests into 2 groups: automated and manual.

## Why do we want this?
:sanic: releases

## :squid: Permissions
🦑


arturs105:
>Are we sure that the UI tests work well enough to test all these things? They run in staging, right? Are we confident that is representative of the app in production?

I see the problem here. Manually testing the app also works partly as integration testing, and part of that would be lost with plain UI tests. We could try running them in production, but that would create a lot of garbage data. Though some tests will have to be done manually anyways, like checking if the web app shows the same data. Staging and production shouldn't be that different, so i think the few manual tests should be enough to verify that production is working ok.

Now regarding UI tests working well enough to test all these things. I think UI tests _can_ test these things better than a real person. Especially if there is only a single person testing and he has to execute tens of tests. At some point the attention is bound to decrease and things can go unnoticed. I emphasised that UI tests "can" be better, because this depends on assertions in UI tests - we need lots of those and they have to be good.

>Do the manual tests cover enough ground that we will notice any non-functional issues (layout glitches, etc.) during the automatically tested features?

This depends on the nature the issues. Since we are gonna add more assertions to UI tests, we will be able to detect if something is not showing up, or if something that shouldn't be visible is visible and maybe even if something is off-screen on smaller devices (gonna test this out sometime soon).

I'll also look into getting screenshots from each step of the test. The RM could glance over those and notice if anything is horribly wrong, though it's true that it would also be possible to miss a lot of things. 🤔 We could even use machine learning, coding, and algorithms to compare these screenshots against a set of "good" screenshots, but that's Loco...

Issue #5340, opened 2019-06-07
Write documentation for push notifications


simonrozsival:
- [x] Firebase settings documentation
- [x] API related documentation
- [x] Shared code documentation (FCM token storage, FCM token syncing, interactors)
- [x] iOS specific documentation
- [x] Android specific documentation

simonrozsival:
Add the docs into the `/docs/push-notifications.md` document.

daividssilverio:
Droid docs are ready on #5480.

simonrozsival:
All docs were merged into a project branch.

PR #5455, opened 2019-06-19
Schedule push notifications handling

### LINKED ISSUE #5450: Sync doesn't work when the app is in background or closed
After testing the droid app, I've found that:
- when the app is in foreground: runs like a charm!
- when the app is in background:
-- if the app was in foreground in the last few seconds/minutes: works alright
-- if the app was closed: nope... doesn't work
-- if the app has been in background for like 10 mins or so: nope... doesn't work

I suggest we either create a `JobService` to light sync the app in background with more time and actually finish it properly.

We can actually throttle incoming push notifications while the app is in background and avoid consuming more battery power than we should.

Another idea is to mark the app somehow to trigger a sync the next the app is opened.

daividssilverio:
## What's this?
This moves the handling of push notifications to an scheduled job

### Relationships
Closes #5450

## Why do we want this?
This way we can take more time to handle incoming push notifications.
It works well even when the application is closed (not in the recents screen). It will schedule a job as soon as possible (when there's internet available).

~It also works well when the application is already running and in foreground because of the way the job was scheduled (`SetImportantWhileForeground`).~

**UPDATED**
It also works well when the application is already running and in foreground because of the way the job was scheduled (`SetImportantWhileForeground`) on devices with Android Pie and above.

On devices with android versions prior to Android Pie, it still runs quickly enough. 
I've tested on 4 devices so far: Samsung Galaxy Note 9 (Pie), Samsung Galaxy S8 (Pie), Nexus 5 (Android 6.0.1) and Positivo Twist Mini (Android Go 8.1).

Specially on lower-end devices, syncs scheduled in background tend to run when the app is resumed from background.

## How is it done?
It moves all the hard work to another service, which will run all the sync logic in a background thread.
Now the entry point of the push notifications on droid will only be responsible for scheduling a job service.
I've added an extra step to avoid multiple jobs being scheduled at the same time.

### Why not another way?
Because we have time limitations to handle incoming push notifications.
Scheduling to run as soon as possible is the best choice here.

### Side effects
There shouldn't be any.

## Review hints
Review by commit.

## :squid: Permissions
🦑 it


Toggl.Droid/Services/TogglFirebaseMessagingService.cs
```diff
@@ -1,51 +1,39 @@
...
         }
-
-        private TogglApplication togglApplication() => (TogglApplication) Application;
-
-        protected override void Dispose(bool disposing)
+        
+        private void scheduleJob(AndroidDependencyContainer dependencyContainer)
         {
-            base.Dispose(disposing);
+            bool hasPendingJobScheduled = dependencyContainer.KeyValueStorage.GetBool(SyncJobService.HasPendingJobScheduledKey);
```
simonrozsival: I don't want to overengineer this, but this isn't thread safe. I don't think it's a good idea to use the key-value storage as a mutex. Can't we check if the service is already running in a different way?

Toggl.Droid/Services/TogglFirebaseMessagingService.cs
```diff
@@ -1,51 +1,39 @@
...
-
-            var shouldHandlePushNotifications = dependencyContainer.RemoteConfigService.ShouldHandlePushNotifications(); 
 
-            syncDisposable = shouldHandlePushNotifications
-                .SelectMany(willHandlePushNotification => willHandlePushNotification
-                    ? syncInteractor.Execute().SelectUnit()
-                    : Observable.Return(Unit.Default))
-                .ObserveOn(dependencyContainerSchedulerProvider.BackgroundScheduler)
-                .Subscribe(_ => StopSelf());
+            scheduleJob(dependencyContainer);
```
simonrozsival: I don't think we need an extra method here, I would put ewverything into `OnMessageReceived`.

Toggl.Droid/Services/SyncJobService.cs
```diff
@@ -0,0 +1,63 @@
...
+namespace Toggl.Droid.Services
+{
+    [Service(
+        Name = "com.toggl.giskard.SyncJobService",
+        Permission = "android.permission.BIND_JOB_SERVICE",
+        Exported = true)]
+    public class SyncJobService : JobService
+    {
+        public const string HasPendingJobScheduledKey = "HasPendingJobScheduledKey";
+        public const int JobId = 1111;
```
simonrozsival: We should maybe put these job IDs into some static class with constants so that we can keep track of them - the `BackgroundSyncJobSchedulerService` uses `JobId = 1`.

Toggl.Droid/Services/SyncJobService.cs
```diff
@@ -0,0 +1,63 @@
...
+
+        public override bool OnStartJob(JobParameters @params)
```
simonrozsival: Please rename this to `parameters`. We shouldn't use `@` in our manually written code whenever possible.

Toggl.Droid/Services/TogglFirebaseMessagingService.cs
daividssilverio: That part of the code only runs in the main thread of the application and it's a blocking call.
It think it's safe to use it in this context.

simonrozsival: What if the app crashes while it's executing the job and it doesn't set the state back to `false`? Will it mean that the user's phone won't be able to receive PNs until he reinstalls the app?

simonrozsival: In this context, I think it should be a static variable with a static lock, which will be used to lock the critical sections.

daividssilverio: But a static variable would be lost when the application is killed and multiple jobs would be scheduled that way.

We need something that is persisted.

daividssilverio:
```
public override bool OnStopJob(JobParameters @params)
        {
            AndroidDependencyContainer.Instance
                .KeyValueStorage
                .SetBool(HasPendingJobScheduledKey, false);

            return false;
        }
```

I've added that ☝️ to reset the key when something goes wrong.
I did not force an exception in the middle of the execution of the job, or tried to kill the application in the middle of its execution. I'll try that.

heytherewill: Also, why are we even checking that? Shouldn't overscheduling be auto-throttled by the system?

daividssilverio: That's for when the app is in foreground. It actually schedules the service as soon as possible and ends up running multiple times.

When it's in background and not run immediately, multiple jobs scheduled with the same jobId will run only once.

simonrozsival: I think it's OK if it runs multiple times. It's much better than deadlocking PN updates. The "lock" should at least expire somehow - we can save the time when it was locked and if the lock is older than 1h, it can be ignored.

daividssilverio: You are right, it deadlocks when an exception is thrown from inside the job service... and `OnStopJob` isn't called.

I'll update the code to take care of that. Good eyes @simonrozsival . Thanks.

Toggl.Droid/Services/SyncJobService.cs
```diff
@@ -0,0 +1,61 @@
...
+        Name = "com.toggl.giskard.SyncJobService",
+        Permission = "android.permission.BIND_JOB_SERVICE",
+        Exported = true)]
+    public class SyncJobService : JobService
+    { 
+        public override bool OnStartJob(JobParameters @params)
+        {
+            Task.Run(() =>
+            {
+                var dependencyContainer = AndroidDependencyContainer.Instance;
```
kornelijepetak: Please extract this into a function. It makes debugging easier and helps with levels of abstraction.

Toggl.Droid/Services/SyncJobService.cs
```diff
@@ -0,0 +1,61 @@
...
+        public override bool OnStopJob(JobParameters @params)
+        {
+            AndroidDependencyContainer.Instance
+                .KeyValueStorage
+                .SetBool(HasPendingSyncJobServiceScheduledKey, false);
+
+            return false;
+        }
+
+        private TogglApplication getTogglApplication() => (TogglApplication) Application;
```
kornelijepetak: Let's not change the member type here, leave it as a property.

```suggestion
        private new TogglApplication Application => base.Application as TogglApplication;

```

daividssilverio:
**Notes about recent changes**
On devices with android versions prior to Android Pie, it still runs quickly enough. 
I've tested on 4 devices so far: Samsung Galaxy Note 9 (Pie), Samsung Galaxy S8 (Pie), Nexus 5 (Android 6.0.1) and Positivo Twist Mini (Android Go 8.1).

Specially on lower-end devices, syncs scheduled in background tend to run when the app enters foreground.

simonrozsival:
> Specially on lower-end devices, syncs scheduled in background tend to run when the app enters foreground. 

I guess there's nothing we can do anything about it.

I'm still not very happy about using the storage for the lock (have you ever used Libre/Open Office and has it crashed? It could be a pain to re-open it afterwards... 😆) but I understand how it is meant to work and I'm glad that there's at least the recovery after 1 hour.

We should definitely describe this behavior in the docs (#5340) and we should remember that we have this code in the app in case it starts derping.

I have one thought 🤔 maybe it would be interesting to track how often it deadlocks. We could add an analytics event into the part of code which checks if the lock is older than 1 hour. If it is older than one hour, than this is a potential deadlock and we could track it. Do you think it would be an overkill?

daividssilverio:
> I'm still not very happy about using the storage for the lock.

I understand your concern. Still, I think this is needed. If you have a safer idea, please show me.

> We should definitely describe this behavior in the docs (#5340)

👌 

> I have one thought 🤔 maybe it would be interesting to track how often it deadlocks. We could add an analytics event into the part of code which checks if the lock is older than 1 hour. If it is older than one hour, than this is a potential deadlock and we could track it. Do you think it would be an overkill?

 I don't think it's an overkill.

PR #5456, opened 2019-06-20
Remove Firebase from debug builds

### LINKED ISSUE #5348: Make firebase configuration optional
Running the app in the simulator or on a device without the (firebase) keys properly substituted causes the app to crash during launching.

We must make this setup conditional again.

Odrakir:
## What's this?
Removes the line that configures Firebase from debug builds again.

### Relationships
Closes #5348 

## Why do we want this?
If we don't do that and don't have the appropriate info in `GoogleService-Info.plist` the app will crash every time we try to debug.

## How is it done?
After much thinking I think this simple line is our best option. Of course it prevents the whole thing with push notifications from working on debug. But everything else works.

## :squid: Permissions
PNs


Toggl.iOS/Startup/AppDelegate.cs
```diff
@@ -22,9 +22,11 @@ public override bool FinishedLaunching(UIApplication application, NSDictionary l
             #if !USE_PRODUCTION_API
             System.Net.ServicePointManager.ServerCertificateValidationCallback
                 += (sender, certificate, chain, sslPolicyErrors) => true;
-            #endif
-            
+#else
```
sofokli1: indentation of `#else` and `#endif` is not the same level as `#if !USE_PRODUCTION_API`

Odrakir:
I've opted for checking for `DEBUG` constant.

simonrozsival:
This works. There's just a problem that somebody might change which build configuration defines DEBUG and which does not. Let's document this in `docs/push-notifications.md` so that we remember what to fix when it breaks.

Odrakir:
@simonrozsival I'll update the docs, but... how's anyone going to change which build config defines DEBUG? Why will it be defined in any config other than debug?

PR #5459, opened 2019-06-20
📦 Detect the install location

### LINKED ISSUE #5435: Track if users install the app internally or externally
Since #5433 allows the app to be moved/installed to the SD card, we can measure, how many users actually do that.

A simple check on startup should be enough. I found [this](https://stackoverflow.com/questions/6256751/android-get-install-location-with-packagename) SO post about getting install location, which might help.

❗️ Though beware of OEMs like Samsung, Huawei etc. - the install location may be different on different devices. This needs to be tested on as many different devices as possible, and best if they are from different manufacturers. ❗️

### LINKED ISSUE #5436: Include install location in Submit Feedback data
Ref: #5433 

Since installing the app in an SD card will reduce performance, the install location could be useful data to include in the feedback report, especially if users are complaining about poor performance, slow startup times etc.

kornelijepetak:
## What's this?
Enables detection of the install location (internal memory or external storage).

* Sends this information with submit feedback form
* Tracks the location

ℹ Has no effect on iOS.

### Relationships
Closes #5435 
Closes #5436

## Why do we want this?
More information

## How is it done?
PlatformInfo.

### Why not another way?
That's where it belongs.

### Side effects
Should be none.

## Review hints
Go commit by commit and it will be super simple, I promise.

## :squid: Permissions
Homo sapiens. 
⚠ But only after we have at least one functional test on a device with an external sd card.


Toggl.Droid/PlatformInfoAndroid.cs
```diff
@@ -7,11 +9,32 @@ public sealed class PlatformInfoAndroid : BasePlatformInfo
     {
         public override string TimezoneIdentifier => TimeZone.Default.ID;
 
-        public PlatformInfoAndroid()
+        private Context context;
```
heytherewill: Please don't cache this. Instead, use `Application.Context` when you need to

PR #4957, opened 2019-04-19
⬆ Update ruleset

kornelijepetak:
## What's this?
Proposition to activate more stylecop rules.

### Relationships
Discusses #1947.

## Why do we want this?
Code quality and streamlining.

## How is it done?
This PR changes the .ruleset file to a proposition that needs to be discussed.
The comment below gives the table to make things simpler.

### Side effects
The code can't be built with this ruleset because it breaks some of them. Let's first discuss and agree on the ruleset, then I will fix the code to conform.

## Review hints
I've tried to help with the table as much as I could.

## :squid: Permissions
Not yet.

PR #5460, opened 2019-06-20
Update style cop ruleset

heytherewill:
## What's this?
This brings order to our codebase

### Relationships
Ref #1947
Supersedes #4957

## Why do we want this?
Better code quality

## How is it done?
I enable the rules and then I fix the violations that appear in our codebase; easy peasy

### Why not another way?
We discussed the rules that should be enabled in 1947. I changed the rule that force System to come first so I could use Visual Studio's auto formatting tool to auto clean our usings and now they are all in alphabetical order.

## Review hints
Review by commit

## :squid: Permissions
Me


Toggl.Tests.UI/Helpers/Attributes.cs
```diff
@@ -2,15 +2,16 @@
 
 [AttributeUsage(AttributeTargets.Assembly | AttributeTargets.Class | AttributeTargets.Method, AllowMultiple = false, Inherited = false)]
 public sealed class IgnoreOnAndroidAttribute
-    #if __DROID__
-    : NUnit.Framework.IgnoreAttribute { }
-    #else
+#if __DROID__
+    : NUnit.Framework.IgnoreAttribute
+{ }
```
sofokli1: why this are down here but in line 9 are inline with `: Attribute`

Toggl.Tests.UI/Helpers/Attributes.cs
```diff
@@ -19,6 +20,7 @@ public sealed class IgnoreOnIosAttribute
 
 }
 #else
-    : Attribute { }
+    : Attribute
+{ }
```
sofokli1: line 9 is different

PR #5462, opened 2019-06-21
Separate remote config reading and update

### LINKED ISSUE #5446: Let's not block anything because of RemoteConfig
## 🚧 Refactor Proposal
During #5411, I noticed that we are using Firebase remote config in a not ideal way.

We are running and the remote config service and fetching values from Firebase in the main views of the app. This isn't bad in the case of the `RatingView` because it subscription runs in background and it shouldn't be blocking anything.
Firebase should be keeping the local cache and only making http requests every 12 hours.

The problem is more serious when we put the remote config calls inside sync states.
More specifically, in #5411 `SyncPushNotificationsTokenState` queries the remote config to check if we should have the app's FCM token registered with the server.

During my tests, even with firebase's remote config developer mode enabled, the remote config takes about 60 seconds to finish fetching the values. 

It sounds a lot like it, and we are most likely just being throttled; and this shouldn't happen in production. Still, in normal circumstances, it will make an http request and take whatever time it takes to complete it every 12 hours. This will slow down sync every 12 hours.

This is also a problem when handling received push notifications. On #5377, the remote config dependency was added to check wether or not we should handle incoming push notifications.
This can happen when the app is in background/not open on android, where there's an execution limit of 10~20 seconds.

## Platforms affected:

- [x] 🤖 Android
- [x] 🍎 iOS
- [x] 🖥️ Shared layer

## Technical aspects
The idea is add another level of cache, handled by us in the application.
Then, we can asynchronously fetch firebase remote config when the application starts.
We can check if the remote config succeed in the last 12 hours or more, running it again if it didn't.
The app would use default values and internally cached values for most of the time.

Also, it seems that iOS isn't displaying that throttling problem I've commented above.
I've also tested firebase with our credentials with the updated Firebase libraries on a kotlin project and it works alright there.

So... even after doing all of this, we might want to revisit this problem when we are able to update the Xamarin Firebase libraries without breaking the app (and the libraries's apis are present).

daividssilverio:
## What's this?
This PR, when done, should separate reading and updating data from our remote config.

### Relationships
Closes #5446

## Why do we want this?
Please read #5446.

But basically, we want this to be able to read a remote configuration fast and with predictable performance.
We can fetch the remote config data asynchronously from somewhere within the app lifecycles in a way that's unobtrusive to the user, every 12 hours or so. 

## How is it done?
The `RemoteConfigService` now lives in shared and only reads the `KeyValueStorage` and exposes getters for cached remote configurations.

The new `IRemoteConfigUpdateService` implementations basically do the same thing as before, but instead of providing observables to the remote config values, it exposes an observable to let know interested parts that the remote config has been updated.

We can trigger the `FetchAndStoreRemoteConfigData` from the `IRemoteConfigUpdateService` to run asynchronously somewhere (maybe when application is resumed from background on both apps).
We can store the last time the cache was updated (not done yet in the droid and iOS implementations).

### Why not another way?
Please, give your suggestions. Close this and supersede it with a new PR if necessary.

### Side effects
There shouldn't be any.

## Review hints
I tried to keep the commits as tidy as possible.

## :squid: Permissions
🦑 after everybody is happy with the code, sync tests, unit tests and the sync diagram generation are fixed.

Note 1: the base is the `project/push-notifications`, but this could be merged into develop as well, after the project is done.
Note 2: I've based this PR's branch off `schedule-pn-handling` (#5455), that needs to be merged first.


Toggl.Droid/Services/RemoteConfigUpdateServiceAndroid.cs
```diff
@@ -0,0 +1,100 @@
...
+        {
+            Ensure.Argument.IsNotNull(keyValueStorage, nameof(keyValueStorage));
+            this.keyValueStorage = keyValueStorage;
+
+            RemoteConfigChanged = remoteConfigUpdatedSubject.AsObservable();
+        }
+        
+        public void FetchAndStoreRemoteConfigData()
+        {
+            var remoteConfig = FirebaseRemoteConfig.Instance;
```
simonrozsival: there should be an early return if it's already running (in some thread) so that we don't trigger this multiple times unintentionally. Use a `bool isFetching` + a lock to ensure atomicity of "compare and set" of the boolean. Also there has to be a big try-finally where the `isFetching` is unset.

The same should be done for the iOS service.

simonrozsival:
It looks reasonable to me. I have no problem using the storage as a cache, it seems reasonable.

This PR doesn't solve the main issue though - when to call `FetchAndStoreRemoteConfigData`. I think we should call it as soon as the user opens the app in a background thread, if it hasn't been called ever or if it is more than 12 hours since it was last called. The appropriate place seems to be the `MainViewModel` (the condition will be checked frequently, but the fetch won't be executed more than once per 12 hours).

daividssilverio:
@simonrozsival @kornelijepetak Yeah, this is a draft.
I said this in the PR's description:

> We can fetch the remote config data asynchronously from somewhere within the app lifecycles in a way that's unobtrusive to the user, every 12 hours or so.

> We can trigger the FetchAndStoreRemoteConfigData from the IRemoteConfigUpdateService to run asynchronously somewhere (maybe when application is resumed from background on both apps).
We can store the last time the cache was updated (not done yet in the droid and iOS implementations).

But anyway, I was thinking about putting this in the implementations of `IBackgroundService`.
As @simonrozsival said, we would check often, but we wouldn't fetch anything unless 12 hours have had passed.

Toggl.iOS/Services/RemoteConfigUpdateServiceIos.cs
```diff
@@ -0,0 +1,63 @@
...
+{
+    public class RemoteConfigUpdateServiceIos : IRemoteConfigUpdateService
+    {
+        private const string remoteConfigDefaultsFileName = "RemoteConfigDefaults";
+        
+        private IKeyValueStorage keyValueStorage;
+        public IObservable<Unit> RemoteConfigChanged { get; }
+
+        private readonly ISubject<Unit> remoteConfigUpdatedSubject = new Subject<Unit>();
+
```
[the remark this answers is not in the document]
daividssilverio: it's blackish for me.

Toggl.Core/Experiments/RatingViewExperiment.cs
```diff
@@ -14,28 +17,33 @@ public sealed class RatingViewExperiment
         private readonly ITogglDataSource dataSource;
         private readonly IOnboardingStorage onboardingStorage;
         private readonly IRemoteConfigService remoteConfigService;
+        private readonly IRemoteConfigUpdateService remoteConfigUpdateService;
 
         public IObservable<bool> RatingViewShouldBeVisible
-            => remoteConfigService
-                .RatingViewConfiguration
+            => remoteConfigUpdateService.RemoteConfigChanged
+                .StartWith(Unit.Default)
```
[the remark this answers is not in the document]
daividssilverio: Do you mean in the droid and iOS implementations?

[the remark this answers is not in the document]
daividssilverio: Well, I took as yes. Take a look again.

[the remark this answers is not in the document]
kornelijepetak: It was a yes 😄

Toggl.Core/Experiments/RatingViewExperiment.cs
```diff
@@ -22,7 +22,8 @@ public sealed class RatingViewExperiment
         public IObservable<bool> RatingViewShouldBeVisible
             => remoteConfigUpdateService.RemoteConfigChanged
                 .StartWith(Unit.Default)
-                .Select(_ => remoteConfigService.GetRatingViewConfiguration())
+                .SelectValue(remoteConfigService.GetRatingViewConfiguration())
```
simonrozsival: No. Don't do it like this. This will cache the value of `remoteConfigService.GetRatingViewConfiguration()` when the observable is created and it will always emit this value even if the remote config changes.

kornelijepetak: Oh, you are right. My bad.

daividssilverio: 🤔 What? 
`Select(_ => something())` it's the same as `SelectValue(something)`, right?

I'll update the tests to make sure these are working properly already.

simonrozsival: No, `SelectValue(something)` is the same as `var x = something; ... Select(_ => x)`.

simonrozsival: Or to give a better example: `SelectValue(something()) <=> var x = something(); ... Select(x)` - the method is called only once.

daividssilverio: Oh... I misunderstood this:
```
public static IObservable<U> SelectValue<T, U>(this IObservable<T> observable, U u)
            => observable.Select(_ => u);
```

Good catch @simonrozsival .

Toggl.iOS/Services/RemoteConfigUpdateServiceIos.cs
```diff
@@ -0,0 +1,87 @@
...
+                remoteConfigUpdatedSubject.OnNext(Unit.Default);
+
+                lock (updateLock)
+                {
+                    isRunning = false;
+                }
+            });
+        }
+        
+        public TimeSpan TimeSpanSinceLastFetch()
```
kornelijepetak: I dislike the fact this is using `default(DateTimeOffset)`.
I believe this should return `TimeSpan?` and return `null` if there is no value in the storage.
This will simplify this method from this obscure use of default DateTimeOffset.
Besides, this returns the value of 700k hours since 1.1.0001. which is theoretically an incorrect value. It is not true that this is the time since last fetch. So, the method should not be returning that.

Instead, handle this on the callsite either in this form (which is also semantically incorrect but is closer to the usage so it's understandable why it's like that)
```
var timeSinceLastRemoteConfigFetch = remoteConfigUpdateService.TimeSpanSinceLastFetch() ?? TimeSpan.MaxValue;
```

or utilize the fact that comparison of timespan and `null` is always false and invert the condition to:
```
var timeSinceLastRemoteConfigFetch = remoteConfigUpdateService.TimeSpanSinceLastFetch();
if (timeSinceLastRemoteConfigFetch < RemoteConfigConstants.RemoteConfigExpiration)
	return;
	
Task.Run(() => remoteConfigUpdateService.FetchAndStoreRemoteConfigData())
	.ConfigureAwait(false);
```

There's also a third option to be more explicit and that is to add a null check to make sure that the reader of the code understands the situation better.  

Any of these is better I think than returning `default(DateTimeOffset)` in this way. It's not as clear.
But since it will work, I will not mandate this change, because I will not be here to approve for two days.

daividssilverio: I've made `TimeSpanSinceLastFetch` return `TimeSpan?`. o/

PR #5466, opened 2019-06-21
Add tips for writing good UI tests to UI test docs

arturs105:
## What's this?
This adds guidelines on how to write good UI tests.

## Why do we want this?
For consistent, high quality UI tests.

## :squid: Permissions
🦑


docs/adding-ui-tests.md
```diff
@@ -139,6 +139,206 @@ public sealed class AmazingScreenTests
...
+//Don't
+[Test]
+public void StopRunningTimeEntry() { }
+
+//Don't
+[Test]
+public void StopRunningTimeEntryWorksCorrectly() { }
+```
+
+If the test method does lots of things and it's hard to come up with a meaningful and easily readable name that follows the pattern Action -> ExpectedBehaviour, then you can use the name of the test case scenario.
```
juanlaube: This also means that the test should probably be broken into smaller scenarios.

PR #5469, opened 2019-06-21
Fix crash in Siri for iOS < 12

### LINKED ISSUE #5438: Custom Siri intents are only available for iOS 12 +
## 💥 Crash Report
Custom Siri intents are only available for iOS 12 and up, which is causing the Siri section of settings to crash for users in iOS 11.

### 👥 Impact
30 users, 36 times.

### 🛰 Link
https://appcenter.ms/orgs/Toggl/apps/Toggl-iOS/crashes/errors/3965596390u/overview

Not much needed to add here, since the issue is pretty self explanatory.

juanlaube:
## What's this?
Fixes a crash introduced in iOS 1.22 and one of the most common crashes in iOS 2.0. Siri Shortcuts are not available before iOS 12.

### Relationships

Closes #5438 

## Why do we want this?
🚫 💥 

## How is it done?
Checking the system version and returning earlier when Siri Shortcuts are not available.

### Why not another way?
🤷‍♂ 

### Side effects
None

## :squid: Permissions
🚫 Only the release manager


Odrakir:
🤔 I'm thinking we should rather completely remove Siri Shortcuts and Workflows from the settings menu when iOS < 12. Otherwise we'll have a section that does nothing at all.

Toggl.iOS/Services/IntentDonationService.cs
```diff
@@ -30,6 +30,11 @@ public IntentDonationService(IAnalyticsService analyticsService)
 
         public IObservable<IEnumerable<SiriShortcut>> GetCurrentShortcuts()
         {
+            if (!UIDevice.CurrentDevice.CheckSystemVersion(12, 0))
+            {
```
heytherewill: No need for braces when returning

PR #5481, opened 2019-06-25
Make fragments undismassible

heytherewill:
## What's this?
This makes the fragments for no workspace view dismissable

## Why do we want this?
Once the user reaches this state, they can only dismiss these views when the workspace problem has been solved.

## How is it done?
Prevent OnCancel from running any code.

### Why not another way?
Glad to hear if there's some other way.

## Review hints
Test on an actual device and, if possible, wait for the debug view PR which will make testing this a lot simpler.

## :squid: Permissions
Only myself


daividssilverio:
I forced the navigation to the affected fragments, but could still dismiss them.

I guess setting the dialog `Cancellable` property in the `DialogFragmentPresenter` is the way to go here:

```
 protected override void PresentOnMainThread<TInput, TOutput>(ViewModel<TInput, TOutput> viewModel, IView sourceView)
        {
            ...
            var dialog = createReactiveDialog(viewModel);
            ...
            dialog.Cancelable = false;
            dialog.Show(fragmentManager, dialog.GetType().Name);
        }
```

something like that ☝️ , but only for the cancellable frags.

PR #5485, opened 2019-06-26
🚧 Replace bgImages in walkthrough with code generated images

### LINKED ISSUE #3868: Replace bgNoiseColor.imageset with a gradient
iOS has only three significantly big assets: `bgNoiseBlue`, `bgNoisePurple` and `bgNoiseYellow`. Together they weight almost 8MB and they are only used during the onboarding phase, so they are pretty much useless.

The easy solution would be simply to delete these files, but in order not to sacrifice A E S T H E T I C S, let's simply draw a gradient from color to transparent using code to achieve the same effect without having to make the user download<sup>1</sup> these images.

<sup>1</sup> Due to app thinning the users are probably not downloading all of that, but using code will nevertheless be lighter

sofokli1:
## What's this?
Replace some assets that took 8 mb with code that weighs nothing

### Relationships
Closes #3868

## Why do we want this?
we want the app to be even smaller and easy to download

## How is it done?
by deleteing the assests and creating similar once in code.

### Side effects
looks even better than before

## Review hints
log out and run the app to see the walkthrough. if you can't see a difference in the background, the job was done right.

## :squid: Permissions
:squid:


heytherewill:
Wouldn't it be simpler to use a `CAGradientLayer` instead of converting the whole thing into an image?

sofokli1:
@heytherewill replaced the image generation with a gradient layer. Feel free to review. The initial approach was something interesting to do but not the easiest thing to read as a dev.

Issue #5237, opened 2019-05-28
Monitor and fix OutOfRangeDateTimeExceptions


arturs105:
#5235 Introduced `OutOfRangeDateTimeException` to track the creation of date times which are out of range defined by Toggl business logic. That includes date times which cause unrepresentable date errors (Jan 1 0001)

We should check all the places, where out of range date times are created, especially date times that cause unrepresentable date errors.

juanlaube:
I've been looking `OutOfRangeDateTimeException ` and `OutOfBoundsDateTimeCreatedException` in AppCenter and I cannot find anything, this doesn't seem to be happening anymore. I will wait for @arturs105 to return from vacations before closing the issue just in case he has another idea.

arturs105:
Okay, seems like the problem was this: #5472. It's fixed and released in iOS 2.1, however only 1% of users have updated and the release is paused at this moment until we fix a sync issue, so we can't really know at the moment if it's actually fixed.

Now there is already some data in App Center from iOS 2.1 and the good news is that neither this nor other weird date related exceptions have occurred in 2.1. This might be because we have fixed the problem, but it could just as well be because not enough people have updated and we haven't received those exception yet.

Either way, moving this issue to a further release, when we will have more data to actually find out if we have fixed the problem.

arturs105:
I checked crashes and events for iOS 2.1 - 2.1.3 and now we have enough data to confidently say, that the problem is fixed and no longer occurs.

PR #5487, opened 2019-06-26
🔥Remove all tracking regarding unrepresentable dates

### LINKED ISSUE #5236: Remove all tracking regarding unrepresentable dates
Once we have fixed all the unrepresentable date errors, there will be no need to track exceptions and events regarding unrepresentable date times, so all code that handles that should be removed.

Here's a list of PRs that introduced the things that should be removed:
* #5023
* #5051
* #5235

sofokli1:
## What's this?
Remove all tracking regarding unrepresentable dates

### Relationships
Closes #5236

## Why do we want this?
looks like we don't need this anymore

## How is it done?
by 🔥

## Review hints
Take a look at the issues where it references the PRs that the code was initially added

## :squid: Permissions
Only me


Toggl.iOS/Views/EditDuration/WheelForegroundView.cs
```diff
@@ -370,14 +370,7 @@ private void updateEditedTime(TimeSpan diff)
...
             {
-                var exceptionProperties = new Dictionary<string, string>
-                {
-                    { "DateTimeOffset", currentDateTimeOffset.ToString() },
-                    { "TimeSpan", difference.ToString() }
-                };
-                var exception = new UnrepresentableDateException("Unrepresentable DateTimeOffset encountered in WheelForegroundView", argumentOutOfRangeException);
-                IosDependencyContainer.Instance.AnalyticsService.Track(exception, exceptionProperties);
-                return currentDateTimeOffset;
+                throw argumentOutOfRangeException;
```
heytherewill: Please don't rethrow exceptions, as this empties the call stack. If we are removing all the logging, this means that the error is fixed, correct? We can safely remove all exceptions in this case

Toggl.Core.UI/Transformations/DateTimeToFormattedString.cs
```diff
@@ -11,38 +11,27 @@ public class DateTimeToFormattedString
...
-                {
-                    { "DateTimeOffset", value.ToString() },
-                    { "TimeZoneName", timeZone.DisplayName },
-                    { "TimeZoneOffset", timeZone.BaseUtcOffset.ToString() }
-                };
-                var exception = new UnrepresentableDateException("Unrepresentable DateTimeOffset when converting to local timezone", argumentOutOfRangeException);
-                analyticsService.Track(exception, exceptionProperties);
-
-                throw exception;
+                throw argumentOutOfRangeException;
```
heytherewill: Same thing about rethrow

Toggl.iOS/Views/EditDuration/WheelForegroundView.cs
sofokli1: Maybe @arturs105 can comment if the error was fixed. I just know that the issue is in the current 2.1 release, so I am just executing the changes.

@arturs105 please review as well

@heytherewill will remove the rethrow

arturs105:
I was gone on vacation for quite a while, so i'm a bit out of the loop on this. Did we analyse the tracked data and did we find where the out of range date times are coming from?

sofokli1:
Personally was never in the loop about this. The issue says you added this to the 2.1 release via automation. @arturs105 who else was involved with this issue?

arturs105:
😅  Yeah, i put this in 2.1 project, though as a rough estimate, since then it was 2 releases away and i estimated that by 2.1 we'd have gathered enough data and analysed it. I guess i should have been more explicit about first getting to the bottom of the Mysterious date times and only then removing the tracking.

Anyways it's not like this PR was wasted time, this is still usable, though a bit later - when we find out (and fix), the thing that is/was creating those date times. Putting this on hold for now.

amulware:
@arturs105 	@sofokli1 	Status?

sofokli1:
@amulware based on this [comment](https://github.com/toggl/mobileapp/issues/5237#issuecomment-510869551), this PR will be dealt with in iOS 2.3

arturs105:
I went over crashes and events in iOS 2.1 - 2.1.3 and found nothing related to unrepresentable dates, so it's now time to move forward with this.

Toggl.iOS/Views/EditDuration/WheelForegroundView.cs
```diff
@@ -363,23 +363,8 @@ private void updateEditedTime(TimeSpan diff)
         private DateTimeOffset getNewDateTimeOffset(
```
arturs105: This method isn't needed anymore. It was introduced to do safe addition with error catching, but now that error catching is removed and this method has become basically this (can be just simplified by using `+` operator):
```
public int sum(int a, int b)
    => a + b;
```

sofokli1:
Not addressed yet

sofokli1:
All addressed now

PR #5489, opened 2019-06-26
Fix disappearing running/idle notifications

### LINKED ISSUE #5122: Notifications disappears when tapped
## 🐛 Bug Report
The `Running Timer` and `Stopped timer` notifications disappear when tapped.

## Context
Build/Branch: `develop` or `scandium (1.15)`
Phone OS: At least on Android 9
Phone model: Most likely unrelated

## Expected behavior
If the option is turned on, the `Running timer` notification should always be there, as long as there is a runnig time entry. Tapping the notification should turn on the app (or bring it into view if it's already running), but should not remove the notification.

## To reproduce
1. Go to settings and turn on the `Running timer` notification
1. Go to Timer view
1. Start a Time entry
1. _Notification should now be visible_
1. Pull down the status bar
1. Tap the notification 
1. _The app opens_
1. ⚠ _The notification is no longer there_

daividssilverio:
## What's this?
This fix the disappearing running/idle notifications reported on #51220

### Relationships
Closes #5122

## Why do we want this?
Because the users who enable running/idle persistent notifications want to see them at all times.

## How is it done?
There was a misplaced `notificationManager.CancelAll();` in the service that schedule calendar smart notifications. Since it runs every time the app is opened, it caused the persistent notifications to be cancelled when calendar events were being re-scheduled.

Now the notifications are being cancelled individually (a notification that has already been fired).

### Why not another way?
We could just drop the cancellation as well. The difference is that events being re-scheduled by the user and already generated a notification would stay in the notification tray.

### Side effects
None so far on my tests. Please do some functional testing.

## Review hints
None.

## :squid: Permissions
🦑 it


Toggl.Droid/BroadcastReceivers/SmartAlertCalendarEventBroadcastReceiver.cs
```diff
@@ -15,8 +18,16 @@ public override void OnReceive(Context context, Intent intent)
...
             var notificationId = intent.GetIntExtra(NotificationId, 0);
+            
+            var sharedPreferences = context.GetSharedPreferences(ScheduledNotificationsSharedPreferencesName, FileCreationMode.Private);
+            var savedNotificationIds = sharedPreferences.GetStringSet(ScheduledNotificationsStorageKey, new List<string>()).ToList();
 
-            notificationManager.Notify(notificationId, notification);
+            var removedNotificationIdsCount = savedNotificationIds.RemoveAll(id => id.GetHashCode() == notificationId);
+
+            if (removedNotificationIdsCount > 0) 
+                notificationManager.Notify(notificationId, notification);
```
heytherewill: Will this not remove the persistent notification if the app is triggered through it or does this service only handle the calendar notifications? 🤔

daividssilverio: No no, this affects only the calendar notifications.

There's no `CancelAll` anymore, the persistent notifications won't be cancelled.

PR #5495, opened 2019-06-27
ReportsViewModel CurrentDateRange tests

### LINKED ISSUE #4991: Add more tests for ReportsViewModel.CurrentDateRangeStringObservable
While working on #4873 i noticed that `ReportsViewModel.CurrentDateRangeStringObservable` wasn't properly tested - it only had 1 test. Now I'm adding a few more tests for the bug that i'm fixing, but more tests are needed.

Odrakir:
## What's this?
Adds some more tests to `ReportsViewModel` `CurrentDateRange`

### Relationships
Closes #4991 

## Why do we want this?
Tests are good

## How is it done?
I'm using `MemeberData` magic to generate a bunch of values for testing.

## :squid: Permissions
Go!


Toggl.Core.Tests/UI/ViewModels/ReportsViewModelTests.cs
```diff
@@ -229,37 +230,132 @@ public async Task IsInitializedToEmptyOrNull()
...
+                    var now = DateTimeOffset.Now;
+                    TimeService.CurrentDateTime.Returns(now);
+                    ViewModel.CurrentDateRange.Subscribe(observer);
+                    await ViewModel.Initialize();
+
+                    await ViewModel.LoadReport(0, startDate, endDate, ReportsSource.Calendar);
 
                     TestScheduler.Start();
-                    observer.LastEmittedValue().Should().Be($"{Resources.ThisWeek} ▾");
+                        observer.LastEmittedValue().Should().Be($"{result} ▾");
```
sofokli1: why the extra indentation?

Toggl.iOS/ViewControllers/ReportsViewController.cs
```diff
@@ -106,7 +106,8 @@ bool isWorkspaceNameTooLong(string workspaceName)
                 .Subscribe(WorkspaceLabel.Rx().Text())
                 .DisposedBy(DisposeBag);
 
-            ViewModel.CurrentDateRangeStringObservable
+            ViewModel.CurrentDateRange
+                .Debug("title")
```
sofokli1: 🔥

Toggl.Core.Tests/UI/ViewModels/ReportsViewModelTests.cs
```diff
@@ -229,37 +230,132 @@ public async Task IsInitializedToEmptyOrNull()
...
                 [Theory, LogIfTooSlow]
-                [InlineData(BeginningOfWeek.Monday)]
-                [InlineData(BeginningOfWeek.Tuesday)]
-                [InlineData(BeginningOfWeek.Wednesday)]
-                [InlineData(BeginningOfWeek.Thursday)]
-                [InlineData(BeginningOfWeek.Friday)]
-                [InlineData(BeginningOfWeek.Saturday)]
-                [InlineData(BeginningOfWeek.Sunday)]
-                public async Task EmitsThisWeekWhenCurrentWeekIsSelected(BeginningOfWeek beginningOfWeek)
+                [MemberData("allCombinations")]
```
arturs105: You can use `nameof(allCombinations)` here and in the other places as well.

Toggl.Core.Tests/UI/ViewModels/ReportsViewModelTests.cs
```diff
@@ -229,37 +230,132 @@ public async Task IsInitializedToEmptyOrNull()
...
+                }
+            }
+
+            public sealed class WhenARandomDateRangeIsSelected : ReportsViewModelTest
+            {
+                public static IEnumerable<object[]> randomCombinations
+                {
+                    get
+                    {
+                        yield return new object[] { new DateTime(2019, 5, 23), new DateTime(2019, 5, 29), "05/23 - 05/29" };
```
arturs105: Testing this with different date formats wouldn't hurt.

Odrakir: I thought about doing it, but we are not testing that really, I think it's an overkill

PR #5502, opened 2019-06-28
Revamp token reset vc

### LINKED ISSUE #5170: Revamp the looks of the TokenResetViewController
## 🚧 Refactor Proposal
That view is so outdated! It still has the original looks from 1.X. Time to change it

## Platforms affected:

- [ ] 🤖 Android
- [X] 🍎 iOS
- [ ] 🖥️ Shared layer

## Technical aspects
We need to update the Layout of the `TokenResetViewController` so that it matches the new Login/Signup screens.

daividssilverio:
## What's this?
This updates the token reset view controller to look more like our login/signup/forgot password views.

### Relationships
Closes #5170

## Why do we want this?
To get more visual consistency.

## How is it done?
With blood and tears. Seriously... I won't complain about using Android Studio to make android layouts ever again.

Other than that, the new layout is based a lot on the login storyboard.

### Why not another way?

### Side effects
I've removed the "next" button from the old layout. I'm not sure it was there... if it should stay... someone please say something.
I've also removed the logic to move the "or Sign out" button when the keyboard moves. The layout is a bit more cramped now and it could look bad on the iPhone SE if I had left that there.

## Review hints
Please do a functional test.

## :squid: Permissions
🦑 after some @toggl/ios-senseis reviewed this.


daividssilverio:
This is how it's looking now:
ezgif com-gif-maker (1)

That's how it looked before:
Simulator Screen Shot - iPhone 8 - 2019-06-28 at 01 35 45

PR #5505, opened 2019-06-28
💄Suggestion minor fixes

sofokli1:
## What's this?
Addresses changes requested to the suggestion PR

### Relationships
none

## :squid: Permissions
Only 🔮


Toggl.Core/Interactors/Suggestions/GetSuggestionsInteractor.cs
```diff
@@ -27,18 +27,35 @@ public sealed class GetSuggestionsInteractor : IInteractor<IObservable<IEnumerab
         public IObservable<IEnumerable<Suggestion>> Execute()
             => getsuggestionProvidersInteractor
                 .Execute()
-                .SelectMany(CommonFunctions.Identity)
+                .Flatten()
```
kornelijepetak: I am also not quite sure what exactly @heytherewill meant, but this whole query can be simplified to:

```
public IObservable<IEnumerable<Suggestion>> Execute()
    => getsuggestionProvidersInteractor
    .Execute()
    .Flatten()
    .SelectMany(provider => provider.GetSuggestions())
    .ToList()
    .Select(removingDuplicates)
    .Select(suggestions => suggestions.Take(suggestionCount));
```

Besides being a bit shorter, it has an added benefit of not tapping into RX for purely enumerable operators. In your case, you are conflating the stream into a single collection element (`:33`) so you can remove the duplicates, but then you are wrapping this into an observable with `SelectMany` to `Take(3)` in RX domain and then conflate it again, instead of working on enumerable itself. This is better this way because we never need a stream of individual suggestions, but return an observable of the whole collection.

PR #5506, opened 2019-06-28
✨Add UserID to AppCenter and Submit feedback

### LINKED ISSUE #5256: Pass user id to App Center
Earlier this year AppCenter introduced a useful search feature to the diagnostics section. Part of this is that we can not provide it with a user id, which can also be searched for. By using this, we will be able to check what errors specific users are experiencing and hopefully make life easier for us, support, and ultimately our users.

Here is the announcement post from App Center: https://devblogs.microsoft.com/appcenter/introducing-diagnostics-search-and-user-id-support/

Things we have to do:
- Tell AppCenter the user id on login
- Make AppCenter forget the user id on logout

And, to catch up on existing users:
- Tell AppCenter the user id on app start, if we're logged in

Whoever works on this please think through if all of this makes sense, and where the best places are to do this. Let's also test and make sure this works in both adhoc and store builds (and maybe should be disabled in debug, that's up to you to figure out).

kornelijepetak:
## What's this?
Adds User ID to

* AppCenter logs
* Submit feedback forms

⬆ Additionally updates the AppCenter to the latest version.
I've checked the release notes, we should be fine.

### ⚠ Important notice
There seems to be no way to literally 'make AppCenter forget user id', as the issue suggests. This ID (I think) is not something that is sent to AppCenter. That's something that's stored locally and whenever other things are sent (such as crashes), it just gets attached to the payload.

The thing I've done is _reset_ it to `""`? And the only reason to do it is to prevent id being sent when nobody's logged in. Examples being signup/login or forgot password screens... Although the chance for a change of id with next login is really rare (mobile phones are usually not shared).

Because it does not seem to be a value that AppCenter stores, unassigning the id will not change anything already sent to appcenter. As far as I can tell, there's no way to scrap that information from existing crashes.

As per [AppCenter docs](https://docs.microsoft.com/en-us/appcenter/sdk/other-apis/xamarin#identify-users), currently setting this id works only with crashes, but according to Microsoft/appcenter/issues/151 it seems that attaching the ID to the events might be implemented in the future.

### Relationships
Closes #5256 

## Why do we want this?
So we can better help the users who experience crashes.

## How is it done?
By calling `AppCenter.SetUserId`.

### Why not another way?
This seemed the best way.

### Side effects
None. One should note thought that if the report fails for whatever reason, it will not crash the app. And it should not because we can live without this id, there's no reason to treat this as crash-the-app error.

## Review hints
Easiest will be commit by commit.

## :squid: Permissions
Me.


Toggl.Core/Analytics/IAnalyticsService.cs
```diff
@@ -170,6 +170,8 @@ public interface IAnalyticsService
 
         IAnalyticsEvent<ApplicationInstallLocation> ApplicationInstallLocation { get; }
 
+        void ReportAppCenterUserId(long id);
```
juanlaube: A better name would be `SetAppCenterUserId` but I will the decision to you.

Toggl.Core.UI/ViewModels/LoginViewModel.cs
```diff
@@ -220,11 +225,19 @@ private async void onAuthenticated()
 
             onboardingStorage.SetIsNewUser(false);
 
+            await reportAppCenterUserId();
```
heytherewill: Do we really gotta `await` for this? I think this is irrelevant to the login process and can safely be a fire and forget operation

Toggl.Core.UI/ViewModels/MainViewModel.cs
```diff
@@ -163,6 +163,9 @@ public override async Task Initialize()
         {
             await base.Initialize();
 
+            var user = await interactorFactory.GetCurrentUser().Execute();
+            analyticsService.ReportAppCenterUserId(user.Id);
```
heytherewill: Why continuously report it when you already reported during login? The user id won't ever change

heytherewill: Either way, I don't think this operation should be awaited.

Toggl.iOS/Services/AnalyticsServiceIos.cs
```diff
@@ -65,6 +66,21 @@ public override void Track(Exception exception, string message)
             return validParameters;
         }
 
+        public override void ReportAppCenterUserId(long id)
+        {
+            try
+            {
+                AppCenter.SetUserId($"{id}");
```
heytherewill: What's wrong with `id.ToString()`?

Toggl.Core.UI/ViewModels/MainViewModel.cs
juanlaube: AppCenter doesn't persist the user id as far as I know, we have to set it per session.

kornelijepetak: The reason why we report this, as the issue states, is to catch up with users that are already logged in. There is an initiative to remove this code once the majority of ids are already in. How exactly we will know that, I don't know. Although we can track how many users have switched to the version that contains this code that does that in MainViewModel. Then, we remove this behavior and it happens only during login.

Toggl.iOS/Services/AnalyticsServiceIos.cs
kornelijepetak: Oh, nothing in particular.

PR #5516, opened 2019-07-01
Remove background overdraw from our views

heytherewill:
## What's this?
Another one of them papercuts, this removes the overdraw from our views.

## Why do we want this?
Overdraw means you are drawing things the user is not even seeing. Why would you do that to yourself?

## How is it done?
Removing the background drawables from our views and using the Window background drawable instead

### Why not another way?
No other way, but I'm open to removing even more if you happen to find opportunities.

## Review hints
Go to settings, search for `Overdraw` and then select `Debug GPU overdraw`. After enabling it, go to our app and it will be shown in shades of Blue, green and red.

Blue: This pixel is being drawn twice. This is fine.
Green: This pixel is being drawn three times. Usually fine as, but could mean something is wrong depending on where it's happening.
Red: This pixel is being drawn more than thrice. Unless there's a good reason for it to happen, it should be optimized.

## :squid: Permissions
Myself.


daividssilverio:
You've also done more than removing the overdrawn 🤔
Please describe the other changes in the PR.

heytherewill:
This has been properly updated to be up-to-date with develop

PR #5519, opened 2019-07-02
Android 2.1 - Baklava

kornelijepetak:
## User facing changelog
```
- Improved time entry suggestions
- Fixed sync failure sometimes being wrongly reported even on successful syncs
- Fixed a crash when detecting country on signup
- Added the possibility to install the app onto external storage
- Changed links to now open in the user’s default browser
- Removed avatar from settings
```

## Changes
[android-2.0.1...baklava](https://github.com/toggl/mobileapp/compare/android-2.0.1...baklava)

⚠ Important! Note that this release also includes push notifications, but they are disabled at the time of this release and will not reach user in `2.1`. However, when a future release comes around, the release manager should take care to add proper changes to the changelog, as those commits will not be visible on github compare link.


amulware:
I’d say:
‘When detecting country on signup’, to be more specific 
‘Added THE possibility’ since there is only one, not multiple
‘Changed links to now open in the user’s default browser’ grammar and more concise

Otherwise no complaints.

PR #5522, opened 2019-07-03
Morphing FAB animation

heytherewill:
## What's this?`
This adds ripple animations to the running TE card.

## How is it done?
A custom `FabTransformationSheetBehavior` + `CoordinatorLayout`

### Why not another way?
This is so much simpler than the current way it can also be considered a positive refactor!

### Side effects
No race condition checks 😄 

## Review hints
Review per commit
Test on device

## :squid: Permissions
Myself


simonrozsival:
GIF!

PR #5529, opened 2019-07-03
Remove swipe onboarding

heytherewill:
## What's this?
This removes swiping onboarding (both animations 

## Why do we want this?
This onboarding is bad and and confusing

## How is it done?
🔥 

## :squid: Permissions
Anyone, anytime


amulware:
Unless there's a technical reason not to, please proceed with this. We want it.

PR #5534, opened 2019-07-03
Add ripples, ripples everywhere on droid

daividssilverio:
## What's this?
This adds missing ripples to clickable items around the app.

### Relationships
Closes #Nothing

## Why do we want this?
Because it looks good.

## How is it done?
Drawable ripples are set either as background or foreground in some places, that's it.

### Why not another way?

### Side effects
Ripples

## Review hints
Please, test this on your phone/emulator.

## :squid: Permissions
🦑


Toggl.Core.UI/ViewModels/SuggestionsViewModel.cs
```diff
@@ -131,7 +131,8 @@ private IObservable<IThreadSafeTimeEntry> startTimeEntry(Suggestion suggestion)
 
             var timeEntry = interactorFactory
                 .StartSuggestion(suggestion)
-                .Execute();
+                .Execute()
+                .SubscribeOn(schedulerProvider.BackgroundScheduler);
```
kornelijepetak: Huh? Why is this in this PR?

daividssilverio:
@kornelijepetak 
> Username and Email in the settings do not need to have ripple. They should not be tappable as there is no interaction with those.

I couldn't find the guidelines on this, and the only example of list item that doesn't lead to another screen I could find is in the `Settings > About this phone`, in which case, some items have ripples but don't do anything.  So I rather keep the behaviour.

Toggl.Core.UI/ViewModels/SuggestionsViewModel.cs
heytherewill: Otherwise the ripple is not even visible

daividssilverio:
> Save button on Start TE view looks like it's going outside of it's container (similar to the callout sign)

The ripples are working there. The problem is the `EditText` with a solid background, which is being worked on `best-colors`.

kornelijepetak:
> I couldn't find the guidelines on this, and the only example of list item that doesn't lead to another screen I could find is in the `Settings > About this phone`, in which case, some items have ripples but don't do anything. So I rather keep the behaviour.

The [material design](https://material.io/design/interaction/states.html#pressed) states that: 
> _Pressed states are initiated by user keyboard or voice input on an interactive element._

 And also:
> _This state applies to all interactive components._


If an element does nothing when tapped, it is not an interactive element at all.

I still vote fore removing ripple from non-interactive elements.


btw. my `about phone` leads to another screen with phone data. I can't understand how the `about phone` wouldn't lead to another screen... 🤔  I mean, it should lead to a screen with data about phone.

Toggl.Core.UI/ViewModels/SuggestionsViewModel.cs
kornelijepetak: I'm still confused. Why is this so? Because the code otherwise executes on the UI thread and prevents ripple effect animation?

kornelijepetak:
> The ripples are working there. The problem is the `EditText` with a solid background, which is being worked on `best-colors`.

I did not say ripples are not working. I was listing all places where I found ripple either missing, or there, but garbled. This is the case of the latter, but as you said, it's being fixed at some other place.

kornelijepetak:
By the power vested in me by the original author of the PR, I'm resuming the PR fix/merge process.

PR #5539, opened 2019-07-04
Move settings to the tab bar

juanlaube:
## What's this?
Moves the settings to the tab bar, just like Android

### Relationships

Closes nothing

## Why do we want this?
Consistency

## How is it done?
1. Add the settings tab
1. Remove the settings button from the main view controller

### Side effects
We need to update the settings icon, looks small next to the other tabs.

## Review hints
Please test.

## :squid: Permissions
🦑


amulware:
Please make sure you get the image files you need and then feel free to go ahead with this.

arturs105:
Adding `on-hold` label until we get the correct icon.

juanlaube:
I had a chat with @maritoggl about this change and we won't merge it right now. The main argument is that the tab bar should be used for common navigation items, and settings is not one of those.

PR #5540, opened 2019-07-04
Customise swipping actions

Odrakir:
## What's this?
This adds the ability to customise main log cells swiping actions. For iOS 11+ it lets you change left and right swipes, for iOS 10- it's only the left swipe (the other one doesn't exist at all).

The possible values for now are: disabled, delete and continue.

This PR doesn't add the necessary UI to the settings section, so users can't change it just yet. This will be done in another PR, as I'm going to go ahead and refactor that so it's easier to add settings.

### Relationships
Doesn't close any issue, this is Buenos Aires.

## Review hints
In order to test it you can edit lines 111 and 113 of `SettingsStorage` setting the defaults for these values.

## :squid: Permissions
🚫 This is missing android. @heytherewill , are you taking care?


amulware:
On hold until we decide if we want to allow full customisation, or just enable/disabling swipe actions. @maritoggl has argued for the latter.

Odrakir:
I'll continue working on this as a decision has been made to have only 1 setting to enable/disable both swipes.

Odrakir:
Alright, I change it so it's just one setting, it'll make things easier. If we ever want to go back to full customisation we can add that again.

PR #5542, opened 2019-07-05
🚧 Rename FB remote config parameters

### LINKED ISSUE #3736: Change Firebase parameter names
Current firebase parameter names for user rating dialogs are not good. They are too vague.

I propose a two-fold change:

1. Add two new parameters with the same values and update the apps with the new parameters
2. Phase out (delete) old parameters when there's only an insignificant number of users on old versions.

Proposed change:
`criterion -> UserRatingDialogTrigger`
`day_count -> UserRatingDialogDelayInDays`

For now, number 1. is already done in code, but as of yet not part of any release.

When we are doing number 2. is completely up to @amulware.

This issue should be closed when BOTH of the points are acted upon.

kornelijepetak:
## What's this?
Renames FB remote config parameters into a more readable format.

### Relationships
Refs #3736

## Why do we want this?
So that code looks nicer and that there's less confusion in the future re: firebase remote config parameters.

## How is it done?
By sheer renaming.

### Side effects
None, but this requires an update to the Firebase parameters. A copy must be kept until all (or most significant number) of users have moved to a new version (and using the new parameters).

## Review hints
Not much to say.

## :squid: Permissions
🚫 Not yet, until I make sure Firebase is ready to go.


kornelijepetak:
⚠ Ok, the Firebase is ready for the change.
This PR can now be merged. 

#3736 should be closed only when the decided majority of users have updated to a version containing this PR's change.

Toggl.Droid/Resources/xml/RemoteConfigDefaults.xml
```diff
@@ -8,20 +8,12 @@
         <key>UserRatingDialogTrigger</key> 
         <value>none</value> 
     </entry> 
-    <entry> 
-        <key>day_count</key> 
-        <value>5</value>
```
sofokli1: aren't you removing the default values here? if yes, why?

kornelijepetak: Because those were superseeded by `UserRatingDialogDelayInDays` and `UserRatingDialogTrigger` ones above. We were dragging both versions for quite some time now.

PR #5543, opened 2019-07-05
Goodbye iOS 10

juanlaube:
## What's this?
Removes support for iOS 10.

## Why do we want this?
Supporting such an old version of iOS is painful and pollutes our codebase with in the effort to maintain compatibility, and iOS 13 is around the corner, supporting 4 major versions will make things even harder.
The users still running iOS 10 are either on devices not supported by iOS 12 or for some reason refuse to update.

## How is it done?
👋 Setting the minimum supported version to iOS 11.0. 

### Why not another way?
🤷‍♂ There's no other way, either we support iOS 10 or we don't.

### Side effects
⚠️ Approximately 3% of active users are still on iOS 10 according to Firebase. The decision is not only to the development team, but also needs product's input.

## :squid: Permissions
🚫 Only after @amulware and @maritoggl approval


Odrakir:
Not sure if this is what William is saying, but we should probably also remove all iOS 10 specific code. For example, the swiping in `TimEntriesLogViewSource` where we do: `if (UIDevice.CurrentDevice.CheckSystemVersion(11, 0))` and all the related methods.

Odrakir:
Although that class specifically is modified in my PR, so maybe wait for that to be merged.

juanlaube:
It's not worth doing this right now, 3% of active users is a lot when increasing weekly active users is our goal. We should revisit after iOS 13 is out.

PR #5559, opened 2019-07-09
Smooth reports loading

dshr:
## What's this?
A way to make reports load the way they designed — smoothly

### Relationships
Closes nothing

## Why do we want this?
It's smooooooove

## How is it done?
1. Show the loading overlay only on first load
2. Add an activity indicator to show when data is loading for subsequent loads.

### Side effects
None

## Review hints
Test pls

## :squid: Permissions
All yours!


arturs105:
Everything seems to be working properly apart from one thing - the activity indicator looks a bit off during the initial load, while the selected period label isn't visible. See screenshot below.

I think this could be fixed in 2 ways:
* Center the activity indicator, when the label is hidden
* Show the "This Week" label immediately so the activity indicator doesn't look off center

Screenshot 2019-07-15 at 11 49 37

arturs105:
Also: what's the relationship with this: #4879.

amulware:
Please present comparative screen recordings or test builds internally for discussion and review by @maritoggl and me. On hold until we determine how far to go with this.

dshr:
@amulware @maritoggl here's a before and after:
_Before_
smooth-reports-before
_After_
smooth-reports-after

dshr:
I've added a default state to the report period observable, so we're presented with "This week" from the get go, no need to wait. All subsequent loads are fine.

PR #5560, opened 2019-07-09
🚧 Filter random forest TEs for description or project

sofokli1:
## What's this?
Filters out all TEs without description or project. This was no check in the situation where the number of TEs without project was not high enough, we would use all available TEs.

### Relationships
none

## Why do we want this?
We do not want suggestions without description or project

## How is it done?
by checking for description or project

## :squid: Permissions
Me


Toggl.Core/Suggestions/RandomForestSuggestionProvider.cs
```diff
@@ -58,6 +58,7 @@ private IEnumerable<IDatabaseTimeEntry> predictUsingRandomForestClassifier(IEnum
                 return predictUsing2Steps(timeEntries);
 
             timeEntries = timeEntriesForPrediction
+                .Where(te => !string.IsNullOrWhiteSpace(te.Description) || (te.ProjectId.HasValue && !string.IsNullOrWhiteSpace(te.Project.Name)))
```
kornelijepetak:
```suggestion
                .Where(te => !string.IsNullOrWhiteSpace(te.Description) || !string.IsNullOrWhiteSpace(te.Project?.Name))
```

kornelijepetak: Just a simplification.

sofokli1: What is the difference?

sofokli1: I was trying to be consistent with the check few lines above. should i change that as well?

kornelijepetak: Don't waste time on it then. I've added it to my list of very small 💄  changes, so once in the future, I'll update it to be concise. You can go with the merge now.

PR #5563, opened 2019-07-09
Track iOS widget (un)installs

### LINKED ISSUE #5090: Track widget install/uninstall
Detect when the widget is installed/uninstalled and track the necessary events.

Add a new event, `TimerWidgetInstallStateChange(bool installed)`.

Installs and uninstall events are not provided by the platform, but we can use two values in user defaults to accomplish this, LastUpdatedDate and InstallTracked. When the widget is updated, we set the current date in that value.
If the InstallTracked value is false, we track an install event.
If the value is older than X days (we need to decide how many days are X), we track an uninstall event and remove both values from user defaults.

While not 100% accurate, it's still better than nothing.

Depends on #5084

juanlaube:
## What's this?
This detects when the widget is installed and uninstalled. Because iOS doesn't tell us when that happens, I had to persist some state in user defaults for that.

### Relationships

Closes #5090

## Why do we want this?
🎛 📈 

## How is it done?
When the widget appears, I set the time to user defaults.
If the value is new in user defaults, the widget was just installed.
If a significant time passed since the view appeared (I set this to 3 days, we can adjust later but I don't see this being a problem other than receiving the events a bit after), the widget was uninstalled.

This doesn't cover the case when the user uninstall the app before removing the widget, but there's nothing we can do in that case, and it's not possible on Android either.

### Why not another way?
I don't see another way.

### Side effects
None.

## :squid: Permissions
🦑


Toggl.iOS/Startup/AppDelegate.Background.cs
```diff
@@ -69,5 +72,26 @@ private void observeAndStoreProperties()
...
+        private void detectTimerWidgetInstallStateChanged()
+        {
+            var analyticsService = IosDependencyContainer.Instance.AnalyticsService;
+            var isCurrentlyInstalled = SharedStorage.Instance.GetWidgetInstalled();
+            var lastUpdated = SharedStorage.Instance.GetWidgetUpdatedDate();
+
+            if (!lastUpdated.HasValue)
+                return;
+
+            if (isCurrentlyInstalled && DateTimeOffset.Now - lastUpdated.Value >= widgetInstallChangedInterval)
```
sofokli1: `DateTimeOffset.Now` I guess there is no `TimeService`

Toggl.iOS/Startup/AppDelegate.Background.cs
```diff
@@ -69,5 +72,26 @@ private void observeAndStoreProperties()
...
+            {
+                SharedStorage.Instance.SetWidgetInstalled(false);
```
sofokli1: what is the point of this?

juanlaube: This is to avoid sending repeated install events.

sofokli1: why set it to false?

juanlaube: This is to keep track of the widget install status, we don't want to send events if this status didn't change.

PR #5566, opened 2019-07-10
Fix push notification integration tests

Odrakir:
## What's this?
This fixes the integration tests that were failing.

### Relationships
Doesn't close any issue.

## How is it done?
We were creating users with the email address `@mock.toggl.com`, but apparently backend implemented a filter to only allow `@toggl.com` addresses. So I've changed that.

## :squid: Permissions
Go!


amulware:
Blocking this until further notice. There's a reason we went with those email domains: to make it 100% clear where these emails come from. There is a single endpoint that has issues with this for reasons that so far have not been explained satisfactorily to me, and I don't think we should make a big change like this without being really really sure we want to.

Odrakir:
Maybe we can ask backend to include those subdomains in the filter, then.

amulware:
If backend won't change their white-list to include subdomains, I'd prefer if we changed the emails only for the tests that are actually failing. And maybe even add another one that checks whether the white-list is in place (i.e. use a failing email and verify that we get a 403 result).

Odrakir:
This is superseded by #5572

PR #5568, opened 2019-07-10
Add Android widget design

### LINKED ISSUE #5093: Implement the time entry widget design
As the title says, implement just the time entry designs (suggestions will be implemented later). Consider the different layouts when resizing.

Designs: https://app.zeplin.io/project/5a0d3cd27a75add1a8924f6e?seid=5d13342f055d93045665f3b4

Depends on #5092

juanlaube:
## What's this?
Adds the widget design in Android. This considers the widget resizing already.

### Relationships

Closes #5093 

## Why do we want this?
🎨 

## How is it done?
Some layouts here and there.
Important considerations:
1. The UI elements we can use in RemoteViews are limited, so I had to make something similar to a fab button.
1. The small 1x1 size uses a different layout.

### Why not another way?
🤷‍♂ 

### Side effects
None.

## Review hints
When trying the 1x1 widget, it may not update immediately, but since this is only the layout, I think it's ok and we can handle the update after.

## :squid: Permissions
🦑


kornelijepetak:
I'm having issues with this 

image

This is all in the same line. When the widget is sized down, things don't go into a new row.

Also, "marquee" does not work in this setup. I can't see anything scrolling. Most probable that the width of particular textview is not fixed.

But otherwise, I'd actually propose a change in the design that puts the project/client/task to new row, as the running TE does. Otherwise, it's a lot of blank space. Even on smaller phones, I think the card is big enough for both rows to be visible. Gives more space for both description and project/client/task.

juanlaube:
With the project paused right now, I created #5607 to handle those comments and any leftover after we resume.

AlicjaSuska:
@kornelijepetak Please, see designs in this section: https://app.zeplin.io/project/5a0d3cd27a75add1a8924f6e/dashboard?seid=5d13342f055d93045665f3b4

 - we move some of the text to the next line.

PR #5579, opened 2019-07-12
♻ Fix sync on startup

### LINKED ISSUE #5578: App not syncing on app start
## 🐛 Bug Report
It appears that the app is not syncing automatically when it starts.

## Context
Build/Branch: 2.1, both platforms

## Expected behavior
The app should always try a full sync when:
- it is cold started
- it was in the background for 5 minutes or more

## To reproduce
- close app
- start/stop entry on the web
- open app
- observe the change from the web not syncing

## Suggestions
It could be that it is syncing but failing due to 429, but we need to make 100% sure.

Please make sure to double-check that the expected behaviour above is implemented and works correctly (**code and actual testing please**). The 5-minute limit should only be relevant if the app stays in the background, that state should not survive the app being shut down, and it should always sync on a cold start.

This issue applies to both platforms and has to be checked and tested on both.

kornelijepetak:
## What's this?
Returns full sync on startup.

> 🍒 I've set this to develop in such a way so that both iOS and Android can cherry-pick into their hotfix branches.

### Relationships
Closes #5578

## Why do we want this?
Cause sync! ♻️ 

## How is it done?
During some refactoring, which were good, a line of code was dropped, which was not good. I put it back there. The reason I did not want to put it into main view model is so that sync can already start while the main screen is loading, so that a bit of time is saved.

### Why not another way?
Inform me.

### Side effects
Should be none.

## Review hints
Please, functional test both platforms.

## :squid: Permissions
me and @amulware .


amulware:
If this is gonna go into a hotfix, this is the wrong base branch, isn't it?

juanlaube:
Looks good to me, I tested on both platforms too. Please change the base branch to the Android hot fix and we can cherry pick on the iOS branch after.

kornelijepetak:
> Looks good to me, I tested on both platforms too. Please change the base branch to the Android hot fix and we can cherry pick on the iOS branch after.

This is the source. iOS and Droid 🍒 from this.

PR #5584, opened 2019-07-12
iOS 2.1.2 - Blackberry hotfix

juanlaube:
## User facing changelog
```
- Improved time entry suggestions
- Fixed sync failure sometimes being wrongly reported even on successful syncs
- Fixed a crash when detecting country on signup
- Changed links to now open in the user’s default browser
- Removed 3 steps on-boarding
```

## Changes
[ios-2.1.1...release/ios-2.1.2](https://github.com/toggl/mobileapp/compare/ios-2.1.1...release/ios-2.1.2)

⚠ Important! Note that this release also includes push notifications, but they are disabled at the time of this release and will not reach user in `2.1`. However, when a future release comes around, the release manager should take care to add proper changes to the change log, as those commits will not be visible on github compare link.


sofokli1:
What is the point of this PR? in the app store there is the 2.1.1 version and this only has a veriosn bump compared to that

PR #5587, opened 2019-07-12
✅ Delete app before each test

dshr:
## What's this?
A temporary fix to delete the app before each run.

### Relationships
Closes nothing

## Why do we want this?
So tests pass.

## How is it done?
Some extra `simctl` commands.

### Why not another way?
All others fails

### Side effects
Test are slower.

## Review hints
Run tests pls

## :squid: Permissions
Go for it


Toggl.iOS.Tests.UI/Configuration.cs
```diff
@@ -28,7 +28,12 @@ public static iOSApp GetApp()
         private static void resetSimulator(string deviceId)
         {
             if (deviceId == null)
+            {
+                var preDeleteCmdLine = string.Format("simctl uninstall booted com.toggl.daneel.debug", deviceId);
```
arturs105: Does this append the device id to the end of the command?

PR #5588, opened 2019-07-12
Dynamic settings screen

### LINKED ISSUE #5710: Refactor Settings
Currently the settings screen is layed out in Interface Builder, this makes that screen really hard to update.

We should make them dynamic from a model in the VM, then each platform can customise it adding/removing/editing as needed for its specific needs.

Odrakir:
## What's this?
I'm trying to make the settings screen dynamic so it's easier to update.

### Relationships
Closes #5710 

## Why do we want this?
Currently this screen is layed out in interface builder and editing it to add or remove any setting is a pain because you have to fight with constraints and it's not really clear how to get the same layout.

## How is it done?
At the beginning I intended to configure a table/recycler view in the View Model and use it for both iOS and Android. 

When I was almost done I changed my mind and moved all the code to the VC and leave this changes only for iOS. My reasoning:
- Each platform includes some common settings and some different, customising the list should be done in the platform specific code, but inserting those platform specific settings is error prone because it has to be done based on position.
- As I was including the actions for each cell in the configuration done in the VM, I stopped exposing those UIActions from the VM, which brakes all current tests and makes editing those tests to work with the new code hard and results in fragile tests.
- This was mainly a problem with the iOS implementation of this screen, I really think there's no need to over complicate things to fix a non-existing problem in Android.

### Side effects
You can thank me later when you need to add a new setting to the app.

## Review hints
Commit history is a mess, but the final result is simple enough. So yeah, ignore commits, check the code.

## :squid: Permissions
Go!


Toggl.iOS/Cells/Settings/SettingCell.cs
```diff
@@ -0,0 +1,99 @@
...
+    public partial class SettingCell : BaseTableViewCell<ISettingRow>
+    {
+        public static readonly string Identifier = nameof(SettingCell);
+        public static readonly UINib Nib;
+
+        private readonly float defaultSeparatorInset = 16;
+
+        private CompositeDisposable disposeBag = new CompositeDisposable();
+
+        private bool _isLast = false;
```
juanlaube:
```suggestion
        private bool isLast = false;
```

Toggl.iOS/DebugHelpers/ErrorTriggeringViewControllerShortcuts.cs
```diff
@@ -10,14 +10,15 @@ public partial class SettingsViewController
         public override void ViewDidAppear(bool animated)
         {
             base.ViewDidAppear(animated);
-
+/*
```
juanlaube: 🔥 the commented block

juanlaube: Or even better, the whole function 😄

Odrakir: Oh, I missed that one... But we still need a way to do this, right?

PR #5590, opened 2019-07-12
Fix missing characters when inputting text during UI tests

dshr:
## What's this?
A temp fix for the stuttering issue on iOS UITests

### Relationships
Closes nothing

## Why do we want this?
So tests pass

## How is it done?
We enter strings char by char

### Why not another way?
This is the only way to enter text reliably

### Side effects
Tests are slower

## Review hints
Run this

## :squid: Permissions
All yours!


dshr:
Just noticed this branched off at the wrong point, I'll change it and force-push the correct version

Toggl.Tests.UI/LoginTests.cs
```diff
@@ -3,6 +3,7 @@
 using System.Threading.Tasks;
 using Toggl.Tests.UI.Helpers;
 using Xamarin.UITest;
+using static Toggl.Tests.UI.Extensions.AppQueryExtensions;
```
arturs105: 🔥

PR #5597, opened 2019-07-15
🚧Fix suggestion execution during background fetch

### LINKED ISSUE #5596: Suggestions are being recalculated in background sync
## 🐛 Bug Report
In appcenter logs i can clearly see that that are suggestion presented events fired during background sync, which does not make much sense but happens due to one of suggestions recalculation trigger being the sync state changes

## Context
Build/Branch: 2.1.1
Phone OS: ios but it could also be android
Phone model: any

## Expected behavior
No need to recalculate suggestion during background fetch

## To reproduce
VS4M go to `Run->Run with -> Custom Configuration` and select `Background Fetch` and Run action `Debug`. you will see `SuggestionsPresented` in the application output view

## Suggestions
none

sofokli1:
## What's this?
makes sure suggestion do not recalculate during background fetch

### Relationships
Closes #5596 

## Why do we want this?
no need to recalculate suggestions during background fetch since we recalculate again when app comes to foreground

## How is it done?
by adding a variable and a function to the background service

### Why not another way?
this made sense

### Side effects
Only love

## Review hints
follow steps in the issues to run the app in background fetch mode

## :squid: Permissions
:squid:


Toggl.Core/Services/BackgroundService.cs
```diff
@@ -18,6 +18,8 @@ public sealed class BackgroundService : IBackgroundService
 
         public IObservable<TimeSpan> AppResumedFromBackground { get; }
 
+        public bool AppIsInBacground { get; private set; }
```
juanlaube:
```suggestion
        public bool AppIsInBackground { get; private set; }
```

Toggl.iOS/Startup/AppDelegate.Background.cs
```diff
@@ -13,6 +13,7 @@ public partial class AppDelegate
 
         public override void PerformFetch(UIApplication application, Action<UIBackgroundFetchResult> completionHandler)
         {
+            IosDependencyContainer.Instance.BackgroundService.EnterBackgroundFetch();
```
juanlaube: Is this really needed? 🤔

sofokli1: when app starts in background fetch mode this is called and the below functions are not (active, foreground, background)

kornelijepetak:
I decided that Juan's approval should be enough. Feel free to proceed with merge.

PR #5598, opened 2019-07-15
🚧Make sure none of the providers failing blocks the suggestions

### LINKED ISSUE #5583: No suggestions visible after some time
## 🐛 Bug Report
After using the new suggestions for maybe a day, they are no longer visible until I restart the app.

## Context

Build/Branch: 2.1 AppStore
Phone OS: 12.3
Phone model: iPhone 6S

## Expected behavior
I have so much data and frequent calendar events, there should always be suggestions (and there have always been in the past).

## To reproduce
Not exactly sure, it just happened after I used the app for a while, possibly a whole day or so.

Nothing I do seems to get them back, except restarting the app completely.

sofokli1:
## What's this?
Makes sure the suggestion provider do no fail thus always having suggestions to choose

### Relationships
Closes #5583

## Why do we want this?
We like suggestions

## How is it done?
by adding `OnErrorResumeNext`

### Why not another way?
I am all ears

### Side effects
only suggestions not disappearing

## Review hints
not sure how to reproduce the issue so just try to check the code

## :squid: Permissions
:squid:


kornelijepetak:
The code looks correct. I'd just like a test for each provider, to prevent regressions.

PR #5605, opened 2019-07-17
Add Cake build configurations for iOS and Android UI tests

arturs105:
## What's this?
This PR changes `build.cake` so we can run UI tests on both platforms with Cake. The commands to run the UI tests are:
* `bash build.sh --target Tests.UI.iOS`
* `bash build.sh --target Tests.UI.Android`

## Why do we want this?
This is necessary to run the UI tests in Bitrise.

## How is it done?
By changing the name of the previous UI test build target and adding a new target for Android UI tests.

### Why not another way?
I though about making a single build target with UI tests for both platforms, but i decided that having 2 separate targets would be better because this allows us to run both in parallel and run only one of them if we wish to do so.

## Review hints
See if you can run UI tests on both platforms.

## :squid: Permissions
:squid:


arturs105:
First try on Bitrise can be found [here](https://app.bitrise.io/build/e769112d178f49f4#?tab=log)

PR #5606, opened 2019-07-17
🚧 Filter out TEs with archived project

### LINKED ISSUE #5591: Suggestions suggest archived projects
## 🐛 Bug Report
I'm seeing suggestions with archived projects.

## Context

Recent iOS, no reason to believe Android would be different.

## Expected behavior
Archived projects should never be seen in suggestions.

## To reproduce
Have archived projects that seem worth suggesting.

## Suggestions
This is probably the ML provider I'm seeing, but it might be others as well, please check them all.

sofokli1:
## What's this?
In random forest provider we did not filter out archived projects when fetching data from the DB

### Relationships
Closes #5591

## Why do we want this?
We do not want suggestions from archived projects

## How is it done?
We filter the data we feed the random forest algo

## :squid: Permissions
:squid:


Toggl.Core/Suggestions/RandomForestSuggestionProvider.cs
```diff
@@ -230,5 +229,10 @@ private int predictTimeEntryID(IEnumerable<IDatabaseTimeEntry> timeEntriesForPre
 
             return uniqueOutputs[prediction];
         }
+
+        private bool isTimeEntryActive(IDatabaseTimeEntry timeEntry)
+            => timeEntry.IsDeleted == false
+               && !timeEntry.IsInaccessible
```
amulware: Why the inconsistency? (comparing to false vs ! operator)

Toggl.Core.Tests/Interactors/Suggestions/RandomForestSuggestionProviderTests.cs
```diff
@@ -145,6 +146,30 @@ public async Task ReturnsEmptyObservableIfThereAreNoTimeEntries()
                 }
             }
 
+            [Fact]
+            public async Task DoesNotReturnSuggestionsWithArchivedProjects()
+            {
+                var numberOfTimeEntrieWithActiveProject = 120;
+                var numberOfTimeEntrieWithoutActiveProject = 120;
```
amulware: Maybe make it have a lot more inactive than active ones to make sure the test doesn't pass randomly? And maybe, in addition, have a test with only inactive projects to check that nothing is returned.

Toggl.Core/Suggestions/RandomForestSuggestionProvider.cs
sofokli1: it was copy pasted 🙈

amulware: Could call on these two conditions btw, we should think about making it harder to get 'bad' time entries like this in the future (except where we really need them). :)

sofokli1:
@kornelijepetak I am merging this due to time constrain, but feel free to review and I can make an other PR if needed

PR #5608, opened 2019-07-17
Stop restarting the iOS simulator on every test run

dshr:
## What's this?
A small change with a big impact — we remove the simulator restart before each test, since all we need is an app reinstall.

### Relationships
Closes nothing

## Why do we want this?
To save time — a full test run takes about 20min less now.

## How is it done?
:fire:

### Why not another way?
This is simpler than the alternative.

### Side effects
None that I've noticed.

## Review hints
Test this.

## :squid: Permissions
All yours


Toggl.iOS.Tests.UI/Configuration.cs
```diff
@@ -26,24 +22,16 @@ public static iOSApp GetApp()
             return app;
         }
 
-        private static void resetSimulator(string deviceId)
+        private static void reinstallApp(string deviceId)
         {
             if (deviceId == null)
             {
-                var preDeleteCmdLine = string.Format("simctl uninstall booted com.toggl.daneel.debug", deviceId);
+                var preDeleteCmdLine = string.Format("simctl uninstall booted com.toggl.daneel.debug");
```
kornelijepetak: You don't seem to be using `Format` facilities, use normal string instead. It's not parameterized.

```suggestion
                var preDeleteCmdLine = "simctl uninstall booted com.toggl.daneel.debug";
```

Toggl.iOS.Tests.UI/Configuration.cs
```diff
@@ -26,24 +22,16 @@ public static iOSApp GetApp()
             return app;
         }
 
-        private static void resetSimulator(string deviceId)
+        private static void reinstallApp(string deviceId)
         {
             if (deviceId == null)
             {
-                var preDeleteCmdLine = string.Format("simctl uninstall booted com.toggl.daneel.debug", deviceId);
+                var preDeleteCmdLine = "simctl uninstall booted com.toggl.daneel.debug";
```
arturs105: I don't think `preDeleteCmdLine` is a good name. If i understand this correctly, this command uninstalls the app from the simulator that is currently booted right? If so, then `uninstallAppFromBootedSimulatorCmdLine` or something along the lines would work better. The same goes for `preDeleteProcess` as well.

PR #5610, opened 2019-07-18
Track if app is launched with accessibility enabled

### LINKED ISSUE #5604: Track how many users open the app with screen reader enabled
An event like `AppLaunchedWithVoiceOverEnabled` should be enough.

arturs105:
## What's this?
This adds an event that tracks if VoiceOver/TalkBack is active, when the app is launched.

### Relationships

Closes #5604 

## How is it done
I got the Android code from stack overflow and can't really test it to see if it's working, so anyone, who has an android phone must review this.

## Why do we want this?
To know how much effort to put into accessibility.

## Review hints
❗️**Important** ❗️

This must be tested on as many Android devices as possible. With screen reader enabled and disabled. I can do this on iOS myself, but i don't have an Android phone 🤷‍♂ .

❗️ **Important** ❗️

## :squid: Permissions
Only after at least 1 approving review that mentions successful functional testing on an Android phone.


Toggl.Core/Analytics/BaseAnalyticsService.cs
```diff
@@ -276,6 +276,9 @@ protected BaseAnalyticsService()
         [AnalyticsEvent("Location")]
         public IAnalyticsEvent<ApplicationInstallLocation> ApplicationInstallLocation { get; protected set; }
 
+        [AnalyticsEvent]
+        public IAnalyticsEvent AppStartedWithScreenReaderEnabled { get; protected set; }
```
kornelijepetak: Hey, I think it'd be wiser for data analysis if this event had a bool parameter stating whether the screen reader is turned on or off. This way, among the events coming in, we'll see the exact ratio of people who have it turned on to those of who don't.

Without this, we can't compare the number of YESes to anything in particular, since the event was not there since the app launch and also since this is done on every app start.

PR #5611, opened 2019-07-18
🚧 Fix possible null indexPath

### LINKED ISSUE #5432: CalendarCollectionViewSource: Object reference not set to an instance of an object
## 💥 Crash Report
Apparently there's something wrong with the calendar layout.
```
CalendarCollectionViewSource.LayoutAttributesForItemAtIndexPath (Foundation.NSIndexPath indexPath)
SIGABRT: Object reference not set to an instance of an object
```

### 👥 Impact
This is affecting 93 users (iOS 12.3.1 for most of them) of 1.22 and it has happened 200 times so far.

### 🛰 Link
https://appcenter.ms/orgs/Toggl/apps/Toggl-iOS/crashes/errors/3567257895u/overview

`CalendarCollectionViewSource.LayoutAttributesForItemAtIndexPath (Foundation.NSIndexPath indexPath)
SIGABRT: Object reference not set to an instance of an object`
```cs
CalendarCollectionViewSource.LayoutAttributesForItemAtIndexPath (Foundation.NSIndexPath indexPath)
CalendarCollectionViewLayout.frameForEditingHour (Foundation.NSIndexPath indexPath)
CalendarCollectionViewLayout.LayoutAttributesForSupplementaryView (Foundation.NSString kind, Foundation.NSIndexPath indexPath)
```

sofokli1:
## What's this?
Fix a crash in the calendar collection view. This is not a complete fix due to the issue being hard to reproduce. My theory is that the `StopEditing` is being used from 2 different gesture recognisers. This might be creating a race condition where 1 of the the 2 gesture recognisers cancels and calls `StopEditing` which in turn nullifies `editingItemIndexPath` while the general logic of the calendar collection view expects it to have a value. `editingItemIndexPath` was being used to get data from an array and creating the crash. If we get this live we can check if we did it from crash reports.

### Relationships
Closes #5432

## Why do we want this?
crash no more

## How is it done?
removed the line that was nulling `editingItemIndexPath`

### Why not another way?
A better way could be implemented if the error can be reproduced consistently

### Side effects
can't find any

## Review hints
test the app because there is not much code change to look at

## :squid: Permissions
:squid:


sofokli1:
This should also take care of a few more calendar collection view crashes from app center

Issue #5569, opened 2019-07-11
Creating tags crashes


amulware:
## 🐛 Bug Report
I don't have any more info than the title, except that several users have reported this through reviews.

## Context

Android. (At least I haven't seen reports on iOS)

## Expected behavior
🚫💥

## To reproduce
Try creating some tags in the start or edit view, see what happens.

## Suggestions
💣🔫

nicktoggl:
**First case** 
App name and version: Giskard 2.1, 
Phone model: 9ONEPLUS A6013 
Platform and OS version:Android 9
conversations/22787786814

"Steps to reproduce: click the tags field, type 1 character, app crashes"

**2nd Case**
App name and version: Giskard/2.0.1
Phone model: SM-G973F
Platform and OS version: Android 9
conversation/22803926749

"the app is crashing a lot. when trying to create an extra tag to track an assignment I'm doing and every time I type the first character for the tag, the app restarts."

lise-toggl:
[case](https://app.intercom.io/a/apps/ayixs927/inbox/inbox/2833855/conversations/22819537259)
Giskard 2.1.1
Android 9
POCOPHONE F1

>when using &#34;Tag&#34; app crashes right away. I have done update also and before last update it was not working also

amulware:
Since we have 2.1 reports, consider checking specific user's exceptions on app center.

juanlaube:
Some crash made it to AppCenter: https://appcenter.ms/orgs/Toggl/apps/Toggl-Android/crashes/errors/1997674262u/overview

PR #5618, opened 2019-07-19
Use id in tags vm to have stable identifiers

juanlaube:
## What's this?

This solves a crash that happens after creating or trying to select tags on a time entry but doesn't close the issue entirely.

It's uncommon, but multiple tags with the same name can be created by repeatedly tapping on the create button, this generates a problem for the tags adapter because two different tags have the same identifiers and the duplicated tags are unsyncable.

This PR doesn't solve the underlying problem though, which is allowing to create multiple tags with the same name.

### Relationships

Refs #5569

## Why do we want this?
We want to app to be like the identifiers, stable.

## How is it done?
Considering the tag id in the `Equals` and `GetHashCode` methods.

### Why not another way?
There are different options to solve the underlying problem, which is allowing to create multiple tags by tapping repeatedly.
One involves blocking all RxAction to wait for the current execution to complete, the other is to lock the `selectTag` method in `SelectTagsViewModel`, I tried some ideas and couldn't make this work, I need help from an expert with Rx.

### Side effects
Instead of crashing, users will see duplicated tags, time entries referencing those duplicated tags won't sync.

## :squid: Permissions
No. Let's see if there's a better solution first.


juanlaube:
This is not a good solution because even if the app doesn't crash, some tags will be duplicated and unsinkable. A better solution proposed by @kornelijepetak is to don't create the tag if it already exists in that workspace.

PR #5627, opened 2019-07-23
Add Edit TE View assertion method for the UI tests

### LINKED ISSUE #5549: Make edit time entry view assertion method
This should be an extension method something like the following examples:
* AssertEditViewVisible(string description, string projectName, string clientName, ... , bool billable)
* AssertEditViewVisible(SomeCustomClassOrStructWithAllTheProperties timeEntry)

The extension method should check:
1. If edit view is visible
2. If it is showing all the correct data

Beware though, there may be problems with asserting loooong project names/descriptions/tags etc., so that needs to be tested out.

Also: asserting the project/task/client combo may be difficult, since (i think) we are using an attributed string for those 3. This method may help with that: `AppQueryExtensions.Contains(this AppQuery q, string text)`.

dshr:
## What's this?
An extra assertion method to check that the edit time entry view is opened and in the correct state.

### Relationships
Closes #5549 

## Why do we want this?
More concise tests.

## How is it done?
By changing some accessibility properties and a whole lot of querying.

### Why not another way?
It's what it says on the tin.

## Review hints
Go commit-by-commit

## :squid: Permissions
@arturs105 only


Toggl.iOS/ViewControllers/EditTimeEntryViewController.cs
```diff
@@ -237,6 +237,21 @@ public override void ViewDidLayoutSubviews()
                 .Subscribe(TagsTextView.Rx().AttributedTextObserver())
                 .DisposedBy(DisposeBag);
 
+            ViewModel.Tags
+                .Select(tags =>
+                {
+                    if (tags.Any())
+                    {
+                        return $"Tags: {string.Join(", ", tags)}";
```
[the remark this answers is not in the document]
dshr: I'm not sure I'm following — here we're actually prefixing the new accessibility label with `Tags: `? Or is there a way to set the prefix in one place, and only add the tags list here?

Toggl.Tests.UI/Extensions/EditTimeEntryExtensions.cs
```diff
@@ -0,0 +1,70 @@
...
+    public static partial class EditTimeEntryExtensions
+    {
+        public static void AssertEditViewIsVisible(
+            this IApp app,
+            string description = null,
+            string projectName = null,
+            string tags = null,
+            string startTime = null,
+            string stopTime = null,
+            bool? isStopped = null,
```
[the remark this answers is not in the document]
dshr: C# `string`s can be null, while `bool`s can not, they default to `false`. We're using nullable behaviour here, so defaulting to false isn't a good option.

Toggl.iOS/ViewControllers/EditTimeEntryViewController.cs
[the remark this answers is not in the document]
arturs105: 🤦‍♂ No, this is actually correct - i didn't notice the "Tags: " prefix 😄

Toggl.Tests.UI/Extensions/EditTimeEntryExtensions.cs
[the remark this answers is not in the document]
arturs105: Okay, okay, i think i get it.

Toggl.Droid.Tests.UI/Extensions/EditTimeEntryExtensions.cs
```diff
@@ -17,14 +17,14 @@ public static void ConfirmEditTimeEntry(this IApp app)
...
-        public static bool CheckIfTagsAreWrong(this IApp app, string tags)
+        public static bool TagsAreCorrect(this IApp app, string[] tags)
         {
             var foundTags = app.Query(x => x.Marked("TagsRecyclerView").Child().Child()).Select(x => x.Text);
-            return tags != string.Join(", ", foundTags);
+            return tags.SequenceEqual(foundTags);
         }
 
-        public static bool CheckIfIsBillableIsWrong(this IApp app, bool isBillable)
+        public static bool IsBillableIsCorrect(this IApp app, bool isBillable)
```
arturs105: There's one too many `Is` here. I recommend removing the first one for consistency with the other method in this file.

Toggl.Tests.UI/Extensions/EditTimeEntryExtensions.cs
```diff
@@ -0,0 +1,79 @@
...
+                wrongFields.Add("duration");
+
+            if (startDate != null && !app.StartDateIsCorrect(startDate))
+                wrongFields.Add("start date");
+
+            if (isBillable != null && !app.IsBillableIsCorrect((bool)isBillable))
+                wrongFields.Add("billable state");
+
+            if (wrongFields.Any())
+                throw new EditViewAssertionException($"Edit view is open, but {string.Join(", ", wrongFields)} {(wrongFields.Count > 1 ? "were" : "is")} wrong.");
```
arturs105: In the code `wrongFields.Count > 1 ? "were" : "is"` one of the verbs is in the past and the other one is in present, this should either be `wrongFields.Count > 1 ? "are" : "is"` or `wrongFields.Count > 1 ? "were" : "was"`. I'd go with the version, where both verbs are in present.

dshr: 👍

Toggl.Droid.Tests.UI/Extensions/EditTimeEntryExtensions.cs
dshr: 👍

Toggl.Droid.Tests.UI/Extensions/EditTimeEntryExtensions.cs
```diff
@@ -15,5 +16,15 @@ public static void ConfirmEditTimeEntry(this IApp app)
 
             app.WaitForNoElement(EditTimeEntry.Confirm);
         }
+
+        public static bool TagsAreCorrect(this IApp app, string[] tags)
+        {
+            var foundTags = app.Query(x => x.Marked("TagsRecyclerView").Child().Child()).Select(x => x.Text);
```
arturs105: `"TagsRecyclerView"` should be extracted to `AndroidAccessibilityNames.cs`.

PR #5628, opened 2019-07-24
Update handler scheduler according to the original

kornelijepetak:
## What's this?
Update HandlerScheduler according to the original source.

## Why do we want this?
If they considered it as a bug, we should as well.

## How is it done?
By following suit.

📝 **EDIT.** This PR also changes:
* all instances of `SynchronizationContext.Current` replaced by DI-based scheduler.
* `GetAwaiter().GetResult()` changed into `SubscribeOn(main)` to avoid deadlocks when crossing sync/async boundary.

### Side effects
All actions (regardless of the thread) are now posted to the handler thread.

## :squid: Permissions
Guys who were present on a CALI meeting.

PR #5640, opened 2019-07-25
Always use the scheduler provider

juanlaube:
## What's this?
This replaces all usages of `SynchronizationContext.Current` with `DependencyInjection.Instance.SchedulerProvider.MainScheduler` when subscribing on a different thread.

## Why do we want this?
`SynchronizationContext.Current` could be null.

## How is it done?
☝️ See `What's this?`

### Why not another way?
🤷‍♂ 

### Side effects
None?

## Review hints
Try the app in both platforms, specially around the parts that changed.

## :squid: Permissions
🦑


juanlaube:
The same code was included in #5628, I'm closing this PR.

PR #5642, opened 2019-07-25
Replace all uses of SynchronizationContext.Current

juanlaube:
## What's this?
As the title says, replaces all uses of sync context for the main scheduler, we still had some left after #5628.

## Why do we want this?
`SynchronizationContext.Current` is null outside the main thread, using the scheduler is safe though.

## How is it done?
Using the main scheduler.

### Side effects
None.

## :squid: Permissions
Only @kornelijepetak, vested in the powers of the release manager.


kornelijepetak:
I'm so confused. I remember searching the solution for `.Current` and I went thru all of the instances to check whether it's one of these. How did I miss 6 occurences. 🤔

PR #5646, opened 2019-07-26
🚧 Track paired apple watches

### LINKED ISSUE #5484: Track how many users have an Apple Watch
We can track how many users have an Apple Watch paired with their iPhones using [this](https://developer.apple.com/documentation/watchconnectivity/wcsession/1615665-ispaired). This will help us decide if it's worth to work on an Apple Watch app.

We need a new analytics event, `WatchPaired` with one single boolean parameter, `Installed`.

In my opinion we should track this only once per user to avoid duplicates, the event could be tracked on login or sign up so we don't add new storage for a single event.

The event name is generic enough to be reused in the future in Android for WearOS devices.

sofokli1:
## What's this?
Tracks paired apple watches

### Relationships
Closes #5484

## Why do we want this?
We want to know how many people have apple watches so that we make an informed decision into starting work on an apple watch app

## How is it done?
the code is straightforward

### Why not another way?
This seemed like the simplest way

### Side effects
not that i know off

## Review hints
Personally I tested with phone with apple watch paired but I guess it could be done also with simulator with a virtual watch paired.

## :squid: Permissions
:squid:


Toggl.iOS/Startup/AppDelegate.cs
```diff
@@ -43,6 +43,9 @@ public override bool FinishedLaunching(UIApplication application, NSDictionary l
 
             UNUserNotificationCenter.Current.Delegate = this;
 
+            var watchservicw = new WatchService();
```
juanlaube:
```suggestion
            var watchService = new WatchService();
```

PR #5656, opened 2019-07-26
Make Login UI tests AAA

### LINKED ISSUE #5548: Refactor LoginTests to AAA
This will probably need some new extension methods, for example `app.AssertLoginError(string message)` or `app.AssertSuccessfulLogin()`, and some other extension methods from those tests will need to be updated a bit.

The key thing here is clear separation of AAA.

dshr:
## What's this?
Just a bunch of deletions and renames really, to make our tests follow AAA logic.

### Relationships
Closes #5548

## Why do we want this?
AAA is a virtue in itself

## How is it done?
🔥 and 💄

## :squid: Permissions
UI test folks


dshr:
Tried adding a test to check the show password functionality, but UI tests don't see the buttons, which is a shame.

PR #5659, opened 2019-07-26
Enable strict mode logging for debug builds

### LINKED ISSUE #5273: Enable Android strict mode for debug builds
While checking the fragment's lifecycle for our custom made framework.
I've enabled Android's Strict mode and stepped upon some problems.

It might be beneficial to enable some configurations and logging for Debug builds.
Maybe even combine it with AppCenter for Adhoc builds.

Strict mode can be misleading and should be used to detect stuff that is happening by mistake or accident. For example, Strict mode can detect IO access, but that's something essential for in our app.

A very simple, not harmful addition would be:
```
StrictMode.SetVmPolicy(
                new StrictMode.VmPolicy.Builder()
                    .DetectAll()
                    .PenaltyLog()
                    .Build());
StrictMode.SetThreadPolicy(
                new StrictMode.ThreadPolicy.Builder()
                    .DetectCustomSlowCalls()
                    .Build());
```

During development, we can add a `.PenaltyDeath()` to make the apps crash if something is detected.

For adhoc, we could use `.PenaltyListener(...)` and log warnings with AppCenter.

daividssilverio:
## What's this?
This adds some Android `StrictMode` loggers to our debug builds.

### Relationships
Closes #5273

## Why do we want this?
To help detect unwanted behaviour in our debug builds.
Added listeners in this PR will help detect leaks and slow calls.

## How is it done?
The strict mode is enabled inside `#if DEBUG ... #endif` directives.

### Why not another way?
The other way is not having it. We can not use as well and always add this snippet when needed only.

### Side effects
There's more listeners around, debug builds will be a bit slower (I couldn't perceive any difference).

## Review hints
None.

## :squid: Permissions
🦑 it (but not before testing it on a device/emulator)


Toggl.Droid/Startup/TogglApplication.cs
```diff
@@ -1,6 +1,7 @@
 using Android.App;
 using Android.Runtime;
 using System;
+using Android.OS;
```
heytherewill: Ideally, code that is inside #if clauses should be self-contained and not depend on other usings, so if possible fully qualify the `StrictMode` class to `Android.OS.StrictMode`

PR #5660, opened 2019-07-26
Close android calendar cursors

### LINKED ISSUE #5271: Close android cursors on CalendarServiceAndroid
While checking the fragment's lifecycle for our custom made framework.
I've enabled Android's Strict mode and stepped upon some problems.

One of them is that we are opening android cursors and not closing after we are done with them in the `CalendarServiceAndroid`.

To verify this, please enable droid's strict mode using:
```
StrictMode.SetVmPolicy(
                new StrictMode.VmPolicy.Builder()
                    .DetectLeakedClosableObjects()
                    .DetectLeakedRegistrationObjects()
                    .DetectLeakedSqlLiteObjects()
                    .Build());
```

before `OnCreate` in the `TogglApplication`.

We just have to close them.

daividssilverio:
## What's this?
This closes the cursors being opened to read calendar events entries for The Calendar ™️ .

### Relationships
Detected thanks to #5659
Closes ~5659~, I mean, Closes #5271 

## Why do we want this?
Those cursors should've been closed already. We let it slip.
We were leaving cursors open every time the calendar service was being used.
That includes: The Calendar™️, when The Suggestions™️ are refreshed (main log, every time the tab is selected, every time the app opens), when the app comes back from background.

## How is it done?
`Close` is called in the cursors after they are used.

### Why not another way?
🤷‍♂ 

### Side effects
There shouldn't be any.

## Review hints
Run the app to see that it still works properly.

## :squid: Permissions
🦑 it after testing


daividssilverio:
@heytherewill @kornelijepetak I've tried using `using`s, but then the leaks stay in place.
Does one of you want to try to make it work?

Add this to your `TogglApplication.cs`, replace the `try-finally`s with `using`s and you should still see violations related to the Calendar.
```
            StrictMode.SetVmPolicy(
                new StrictMode.VmPolicy.Builder()
                    .DetectLeakedClosableObjects()
                    .DetectLeakedRegistrationObjects()
                    .DetectLeakedSqlLiteObjects()
                    .PenaltyLog()
                    .Build());
            StrictMode.SetThreadPolicy(
                new StrictMode.ThreadPolicy.Builder()
                    .DetectCustomSlowCalls()
                    .PenaltyLog()
                    .Build());
```

daividssilverio:
Oops, thanks for the correction, this doesn't close #5659.

PR #5664, opened 2019-07-29
Add fix translation request templates

### LINKED ISSUE #4837: Create issue template to request translations fixes/improvements
Following more or the same format as whatever is defined by #4835, but having in mind that the issue will be used for users and ourselves to submit requests that will fix bad/wrong translations.

### LINKED ISSUE #4838: Create a manual PR template (to copy and paste) for translations fixes/improvements
This won't be much different from whatever is defined on #4836, except that it will touch existing translations.

The PR should explain why the new translation is better/right and why what is already there was worse/wrong.

daividssilverio:
## What's this?
This PR adds the issue and pull request templates for translation fix/improvement requests.

### Relationships
Closes #4837 
Closes #4838 

## Why do we want this?
To be organized when fixing/improving translations from our codebase.

## How is it done?
It just adds two new files.
One to be used when a new translation fix/improvement is to be requested either by us or clients themselves.
Another to be used as a manual (or not) pull request template when the translation fix/improvement is done and needs to be merged into our codebase.

### Why not another way?
Feel free to make suggestions, improve wording, anything that could improve the experience for translations fixes/improvements.

### Side effects
There shouldn't be any.

## Review hints
Check the .md files.

## :squid: Permissions
🦑 i18n team + This needs a look from @maritoggl and/or @amulware. No hurry in merging these.


daividssilverio:
> I like this, but please move the existing pull request template to the new folder.

I'll check if this is strictly necessary, I was under the impression that these wouldn't affect anything and the default would still be used.

daividssilverio:
> I like this, but please move the existing pull request template to the new folder.

@juanlaube, based on my tests, that's not necessary. Unless you have a reason for that, even if it's for keep it organized, what we have works.

Issue #5561, opened 2019-07-09
Left to right languages issues


sofokli1:
## 🐛 Bug Report
- [ ] Create tag and client view looks ok but in selection view shouldn't the options be on the left and the switches on the right?
![IMG_1966](https://user-images.githubusercontent.com/885122/60841618-9f422280-a1a8-11e9-8b0b-31c2bba0c49a.png)

- [x] Create project looks strange (text starts at the middle of the view)
![IMG_1960](https://user-images.githubusercontent.com/885122/60841632-a6693080-a1a8-11e9-9881-5b6526009875.png)

- [x] Description text field does not start on the correct side regardless if start view edit view or group edit view.
![IMG_1961](https://user-images.githubusercontent.com/885122/60841665-c0a30e80-a1a8-11e9-971e-96c095d06af3.png)

- [x] Worst of all the description in start view can't create projects or tags, it just adds the #something or @something to the description as plain text. (alos no new project option below the description)
![IMG_1964](https://user-images.githubusercontent.com/885122/60841698-d6183880-a1a8-11e9-9e6c-cc697f789eca.png)

- [x] Regardless of the project color the project in arabic alphabet have the same plain black color
![IMG_1963](https://user-images.githubusercontent.com/885122/60841710-ddd7dd00-a1a8-11e9-97cc-7c61c6945ebc.png)

## Context
Build/Branch: 2.01, 2.1
Phone OS: ios but could also have implications in android
Phone model: iphone x

## Expected behavior
Not sure

## To reproduce
Just go to the specified views above and try to type in arabic. make sure the keyboard is selected before the textfields are selected

## Suggestions
none

sofokli1:
un-assigning self due to other obligations

PR #5669, opened 2019-07-30
🚧Improve RTL support in edit view

sofokli1:
## What's this?
Adds better support for RTL languages in edit TE view since the underling use of NSAttributtedString in the TextView subclass was conflicting with the RTL capabilities of the view.

### Relationships
Ref #5561

## Why do we want this?
We want better support for RTL languages

## How is it done?
~I tokenise the text into words and use NLLanguageRecognizer to detect the language. If it is par of the 12 RTL languages it will use UIStringAttributes that specify RTL direction for the range of the word, otherwise it will return LTR direction UIStringAttributes. It supports also mix text of RTL and LTR~
I removed the attributed text completely since it was only used to show the placeholder text. Now it uses the default Text and supports and has default os behaviour for RTL. Also increased text size in the description to be the same as zeplin designs.

### Side effects
When using mix of RTL and LTR the cursor may not be in the correct location even though the text is being inputed in the correct location. Not sure why this happens.

## Review hints
type type type

## :squid: Permissions
:squid:


Toggl.iOS/Views/TextViewWithPlaceholder.cs
```diff
@@ -14,10 +17,14 @@ public class TextViewWithPlaceholder : UITextView, IUITextViewDelegate
         private readonly int defaultPlaceholderSize = 14;
         private readonly UIColor defaultPlaceholderColor = Colors.Common.PlaceholderText.ToNativeColor();
         private readonly ISubject<String> textSubject = new Subject<string>();
+        private readonly String[] rightToLeftLanguages = {"ar", "arc", "dv", "fa", "ha", "he", "khw", "ks", "ku", "ps", "ur", "yi"};
+
 
         private bool isFocused;
         private UIStringAttributes placeholderAttributes;
-        protected UIStringAttributes DefaultTextAttributes { get; private set; }
+        private readonly NLTokenizer tokenizer = new NLTokenizer(NLTokenUnit.Word);
```
juanlaube: ⚠️ Careful! this `NaturalLanguage.framework` is available since iOS 12, this will crash badly in iOS 10 and 11.

sofokli1: we still support those? :trollface:

amulware: for now, yes - feel free to gather stats on that and we can talk about it, especially after 13 is out

Odrakir: We dropped iOS 10, didn't we?

Odrakir: Oh, we did not: https://github.com/toggl/mobileapp/pull/5543

sofokli1:
@amulware and @Odrakir anything to add to this PR based on the latest changes?

amulware:
I'm not even here. 👀

PR #5670, opened 2019-07-30
Remove Firebase.Performance from our apps

### LINKED ISSUE #3542: Replace the manually built Xamarin.Firebase.Perf.dll
We need to replace the dll added by #3541 once Xamarin releases a new version of the Firebase.Perf package.

### LINKED ISSUE #5637: Get rid of Firebase Performance
## 🚧 Refactor Proposal
Xamarin has just admited that [they will not support this library properly](https://github.com/xamarin/GooglePlayServicesComponents/issues/151#issuecomment-514289605). It took them almost a year, but at least we can now part our ways with it.

Also, as @amulware found out in Buenos Aires, there are some issues with the code that measures a couple of things, so it'll be a win-win.

## Dependencies
Supersedes #3542 if we decide to go through with this

## Technical aspects
Literally just get :fire: the whole `Toggl.Core.Diagnostics` folder and then make the app compile again :sogood:

heytherewill:
## What's this?
This removes Firebase.Performance from our apps

### Relationships
Closes #5637
Closes #3542

## Why do we want this?
Reasons are outlined in the issue. After discussing this with Paul during the architecture meeting, he agreed that we are barely using that data anyways and that it's better to get rid of the tracking for a package that is not even maintained properly

## How is it done?
🔥 

## :squid: Permissions
Only myself


daividssilverio:
Also, we could develop our own solution later on if we really want it.

amulware:
Develop our own performance solution? I'd like to see us wrangling that amount of data. :D

Issue #3672, opened 2018-11-07
 Improve UX for changing manual/timer modes


kornelijepetak:
### Current state of affairs

Users can currently switch between manual and timer tracking modes in the settings. This will change the default behavior of the ➡️button in the Main Log. However, they can execute the Start in the other mode by long pressing the ➡️button. 

The problem is... many users are not really aware of this.

### Suggestions

We could improve the description in the settings, to include something like: 
> This option sets the default behavior for the start button. Regardless of the current option, long pressing the button will always activate the alternative mode.

We need a better copy, but something along those lines.

The alternative would be to use the new onboarding step, but it may be an overkill at this point. And if you think the long description would clutter the settings too much, we could maybe remove the description (or shorten it) once the user has used the long press once or twice (or whatever). Or not.

amulware:
I'm very much against an onboarding step. Those should be for the most important features, and manual mode is not something we are promoting actively.

I agree with adding a line to the settings though, and also an animation to the start button itself (a circular bar filling up around the button while holding I'm thinking).

amulware:
Why haven't we done this yet? We really should.

I propose:
> You can add your data manually instead of tracking it in real time. You can also long press the start button to use the alternative mode.

Unless someone comes up with something better (ping @maritoggl) I say we go with this.

maritoggl:
This sounds reasonable enough.

heytherewill:
I think this is a good issue for test-weekers. @amulware could you just make it clearer what was decided and what should be implemented?

amulware:
My last comment above goes.

semanticer:
"good issue for test-weekers" caught my eye.  I think I get what should be done, only the part "a circular bar filling up around the button while holding I'm thinking" - do ve need some visual spec for this? I can imagine a solution but don't know if it is the right one :) @amulware

amulware:
I'd skip the animation for now and do just the thing in the settings.

kornelijepetak:
Usually we wait for designs, but they are hard to come by in recent times, so if it doesn't take too much of your time, you can add something on your own and we can peer review. 

Also, it's totally fine to solve one part of the issue (the copy in settings screen) with a PR that closes this issue on a contigency to create another issue that deals only with animation part of it until visuals are specified better.

amulware:
Something like this is what _I_ had in mind. If you feel like implementing this, or you think you got a better idea, don't let us stop you. But like Kornelije said, let's not spend too much time on it.

[...]

PR #5678, opened 2019-07-31
Change Manual mode explanation on android (#3672)

semanticer:
## What's this?

Change of the string for Manual Mode settings item as well some layout improvements in order to make it more scalable across different screen sizes and for longer texts like this new one.

### Relationships

The part of https://github.com/toggl/mobileapp/issues/3672 where we need to change the label on Android.


Toggl.Droid/Resources/layout/SettingsFragment.axml
```diff
@@ -315,52 +315,47 @@
                 android:layout_height="0.5dp"
                 android:layout_marginTop="24dp"
                 android:layout_width="match_parent" />
-            <FrameLayout
+            <LinearLayout
```
heytherewill: I'd rather see this as a `CoordinatorLayout` or `FrameLayout` instead of nested `LinearLayout`s

Toggl.Droid/Resources/layout/SettingsFragment.axml
```diff
@@ -315,52 +315,47 @@
...
-                    android:gravity="center_vertical"
-                    android:text="@string/ManualMode"
-                    android:textColor="@android:color/black"
-                    android:layout_marginTop="8dp"
-                    android:layout_marginLeft="16dp"
-                    android:layout_width="wrap_content"
-                    android:layout_height="wrap_content" />
+                <LinearLayout
+                    android:orientation="horizontal"
+                    android:layout_height="40dp"
```
daividssilverio: Instead of using a fixed height, we could use `wrap_content` in combination with `android:minHeight="40dp"` to avoid cutting the text on larger fonts.

Toggl.Droid/Resources/layout/SettingsFragment.axml
semanticer: Agreed, I just didn't want to go into `CoordinatorLayout` without layout preview working :) Also didn't think layout flatness is that important considering the rest of this layout. So should I refactor the whole layout to make it flat or just this part?

daividssilverio: I'm guessing this is a bit out of scope of the original issue.
We do think layout flatness is important, it's just that this screen has been like this for a while and it's not performance critical.
We just try not to make it worse when making changes to it.
I wouldn't be against an issue to revamp this layout though. That would be a nice `up-for-grabs` IMHO.

semanticer: Ok, I will solve this with the nested constraint layout tomorrow and create another task for screen layout refactor.

amulware: We actually were talking about generating the entire view from actual code instead of hardcoding the layout - I believe Android supports that quite easily? If so, maybe THAT should be the issue instead (though the solution will also have to accommodate iOS). :)

amulware:
Don't wait for my review on this one, I have neither the time nor the Android experience to judge. :)

Toggl.Droid/Resources/layout/SettingsFragment.axml
semanticer: Yeah, this screen is quite consistent so I think it might be doable. I just don't know if doing constrains manually would produce a nice code, it could get pretty messy and we would lose some of the goodies that android resource system is giving us.

semanticer: What's your opinion on this @heytherewill  ?

kornelijepetak: @heytherewill did you mean ConstraintLayout instead?

kornelijepetak: @amulware why would we do it from code? I mean, I can't see any particular benefits and it will introduce weirdness cause all other things are in resource files (layouts) and then all of a sudden this one is not 🤷‍♂ 

I know there's a lot of hype about swift-ui, but until something similar comes to xamarin, I think we should go with whatever is the most consistent with the rest of our codebase.

But in any case, I'd like to hear the arguments you talked about (in BA I assume?)

amulware: This was a BA thing, yes. https://www.notion.so/What-ruffles-your-feathers-discussion-04ee84d032f44fac884a2d46dfbf6c44#87878742c56242c09d03331de0653695

According to this, @Odrakir is taking care of it.

amulware: But perhaps this is indeed a very big change with potential unintended consequences and should not be done lighthandedly.

kornelijepetak: I'm confused. Notion says:

- Would allow iOS to use `UITableView`
- Would allow Android to use `SettingsFragment`

Like, what?

1. Android already uses `SettingsFragment` whatever that means 😄 
2. I assume it was about RecyclerViews?! (as an equivalent of tableview on iOS)

In any case, I understand now. For droid, I believe it will just introduce more layouts because we now need a cell for each setting type (button, switch, choice)... But it's not like that's a big problem.

amulware: I've been told this would allow us to expose our settings in the actual settings app.

daividssilverio: Everybody means the `PreferenceFragment` or its jetpack variant.

kornelijepetak: Oh I see...

PR #5686, opened 2019-08-01
🤖 Target Android SDK 28

kornelijepetak:
## What's this?
Well, it's Aug 1st.
Read why do we want this.

## Why do we want this?

image

## How is it done?
⏫ 

## :squid: Permissions
🦑


kornelijepetak:
I just noticed that the August warning is for new apps. Updates could have waited November. But I don't think it hurts.

PR #5687, opened 2019-08-01
:bug: Make sure the root view controller is set synchronously

juanlaube:
## What's this?
Fixes a crash in `develop` when the window's root view controller is not the main tab bar view controller.

## Why do we want this?
No crashes.

## How is it done?
The presenter dispatches an action asynchronously to the main queue, only if the current queue is a background queue.

### Why not another way?
🤷‍♂ 

### Side effects
None.

## :squid: Permissions
🦑


heytherewill:
Why no just use InvokeOnMainThread here?

juanlaube:
`InvokeOnMainThread` is dangerous and I would avoid that, we had a nasty deadlock before because of that.

Odrakir:
I don't like it, but if you say this is the only way...

PR #5689, opened 2019-08-01
Show archive icon in row if project is archived

### LINKED ISSUE #2583: Timer page: Show archived icon instead of the dot when the project is archived
In addition to (archived) text, we want to show a little archived icon at the front of the project name instead of the dot. See #2366

screen shot 2018-07-02 at 14 16 10

Zeplin: https://zpl.io/a74eDz8

semanticer:
## What's this?

Add the archive icon to log entries

### Relationships
Closes #2583


Toggl.Core.UI/ViewModels/TimeEntriesLog/LogItemViewModel.cs
```diff
@@ -13,6 +13,7 @@ public sealed class LogItemViewModel : IDiffable<IMainLogKey>, IEquatable<LogIte
         public LogItemVisualizationIntent VisualizationIntent { get; }
 
         public bool IsBillable { get; }
+        public bool HasArchivedProject { get; }
```
juanlaube: I would use `IsActive` directly because that's how we represent the property in the model.

Toggl.Core.Tests/UI/Transformations/TimeEntriesFlatenningTests.cs
```diff
@@ -326,6 +326,7 @@ private static IEnumerable<LogItemViewModel> expanded(IThreadSafeTimeEntry[] gro
                 representedTimeEntriesIds: group.Select(timeEntry => timeEntry.Id).ToArray(),
                 visualizationIntent: visualizationIntent,
                 isBillable: sample.Billable,
+                hasArchivedProject: sample.Project?.Active == false,
```
juanlaube: You can use the null coalescing operator here, I think it reads better.

Toggl.Core.UI/Extensions/TimeEntryExtensions.cs
```diff
@@ -24,6 +24,7 @@ public static class TimeEntryExtensions
                 new[] { timeEntry.Id },
                 visualizationIntent,
                 timeEntry.Billable,
+                timeEntry.Project?.Active == false,
```
juanlaube: Same thing here.

Toggl.Core.UI/Transformations/TimeEntriesGroupsFlattening.cs
```diff
@@ -141,6 +141,7 @@ private IEnumerable<LogItemViewModel> flattenGroup(IThreadSafeTimeEntry[] group,
                 representedTimeEntriesIds: group.Select(timeEntry => timeEntry.Id).ToArray(),
                 visualizationIntent: visualizationIntent,
                 isBillable: sample.Billable,
+                hasArchivedProject: sample.Project?.Active == false,
```
juanlaube: And here.

daividssilverio:
Please add more details to your PR. I recommend at least:
`## What's this?` <-- fill this
`### Relationships` <-- fill this
`## Why do we want this?` <-- Sometimes not needed, but useful
`## How is it done?` <-- Useful for reviewers, specially when touching layout, unless it's too simple
`### Why not another way?` <-- Useful when doing something that could be done in another way and you decided not to go that way
`### Side effects` <-- Useful when you know this causes side-effects or think it could affect something
`## Review hints` <-- meh
`## Tests` <-- very specific, not usually filled
`## :squid: Permissions` <-- fill this; Usually 🦑 to indicate that it should be Squashed; We don't merge regular PRs

Toggl.Core.Tests/UI/Transformations/TimeEntriesFlatenningTests.cs
kornelijepetak: To help with this:
```suggestion
                hasArchivedProject: !sample.Project?.Active ?? false,
```

`??` has this kind of conventional reading as a `default in case of null`, especially in combination with a null-propagation operator. I believe this is the only reason why C# devs would find this more readable than what you provided initially. Otherwise @semanticer 's solution is completely correct.

semanticer: I was trying elvis operator `?:` from Kotlin but it wasn't there so I thought there is no such feature in C#

semanticer:
All addressed all the issues in the latest commit.

Toggl.Core.UI/Extensions/TimeEntryExtensions.cs
```diff
@@ -24,6 +24,7 @@ public static class TimeEntryExtensions
                 new[] { timeEntry.Id },
                 visualizationIntent,
                 timeEntry.Billable,
+                timeEntry.Project?.Active ?? false,
```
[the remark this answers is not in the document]
semanticer: Makes total sense 👍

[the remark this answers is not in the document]
semanticer: Solved with the latest commit.

Toggl.Droid/Toggl.Droid.csproj
```diff
@@ -50,7 +50,7 @@
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release.PlayStore|AnyCPU' ">
     <Optimize>false</Optimize>
     <OutputPath>bin\Release.PlayStore</OutputPath>
-    <DefineConstants>USE_ANALYTICS;USE_PRODUCTION_API</DefineConstants>
+    <DefineConstants>USE_ANALYTICS;USE_PRODUCTION_API;USE_APPCENTER</DefineConstants>
```
kornelijepetak: Why is this change in this PR?

I don't think this PR has anything to do with analytics... 🤔 

Also, from this file, only the line

`<AndroidResource Include="Resources\drawable\ic_archived.xml" />` 

should be commited. Everything else seems to be a different tooling settings (BOM and EOF `\n`).

semanticer: You are right, my bad.

kornelijepetak: no worries.

PR #5693, opened 2019-08-01
Use Tasks instead of Observables in our API

### LINKED ISSUE #5631: The API layer should be Task<T> and not Observable<T>
## 🚧 Refactor Proposal
Rx.NET doesn't have the concept of an Observable that returns a single value and then finishes (like RxSwift's `Single` Trait). This makes it hard to communicate whether an observable is meant to be awaited or if it returns multiple values.

Our API, in specific, literally never has the multiple values use case: We just wait for a value and then return it or error. This means the API layer can be massively simplified (both in terms of code but also in terms of semantics) if we just return `Task<T>` instead of `IObservable<T>`.

We won't lose the benefit of composing observables, since both APIs have methods to convert between each other.

## Dependencies
This can be worked right away. We just need to make sure no projects are touching the API in the meantime.

Also, it's extremely important that someone with sync knowledge works on this; the sync manager is the biggest consumer of our API.

## Technical aspects
The core change is simple: We just change all public facing API interfaces to return `Task<T>` and then make the app compile again. This is simple in principle, but will definitely be a boring task in practice.

heytherewill:
## What's this?
This changes all Observables in the public interfaces of our API into Tasks

### Relationships
Closes #5631

## Why do we want this?
In order to be more consistent in the Task 1 Observable Many principle. The fact that this change was so simple actually shows that this is what we should have done from the ground up (after all, the underlying implementation already used `Task<T>`)

## How is it done?
Changing return types, pretty much

### Side effects
This might've breaked everything; test with care

## Review hints
Just, like, use the entire app! This change affects the following things directly:

- [ ] Reports
- [ ] Login
- [ ] Signup 
- [ ] Siri integration
- [ ] Syncing

So make sure to test all corner cases of these components

## Tests
[Integration tests](https://app.bitrise.io/build/32c6098818158e8f#?tab=log) - do not merge this branch unless these tests are passing.
[Syncing tests](https://app.bitrise.io/build/b065604118bf4044#?tab=log) - do not merge this branch unless these tests are passing.

## :squid: Permissions
Only myself


Toggl.Networking.Tests/ApiClients/StatusApiTests.cs
```diff
@@ -36,12 +37,14 @@ public void DoesNotHideThrownExceptions()
...
-                    .Catch((WebException exception) =>
-                    {
-                        caughtException = true;
-                        return Observable.Return(Unit.Default);
-                    }).Wait();
+                try
+                {
+                    await statusApi.IsAvailable();
+                }
+                catch
```
daividssilverio: Shouldn't the catch be specific?

Toggl.Networking/ApiClients/BaseApi.cs
```diff
@@ -38,65 +38,72 @@ protected BaseApi(IApiClient apiClient, IJsonSerializer serializer, Credentials
...
+        protected Task<string> SendRequest(Endpoint endpoint, HttpHeader header, string body = "")
             => SendRequest(endpoint, new[] { header }, body);
 
-        protected IObservable<string> SendRequest(Endpoint endpoint, IEnumerable<HttpHeader> headers, string body = "")
+        protected async Task<string> SendRequest(Endpoint endpoint, IEnumerable<HttpHeader> headers, string body = "")
         {
             var request = new Request(body, endpoint.Url, headers, endpoint.Method);
-            return sendRequest(request, headers)
-                .Select(response => response.RawData);
+            var response = await sendRequest(request, headers).ConfigureAwait(false);
```
daividssilverio: Should this be run on background by default?

Toggl.Networking/ApiClients/ClientsApi.cs
```diff
@@ -17,18 +19,18 @@ public ClientsApi(Endpoints endPoints, IApiClient apiClient, IJsonSerializer ser
...
 
-        public IObservable<IClient> Create(IClient client)
+        public Task<IClient> Create(IClient client)
         {
             var endPoint = endPoints.Post(client.WorkspaceId);
             var clientCopy = client as Client ?? new Client(client);
-            var observable = SendRequest(endPoint, AuthHeader, clientCopy, SerializationReason.Post);
-            return observable;
+            return SendRequest(endPoint, AuthHeader, clientCopy, SerializationReason.Post)
+                .Upcast<IClient, Client>();
```
daividssilverio: It's so weird to need to do this 🤔

Toggl.Networking/ApiClients/UserApi.cs
```diff
@@ -67,14 +70,20 @@ string timezone
...
+            try
+            {
+                var user = await SendRequest<User>(endPoints.Post, new HttpHeader[0], json)
+                    .ConfigureAwait(false);
+                return user;
+            }
+            catch (BadRequestException ex)
+            when (ex.LocalizedApiErrorMessage == userAlreadyExistsApiErrorMessage)
+            {
+                throw new EmailIsAlreadyUsedException(ex);
```
daividssilverio: Shouldn't this be re-throwing the `BadRequestException` when `ex.LocalizedApiErrorMessage != userAlreadyExistsApiErrorMessage`?

Toggl.Networking/ApiClients/ClientsApi.cs
kornelijepetak: In c# classes can't be co/contravariant, only interfaces and delegates. And since `Task<T>` is a class, that's what it is.

Just as `List<Derived>` does not inherit from `List<Base>` and can't be used in its stead.

Toggl.Networking/ApiClients/BaseApi.cs
heytherewill: It's not a CPU bound operation, so it doesn't `Run` on another thread, but once the response comes back it continues in another thread, yes.

Toggl.Networking/ApiClients/ClientsApi.cs
heytherewill: What he said :)

Toggl.Networking/ApiClients/UserApi.cs
heytherewill: The exception is not caught when`LocalizedApiErrorMessage != userAlreadyExistsApiErrorMessage`, so there's no need to rethrow it

Toggl.Shared/Extensions/TaskExtensions.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System.Threading.Tasks;
+
+namespace Toggl.Shared.Extensions
+{
+    public static class TaskExtensions
+    {
+        public static Task<TBase> Upcast<TBase, TDerived>(this Task<TDerived> task)
+            where TDerived : TBase
+            => task.ContinueWith(t => (TBase)t.Result);
```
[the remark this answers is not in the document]
heytherewill: I wouldn't introduce another context just for that, to be honest. I'd rather not care about the thread

heytherewill:
@kornelijepetak There are other issues to further change the interactors and friends into using Tasks. This is, indeed, meant to be a lighter change.

heytherewill:
@daividssilverio With regards to the `ToObservable` calls, these will be further simplified when we move the rest of the things to the same Task model :)

Toggl.Networking/ApiClients/ClientsApi.cs
```diff
@@ -17,18 +19,18 @@ public ClientsApi(Endpoints endPoints, IApiClient apiClient, IJsonSerializer ser
             this.endPoints = endPoints.Clients;
         }
 
-        public IObservable<List<IClient>> GetAll()
+        public Task<List<IClient>> GetAll()
```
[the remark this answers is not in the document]
heytherewill: > You said that that code would be further simplified in the future, but I have my doubts that it will be actually simpler, mostly when we are dealing with the composition of multiple observables. Is there a simpler way to do that with tasks?

Yes. C# has language features for using tasks, so composing them is as simple as calling `await` or doing chaining using `linq` (the language feature, not the library).

>I don't know... I see where all this change comes from, but I think, in order to see if it actually makes sense, it would be better to take a specific type, change its return value to Task and do the necessary changes in call site, to see where that leads us. Instead of reaching wide, reach deep.

I could try that in one specific type if that helps you ease your mind. I will create a branch with that later today :)

> I still see benefits in using only Observables (and I might be alone on this one), it's just one API and they are composable

Both apis are very interchangeable when one needs to compose and, in general, we are making it really confusing by exposing observables that are meant to be awaited.

> If we need to differentiate single vs many values we could create another type for that 

We don't need to introduce a type for single result computations because [Task is that type](https://github.com/dotnet/reactive#a-brief-intro). I agree that we would greatly benefit from the Driver type tho.

heytherewill:
Putting this on hold so we can discuss it further in Tallinn

Toggl.Networking/ApiClients/ProjectsSummaryApi.cs
```diff
@@ -26,24 +26,21 @@ public ProjectsSummaryApi(Endpoints endPoints, IApiClient apiClient, IJsonSerial
             this.credentials = credentials;
         }
 
-        public IObservable<IProjectsSummary> GetByWorkspace(long workspaceId, DateTimeOffset startDate, DateTimeOffset? endDate)
+        public Task<IProjectsSummary> GetByWorkspace(long workspaceId, DateTimeOffset startDate, DateTimeOffset? endDate)
```
simonrozsival: Why not async? I don't like `ContinueWith` and it seems we could avoid it here.

Toggl.Networking/ApiClients/TimeEntriesApi.cs
```diff
@@ -23,43 +24,50 @@ internal sealed class TimeEntriesApi : BaseApi, ITimeEntriesApi
...
             var timeEntryCopy = timeEntry as TimeEntry ?? new TimeEntry(timeEntry);
             if (reason == SerializationReason.Post)
             {
                 timeEntryCopy.CreatedWith = userAgent.ToString();
             }
 
-            var observable = SendRequest(endPoint, AuthHeader, timeEntryCopy, reason);
-            return observable;
+            return SendRequest(endPoint, AuthHeader, timeEntryCopy, reason)
+                .Upcast<ITimeEntry, TimeEntry>();
```
simonrozsival: Could we maybe add two generic params to `SendRequest` and hide the upcast inside?

Toggl.Networking/ApiClients/TimeEntriesReportsApi.cs
```diff
@@ -40,27 +41,18 @@ public TimeEntriesReportsApi(Network.Endpoints endPoints, IApiClient apiClient,
             var parameters = new TimeEntriesTotalsParameters(userId, startDate, endDate);
             var json = serializer.Serialize(parameters, SerializationReason.Post);
 
-            return Observable.Create<ITimeEntriesTotals>(async observer =>
-            {
-                var response = await SendRequest<TotalsResponse>(endPoints.Totals(workspaceId), credentials.Header, json);
-
-                var totals = new TimeEntriesTotals
+            return SendRequest<TotalsResponse>(endPoints.Totals(workspaceId), credentials.Header, json)
+                .ContinueWith<ITimeEntriesTotals>(t => new TimeEntriesTotals
```
simonrozsival: 😢

Toggl.Networking/ApiClients/TimeEntriesApi.cs
heytherewill: Not sure it's worth the hassle

PR #5699, opened 2019-08-02
Make sure all the popup windows are deallocated

### LINKED ISSUE #4035: Fix leaky onboarding leaky windows
I was testing the background sync and realized we are leaking the popup windows when we go into background while a popup is visible. 

This should be as easy as dismissing the popups when the activity goes into background.
This has to be fixed on all the activities that show a window popup.

A leak warning is shown when you have a leak:
```
android.view.WindowLeaked: Activity md5a44fea7db10133e9f6bddee3bde9d640.MainActivity has lea
                            ked window android.widget.PopupWindow$PopupDecorView{8f03aae V.E...... ......I. 0,0-547,172}
                             that was originally added here
                         E      at android.view.ViewRootImpl.<init>(ViewRootImpl.java:368)
                         E      at android.view.WindowManagerGlobal.addView(WindowManagerGlobal.java:299)
                         E      at android.view.WindowManagerImpl.addView(WindowManagerImpl.java:85)
                         E      at android.widget.PopupWindow.invokePopup(PopupWindow.java:1258)
                         E      at android.widget.PopupWindow.showAsDropDown(PopupWindow.java:1110)
                         E      at android.widget.PopupWindow.showAsDropDown(PopupWindow.java:1069)
                         E      at mono.java.lang.RunnableImplementor.n_run(Native Method)
                         E      at mono.java.lang.RunnableImplementor.run(RunnableImplementor.java:30)
                         E      at android.os.Handler.handleCallback(Handler.java:739)
                         E      at android.os.Handler.dispatchMessage(Handler.java:95)
                         E      at android.os.Looper.loop(Looper.java:148)
                         E      at android.app.ActivityThread.main(ActivityThread.java:5417)
                         E      at java.lang.reflect.Method.invoke(Native Method)
                         E      at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:726)
                         E      at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:616)
```

Get rid of those and you don't have any more leaks 👍 

Why? We don't want our app to be holding references it doesn't need, we need to free resources when we are not using them anymore, let's be good neighbours to other applications.

semanticer:
## What's this?

Make sure all the popup windows are not only dismissed but also nullified when activity/fragment goes into the background.  Some of the PopupWindows were originally only dismissed but not nullified and one was forgotten completely.

### Relationships

Closes #4035

## Why do we want this?

We are not fans of memory leaks

### Why not another way?

This is probably the simplest solution given the current code, but it might be more reasonable to come up with some solution where we don't have to think about every single popup being deallocated. This might be solved if we make PopupWindows Lifecycle observers.


Toggl.Droid/Fragments/MainFragment.Onboardings.cs
```diff
@@ -52,6 +52,10 @@ public override void OnStop()
             base.OnStop();
             playButtonTooltipPopupWindow.Dismiss();
```
juanlaube: Probably not related to this leak, but I just had a crash here (null ref) when sending the app to background and the first popup. Consider doing some null checking when calling the `.Dismiss()` methods.

semanticer: I just made those call null-safe

PR #5701, opened 2019-08-02
Run UI updates on main thread

### LINKED ISSUE #4960: NSConstraintExtensions.ConstantAnimated 💥
SIGABRT: Objective-C exception thrown. Name: NSGenericException Reason: *** Collection <__NSArrayM: 0x2827d3de0> was mutated while being enumerated. 

### 👥 Impact
Highest crashes in `1.20.1`

### 🔗 Context
Something to do with `AnimationExtensions` 

### 🛰 Link
https://appcenter.ms/orgs/Toggl/apps/Toggl-iOS/crashes/errors/176021915u/overview

https://appcenter.ms/orgs/Toggl/apps/Toggl-iOS/crashes/errors/176021915u/overview

juanlaube:
## What's this?
Potentially solves the second most common crash in iOS. While I couldn't reproduce the crash, I noticed a sync error when the reports calendar is open and the beginning of week changes while syncing,.

### Relationships

Closes #4960 

## Why do we want this?
No crashes, no sync errors.

## How is it done?
`.ObserveOn(schedulerProvider.MainScheduler)`.

### Why not another way?
I was going for a driver first, but that means giving a default value in case of an error, I don't want that.

### Side effects
None.

## Review hints
To reproduce the sync error I've found:
1. Go to the reports tab
1. Open the calendar
1. Go back to the main log
1. Change the beginning of week setting in the webapp
1. Pull to refresh
1. 🐛 

## :squid: Permissions
🦑


juanlaube:
For some reason using drivers didn't work first, but after another try it seems to work, maybe I missed something the first time? 🤷‍♂

PR #5703, opened 2019-08-04
Cancel login process when back pressed

### LINKED ISSUE #1411: Cancel login/sync on BACK
Basically, if the user uses the hardware back button during the login/initial sync we need to cancel all that and return back to the login screen.

To make sure that no data is persisted, and won't be persisted asynchronously, we should call the regular logout which makes sure to freeze the syncing and cleans the database.

pdrj:
## What's this?
This cancels the login process when a user presses the back button during the login execution.

### Relationships

Closes #1411

## Why do we want this?
To enable users to bail out of logging in if they wish to do so.

## How is it done?
Cancellation tokens are used to cancel the ongoing login process.

There are still several issues with this:
- if we reach `onAuthenticated` callback before back is pressed, the user will still get logged in. We should decide if that's what we want, or do we still want to rollback the login.
- the `await` used in the `CancelLogin` function may cause the spinner over the login button to stutter
- signins with Google seem to handle the back button fine (without modifications), but I still added cancellation tokens to that operation as well because I didn't have enough time to investigate if it's really needed or not
- I feel there are still some race conditions that might happen here, mainly between the `onAuthenticated` and `CancelLogin` functions, which I haven't had time to dig into

## :squid: Permissions
Anyone, if they feel this is good to go.


amulware:
If this gets continued, I want to look over the logic since it has to do with sync stuff.

pdrj:
I'm slowly getting back to work on this issue.

@amulware what do you think about the question I posted in this PR? Specifically:

> if we reach onAuthenticated callback before back is pressed, the user will still get logged in. We should decide if that's what we want, or do we still want to rollback the login.

Being more specific, what happens in this scenario: 
1. User initiates login, spinner appears over the button.
2. `onAuthenticated` fires, causing first a sync, then navigation to the main view (spinner is still visible over the button).
3. During point 2 above, ie. during the sync, user presses back, as she doesn't know she's already authenticated, but waiting for other operations to complete. 
4. Even though back was pressed, the login still completes - user gets taken to the main view.

What do you think should happen? Some suggestions:
- roll back the login after sync completes (don't show the main view)
- as soon as `onAuthenticated` fires, navigate to the main view and init sync operation in  `MainViewModel` - if we do that, user would know that login is done and now she's waiting for something else (here more research should be done to identify any side effects moving sync out of login screen causes)
- leave it as it is.

amulware:
I think we should completely rollback login (i.e. freeze the sync and logout - you can check the actual logout flow from the settings view to make sure we definitely wait for everything to finished so no garbage data remains in the database afterwards).

Alternatively, we could have another loading state that's shown during the initial sync, but that'll be a bit of a bigger change and require designs and stuff. However, shouldn't the app still get the user 'back' i.e. to the login screen if the user navigates back while on this screen? I'm not exactly sure what an Android user would expect.

pdrj:
@amulware I've added a bit more logic around cancelling while already authenticated. Would you like to review it now? Honestly, I'm not yet super comfortable with the `C#` rx stuff, so please share any concerns ;).

Toggl.Core.UI/ViewModels/LoginViewModel.cs
```diff
@@ -172,37 +175,70 @@ public void Login()
...
+            setupLoginCancellationTokenSource();
+
+            View.GetGoogleToken()
                 .SelectMany(userAccessManager.LoginWithGoogle)
                 .Track(analyticsService.Login, AuthenticationMethod.Google)
-                .Subscribe(_ => onAuthenticated(), onError, onCompleted);
+                .Subscribe(_ => onAuthenticated(), onError, onCompleted, loginCancellationToken);
+        }
+
+        public bool IsLoginInProgress()
```
amulware: Could be made a property instead of a method.

Toggl.Core.UI/ViewModels/LoginViewModel.cs
```diff
@@ -172,37 +175,70 @@ public void Login()
...
+            if (loginCancellationTokenSource == null || !loginCancellationToken.CanBeCanceled)
+            {
+                return;
+            }
+
+            loginCancellationTokenSource.Cancel();
+            cancelLoginViaLogout();
+        }
+
+        private async void cancelLoginViaLogout()
```
amulware: Async Void is generally not advisable, is there a reason to use it here?

Toggl.Core.UI/ViewModels/LoginViewModel.cs
```diff
@@ -172,37 +175,70 @@ public void Login()
...
+                shouldCancelAfterSync = true;
+            }
+
+            if (loginCancellationTokenSource == null || !loginCancellationToken.CanBeCanceled)
+            {
+                return;
+            }
+
+            loginCancellationTokenSource.Cancel();
+            cancelLoginViaLogout();
```
amulware: This async method is not being awaited, so I'm really not sure what's gonna happen with that.

Toggl.Core.UI/ViewModels/LoginViewModel.cs
```diff
@@ -172,37 +175,70 @@ public void Login()
...
-            loginDisposable = View
-                .GetGoogleToken()
+            shouldCancelAfterSync = false;
+            setupLoginCancellationTokenSource();
+
+            View.GetGoogleToken()
                 .SelectMany(userAccessManager.LoginWithGoogle)
                 .Track(analyticsService.Login, AuthenticationMethod.Google)
-                .Subscribe(_ => onAuthenticated(), onError, onCompleted);
+                .Subscribe(_ => onAuthenticated(), onError, onCompleted, loginCancellationToken);
```
amulware: Do we really need the cancellation token? Where are we cancelling it?

Toggl.Core.UI/ViewModels/LoginViewModel.cs
```diff
@@ -231,6 +273,12 @@ private async void onAuthenticated()
 
             await UIDependencyContainer.Instance.SyncManager.ForceFullSync();
 
+            if (shouldCancelAfterSync)
```
amulware: Is this complexity needed? I'm pretty sure the regular logout will freeze and cancel sync at the earliest opportunity, which will also be much faster than waiting for it to finish.

Toggl.Core.UI/ViewModels/LoginViewModel.cs
pdrj: I misunderstood how these work. I'll refactor to get rid of them.

Toggl.Core.UI/ViewModels/LoginViewModel.cs
pdrj: Changed it, thanks.

Toggl.Core.UI/ViewModels/LoginViewModel.cs
pdrj: Refactored, it's no longer there 😄.

Toggl.Core.UI/ViewModels/LoginViewModel.cs
pdrj: It was overcomplicated - I think it was supposed to be handling a case when back was pressed after successful auth. It's done a bit differently now.

amulware:
Sorry, I won't find time for this anytime soon.

kornelijepetak:
@pdrj I agree with @amulware. I've checked the code for the logout interactor and it does seem to
* freeze the sync
* clear the data from DB and other user storage

so it seems safe to just interrupt the sync with this as soon as the user presses the back button.

However, the question for @amulware is whether we want this to be considered a logout in our analytics (which will happen if you just call the current interactor).

Otherwise, you either need a different, but very similar interactor, or parameterize that interactor and skip tracking when it's an abort. Or if @amulware would find this useful, maybe even expand the event with the information for the source of the logout (login abort, settings screen...).

Also, avoid using `volatile`s. They seem simple, but their intricacies are dirty. Heed Eric Lippert's advice: `Volatile fields are a sign that you are doing something downright crazy.` Use locks. They have stricter guarantees for ensuring thread safety than volatile.

kornelijepetak:
@amulware I guess that depends on how far the process comes to. But like you said, it's not really a frequent case.

heytherewill:
I just tested it and it works quite well! 

One thing though is that it cancels the thing and only yields control back after a while (when it reaches the cancel point). Is that by design? We could either give control back to the user immediately (which is hard, since it might run into race conditions) or at least show a toast saying `cancelling your login attempt` when they hit back, so they know we're working on it

daividssilverio:
This seems to work alright, but I agree with what @heytherewill said, it'd be nice to at least have some indicator that the login was cancelled.

Btw, I managed to get into a state where the initial loading never finishes and pressing back does nothing.

semanticer:
I think we shout also disable text fields and sign in with google button when login is in progress. We effectively block the functionality but the user can still click around and expect some results.

semanticer:
During my 40 min testing session, I can now replicate easily what @daividssilverio mentioned. 
You simply:
Tap login button with correct credentials
Tap home right after that as soon as you can (before login process finishes) 
Wait for 5s
Re-enter the app
You should find yourself on the login screen with never-ending loading and unable to hit back (back is just trying to stop the login request again and again)

Fun fact - this can happen in our production version as well, it's just less likely for some reason. 

The second issue mentioned by @pdrj https://toggl.slack.com/archives/C02APMYTM/p1577963967333700 was not reproducible for me at all. 

Testing devices: OG Pixel, Android emulator Nexus 5 (Both API 29)

semanticer:
I'm pretty sure I fixed the stuck login issue with the last commit but I was unable to reproduce the second issue mentioned by @pdrj again. I'm gonna assign this PR back to @pdrj to verify if it still sucks.

pdrj:
I can no longer reproduce the other issue.

semanticer:
@pdrj Cool, I guess we can merge it then?

pdrj:
Yeah, we're just waiting for an OK from @daividssilverio.

daividssilverio:
Not sure if it's intended or not.
Bug my tests yielded some possible problems:

1. The loading progress indicator never finished for me after pressing "back" in the login screen. (Happened once).
2. I've managed to get logged in after pressing back in the login screen. The app was closed by VS during (1)☝ 

Not sure if it's even worth taking a look. If I manage to replicate it soon I'll ask for changes, otherwise it could be worth just creating an issue to remind us to take another look later.

daividssilverio:
Eh... Actually, I can replicate it pretty easily. And it it seems to be happening when the login succeeds (at least partially) and I press `back` as soon as I press the button to login. Then If let the app go to bg and back again, it goes to the main screen and never finishes syncing.

Toggl.Core.UI/ViewModels/LoginViewModel.cs
```diff
@@ -200,9 +213,30 @@ public void GoogleLogin()
                 .Subscribe(_ => onAuthenticated(), onError, onCompleted);
         }
 
+
+        public void CancelLoginAtEarliestOpportunity()
```
daividssilverio: Also, this is a public method in the ViewModel and its behaviour wasn't tested.

daividssilverio:
No news on this PR? @semanticer @pdrj who's taking care of this PR now?

heytherewill:
Do we still want this? I say we don't

pdrj:
We'll revisit this (if it makes sense) once the great project with no name is ongoing.

PR #5721, opened 2019-08-06
📐 Use project color in edit view project name

sofokli1:
## What's this?
Improves the UI of project color in edit view and also works for RTL languages. Also looks more like the zeplin designs.

### Relationships
Ref #5561

## Why do we want this?
Better UI for all RTL or not

## How is it done?
1 bool change

### Why not another way?
this is the simplest

### Side effects
none

## Review hints
run the app

## :squid: Permissions
:squid:


sofokli1:
@Odrakir I went with this one https://app.zeplin.io/project/5728ce853dd1b182475f25e9/screen/5c4209831f32ea378bfa6866

PR #5722, opened 2019-08-06
Fix crash when starting time entry in empty log

juanlaube:
## What's this?
Found this crash in `develop` when starting a time entry in an empty log. It's caused by a leftover of `Firebase.Perf`.

## Why do we want this?
No crashes are the best crashes.

## How is it done?
🔥 

## :squid: Permissions
🦑


heytherewill:
Kind of a duplicate of #5706

juanlaube:
Thanks, I didn't see that one, we still need the fix in `develop` though 🤔

juanlaube:
I'm gonna close since this is a duplicate, if we really really really need this soon in `develop` we can cherry-pick #5706.

Issue #5698, opened 2019-08-02
Unify toolbars across the app


amulware:
## 🐛 Bug Report
In the wheel view, the save button in the toolbar is too close to the edge of the screen (or at least it's closer than the other various save buttons in the app).

![Screenshot_2019-08-02-11-19-44-442_com toggl giskard](https://user-images.githubusercontent.com/3987892/62364689-1ce41e80-b51a-11e9-84b7-ebace4b8cb8e.png)

## Context

Build/Branch: 2.1.6
Phone OS: Andriod 8
Phone model: redmi 5a

## Expected behavior
The toolbar layout should confirm to designs and be the same everywhere.

## To reproduce
Check the wheel view, compare to the save button in the edit/start views.

## Suggestions
Check designs. Also check if the title text in the same view is aligned correctly.

daividssilverio:
🤔 The `EditDuration` is the only `Activity` that's using the system provided toolbar menus.
The other `Activities` are using a custom fake toolbar. 
I think the right thing to do is to update the other activities.

amulware:
@maritoggl do you have a strong preference? My main concern is that they should all look the same.

maritoggl:
I agree that all should look the same, what's happening under the hood is up to you - do what's most reasonable.

dshr:
There are several other toolbars that need some attention, as flagged by @shantanuraj:

PR #5731, opened 2019-08-07
Bug Fix For Wheel View Toolbar

kpv1993:
## What's this?
It is a bug fix for "Save" button in wheel view. There was a very less padding on the right side.

### Relationships
None

## Why do we want this?
It is a design change.

## How is it done?
Added a new layout to inflate inside the item to add padding. I didn't find any other way to add margin or padding

### Side effects
No Side Effects

## References
https://github.com/toggl/mobileapp/issues/5698

## Review hints
The wheel view screen should be tested to see if the design is good.

## :squid: Permissions
No


Toggl.Droid/Activities/EditDurationActivity.cs
```diff
@@ -208,17 +208,19 @@ protected override void OnCreate(Bundle bundle)
         public override bool OnCreateOptionsMenu(IMenu menu)
         {
             MenuInflater.Inflate(Resource.Menu.GenericSaveMenu, menu);
+            var item = menu.FindItem(Resource.Id.SaveMenuItem);
+            item.ActionView.Click += (sender, e) =>
```
heytherewill: Why do it like this instead of using `OnOptionsItemSelected`?

heytherewill: Also, if you really must do it like so, prefer the functional approach of using `item.ActionView.Rx().Tap()` instead of using an event; the latter is much more error prone (e.g.: You forgot to unsubscribe to the event, something that is handled automatically by the observable approach).

Toggl.Droid/Resources/layout/CustomMenuItem.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+    <TextView
```
heytherewill: Weird indenting

kpv1993: Will make it proper

Toggl.Droid/Activities/EditDurationActivity.cs
kpv1993: I tried in OnoptionsItemSelected, but it was not working. So I had to do this way. 
I didn't know about this Tap() that's why I used the event. Will change that.

kpv1993: Requested changes are done

Toggl.Droid/Activities/EditDurationActivity.cs
```diff
@@ -208,17 +208,21 @@ protected override void OnCreate(Bundle bundle)
         public override bool OnCreateOptionsMenu(IMenu menu)
         {
             MenuInflater.Inflate(Resource.Menu.GenericSaveMenu, menu);
+            var item = menu.FindItem(Resource.Id.SaveMenuItem);
+            item.ActionView.Rx().Tap()
+                .Subscribe(_ =>
```
daividssilverio: We can somehow improve this by changing it to:

```
.Select(_ => item)
.Subscribe(_ => OnOptionsItemSelected(item))
```

That would minimize the changes and keep things more consistent.

Toggl.Droid/Activities/EditDurationActivity.cs
```diff
@@ -208,17 +208,21 @@ protected override void OnCreate(Bundle bundle)
...
+                })
+                .DisposedBy(DisposeBag);
             return true;
         }
 
         public override bool OnOptionsItemSelected(IMenuItem item)
         {
             switch (item.ItemId)
             {
-                case Resource.Id.SaveMenuItem:
```
daividssilverio: Then you can leave this as it was.

heytherewill:
I vote for closing this PR, going back to the drawing table and figuring out how to best add uniformity to all toolbars in the app.

heytherewill:
This has conflicts and is not currently active, so I'm closing it until we figure out a better way of doing it.

PR #5732, opened 2019-08-07
Expose ImmutableLists in our ViewModels

### LINKED ISSUE #5643: Use collections more consistently in our codebase
## 🚧 Refactor Proposal
@juanlaube has pointed out during our Buenos Aires discussions that we've had problems with enumerating sequences multiples times. A good way to avoid that is making sure that we are enumerating things that do expensive operations.

## Technical aspects
The safest route I see is simply making sure that we are always exposing `IImmutableList` from ViewModels. This will ensure that we enumerate only once before displaying any kind of list. 

Whoever works on this, please keep an eye open for possibilities of improvement. While  we want to enumerate, we only wanna do it at the last possible moment

### LINKED ISSUE #5782: Reports tab isn't updating its UI until you touch it
## 🐛 Bug Report
The reports's tab isn't updating/showing the UI the first time you open it.
It looks like this:

## Context
Build/Branch: This was happening sometimes on 2.2.1 and seems to have been intensified with #5767.
Phone OS: Android

## Expected behavior
All of the reports tab's content should be loaded when the tab is opened.

## To reproduce
Close the app, open it again, select the reports tab.
If you tap the screen, the content is loaded.

## Suggestions
This seems be a problem with the custom views in that tab, somehow they are not being re-drawn the first time they are opened.
What's even weirder is that it seems that the content loads very briefly and then everything gets all white.

heytherewill:
## What's this?
This makes sure that all ViewModels expose the same type of collection: `IImmutableList`
Collaterally, I'm also ensuring that collections get enumerated only where needed.

### Relationships
Closes #5643 
Closes #5782

## Why do we want this?
To avoid enumerating things more than once + Ensuring our lists are unmodifiable.

## How is it done?
Exposing `IImmutableList` everywhere

### Side effects
I'll need a .net person to take a look at things here to ensure I'm not enumerating things more than once. For sanity checking.

## :squid: Permissions
Only myself


Toggl.Droid/Adapters/BaseSectionedRecyclerAdapter.cs
```diff
@@ -136,9 +137,9 @@ private void setItems(IList<SectionModel<TSection, TItem>> newItems)
             }
         }
 
-        private IList<Either<TSection, TItem>> flattenItems(IList<SectionModel<TSection, TItem>> newItems)
+        private IImmutableList<Either<TSection, TItem>> flattenItems(IImmutableList<SectionModel<TSection, TItem>> newItems)
         {
-            var flattenedItems = new List<Either<TSection, TItem>>();
+            var flattenedItems = ImmutableList<Either<TSection, TItem>>.Empty;
```
daividssilverio: This doesn't need to be immutable, right?
We are building this list here and those `flattenedItems = flattenedItems.Add` inside the loop looks more expensive than it should in this case.

I think all we need to do is to convert the list to `Immutable` before returning it.

Toggl.iOS/ViewSources/Generic/TableView/BaseTableViewSource.cs
```diff
@@ -12,41 +11,42 @@ namespace Toggl.iOS.ViewSources
...
-        public void SetItems(IEnumerable<TModel> items)
+        public void SetItems(IImmutableList<TModel> items)
         {
-            IEnumerable<TSection> sections = new List<TSection>();
+            var sections = ImmutableList<TSection>.Empty;
             if (items != null)
             {
                 var newSection = new TSection();
-                newSection.Initialize(default(THeader), items);
+                newSection.Initialize(default, items);
```
juanlaube: There's a null ref exception here when items is empty (well, not exactly here, but in one of the `TSection`s, `AnimatableSectionModel`). Adding `items.Count > 0` to the `if` should solve the problem.

heytherewill:
Addressed your requests. Please test the whole thing again, as I merged develop in here :)

daividssilverio:
I'm comparing against production 2.2.1:
Problems so far:
- Start time entry list on droid is janking a bit.
- Reports were all blank every time I opened the reports tab when I tried this branch before. Now I can't replicate it 🤷‍♂ .
- Reports Calendar shows the minimum value when you select a date, close the sheet and open it again.
- Reports Calendar is jumpy. The first time it's opened it's huge. It's quite fast though.

I found other issues, but they are all in develop/2.2.1, so I'll just open some issues for them.

heytherewill:
> Start time entry list on droid is janking a bit.
Reports were all blank every time I opened the reports tab when I tried this branch before. Now I can't replicate it 🤷‍♂ .

These two seem to be happening in develop as well, could you confirm?

daividssilverio:
Hey @heytherewill only the sheet-related issues are from this branch. Please take a look at:

- Reports Calendar shows the minimum selectable date value (month - year label) when you select a date, close the sheet and open it again.
- Reports Calendar is jumpy. The first time it's opened it's huge. It's quite fast though

PR #5733, opened 2019-08-07
Load culture info on startup

### LINKED ISSUE #5702: Handle fallback language on droid
When looking for weirdness on #4831, I found a weird behaviour: When you select a language that's not currently supported by us, the framework somehow keeps the last supported previously selected language instead of going back to the default English.

Let's make it so that when selecting an unsupported language English should be used.

### LINKED ISSUE #5712: Load CultureInfo on startup
It looks like iOS doesn't load and setup the `CultureInfo` properly during startup like Android.

We need to make sure the user language is properly loaded in the app if we do support it.

We need to call:

```
Thread.CurrentThread.CurrentCulture = new CultureInfo(netLanguage);
Thread.CurrentThread.CurrentUICulture = new CultureInfo(netLanguage);
```

during startup before strings start being loaded.
Since we'll need to do something similar on Droid, We could set this up on `AppStart` and also handle fallback languages in the same place.

### LINKED ISSUE #5713: Handle fallback language on iOS
While testing and trying out languages on iOS for #4831, I found some other interesting behaviour on iOS. For some reason, the language code in the phone accessed by `NSLocale.PreferredLanguages[0]` can be something like "ja-US"... which is bananas to me (maybe because I had the system language to Japanese, but still had a US keyboard available... idk).

Either way, we need to support that sort of behaviour. And Since droid might show similar behaviour as described on #5702, let's make this shared as possible. The Xamarin samples for handling `Resx` files with `Xamarin.Forms` looks  like a reasonable example: https://github.com/xamarin/xamarin-forms-samples/tree/master/UsingResxLocalization.

daividssilverio:
## What's this?
This PR adds a small service to setup the right culture info during app startup.

### Relationships
Closes #5712
Should be enough to close #5702 as well. 
Should be enough to close #5713 as well.

## Why do we want this?
To be able to load localization resources properly for both droid and iOS.

## How is it done?
The `BaseLocalizationService` class does most of the job; 
Android and iOS sides only implement `GetCurrentNativeLanguageCode` and return the current language of the system defined by the user.
Other methods were left to be overridden when necessary, that is, when we find an issue with some language during the translation process.

### Why not another way?
This shares most of the code for now and seems flexible enough.

### Side effects
None.

## Review hints
You may create a translated resource, for example `Resources.ja.resx`, change some text entries, change the device language to Japanese and see the results. Change it back to your language or English, and check that the apps should be back to English (the old text entries at least).

## :squid: Permissions
🦑 i18n team


Toggl.iOS/Services/LocalizationServiceIos.cs
```diff
@@ -0,0 +1,15 @@
+using Foundation;
+using Toggl.Core.UI;
+
+namespace Toggl.iOS.Services
+{
+    public class LocalizationServiceIos : BaseLocalizationService
+    {
+        protected override string GetCurrentNativeLanguageCode()
+        {
+            return NSLocale.PreferredLanguages.Length > 0
```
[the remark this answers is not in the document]
heytherewill: Also, do we really want this to return null? Wouldn't it be better to just return english as a fallback for maximum safety?

daividssilverio:
> Since we are running away from creating new services, I'll suggest you move the method to getting the current language code into the a property of IPlatformInfo. This way you can make the service initialization become part of the App.cs class :)

I don't like this idea really much, the `IPlatformInfo` sounds so query-related, it feels weird put this stuff there.

> Since we are running away from creating new services

And I thought we are running way from new data sources only 🤔

Toggl.Core.Tests/UI/AppStartTests.cs
```diff
@@ -182,5 +193,97 @@ public void ReturnsLoggedInWhenTheUserHasLoggedInPreviously()
...
+            
+            [Fact, LogIfTooSlow]
+            public void SetsTheCultureInfoToEnglishWhenWhenThePlatformInfoReturnsAnInvalidCurrentNativeLanguageCodeWithAnInvalidLanguageAndCultureCodes()
+            {
+                string invalidLanguageCode = "00_11";
+                PlatformInfo.CurrentNativeLanguageCode.Returns(invalidLanguageCode);
+
+                App.LoadLocalizationConfiguration();
+
+                Thread.CurrentThread.CurrentCulture.TwoLetterISOLanguageName.Should().Be("en");
```
heytherewill: This feels very error-prone because of the way you check a static property 🤔

Toggl.iOS/PlatformInfoIos.cs
```diff
@@ -163,5 +164,13 @@ private static string toDeviceModel(string identifier)
             }
         }
 
+        public override string CurrentNativeLanguageCode => getPreferredLanguageCode();
+
+        private string getPreferredLanguageCode()
+        {
+            return NSLocale.PreferredLanguages.Length > 0
```
heytherewill: Why didn't you follow my suggestion of using `NSLocale.PreferredLanguages.FirstOrDefault() ?? DefaultLanguageCode`?

Toggl.Core.UI/AppStart.cs
```diff
@@ -1,16 +1,55 @@
 using System;
+using System.Globalization;
+using System.Linq;
+using System.Threading;
 
 namespace Toggl.Core.UI
 {
     public sealed class AppStart
     {
+        public static readonly string DefaultLanguageCode = "en";
```
heytherewill: Don't we have a constant file to put this on?

Toggl.Core.Tests/UI/AppStartTests.cs
daividssilverio: I don't see another way of checking this 🤔

Toggl.iOS/PlatformInfoIos.cs
daividssilverio: Well... it's that I've checked the implementation details of `FirstOrDefault` and it looked like a bit of an overkill. But I can see that it would read better.

Toggl.Core.UI/AppStart.cs
daividssilverio: Sure, we can put it on `Constants`.

PR #5735, opened 2019-08-08
Fixed a Date Picker Bug

kpv1993:
## What's this?
Bugfix for the date-picker dialog whose title is getting automatically added in some phones

### Relationships
None

## Why do we want this?
To remove the redundant data

## How is it done?
A blank title has been added 

### Why not another way?
Open for suggestions

### Side effects
None

## Review hints
Before:

Screenshot_20190808-143102

After:

Screenshot_20190808-183226

## :squid: Permissions
None


daividssilverio:
Will and Juan already approved this.

PR #5740, opened 2019-08-08
Fix description textfield cursor color

### LINKED ISSUE #5736: Cursor is not visible in start and edit view
## 🐛 Bug Report
The cursor when editing the description text field is not visible in the start and edit views.

## Context

Build/Branch: creme-brulee-2.2
Phone OS: Android
Phone model: 

## Expected behavior
The cursor should be visible.

## To reproduce
Try to edit the description

## Suggestions
Probably related to the toolbar colors change, my guess is the cursor is still white.

juanlaube:
## What's this?
Fixes the cursor color for the description text field in the start and edit views.

### Relationships

Closes #5736 

## Why do we want this?
No bugs.

## How is it done?
XML magic.

### Why not another way?
🤷‍♂ 

### Side effects
None.

## :squid: Permissions
🦑


heytherewill:
Won't black, which is the same color as the text, become hard on the eyes when the user selects text?

juanlaube:
Nope, I tried and for some reason it is blue-ish ¯\_(ツ)_/¯

PR #5758, opened 2019-08-13
iOS 2.2.1 Cantaloupe hotfix

arturs105:
## What's this?
This is a mergeback PR for iOS 2.2.1 release. The only change from 2.2 is fixing a null exception in `MainViewController.ViewDidDisappear`.

## Why do we want this?
Because that null exception is the [biggest crash in 2.2](https://appcenter.ms/orgs/Toggl/apps/Toggl-iOS/crashes/errors?version=2.2&appBuild=&period=last30Days&status=&errorType=unhandlederror&sortCol=count&sortDir=desc) _and_ it is so easy to fix.

## How is it done?
A simple null check.

### Why not another way?
I could have used a combination of `.?` and `?? [fallbackValue]`, but this approach seemed easier to understand.

## :squid: Permissions
Only me, when 2.2.1 goes out.


heytherewill:
Why has this not been merged into develop yet?

arturs105:
😬

PR #5767, opened 2019-08-14
Improve bailout code

### LINKED ISSUE #5307: Improve safety measures to make sure ReactiveActivities always have valid states
Let's make what was introduced in #5244 more fool-proof.
We need code, but it could be improved.

@heytherewill suggested something like sealing `OnCreate` in the `ReactiveActivity` and add methods that inheritors would override; something like:

```
protected sealed override void OnCreate(Bundle bundle)
{
  BeforeBaseOnCreate(bundle);
  base.OnCreate(bundle);
  ViewModel?.ViewCreated();
  AfterBaseOnCreate(bundle);
}

protected virtual void BeforeBaseOnCreate(Bundle bundle) {}
protected virtual void AfterBaseOnCreate(Bundle bundle) {}
```

heytherewill:
## What's this?
This moves the bailout boilerplate to the ReactiveActivity, making the rest of our code significantly clearer.

### Relationships
Closes #5307

## Why do we want this?
First, to make sure that new Activities don't introduce breaking behavior
Second. to make our Activity code more readable (and, oh boy, it is a lot more readable).

## How is it done?
Moving the code of OnCreate to base, sealing base and exposing two methods that are extension points of our lifecycle. Most Activities only need one of the two.

### Why not another way?
I'm open to suggestions

### Side effects
Not only is it impossible to introduce failing activities, it's also impossible to have wrong transition animations as well.

## Review hints
Make sure to check each activity to make sure I didn't derp and change some style/layout/whatever.
Also, check all activities in app. **All of them**

## :squid: Permissions
Only myself


Toggl.Droid/Activities/EditTimeEntryActivity.Onboarding.cs
```diff
@@ -11,6 +11,18 @@ public sealed partial class EditTimeEntryActivity
     {
         private PopupWindow projectTooltip;
 
+        protected override void OnResume()
```
daividssilverio: Just my opinion. But I don't like the lifecycle stuff outside of the main Activity part.
Why: 
- It will cause people to come over to this dreaded file
- It can make people question about consistency when overriding lifecycle callbacks
- It's harder to get a grasp of what's going on in the Activity just from looking in the main part.

Toggl.Droid/Activities/ReactiveActivity.cs
```diff
@@ -7,55 +7,65 @@
...
+            int layoutResId,
+            int correctTheme,
+            ActivityTransitionSet transitions)
         {
+            this.layoutResId = layoutResId;
+            this.transitions = transitions;
+            this.correctTheme = correctTheme;
         }
 
         protected ReactiveActivity(IntPtr javaReference, JniHandleOwnership transfer)
```
daividssilverio: Question: Why is it now necessary to have this other constructor declared in all other Activities? If it's mandatory, why it wasn't there before?

Toggl.Droid/Activities/ReactiveActivity.cs
```diff
@@ -7,55 +7,65 @@
...
                 .Get<TViewModel>();
 
+            if (ViewModel == null)
+            {
+                bailOutToSplashScreen();
+                return;
+            }
+
             ViewModel?.AttachView(this);
+            OnViewModelInitialized(bundle);
```
daividssilverio: I feel like it would be safer to have:

```
ViewModel?.AttachView(this);
OnViewModelInitialized(bundle);
```

after `InitializeViews`.

Why: You've used `OnViewModelInitialized` as a restore point (and it looks like one because of the `bundle`).
It would be better to enforce that `OnViewModelInitialized` can only be used to restore view model related state, that is, it can't restore any view state; or just move it after `InitializeViews`.

Toggl.Droid/Activities/EditTimeEntryActivity.Onboarding.cs
heytherewill: I get your concern. I can revert these changes. My hope was to get everything related to onboarding on a single file to make it easier for deleting 😄. Your point is valid

Toggl.Droid/Activities/ReactiveActivity.cs
heytherewill: Before the activities didn't have a constructor of their own. Now they do. This constructor is used by Xamarin when the .net object is ded but the Java object is still alive (multiple GCs are hard)

Toggl.Droid/Activities/ReactiveActivity.cs
heytherewill: I don't think I follow. How can we make it sure that consumers will only initialize ViewModel code if we give them initialized Views? 🤔

heytherewill:
I'd keep the ViewBinds naming, mostly because this is the same nomenclature that ButterKnife uses 🤷‍♂

Toggl.Droid/Activities/ReactiveActivity.cs
daividssilverio: Multiple GCs suck.

Toggl.Droid/Activities/ReactiveActivity.cs
daividssilverio: By enforcing, I mean adding a comment or something like that. 

Moving it after `InitializeViews` is another option, that way if a consumer decides for some reason to do something with the views, they will be already bound. Moving it after `InitializeViews` won't make sure users will only initialize VM code, it will just make this more flexible.

daividssilverio:
Found some issues while testing:

- Transitions from "Select Item From List" back to previous screen looks very off. The items from the selection screen flash in a weird way.
- Transition from the wheel back to previous screen also looks off, like ☝️.
- Transition from start time entry back to main tab also looks off, like ☝️.
- Reports screen sometimes it's all white or some of the inner charts are.
- Timer tab is janking when you go back to it.
- Transition between login and signup screens fading transitions are also looking off.
- Transition between forgot password and login screen is a bit off as well.

daividssilverio:
> Timer tab is janking when you go back to it.
This didn't come from this branch, it's on develop and production.

daividssilverio:
> Reports screen sometimes it's all white or some of the inner charts are.

Yeah, that's also in develop, but not in production. We can bisect it ;)

daividssilverio:
Also, your last commit didn't fix the transition problems.

daividssilverio:
> Transition between login and signup screens fading transitions are also looking off.

Also present on develop.

PR #5768, opened 2019-08-14
Extract all the Droid strings into Resources.resx

### LINKED ISSUE #4827: Extract strings from `strings.xml` into Resources.resx
There's a bunch of strings yet on android.
We've got to extract them from `Strings.xml`, and of course, rebind all activities, fragments, viewholders, dialogs, custom views, etc.

There's gonna be a lot of changes, but they are simple changes.

Don't forget to leave a comment in the extracted string indicating `ViewModel` relationships.
If the string doesn't belong to any `ViewModel`, indicate any other context to help understand the meaning of the text.

dshr:
## What's this?
A truckload of small changes that move all string names into Resources.resx and set them in the appropriate activity code.

### Relationships
Closes #4827 

## Why do we want this?
🌏 

## How is it done?
By setting the appropriate element's text/hint/helper text to a value from Resources.resx in activity/view holder code

### Why not another way?
🤷‍♂ 

### Side effects
See comments

## :squid: Permissions
i18n crew, I choose you


Toggl.Core.UI/ViewModels/RatingViewModel.cs
```diff
@@ -132,7 +132,7 @@ private string callToActionDescription(bool? impressionIsPositive)
                 return string.Empty;
 
             return impressionIsPositive.Value
-                   ? Resources.RatingViewPositiveCallToActionDescription
+                   ? Resources.RatingViewPositiveCallToActionDescriptionIos
```
dshr: @daividssilverio since we're setting the string from Core.UI code, is there a way to use a compiler flag to differentiate between iOS and Droid?

Toggl.Shared/Resources.resx
```diff
@@ -532,21 +535,36 @@ Start the timer and your most used time entries will appear here.</value>
...
     <value>By tapping I Agree, you understand and agree to Toggl’s Terms of Service and Privacy Policy.</value>
+  </data>
+    <data name="TermsOfServiceDialogMessageDroid" xml:space="preserve">
+    <value>By tapping "I Agree", you
+understand and agree to Toggl\'s</value>
   </data>
     <data name="CountryOfResidence" xml:space="preserve">
     <value>Country of residence</value>
   </data>
+  <!-- iOS -->
```
dshr: There are numerous cases like this where the copy between apps differs ever so slightly, I've marked then with comments. @maritoggl @daividssilverio what shall we do in cases like this?

daividssilverio: I think it makes sense to unify most of the strings, with the exception of actually platform specific stuff.

Toggl.Droid/Resources/layout/CalendarSettingsActivity.axml
```diff
@@ -36,6 +37,7 @@
             android:textColor="@android:color/black"
             android:text="@string/LinkCalendars"/>
```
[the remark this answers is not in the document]
daividssilverio: There's a lot of these, so I didn't leave a comment for each.

daividssilverio:
Yeah... we'll need to keep the `@string/Reports` and `@string/NewTimeEntry` on `strings.xml` for the droid tiles. Please leave the copy in `Resources.resx` so that we can update the tiles manually when there's a new language added.

Toggl.Core.UI/ViewModels/RatingViewModel.cs
daividssilverio: 1. A little bit 🤢 , but we can check the PlatformInfo to check which OS we are handling.

2. Or we could just ditch "App Store" and "Play Store" with something neutral "applications store", which also isn't really good... is it? Oh boy... Maybe option 1 is the way to go.

dshr: Option 1 will be a nightmare in tests tbh. I guess we can refactor this so we set the string in the Activity/VC, while the VM handles the rest?

dshr:
> It looks good overall.
> I have two requests so far:
> 
>     * Please delete the strings from `Toggl.Droid/Resources/values/Strings.xml` and `Toggl.Droid/Resources/values-v23/Strings.xml`, except by the `AppName`
> 
>     * Please remove every `android:text="@string/ref"`, `android:hint="@string/ref"` and `android:ContentDescription` (which needs to be updated as well). Please replace them with their `tools:*` equivalent if there's one.
> 
> 
> And I'll investigate whether or not we can replace the Labels from the multiple `TileServices` we have.

Removing all entries from Strings.xml breaks builds since the `.axml`s still contain references to them. We could replace the values in there, but that's a crapload of work :(

Toggl.Core.UI/ViewModels/RatingViewModel.cs
daividssilverio: That sounds fair enough.

daividssilverio:
Hey @dshr, please take a look at #5772, merge it into your branch if it is of your liking.
It should fix the droid build here as well.

PR #5775, opened 2019-08-16
Fix cursor position when holding backspace

### LINKED ISSUE #5474: Start View holding backspace doesn't work properly
## 🐛 Bug Report

When I have a longer sentence entered in the start view and hold the backspace button, it starts deleting single letters fine, but once it starts deleting entire words at a time, it fails, and the cursor just moves to the start of the text.

## Context

Build/Branch: 2.0 app store build
Phone OS: iOS 12.3.1
Phone model: iPhone 6

## Expected behavior

It should delete word by word, instead of moving just the cursor.

## To reproduce

- enter a longer sentence, like this one, in the start view
- hold and continue holding backspace 

## Suggestions
Probably something related to our fancy custom parsing stuff.

juanlaube:
## What's this?
Fixes a bug on the cursor position when holding the backspace button in the virtual keyboard.

### Relationships

Closes #5474 

## Why do we want this?
We don't like bugs

## How is it done?
For some unknown reason we were updating the cursor position when setting the text field, removing that solves the bug.

### Why not another way?
🤷‍♂ 

### Side effects
I couldn't find any, but I have no idea why were setting the cursor position so I might be missing something.

## Review hints
Test on a device, use projects and tags, in the different parts of the description.

## :squid: Permissions
No, and I'd like to have two reviews.


arturs105:
This is working better than previously, but now the first typed word (last deleted) can't be deleted by holding the backspace. Steps to reproduce:
1. Enter a long sentence
2. Start holding the backspace key
3. Keep holding it until it starts deleting whole words
4. Hold it a bit more until it gets to the last word, which isn't deleted

juanlaube:
Follow up issue #5785

PR #5780, opened 2019-08-19
Add shared element transitions

heytherewill:
## What's this?
An old branch I had lying around since BA. Updated it and I figured I might as well see what people think of it

## Why do we want this?
Prettier animations when going to the EditTimeEntry view.

## How is it done?
Android shared element transitions

### Side effects
A pretty darn cool applications

## :squid: Permissions
Only myself


daividssilverio:
This is looking cool. But it needs some work.
The duration could also be animated.
The stop button is weirdly positioned when you open the edit time entry view

| grouped time entries | single time entry |
| ---- | ---- |
|  |  |

daividssilverio:
It looks like the constraints are being messed up during the transition somehow.

heytherewill:
Closing this for the time being, might revive it later after going through a designer

PR #5787, opened 2019-08-20
Fix missing iOS translatables

### LINKED ISSUE #5770: Fix missing iOS translatable items
I found some strings views that still have their text fixed in the xibs, storyboards, cells, etc:

Let's fix them:

- Fix CalendarSettingsTableViewHeader: "Toggl needs access to your calendar in order to display events."
- Reports Calendar Month/Year
- Settings Siri "Shortcuts"/"Workflows"
- Login/SignUp Email/Password hints
- Navigation "Back" on settings
- Calendar settings user calendars sections
- Calendar ui labels when editing a time entry, it shows PM/AM only. Maybe this isn't fixable :thinking:

dshr:
## What's this?
A set of fixes that translate what hasn't been translated yet.

### Relationships
Closes  #5770

## Why do we want this?
i18n!

## How is it done?
By adding explicit back buttons, using appropriate CultureInfo's, formatting things where needed, etc

### Why not another way?
This is the right way tho

### Side effects
Not yet.

## Review hints
Cross-check against the source issue.

## :squid: Permissions
Globe people


Toggl.iOS/ViewControllers/MainViewController.cs
```diff
@@ -101,6 +101,12 @@ public override void ViewDidLoad()
             StartTimeEntryButton.AccessibilityLabel = Resources.StartTimeEntry;
             StopTimeEntryButton.AccessibilityLabel = Resources.StopCurrentlyRunningTimeEntry;
 
+            var backButton = new UIBarButtonItem
+            {
+                Title = Resources.Back
+            };
+            NavigationItem.BackBarButtonItem = backButton;
+
```
Odrakir: Why do we need this? Shouldn't this button be translated by the system?

Toggl.iOS/ViewControllers/ReportsCalendarViewController.cs
```diff
@@ -59,14 +61,28 @@ public override void ViewDidLoad()
                 .DisposedBy(DisposeBag);
 
             ViewModel.CurrentMonthObservable
-                .Select(month => month.Year.ToString())
-                .Subscribe(CurrentYearLabel.Rx().Text())
```
Odrakir: It seems like `CurrentYearLabel` doesn't exist anymore, but the outlet is still there in the `.designer.cs` file, please :fire:

Toggl.iOS/ViewControllers/Settings/SettingsViewController.cs
```diff
@@ -50,6 +50,12 @@ public override void ViewDidLoad()
             SiriSettingsHeaderLabel.Text = Resources.SiriSettings;
             TimerSettingsHeaderLabel.Text = Resources.TimerSettings;
 
+            var backButton = new UIBarButtonItem
+            {
+                Title = ViewModel.Title
+            };
+            NavigationItem.BackBarButtonItem = backButton;
+
```
Odrakir: Same deal as before, this should be done by the system, doesn't it work?

Toggl.iOS/ViewControllers/ReportsCalendarViewController.cs
```diff
@@ -59,14 +61,28 @@ public override void ViewDidLoad()
...
+                .Select(month =>
+                {
+                    var dateTime = month.ToDateTime();
+                    var pattern = CultureInfo.CurrentCulture.DateTimeFormat.YearMonthPattern;
 
-            ViewModel.CurrentMonthObservable
-                .Select(month => month.Month)
-                .Select(CultureInfo.GetCultureInfo("en-US").DateTimeFormat.GetMonthName)
-                .Subscribe(CurrentMonthLabel.Rx().Text())
+                    var yearMonthString = dateTime.ToString(pattern);
```
Odrakir: This is a lot of logic for the view (which is not tested) I liked the separate month-year approach better. But I'm not going to fight you about it, I just wanted to mention it.

Toggl.iOS/ViewControllers/MainViewController.cs
dshr: I'm afraid it doesn't get translated for whatever reason :(

Toggl.iOS/ViewControllers/ReportsCalendarViewController.cs
dshr: The separate year month approach would require some view reordering, since in some formats you have the year come first, like "2019, August"

dshr:
🔥'd the lines from the .designer.cs file

Toggl.iOS/ViewControllers/MainViewController.cs
Odrakir: I've been testing this and I know exactly why that's happening. You are missing the following in the `Info.plist`
```
<key>CFBundleDevelopmentRegion</key>
	<string>en</string>
	<key>CFBundleLocalizations</key>
	<array>
		<string>en</string>
		<string>es</string>
                ...
	</array>
```

dshr: This worked a charm, cheers ✨

PR #5747, opened 2019-08-09
Project: i18n preparations

daividssilverio:
## Why do we want this?
Code changes are there to make it easier to have a streamlined translation process.
We'll have a single source of truth for the translations, the `Resources.resx` file.

The new documentations, issue and pull request templates, together with the new [GitHub project](https://github.com/toggl/mobileapp/projects/74) are there to make it possible for anyone to contribute with translations for the Toggl apps.

## How is it done?
The iOS and Droid projects had all of their static text entries moved from views and other native solutions (xibs, storyboards, xmls) to the `Resources.resx` file.

The language is loaded during app startup in the `AppStart`.

### Side effects
It's gonna be more complicated to make releases and make text changes now.
Everybody is gonna have to be more careful when making changes that affect texts.

## Review hints
Review by commit.
If you want to test if the translations work, please add a translated file, `Resources.ja.resx` for example, prefixing every text entry or something like that, change the device preferred language to the language you've translated and check the results. Yes, the app must be restarted for the translation to be kick in.

## :squid: Permissions
Merge, don't squash. Only me.

PR #5791, opened 2019-08-21
Replace manual mode text

### LINKED ISSUE #3672: Improve UX for changing manual/timer modes

Odrakir:
## What's this?
Changes the text for the manual mode switch in settings

I did change the text to make it a little shorter from:

> This option sets the default behavior for the start button. Regardless of the current option, long pressing the button will always activate the alternative mode.

to

> Sets the default behavior for the start button. Long pressing the button will always activate the alternative mode.

Let me know if that's ok, maybe @toggl/androids should change it too.

### Relationships
Closes #3672 

Unless @amulware says we should change the start button UI in this same PR (otherwise I'll create another issue)

## Why do we want this?
This is very straight forward.

## :squid: Permissions
Wait for @amulware


amulware:
Why do we want this though? It's not clear to me from the issue.

Odrakir:
🤔 Why do we want what? The text? As described in the issue this is so the long-press thing is a little bit clearer. We are taking the same approach as Android did. I still thing this is not enough and we should do something in the button itself.

amulware:
I mean: The text was already changed - why are we changing it again to a shorter version?

amulware:
No wait.. the text was not changed yet? I thought it was... My bad. Why did you go with this shorter version though?

Odrakir:
I merely took out the redundant parts to make it shorter. I just seen you had posted another version in the issue, maybe I can use that one instead. Or maybe we could ask @maritoggl to give us the definitive one.

amulware:
I thought about it and I think I prefer your option. It might be clearer, being shorter, and it contains the same information.

amulware:
The same string should be used by Android right? Or do we need to do something separately there? If so, please make sure that happens. :)

Odrakir:
Judging by #3672 Android is not using the Resources file for this (and many other texts, apparently). They should, right?

Also... What happens with the animation? Should I create another issue?

amulware:
I'd check with @daividssilverio on what Android is doing, especially now with the i18n project changing a lot. But I see now that the previous pull request only changed it on Android then. This should really be simpler then, I hope it is with the i18n changes.

And yes, if you close the issue, please create follow ups (one per platform) about the animation, and link this thread where it's being discussed: https://toggl.slack.com/archives/CMB9FDE5D/p1565778022020100

maritoggl:
> I merely took out the redundant parts to make it shorter. I just seen you had posted another version in the issue, maybe I can use that one instead. Or maybe we could ask @maritoggl to give us the definitive one.

The one you proposed is good 👍

daividssilverio:
Hey y'all, both droid and iOS should use only `Resources.resx` defined string, unless ofc it's a platform limitation on specific use case. Either way, every text entry should be at least present in the `Resources.resx` so that it can be translated.

Odrakir:
So #5678 should be changed, right?

daividssilverio:
Yeah, and #5747 actually changed it already.

PR #5794, opened 2019-08-22
Fix system's autocomplete on start view

### LINKED ISSUE #5777: iOS autocomplete isn't working properly in the start view
## 🐛 Bug Report
When a word from iOS autocomplete suggestions (the bar right above the keyboard) in the start time entry view is selected, there should be a space after the word, but in our custom start time entry view that space gets removed. But wait, there's more - when you try to enter the space manually, the last character in the last word gets _changed_ to a space.

## Context

Build/Branch: Latest `develop`
Phone OS: iOS 12
Phone model: tested on an iPhone X and iPhone 8 simulator

❗️ This does not happen in iOS 2.2.1 (from App Store), so this is some kind of regression ❗️ 

## Expected behavior
When you select a word from the bar above the keyboard that word should be followed a by a space.

## To reproduce
1. Open edit view
2. Enter a character or 2 so the iOS autocomplete suggestions show up
3. Tap on one of the suggestions
4. 💥 No space after the word
5. Now if you press space, the last character gets replaced by a space

Odrakir:
## What's this?

This fixes an error when using system's autocomplete in the start view. It wasn't leaving a space after the word plus it was doing some other weird stuff.

## Relationships
Closes #5777

## How is it done?
After some research (1'5 hours according to my Toggl timer) everything seemed to point to some threading issue. When using the system's autocompletion the text didn't get there until after the event was triggered.

The culprit is the line in StartTimeEntryViewController in which we try to move the parsing operations to a background thread, but instead of using ObserveOn as we should, we were using SubscribeOn, which doesn't move the next operators to another thread, but will change the thread in which the observable operates (generates the event). Which in this case can cause any kind of weird effects as we are dealing with UI and thus the main thread.

@heytherewill As I think you wrote this, please take a look at how SubscribeOn and ObserveOn operate, as this is important and I think the whole deal happens because of that confusion. Also, we might need to check other places for this mistake.

Some info: http://reactivex.io/documentation/operators/subscribeon.html

I think this image is specially helpful

## Why not another way?
While working on this I do get the feeling that this whole deal is far more complicated than it needs to be. Maybe some day we could spend some time refactoring it (again). I can't pin point anything right now, but I still have the feeling that it behaves a little weirdly sometimes.

## 🦑 Permissions

Go, but let @heytherewill read this.


Odrakir:
Ummm... It's hard to tell what's going on in the Calendar, a `SubscribeOn` followed by `ObserveOn` both to a background thread... 🤔 I'm not saying that could not work, but there are clearer ways to do what that does (if that is the intention). 

Maybe @daividssilverio should take a look, but it smells like the same thing to me, yes.

PR #5798, opened 2019-08-22
🐎 Move suggestions binding to a background thread

### LINKED ISSUE #5665: Profile the Start Time Entry View model startup
This view takes a while to load on older devices. We should measure what are the most problematic methods and figure out what can be improved to make it load faster.

pdrj:
## What's this?
This moves binding suggestions in start time entry activity to a background thread. 
### Relationships

[Slack thread](https://toggl.slack.com/archives/CD40QCV25/p1566313600008000?thread_ts=1566301895.001800&cid=CD40QCV25)
Closes #5665 

## Why do we want this?
To improve the activity startup time. In my testing the improvement was in 80-100ms range.

## How is it done?
Moving the most time-consuming binding to a background thread.

### Why not another way?
🤷‍♂ open to suggestions.

## Review hints
Either use a very slow device, or add a `Thread.Sleep` in `bindSuggestions` to see the progressbar.

## Next steps
Show the progress bar when user is waiting for suggestions (as mentioned in the slack thread) - I originally planned to do this in this PR, but [encountered time-consuming problems with Rx](https://toggl.slack.com/archives/C02APMYTM/p1566468794337300) and decided to do it as a part of another issue.

## :squid: Permissions
Anyone.


Toggl.Droid/Activities/StartTimeEntryActivity.cs
```diff
@@ -41,12 +45,24 @@ public StartTimeEntryActivity(IntPtr javaReference, JniHandleOwnership transfer)
...
         }
 
-        protected override void InitializeBindings()
+        private void bindSuggestions()
         {
             ViewModel.Suggestions
                 .Subscribe(adapter.Rx().Items())
                 .DisposedBy(DisposeBag);
 
+            RunOnUiThread(() => isLoadingSubject.OnNext(false));
```
Odrakir: Maybe I'm missing something, so bear with me, but couldn't this be done with an Rx operator when subscribing (right before line 60) `.ObserveOn(...)`

Toggl.Droid/Activities/StartTimeEntryActivity.cs
```diff
@@ -41,12 +45,24 @@ public StartTimeEntryActivity(IntPtr javaReference, JniHandleOwnership transfer)
...
+        }
+
+        protected override void InitializeBindings()
+        {
+            isLoadingSubject
+                .DistinctUntilChanged()
+                .Subscribe(suggestionsProgressBar.Rx().IsVisible(useGone: false))
+                .DisposedBy(DisposeBag);
+
+            ThreadPool.QueueUserWorkItem(_ => bindSuggestions());
```
Odrakir: The same way this could use the `SubscribeOn` operator before line 51.

Toggl.Droid/Activities/StartTimeEntryActivity.cs
pdrj: Yeah, I think you're right. Will change that, thanks.

Toggl.Droid/Activities/StartTimeEntryActivity.cs
pdrj: I can use `SubscribeOn` and that looks better. But how to then make sure that `        isLoadingSubject.OnNext(false);` doesn't get executed immediately on the main thread, before `.Subscribe(adapter.Rx().Items())` happens?

Odrakir: Ok, let me see if I get this. You only need to move `.Subscribe(adapter.Rx().Items())` to a background thread, right?

- First, are you sure you can do that? Does nothing in there happen in the UI?
- And in case you are sure, all you have to do is:

```
protected override void InitializeBindings()
{
    ViewModel.Suggestions
        .ObserveOn(AndroidDependencyContainer.Instance.SchedulerProvider.BackgroundScheduler)
        .Subscribe(adapter.Rx().Items())
        .DisposedBy(DisposeBag);

    isLoadingSubject.StartWith(true)
        .DistinctUntilChanged()
        .Subscribe(suggestionsProgressBar.Rx().IsVisible(useGone: false))
        .DisposedBy(DisposeBag);
```

I don't like it much, because I think the views should work on the main thread all the time to avoid too much trouble, but I think this is equivalent to the code you were writing.

pdrj: Apart from moving it to a background thread, I need to display a progressbar before the binding is complete and then hide it once binding completes. Hence the introduction of, in this commit, `isLoadingSubject` that tracks the binding progress and then the need to update it, after subscription completes. When I used `QueueUserWorkItem`, the binding and the subsequent subject update happened on the same, background thread. If I just do ```ViewModel.Suggestions
        .ObserveOn(AndroidDependencyContainer.Instance.SchedulerProvider.BackgroundScheduler)``` on the main thread and then `isLoadingSubject.OnNext(false)` straight after that, the subject line is executed without waiting for the binding (which was moved to the background thread).

Odrakir: When you say you need to show a progress bar for the binding... you mean the actual binding, right? meaning the declaration and not the event that gets sent through `ViewModel.Suggestions` and to `adapter.Rx().Items()`

I misunderstood, I thought you were trying to track the actual event sending through the observable chain.

In that case I don't understand how that code could take so long to execute that it needs a progress bar 😳. We need to know what's going in there, maybe the `ViewModel.Suggestions` observable is firing some side effects that take some time when subscribed, if that's the case we should be moving some of that stuff to the background in the view model. 

But first we need to find out what exactly is going on.

pdrj: > I misunderstood, I thought you were trying to track the actual event sending through the observable chain.

That will also be done, but later. And for that, we'll use your code from #5804 :).

I agree that with this problem here ie. the actual binding taking long, it'd be better to actually get to the bottom of it rather than moving stuff to different threads in the activity :see_no_evil:. I'll do some more digging.

pdrj:
After giving it more thought, I came up with a cleaner solution. 

For now, I've reverted the commit introducing the progress bar, I'll re-introduce it when we have a way of tracking `Suggestions` observable progress.

I couldn't test it on iOS, as I lack the provisioning profile.

The new changes broke a unit test, which I'll either fix, or I can move the subscribing to a background thread on Android only - any thoughts on what would be best?

pdrj:
Although I don't like it, I decided to only make Android subscribe on a background thread. Motivation:

1. The issue is Android related.
2. We'll want to create 2.3 branch soon and this is part of the project.
3. No time to properly test it in iOS and fix the unit test.

Ideally, sometime after go loco, I'll come back to moving the subscription to the background in the viewmodel and getting to the bottom of that broken unit test (which is the result of doing that).

PR #5804, opened 2019-08-23
Add a way to track Observable execution

Odrakir:
## What's this?
This comes from a problem @pdrj was having the other day with Observables. For now is just a draft, and it's loosely based on this: https://github.com/ReactiveX/RxSwift/blob/master/RxExample/RxExample/Services/ActivityIndicator.swift

It adds a new class and an rx operators to make it easier to track the execution of an Observable.

Still needs some touches and some tests, but I would like to know what you think.

## :squid: Permissions
🚫

### DIFF
--- a/Toggl.Core/Extensions/ObservableExtensions.cs
+++ b/Toggl.Core/Extensions/ObservableExtensions.cs
@@ -1,6 +1,8 @@
 ﻿using System;
 using System.Reactive;
+using System.Reactive.Disposables;
 using System.Reactive.Linq;
+using System.Reactive.Subjects;
 using Toggl.Core.Analytics;
 using Toggl.Core.Sync;
 using Toggl.Shared.Extensions;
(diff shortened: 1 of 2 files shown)


Toggl.Core/Extensions/ObservableExtensions.cs
```diff
@@ -49,5 +51,61 @@ public static IObservable<ITransition> OnErrorReturnResult<T>(this IObservable<I
...
+            return _source;
+        }
+
+        public void Dispose()
+        {
+            _dispose.Dispose();
+        }
+    }
+
+    public class ActivityObserver : IDisposable
```
heytherewill:
```suggestion
    public class ExecutionObserver : IDisposable
```

Activity is a very specific Android keyword and I'd rather avoid confusion

Toggl.Core/Extensions/ObservableExtensions.cs
```diff
@@ -49,5 +51,61 @@ public static IObservable<ITransition> OnErrorReturnResult<T>(this IObservable<I
...
+    {
+        public IObservable<bool> running { get; }
```
heytherewill:
```suggestion
        public IObservable<bool> Running { get; }
```

Toggl.Core/Extensions/ObservableExtensions.cs
```diff
@@ -49,5 +51,61 @@ public static IObservable<ITransition> OnErrorReturnResult<T>(this IObservable<I
...
+            runningSubject.OnNext(false);
+        }
+
+        internal IObservable<T> trackActivityOfObservable<T>(IObservable<T> source)
+        {
+            return Observable.Using(
+                () =>
+                {
+                    runningSubject.OnNext(true);
+                    return new ActivityToken<T>(source, runningSubject);
```
heytherewill: Is this equivalent to this:

```csharp
internal IObservable<T> trackActivityOfObservable<T>(IObservable<T> source)
    => Observable.Using(
        () => 
        {
            runningSubject.OnNext(true);
            return Disposable.Create(() => runningSubject.OnNext(false));
        },
        _ => source);
```

Or does the introduced type provide some other benefit?

Toggl.Core/Extensions/ObservableExtensions.cs
```diff
@@ -49,5 +51,61 @@ public static IObservable<ITransition> OnErrorReturnResult<T>(this IObservable<I
...
+            running = runningSubject.AsObservable();
+            runningSubject.OnNext(true);
+        }
+
+        public void Dispose()
+        {
+            runningSubject.OnNext(false);
+        }
+
+        internal IObservable<T> trackActivityOfObservable<T>(IObservable<T> source)
```
heytherewill:
```suggestion
        internal IObservable<T> TrackActivityOfObservable<T>(IObservable<T> source)
```

heytherewill:
Is this still needed/being worked on?

heytherewill:
pinging @Odrakir for maximum annoyance

Odrakir:
There's no need for this right now, and didn't quite managed to make it work, so I'm closing it.

PR #5814, opened 2019-08-26
Allow users to change dates in the calendar

heytherewill:
## What's this?
This adds the possibility yo change days in the calendar view!!!!!!!!!!1!1!1!!111!!ELEVEN!!!!

## Why do we want this?
Users really want that, and our codebase already supports it, so WHY NOT?!?!?!

## How is it done?
I ressurected a dead fragment to write less code, wired some RxActions and boom! it works
Making it work on iOS shouldn't be hard, BTW, we just need specs

I'm constraining the selected date to be from `now - numberOfMonthsWeFetchData` to `now`

### Side effects
The app is objectively better now

### Review hints

Just tap the current date circle on Android and a picker will appear. Not the most discoverable, I know, but that's just a Go Loco project; we can improve designs later

## :squid: Permissions
Only myself, and only after @maritoggl approves.

⚠️ Tests will be written once Mari approves and we settle implementation details ⚠️

### DIFF
--- a/Toggl.Core.UI/ViewModels/Calendar/CalendarViewModel.cs
+++ b/Toggl.Core.UI/ViewModels/Calendar/CalendarViewModel.cs
@@ -56,7 +56,7 @@ namespace Toggl.Core.UI.ViewModels.Calendar
 
         public IObservable<string> TimeTrackedToday { get; }
 
-        public IObservable<string> CurrentDate { get; }
+        public IObservable<string> FormattedDate { get; }
 
         public UIAction GetStarted { get; }
 
--- a/Toggl.Core.UI/ViewModels/Calendar/CalendarViewModel.cs
+++ b/Toggl.Core.UI/ViewModels/Calendar/CalendarViewModel.cs
@@ -24,6 +24,7 @@ using Toggl.Core.UI.Views;
 using Toggl.Shared;
 using Toggl.Shared.Extensions;
 using Toggl.Storage.Settings;
+using static Toggl.Core.Helper.Constants;
 
 namespace Toggl.Core.UI.ViewModels.Calendar
 {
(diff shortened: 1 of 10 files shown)


maritoggl:
Let's do this - show me what you've got and let me know of any details that you want me to confirm or clarify.

Toggl.Core.UI/ViewModels/Calendar/CalendarViewModel.cs
```diff
@@ -176,6 +171,12 @@ public sealed class CalendarViewModel : ViewModel
                 indexKey: item => item.StartTime,
                 orderingKey: item => item.StartTime,
                 groupingKey: _ => 0);
+
+            timeService.CurrentDateTimeObservable
+                .Select(dateTime => dateTime.Date)
+                .DistinctUntilChanged(date => date.Day)
```
amulware: Isn't `Day` the day of the month, so if I switch from 01-01 to 02-01 it'll ignore the second?

Toggl.Core.UI/ViewModels/Calendar/CalendarViewModel.cs
```diff
@@ -307,6 +309,18 @@ private async Task selectUserCalendars(bool isOnboarding)
             }
         }
 
+        private async Task pickNewCalendarDate()
+        {
+            var now = timeService.CurrentDateTime;
+            var minDate = now.AddMonths(-FetchTimeEntriesForMonths);
```
amulware: This maxmimum should be confirmed carefully. What's the TE prune value? Should we really allow that far back? If we add other sources of data, we won't be able to show them nearly as far back most likely. Just some things to think about.

amulware:
No swiping between days? 😭

Toggl.Core.UI/ViewModels/Calendar/CalendarViewModel.cs
amulware: Then I don't know what this does, but if you say it's fine, feel free to resolve.

amulware:
@sofokli1 Whether to merge this is pretty much up to you. See if it's helpful for your project or not, maybe merge before, or merge into project branch, or whatever makes sense.

sofokli1:
I wouldn't merge this yet. Will wait for designs for the calendar project first and afterword will see if it still fits.

sofokli1:
@heytherewill  we should close this without merge

heytherewill:
Closing as obsolete, but I hope the code can help future calendarians 👐

PR #5818, opened 2019-08-27
🐛 Remove billable switch ripple effect on appbar tap

### LINKED ISSUE #5817: Tapping the description panel in edit view interacts with the billable switch
## 🐛 Bug Report
The title says it all. Also look at repro steps.

## Context
Build/Branch: `develop`
Phone OS: `Android 9`
Phone model: `OP5T`

## Expected behavior
Tapping description field should have nothing to do with the billable switch.

## To reproduce
1. Open edit view for a TE that has a relatively short description.
1. Tap the bottom right corner of the description background panel (just below the `SAVE` button).
1. You should see the ripple effect activating on the billable switch.

## Suggestions
Probably just a layouting issue.

kornelijepetak:
## What's this?
Removes the interaction between the appbar and billable switch.

### Relationships
Closes #5817

## Why do we want this?
So that it doesn't look weird.

## How is it done?
Very simple.

### Why not another way?
Honestly, I don't even understand why that was happening in the first place. The two layouts didn't seem related at all... 

If any of the @toggl/androids have better solutions or even better insight into why it was happening in the first place, please suggest/explain.

### Side effects
Couldn't find any after a while of testing all UI elements on that screen, but please make sure to do a proper functional test of Edit TE view before approving.

## :squid: Permissions
me


kornelijepetak:
@pdrj yeah, honestly I don't even know why `clickable` was set to `true` on the root layout at all. The problem with the clickable switch is in the fact that the background behind the switch is also clickable so, tapping the switch would execute both events, so since geometrically the switch is encompassed in it's background's geometry, handling taps on the background is enough (without resorting to the onTouchEvent-based prevention of event bubbling). 

However, that does prevent a ripple effect on the switch itself. So, let's invoke @maritoggl to confirm whether it's ok to have only the ripple on the row itself.

PR #5832, opened 2019-08-29
Configure build/performance settings

heytherewill:
## What's this?
This properly configures our .csprojs for faster deploy times and to use Bundles

## Why do we want this?
Faster and smaller apps

## How is it done?

- Aligned `Release` with  Release.PlayStore` for easier local testing
- Enabled aapt2 for faster builds
- Enabled R8 and D8
- Enabled [Startup Tracing](https://devblogs.microsoft.com/xamarin/faster-startup-times-with-startup-tracing-on-android/)
- Using AppBundles instead of APKS

### Side effects
Faster app builds, but we also gotta test the new aab. Once we test it and it works we can shift the focus to fixing our CI builds to use the new format.

## Review hints
Carefully check configurations and point out possible mistakes

## :squid: Permissions
Only myself


daividssilverio:
It looks great to me, but in the little amount of testing I just did on an emulator I got a little bit of jank in the main log. 🤔 I'll run more tests in multiple actual devices tomorrow.

heytherewill:
On hold until we figure out when the move to .aab will happen

semanticer:
It looks like bundles are not to blame after all https://github.com/toggl/mobileapp/issues/6873

PR #5834, opened 2019-08-29
Fix a bug where the reports view wouldn't show the placeholder

dshr:
## What's this?
A little fix to a presentation problem in Reports introduced in #5732.

### Relationships
Closes nothing

## Why do we want this?
So things work properly

## How is it done?
So, the way our Reports view is set up is that all auxiliary views, apart from the chart segment list, are all in a table view section header. The problem that we got is that in the refactor a TV without cells would not render any headers, which is expected behaviour under normal circumstances. But this case is not — we need the header regardless. So I added a simple override to the `ReportsTableViewSource` that creates a single empty section when the underlying collection is empty.

### Why not another way?
This seemed like the simplest one to me.

### Side effects
None.

## Review hints
Check develop on a fresh account that has no TEs — that's the only way to trigger the bug.

## :squid: Permissions
All yours!


heytherewill:
Please wait for some @toggl/ios-senseis to approve, though, since I didn't functional test

dshr:
> Please wait for some @toggl/ios-senseis to approve, though, since I didn't functional test
Yup, that's why I've requested a separate review from them ;)

PR #5836, opened 2019-08-30
Replace '\n' with actual new lines

### LINKED ISSUE #5835: Overlapping Terms & Conditions text
## 🐛 Bug Report
The popup to accept terms and conditions shown to users creating account has some overlapping text, as in the attached screenshot.

## Context

Build/Branch: development
Phone OS: Android 6
Phone model: XT1092

## Expected behavior
Text displayed properly, without overlapping. 

## To reproduce
Try to sign up for a new account.

## Suggestions
Might be related to the i18n changes.

device-2019-08-29-163444

daividssilverio:
## What's this?
This fixes a bug introduced by the i18n project.
We've formatted the `Resources.resx` file to look well indented, but it turns out 
```
<data name="PrivateProjectHint" xml:space="preserve">
        <value>Private projects are only accessible
to project members</value>
    </data>
```
,
```
<data name="PrivateProjectHint" xml:space="preserve">
        <value>Private projects are only accessible\n to project members</value>
    </data>
```

and
```
<data name="PrivateProjectHint" xml:space="preserve">
        <value>Private projects are only accessible\n
                      to project members</value>
    </data>
```

are all different things, thanks to the `xml:space="preserve"` attribute, which does what it says it does, it preserves spaces.

### Relationships
Closes #5835

## Why do we want this?
Because we don't want to see actual `\n` characters in our text entries and we still need those new lines (debatable)

## How is it done?
It replaces every usage of `\n` with an actual new line.
It removes the extra spaces that were added to make the `Resources.resx` file look good.

### Why not another way?
I don't know if there's another way, maybe @kornelijepetak, @heytherewill or @amulware could tell.

### Side effects
If in any of the places the actual extra spaces were intentional, they have been removed and the results should be different.

## Review hints
Please test this on an actual device.

## :squid: Permissions
🦑 it anyone.


Toggl.Shared/Resources.resx
```diff
@@ -538,7 +538,7 @@
     </data>
     <data name="TermsOfServiceDialogMessageDroid" xml:space="preserve">
         <value>By tapping "I Agree", you
-            understand and agree to Toggl\'s</value>
+understand and agree to Toggl’s</value>
```
kornelijepetak: This change, however, is not equivalent. The characters are different. In some way, I don't care, but in digital world (as in not a printed text), a `'` character is used as an apostrophe. Of course, the way it looks is a bit different depending on the font, but still.

This change does not really have anything to do with whitespace. Why is it escaped in the first place? As far as I know, `'` is **not** a special character in resx file and does not need escaping.

daividssilverio: I just took the chance the make it the same as the iOS equivalent string.
Indeed, using `'` should work just fine.

PR #5837, opened 2019-08-30
Implement Handoff to web app

arturs105:
## What's this?
This adds Handoff support to the following screens:
* Timer page
* Settings page
* Reports page (with support for workspace and selected period changes)

## How is it done?
With `NSUserAction`

### Side effects
Should be zero to none

## Review hints
Handoff doesn't work on simulators, so you'll need to test on an actual device.

## :squid: Permissions
Only me or the release manager for iOS 2.4


dshr:
Code's also on point 👌

Odrakir:
@heytherewill What frontend bug? I can't test this right now.

PR #5843, opened 2019-09-05
Fix a bug where the reports calendar would start at the wrong page

dshr:
## What's this?
A release branch fix for the little bug that looked as if the reports calendar selection was not working right. Turns out it was stuck on the wrong page, in this case September 2017.

### Relationships
Closes nothing.

## Why do we want this?
No bugs.

## How is it done?
By using `.AsDriver`.

### Why not another way?
Seems simple enough.

### Side effects
None.

## Review hints
Check before and after: start app, go to reports, open calendar and check the date.

## :squid: Permissions
Only me


dshr:
Changed the fix due to some bugs in the Android implementation of the view which caused `AsDriver` to start a feedback loop. Now moving the observer to the main thread is done in the ViewController code.

PR #5881, opened 2019-09-11
Add common selector view for iOS

### LINKED ISSUE #5714: Settings page is cut off
## 🐛 Bug Report
The settings page is cut off at the sides for this [user](https://app.intercom.io/a/apps/ayixs927/inbox/inbox/3155115/conversations/23105805184):
image

Build/Branch:
Phone OS: iOS 12.3.1
Phone model: iPhone 7

## Expected behavior
The page should not be cut off

## To reproduce
Could not reproduce with a long workspace name on iPhone SE so probably not that 🤷‍♀ 

## Suggestions
Nada 👶

### LINKED ISSUE #5803: Calendar settings go over the sides of the screen
## 🐛 Bug Report
This only happens in iOS 13 beta, apparently, but I don't think it'll work in the final version. When accessing the calendar settings from the app settings (via the main view) that table view goes over the sides of the screen and it's unusable.

I'm attaching a screenshot.

image

## Context
Build/Branch: This happens in the App Store version (can't test on develop because VS doesn't want to)
Phone OS: iOS 13 (beta 8)
Phone model: iPhone Xs

## To reproduce
Open the app, got to the main timeline -> settings -> Calendar settings
(for the calendar settings to show you have to enabled the calendar integration first)

## Suggestions
This is probably related to how we layout those table views for them to show properly in the iPad.

### LINKED ISSUE #5848: Refactor selector table views
This should be done on top of Will's go-loco project that simplifies Select ViewModels #5819.

That PR results in settings selectors for iOS being action sheets, which I think is not the best component for that.

My idea is to make the main settings screen modal and those selector a pushed view controller with a table view that gets its data from the view model in a common manner using the code in #5819

Odrakir:
## What's this?
This builds over #5819 to replace selectors by new view controllers with table views in iOS. 

While on it, it also changes the way settings are shown in iOS, now they show as a page sheet, which will also fix layout problems due to some constraints we needed for the iPad layout.

### Relationships
Closes #5803 
Closes #5848 
Closes #5714 (I'm not 100% sure about this one, but I haven't been able to reproduce it) 

## Why do we want this?
We want:
- A common way to present selectors + a UI for it that follows design guidelines
- A layout that doesn't break depending on the device
- Settings being a modal view. This is me, personally, but it'll also make them look cool on iOS 13 (that is, when we update our build stack)

## How is it done?
Given that some of these views require an initial selection index (default workspace) and some don't (calendar event long-press). I made that parameter optional and then the platform specific view chooses the appropriate interface depending on that.

If there's no initial index or the current vc is not in a navigation stack -> Action sheet
Otherwise -> Push a table view selector

The rest is just a simple table view.

By the way, this is how it looks in iPad in landscape.

Simulator Screen Shot - iPad Pro (11-inch) - 2019-09-11 at 17 51 37

and iPhone:
iphone

## :squid: Permissions
Go!


arturs105:
From a functional test point looks good, apart from 2 visual things:
* The background of the selector view is white. I think it would make more sense for it to be the same background color from settings view (at least for the selector views that are accessible from settings)
* Having no separator below the navigation bar looks a bit weird, i think there should be one

Both of these are design issues, so in the end it's probably a designers decision.

Toggl.iOS/ViewControllers/Common/SelectorViewController.cs
```diff
@@ -0,0 +1,70 @@
...
+
+            var selected = options[selectedIndex];
+            onChosen(selected.Item);
+        }
+
+        private void onItemSelected(int selectedIndex)
+        {
+            this.selectedIndex = selectedIndex;
+
+            if (NavigationController != null)
```
dshr: Can we drop the `{}`s?
```suggestion:0+8
            if (NavigationController != null)
                NavigationController.PopViewController(true);
            else
                this.Dismiss();
```

Toggl.Droid/Activities/ReactiveActivity.DialogProvider.cs
```diff
@@ -12,8 +12,8 @@ public abstract partial class ReactiveActivity<TViewModel>
         public IObservable<bool> Confirm(string title, string message, string confirmButtonText, string dismissButtonText)
             => this.ShowConfirmationDialog(title, message, confirmButtonText, dismissButtonText);
 
-        public IObservable<T> Select<T>(string title, IEnumerable<SelectOption<T>> options, int initialSelectionIndex = 0)
-            => this.ShowSelectionDialog(title, options, initialSelectionIndex);
+        public IObservable<T> Select<T>(string title, IEnumerable<SelectOption<T>> options, int? initialSelectionIndex = null)
+            => this.ShowSelectionDialog(title, options, initialSelectionIndex.HasValue ? initialSelectionIndex.Value : 0);
```
arturs105: Won't this mark the first option as selected if `null` is passed?

Toggl.Droid/Fragments/ReactiveTabFragment.DialogProvider.cs
```diff
@@ -18,8 +18,8 @@ public abstract partial class ReactiveTabFragment<TViewModel>
             return Activity.ShowConfirmationDialog(title, message, confirmButtonText, dismissButtonText);
         }
 
-        public IObservable<T> Select<T>(string title, IEnumerable<SelectOption<T>> options, int initialSelectionIndex = 0)
-            => Activity.ShowSelectionDialog(title, options, initialSelectionIndex);
+        public IObservable<T> Select<T>(string title, IEnumerable<SelectOption<T>> options, int? initialSelectionIndex = null)
+            => Activity.ShowSelectionDialog(title, options, initialSelectionIndex.HasValue ? initialSelectionIndex.Value : 0);
```
arturs105: Same question as above.

Toggl.iOS/Cells/Common/SelectorCell.cs
```diff
@@ -0,0 +1,40 @@
+﻿using System;
+
```
arturs105: Empty line

Toggl.iOS/Extensions/ViewControllerExtensions.cs
```diff
@@ -115,7 +116,19 @@ internal static IObservable<bool> ShowConfirmDestructiveActionDialog(this UIView
             });
         }
 
-        internal static IObservable<T> ShowSelectDialog<T>(this UIViewController viewController, string title, IEnumerable<SelectOption<T>> options, int initialSelectionIndex)
+        internal static IObservable<T> ShowSelectDialog<T>(this UIViewController viewController, string title,
+            IEnumerable<SelectOption<T>> options, int? initialSelectionIndex = null)
```
arturs105: The way that the arguments are split irritates me, though there's nothing in the style guide that forbids this (yet), so 🤷‍♂

Toggl.iOS/ViewSources/Generic/Common/SelectorTableViewSource.cs
```diff
@@ -0,0 +1,40 @@
...
+using Toggl.Core.UI.Views;
+using Toggl.iOS.Cells.Common;
+using UIKit;
+
+namespace Toggl.iOS.ViewSources.Common
+{
+    public sealed class SelectorTableViewSource<T> : BaseTableViewSource<SectionModel<Unit, SelectOption<T>>, Unit, SelectOption<T>>
+    {
+        private readonly int initialSelection;
+        private Action<int> selectItem;
```
arturs105: This can be made `readonly`.

Toggl.Core.UI/ViewModels/Calendar/CalendarViewModel.cs
```diff
@@ -342,7 +342,7 @@ private async Task handleCalendarEventLongPressed(CalendarItem calendarItem)
                 options.Add(option);
             }
 
-            var selectedOption = await View.Select(Resources.CalendarWhatToDoWithCalendarEvent, options, initialSelectionIndex: 0);
+            var selectedOption = await View.Select(Resources.CalendarWhatToDoWithCalendarEvent, options);
```
heytherewill: Why make it optional? I don't get it

Toggl.iOS/Presentation/PageSheetPresenter.cs
```diff
@@ -0,0 +1,30 @@
...
+            typeof(SettingsViewModel),
+        };
+
+        public PageSheetPresenter(UIWindow window, AppDelegate appDelegate) : base(window, appDelegate)
+        {
+        }
+
+        protected override void PresentOnMainThread<TInput, TOutput>(ViewModel<TInput, TOutput> viewModel, IView sourceView)
+        {
+            UIViewController viewController = ViewControllerLocator.GetNavigationViewController(viewModel);
```
heytherewill: var

Toggl.iOS/Presentation/PageSheetPresenter.cs
```diff
@@ -0,0 +1,30 @@
...
+
+            viewController.ModalPresentationStyle = UIModalPresentationStyle.PageSheet;
+
+            UIViewController topmostViewController = FindPresentedViewController();
```
heytherewill: var

heytherewill:
I just saw that you changed this PR quite a bit and that's why you did the `optional` thing. I don't like it because it will create problems in Android land; the select will return the 0th index instead of the currently selected value. Unless you make the return type of the `Select` and `Option<T>`, which will make it simple to check whether the method was canceled or not.

Toggl.Core.UI/ViewModels/Calendar/CalendarViewModel.cs
Odrakir: At some places were using this component there's no concept of a previously selected option. Like when long pressing a calendar event on iOS. You get an action sheet with a set of actions, but none of those is pre-selected. 

I'm not sure if something like that happens on Android as well.

What I do is, when that is null, we show an action sheet, when it is not, we show the table with the preselected value.

I also use it when switching workspaces from the reports screen, because we are showing an action sheet there as well for now and presenting one as preselected is not so important.

Toggl.iOS/Presentation/PageSheetPresenter.cs
Odrakir: Ok, but why is that important?

Toggl.iOS/ViewControllers/Common/SelectorViewController.cs
Odrakir: Mmmm... No? 😄

Seriously, we should write guidelines on this. I do like my `{}` even for single statements, it prevents mistakes when adding more lines.

Odrakir:
I addressed everything. I think colors look better now, but in any case, this would need a review when we merge Dark Mode.

Toggl.iOS/ViewControllers/Common/SelectorViewController.cs
kornelijepetak: When `if` has an `else`, I am also in favor of putting stuff into the `{}`. Although I think that mistake prevention is a non-issue. Everybody uses that argument, but we simply don't change code that caressly. Besides, the IDE will format your code in such a way that you will see the incorrect indentation anyway. 🤷‍♂

Toggl.Droid/Activities/ReactiveActivity.DialogProvider.cs
```diff
@@ -12,9 +12,12 @@ public abstract partial class ReactiveActivity<TViewModel>
         public IObservable<bool> Confirm(string title, string message, string confirmButtonText, string dismissButtonText)
             => this.ShowConfirmationDialog(title, message, confirmButtonText, dismissButtonText);
 
-        public IObservable<T> Select<T>(string title, IEnumerable<SelectOption<T>> options, int initialSelectionIndex = 0)
+        public IObservable<T> Select<T>(string title, IEnumerable<SelectOption<T>> options, int initialSelectionIndex)
             => this.ShowSelectionDialog(title, options, initialSelectionIndex);
 
+        public IObservable<T> SelectAction<T>(string title, IEnumerable<SelectOption<T>> options)
+            => Observable.Throw<T>(new InvalidOperationException("This is not implemented for Android"));
```
heytherewill:
```suggestion
            => this.ShowSelectionDialog(title, options, initialSelectionIndex, 0);
```

Toggl.Droid/Fragments/ReactiveDialogFragment.cs
```diff
@@ -125,6 +125,9 @@ public IObservable<bool> ConfirmDestructiveAction(ActionType type, params object
 
         public IObservable<T> Select<T>(string title, IEnumerable<SelectOption<T>> options, int initialSelectionIndex)
             => throw new InvalidOperationException("You shouldn't be doing this from the Dialog. Use the parent activity/fragment");
+        
+        public IObservable<T> SelectAction<T>(string title, IEnumerable<SelectOption<T>> options)
+            => throw new InvalidOperationException("This is not implemented for Android");
```
heytherewill:
```suggestion
            => throw new InvalidOperationException("You shouldn't be doing this from the Dialog. Use the parent activity/fragment");
```

Toggl.Droid/Fragments/ReactiveTabFragment.DialogProvider.cs
```diff
@@ -18,9 +18,12 @@ public abstract partial class ReactiveTabFragment<TViewModel>
             return Activity.ShowConfirmationDialog(title, message, confirmButtonText, dismissButtonText);
         }
 
-        public IObservable<T> Select<T>(string title, IEnumerable<SelectOption<T>> options, int initialSelectionIndex = 0)
+        public IObservable<T> Select<T>(string title, IEnumerable<SelectOption<T>> options, int initialSelectionIndex)
             => Activity.ShowSelectionDialog(title, options, initialSelectionIndex);
 
+        public IObservable<T> SelectAction<T>(string title, IEnumerable<SelectOption<T>> options)
+            => Observable.Throw<T>(new InvalidOperationException("This is not implemented for Android"));
```
heytherewill:
```suggestion
            => this.ShowSelectionDialog(title, options, initialSelectionIndex, 0);
```

Toggl.Droid/Activities/ReactiveActivity.DialogProvider.cs
Odrakir: Are you sure about this? This is supposed to show a view in which the idea of an initial selection makes no sense. Wouldn't this cause confusion if you ever need that view?

Toggl.iOS/Presentation/PageSheetPresenter.cs
arturs105: because it's less code 😄 We'd loose 13 characters of code from this change 😄

Toggl.iOS/ViewSources/Generic/Common/SelectorTableViewSource.cs
```diff
@@ -0,0 +1,54 @@
...
+            base.RowSelected(tableView, indexPath);
+            selectItem(indexPath.Row);
+        }
+        
+        public override nfloat GetHeightForHeader(UITableView tableView, nint section)
+        {
+            return 1;
+        }
+
+        public override UIView GetViewForHeader(UITableView tableView, nint section)
```
arturs105: I think you should do this for the footer as well, because the last separator line not being full width looks weird.

Odrakir: I took a different approach, because if I did that I would end up with 2 separators.

Toggl.iOS/ViewControllers/Common/SelectorViewController.cs
Odrakir: There are two things I don't trust:
- Developers care (even my own, we all have those days)
- IDEs (mostly since I do Xamarin)

;)

Issue #5878, opened 2019-09-10
Polish and ship dark mode


heytherewill:
We need to figure out what still needs to be done for Dark Mode (#5811) and do it so we can include it into the next release!

semanticer:
@heytherewill We still need to fix how we communicate elevation in dark mode. You can read about google's recommendations here https://material.io/design/color/dark-theme.html#properties and more about the implementation in section Elevation Overlays here https://material.io/develop/android/theming/dark/ but I don't know if we can take advantage of that surface magic without switching to material design library. 

Having said that, the only place where I see this as a problem is the bottom tab bar and current running TE view. They both blend into the rest of the content and that's not ideal.

AlicjaSuska:
Hi! In Zeplin / design files there are no dark mode designs ready to link. If some design work is needed it has to be planned as a project for design team. @maritoggl that would be great if you could see this thread and please, let me know if you already have schedule in mind for this project:) Thank you!

PR #5882, opened 2019-09-11
📐 🚧 Unify toolbars across the app

### LINKED ISSUE #5698: Unify toolbars across the app

pdrj:
## What's this?
This makes all of our toolbars look the same (and be actual `Toolbars`), with the exception of the one from calendar activity (it's very customized, not worth the effort to make it a `Toolbar`).

### Relationships

Closes #5698

## Why do we want this?
This makes all of the toolbars look the same and improves UX. Especially it makes sure that the action buttons are always in the same position and have the same styling.

## How is it done?
Lots of layout changes :).

### Why not another way?
🤷‍♂ 

### Side effects
In some cases, I had to let go of our white-themed activities. These white activities caused a different color of the status bar and the toolbar, which looks pretty bad. The default background color is also close to white, hence I think this change should be OK.

## Review hints
I've tested with phones running Android 5, 6, 8. It'd be nice if someone took a look on a tablet and on different phones too ;).

## :squid: Permissions
Me


heytherewill:
A lot of the changes here conflict with the Dark mode PR, which is also being worked on for this release but has, arguably, a higher priority because of the release of Android Q. I suggest putting this `on-hold` until we merge Dark Mode, but this is ultimately up to the release manager

semanticer:
I agree with @heytherewill , let's give our Pixel users(like me) some extra treatment and make sure they have a black celebration as soon as possible :-)

semanticer:
Ok I've read the rest of the conversation in related [slack thread](https://toggl.slack.com/archives/C02APMYTM/p1568113730114600) and given that we are still waiting for designers to look at it and haven't decided if we want to do anything about the theme immediate updates etc I think we shouldn't hold this nice refactor back if we don't know for sure dark mode will be even ready for 2.4.  

I'm removing `on-hold` and in the worst case, a few Pixel users will miss dark mode for one more release. Honestly, as much as I want it for myself, I think dark mode on iOS will be a much bigger deal because of the quick adoption.

kornelijepetak:
In the end, I think this is a decision to be made by designer/product... The PR itself seems fine in any case.

Toggl.Droid/Resources/drawable/black_close.xml
```diff
@@ -0,0 +1,12 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="24dp"
+    android:height="24dp"
+    android:viewportWidth="24"
+    android:viewportHeight="24">
+  <path
```
semanticer: The formating seems to be a bit off here

pdrj: Yeah there seems to be too much space - will fix it, thanks :).

semanticer:
So, know when 🇵🇱 and ship dark mode (#5878)  has a green light, what should we merge first @heytherewill @pdrj

pdrj:
I think users will enjoy the Dark Side more than this. Maybe, after merging dark mode, we'll be able to squeeze this one in as well for 2.4? From today's meeting it seems we don't have to wait for a lot of design feedback, so I hope dark mode will be merged quite soon.

Toggl.Droid/Resources/drawable/black_close.xml
pdrj: Fixed.

pdrj:
Since this has a lot of changes and now also works in dark mode, it'd be nice if someone additional from @toggl/androids could test it as well.

Toggl.Droid/Resources/layout/EditTimeEntryActivity.axml
```diff
@@ -16,34 +16,28 @@
...
-                android:tint="@color/primaryText" />
+            android:minHeight="@dimen/expandedToolbarHeight"
+            android:elevation="@dimen/toolbarElevation"
+            android:paddingBottom="16dp"
+            android:background="@color/background"
+            >
+            <android.support.v7.widget.Toolbar
+                android:id="@+id/Toolbar"
+                style="@style/ToolbarStyle"
+                android:layout_height="match_parent" />
```
heytherewill: Should height really match parent?

daividssilverio: it looks like a leftover only, the style `ToolbarStyle` is setting everything.

Toggl.Droid/Activities/EditTimeEntryActivity.cs
```diff
@@ -25,6 +26,8 @@ namespace Toggl.Droid.Activities
               ConfigurationChanges = ConfigChanges.Orientation | ConfigChanges.ScreenSize)]
     public sealed partial class EditTimeEntryActivity : ReactiveActivity<EditTimeEntryViewModel>
     {
+
```
daividssilverio: 🔥 extra line

daividssilverio:
Oh, using `app:titleMarginStart="0dp"` also helps o/
I guess it's just a matter of tweaking these attributes and testing on different devices o/

Toggl.Droid/Resources/layout/EditTimeEntryActivity.axml
pdrj: This is actually on purpose. Otherwise, the edit text has a different background and it looks especially bad when it's expanded to multiple lines.
Screenshot 2019-09-20 at 09 35 03

I can't put the `EditText` inside the `Toolbar` because it'll get a left margin then, which can't be removed with `app:contentInsets` of any kind ;).

Toggl.Droid/Activities/EditTimeEntryActivity.cs
pdrj: Done.

pdrj:
> Oh, using `app:titleMarginStart="0dp"` also helps o/

This didn't work for me, but I used a content inset of `2dp`. Tested on a big/small phone and a tablet (emulator) - looks alright. Please let me know how you like it now @daividssilverio.

BTW, I think the default toolbar follows [the specs](https://material.io/components/app-bars-top/#specs), which specify quite a big margin. What also doesn't help is that the button itself takes up a lot of space and it makes it seem like the margin is even bigger:
Screenshot 2019-09-20 at 09 46 54

That being said, I agree with you that the margin looks a bit too big :).

Toggl.Droid/Resources/layout/ReportsFragment.axml
```diff
@@ -20,12 +20,7 @@
...
         <android.support.v7.widget.Toolbar
             android:id="@+id/Toolbar"
-            android:minHeight="?attr/actionBarSize"
-            android:background="@color/background"
-            app:popupTheme="@style/ThemeOverlay.AppCompat.Dark"
-            android:theme="@style/ThemeOverlay.AppCompat.Dark.ActionBar"
-            android:layout_width="match_parent"
-            android:layout_height="?attr/actionBarSize">
+            style="@style/ToolbarStyle">
             <TextView
```
daividssilverio: Also, the reports' title could've been left aligned instead of centred, right? But I guess this is alright.

image
image

pdrj:
Last commit 'approved' in https://toggl.slack.com/archives/C02APMYTM/p1568987178332400

PR #5883, opened 2019-09-11
Proper project visibility

### LINKED ISSUE #5324: Only administrators should be able to create public projects
Currently, our mobile apps allow all workspace users to create public projects. This is not expected behavior. Only administrators should be allowed to create public projects, all projects should be private by default. 

For that reason, we need the following changes:
* when creating new projects, public/private switch should be visible only to workspace administrators
* when creating new projects, projects should be by default private

Related to https://github.com/toggl/discussions/issues/53

heytherewill:
## What's this?
This prevents Ricardo from polluting the Toggl Workspace with projects made public by accident

### Relationships
Closes #5324 

## Why do we want this?
For consistency across platforms

## How is it done?
- The relay that was responsible for the `IsPrivate` field is now set to true at the beginning
- There's a property that makes the whole row disappear if the user is not an Admin
  - If they are not an admin, we also make the project private by default, regardless of the previous choice.
- Turns out we were completely ignoring the `IsPrivate` field of the DTO, so now we are actually using it

### Why not another way?
This makes sense

## Review hints
Test the app functionally on both platforms and ensure that the field gets hidden properly when you change the project's WS to one you are not an admin of. 
Also check the project on webapp's project management screen to see that the project is indeed being created as a private project.

## :squid: Permissions
Only myself


kornelijepetak:
Approving after creating https://github.com/toggl/mobileapp/issues/5901

Toggl.iOS/ViewControllers/EditProjectViewController.cs
```diff
@@ -93,15 +93,11 @@ public override void ViewDidLoad()
                 .DisposedBy(DisposeBag);
 
             // Is Private
-            PrivateProjectSwitchContainer.Rx().Tap()
-                .Select(_ => !PrivateProjectSwitch.On)
+            PrivateProjectSwitch.Rx().Changed()
```
Odrakir: I'll leave this to you, but... How about being the good boy scout and changing that `Changed()` extension to return `Observable<bool>` instead of `Observable<Unit>`? That way you could delete the next line and it'll only require changes in a couple of places (maybe you could even avoid that by creating an alternative extension `ValueChanged` or something)

arturs105: I had exactly the same though, but as i looked into the code, i noticed that the `Changed` extension method could be made more generic instead: #5910.

Odrakir: 👌

Issue #4986, opened 2019-04-25
Consider using dedicated views for Create Project suggestion


kornelijepetak:
When typing into a filter in project selection, the app will prepend a `Create Project XYZ` item to the beginning of the list. In order for this item to be visible, the list has to be scrolled to the beginning, because in the current implementation, it's just a regular row in the tableview/recycler. In order to do that, the scroll state has to be reset to top.

If we extract this row and use a dedicated view (with toggling its visibility), we could have the scroll independent of the creation process, leading to better UX.

This is what @maritoggl suggested as a desired behavior.

@amulware If you agree with her suggestion, make sure to schedule this issue somewhere.

amulware:
Shouldn't this have platform labels? Doesn't seem like an actual refactor to me.

Also, I'm fine with this, assuming there are proper designs.

maritoggl:
Hmm. I don't think this needs new designs, perhaps using the term "view" brought about some confusion. I have not proposed using entirely new visuals, but an improvement to how the list that we already have is being scrolled up and down - basically removing automatic scrolling, but still displaying the create new project row at the beginning of the list.

kornelijepetak:
@amulware yes, as @maritoggl said, this is supposed only to stop adding "Create new" label as a row in a table and instead show it (when needed) as a separate view so that it is indepenedent of scrolling.

amulware:
Have we considered how this will look on small screens? Will there be enough space for anything to actually scroll if more space is taken up by this additional view?

kornelijepetak:
Well, if there's a user on whose phone only one row can be visible, trust me, he will not be able to use other parts of the app anyway...

This may be a radical comment, but I don't think it's gonna be a problem. But that is a product question, @maritoggl should decide.

maritoggl:
I approve.

amulware:
Before I schedule this.. @maritoggl why do we actually need this? Being scrolled to the top is the default already... It seems inelegant to me to put this view always at the top.

kornelijepetak:
@amulware for reasons described here https://github.com/toggl/mobileapp/issues/4822#issuecomment-486990846, it is not being scrolled to the top by default. 

I've added that behavior (hard scroll to top) at some point in a fix.

However, I'm revisiting this issue, because my _fix_ introduced another issue:
When tasks are expanded, the list is reset to top even thought the tasks and their project are somewhere down the list.

kornelijepetak:
[...]

PR #5893, opened 2019-09-13
🐛 Prevent RecyclerView scrolling to top when expanding tasks

### LINKED ISSUE #5779: Select project view recycler view is updating more items than it should when opening tasks
## 🐛 Bug Report
When you go to the Select project view (from edit view) and expand tasks, the recycler view updates like the adapter is adding more items than the expansion is causing.

## Context
Build/Branch: Develop at time of creation of this issue; 2.2.1
Phone OS: Android
Phone model: Samsung Note 9

## Expected behavior
Only related items should be updating when expanding tasks.

## To reproduce
1. Make sure you have a lot of projects/workspaces
2. Go to Edit Time Entry View
3. Go to the Select Project View
4. Scroll down a bit
5. Select a project with tasks to expand it
6. There you go.

## Suggestions
It looks like the recycler view's adapter update logic in the select project view is flawed.

pdrj:
## What's this?
This makes the RecyclerView holding project suggestions not scroll to the top when project's tasks are expanded.

### Relationships

Closes #5779

## Why do we want this?
Otherwise the recycler scrolls to the top and blinks in a weird way, as in the gif attached to the related issue. It's bad UX.

## How is it done?
Dirtyly, with side effects that track if tasks were toggled.

### Why not another way?
If rx-experts can suggest something nicer, I'll be happy to modify this. That being said, the whole scrolling to the top business introduced in #4979 is not ideal either, so maybe we should again think about #4986 or make the scroll to the top execute only for item updates that introduce the `create` suggestion to the recycler.

## Review hints
Follow the steps from the issue and see there's no longer a blink at the top of the recycler.

## :squid: Permissions
Me


kornelijepetak:
Oh, wait, that will not work. It would require another reset of the flag.
Well, never mind my previous comment.
I think the PR is fine.

Toggl.Droid/Activities/SelectProjectActivity.cs
```diff
@@ -65,8 +68,13 @@ protected override void InitializeBindings()
 
             void scrollToTop()
             {
-                recyclerView.GetLayoutManager().ScrollToPosition(0);
+                if (!hasToggledTasks)
```
semanticer: Maybe it would be nicer to go back to the previous simple implementation of scrollToTop so it actually does what it says and add `.Where()` to the stream instead. (`Line 46`)

```
                .Where(_ =>
                {
                    if (!hasToggledTasks) return true;
                    hasToggledTasks = false;
                    return false;
                })
```

This lambda could be extracted as a method to further improve readability. 
It does express the intentions a bit more clearly imo...

semanticer:
Ou, now I see @kornelijepetak was first with a similar suggestion 👍

Toggl.Droid/Activities/SelectProjectActivity.cs
pdrj: Honestly, AFAIC, using `Where` or any operator that takes arguments, then discarding them and introducing side effect inside that operator, is a bit of a smell. IMO `Where` should be used to filter and that's it ;). `Do` on the other hand, is supposed to be used for side effects, so I think in this case the usage is warranted. Please let me know if you agree.

I agree with your point about the scroll method doing actually what the name suggests. How about I rename it to `scrollToTopIfHasntToggledTasks` or maybe something shorter (ideas welcome ;))?

semanticer: Yeah, you are probably right. The side effect in `.Where` is no improvement over the existing solution. My original idea was the same as @kornelijepetak suggested and then I realized that you also have to reset the flag.  

I'm not great at naming, can't think of anything less horrible :)

Toggl.Droid/Activities/SelectProjectActivity.cs
```diff
@@ -48,6 +50,7 @@ protected override void InitializeBindings()
                 .DisposedBy(DisposeBag);
 
             adapter.ToggleTasks
+                .Do(() => hasToggledTasks = true)
```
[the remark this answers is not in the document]
pdrj: Yep, I agree it's not ideal.

PR #5908, opened 2019-09-16
Allow app to fetch archived projects

### LINKED ISSUE #5595: Pull recently archived projects
With https://github.com/toggl/backend/issues/6804 implemented, we should be able to include an `?include_archived=true` parameter when pulling time entries with '?since`.

Ideally, this should automatically archive the projects locally and lead to all the desired UI changes, but do make sure to test this.

I don't think we want to include the new parameter for fetching all projects on login, since that might pull way too many projects.

However, this shows a bit of an edge case if users left the app closed for 2+ months, which will activate pulling data without ?since, and miss changes. This problem already exists right now though and we shouldn't do anything about that here.

To complete this issue, we should:
- [x] add the parameter to the api (optionally defaulting to false/ommitted)
- [x] add integration tests (for both false/true)
- [x] use the parameter when fetching projects with ?since in sync
- [x] add at least one sync test verifying that archived projects will be correctly pulled
- [x] manual testing of all this of course

If you think the issue would be better split up into multiple ones, feel free to do so.

heytherewill:
## What's this?
This adds the ability to fetch archived projects, making our UI better and the app more resilient to changes made in other clients

### Relationships
Closes #5595 

## Why do we want this?
We already have the UI foundation to show that projects are archived, but the API wouldn't return those projects. Now it does, so we might as well

## How is it done?
I've simply added an option to fetch archived projects and made it the default behavior for the sync algorithm

### Why not another way?
I honestly don't see why even keep the option to not fetch archived projects in the `?since` endpoint. IMO we should make that the default behavior, specially so because the method to `FetchSince` comes from an interface that we don't wanna touch to add an extra parameter. I'm pinging @amulware here since he probably knows better, but I think we should just remove the `?fetch_archived=false` behavior from the app, since it's not used anywhere

### Side effects
None that I'm aware of.

## Review hints
Create a bunch of projects, create TEs with them and sync. Then, on web, archive the projects and sync again. The nice `Project (archived)` UI should be shown. in the TEs with archived projects. Also, suggestions from archived projects should disappear

## Tests
[Integration tests](https://app.bitrise.io/build/8d1bddc60a9adcef#?tab=log)
[Syncing tests](https://app.bitrise.io/build/b358ac9379626fe6#?tab=log)

## :squid: Permissions
Only myself, and only after explicit approval from one @toggl/sync-guardians and functional tests of both @toggl/androids and @toggl/ios-senseis


amulware:
Does the endpoint allow using the new parameter WITHOUT using ?since? If so, we definitely shouldn't do that, since it will pull TONS of data.

amulware:
I see from our code that we don't do that in either case. So that's good. And yeah, I don't see a reason to ever use ?since=*?withArchived=false either - if it simplifies the code, remove it, but if it doesn't, then don't bother?

amulware:
Note to release managers: this improvement is a big deal. Please include it in the changelog, and notify support especially, since they in the past would recommend users to logout/in to get rid of archived projects.

sofokli1:
@heytherewill sync test has failed, ping me when fixed

dshr:
Same as @sofokli1 — code looks good and all works fine, but the sync tests are failing :(

heytherewill:
@sofokli1 I've spoke to @amulware and we don't really have an easy way to add a sync test to this PR. 

With that in mind, I've created #5915 so we can move on with this. The Integration test already covers the cases we use anyways

PR #5910, opened 2019-09-17
Make UISwitch.Changed() extension more generic

arturs105:
## What's this?
This PR moves `UISwitch.Changed()` extension method to the more generic class `UIControl`, since everything that's  needed for the extension is available on `UIControl`.

## Why do we want this?
So all `UIControl` subclasses can benefit from this extension.

## :squid: Permissions
:squid:


Odrakir:
Won't you make an extension specific to UISwitch that returns the bool value?

arturs105:
@Odrakir Naah, since it would need a different name and i couldn't come up with a good one.

Odrakir:
Let me help you: I think there’s a hint in the name of the company you work for...

PR #5912, opened 2019-09-17
Fix layout issues in iPad

### LINKED ISSUE #5531: Edit view cut off in iPad hover view
## 🐛 Bug Report
When I edit a time entry on iPad with the app in the small hover state, it cuts off the bottom of the edit view, making it hard to delete/confirm changes.

5D4A8B4E-382C-497A-A3B6-D3EB4BE4E628

## Context

Build/Branch: 2.0.1 release
Phone OS: iOS 12
Phone model: iPad Pro

## Expected behavior

There should be enough padding for all buttons to be visible and looking nice.

## To reproduce

Open the app on iPad in the small hover mode and tap a time entry.

## Suggestions
Probably a constraint issue? Or something about this particular app layout that’s not taken into account correctly. Maybe worth checking if this affects other modals as well.

### LINKED ISSUE #5533: Start view resizes and gets cut off in iPad hover view
## 🐛 Bug Report
When having the app open in the hover view on iPad with the hardware keyboard attached, the model changes size depending on whether the keyboard is ‘active’ and the bottom row of buttons, including the confirm button are hidden by the active keyboard auto completion bar. The size change happens in other layouts as well, but the hidden bottom area only happens in the hover state and is probably related to #5531

Video in Slack: https://toggl.slack.com/archives/CD40QCV25/p1562180033011100

## Context

Build/Branch: 2.0.1
Phone OS: iOS 12.3
Phone model: iPad Pro

## Expected behavior
The size should not change (I’m not sure if it should always be big or small, do our designs specify?).
And the bottom bar should never be hidden.

## To reproduce
Open the app in hover layout, hit the start button, type and dismiss the keyboard to change the model size back and forth.

## Suggestions
If our designs cover how this should look, let’s make it look that way. If not, let’s figure out how it should look/behave first.

Odrakir:
## What's this?
Exactly what the title says, I'll get into detail.

### Relationships
Closes #5531 and closes #5533

## Why do we want this?
We want our views to be beautiful for every device. And we also want our code to not be too messy.

## How is it done?
Basically I had to replace most the checks for iPad `UIDevice.CurrentDevice.UserInterfaceIdiom == UIUserInterfaceIdiom.Pad` for checks for regular horizontal size class: `TraitCollection.HorizontalSizeClass == UIUserInterfaceSizeClass.Regular`.

This way the hover view on iPad (and the 2 smaller split view ones) behave exactly like the iPhone does in terms of layout.

Then I also did some cleaning. I hope I didn't break anything, so I would like @dshr to check, as I think he was the one in charge of the iPad project.

### Some other notes

- Generally we should try to check for size classes and not specific device families when possible
- I (kind of) kept it for now, but I'm not sure about the top margin for modals in iPad. Does it make sense if the whole screen (including the status bar) is being darkened?
- I think we should make some changes to the edit and start views. For starters the content should be in scroll views. If you look at landscape on a new iPad Pro everything seems fine, but if you test it on and old one, the buttons get cut, and there's no way to use them without dismissing the keyboard.
ipads
- In addition to that, I think the content inside those modals should live in a navigation bar. Meaning the title and close buttons should all be in that bar. Visually won't change much, but it'll allow us to push views and also it facilitates the layout of that top elements. I would like you to look at how the AppStore (for example) handles similar modals (and also see how they are done in iPhone)
appstore
They look similar, but not quite.
- Another option is to use pop overs for this, like the calendar app does... Not sure if that would fit, maybe for some stuff.
- Finally, the stacking depth thing... I don't know why we are doing this. It makes the code more complicated and it's barely visible in a couple of occasions and it really isn't an standard behaviour.

Anyway. My point is I would like to go over all of this with a designer and redo some of it.

## Review hints
I don't know, there's just one commit, but this kind of code is not too easy to review. Test it as much as you can.

## :squid: Permissions
Go, but let @dshr review.


Odrakir:
I completely broke this views on iPhone. Let me try again. I'll close this PR.

PR #5918, opened 2019-09-17
Prevent fragment being added twice

### LINKED ISSUE #5917: java.lang.IllegalStateException: Fragment already added: MainFragment
### 👥 Impact
2 times in 2.3.1 (0.01% of users)

### 🔗 Context
The app is trying to add the same fragment twice in the `MainTabBarActivity` and then things go boom boom

### 🛰 Link
https://appcenter.ms/orgs/Toggl/apps/Toggl-Android/crashes/errors/2837185150u/reports

`Exception type and message`
```cs
Stacktrace
    android.support.v4.app.FragmentManagerImpl.addFragment FragmentManager.java:1916
    android.support.v4.app.BackStackRecord.executeOps BackStackRecord.java:765
    android.support.v4.app.FragmentManagerImpl.executeOps FragmentManager.java:2625
    android.support.v4.app.FragmentManagerImpl.executeOpsTogether FragmentManager.java:2411
    android.support.v4.app.FragmentManagerImpl.removeRedundantOperationsAndExecute FragmentManager.java:2366
    android.support.v4.app.FragmentManagerImpl.execPendingActions FragmentManager.java:2273
    android.support.v4.app.FragmentManagerImpl$1.run FragmentManager.java:733
    android.os.Handler.handleCallback Handler.java:873
    android.os.Handler.dispatchMessage Handler.java:99
    android.os.Looper.loop Looper.java:214
    android.app.ActivityThread.main ActivityThread.java:7037
    java.lang.reflect.Method.invoke Method.java
    com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run RuntimeInit.java:494
    com.android.internal.os.ZygoteInit.main ZygoteInit.java:965
```

heytherewill:
## What's this?
This prevents the app from adding the same fragment twice

### Relationships
Closes #5917

## Why do we want this?
No boom-boom

## How is it done?
There basically are only two places where we do fragment transactions in our app. The first one checks for `IsAdded`, the second doesn't because it assumes the app is starting fresh. It's now checking.

### Side effects
Since the fragment is already there, none. I want the reviewers to make sure that the rest of the code should indeed run and not early return.

## :squid: Permissions
Only myself


Toggl.Droid/Activities/MainTabBarActivity.cs
```diff
@@ -233,10 +233,13 @@ private void showInitialFragment(int initialTabItemId)
             SupportFragmentManager.RemoveAllFragments();
 
             var initialFragment = getCachedFragment(initialTabItemId);
-            SupportFragmentManager
-                .BeginTransaction()
-                .Add(Resource.Id.CurrentTabFragmmentContainer, initialFragment)
-                .Commit();
+            if (!initialFragment.IsAdded)
```
daividssilverio: I'm confused by this 🤔. There shouldn't be any fragments in the fragment manager at this point 🤔 .

Unless a fragment gets added by the system between `SupportFragmentManager.RemoveAllFragments();` and the `initialFragment` is added.

I guess there's no harm in adding this check, even though it looks like it's going to run at all times.

PR #5922, opened 2019-09-18
Shake e-mail field in forgot password view when the email is invalid

### LINKED ISSUE #5718: Shake email field in forgot password field when e-mail is invalid
Do these steps in the login or signup view:
1. Enter an invalid e-mail address (for example: `not.a.valid.email.address`)
2. Press the login/signup button
3. The e-mail field with the invalid email did a little shake to indicate that the e-mail isn't valid

Now repeat the same steps in forgot password view - the e-mail input field does not shake there. We should add the shaking behaviour to it as well for consistency.

arturs105:
## What's this?
Apart from the thing described in the issue, this PR also fixes the activity indicator not being visible, when it should - it was behind the login button, so i just moved it on top of the login button in the xib file.

### Relationships

Closes #5718 

## Review hints
Do lots of functional tests and try to break things.

## :squid: Permissions
🦑


heytherewill:
TFW no droid shake

Toggl.Core.UI/ViewModels/ForgotPasswordViewModel.cs
```diff
@@ -49,7 +52,8 @@ public sealed class ForgotPasswordViewModel : ViewModel<EmailParameter, EmailPar
             this.analyticsService = analyticsService;
             this.rxActionFactory = rxActionFactory;
 
-            Reset = rxActionFactory.FromObservable(reset, Email.Select(email => email.IsValid));
+            Reset = rxActionFactory.FromObservable(reset);
+            PasswordResetWithInvalidEmail = passwordResetWithInvalidEmailSubject.AsObservable();
```
Odrakir: I would make this a Driver

Odrakir: And `PasswordResetSuccessful`, even if it's not part of this PR.

maritoggl:
I see two options actually. One is that in case of an invalid e-mail we disable the button so that it cannot be interacted with at all. Another is that we allow interaction and thus shake the button, but then also display an error text that says what's wrong. I'm leaning towards the other option whereby
* if user enters an invalid e-mail (invalid format) and hits the reset button, the field is being shaked and "The e-mail doesn't look right" error is displayed
* if user enters a valid e-mail address that has not registered with Toggl, and hits the reset button, the field is being shaked and "E-mail address does not exist" error is displayed.
☝️both of these texts have been taken from webapp, we might want to review the copy (mainly for the second error), but in either case, I would keep away from disabling any buttons as not being able to interact with the button also eliminates triggering an error that helps the user overcome the issue they're having. 
Also, if this PR is for shaking only, we can handle adding these error texts in a followup issue.

arturs105:
> I think we should make up our minds about that, the button is either disabled, so it shouldn't do anything or not

I think we have made up our minds. We need the button to _look_ disabled, so the user sees that logging in can't be finished, but at the same time we need to show the user which fields are filled incorrectly, so we can't really do `.Enabed = false`, because we need to get the button press events if the user tries to "log in" with invalid credential(s). Now there are tons of ways to show, which fields are filled incorrectly, but this is how it's done in the login and signup views and the initial issue was about increasing consistency between these views.

I think shaking the field is really efficient for error that we can detect quickly (without making any network calls). Shaking the field after 4 seconds of loading time isn't as useful i think. We could add shaking to network login errors as well, but that's not what the initial issue is about.

PR #5927, opened 2019-09-18
📐 Prevent text in pie chart from being upside down

### LINKED ISSUE #5511: Pie chart text is upside down on the right
## 🐛 Bug Report
_This is not really a bug_ - the current design of the pie chart means text labels on the right side appears upside down.
image

## Context
iOS 2.0.1, Android 2.0.1 - reports page

## Expected behavior
It would be nice if the text always appeared right side up :)

## To reproduce
Create a report with four equal project durations in the app. The top right slice's text will be upside down.

## Suggestions
[Slack](https://toggl.slack.com/archives/C029WU7AY/p1561444620161300)

pdrj:
## What's this?
This fixes the text being upside down in the pie chart from reports view.

### Relationships

Closes #5511

## Why do we want this?
Because it's hard to read stuff upside down.

## How is it done?
Pie slices on the right are rendered differently. Screenshot explanation (left is iOS, right Android):

Screenshot 2019-09-18 at 17 27 59
Screenshot 2019-09-18 at 17 27 34
Screenshot 2019-09-18 at 17 31 27
Screenshot 2019-09-18 at 17 29 50
Screenshot 2019-09-18 at 17 35 29
Screenshot 2019-09-18 at 17 34 13
Screenshot 2019-09-18 at 17 32 37

As you can see, especially on iOS there seems to be a problem with long project names and wide letters - but it's not caused by this PR and I'll create an issue for that.

### Why not another way?
In the issue we decided on a slightly different design, but that causes problems like:
Screenshot 2019-09-18 at 15 35 27
Hence, on the right portion of the pie, percentages are displayed above the text, to let the beginning of the text use as much height as possible. It also has a side effect of better 'point symmetricalness' ;).

## Review hints
Create a bunch of TEs and assign different projects to them. Then go to the reports view.

## :squid: Permissions
Me


Toggl.Droid/Views/PieChartView.cs
```diff
@@ -64,12 +66,22 @@ protected override void OnMeasure(int widthMeasureSpec, int heightMeasureSpec)
             base.OnMeasure(widthMeasureSpec, widthMeasureSpec);
         }
 
+        private bool isSegmentOnTheRight(float endDegrees) => endDegrees > 270 || (endDegrees >= 0 && endDegrees <= 90);
+
+        private (int width, int height) getTextWidthAndHeight(string text)
+        {
+            var bounds = new Rect();
```
daividssilverio: Could create a field and reuse this rect?

Toggl.Droid/Views/PieChartView.cs
```diff
@@ -94,21 +106,42 @@ public override void Draw(Canvas canvas)
...
 
+                    var isOnTheRight = isSegmentOnTheRight(endDegrees);
+                    var integerPercentage = (int) (percent * 100);
+                    var nameToDraw = segment.FormattedName();
+                    var percentageToDraw = $"{integerPercentage}%";
+
+                    var (textWidth, textHeight) = getTextWidthAndHeight(nameToDraw);
+                    var (percentWidth, percentHeight) = getTextWidthAndHeight(percentageToDraw);
+
+                    var nameCoordinates = new Point();
```
daividssilverio: Same for these points, please create fields for them and reuse as necessary.

Toggl.Droid/Views/PieChartView.cs
```diff
@@ -64,12 +66,22 @@ protected override void OnMeasure(int widthMeasureSpec, int heightMeasureSpec)
...
+
+        private (int width, int height) getTextWidthAndHeight(string text)
+        {
+            var bounds = new Rect();
+            textPaint.GetTextBounds(text, 0, text.Length, bounds);
+
+            return (bounds.Width(), bounds.Height());
+        }
+
         public override void Draw(Canvas canvas)
```
daividssilverio: Hey @heytherewill , why is this view overriding `Draw` instead of `OnDraw`? The latter should be used for most cases. Ref: https://developer.android.com/reference/android/view/View.html#draw(android.graphics.Canvas)

Toggl.Droid/Views/PieChartView.cs
```diff
@@ -94,21 +106,42 @@ public override void Draw(Canvas canvas)
...
 
-                    // Draw the text
-                    var integerPercentage = (int)(percent * 100);
-                    var nameToDraw = segment.FormattedName();
-                    var percentageToDraw = $"{integerPercentage}%";
+                        nameCoordinates.X = radius - padding - textWidth;
+                        nameCoordinates.Y = -padding + textHeight;
+
+                        percentageCoordinates.X = radius - padding - percentWidth;
+                        percentageCoordinates.Y = -1.5 * padding + -textHeight + percentHeight;
```
daividssilverio: what is this `1.5f`? I know it was there before, it seems to be related to the line height between the percentage and the name, but it could have a name.

Toggl.Droid/Views/PieChartView.cs
pdrj: Sure, done.

Toggl.Droid/Views/PieChartView.cs
pdrj: Done.

Toggl.Droid/Views/PieChartView.cs
pdrj: Extracted - I think it's more descriptive to extract `0.5 * padding` as this is the height of the separator between lines.

Toggl.Droid/Views/PieChartView.cs
pdrj: Good catch, this does seem a bit unsafe, especially since the docs say 
> If you override this method you must call through to the superclass implementation.

and we don't. I've tested it with `onDraw` and it seems to work fine. Let's wait for @heytherewill to see if there's a reason to do `draw` instead of `onDraw`.

heytherewill: No reason to, please change it

pdrj: Done.

Issue #5820, opened 2019-08-28
Multiple crashes in recycler views


amulware:
## 💥 Crash Report
Users report that trying to type anything in the text box of the tag selection view (opened from the edit view) crashes the app.

### 👥 Impact
- 5% of total crashes on 2.2.1

### 🔗 Context
- Could be the same as #5691
- and perhaps related to #5692

### 🛰 Link
https://appcenter.ms/orgs/Toggl/apps/Toggl-Android/crashes/errors/3633119050u/overview

### 📝  Exception & Stack Trace
```cs
android.support.v7.widget.RecyclerView$Recycler.tryGetViewHolderForPositionByDeadline
RecyclerView.java, line 5923
```
`
java.lang.IndexOutOfBoundsException: Inconsistency detected. Invalid item position 4(offset:-1).state:13 android.support.v7.widget.RecyclerView{9bf8678 VFED.V... ......ID 0,196-1440,1731 #7f0d01ee app:id/SelectTagsRecyclerView}, adapter:md5e6930ea5438967c9497bdc857a01d090.SelectTagsRecyclerAdapter@82aa065, layout:android.support.v7.widget.LinearLayoutManager@a550151, context:md56d31f6f2c59878fdf075cc45b4e7258a.SelectTagsActivity@582a55c
`
```cs
android.support.v7.widget.RecyclerView$Recycler.tryGetViewHolderForPositionByDeadline RecyclerView.java:5923
android.support.v7.widget.RecyclerView$Recycler.getViewForPosition RecyclerView.java:5858
android.support.v7.widget.RecyclerView$Recycler.getViewForPosition RecyclerView.java:5854
android.support.v7.widget.LinearLayoutManager$LayoutState.next LinearLayoutManager.java:2230
android.support.v7.widget.LinearLayoutManager.layoutChunk LinearLayoutManager.java:1557
android.support.v7.widget.LinearLayoutManager.fill LinearLayoutManager.java:1517
android.support.v7.widget.LinearLayoutManager.onLayoutChildren LinearLayoutManager.java:622
android.support.v7.widget.RecyclerView.dispatchLayoutStep1 RecyclerView.java:3875
android.support.v7.widget.RecyclerView.dispatchLayout RecyclerView.java:3639
android.support.v7.widget.RecyclerView.onLayout RecyclerView.java:4194
android.view.View.layout View.java:20715
android.view.ViewGroup.layout ViewGroup.java:6202
[...]

Issue #5692, opened 2019-08-01
IndexOutOfBoundsException on RecyclerView$Recycler.tryGetViewHolderForPositionByDeadline


daividssilverio:
## 💥 Crash Report
It seems to be a crash that happens when adding/selecting tags.

### 👥 Impact
I has affected 167 users, with 446 occurrences; 
About 0.2% of our users in the last 30 days. 

### 🔗 Context
Very similar to #5691 and possibly #5569
Could have been already fixed by #5624

### 🛰 Link
https://appcenter.ms/orgs/Toggl/apps/Toggl-Android/crashes/errors/917105280u/overview

### 📝  Exception & Stack Trace
`Exception type and message`
```cs
java.lang.IndexOutOfBoundsException: Inconsistency detected. Invalid item position 5(offset:-1).state:12 android.support.v7.widget.RecyclerView{f6b25a3 VFED.V... ......ID 0,168-1080,1050 #7f0d01e9 app:id/SelectTagsRecyclerView}, adapter:md5e6930ea5438967c9497bdc857a01d090.SelectTagsRecyclerAdapter@fab1958, layout:android.support.v7.widget.LinearLayoutManager@528e0a0, context:md56d31f6f2c59878fdf075cc45b4e7258a.SelectTagsActivity@bd46e3b
  at android.support.v7.widget.RecyclerView$Recycler.tryGetViewHolderForPositionByDeadline(RecyclerView.java:5923)
  at android.support.v7.widget.RecyclerView$Recycler.getViewForPosition(RecyclerView.java:5858)
  at android.support.v7.widget.RecyclerView$Recycler.getViewForPosition(RecyclerView.java:5854)
...
```

pdrj:
@daividssilverio this seems to be the same as #5820 - are you OK with closing this one? I know this issue has been created first, but the other one has a bit more info in it.

daividssilverio:
Alright @pdrj . Superseded by #5820 .

PR #5940, opened 2019-09-23
Base our Recycler Adapters on Android's ListAdapter

### LINKED ISSUE #5925: Move all our adapters to ListAdapter base class
## 🚧 Refactor Proposal

Related slack thread https://toggl.slack.com/archives/C02APMYTM/p1568810483151900

Our implementation of `BaseRecyclerAdapter` is complicated and there are now several issues that might be related to some unpredictable behaviors (#5692 #5691 ). Therefore, we want to move to standard `ListAdapter` [docs](https://developer.android.com/reference/android/support/v7/recyclerview/extensions/ListAdapter) and let google handle all of these issues.

## Dependencies

First, we need to verify that `ListAdapter` is available and works correctly on Xamarin.

## Technical aspects

Basically we need to replace `BaseRecyclerAdapter` everywhere as a base class and use the `ListAdapter` instead.  Extending the `ListAdapter` is pretty straight forward, just follow the [docs](https://developer.android.com/reference/android/support/v7/recyclerview/extensions/ListAdapter)

semanticer:
## What's this?

We move from our custom implementation of `BaseRecyclerAdapter` to implementation that delegates much of is logic to Android's `ListAdapter`

### Relationships

Closes #5925 

Maybe also #5692 and #5820 - these are hard to reproduce and verify but there is a good chance they are fixed since the implementation completely changed towards standard Android way

## Why do we want this?

We want to use existing battle-tested solution instead of our DIY implementation

## How is it done?

`BaseRecyclerAdapter` now extends `ListAdapter` and bridges it's methods.

### Why not another way?

Let me know if you have any other ideas

### Side effects

Since `ListAdapter` returns java objects we need to cast them to our C# equivalents.  `BaseRecyclerAdapter` now provides 'safe' methods like `getTypedItem(pos)` in order to shield all its subclasses from this casting magic shenanigans. This can theoretically be a performance problem, especially in the DiffingCallback.

## Review hints

Well if you have time you can double-check all our recycler views are working and updating the same way as before. 

## :squid: Permissions

Me or @heytherewill

### DIFF
--- a/Toggl.Droid/Extensions/ObjectTypeExtensions.cs
+++ b/Toggl.Droid/Extensions/ObjectTypeExtensions.cs
@@ -1,0 +1,11 @@
+namespace Toggl.Droid.Extensions
+{
+    public static class ObjectTypeExtensions
+    {
+        public static T Cast<T>(this Java.Lang.Object obj) where T : class
+        {
+            var propertyInfo = obj.GetType().GetProperty("Instance");
+            return propertyInfo == null ? null : propertyInfo.GetValue(obj, null) as T;
+        }
+    }
+}
--- a/Toggl.sln
+++ b/Toggl.sln
@@ -21,21 +21,21 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Toggl.Shared.Tests", "Toggl
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Toggl.Networking", "Toggl.Networking\Toggl.Networking.csproj", "{5ACEB155-9DD4-4309-BCC9-997A09950393}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Toggl.Networking.Tests", "Toggl.Networking.Tests\Toggl.Networking.Tests.csproj", "{8598D5CB-4E3A-4F99-8E80-CC4EA8609ED8}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Toggl.Networking.Tests", "Toggl.Networking.Tests\Toggl.Networking.Tests.csproj", "{8598D5CB-4E3A-4F99-8E80-CC4EA8609ED8}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Toggl.Networking.Tests.Integration", "Toggl.Networking.Tests.Integration\Toggl.Networking.Tests.Integration.csproj", "{C400D1A6-5DC5-4A92-8DD0-FF272D0D1189}"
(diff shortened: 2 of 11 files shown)


Toggl.Droid/Extensions/ObjectTypeExtensions.cs
```diff
@@ -0,0 +1,11 @@
+namespace Toggl.Droid.Extensions
+{
+    public static class ObjectTypeExtensions
+    {
+        public static T Cast<T>(this Java.Lang.Object obj) where T : class
+        {
+            var propertyInfo = obj.GetType().GetProperty("Instance");
```
heytherewill: Shit, is this needed? Won't `obj as T` suffice?

Alternatively, couldn't you simply `obj.JavaCast<T>`? I don't really remember what JavaCast does, I'm just trying to figure out a solution that doesn't involve using reflection 🙈

Toggl.sln
```diff
@@ -21,29 +21,29 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Toggl.Shared.Tests", "Toggl
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Toggl.Networking", "Toggl.Networking\Toggl.Networking.csproj", "{5ACEB155-9DD4-4309-BCC9-997A09950393}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Toggl.Networking.Tests", "Toggl.Networking.Tests\Toggl.Networking.Tests.csproj", "{8598D5CB-4E3A-4F99-8E80-CC4EA8609ED8}"
```
heytherewill: Not relevant to this PR

Toggl.Droid/Extensions/ObjectTypeExtensions.cs
kornelijepetak: I am almost 100% sure that this can be done with `as T`.
Avoid reflection as hell 😛

kornelijepetak: Especially since this code here will be called quite a lot!

semanticer: `obj as T` results in `null` :-( 

I will try the `JavaCast` variant

semanticer: `obj.JavaCast<T>` requires T to implement `IJavaObject` https://docs.microsoft.com/en-gb/dotnet/api/Android.Runtime.Extensions.JavaCast?view=xamarin-android-sdk-9

pdrj:
Regarding #5820 - good news - I can't reproduce it with data that crashes each time on develop :).

Will test the other adapters too.

semanticer:
After discussion here https://toggl.slack.com/archives/CD40QCV25/p1569248243020200 and the bug fund by @pdrj above ☝️ we decided to move the related issue to the next release and try some other approach with 0 reflection involved.

heytherewill:
Is it worth to keep this open?

semanticer:
We agreed that reflection is unacceptable and this solution depends on it. I'm closing this PR, should try a different approach. 
Related slack thread: https://toggl.slack.com/archives/CD40QCV25/p1569248243020200

PR #5941, opened 2019-09-23
Re-enable linker all for release builds

heytherewill:
## What's this?
This (hopefully) makes our apps smaller again

## Why do we want this?
Our binaries are so massive they are starting to have their own gravity
 
## How is it done?
[I'm linking everything *but* the faulty assembly](https://github.com/xamarin/xamarin-android/issues/3419#issuecomment-533703511)

### Why not another way?
There's a bug in Xamarin that prevents us to link everything

### Side effects
Smaller apps

## Review hints
Test the app and specially the [Play Store](https://app.bitrise.io/build/e1db2896263a549f#?tab=log) and [AppStore](https://app.bitrise.io/build/3ffcf75979fbec75#?tab=log) builds

## :squid: Permissions
Only myself


juanlaube:
The change looks good, the builds are taking forever tho 😄

PR #5946, opened 2019-09-24
Lazy initialize views

### LINKED ISSUE #4423: Android Bottom Bar: Create placeholders for pages
The bottom bar is cool, but switching pages can be a bit sluggish, specially on the Reports view and most likely the calendar page when it's ready.

We can add a fade transition to reduce a bit of the problem, but ideally, we would need placeholders to show while the data is loading.

### LINKED ISSUE #6677: Run initial sync in a background thread
Starting background sync during startup (https://github.com/toggl/mobileapp/blob/develop/Toggl.iOS/Startup/AppDelegate.cs#L116) takes non-trivial amount of time (ca 180ms on my iPhone 7, see the full report in #5954).

The reason why we trigger it during startup and not when the app reaches the main screen is to get the data from the server as soon as possible and to decrease the time the user will see old data when the app is shown. This effort shouldn't block app's startup process. It would be safe to wrap this in a `Task.Run(() => ...)` and run this in a background thread and let the UI thread initialize the rest of the app.

heytherewill:
##  What's this?
This deduplicates some of our startup calls, changes some things to run on a BG thread and then makes startup feel faster by changing tabs immediately and waiting for initialization asynchronously

### Relationships
Closes #4423
Closes #6677

## Why do we want this?
Better UX

## How is it done?
- Some calls (namely the creation of the TimeEntriesLog and the calculation of suggestions) were being called too often. 
- Many things were happening sequentially on the main thread during the app's startup
- We were awaiting all the ViewModels at once during startup and before showing the first VM.

All of the above is now false

### Why not another way?
Feel free to suggest improvements!

### Side effects
I haven't at all tested iOS because my Mac is still getting repaired, but in principle it should be fine.
We need to test this thoroughly, though, specially on mid-end phones

## Review hints
Looking at it commit by commit should be enough.
Also, don't test using Debug, use [this AdHoc build instead](https://app.bitrise.io/build/4d278b55018587e7#?tab=log) so you don't have Debug shenanigans in the way of the speed :sanic:

## :squid: Permissions
Only myself


Toggl.Core.UI/AppStart.cs
```diff
@@ -96,10 +97,11 @@ public void UpdateOnboardingProgress()
...
 
-        public void ForceFullSync()
-        {
-            dependencyContainer.SyncManager.ForceFullSync().Subscribe();
-        }
+        public Task ForceFullSync()
+            => Task.Run(() => dependencyContainer
+                .SyncManager
+                .ForceFullSync()
+                .Subscribe());
```
amulware: Does this need to be async? The operation should be reasonably light, no? unless... this might trigger the creation of all the needed dependencies in the first place?

Toggl.Core.UI/ViewModels/MainTabBarViewModel.cs
```diff
@@ -1,153 +1,107 @@
...
         }
 
-        public override async Task Initialize()
+        public TViewModel GetViewModel<TViewModel>()
+            where TViewModel : class, IViewModel
         {
-            await base.Initialize();
+            var expectedType = typeof(TViewModel);
+            if (expectedType == typeof(MainViewModel))
+                return mainViewModel.Value as TViewModel;
```
amulware: Turn these into explicit casts so it fails fast on null.

Toggl.Core.UI/ViewModels/TimeEntriesViewModel.cs
```diff
@@ -22,10 +22,12 @@
 
 namespace Toggl.Core.UI.ViewModels
 {
+    using MainLogSection = AnimatableSectionModel<DaySummaryViewModel, LogItemViewModel, IMainLogKey>;
 
     [Preserve(AllMembers = true)]
     public sealed class TimeEntriesViewModel
     {
+        private static readonly TimeSpan timeEntriesThrottlePeriod = TimeSpan.FromSeconds(0.1);
```
amulware: Why such a large throttle? Why this value? How does the below throttle help at all? (i.e. what events are coming in too often/fast that need throttling?)

Toggl.Core.UI/ViewModels/MainViewModel.cs
```diff
@@ -209,13 +193,15 @@ public override async Task Initialize()
 
             CurrentRunningTimeEntry = dataSource.TimeEntries
                 .CurrentlyRunningTimeEntry
-                .AsDriver(schedulerProvider);
+                .AsDriver(schedulerProvider)
```
juanlaube: Would an extension method like `.AsDriver(ISchedulerProvider, IScheduler)` make sense? 🤔

Toggl.Core.UI/AppStart.cs
juanlaube: I agree with @amulware, this a fire and forget, the side effect of the subscription triggers the sync manager.

Toggl.Droid/Presentation/ActivityPresenter.cs
```diff
@@ -36,7 +36,7 @@ public sealed class ActivityPresenter : AndroidPresenter
             typeof(TokenResetViewModel)
         };
 
-        private readonly Dictionary<Type, ActivityPresenterInfo> presentableActivitiesInfos = new Dictionary<Type, ActivityPresenterInfo>
+        private readonly Dictionary<Type, ActivityPresenterInfo> presentableActivitiesInfos = new Dictionary<Type, ActivityPresenterInfo>(18)
```
juanlaube: I understand why, but I don't like it, it's super easy to forget about this 😞

Toggl.Core.UI/AppStart.cs
heytherewill: This is the first point sync is called and it makes the call expensive because it initializes a bunch of dependencies. `Task.Run` is there to make it run on the thread pool

Toggl.Core.UI/ViewModels/TimeEntriesViewModel.cs
heytherewill: The value was there already, I merely copied it over from another file

Toggl.Core.UI/ViewModels/MainViewModel.cs
heytherewill: Not really

Toggl.Droid/Resources/layout/MainTabBarActivity.axml
```diff
@@ -6,15 +6,30 @@
...
         android:layout_height="0dp"
         app:layout_constraintTop_toTopOf="parent"
-            app:layout_constraintBottom_toTopOf="@+id/MainTabBarBottomNavigationView"
-            app:layout_constraintStart_toStartOf="parent"
-            app:layout_constraintEnd_toEndOf="parent">
+        app:layout_constraintBottom_toTopOf="@+id/MainTabBarBottomNavigationView"
+        app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintEnd_toEndOf="parent" />
+
+    <FrameLayout
```
daividssilverio: why the `FrameLayout`? Using just the `ProgressBar` showed any problems?

Toggl.Droid/Activities/MainTabBarActivity.cs
```diff
@@ -205,67 +216,86 @@ public override void OnBackPressed()
...
+                    var fragment = t.Result;
+                    if (fragment is IScrollableToTop scrollableToTop)
+                    {
+                        scrollableToTop.ScrollToTop();
+                    }
+                });
         }
 
-        private void showFragment(Fragment fragment)
+        private async Task showFragment(int itemId)
```
daividssilverio: 🤔 While testing, I managed to fairly easily crash the app by selecting the bottom bar tabs faster than it loads.
Some quick crashes:
- select all (or any two) the tabs quickly as soon as I see the bottom I've got
```
java.lang.IllegalStateException: Fragment already added: CalendarFragment{9a60f40 #3 id=0x7f0d0181}
```
- pressing the settings tab as soon as the app opens multiple times:
```
java.lang.NullPointerException: Attempt to invoke virtual method 'void android.support.v4.app.Fragment.setNextAnim(int)' on a null object reference
        at android.support.v4.app.BackStackRecord.executeOps(BackStackRecord.java:772)
        at android.support.v4.app.FragmentManagerImpl.executeOps(FragmentManager.java:2625)
        at android.support.v4.app.FragmentManagerImpl.executeOpsTogether(FragmentManager.java:2411)
        at android.support.v4.app.FragmentManagerImpl.removeRedundantOperationsAndExecute(FragmentManager.java:2366)
        at android.support.v4.app.FragmentManagerImpl.execPendingActions(FragmentManager.java:2273)
        at android.support.v4.app.FragmentManagerImpl$1.run(FragmentManager.java:733)
```

juanlaube:
I did some functional testing, the app feels faster and didn't find any bugs.

heytherewill:
Putting this on-hold until #6083 gets resolved

daividssilverio:
Removed `on-hold` as #6083 got merged.

sofokli1:
For me this crashes in `MainViewController` line 737 when trying to access `ViewModel.IsTimeEntryRunning`

heytherewill:
I'm closing this for the time being, for it's:

- Old
- Complex
- Full of possibly useless changes
- Hard to review
- Error prone

Fret not, for most of the code here will be introduced in the codebase. I will just do so in more manageable ways so that we don't depend on a single god PR and fail our OKR again :)

Issue #5854, opened 2019-09-10
Calendar: Add option to move time entries in the calendar view


sofokli1:
Option to move time entries in the calendar view by dragging them from one place to another (supported on iOS)

daividssilverio:
Closed by #5978

PR #5957, opened 2019-09-25
Handle drag to edit in the droid calendar

### LINKED ISSUE #5856: Calendar: Add option to edit the start and end time of a time entry
Option to edit the start and end time of a time entry by dragging the start/end time in the calendar view (supported on iOS)

daividssilverio:
## What's this?
This adds the basic drag to edit functionality to the droid calendar.

### Relationships
Closes #5856 and part of #5854

## Why do we want this?
The droid calendar has to catch up.

## How is it done?
Except for some details, it's a copy of the `CalendarCollectionViewEditItemHelper`, but adapted for android with some extra details from `CalendarCollectionViewAutoScrollHelper`.

Most of the logic is the same.

### Why not another way?
We could also make the methods adapted from `CalendarCollectionViewAutoScrollHelper` shared. Not sure if would cause problems though (maybe we could do it later).

## Review hints
Please start reviewing, but I'll leave this as a draft until I finish cleaning up and fixing some details.
For now, all changes only affect the UI.

## :squid: Permissions
🦑 it 📆 team.


daividssilverio:
I wanted to create it as a draft... but I can't revert. I'm closing it and open again as a draft.

PR #5960, opened 2019-09-25
Use attributed string for suggestions cell

### LINKED ISSUE #5661: Padding issue in suggestion cells
## 🐛 Bug Report
There's a small whitespace/padding issue in the suggestion cells: the project name is too close to the project dot.

IMG_2347

## Context

Build/Branch: 2.1.3 AppStore build
Phone OS: latest iOS 12
Phone model: iPhone 6S

## Expected behavior
The padding should be the same as in the TE cells. I didn't check designs (please do) but I'm quite sure of that.

## To reproduce
Have suggestions with projects.

## Suggestions
Probably just a single number in some constraint somewhere.

juanlaube:
## What's this?
Fixes a layout issue in the suggestions cell by using an attributed string with the dot as an attachment, like the main log cells.

### Relationships

Closes #5661 

## Why do we want this?
📐 

## How is it done?
I removed the dot and some labels simplifying the layout and used the same attributed string builder we use in other places like the main log.

### Why not another way?
This is simple, more consistent and resilient to changes.

### Side effects
None.

## Review hints
Try different scenarios like without a project, only with project, project and client, etc.
Also try on iPhone and iPad.

## :squid: Permissions
🦑


heytherewill:
Shouldn't this be a hotfix against 2.4?

PR #5975, opened 2019-09-27
✨ Display sync status next to the logout button

### LINKED ISSUE #4907: Add sync status next to a sign out button in settings view
Users should be able to check the status of the sync process by looking at the settings screen. 
The best place to put it is close to the sign-out button because when they sign out the entirety of their local data is wiped out so everything that has not been synced would be lost.

~If the user taps sign out, maybe we should check whether there's unsynced data and warn the users about the possible loss of data.~ (read comments)

pdrj:
## What's this?
This displays sync status next to the logout button in settings.

### Relationships

Closes #4907

## Why do we want this?
iOS has it and it gives users more info about their state.

## How is it done?
Moving an observable from iOS solution to the view model and then some layout work.

### Why not another way?
This seems pretty clean.

## Review hints
Trigger sync, go to settings quickly. You should see the following.

Screenshot 2019-09-27 at 11 15 46
Screenshot 2019-09-27 at 11 15 57
Screenshot 2019-09-27 at 11 17 03
Screenshot 2019-09-27 at 11 16 41

Try to logout as well.

If we want to add some fancy design, I'd say we can do it later and now merge it as it is. @maritoggl please have a look at the above screenshots and say if you're OK with how it looks. It is done in a similar way on iOS.

## :squid: Permissions
Me,


Toggl.Core.UI/ViewModels/Settings/SettingsViewModel.cs
```diff
@@ -54,13 +54,21 @@ public sealed class SettingsViewModel : ViewModel
         public string Title { get; private set; } = Resources.Settings;
         public bool CalendarSettingsEnabled => onboardingStorage.CompletedCalendarOnboarding();
         public string Version => $"{platformInfo.Version} ({platformInfo.BuildNumber})";
+        
+        public enum SyncStatus
```
dshr: Perhaps could extract this into a separate file in `Toggl.Core.Sync`, like `SyncState.cs`? The iOS side of things looks great otherwise.

pdrj: Yeah I thought about it, but then we'd have `SyncState` and `SyncStatus` there - maybe a bit confusing? 🤷‍♂ I also thought about just reusing the one from `Core.Sync` but it doesn't quite match the values needed on the UI.

maritoggl:
I’m ok with this approach, but do we cover all instances with just “Synced” and “Syncing”? I think we should have a status that reflects there are unsynced entries as well as those are the most important entries in the context of logging out

pdrj:
@maritoggl I'm doing exactly what iOS is - there is 'synced', 'syncing' and 'logging out'. It would be useful to have what you propose, but maybe we should make a separate issue for that? What do you think?

maritoggl:
I agree, no need to extend the scope of this PR.

pdrj:
Thanks Mari. Only waiting for @toggl/androids review then.

Toggl.Core.UI/ViewModels/Settings/SettingsViewModel.cs
```diff
@@ -25,6 +25,7 @@
 using Toggl.Shared.Extensions;
 using Toggl.Storage.Settings;
 using Xamarin.Essentials;
+using PresentableSyncStatus = Toggl.Core.Sync.PresentableSyncStatus;
```
Odrakir: I think you don't need this line, given that you are importing `Toggl.Core.Sync`

Toggl.iOS/Cells/Settings/SettingsSyncCell.cs
```diff
@@ -7,10 +7,11 @@
 using Toggl.Shared;
 using UIKit;
 using Colors = Toggl.Core.UI.Helper.Colors;
+using PresentableSyncStatus = Toggl.Core.Sync.PresentableSyncStatus;
```
Odrakir: This is probably me not understanding the way things are done in C#, but isn't it clearer (more consistent) to just do `using Toggl.Core.Sync;'?

kornelijepetak: No, @Odrakir, your understanding of C# is on point.
The only reason why one would use an alias in this case is if `using Toggl.Core.Sync` would make visible some other type that would clash with another namespace. Which in this situation is not the case.

Toggl.Core.UI/ViewModels/Settings/SettingsViewModel.cs
```diff
@@ -207,6 +208,16 @@ public sealed class SettingsViewModel : ViewModel
 
             LoggingOut = loggingOutSubject.AsObservable()
                 .AsDriver(schedulerProvider);
+            
+            CurrentSyncStatus = Observable.CombineLatest(
+                IsSynced,
+                IsRunningSync,
+                LoggingOut.SelectValue(true).StartWith(false),
+                (synced, syncing, loggingOut) =>
```
kornelijepetak: Please extract this into a separate method to respect the levels of abstractions, especially when you are not utilizing closures.

Please do not write returns in the same line as `if`s.

Toggl.Droid/Fragments/SettingsFragment.ViewBinds.cs
```diff
@@ -79,7 +88,8 @@ protected override void InitializeViews(View fragmentView)
             groupTimeEntriesView = fragmentView.FindViewById(Resource.Id.GroupTimeEntriesView);
             defaultWorkspaceView = fragmentView.FindViewById(Resource.Id.DefaultWorkspaceView);
       
-            logoutView = fragmentView.FindViewById<TextView>(Resource.Id.SettingsLogoutButton);
+            txtLogout = fragmentView.FindViewById<TextView>(Resource.Id.TxtSettingsLogout);
+            logoutView = fragmentView.FindViewById<View>(Resource.Id.SettingsLogoutButton);
```
kornelijepetak:
```suggestion
            logoutView = fragmentView.FindViewById(Resource.Id.SettingsLogoutButton);
```

Toggl.Droid/Fragments/SettingsFragment.ViewBinds.cs
```diff
@@ -121,8 +131,16 @@ protected override void InitializeViews(View fragmentView)
             appBarLayout = fragmentView.FindViewById<AppBarLayout>(Resource.Id.AppBarLayout);
 
             scrollView = fragmentView.FindViewById<NestedScrollView>(Resource.Id.ScrollView);
-
-            logoutView.Text = Shared.Resources.SignOutOfToggl;
+            
+            syncStateSyncedHolder = fragmentView.FindViewById<View>(Resource.Id.SyncStateSyncedHolder);
```
kornelijepetak:
```suggestion
            syncStateSyncedHolder = fragmentView.FindViewById(Resource.Id.SyncStateSyncedHolder);
```

Toggl.Droid/Fragments/SettingsFragment.ViewBinds.cs
```diff
@@ -121,8 +131,16 @@ protected override void InitializeViews(View fragmentView)
...
+            syncStateInProgressHolder = fragmentView.FindViewById<View>(Resource.Id.SyncStateInProgressHolder);
```
kornelijepetak:
```suggestion
            syncStateInProgressHolder = fragmentView.FindViewById(Resource.Id.SyncStateInProgressHolder);
```

Toggl.Droid/Fragments/SettingsFragment.cs
```diff
@@ -164,6 +167,20 @@ public override void OnViewCreated(View view, Bundle savedInstanceState)
...
+                .DisposedBy(DisposeBag);
+        }
+
+        private void setSyncStatusView(PresentableSyncStatus status)
+        {
+            syncStateViews.Values.ForEach(view => view.Visibility = ViewStates.Gone);
+
+            txtStateInProgress.Text = status == PresentableSyncStatus.Syncing ? Syncing : LoggingOutSecurely;
+
+            syncStateViews.TryGetValue(status, out var visibleView);
```
kornelijepetak: If you are not using the `bool` return value to check for existance of the key, please use the dictionary access syntax.
```suggestion
            var visibleView = syncStateViews[status];
```

Toggl.iOS/Cells/Settings/SettingsSyncCell.cs
```diff
@@ -46,19 +47,19 @@ protected override void UpdateView()
         {
             switch (Item.CustomValue)
             {
-                case SyncStatus.Synced:
+                case PresentableSyncStatus.Synced:
```
kornelijepetak: You can avoid writing these with
`using static Toggl.Core.Sync.PresentableSyncStatus;`

Then just

`case Synced:`
`case Syncing:`
...

Toggl.Droid/Resources/layout/SettingsFragment.axml
```diff
@@ -613,18 +613,63 @@
...
+                <TextView
+                    android:id="@+id/TxtSettingsLogout"
+                    android:textSize="15sp"
+                    tools:text="Sign out of Toggl"
+                    android:gravity="bottom"
+                    android:textColor="@color/errorText"
+                    android:layout_height="32dp"
+                    android:layout_width="match_parent"
+                    android:layout_marginBottom="2dp"/>
+                <FrameLayout
```
semanticer: I don't love how deep the layout hierarchy is but it's consistent with the rest of the layout and it looks like this is how we roll for now :-(

Toggl.iOS/Cells/Settings/SettingsSyncCell.cs
pdrj: If I do this and remove `using Toggl.Core.Sync`, this line no longer compiles: `    public partial class SettingsSyncCell : BaseTableViewCell<CustomRow<PresentableSyncStatus>>`. Am I doing something wrong?

pdrj:
@kornelijepetak - I addressed all your comments except the one about static `using`. Please let me know what would you like to be done there, WRT to my reply to your comment.

Toggl.iOS/Cells/Settings/SettingsSyncCell.cs
kornelijepetak: Yeah, just add a `using static Toggl.Core.Sync.PresentableSyncStatus` without removing the `using Toggl.Core.Sync`.

`using static` just tells the compiler that for this file, the enum elements do not need to be fully qualified.

PR #5978, opened 2019-09-27
Add droid calendar drag auto scrolls

### LINKED ISSUE #5854: Calendar: Add option to move time entries in the calendar view

daividssilverio:
## What's this?
This does what is left to complete #5854 (the auto scrolls)
Auto scroll -> scroll that happens when the user drags entries over the screen edges.

### Relationships
Closes #5854

## Why do we want this?
It's a basic functionality

## How is it done?
With a scroller. Thanks android sdk.

### Why not another way?
Feel free to suggest improvements.

## Review hints
Give it a try. But the code should be easy to read.

## :squid: Permissions
🦑 📆 team


Toggl.Droid/Views/Calendar/CalendarDayView.EditItem.cs
```diff
@@ -219,6 +231,37 @@ private void changeOffset(float touchY)
...
+
+        private void autoScrollTop(int deltaY)
+        {
+            scroller.ForceFinished(true);
+            isScrolling = true;
+            scroller.StartScroll(0, scrollOffset, 0, -deltaY);
+            handler.Post(continueScroll);
+        }
+
+        private void autoScrollBottom(int deltaY)
```
pdrj: Maybe it's a matter of personal taste, but can we have only one `autoScroll` function and then call it with a positive or negative `deltaY`? Or have another argument `isScrollTop` or an `enum` with scroll mode and modify the delta according to that? What do you think?

sofokli1:
The above 3 points about functionality make sense. I will add that if the TE is 15min tall it will disappear under the top and bottom bar while scrolling.

Toggl.Droid/Views/Calendar/CalendarDayView.EditItem.cs
daividssilverio: I was aiming to make it similar to the iOS version of the code.
But I like your idea better.

daividssilverio:
@pdrj 

> The scrolling is not completely smooth.

I'm guessing this is related to the fact we are using 15 min snaps to scroll the view.
I will to limit the scrolling to see if it helps.


> If move the TE quickly, it can become almost not visible (disappears for a bit below the top or bottom bar).

> If I move the start down, when the end of TE is not visible on the screen, scroll won't be triggered, same goes with the end going up.

🤔 I'll see if we can improve this by maybe scrolling a bit before the user reaches the edges.

daividssilverio:
Hey @pdrj and @sofokli1 , please take another look.

> If move the TE quickly, it can become almost not visible (disappears for a bit below the top or bottom bar).

I've made the calendar view auto scroll so that the current item is visible when the user lifts the touch.

> If I move the start down, when the end of TE is not visible on the screen, scroll won't be triggered, same goes with the end going up.

I couldn't fix this for now. Also, rather fix this later if alright with @sofokli1 .

> The scrolling is not completely smooth.

I tried fixing this today and I couldn't come with a quick solution. I've tried adding an interpolator for the auto scroll but couldn't come with nice values and ended up with the events jumpy because of the way we are snapping it right now.

If @sofokli1 is alright with it, I rather create an issue for this and work on it later. It's gonna be time consuming to tune it just right.

daividssilverio:
I've created #6002 and #6003 to handle issues related to smooth auto scrolling and better visualization of items during auto scrolls.

PR #5979, opened 2019-09-27
Scroll to current hour of day (and indicator) on droid calendar

### LINKED ISSUE #5859: Calendar: Scroll to current time of day when opening the calendar view
Scroll to current time of day when opening the calendar view

daividssilverio:
## What's this?
This makes the calendar fragment scroll to the current hour of day on the calendar day view when the fragment opens.
It also adds the missing current hour indicator.

### Relationships
Closes #5859 

## Why do we want this?
To catch up with iOS

## How is it done?
With math and canvas.

## Review hints
🔍 by commit.

## :squid: Permissions
🦑 📆 team


Toggl.Droid/Views/Calendar/CalendarDayView.cs
```diff
@@ -303,6 +338,8 @@ private void scrollView(MotionEvent e1, MotionEvent e2, float deltaX, float delt
 
         private void flingView(MotionEvent e1, MotionEvent e2, float velocityX, float velocityY)
         {
+            if (editAction != EditAction.None) return;
```
heytherewill: y tho?

Toggl.Droid/Fragments/CalendarFragment.cs
```diff
@@ -93,7 +93,13 @@ public override void OnViewCreated(View view, Bundle savedInstanceState)
 
         public void ScrollToTop()
         {
-            //Todo: Scroll To Top
+            calendarDayView?.ScrollToTop();
```
heytherewill: I wonder whether this should scroll to the current hour or to the top. 

pinging @maritoggl for decisions. Context: What should happen when the user has the calendar tab selected and then clicks it again? All other views scroll to the top, but I think this one should scroll to the current time instead.

Toggl.Droid/Views/Calendar/CalendarDayView.cs
daividssilverio: Because it looks buggy if you try to fling while dragging the event around.
The frame would scroll if you drag fast and stop dragging without lifting your finger, causing a false fling.

Toggl.Droid/Fragments/CalendarFragment.cs
maritoggl: Go with current time.

daividssilverio: @sofokli1 ➿

PR #5982, opened 2019-09-27
Fix start/edit modals layout

### LINKED ISSUE #5531: Edit view cut off in iPad hover view

### LINKED ISSUE #5533: Start view resizes and gets cut off in iPad hover view

Odrakir:
## What's this?
This fixes a couple of issues with modals in iPad.

### Relationships
Closes #5531 and closes #5533 

## How is it done?
Basically using Size Classes instead of Interface idioms.

**THIS IS IMPORTANT** Always check for size classes (regular and compact) as opposed of devices (iPhone and iPad), because small split views and hover views in iPad should behave like iPhone in most cases.

My changes to how the modal is showing are breaking some other views in the app (duration edit for iPad hover views, for example). I'm going to go over those in another PR and change them to behave like they do on iPhone as well, this ignores some of the designs at Zeplin, but I think it makes much more sense. We can discuss it in another PR.

## :squid: Permissions
Please go!


dshr:
Just to double-check — this is only for the Start and Edit TE, all other modals, like Select Project/Tags are okay to be broken-ish?

Odrakir:
Yes, I have another PR ready for those.

PR #5991, opened 2019-09-30
Fix an autocomplete-autocorrect clash

### LINKED ISSUE #5662: Autocorrect and project/tag completion cause graphical glitch
## 🐛 Bug Report
If you search for a project in the start view, and they keyboard thinks that your search string should be corrected to something else, it will select the whole word (including the @) and that selection will still be visible if you then select a project.

Screenshot 2019-07-29 at 12 37 33

## Context

Build/Branch: 2.1.3 Appstore build (has been this way for months though)
Phone OS: 12.3 or so
Phone model: iPhone 6S

## Expected behavior
EITHER
1. the selection should go away after the project was selected
OR
2. it shouldn't even autocorrect when searching for tags/projects

## To reproduce
- be in the start view
- search for a project (very sure this happens for tags too)
- type something that results in a project suggestion, but also makes iOS want to correct your spelling
- select a project
- see the selection still visible until you continue typing

## Suggestions
Depending on the desired behaviour (and maybe technical feasibility) we should either:
1. clear the selection when selecting a project or tag
OR
2. disable auto correction if we're searching for projects/tags (and only then, we do want it on otherwise)

I _think_ we should go with 1. because we want users to be able to have correctly spelled project names when creating new projects, but maybe @maritoggl should weigh in.

dshr:
## What's this?
A slightly unsightly fix to a little clash we had between our suggestions and iOS' autocomplete.

### Relationships
Closes #5662

## Why do we want this?
So things look spotless!

## How is it done?
By walking across all the subviews of `DescriptionTextView`, finding the offending `UIAutocorrectInlinePrompt` (which is defined in a runtime-only header) by classname, and removing it from the `DescriptionTextView`.

### Why not another way?
Looked at other options, but due to the runtime-only nature of `UIAutocorrectInlinePrompt` I can't come up with anything.

## Review hints
Reproduce the error from the source issue on `develop`, then here.

## :squid: Permissions
All yours!

### DIFF
--- a/Toggl.iOS/Extensions/TextViewExtensions.cs
+++ b/Toggl.iOS/Extensions/TextViewExtensions.cs
@@ -1,0 +1,28 @@
+using CoreGraphics;
+using Foundation;
+using UIKit;
+
+namespace Toggl.iOS.Extensions
+{
+    public static class TextViewExtensions
+    {
+        public static void rejectAutocorrect(this UITextView textView)
+        {
+            var originalText = textView.Text;
+            NSRange originalRange = textView.SelectedRange;
+            CGPoint originalOffset = textView.ContentOffset;
+
+            //Force any pending autocorrection to be applied
+            textView.ResignFirstResponder();
+            textView.BecomeFirstResponder();
+
+            string finalText = textView.Text;
+            if (originalText != finalText)
+            {
+                textView.Text = originalText;
+                textView.SelectedRange = originalRange;
+                textView.SetContentOffset(originalOffset, false);
+            }
+        }
+    }
+}
--- a/Toggl.iOS/Extensions/TextViewExtensions.cs
+++ b/Toggl.iOS/Extensions/TextViewExtensions.cs
@@ -6,7 +6,7 @@ namespace Toggl.iOS.Extensions
 {
     public static class TextViewExtensions
     {
-        public static void rejectAutocorrect(this UITextView textView)
+        public static void RejectAutocorrect(this UITextView textView)
         {
             var originalText = textView.Text;
             NSRange originalRange = textView.SelectedRange;
--- a/Toggl.iOS/ViewControllers/StartTimeEntryViewController.cs
+++ b/Toggl.iOS/ViewControllers/StartTimeEntryViewController.cs
@@ -26,7 +26,6 @@ namespace Toggl.iOS.ViewControllers
     public sealed partial class StartTimeEntryViewController : KeyboardAwareViewController<StartTimeEntryViewModel>
     {
         private const double desiredIpadHeight = 360;
-        private const string autocorrectPromptClassName = "UIAutocorrectInlinePrompt";
 
         private bool isUpdatingDescriptionField;
 
@@ -213,16 +212,7 @@ namespace Toggl.iOS.ViewControllers
             DescriptionTextView.InputDelegate = emptyInputDelegate;
             DescriptionTextView.AttributedText = attributedText;
 
-            // This line is needed for when the user selects from suggestion and
-            // the iOS autocorrect is ready to add text at the same time too.
-            // Without this the little selection will persist even after the
-            // suggestion has been selected
-            foreach (var view in DescriptionTextView.Subviews)
-                if (view.Class.Name == autocorrectPromptClassName)
-                {
-                    view.RemoveFromSuperview();
-                    break;
-                }
+            DescriptionTextView.rejectAutocorrect();
 
             updatePlaceholder();
         }
--- a/Toggl.iOS/ViewControllers/StartTimeEntryViewController.cs
+++ b/Toggl.iOS/ViewControllers/StartTimeEntryViewController.cs
@@ -26,6 +26,7 @@ namespace Toggl.iOS.ViewControllers
     public sealed partial class StartTimeEntryViewController : KeyboardAwareViewController<StartTimeEntryViewModel>
     {
         private const double desiredIpadHeight = 360;
+        private const string autocorrectPromptClassName = "UIAutocorrectInlinePrompt";
 
         private bool isUpdatingDescriptionField;
 
@@ -217,7 +218,7 @@ namespace Toggl.iOS.ViewControllers
             // Without this the little selection will persist even after the
             // suggestion has been selected
             foreach (var view in DescriptionTextView.Subviews)
-                if (view.Class.Name == "UIAutocorrectInlinePrompt")
+                if (view.Class.Name == autocorrectPromptClassName)
                 {
                     view.RemoveFromSuperview();
                     break;
--- a/Toggl.iOS/ViewControllers/StartTimeEntryViewController.cs
+++ b/Toggl.iOS/ViewControllers/StartTimeEntryViewController.cs
@@ -212,7 +212,7 @@ namespace Toggl.iOS.ViewControllers
             DescriptionTextView.InputDelegate = emptyInputDelegate;
             DescriptionTextView.AttributedText = attributedText;
 
-            DescriptionTextView.rejectAutocorrect();
+            DescriptionTextView.RejectAutocorrect();
 
             updatePlaceholder();
         }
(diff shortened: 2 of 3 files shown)


Toggl.iOS/ViewControllers/StartTimeEntryViewController.cs
```diff
@@ -209,6 +209,10 @@ private void onTextFieldInfo(TextFieldInfo textFieldInfo)
             DescriptionTextView.InputDelegate = emptyInputDelegate; //This line is needed for when the user selects from suggestion and the iOS autocorrect is ready to add text at the same time. Without this line both will happen.
             DescriptionTextView.AttributedText = attributedText;
 
+            foreach (UIView view in DescriptionTextView.Subviews) // This line is needed for when the user selects from suggestion and the iOS autocorrect is ready to add text at the same time too. Without this the little selection will persist even after the suggestion has been selected
+                if (view.Class.Name == "UIAutocorrectInlinePrompt")
```
heytherewill: There's no constant for this? I don't like this magic string

Toggl.iOS/ViewControllers/StartTimeEntryViewController.cs
```diff
@@ -209,6 +209,10 @@ private void onTextFieldInfo(TextFieldInfo textFieldInfo)
             DescriptionTextView.InputDelegate = emptyInputDelegate; //This line is needed for when the user selects from suggestion and the iOS autocorrect is ready to add text at the same time. Without this line both will happen.
             DescriptionTextView.AttributedText = attributedText;
 
+            foreach (UIView view in DescriptionTextView.Subviews) // This line is needed for when the user selects from suggestion and the iOS autocorrect is ready to add text at the same time too. Without this the little selection will persist even after the suggestion has been selected
```
heytherewill: Please put the comment above the for each and make it multiline.

Toggl.iOS/ViewControllers/StartTimeEntryViewController.cs
```diff
@@ -209,6 +209,10 @@ private void onTextFieldInfo(TextFieldInfo textFieldInfo)
...
+                if (view.Class.Name == "UIAutocorrectInlinePrompt")
+                    view.RemoveFromSuperview();
```
heytherewill: Can you break the loop as soon as you find the view or there might be multiple views?

Toggl.iOS/ViewControllers/StartTimeEntryViewController.cs
heytherewill: Also `var view`

Toggl.iOS/ViewControllers/StartTimeEntryViewController.cs
```diff
@@ -206,9 +207,23 @@ private void onTextFieldInfo(TextFieldInfo textFieldInfo)
...
+            // the iOS autocorrect is ready to add text at the same time.
+            // Without this line both will happen.
+            DescriptionTextView.InputDelegate = emptyInputDelegate;
             DescriptionTextView.AttributedText = attributedText;
 
+            // This line is needed for when the user selects from suggestion and
+            // the iOS autocorrect is ready to add text at the same time too.
+            // Without this the little selection will persist even after the
+            // suggestion has been selected
+            foreach (var view in DescriptionTextView.Subviews)
```
Odrakir: Please, please, please. Use `{` for this `foreach`, even if the `if` is just one statement, it's multiple lines and it's really confusing.

Toggl.iOS/ViewControllers/StartTimeEntryViewController.cs
```diff
@@ -206,9 +207,23 @@ private void onTextFieldInfo(TextFieldInfo textFieldInfo)
...
+                if (view.Class.Name == autocorrectPromptClassName)
+                {
+                    view.RemoveFromSuperview();
```
Odrakir: The problem with this approach is, once you remove that view, it's never going back, so we are not getting autocorrect indicators for that textfield any more.

Odrakir: You would have to use `view.Hidden = true`, but then you would have to unhide it later, maybe when the user starts typing?

Toggl.iOS/ViewControllers/StartTimeEntryViewController.cs
```diff
@@ -26,6 +26,7 @@ namespace Toggl.iOS.ViewControllers
     public sealed partial class StartTimeEntryViewController : KeyboardAwareViewController<StartTimeEntryViewModel>
     {
         private const double desiredIpadHeight = 360;
+        private const string autocorrectPromptClassName = "UIAutocorrectInlinePrompt";
```
Odrakir: This seems really fragile...

Toggl.iOS/ViewControllers/StartTimeEntryViewController.cs
kornelijepetak: Yes, please!

Toggl.iOS/ViewControllers/StartTimeEntryViewController.cs
dshr: Nah, the view is created dynamically when there's an autocomplete situation:
autocomplete

Odrakir: 🤷‍♂ For some reason that wasn't working for me. But anyway, did you see my comment with the alternative approach?

dshr: Yeah, just tried it, it works, albeit with the jumps. I'm not sure we can get rid of the jumps though — we're effectively just hiding and showing the KB real quick, so a jump is bound to happen.

Odrakir: I think the jump is happening in the tools view over the keyboard, not the keyboard itself, and it's probably due to how we animate those modals in the case of a keyboard notification. I think this might get fixed when we re-do the modals.

But I, personally, would rather have the jump for a bit than the looking for the `UIAutocorrectInlinePrompt` named view, which is more prone to failing when Apple decides to change something.

dshr: Yeah, that's what I meant — we'd have to add a way to stop the animations, then reject the autocomplete, then reenable them. I've gone with your solution then, since I generally feel off about the string lookup too.

Toggl.iOS/Extensions/TextViewExtensions.cs
```diff
@@ -0,0 +1,28 @@
+using CoreGraphics;
+using Foundation;
+using UIKit;
+
+namespace Toggl.iOS.Extensions
+{
+    public static class TextViewExtensions
+    {
+        public static void rejectAutocorrect(this UITextView textView)
```
simonrozsival: This is a public method, it should be capitalized - `RejectAutocorrect`.

Toggl.iOS/ViewControllers/StartTimeEntryViewController.cs
```diff
@@ -206,9 +206,14 @@ private void onTextFieldInfo(TextFieldInfo textFieldInfo)
...
                 return;
 
-            DescriptionTextView.InputDelegate = emptyInputDelegate; //This line is needed for when the user selects from suggestion and the iOS autocorrect is ready to add text at the same time. Without this line both will happen.
+            //This line is needed for when the user selects from suggestion and
+            // the iOS autocorrect is ready to add text at the same time.
+            // Without this line both will happen.
+            DescriptionTextView.InputDelegate = emptyInputDelegate;
             DescriptionTextView.AttributedText = attributedText;
 
+            DescriptionTextView.rejectAutocorrect();
```
heytherewill:
```suggestion
            DescriptionTextView.RejectAutocorrect();
```

dshr: Yeah, maybe the online editor is not that great sometimes 🙈

Issue #5973, opened 2019-09-26
Calendar: Top toolbar improvements


sofokli1:
Set the date and the total duration of the day as in the design below.

- [ ] Label at the top of the screen will show selected date with format "EEEE, MMM d"
- [ ] Daily tracked hours on the left

Design: https://app.zeplin.io/project/5728ce853dd1b182475f25e9/screen/5d8bbabe1b729d1611ddd55d

PR #5992, opened 2019-09-30
Calendar: add swipe-ability between days

### LINKED ISSUE #5853: Calendar: Non-confusing settings button
Settings button will be in the top right corner of the calendar view (same as in main log).

It leads to general settings (not calendar settings any more)

(https://app.zeplin.io/project/5728ce853dd1b182475f25e9/screen/5b34f15671d2e4ef0dedd0e3) positioning if easily possible

### LINKED ISSUE #5863: Calendar: Support changing date by swiping
Ability to swipe left and right to move a single day back and forth

your mission should you choose to accept, will be:
- [x] add ability to swipe left and right
- [x] add needed VM logic (be aware that will be used by android too so feel free to talk with your project members dev if in doubt when implementing)
- [x] make sure it can only swipe for 2 weeks (starting from today and back). it should just bounce when trying to swipe to a date outside the 2 weeks range.
- [x] make sure it works on iPad
- [x] make sure there is enough VM logic so that the weekly date picker can be updated based on the swipe of the main VM and back (this is optional in order to complete this issue since it will be worked again in #5865 )

Feel free to copy this list in the PR so that the reviewer knows what to check

arturs105:
## What's this?
This PR:
* adds swiping to the calendar
* adds a placeholder for #5865
* implements the current date part of #5973

### Relationships

Closes #5863 
Closes #5853

## How is it done?
In the foundation layer we now have a `CalendarViewModel` and `CalendarDayViewModel`. The former is like an overall view model, which contains the current date, daily tracked time and a bunch of `CalendarDayViewModel`s. And the `CalendarDayViewModel ` represents, well, a day in the calendar - it is what `CalendarViewModel` used to be - a collection of calendar events displayed in a vertical scroll view.

In the iOS layer we have the `CalendarViewController`, which is the overall view controller, that controls the current date label, daily tracked time label, and it has a nested `UIPageViewController`, which shows a bunch of `CalendarDayViewControllers`.

## :squid: Permissions
🦑


Toggl.Core.UI/ViewModels/Calendar/CalendarDayViewModel.cs
```diff
@@ -0,0 +1,211 @@
...
+            this.analyticsService = analyticsService;
+            this.interactorFactory = interactorFactory;
+
+            OnItemTapped = rxActionFactory.FromAsync<CalendarItem>(handleCalendarItem);
+            OnUpdateTimeEntry = rxActionFactory.FromAsync<CalendarItem>(updateTimeEntry);
+            OnCalendarEventLongPressed = rxActionFactory.FromAsync<CalendarItem>(handleCalendarEventLongPressed);
+            OnDurationSelected = rxActionFactory.FromAsync<(DateTimeOffset StartTime, TimeSpan Duration)>(
+                tuple => durationSelected(tuple.StartTime, tuple.Duration));
+
+
```
daividssilverio: 🔥 Extra space.

Toggl.Core.UI/ViewModels/Calendar/CalendarDayViewModel.cs
```diff
@@ -0,0 +1,211 @@
...
+            var preferences = dataSource.Preferences.Current;
+            TimeOfDayFormat = preferences
+                .Select(current => current.TimeOfDayFormat)
+                .AsDriver(schedulerProvider);
+
+            CalendarItems = new ObservableGroupedOrderedCollection<CalendarItem>(
```
daividssilverio: The `ObservableGroupedOrderedCollection` has been deprecated, right? Shouldn't we use the `SectionModel` or `AnimatableSectionModel` now?

Toggl.Core.UI/ViewModels/Calendar/CalendarViewModel.cs
```diff
@@ -36,388 +24,77 @@ public sealed class CalendarViewModel : ViewModel
...
-
-        public IObservable<string> TimeTrackedToday { get; }
-
-        public IObservable<string> CurrentDate { get; }
-
-        public UIAction GetStarted { get; }
-
-        public UIAction SkipOnboarding { get; }
-
-        public UIAction LinkCalendars { get; }
```
daividssilverio: Are we removing the onboarding completely? @sofokli1 
As far as I understood, I thought at least the first placeholder page of the calendar onboarding would stay intact.

Toggl.iOS/ViewControllers/Calendar/CalendarViewController.cs
```diff
@@ -1,185 +1,101 @@
...
-                .Subscribe(ViewModel.OnDurationSelected.Inputs)
-                .DisposedBy(DisposeBag);
+        public UIViewController GetPreviousViewController(UIPageViewController pageViewController, UIViewController referenceViewController)
+        {
+            var referenceTag = referenceViewController.View.Tag;
+            if (referenceTag == minAllowedPageIndex)
+                return null;
 
-            CalendarCollectionView.LayoutIfNeeded();
+            return viewControllerAtIndex(referenceTag -1);
```
daividssilverio:
```suggestion
            return viewControllerAtIndex(referenceTag - 1);
```

Toggl.Core.UI/ViewModels/Calendar/CalendarDayViewModel.cs
arturs105: Yes, i noticed that it was depreciated, however, this is how the calendar view was implemented and i didn't want to refactor it - i just copy+pasted the existing code to make swipes possible.

Toggl.Core.UI/ViewModels/Calendar/CalendarViewModel.cs
arturs105: I understood that all the onboarding would be removed. And, to be honest, i got excited about deleting so. much. code. 😄

arturs105: And yes, summoning @sofokli1

arturs105:
> I didn't get it from the code, but is your pager thingy actually loading the next/previous views before they appear in screen?

Yes - they are loaded _right_ before the swipe (or as the swipe is being started). That happens in `CalendarViewController. GetPreviousViewController` and `GetNextViewController`.

Not sure, how droid works, so this is something we might need to change. We could add some view model caching with a list of view models, or whatever is necessary for Droid.

Toggl.Core.UI/ViewModels/Calendar/CalendarViewModel.cs
sofokli1: I understood it that we will be removing it completely but @maritoggl can clarify for us

Toggl.Core.UI/ViewModels/Calendar/CalendarDayViewModel.cs
Odrakir: Oh yeah, no please, don't keep using the `ObservableGroupedWhateverBlaBla`, it's evil and we should get rid of it as soon as possible.

Odrakir: And I should know, because I wrote it. Using `ISectionModel` is much, much easier.

daividssilverio:
By the way. I'm ok with how it works right now. It should be ok to use it Android and we can keep the local cache platform specific.

Toggl.Core.UI/ViewModels/Calendar/CalendarViewModel.cs
maritoggl: We are removing the current onboarding screen, yes, so that users would see the time entries in the calendar view straight away. With that said, we'll revisit the onboarding tips etc in a separate future project.

daividssilverio:
Hey @arturs105 I just noticed that there's no equivalent of `TimeTrackedToday` from the old view model in the new `CalendarDayViewModel`. We should be showing the total duration of the time entries in the currently selected day.

Since we have `CurrentlyShownDateString`, might as well have an observable for durations as well, right ?

arturs105:
@sofokli1 
> I see the current time blue line in all days, I think that should be visible only in the todays view.

My thoughts exactly, but the designs have the blue line in all days, so i didn't bother with that. But like i said, i agree that it should only be shown on today. Summoning @maritoggl to clear things up.

@daividssilverio 
> Hey @arturs105 I just noticed that there's no equivalent of TimeTrackedToday from the old view model in the new CalendarDayViewModel. We should be showing the total duration of the time entries in the currently selected day.

I left that out on purpose, since there's an issue for that: #5973. The issue also mentions displaying the current time, which i did implement, but that's only because i noticed the issue after i had already implemented it.

Toggl.Core.UI/ViewModels/Calendar/CalendarDayViewModel.cs
```diff
@@ -0,0 +1,210 @@
...
+    {
+        private readonly ITimeService timeService;
+        private readonly IAnalyticsService analyticsService;
+        private readonly IInteractorFactory interactorFactory;
+
+        private readonly CompositeDisposable disposeBag = new CompositeDisposable();
+
+        public DateTimeOffset Date { get; }
+
+        public ObservableGroupedOrderedCollection<CalendarItem> CalendarItems { get; }
```
Odrakir: We should not use this, it's deprecated (and only used by the calendar rebels now). Instead we should be using `ISectionModel` and some extensions that use diffing.

Toggl.Core.UI/ViewModels/Calendar/CalendarDayViewModel.cs
```diff
@@ -0,0 +1,210 @@
...
+            Ensure.Argument.IsNotNull(schedulerProvider, nameof(schedulerProvider));
+            Ensure.Argument.IsNotNull(interactorFactory, nameof(interactorFactory));
+
+            Date = date;
+            this.timeService = timeService;
+            this.analyticsService = analyticsService;
+            this.interactorFactory = interactorFactory;
+
+            OnItemTapped = rxActionFactory.FromAsync<CalendarItem>(handleCalendarItem);
+            OnUpdateTimeEntry = rxActionFactory.FromAsync<CalendarItem>(updateTimeEntry);
```
Odrakir: The name of this action doesn't match the rest, which directly represent user actions... Maybe something like `OnTimeEntryEdited`? Feel free to ignore.

Toggl.Core.UI/ViewModels/Calendar/CalendarDayViewModel.cs
```diff
@@ -0,0 +1,210 @@
...
+            TimeOfDayFormat = preferences
+                .Select(current => current.TimeOfDayFormat)
+                .AsDriver(schedulerProvider);
+
+            CalendarItems = new ObservableGroupedOrderedCollection<CalendarItem>(
+                indexKey: item => item.StartTime,
+                orderingKey: item => item.StartTime,
+                groupingKey: _ => 0);
+
+            var selectedCalendarsChangedObservable = userPreferences
```
Odrakir: I might be wrong, but didn't we agree on not naming every Observable `whateverObservable` and name it only `whatever`?

Toggl.Core.UI/ViewModels/Calendar/CalendarDayViewModel.cs
```diff
@@ -0,0 +1,210 @@
...
+            var appResumedFromBackgroundObservable = backgroundService
+                .AppResumedFromBackground
+                .SelectUnit();
+
+            dataSource.TimeEntries
+                .ItemsChanged()
+                .Merge(selectedCalendarsChangedObservable)
+                .Merge(appResumedFromBackgroundObservable)
+                .SubscribeOn(schedulerProvider.BackgroundScheduler)
+                .ObserveOn(schedulerProvider.BackgroundScheduler)
```
Odrakir: What are you trying to do here? Are you sure you need both of those?

Toggl.Core.UI/ViewModels/Calendar/CalendarDayViewModel.cs
```diff
@@ -0,0 +1,210 @@
...
+                .SelectUnit();
+
+            var appResumedFromBackgroundObservable = backgroundService
+                .AppResumedFromBackground
+                .SelectUnit();
+
+            dataSource.TimeEntries
+                .ItemsChanged()
+                .Merge(selectedCalendarsChangedObservable)
+                .Merge(appResumedFromBackgroundObservable)
```
Odrakir: This is personal taste, but to me intention is much more clear this way:
```c#
Observable.Merge(
        dataSource.TimeEntries.ItemsChanged(),
        selectedCalendarsChangedObservable,
        appResumedFromBackgroundObservable
    )
```

Toggl.Core.UI/ViewModels/Calendar/CalendarViewModel.cs
```diff
@@ -36,388 +24,77 @@ public sealed class CalendarViewModel : ViewModel
...
-            TimeTrackedToday = timeTrackedToday
-                .StartWith(TimeSpan.Zero)
-                .CombineLatest(durationFormat, DurationAndFormatToString.Convert)
-                .AsDriver(schedulerProvider);
-
-            CurrentDate = timeService.CurrentDateTimeObservable
-                .Select(dateTime => dateTime.ToLocalTime().Date)
+            var dateFormatObservable = dataSource.Preferences.Current
+                .Select(current => current.DateFormat);
+            CurrentlyShownDateString = CurrentlyVisiblePage.AsObservable()
```
Odrakir: Add a new line.

Toggl.iOS/ViewControllers/Calendar/CalendarViewController.cs
```diff
@@ -1,185 +1,101 @@
...
-            zoomHelper = new CalendarCollectionViewZoomHelper(CalendarCollectionView, layout);
-
-            CalendarCollectionView.SetCollectionViewLayout(layout, false);
-            CalendarCollectionView.Delegate = dataSource;
-            CalendarCollectionView.DataSource = dataSource;
-            CalendarCollectionView.ContentInset = new UIEdgeInsets(20, 0, 20, 0);
-
-            dataSource.ItemTapped
-                .Subscribe(ViewModel.OnItemTapped.Inputs)
+            SettingsButton.Rx()
```
Odrakir: The settings button is not positioned in the same place as in the timeline, so it's a little weird when you switch tabs.

Also, if this is going to be a general button I think it should be in Reports as well.

My advice is to embed each of these screens in a `UINavigationController` and add the button as a `UIBarButtonItem`.

Toggl.iOS/ViewControllers/Calendar/CalendarViewController.cs
```diff
@@ -1,185 +1,101 @@
...
-                    - CalendarCollectionView.ContentInset.Top
-                    - CalendarCollectionView.ContentInset.Bottom;
-            var hoursOnScreen = frameHeight / (CalendarCollectionView.ContentSize.Height / 24);
-            var centeredHour = calculateCenteredHour(timeOfDay.TotalHours, hoursOnScreen);
-
-            var offsetY = (centeredHour / 24) * CalendarCollectionView.ContentSize.Height - (frameHeight / 2);
-            var scrollPointY = offsetY.Clamp(0, CalendarCollectionView.ContentSize.Height - frameHeight);
-            var offset = new CGPoint(0, scrollPointY);
-            CalendarCollectionView.SetContentOffset(offset, false);
+            var date = DateTimeOffset.Now.AddDays(index);
```
Odrakir: The date should come from the ViewModel

Toggl.Core.UI/ViewModels/Calendar/CalendarDayViewModel.cs
Odrakir: Oh, also, this never goes back to the main thread... 🤔

Toggl.Core.UI/ViewModels/Calendar/CalendarDayViewModel.cs
sofokli1: This was discussed and we agreed to keep it as is for now due to lack of time (also refactoring is not the goal of the project). We also have an issue in at the end of the project to try and refactor this with `ISectionModel`.

Toggl.iOS/ViewControllers/Calendar/CalendarViewController.cs
sofokli1: It was agree that reports view will not be touched during this project.

The position of the settings button between log and calendar view tabs should be fixed.

Toggl.Core.UI/ViewModels/Calendar/CalendarDayViewModel.cs
Odrakir: Understood, let's hope we can get rid of that soon 🤞

Toggl.iOS/ViewControllers/Calendar/CalendarViewController.cs
arturs105: The `UINavigationController` will be added as a part of #5865.

Toggl.iOS/ViewControllers/Calendar/CalendarViewController.cs
```diff
@@ -81,7 +81,8 @@ public void WillTransition(UIPageViewController pageViewController, UIViewContro
 
         private CalendarDayViewController viewControllerAtIndex(nint index)
         {
-            var date = DateTimeOffset.Now.AddDays(index);
+            var now = IosDependencyContainer.Instance.TimeService.CurrentDateTime.ToLocalTime().Date;
```
Odrakir: But why not though the ViewModel? I'm guessing it could even just return the DayViewModel for a specific index. Doing it like this hides the dependency making it's harder to test and debug.

How about `var viewModel = ViewModel.DayViewModelAt(index)`

Toggl.Core.UI/ViewModels/Calendar/CalendarDayViewModel.cs
```diff
@@ -83,21 +83,18 @@ public sealed class CalendarDayViewModel : ViewModel
...
 
-            dataSource.TimeEntries
-                .ItemsChanged()
-                .Merge(selectedCalendarsChangedObservable)
-                .Merge(appResumedFromBackgroundObservable)
-                .SubscribeOn(schedulerProvider.BackgroundScheduler)
+            Observable.Merge(dataSource.TimeEntries.ItemsChanged(), selectedCalendarsChanged, appResumedFromBackground)
                 .ObserveOn(schedulerProvider.BackgroundScheduler)
                 .SelectMany(_ => reloadData())
+                .SubscribeOn(schedulerProvider.MainScheduler)
```
Odrakir: Nope, this is not doing what you want it to do. Please read: http://reactivex.io/documentation/operators/observeon.html

arturs105: Confusing names are confusing. So let me see if i get this correctly. `SubscribeOn` moves the _whole_ observable operator sequence to the specified thread. And it can be overrided by a `ObserveOn` later in the chain. So could we do `SubscribeOn(backgroundScehduler)` and then bring it back to main thread by `ObserveOn(mainScheduler)` right before `Subscribe` is called?

Odrakir: Yes, but with the caveat that `SubscribeOn` will also change the thread in which the closure that generates the events works, so that might be delicate. Maybe just use `ObserveOn` twice, once to send it to a background thread (in case it's not there already) and then again to bring it back to the main thread (which is something that AsDriver usually takes care of, but not in this case)

arturs105: The `AsDriver` doesn't take care of this because it simply isn't there? Or is there some other reason why it wouldn't work?  If it would work could i just omit the second `ObserveOn` and use `AsDriver?`.

Odrakir: We use Drivers to expose properties from the VMs to the views, which is not what you are doing here, so it doesn't make sense to use it. 

You could use it when you move this code into using the new `ISectionModel`, though.

PR #5993, opened 2019-09-30
Start and stop time entries from the Android widget

### LINKED ISSUE #5094: Handle start and stop actions using intents
Send an intent from the widget buttons to start stop last TE.
Handle those intents from the widget provider.

Starting TEs from suggestions will be handled in a different issue.

Depends on #5092

juanlaube:
## What's this?
This handles the start and stop actions when pressing the buttons in the widget.

### Relationships

Closes #5094 

## Why do we want this?
We want to start and stop time entries from the widget.

## How is it done?
The buttons send a pending intent on click handled by the `AppWidgetProvider` which is a broadcast receiver, then a service is started to start and stop the time entries.

### Why not another way?
Simply starting the service with a pending intent doesn't work after Oreo when the app is not running, we could have used a foreground service but that puts a silent notification in the status bar and it's ugly.

### Side effects
The `TimerBackgroundService` can be used after by other Android integrations, e.g: the enhanced notifications we need to discuss.

## Review hints
Try it! Make sure to try with the app in background and not started.
Notice that the pending intents are set only when the widget is updated, so resize it first (that should be fixed in another PR).

## :squid: Permissions
🦑


Toggl.Droid/Services/TimerBackgroundService.cs
```diff
@@ -0,0 +1,72 @@
...
+        protected override void OnHandleWork(Intent intent)
+        {
+            var action = intent.Action;
+
+            AndroidDependencyContainer.EnsureInitialized(Application.Context);
+
+            switch (action)
+            {
+                case StartTimeEntryAction:
+                    handleStartTimeEntry();
```
kornelijepetak: Discard this result, to avoid the async in non-async method warnings and to signify that you are aware that the task is running async.

```suggestion
                    _ = handleStartTimeEntry();
```

Toggl.Droid/Services/TimerBackgroundService.cs
```diff
@@ -0,0 +1,72 @@
...
+                    break;
+                case StopRunningTimeEntryAction:
+                    handleStopRunningTimeEntry();
```
kornelijepetak: Discard this result, to avoid the async in non-async method warnings and to signify that you are aware that the task is running async.
```suggestion
                    _ = handleStopRunningTimeEntry();
```

Toggl.Droid/Widgets/TimeEntryWidget.cs
```diff
@@ -12,6 +13,21 @@ namespace Toggl.Droid.Widgets
...
     public class TimeEntryWidget : AppWidgetProvider
     {
+        public override void OnReceive(Context context, Intent intent)
+        {
+            base.OnReceive(context, intent);
+
+            var action = intent.Action;
+            if (action == TimerBackgroundService.StartTimeEntryAction)
+            {
+                TimerBackgroundService.EnqueueStartTimeEntry(context, intent);
```
kornelijepetak: I don't know if I'm doing something wrong, but this is never called for me.

`OnReceive` gets called on widget install/remove etc, but pressing the button does nothing.

Also, `getRemoteViews` on line `:62` is never called for me either after adding a widget, which is of course the cause for the problem above (click intent never registered).

However, after I resize the widget, then it starts responding to clicks. But even that only invisibly. I can see created entries in the app, but the widget does not change visually. I can't see the Stop button. Instead, multiple taps on the are all `Continue` action...

kornelijepetak: Ah, of course I was more interested in code than reading your PR description in full. My bad! 😛 There go half an hour of testing my environment. 🤷‍♂

juanlaube: 😄

PR #5994, opened 2019-10-01
Remove UIUserInterfaceIdiom checks

Odrakir:
## What's this?
This changes most of the layout checks we were doing to differentiate iPad/iPhone. This was done following the designs at Zepplin, but I think those are wrong, so I'm changing them. More info below.

### Relationships
Doesn't close any issue.

## Why do we want this?
Using `UIUserInterfaceIdiom` instead of size classes makes layout harder, mostly in the case of iPad, where in a single device, you can have different widths. The way this is meant to be done is with size classes, which tells us if the content is being displayed in a compact or regular width window.

See [Multitasking Size Classes](https://developer.apple.com/design/human-interface-guidelines/ios/visual-design/adaptivity-and-layout/)
image

As you can see, there's no Regular width size for portrait mode, no matter the iPad model. This changes a little bit with respect to what's in Zepplin. For example, in Zeppelin we can find this:
Captura de pantalla 2019-10-01 a las 10 12 46
Which makes no sense, because that 1/2 layout is just not possible in portrait mode. But the closer thing you can have will look like this, in compact width mode.
image

## How is it done?
Basically going over every `UIUserInterfaceIdiom.` and changing it to the equivalent `TraitCollection`.

Then I also changed how some of the modals behave in terms of moving when the keyboard shows and so on. The idea is not to resize them when the keyboard shows, but to move its frame just so the first responder is visible.

Then I also made it so the keyboard hides when the user scrolls.

### Why not another way?
This is the best way to do multidevice layout and it's how Apple intended us to do it.

## Review hints
It's a lot of commits, and a lot of files. I know. But if you go one commit at a time, you'll find out most of them are the same thing just for different screens.

## :squid: Permissions
This is ready. Go!


Odrakir:
I agree those things should be fixed, but that doesn't mean we don't follow guidelines. The way to do this is: First we use size classes and then we look at specific cases. As a matter of fact I think those two are easily fixed, both have problems with left over code from the previous approach. I'll fix them right away.

On the topic on following guidelines, I think we should follow them as much as possible, and then break them, but only when we are aware that we are doing it. In my opinion whoever designed this wasn't aware of the way size classes or even multitasking works on iPad.

Odrakir:
Those two are fixed.

I think the date selector is ok-ish now, in the future, when (if) we remodel those screens, I would very much like to make the date picker inline (and probably the color picker too)

PR #6000, opened 2019-10-01
Track widget resize

### LINKED ISSUE #5098: Track widget resize
Detect when the widget is resized and track the necessary events.

Add a new event, `TimerWidgetSizeChanged` and pass the new size as a parameter.

We can get notified by the platform when a widget is resized by listening to some breadcast intents, here's more info about that: https://developer.android.com/guide/topics/appwidgets#ProviderBroadcasts

Depends on #5092 and #5404

juanlaube:
## What's this?
Tracks the widget resizing in Android.

### Relationships

Closes #5098 

## Why do we want this?
📈 We want to understand how users use the widget.

## How is it done?
Similar to how we track installations.

### Side effects
None.

## :squid: Permissions
🦑


Toggl.Droid/Widgets/TimeEntryWidget.cs
```diff
@@ -84,5 +91,8 @@ private PendingIntent stopTimeEntryPendingIntent(Context context)
             intent.SetAction(TimerBackgroundService.StopRunningTimeEntryAction);
             return PendingIntent.GetBroadcast(context, 0, intent, PendingIntentFlags.UpdateCurrent);
         }
+
+        private int getColumns(int width)
```
kornelijepetak: Can you please move the comment into a `///` comments above the method, in the `<remarks> tag`?

Also, rename this into `getColumnsCount` because what you get from this method is the *count*, not the *columns*.

Toggl.Droid/Services/WidgetsAnalyticsService.cs
```diff
@@ -0,0 +1,61 @@
...
+using Android.Runtime;
+using Android.Support.V4.App;
+using static Toggl.Droid.Services.JobServicesConstants;
+
+namespace Toggl.Droid.Widgets.Services
+{
+    [Service(Permission = "android.permission.BIND_JOB_SERVICE", Exported = true)]
+    class WidgetsAnalyticsService : JobIntentService
+    {
+        public const string TrackTimerWidgetInstallAction = nameof(TrackTimerWidgetInstallAction);
```
kornelijepetak: I'd rather group these by purpose, not the type.

PR #6006, opened 2019-10-01
Save running time entry data in shared storage

### LINKED ISSUE #5087: Observe and persist current running time entry in SharedStorage
Observe the running time entry and persist the necessary information to populate the widget in ExtensionKit.SharedStorage.

Depends on #5085

juanlaube:
## What's this?
Saves the running time entry data in the shared storage so the widget can use it to populate the UI.

### Relationships

Closes #5087 

## Why do we want this?
To show the running time entry in the widget.

## How is it done?
Observing the running time entry and saving what we need in the shared user default.
I added a few methods and some helper extensions to save the time as a dictionary because we will add continue saving more data in milestone 3 for the suggestions.

### Why not another way?
🤷‍♂ 

### Side effects
None.

## :squid: Permissions
🦑


Toggl.iOS.ExtensionKit/SharedStorage.cs
```diff
@@ -120,6 +129,35 @@ public void SetWidgetInstalled(bool installed)
...
         public bool GetWidgetInstalled() => userDefaults.BoolForKey(widgetInstalledKey);
 
+        public void SetRunningTimeEntry(
+            long timeEntryId,
+            string description,
+            DateTimeOffset startTime,
+            string projectName,
+            string projectColor,
+            string taskName,
+            string clientName)
```
simonrozsival: Couldn't this just take `ITimeEntry` directly instead of all those arguments? It is a very minor thing, but it would make the interface much nicer. I also wanted to try multi-line comments. So feel free to dismiss this 😄

juanlaube: That was my first thought too, but `ITimeEntry` doesn't know about the project, task and client properties we need, only their ids.

Odrakir: My ears were ringing. There’s this for the Siri extension. Maybe we should refactor it and move it to the extension kit. https://github.com/toggl/mobileapp/blob/develop/Toggl.iOS.SiriExtension/Models/TimeEntry.cs

juanlaube: It takes `ITimeEntry` now.

PR #6010, opened 2019-10-02
Remove ARMv7 support for iOS

simonrozsival:
## What's this?
After updating to the newest version of Xcode and Xamarin, I was not able to build the app in Visual Studio locally or in Bitrise and I got this message:

```
MTOUCH : error MT0116: Invalid architecture: ARMv7. 32-bit architectures are not supported when deployment target is 11 or later. [/Users/vagrant/git/Toggl.iOS/Toggl.iOS.csproj]
```

### Relationships
I haven't created an issue for this problem.

## Why do we want this?
We want to be able to build and deploy our app.

## How is it done?
I got rid of `ARMv7` and I left just `ARM64` in `Toggl.iOS.csproj`.

### Why not another way?
I don't think we have any other options.

### Side effects
We won't be able to ship the app to iPhone 5 users:

```
arm64 is the current 64-bit ARM CPU architecture, as used since the iPhone 5S and later (6, 6S, SE and 7), the iPad Air, Air 2 and Pro, with the A7 and later chips.
```

## Review hints
Try deploying this to your iPhone from Visual Studio.

## :squid: Permissions
🦑


simonrozsival:
Oh you're right, we must change `UIRequiredDeviceCapabilities` (currently `armv7`)

Issue #5871, opened 2019-09-10
Calendar: Add contextual menu


sofokli1:
VM work should be done in this issue

For all contextual menus, if the user taps outside the the contextual menu should shake (this is to show the user that it needs to make a decision in order to dismiss it). There is one exception to the above and that is for the situations where the user drags a TE start, stop or the whole thing (only for the one that is in focus in the contextual menu).

Ideally we want the start and end time that is shown in the contextual menus (where it applies) to update when the user drags the handles of the TE or the whole TE. If this is hard to manage please let me know so that we can act accordingly.

### Newly created TE contextual menu (triggered when creating a TE with long press)
- [x] Discard (deletes the TE)
- [x] Edit (opens edit view)
- [x] Save

- [x] New `CalendarNewTimeEntryCM` event with 1 parameters (triggered either when the contextual menu is dismissed or when one of the options is tapped)
- `SelectedOption` string with values from the following list
    - `Dismiss` (this is for the x button interaction)
    - `Discard`
    - `Edit`
    - `Save`

Design: https://app.zeplin.io/project/5a0d3cd27a75add1a8924f6e/screen/5d8a2802a3c39b61c2643a26

### Existing TE contextual menu (triggered when tapping on an existing TE)
- [x] Delete
- [x] Edit (open edit view)
- [x] Save (to save changes when dragging the handles or the whole view)
- [x] Continue (starts a new TE with same metadata project, tags, etc at current time. Same as in main log)

- [x] New `CalendarExistingTimeEntryCM` event with 1 parameters (triggered either when the contextual menu is dismissed or when one of the options is tapped)
- `SelectedOption` string with values from the following list
    - `Dismiss` (this is for the x button interaction)
    - `Delete`
    - `Edit`
    - `Save`
    - `Continue`

Design: https://app.zeplin.io/project/5a0d3cd27a75add1a8924f6e/screen/5d8bb8fbf389236569b71bc2

### Running TE contextual menu (triggered when tapping on a running TE)
- [x] Discard (deletes the TE)
- [x] Edit (opens the running TE edit view)
- [x] Save (to save changes when dragging the handle on top)
- [x] Stop (stops the running TE)

- [x] New `CalendarRunningTimeEntryCM` event with 1 parameters (triggered either when the contextual menu is dismissed or when one of the options is tapped)
- `SelectedOption` string with values from the following list
    - `Dismiss` (this is for the x button interaction)
    - `Discard`
    - `Edit`
    - `Stop`
[...]

PR #6018, opened 2019-10-03
Add the contextual menu view model

daividssilverio:
## What's this?
This adds the contextual menu view model logic.

### Relationships
Part of #5871 (Foundation part)

## Why do we want this?
To have the contextual menu view model shared logic.

## What's missing (Why is it draft?)
- [x] all proper actions (some are missing, some are not right yet)
- [x] the events
- [x] observable for start/end text
- [x] observables for the description, project, task related texts
- [x] logic to abort/dismiss actions
- [x] tests for everything in this list (except for the events)

## How is it done?
It's made to be used as a child VM for the `CalendarDayViewModel` from the yet to be merged #5992, but it could also be used in some other way.

### Why not another way?

### Side effects
It will deprecate a bunch of other view model methods from the related old calendar view models.

## Review hints
Review by commit;
Please help us find any problems that could arise when using this VM in the iOS and droid views.

## :squid: Permissions
📆 team after the draft becomes a PR and it's approved.

@heytherewill please take a look.
@arturs105 please make sure this ok to use on the iOS side.


daividssilverio:
Hey @arturs105 ,

The usage I had in mind was similar to the running time entry card from the main fragment.

We'd have the binds for each calendar day view page:

```
// menu is a simple view on android, I don't know what it would be on iOS
contextualMenuVM.ShouldBeOpen.Subscribe(menu.Rx().IsVisible()); // or similar
contextualMenuVM.CurrentActions.Subscribe(menu.UpdateItems);
contextualMenuVM.Description.Subscribe(menu.UpdateDescription);
contextualMenuVM.Period.Subscribe(menu.UpdatePeriod);
menu.Rx().Dismiss().Subscribe(contextualMenuVM.Dismiss);
```

Then, the current `handleCalendarItem` from `CalendarDayViewModel` would link to the `contextualMenuVM.OnItemSelected`

When the current item in edit mode changes, we would call.
```
contextualMenuVM.Update(calendarItem);  // We would expose an observable from the calendar day view VM or directly from the view
```

The menu view would have something like:
```
void UpdateItems(ImmutableList<CalendarMenuAction> actions) {
    bind each action name, icon, `InputAction`
}
```

Regarding:
> Also: how would this be dismissed?

I'd add an observable for dismissal that would be used to revert changes where possible.
Some of the actions aren't reversible (like edit).

daividssilverio:
@arturs105 
The way I'm proposing, the `CalendarDayViewModel` would loose most of its functionality that commits changes calling interactors and navigations in favor to the `CalendarContextualMenuViewModel` which would handle all actual changes.

The `CalendarDayViewModel` would be only responsible for:
- displaying items
- display items in edit mode
- notify the contextual menu about the item in edit mode

The actual changes that happen as side effects of the contextual menu actions would be propagated to the `CalendarDayViewModel`, we'd just have end edit mode and dismiss the menu and profit.

arturs105:
Okay, fair enough. This looks like it could work, so if @heytherewill approves, then let's go with this.

daividssilverio:
@sofokli1 could you review the events added specifically in this commit?

https://github.com/toggl/mobileapp/pull/6018/commits/1ff11ad0e1baae5c804ca2f307cfc542697e43aa

Just to make sure they are catching what they were supposed to catch.

daividssilverio:
Hey @sofokli1 , that event is very similar to `CalendarTimeEntryCreated`. I'll add the other one. I missed it in the original issue.

Toggl.Core.UI/ViewModels/Calendar/CalendarContextualMenuViewModel.cs
```diff
@@ -0,0 +1,524 @@
...
+                        return deleteTimeEntry(item); 
+                    })
+                },
+                new CalendarMenuAction
+                {
+                    Title = Resources.Edit,
+                    Kind = CalendarMenuActionKind.Edit,
+                    MenuItemAction = rxActionFactory.FromAsync<CalendarItem>(item => 
+                    {
+                        analyticsService.CalendarRunningTimeEntryCM.Track("Edit");
```
heytherewill: Why so many raw string passing when these could be an enum?

Toggl.Core.UI/ViewModels/Calendar/CalendarContextualMenuViewModel.cs
```diff
@@ -0,0 +1,524 @@
...
+                .Execute();
+            var prototype = timeEntryToStart.AsTimeEntryPrototype(workspace.Id);
+            await interactorFactory.CreateTimeEntry(prototype, TimeEntryStartOrigin.CalendarEvent).Execute();
+            closeMenuWithCommittedChanges();
+        }
+
+        private CalendarContextualMenu setupRunningTimeEntryActions()
+        {
+            var actions = ImmutableList.Create(
+                new CalendarMenuAction
```
heytherewill: Couldn't you you create static functions like `CalendarMenuAction.ForDiscard(actionToRunAfterTheAnalyticsAndStufff)`? You seem to be repeating yourself quite a lot

Toggl.Core/Analytics/BaseAnalyticsService.cs
```diff
@@ -195,6 +195,16 @@ public abstract class BaseAnalyticsService : IAnalyticsService
         public IAnalyticsEvent<bool> AccessibilityEnabled { get; }
 
         public IAnalyticsEvent<bool> WatchPaired { get; }
+        
+        public IAnalyticsEvent<string> CalendarEventCM { get; }
```
heytherewill: Please don't abbreviate, your future self will be grateful

Toggl.Shared/Resources.resx
```diff
@@ -1368,4 +1368,7 @@ to project members</value>
     <data name="No" xml:space="preserve">
         <value>No</value>
     </data>
+    <data name="Now" xml:space="preserve">
```
heytherewill: There was no now string before? Wow, weird

Toggl.Core/Analytics/BaseAnalyticsService.cs
daividssilverio: @sofokli1 , what do you think? change it or keep it? `CalendarEventContextualMenu` or `CalendarEventCM`? That will change all the other related events as well.

Toggl.Core.UI/ViewModels/Calendar/CalendarContextualMenuViewModel.cs
```diff
@@ -0,0 +1,524 @@
...
+        private void closeMenuWithCommittedChanges()
+        {
+            currentMenuSubject.OnNext(contextualMenus[ContextualMenuType.Closed]);
+            menuVisibilitySubject.OnNext(false);
+        }
+
+        private CalendarContextualMenu setupCalendarEventActions()
+        {
+            var actions = ImmutableList.Create(
+                new CalendarMenuAction
```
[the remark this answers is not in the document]
daividssilverio: It seems like an overkill TBH, these lists are built only here and we are not exposing the API to be used anywhere else for now.

Do you mind showing a snippet of how you think it should be?
Are you talking about something like this?
```
CalendarContextualMenu
.WithAction(title, kind, action) // manual
.WithAction(Discard.BuildForX) // using your idea in a comment below
.Build(DismissAction)
```

I'll consider it if you think it's worth it.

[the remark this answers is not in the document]
heytherewill: I think using *only* the idea from the comment below is actually better than the builder

Toggl.Core.UI/ViewModels/Calendar/CalendarContextualMenuViewModel.cs
daividssilverio: I did, but not for the static functions for struct itself because of the `rxActionFactory`, I've added local functions.

Toggl.Core.UI/ViewModels/Calendar/CalendarContextualMenuViewModel.cs
daividssilverio: All of them are enums now (except the exception types).

Toggl.Core/Analytics/BaseAnalyticsService.cs
sofokli1: @maritoggl can you help us out with this. How would you like the name since you are going to use these more?

maritoggl: If there's no char limit, then, true, CalendarEventContextualMenu is way more straightforward and will not necessarily require any additional documentation.

PR #6024, opened 2019-10-03
Update Xamarin.Android.Arch packages

heytherewill:
## What's this?
This bundle the arch package updates

### Relationships
Supersedes #6012 and #6014 

## :squid: Permissions
Only myself


heytherewill:
That's actually a fair point. We should wait until AndroidX to update these dependencies

PR #6030, opened 2019-10-03
Handle start and stop actions from widgets

### LINKED ISSUE #5086: Handle start and stop actions
Call the API to start and stop actions, the API client instance moved to `ExtensionKit` contains all we need to start and stop time entries.

Check `StartTimerIntentHandler` and `StopTimerIntentHandler` in the Siri extension for reference.

Depends on #5085 and #5500

dshr:
## What's this?
Teaches our widgets how to call the API and start/stop TEs

### Relationships
Closes #5086

## Why do we want this?
For the greater good

## How is it done?
By adding a dedicated networking handler which will take care of all our networking needs.

## :squid: Permissions
🎛 people only, and only after this is not a draft.


Toggl.iOS.TimerWidgetExtension/TodayViewController.cs
```diff
@@ -1,13 +1,17 @@
 ﻿using System;
-using NotificationCenter;
 using Foundation;
+using NotificationCenter;
```
juanlaube: 👀

Toggl.iOS.TimerWidgetExtension/TodayViewController.cs
```diff
@@ -16,6 +20,13 @@ public override void ViewDidAppear(bool animated)
         {
             base.ViewDidAppear(animated);
             SharedStorage.Instance.SetWidgetUpdatedDate(DateTimeOffset.Now);
+
+            StopButton.AddTarget(stopTimeEntry, UIControlEvent.TouchUpInside);
```
juanlaube: This should be in `viewDidLoad`, right? Also, no start action? 🙂

Toggl.iOS.TimerWidgetExtension/TodayViewController.cs
dshr: That's needed for `WidgetPerformUpdate`and `INCWidgetProviding`

Toggl.iOS.TimerWidgetExtension/TodayViewController.cs
dshr: D'oh! Completely missed the `ViewDidAppear`. As for the start action — isn't that dependent on #5087?

PR #6031, opened 2019-10-03
📐 Prevent text in pie chart from escaping its slice

### LINKED ISSUE #5937: Text in pie chart escapes the containing slice
## 🐛 Bug Report
The title says it all. Screenshot for reference (left - iOS, right - Android):
Screenshot 2019-09-18 at 17 31 27

## Context

Build/Branch: develop
Phone OS: Android and iOS
Phone model: Pixel 3 and iPhone Xs (emulators)

## Expected behavior
Text being contained in the corresponding slice.

## To reproduce
Create 9 time entries and assign them to different projects. Make the project names long and wide (letter `M` is pretty wide :)).
 
## Suggestions

- Do some measurements instead of just capping the project title length at 18.
- For Android, think what to do if the device's font is bigger than regular. Maybe use `dp` instead of `sp`.

pdrj:
## What's this?
This prevents text in pie chart from escaping its slice.

### Relationships

Closes #5937

## Why do we want this?
It looks better.

## How is it done?
By comparing the length of the text that we want to print with the diameter of the pie chart. If the text is too long, we try to figure out how many letters shorter it has to be to fit.

### Why not another way?
Suggestions welcome. My approach is not ideal, as sometimes the text might not fit vertically near the pie chart center - I tried to compute the min radius at which the text is guaranteed to fit, but it's been a while since I dealt with math equations involving sines and stuff ;).

## Review hints
Create several TEs with similar lengths and see how the pie chart presents on different screens.

Before and after:

- iPad emulator
Screenshot 2019-10-03 at 15 22 18
Screenshot 2019-10-03 at 16 10 35
- Pixel emulator
beforepixel
afterpixel

- Pixel with font set to largest
Screenshot 2019-10-03 at 12 33 29

## :squid: Permissions
Me


Toggl.iOS/Views/Reports/PieChartView.cs
```diff
@@ -84,6 +86,17 @@ public override void Draw(CGRect rect)
 
                     var textWidth = nameToDraw.Size.Width;
                     var textHeight = nameToDraw.Size.Height;
+                    var initialTextLength = nameToDraw.Length;
+                    
+                    for (var shortenBy = 1;
```
heytherewill: A conditional loop would be much easier to reason about than a counted loop here

Toggl.Droid/Views/PieChartView.cs
```diff
@@ -116,6 +120,17 @@ protected override void OnDraw(Canvas canvas)
                     var percentageToDraw = $"{integerPercentage}%";
 
                     var (textWidth, textHeight) = getTextWidthAndHeight(nameToDraw);
+                    var initialTextLength = nameToDraw.Length;
+
+                    for (var shortenBy = 1;
+                        shortenBy < initialTextLength - minVisibleLetters
```
heytherewill: Same thing applies here

Toggl.iOS/Views/Reports/PieChartView.cs
pdrj: I started with a `while` there, but thought this one is better 😃 I don't have a strong preference though - will change it.

simonrozsival: I actually had a comment suggesting to rewrite the `while` with `for`, because you have initialization, condition, and then you have incrementation... that's what you use `for` for. But it's not a big deal for me I guess.

pdrj: Yeah that was exactly my thinking as well 😄 Not a big deal for me either though.

PR #6036, opened 2019-10-04
Fix project dot colors on iOS 13

### LINKED ISSUE #5981: Project colours not showing correctly in iOS 13 log
The project-bullet on time entry cells in the time entry log all have the same (and seemingly random?) colour on iOS13.

They are supposed to have the same colour as the project name right next to them (which looks fine).

simonrozsival:
## What's this?
This is a UI fix for iOS 13.

### Relationships
Closes #5981 

## Why do we want this?
We want the app to work well on iOS 13

## How is it done?
We used a hacky solution to set the color of the attributed string which does not work anymore in iOS 13. We have to leave the hack for older versions of iOS, otherwise it'll look bad (black) on older phones.

### Why not another way?
This solution is simple and it works as it should.

### Side effects
No side effects.

## Review hints
Test it in simulator on a device with iOS 11 or 12 and then on iOS 13. Both should look the same:

fixed

## :squid: Permissions
🦑


Toggl.iOS/Extensions/TextExtensions.cs
```diff
@@ -68,9 +68,15 @@ private static NSAttributedString endingWithIcon(string text, NSMutableAttribute
             var y = (fontCapHeight - imageSize.Height) / 2;
             attachment.Bounds = new CGRect(0, y, imageSize.Width, imageSize.Height);
 
-            //There neeeds to be a space before the dot, otherwise the colors don't work
-            var result = new NSMutableAttributedString(" ");
-            result.AddAttribute(UIStringAttributeKey.Font, UIFont.SystemFontOfSize(0), new NSRange(0, 1));
+            var result = new NSMutableAttributedString("");
+
+            if (!UIDevice.CurrentDevice.CheckSystemVersion(13, 0))
```
heytherewill: I would actually like to have an API similar to what we have on Android, so that when we deprecate versions it's easier to remove branches. Something like `Ios13Apis.AreNotAvailable`. I guess it should be done in another PR tho

simonrozsival: There are so few places where we check the OS version that I don't think that it's worth the effort 😄

PR #6037, opened 2019-10-04
✨Add new analytic event for selected cals

### LINKED ISSUE #5998: Calendar: Add analytics event for count the linked calendars
IMPORTANT This should be done against develop since it is going to be used to measure this projects outcome, so we need it already out there before we are done with the project in order to have something to compare.

- [x] New `NumberOfLinkedCalendarsChanged` event with one parameter (triggered when user updates the connected calendars) 
    - `Count` int, number of linked calendars
- [x] New `NumberOfLinkedCalendarsNewUser` event with one parameter (triggered when user connects calendar for the first time)
    - `Count` int, number of linked calendars

sofokli1:
## What's this?
Tracks the number of selected calendars for from new users and returning users

### Relationships
Closes #5998

## Why do we want this?
We want to check the how many calendars people connect before the calendar project is done and after it is out to the users so that we can measure if we have an increase in calendars connected or a drop in them.

## How is it done?
2 analytics events one for updates to the connected calendars changes and one for the first time of connecting calendars. Also added a flag in `onboarding storage` to keep track if the user has already connected calendars or not. Also checks if the user has connected calendars together with the flag (since the flag was not there before).

### Why not another way?
you tell me

### Side effects
none that i know

## Review hints
check the tests

## :squid: Permissions
:squid:


Toggl.Core.UI/ViewModels/Settings/CalendarSettingsViewModel.cs
```diff
@@ -83,6 +86,22 @@ protected override void Done()
...
+            if (OnboardingStorage.IsFirstTimeConnectingCalendars() && InitialSelectedCalendarIds.Count == 0)
+            {
+                AnalyticsService.NumberOfLinkedCalendarsNewUser.Track(SelectedCalendarIds.Count);
+            }
+            else
+            {
+                AnalyticsService.NumberOfLinkedCalendarsChanged.Track(SelectedCalendarIds.Count);
+            }
+
+            OnboardingStorage.SetIsFirstTimeConnectingCalendars();
```
pdrj: This is the same code as in base. Can we remove it from here and have a method in the base class that this will call?

Toggl.Core.UI/ViewModels/Calendar/SelectUserCalendarsViewModelBase.cs
```diff
@@ -122,6 +133,17 @@ public override void CloseWithDefaultResult()
 
         protected virtual void Done()
```
pdrj: The event sometimes can be sent twice, once because of `Done` and then once again - because of `ViewDisappeared` in `CalendarSettingsViewModel`. Is it safe to do so?

pdrj:
Thanks for the DRY change, but I think my other comment still applies. If I'm not clear enough - if you use the 'done' button on android, the analytics code will get executed twice: once as a result of clicking the 'done' button and a second time, because of the call in 'ViewDissapeared'. Is it OK to do so?

sofokli1:
@pdrj I think the new commit fixes the possible double trigger of the `Done` button. Please take a look again

Issue #5868, opened 2019-09-10
Calendar: Add manual time entry creation


sofokli1:
Depends on #5867 

Manual TEs
- [x] Use long press to create new Entry with handlers and contextual menu appear.
- [x] New entries should snap to quarter hour and other calendar events start and end times
- [x] Should have a default duration of 30 min
- [x] If created in the gap between 2 TEs that is 2 hours or less, it should fill the whole gap by default

Design: https://app.zeplin.io/project/5a0d3cd27a75add1a8924f6e/screen/5d8a2802a3c39b61c2643a26

daividssilverio:
Closed by #6300

PR #6052, opened 2019-10-08
Add droid calendar swipe to change day

### LINKED ISSUE #5863: Calendar: Support changing date by swiping

### LINKED ISSUE #5864: Calendar: Support changing date by swiping
Ability to swipe left and right to move a single day back and forth

daividssilverio:
## What's this?
This makes the calendar tab fragment a view pager for calendar days.

### Relationships
Closes #5864  

## How is it done?
With a view pager and some relays.

### Why not another way?
I thought about using a raw `PagerAdapter` and handle `CalendarDayView`s directly... but the `FragmentStatePagerAdapter` worked well... so I just used it.

## Review hints
None this time.

## :squid: Permissions
🦑 it 📆 team


semanticer:
Clicking on a calendar tab moves me to the current time right? But maybe it should also move the page to the current date? At least that's what I would expect as a user.

semanticer:
It's very easy to move back accidentally instead of moving between days with the new Android 10 gesture navigation but I guess there is nothing we can do about that.

semanticer:
Maybe you should consider new ViewPager2 - https://medium.com/google-developer-experts/exploring-the-view-pager-2-86dbce06ff71 https://developer.android.com/jetpack/androidx/releases/viewpager2#1.0.0-beta04

Oh no wait - it's probably only for Androidx 🤦‍♂

sofokli1:
I think the issues this closes is for iOS (and it closed already), I think it should be #5864

daividssilverio:
> I think the issues this closes is for iOS (and it closed already), I think it should be #5864

Thanks.

daividssilverio:
> As I mentioned above, I don't think the current calendar tab click behavior is correct. I'm also not sure if we should display the line with current time in all of the pages - I would prefer it only on today's page.

Let's just confirm that with @maritoggl . The change is trivial.

> The interactions like creating a new event by clicking into calendar view don't work for me - but I guess that's not the point of this PR and you are in the middle of polishing these things?

There's an issue for that, which will work nicer when we finish the contextual menu as well #5868 
 
> Given that we probably can't use the new viewPager2 (maybe with reverse jetifier?) the code looks fine.

Oh man... you make me dream, I'll take a look.

daividssilverio:
> Given that we probably can't use the new viewPager2 (maybe with reverse jetifier?) the code looks fine.
> Oh man... you make me dream, I'll take a look.

@semanticer as far as I checked, maybe not for now. We can revisit all the view pagers in the app later when we move to androidX as well.

daividssilverio:
@semanticer 
>It's very easy to move back accidentally instead of moving between days with the new Android 10 gesture navigation but I guess there is nothing we can do about that.

We can opt-out for navigation gestures while in this lab when we update our stack to use Xamarin 10. But that should be done first in develop, then merged back into the project branch to then be fixed. I'll create an issue for that.

PR #6053, opened 2019-10-08
🕶 iOS 13 dark mode

### LINKED ISSUE #5969: Implement dark mode for iOS
Most of the heavy lifting was done by me during go-loco, there's just some color tweaking and some screens left. I'll follow the comments on #design-discussions for this.

Odrakir:
## What's this?
This PR implements dark mode for iOS, it also changes the way we handle colors and a couple of other things along the way, more on this below.

### Relationships
Closes #5969 

## How is it done?
Basically we'll use a Color Asset Catalog instead of the shared colors from now on. This is an special json file that will have two versions for every color we add there (one for light mode and one for dark mode), plus another one for previous iOS versions, which will match light mode.

Some things to take into account:
- We can't use `Appearance` for `UITabBar` and `UINavigationBar` anymore because they won't refresh properly when switching style from light to dark or back.
- I replaced all separators by ones created in code, because they have to be 1 pixel thick, and if we create them in IB they are 1 point thick (which translate to ~3 pixels in most phones)
- `UIBarButton` item customization is moved to the `UINavigationController` class because otherwise it doesn't work
- Doing this, I noticed there's A LOT of repetition in all the entity edition/creation table views, we should refactor it.

## Review hints
One commit after another, commit history is good. But mostly, use the app in both modes and see if you find anything funny.

## :squid: Permissions
🚫  Not yet!


Toggl.iOS/Extensions/ColorAssets.cs
```diff
@@ -5,5 +5,7 @@ namespace Toggl.iOS.Extensions
     public class ColorAssets
     {
         public static readonly UIColor Background = UIColor.FromName("Background");
+        public static readonly UIColor Text = UIColor.FromName("Text");
+        public static readonly UIColor Text2 = UIColor.FromName("Text2");
```
dshr: Is it possible to have a bit more descriptive colour names? Like, I won't know the difference between `Text` and `Text2` without peeking into the ColorAssets files. OTOH, they can stay like this, and have them used as preset colours for Sketch and Zeplin, so when we inspect an element there we'd get `Text2` instead of `rgba(0.235294119, 0.235294119, 0.2627451, 0.6784314)`, but I don't know if this is possible.

Odrakir: I agree with you. I think we definitely should do something about this, but since it has to be done together with design, we can probably wait for the design changes Alicja is working on. We should then have a few semantic colors for both light and dark mode that can be used easily all over the app.

Toggl.iOS/ViewSources/StartTimeEntryTableViewSource.cs
```diff
@@ -81,8 +81,8 @@ public override UITableViewCell GetCell(UITableView tableView, NSIndexPath index
                             .Subscribe(toggleTasks.Accept)
                             .DisposedBy(cell.DisposeBag);
 
-                        cell.TopSeparatorHidden = true;
-                        cell.BottomSeparatorHidden = true;
+                        //cell.TopSeparatorHidden = true;
```
dshr: 👀

Toggl.iOS/Views/StartTimeEntry/ProjectSuggestionViewCell.cs
```diff
@@ -31,7 +31,7 @@ public bool BottomSeparatorHidden
         {
             get => BottomSeparatorView.Hidden;
             set => BottomSeparatorView.Hidden = value;
-        }
+        }*/
```
dshr: Same thing, let's not merge commented-out code

Odrakir:
About the overlay of card views... I don't think there's a guideline for that, Apple uses transparency in the music app, maybe we could go with that, but I would like the input of a designer on this. I would release as it is right now and fix it later.

IMG_CB95E6766327-1

Toggl.iOS/Views/Log/RefreshControl.cs
```diff
@@ -27,10 +27,10 @@ public sealed class RefreshControl
 
         private static readonly float scrollThreshold = 3 * syncBarHeight;
 
-        private readonly UIColor pullToRefreshColor = Colors.Main.PullToRefresh.ToNativeColor();
-        private readonly UIColor syncingColor = Colors.Main.Syncing.ToNativeColor();
-        private readonly UIColor syncFailedColor = Colors.Main.SyncFailed.ToNativeColor();
-        private readonly UIColor offlineColor = Colors.Main.Offline.ToNativeColor();
+        private readonly UIColor pullToRefreshColor = ColorAssets.CustomGray2;
+        private readonly UIColor syncingColor = ColorAssets.CustomGray2;
+        private readonly UIColor syncFailedColor = ColorAssets.CustomGray2;
```
simonrozsival: "Sync failed" should be red

PR #6065, opened 2019-10-08
Delete everything from SharedStorage on logout

### LINKED ISSUE #5498: Delete data from shared storage on logout
On log out, we need to delete all the suggestions and running time entry information from the shared storage.

dshr:
## What's this?
A simple PR to 🔥 the whole SharedStorage

### Relationships
Closes #5498 

## Why do we want this?
🎛 

## How is it done?
By injecting the UserAccessManager into our TimerWidgetService and then watching the UserLoggedOut observable to trigger the :fire: 

## :squid: Permissions
🎛 folx only


Toggl.Core.UI/Services/TimerWidgetService.cs
```diff
@@ -8,7 +9,7 @@ public abstract class TimerWidgetService
     {
         private IDisposable runningTimeEntryDisposable;
 
-        protected TimerWidgetService(ITogglDataSource dataSource)
+        protected TimerWidgetService(ITogglDataSource dataSource, IUserAccessManager userAccessManager)
```
heytherewill: Why include something in base if it isn't being used?

Toggl.iOS/Services/TimerWidgetServiceIos.cs
```diff
@@ -11,14 +13,19 @@ namespace Toggl.iOS.Services
...
-        public TimerWidgetServiceIos(ITogglDataSource dataSource) : base(dataSource)
+        public TimerWidgetServiceIos(ITogglDataSource dataSource, IUserAccessManager userAccessManager) : base(dataSource, userAccessManager)
         {
             durationFormatDisposable = dataSource
                 .Preferences
                 .Current
                 .Select(preferences => preferences.DurationFormat)
                 .Subscribe(onDurationFormat);
+
+            userLoggedOutDisposable = userAccessManager
```
heytherewill: Arguably this should be part of the logout interactor

dshr: Why? I don't think that the logout interactor should care about SharedStorage at all, it should just log the user out and let everybody know via the Observable.

PR #6068, opened 2019-10-08
🕶 Dark mode for reports

simonrozsival:
## What's this?
This is dark mode for the Reports screen + the selection screen (it was necessary for the workspace selection, I hope it won't cause merge conflicts).

### Relationships
Ref #5969

## Why do we want this?
We want dark mode in the next release.

## How is it done?
Using `ColorAssets`. I tried to set most of the colors in the interface builder and change stuff in code only when it was necessary (e.g., backgrounds of table views).

The view page should look good in the Dark mode and the "new light look". There were no designs and I hope it looks good 😄.

### Why not another way?
There is no other way.

### Side effects
Some people might not like it (or some details of it).

## Review hints
Try it in the simulator, both the light and dark modes. Here are some screenshots:

| Dark | Light |
|:----:|:----:|
| image | image |
| image |  image| 
|  image | image |
| image | image |
| image |  image |

I think we should merge this ASAP to avoid conflicts and tune the colors with a designer or with the product manager(s) and tune it later.

## :squid: Permissions
@Odrakir


simonrozsival:
BTW the iPad version is kinda broken and sad :D 
image
image

image
image


... but the white background stripes on the sides are on the timer page as well, so I guess we can fix it later (@dshr could help with that - where do I set that background color? I'm not sure where to look for it and it might be very easy, so please show me :)

simonrozsival:
(also the borders on the iPad are kinda weird - how to get rid of those?)

Toggl.iOS/Extensions/ColorAssets.cs
```diff
@@ -6,6 +6,7 @@ public class ColorAssets
     {
         public static readonly UIColor Background = UIColor.FromName("Background");
         public static readonly UIColor Text = UIColor.FromName("Text");
+        public static readonly UIColor InverseText = UIColor.FromName("InverseText");
```
Odrakir: This one is basically the same as `Background`, right? I'm not saying we shouldn't have it, but if we go down that road we might end up with a bunch of colors that are the same but with different names. What's your thinking? Should we instead have just one with another name?

Toggl.iOS/Cells/Common/SelectorCell.cs
```diff
@@ -34,6 +35,7 @@ protected override void UpdateView()
         {
             NameLabel.Text = Item;
             CheckMark.Hidden = !OptionSelected;
+            BackgroundColor = ColorAssets.CellBackground;
```
Odrakir: In the case of cells I'm also removing the separator that's created in IB and adding it in code with the `InsertSeparator` extension method. This way they'll be 1 pixel thick, and not 1 point thick.

Toggl.iOS/Views/Reports/ReportsCalendarViewCell.cs
```diff
@@ -42,28 +42,27 @@ private void prepareViews()
...
-
-            if (Item.IsSelected(dateRange))
-            {
-                Text.TextColor = selectedColor;
-            }
-            else
-            {
-                Text.TextColor = Item.IsInCurrentMonth ? thisMonthColor : otherMonthColor;
-            }
+            Text.TextColor = (Item.IsToday && !Item.IsSelected(dateRange))
```
Odrakir: I am dizzy reading this...

Toggl.iOS/Extensions/ColorAssets.cs
simonrozsival: I'd like to keep it. I know it's the same and I thought about it for a while, but I didn't want to use a color which is used for backgrounds as a text color. If you don't think it's a good reason, we can get rid of it.

PR #6074, opened 2019-10-09
🕶 Calendar dark mode

simonrozsival:
## What's this?
Dark mode for the calendar view.

### Relationships
Ref #5969 

## Why do we want this?
We want dark mode in the next release.

## How is it done?
I replaced some colors + I also adjusted the color of the spider for the dark appearance (I made a bit lighter so it's not dark against dark background).

### Why not another way?
There's no other way.

### Side effects
TEs with dark project colors (e.g., black) blend with the background. People are used to light background in the webapp and in the mobile app, so they might be using dark project colors and they might not like it. Maybe we should add some light border/shadow to these dark colors in the dark mode? No idea... we'll have to fix this later I guess. Or maybe nobody will notice.

## Review hints
I made some screenshots:

| New light mode | Dark mode |
| :--------------: | :----------: |
| image | image |
| image | image |
| image | image |
| image | image |
| image | image |

I fixed the top separator after I took some of these screenshots:

image

image

## :squid: Permissions
@Odrakir


Toggl.iOS/ViewSources/SelectUserCalendarsTableViewSource.cs
```diff
@@ -42,14 +41,14 @@ public override UITableViewCell GetCell(UITableView tableView, NSIndexPath index
             var cell = (BaseTableViewCell<SelectableUserCalendarViewModel>)tableView.DequeueReusableCell(
                 SelectableUserCalendarViewCell.Identifier, indexPath);
             cell.Item = ModelAt(indexPath);
+            cell.ContentView.InsertSeparator();
```
Odrakir: This should be in the cell class in `AwakeFromNib`, otherwise it might add it multiple times when the cell is reused

PR #6080, opened 2019-10-09
Fix settings not opening after swiping to close

### LINKED ISSUE #6011: Settings view cannot be re-opened when dismissed by swiping down
## 🐛 Bug Report
The settings view becomes "stuck" when it is swiped down and it cannot be re-opened.

Build/Branch: `develop`
Phone OS: iOS 13, built with Xcode 11
Phone model: iPhone

## Expected behavior
The settings view can be re-opened.

## To reproduce
1. go to settings
1. dismiss with the the "done" button
1. go to settings again
1. dismiss by swiping down
1. go to settings
1. oops

Odrakir:
## What's this?
This fixes an issue that prevented the settings modal from opening after it was dismissed by swiping.

### Relationships
Closes #6011 

## How is it done?
I haven't found what is going on, but something tells me is some kind of Xamarin bug, there seems to be an [issue](https://github.com/xamarin/Xamarin.Forms/issues/7145) with Xamarin Forms as well.

Anyway, I fixed it by completing the navigate task right when its shown, I think this causes no other problems and fixes the bug.

While on it I also removed some old code.

## :squid: Permissions
Go!


heytherewill:
This is pretty palliative, though. We should handle the swipes properly instead.

Odrakir:
William, I am not sure what you mean, we are not handling swipes at all there, it's all taken care by the system.

juanlaube:
I didn't test this branch, but looking at the code I think this is the way to go. Unfortunately we can only know when the view controller is dismissed by the swipe gesture using `UIAdaptivePresentationControllerDelegate`.

PR #6088, opened 2019-10-10
Handle logged out case in the widget

### LINKED ISSUE #5089: Handle the case when user is not logged in
Show a "Login to track time" message instead of the widget content and open the app  when tapping on the widget.

You can check if the api token is present in the shared storage to determine if the user is logged in or not.

Depends on #5085 and #5400

dshr:
## What's this?
A catch to the logged out case in our iOS widget

### Relationships
Closes #5089 

## Why do we want this?
Better UX and no 💥 

## How is it done?
By checking if the API token is save in shared storage and then showing the appropriate error view.

### Why not another way?
Why not this way?

## :squid: Permissions
🎛


Toggl.iOS.TimerWidgetExtension/TodayViewController.cs
```diff
@@ -25,22 +24,31 @@ protected TodayViewController(IntPtr handle) : base(handle)
...
             StartButton.AddTarget(startTimeEntry, UIControlEvent.TouchUpInside);
             StopButton.AddTarget(stopTimeEntry, UIControlEvent.TouchUpInside);
-        }
 
-        public override void ViewDidAppear(bool animated)
-        {
-            base.ViewDidAppear(animated);
-            SharedStorage.Instance.SetWidgetUpdatedDate(DateTimeOffset.Now);
 
+            if (SharedStorage.Instance.GetApiToken() == null)
```
juanlaube: This should be in `viewDidAppear`. `viewDidLoad` is called when the widget is created but can appear and disappear mutiple times before it's created again.

Toggl.iOS.TimerWidgetExtension/MainInterface.storyboard
```diff
@@ -21,6 +19,12 @@
                         <rect key="frame" x="0.0" y="0.0" width="320" height="120"/>
                         <autoresizingMask key="autoresizingMask" widthSizable="YES" heightSizable="YES"/>
                         <subviews>
+                            <label hidden="YES" opaque="NO" userInteractionEnabled="NO" contentMode="left" horizontalHuggingPriority="251" verticalHuggingPriority="251" text="Login to track time" textAlignment="natural" lineBreakMode="tailTruncation" baselineAdjustment="alignBaselines" adjustsFontSizeToFit="NO" translatesAutoresizingMaskIntoConstraints="NO" id="Ga5-pU-ZGk" userLabel="Error Message Label">
+                                <rect key="frame" x="89" y="40" width="142.5" height="20.5"/>
+                                <fontDescription key="fontDescription" type="system" pointSize="17"/>
+                                <nil key="textColor"/>
```
juanlaube: The text color is not black (nil) in the design.

Toggl.iOS.TimerWidgetExtension/TodayViewController.cs
juanlaube: Another option is to recreate the api client if the token changes.

juanlaube:
We also need to launch the app when tapping on the widget if it's showing an error, there's an url scheme for that.

Toggl.iOS.TimerWidgetExtension/TodayViewController.cs
```diff
@@ -25,20 +25,39 @@ protected TodayViewController(IntPtr handle) : base(handle)
         public override void ViewDidLoad()
         {
             base.ViewDidLoad();
+
             StartButton.AddTarget(startTimeEntry, UIControlEvent.TouchUpInside);
             StopButton.AddTarget(stopTimeEntry, UIControlEvent.TouchUpInside);
+
+            tapGestureRecognizer = new UITapGestureRecognizer(() =>
+            {
+                ExtensionContext?.OpenUrl(new Uri("toggl://tracker"), null);
```
heytherewill: We've had problems with using raw strings in the past when the URL becomes unused; this should be using the helpers from Core instead.

dshr: Can't seem to find anything that would generate deeplinks in our codebase, I only see the `DeeplinkParser` and its params :(

PR #6098, opened 2019-10-10
🎨 Add not-logged-in view to widgets

### LINKED ISSUE #5096: Handle the case when user is not logged in
Show a "Login to track time" message instead of the widget content and open the app when tapping on the widget.

Depends on #5093 and #5405

kornelijepetak:
## What's this?
Does two things.

1. Implements the not-logged-in widget.

2. While doing 1., I also reworked the way we are handling the widget form factors. The previous solution was full of conditionals and branching to get to what we need. In this PR I've tried to separate concerns properly.

### Relationships
Closes #5096 

## Why do we want this?
So that widgets leak no data.

## How is it done?
By 1. and 2. 

### Why not another way?
You tell me. That's what review process is for. 

### Side effects
Should be none.

## Review hints
Do a proper functional. A lot of it.

## :squid: Permissions
Me.


Toggl.Shared/Resources.resx
```diff
@@ -1365,4 +1365,7 @@ to project members</value>
     <data name="No" xml:space="preserve">
         <value>No</value>
     </data>
+    <data name="LoginToTrack" xml:space="preserve">
+        <value>Log in to track time</value>
```
juanlaube: You need to push the changes in `Resources.designer.cs` too 🙂

PR #6099, opened 2019-10-10
Add the droid calendar contextual menu UI

### LINKED ISSUE #5871: Calendar: Add contextual menu

daividssilverio:
## What's this?
This adds the calendar contextual menu UI to droid and makes some minor changes to the `ContextualMenuViewModel`.

### Relationships
Closes #5871 

## Why do we want this?
Because it's dope.

## How is it done?
With relays and love.
The calendar page in the calendar day view pager is now a Fragment.
Each page contains a `CalendarDayView` and the views for the `ContextualMenu`.
The items being edited in the `CalendarDayView` are forwarded to the `ContextualMenuViewModel` which then handles changes and notifies the page to update the UI values.

### Why not another way?

### Side effects
None that I'm aware of.

## Review hints
**Please start reviewing.**
Review by commit.
Please note that the contextual menu doesn't animate visibility changes yet and I rather do it some other time because it's tricky at the moment to make it work beautifully.

## :squid: Permissions
🦑 it 📆 team after the Draft is set to be ready to review.


daividssilverio:
> A few thoughts from functional testing. These these are just my thoughts, so feel free to argue otherwise:
Deletion should show a dialog asking "are you sure?"
Closing the view when there are changes should should show a similar dialog that's asking: "Do you want to discard the changes?"

That sounds cool, but I guess that's something that has to be talked with @sofokli1 and @maritoggl and go through design.

>Tapping outside of the selected should close the contextual menu
Now that you've said, I think we agreed on that doing that should "shake" the contextual menu, but I do think it would make more sense to just close the menu (maybe give an alert to confirm).

>A bug:
Tap on a time entry to bring up contextual menu
Increase the duration of the entry by dragging the handles in the calendar view
Press "Edit"
Edit view opens with the old / un-updated duration + start/end times
Bonus bug: after pressing "save" in the edit view, the modal gets closed as well and the edits from step 2 are discarded.

That does sound like a bug, but it's a limitation of the current `EditTimeEntryViewModel`.  There's a thread in slack about it: https://toggl.slack.com/archives/CMWQK3GS2/p1570119172013100. And it shouldn't be fixed in this PR.

> Another bug:
Have some time entries visible in the calendar
Start a new time entry from the main view
Tap on the running entry in the calendar
Close the contextual menu
Now tap on one of the existing entries
💥 The menu for a running entry is shown

Ok, I'll take a look at that and fix it. Thanks!

> Another thing that bothers me but is design related: the fact that "Save" button is not green and isn't at the very end of the list - multiple times i had the urge to press the green button to save my edits, but that button is the continue button.

That's something that should be discussed somewhere else as @sofokli1 suggested. Please ping @maritoggl and let's discuss it in slack.

daividssilverio:
@heytherewill 

>From functional tests:
The header information of the menu is positioned according to where the item is and not always centred, which looks super bad when there are three items on the same slot

Thanks, I'll take a look at that.

> Time entries with no description feel really weird because there's no info shown on the contextual menu at all

Ok, that's a bug. I'll fix it.

> Same thing for calendar events, the contextual menu has no header information about them

You mean when there's no description, right? I'll fix it if that's the case.


> The contextual menu is not respecting dark mode

:wat: I'm using the droid colours 🤔 in the resources file.

> It's not clear at all that the contextual menu is locking the swiping of days. I thought it was a bug before looking at code. I advise changing this behavior or making it more clear by adding some UI indication

Thanks for the feedback. I'll pass along to @maritoggl @sofokli1 and Alicja, but that's been discussed here: https://toggl.slack.com/archives/CMWQK3GS2/p1570036959022800

>Forcing the user to dismiss only by pressing the X is not good. We should allow swiping down (or dismiss the contextual menu when pressing anywhere that is not a calendar item)

Thanks for the feedback. I'll pass along to @maritoggl @sofokli1 and Alicja.

>If you press the physical back (or make the swipe gesture) while the contextual menu is open, the bottom navigation bar will be gone forever. Pressing back should, instead, dismiss the contextual menu.

That makes all the sense, I will fix it.

AlicjaSuska:
@daividssilverio 

Swiping days when contextual menu is open - changing day in this situation doesn't make sense so I would keep this behaviour. We can disable temporarily other days if visual indication is needed. 

Closing contextual menu - is shaking the menu that @sofokli1 mentioned already implemented?

arturs105:
@AlicjaSuska i"m pretty sure that the shake isn't implemented yet, because it wasn't there when i tested this and no new code has been added to this PR since then.

AlicjaSuska:
@arturs105 thank you for the update. I think we should test how shaking menu feels like and then make the decision about the discarding/closing flow. Is that ok?

arturs105:
I don't really see the point of shaking the thing - it seems like the confirmation dialog should be there even if we have the shake behaviour because:
1. User makes some changes
2. User changes his mind and wants to close the thing, so taps on the `X`/ taps on background / swipes down or whatever
3. The dialog shakes indicating there are unsaved changes
4. Wat do? I want to close the dialog. How to discard the changes?
5. At some point the user _will_ have to press a button that means "discard these changes", so i don't see the benefit for the shake.

Am i getting the shake idea correctly? Or am i missing something?

daividssilverio:
@arturs105 @AlicjaSuska sorry for not replying earlier. 
We've discussed this internally https://toggl.slack.com/archives/CMWQK3GS2/p1571055400024900

So, yeah, we are going towards killing the shake and adding confirmation dialogs for destructive actions.

daividssilverio:
Hey @arturs105 ,  how did you manage to do this? I'm not able to replicate. I'll try to improve the blinking that happens when you click on a different time entry, is that what you are talking about?

> Have some time entries visible in the calendar
Start a new time entry from the main view
Tap on the running entry in the calendar
Close the contextual menu
Now tap on one of the existing entries
💥 The menu for a running entry is shown

daividssilverio:
Hey @arturs105 and @heytherewill , please take another look. 
I've added the logic to handle destructive actions (dismisses/discards/click outside of menu/click on different items/back press).
I've fixed the back press handling when the menu is open.

Toggl.Core.UI/ViewModels/Calendar/ContextualMenu/CalendarContextualMenuViewModel.cs
```diff
@@ -91,22 +93,78 @@ public IObservable<string> TimeEntryPeriod
             currentMenuSubject = new BehaviorSubject<CalendarContextualMenu>(closedMenu);
         }
 
-        private void handleCalendarItemInput(CalendarItem? calendarItem)
+        private async void handleCalendarItemInput(CalendarItem? calendarItem)
```
heytherewill: Can this ever throw? Why is this not a Task?

daividssilverio: I thought that adding the `async` was enough.
What's the difference of it being a `async Task` and `async void`?

daividssilverio: It could throw, if the reference to the `View` is lost and not re-attached before this is called again.

daividssilverio:
I've created issues for comments that won't be addressed on this PR. #6139 #6140 #6141

arturs105:
Oh, and this problem no longer occurs:

>Have some time entries visible in the calendar
Start a new time entry from the main view
Tap on the running entry in the calendar
Close the contextual menu
Now tap on one of the existing entries
💥 The menu for a running entry is shown

sofokli1:
when continuing a TE it would make sense to scroll to today and current time. What did we end up agreeing on the latest changes?

daividssilverio:
> when continuing a TE it would make sense to scroll to today and current time. What did we end up agreeing on the latest changes?

It does make sense, but as @maritoggl said, it's something that could be nice and could be confusing as well. (https://toggl.slack.com/archives/CMWQK3GS2/p1571053695013800)

We could just show a toast (Idk how it's called in iOS, those flash messages) saying that a time entry has been continued. Or maybe a snackbar with an action takes you to "Today".

sofokli1:
@daividssilverio leave it for now. we can talk about this again when people are testing next week

arturs105:
@heytherewill Get a better phone

daividssilverio:
@heytherewill the contextual menu should be smooth now. Please take another look.
I had too much trouble trying to fix the tests that broke because of https://github.com/toggl/mobileapp/pull/6099/commits/ba38a5e7cee925d88e721b3a68a05a7830ccc6a3, so I'll be creating an issue to fix them later as suggested by @sofokli1 .

PR #6114, opened 2019-10-11
🕶 Navbar separator

### LINKED ISSUE #5969: Implement dark mode for iOS

simonrozsival:
## What's this?
This PR adds navbar separator for the dark mode.

### Relationships
Closes #5969 

## Why do we want this?
We want nice UI.

### Why not another way?
I tried changing [ShadowImage](https://developer.apple.com/documentation/uikit/uitabbar/1623452-shadowimage), but that didn't work. So I just inserted a separator... it works and looks good.

## Review hints

| Before | After |
| :-----: | :----: |
| image | image |
| image | image |

## :squid: Permissions
🦑


Toggl.iOS/ViewControllers/MainTabBarController.cs
```diff
@@ -112,6 +112,9 @@ private void setupAppearance()
             TabBar.BackgroundImage = ImageExtension.ImageWithColor(ColorAssets.Background);
             TabBar.SelectedImageTintColor = Colors.TabBar.SelectedImageTintColor.ToNativeColor();
             TabBarItem.TitlePositionAdjustment = new UIOffset(0, 200);
+
+            var separator = TabBar.InsertSeparator(UIRectEdge.Top);
```
Odrakir: The problem I see with this is that we end up with a very thick bar in light mode (which already has a separator)... 

Looking at system apps, they don't add a separator in dark mode. Not here and not in the nav bar.

simonrozsival: It looks very weird without the separator. The Clock app definitely has a separator in the dark mode - and it's the example app in the NavBar documentation. Also I'm not sure if I'm holding it wrong, but I wasn't able to modify the `ShadowImage`, which I guess is the way to change the color of the "separator".

Feel free to close this PR if you decide it's not worth dealing with.

Odrakir: 🤔 You are right about the clock. Calendar doesn't have one on the other hand...

Odrakir: I'm also unable to remove that line... I think I'll close this for now, let's see in the future.

simonrozsival: OK

PR #6116, opened 2019-10-13
Clean up calendar settings

### LINKED ISSUE #5955: Calendar: Drop calendar on-boarding
Drop the old calendar on-boarding that was shown in the calendar view when first opening the calendar view.

- [x] delete the old logic
- [x] make sure the user can still connect calendars from the settings
- [x] make sure the permissions asking logic is still working

sofokli1:
### Relationships
Closes #5955 

## Why do we want this?
we do not want dead code in the project

## How is it done?
Deleting the old calendar selection that was used in the calendar view. Removed a check in the settings that was not showing the calendar settings.

## Review hints
test the app with a fresh install and go to settings to enable the calendars.

## :squid: Permissions
🦑 cal pals


Toggl.Core.UI/Toggl.Core.UI.csproj
```diff
@@ -24,6 +24,9 @@
   <ItemGroup>
     <Service Include="{508349b6-6b84-4df5-91f0-309beebad82d}" />
   </ItemGroup>
+  <ItemGroup>
+    <Compile Remove="ViewModels\Calendar\SelectUserCalendarsViewModel.cs" />
+  </ItemGroup>
```
arturs105: I think we can even remove
* all of the `SelectUserCalendarsViewModel.cs` and related tests.
* `SelectUserCalendarsViewModelBase` and refactor the used bits to the `CalendarSettingsViewModel`, since we don't need 2 different calendar picker dialogs anymore

Toggl.Core.Tests/Toggl.Core.Tests.csproj
```diff
@@ -32,4 +32,7 @@
       <HintPath>..\packages\NUnit.2.6.3\lib\nunit.framework.dll</HintPath>
     </Reference>
   </ItemGroup>
+  <ItemGroup>
+    <Compile Remove="UI\ViewModels\SelectUserCalendarsViewModelTests.cs" />
```
heytherewill: We should not have Remove directives. If the file no longer exists, this is not needed. If the file exists, this should not be here.

Toggl.Core.UI/Toggl.Core.UI.csproj
sofokli1: 1. fixed
2. will make an issue for now, and see if there is still time later

Toggl.Core.Tests/Toggl.Core.Tests.csproj
sofokli1: fixed

Toggl.Core.UI/Toggl.Core.UI.csproj
sofokli1: for reference https://github.com/toggl/mobileapp/issues/6130

PR #6127, opened 2019-10-14
Handle API errors in the iOS widget

### LINKED ISSUE #6079: Handle API errors in the widget
When we get an API error in the widget, show the correct message in the UI. Designs can be found in https://app.zeplin.io/project/5728ce853dd1b182475f25e9/screen/5d23165906494a2e2552ed27/

When the user taps on the widget, launch the app to continue the action from there.
Hide the message when the widget appears again.

dshr:
## What's this?
A set of changes that allow us to gracefully handle API errors in the widget.

### Relationships
Closes #6079

## Why do we want this?
No 💥, only ✨ 

## How is it done?
By wrapping out network calls into `try..catch`es and showing the error view when there's an exception.

### Why not another way?
🤷‍♂ 

## Review hints
Try throwing an error from our NetworkHandler or just cut your internet connection and try stopping a running TE.

## :squid: Permissions
🎛 crew


Toggl.iOS.TimerWidgetExtension/TodayViewController.cs
```diff
@@ -69,39 +69,52 @@ public override void ViewDidDisappear(bool animated)
...
+                    projectName: null,
+                    projectColor: null,
+                    taskName: null,
+                    clientName: null
+                );
+                await networkingHandler.StartTimeEntry(timeEntry);
+                SharedStorage.Instance.SetRunningTimeEntry(timeEntry);
+                renderRunningTimeEntry(timeEntryViewModel);
+            }
+            catch
```
heytherewill: Should this really be a Pokémon exception? Should you not handle different errors differently?//

ping @juanlaube

Toggl.iOS.TimerWidgetExtension/TodayViewController.cs
```diff
@@ -159,6 +172,7 @@ private void renderErrorState(string error)
         {
             ErrorMessageLabel.Hidden = false;
             ErrorMessageLabel.Text = error;
+            ErrorMessageLabel.SetLineSpacing(2, UITextAlignment.Center);
```
heytherewill: Magic constant

Toggl.iOS.TimerWidgetExtension/TodayViewController.cs
juanlaube: Makes sense to show the network error message for all `ApiException`s and have another message for any other kind of exception, we don't have a copy though.

dshr: Check it out now, I've added some copy on my own, but we might need to discuss the way we handle the no running TE case (like, the TE has been stopped from Web, but the user hasn't opened the app, so SharedStorage still thinks it's running). We can either show a message, or just render the UI as if the TE has been stopped — since there's no more running TEs already.

A similar problem can happen if the user with a running TE starts a new one from the web, and then tries to stop it from mobile — we won't be stopping the SharedStorage TE, but the one currently on the server, which will def bring some confusion to the user.

PR #6144, opened 2019-10-16
🐛 Infinite syncing bugfix

### LINKED ISSUE #5939: User unable to log out
## ⛑️ User Support

A user reached out with the usual syncing issues when logging out was suggested he was unable to. They said that when they tap the logout button the app would say it's logging out but would hang on that step. 

They have since decided to uninstall and reinstall the app but still, have issues logging out. When they reached out they were on iOS 12.4 and now they are on iOS 13

## About the user

App version: Daneel/2.2.1
Phone OS: iOS 13
Phone model: iPhone SE

simonrozsival:
## What's this?
This is a bugfix for a syncing problem which we have been observing for the couple of last months. It manifested itself by looking as if the app got stuck syncing.

### Relationships
Closes #5939

## Why do we want this?
We want our app to work seamlessly even for users, who have many TEs.

## How is it done?

Syncing got stuck in the `PersistListState` state for time entries. This state calls the `ObservableDataSource.BatchUpdate` method, which stores all the data pulled from the server in the database (with conflict resolution, in a single transition) and then it emits a message through three observables `Deleted`, `Updated`, `Created` with the ID and additional data of the TE which changed. These three observables were then merged into a single stream of `Unit`s, which were observed by the `ObserveTimeEntriesVisibleToTheUserInteractor` interactor, which internally calls `GetAllTimeEntriesVisibleToTheUserInteractor`. This interactor queries the database and fetches all of the visible time entries and enumerates them into an array. This forces Realm to load all the data from the database into memory, we then convert each entity into a plain C# object (POCO). The conversion also requires querying the database for the linked workspace, the user, the project (if any), the task (if any), and the tags (if any), and converting them into POCOs as well. The dependencies of each of these linked entities are loaded and converted as well (each of the entities requires fetching the workspace again; there are several levels of dependencies, e.g., time entry -> project -> client -> workspace). Once all this is loaded, this data is diffed with the current time entries log. It is a good thing that we throttle the UI updates, because otherwise the UI would be overwhelmed as well (at least on iOS). For `N` time entries fetched from the server, this whole process would be repeated `N` times. For `N` time entries fetched from the server, this whole process would be repeated `N` times, and only the last change actually propagates to the UI.

For all of this time, the state machine is stuck in the `PersistListState`. Only when the whole `N` repetitions of the process are finished, we continue to the next state, which is `UpdateSinceDateState`. If the app is killed before it reaches this state, the next time it is started it will repeat the whole process of fetching those `N`. Maybe even more, because the user can keep tracking on a different device or in a different client app because the mobile app didn't work for them.

I decided to get rid of the `Created`, `Updated`, and `Deleted` observables and use only a single `ItemsChanged` observable property (not an extension method anymore). This means that we can now emit just 1 event per `BatchUpdate`. We don't need any of the data in the `Update/Delete/Create` observables' metadata, because we load the visible TEs in the main log through the `GetAllTimeEntriesVisibleToTheUser` (or something like that) and we ignore the metadata anyway. The only place slightly affected by this change is the running TE observable, but that was very easy to change (and I think this way it's still very  effective).

### Side effects
I don't think there are any side effects. But please review this PR carefully.

## Review hints
Use this bash script to create many TEs in your staging account:

```bash
for i in {1..1000}
do

curl -X POST \
  https://mobile.toggl.space/api/v9/workspaces/<workspace id>/time_entries \
  -H 'Authorization: Basic <base64(email:password)>' \
  -H 'Content-Type: application/json' \
  -d '{
	"description": "",
	"workspace_id": <workspace id>,
	"start": "2019-10-14T13:04:50+00:00",
	"duration": 1,
	"tag_ids": [<tag id>],
	"project_id": <project id>,
	"at": "2019-10-14T13:04:50+00:00",
	"user_id": <user id>,
	"created_with": "postman"
}'

sleep .5

done
```

Feel free to ask me for help if you have trouble getting this to work.

## :squid: Permissions
I'd rather do it myself once I'm sure this has received enough reviews.


amulware:
I don't have the time to review the code, but from the explanation, I have a question:

Doing lots of redundant work was an issue, but so was the UI blocking sync in the first place - if it hadn't, sync would have finished fast, and at worst a simple restart of the app would have made the UI refresh fast. After these changes is the UI still blocking sync? I really think it shouldn't. Sync is a black box that should run on its own and not be affected by anyone observing it.

PR #6156, opened 2019-10-16
Fix widget layout when resizing to 1x1

### LINKED ISSUE #6136: Starting a time entry from a 1x1 widget shows incorrect layout
The widget enters in a weird state when starting a time entry from the 1x1 layout, it shows the extended 1xN layout but in 1x1, and it doesn't look good.

Steps to reproduce:
- Don't have any running time entry
- Resize the widget from 1xN to 1x1
- Start a time entry from the widget

juanlaube:
## What's this?
Fixes a layout problem when starting a time entry after resizing to a 1x1 widget.

### Relationships

Closes #6136 

## Why do we want this?
We don't want bugs.

## How is it done?
Considering the widget dimensions when updating the content.

## :squid: Permissions
🦑


Toggl.Droid/Widgets/TimeEntryWidget.cs
```diff
@@ -53,7 +53,9 @@ public override void OnUpdate(Context context, AppWidgetManager appWidgetManager
 
             foreach (var widgetId in appWidgetIds)
             {
-                updateWidget(context, appWidgetManager, widgetId);
+                var options = appWidgetManager.GetAppWidgetOptions(widgetId);
```
kornelijepetak: This was initially my idea, but then I found out that there are calls to `OnUpdate` that do not have any dimensions attached to the intent, such as what is guaranteed in `OnAppWidgetOptionsChanged`.

PR #6158, opened 2019-10-16
💚 Use a separate firebase config for staging

### LINKED ISSUE #5951: Figure out production/stating Firebase configuration
Currently we only one Firebase project for our apps's configurations.

Ideally, we should have different api keys for production and staging environments.
Using the current configuration, I couldn't find a way to generate a new api key that would isolate production apps from staging/debug apps.

The way I see we have two options:
1 - Contact Firebase support and ask to make sure there isn't a way to create an isolated api key without creating another firebase project.
2 - Create another firebase project for staging (and use the current one as production).

Pinging @amulware ➿ and _🔨 (loop & decisions).

pdrj:
## What's this?
This makes adhoc builds use a separate firebase project.

### Relationships

Closes #5951 

## Why do we want this?
It makes it easier to test stuff, for instance remote config.

## How is it done?
Adding new secrets to bitrise, using them in cake, setting up apps in firebase. I wanted to touch as little of current workflows as possible (to not break current adhoc builds by modifying vars in bitrise).

### Why not another way?
🤷‍♂ 

## Review hints
Please see if it the code makes sense and maybe for Android also try to use [this build](https://toggl.slack.com/archives/C02APMYTM/p1571232876020700)

I couldn't login with this apk for some reason, but @semanticer could... I also did a playstore build and it worked fine.

## :squid: Permissions
Me


pdrj:
I've been able to test an Android prod apk too - works fine.

pdrj:
Adhoc iOS tested and working as well (it pulls the staging config) 
https://app.bitrise.io/build/aecb651477ecad34#?tab=artifacts

build.cake
```diff
@@ -153,18 +153,29 @@ private TemporaryFileTransformation GetAndroidProjectConfigurationTransformation
     };
 }
 
+private string GetStagingAwareEnvironmentVariable(string varName)
+{
+    if(target.ToUpper().Contains("ADHOC"))
```
simonrozsival: I'm afraid this could break in the future. I have two suggestions:
- create a variable `var isStaging = <condition>;` on line 15 of `build.cake`
- could you list all the current targets, which would be currently affected by this? I'm not sure we always follow the "rule" that it contains adhoc or that we will follow this unwritten rule in the future and then it'll stop working for some reason. We should at least document this somewhere 🤔 Maybe we should describe how this works in https://github.com/toggl/mobileapp/blob/develop/docs/env-vars.md - well we definitely should add documentation for this behavior here! So please add docs.

PR #6159, opened 2019-10-16
iOS 2.5 Fig

Odrakir:
# User Facing
- Added support for dark mode
- Added shake animation to email field when an error happens in the password forgotten screen
- Fixed a bug that could lead the app to sync indefinitely
- Fixed pie chart labels appearing upside down
- Fixed not being able to cancel confirmation dialogs
- Fixed layout issues in suggestion cells
- Fixed possible crash in the Calendar
- Fixed system's autocomplete and project suggestions clashing
- Fixed start and edit modal windows layouts on iPad
- Fixed color of project dot in Timeline cells
- Fixed crash when syncing in background
- Fixed glitch when layout suggestions

# Internal
- Re-enabled the linker, so app size should be smaller
- Added tracking for the number of user calendars
- Dropped support for iOS 10

# Changes
https://github.com/toggl/mobileapp/compare/ios-2.4-2...ios-2.5-fig


amulware:
Couple of notes about the changelog:
- Reorder to have all "Added" points first
- Let's leave 'Fixed rating view not showing' out because it's not really a user-facing feature

Otherwise looking good :)

PR #6202, opened 2019-10-17
Dark mode: fix reports for iPad

### LINKED ISSUE #6190: Dark Mode: Reports project list has white border on iPad
IMG_FA732C69042C-1

simonrozsival:
## What's this?
UI fix for reports.

### Relationships
Closes #6190 

## Why do we want this?
We want nice iPad dark mode.

## How is it done?
The usual way.

### Why not another way?
I tried figuring out the shadows in the light mode, but I wasn't successful so far. I think it's a separate issue, because there were no shadows in the light mode up to now anyway.

## Review hints
iPhones still look good:
image
image

iPads are now fixed:
image
image

## :squid: Permissions
🦑


Toggl.iOS/Views/Reports/ReportsHeaderView.cs
```diff
@@ -50,6 +50,9 @@ public override void AwakeFromNib()
 
             EmptyStateTitleLabel.Text = Resources.ReportsEmptyStateTitle;
             EmptyStateDescriptionLabel.Text = Resources.ReportsEmptyStateDescription;
+
+            var iPad = TraitCollection.HorizontalSizeClass == UIUserInterfaceSizeClass.Regular;
```
Odrakir: Not really `iPad`. Portrait mode on small iPads, or any split view are of `Compact` horizontal size class and still is an iPad. If you want a variable name call it `regularWidth` or something like that.

PR #6204, opened 2019-10-17
Fix crash in settings view

heytherewill:
## What's this?
This fixes a tiny crash in the settings page

## Why do we want this?
Crashes no good

## How is it done?
🚗 

## :squid: Permissions
Anyone, anytime


Toggl.Core.UI/ViewModels/Settings/SettingsViewModel.cs
```diff
@@ -200,11 +200,13 @@ public sealed class SettingsViewModel : ViewModel
                     .AsDriver(false, schedulerProvider);
 
             IsCalendarSmartRemindersVisible = calendarPermissionGranted.AsObservable()
-                .CombineLatest(userPreferences.EnabledCalendars.Select(ids => ids.Any()), CommonFunctions.And);
+                .CombineLatest(userPreferences.EnabledCalendars.Select(ids => ids.Any()), And)
+                .AsDriver(false, schedulerProvider);
```
kornelijepetak: Considering that we have an overload with a `default(T)`, you can omit first argument.

Toggl.Core.UI/ViewModels/Settings/SettingsViewModel.cs
```diff
@@ -197,16 +197,16 @@ public sealed class SettingsViewModel : ViewModel
...
             IsCalendarSmartRemindersVisible = calendarPermissionGranted.AsObservable()
                 .CombineLatest(userPreferences.EnabledCalendars.Select(ids => ids.Any()), And)
-                .AsDriver(false, schedulerProvider);
+                .AsDriver(schedulerProvider);
 
             CalendarSmartReminders = userPreferences.CalendarNotificationsSettings()
                 .Select(s => s.Title())
                 .DistinctUntilChanged()
-                .AsDriver("", schedulerProvider);
+                .AsDriver(schedulerProvider);
```
kornelijepetak: Be careful with this change. This is not equivalent. It used to be `""` but with your latest change, `default(string)` is `null`. Consider whether it's ok to propagate `null` value up to the subscribers...

Issue #5867, opened 2019-09-10
Calendar: Improve manual time entry creation


sofokli1:
Manual TEs
- [x] Use long press to create new Entry with handlers and contextual menu appear.
- [x] New entries should snap to quarter hour
- [x] Should have a default duration of 30 min
- [x] If created in the gap between 2 TEs that is 2 hours or less, it should fill the whole gap by default

- [x] New `CalendarTimeEntryCreated` event with 3 parameters (even may be there already but not support the below type)
    - `Type` string with possible values from the following list
        - `LongPress` (triggered when TE is created with long pressed)
    - `DaysSinceToday` int. Difference in days between the start date of the TE and today. If negative it is in the past, if positive it is in the future (with our  implementation all the values should be negative since we do not allow swiping to future dates, for now)
    - `DayOdTheWeek` string. The name of the day of the week

Designs:
https://app.zeplin.io/project/5728ce853dd1b182475f25e9/screen/5d88b3c9f1714261f39bed5a

arturs105:
I think copying the iOS calendar behaviour would work really well, since most users would already be familiar with it.
The way iOS calendar works:
1. 3D touch on an empty space in the calendar (long press works as well, probably for phones without 3D touch)
2. An hour long entry appears, that is immediately draggable
3. After releasing the finger, the entry sets in place and some editing options appear

I think this behaviour would be better than the one we have now - tapping and holding to create an entry with fixed start time at the point of touch, and then allowing the end time to be modified.

sofokli1:
Main issue with 3d touch and long press is that it is not very intuitive since a lot of users do not know those exist (it is not obvious those options are there). Let's wait for designs and talk agin about it at that point.

PR #6205, opened 2019-10-17
📆 Improve manual Time entries

sofokli1:
## What's this?
Sets default duration of manual TEs to 30min. also if the gap between 2 TEs is less than 2h it will fill that gap instead of the 30 min default duration.

### Relationships
Partially deals with #5867

## How is it done?
The 30 min duration s it is simple constant change.
The 2h gap is done by first calculation all the available gaps that are 2h or less and check if the long press was started in one of those. if it is the start and duration of the new TE are overridden with the start and duration of the gap.

### Why not another way?
didn't think of any

### Side effects
none

## Review hints
run the app

## :squid: Permissions
📆 pals only


Toggl.iOS/ViewSources/CalendarCollectionViewSource.cs
```diff
@@ -183,6 +184,55 @@ public List<DateTimeOffset> AllItemsStartAndEndTime()
             return startTimes.Concat(endTimes).ToList();
         }
 
+        public List<(DateTimeOffset, TimeSpan)> GapsBetweenTimeEntriesOf2HoursOrLess()
```
daividssilverio: Hey, could you put this in the shared layers and add some tests? Something like what we have for `CalendarLayoutCalculator`.
That way we could just use it on droid directly and profit :) That would be real cool.

PR #6206, opened 2019-10-17
Swipe down to dismiss fix

### LINKED ISSUE #6181: Cancelling dismissing edit view by swiping breaks UI
Steps to reproduce:
- edit a time entry
- make any change (e.g., change description)
- drag the view down to dismiss it
- select "Continue editing" when you're prompted
- the bug: edit view does not go back to full size but it stays as small as it was when you dragged it down

heytherewill:
## What's this?
This fixes the swipe to dismiss bug

### Relationships
Closes #6181

## Why do we want this?
Users can change their minds

## How is it done?
I'm making close cancellable and each ViewModel can implement said cancellation as it pleases.

## :squid: Permissions
Anyone


Toggl.Core.UI/ViewModels/EditTimeEntryViewModel.cs
```diff
@@ -436,26 +436,30 @@ private void dismissSyncErrorMessage()
...
-                if (view != null)
-                {
-                    var userConfirmedDiscardingChanges = await view.ConfirmDestructiveAction(ActionType.DiscardEditingChanges);
-                    if (!userConfirmedDiscardingChanges)
-                        return;
-                }
+                if (view == null)
+                    return true;
+
+                    return await view
```
Odrakir: Indentation

Odrakir: Indentation

Odrakir:
As the change I requested was so minor, I did it my self so we don't have to wait for Will to be online.

Issue #6196, opened 2019-10-17
Rating view in dark mode


Odrakir:
The rating view doesn't have a dark mode version.

![IMG_2FEB66C14B25-1](https://user-images.githubusercontent.com/724743/67015223-7fe75980-f0f6-11e9-8d64-ee796714d7ca.jpeg)

simonrozsival:
Fixes were merged into 2.5 release branch.

PR #6214, opened 2019-10-18
Dark mode: Fix rating view background color

simonrozsival:
## What's this?
I was too trigger happy when I was reviewing #6199 and I didn't check how it looks in the light mode. I tested it after I merged the branch and I realized that there's a problem with the white vs. gray background.

### Relationships
Ref #6196, ref #6199.

## How is it done?
I made the background of the rating view transparent and so there are no more problems.

## Review hints
Before:
image

After:
image

## :squid: Permissions
🦑


Odrakir:
Good catch!

PR #6216, opened 2019-10-18
Tap widget-based suggestion to start a new TE

### LINKED ISSUE #5407: Handle the continue action using intents
Use intents to handle the start from suggestion action.

Depends on #5406

kornelijepetak:
## What's this?
Continues time entries by tapping suggestions.

## Issues
I'd like this merged and I suggest we fix the bugs in a separate issue.

1. This does not completely solve the issue, but I wasn't able to find out the cause. What happens is that starting a new time entry from a widget takes a bit more of time, causing the UI to not be responsive. 

2. Sometimes, even after the tap, the widget does not start a new time entry. 

3. When the app is not open, the suggestions don't create anything at all. Ever.

### Relationships
Closes #5407 

## Why do we want this?
So that suggestions actually make sense.

## How is it done?
Android shenanigans.

### Why not another way?
There's no other way with droid lists on widgets.

## Review hints
I need a proper functional test. Try to add both widgets and then play with suggestions.

## :squid: Permissions
🎛 guys.


juanlaube:
Looks good to me.
About the problems you mention, I cannot reproduce 1 and 2, but I was able to reproduce 3. Lets find a solution and handle that in a different issue.

kornelijepetak:
@juan the truth is, I've experienced 1 and 2 only on emulator with debugger attached.
I don't see the same issues on my phone (Debug build).

The 3rd one is observable everywhere.

PR #6218, opened 2019-10-18
Fix main log header layout

### LINKED ISSUE #5909: Layout messed up on startup before first sync
## 🐛 Bug Report
When starting the app (not when opening the app from background), the time log view layout is messed up until the first sync completes. Suggestions seem to overlap the time entries. 

## Context

IMG_3632

Build/Branch: Daneel 2.3
Phone OS: iOS 12.4.1
Phone model: iPhone 7

## Expected behavior
No overlap of different sections. 

## To reproduce
Start the app from scratch (not from background)

Odrakir:
## What's this?
Fixes the weird glitch with the suggestions that was happening for a second when the app started.

### Relationships
Closes #5909 

## How is it done?
I think most of it comes down to not knowing the height of suggestions until those load, and then not reacting properly to that. I've extracted the logic into a method that gets called once when the header is setup and another time when suggestions change. 

The method itself is a little weird, but judging by what I read online, this is what works with UIKit.

## :squid: Permissions
Do it!


Toggl.iOS/ViewControllers/MainViewController.cs
```diff
@@ -371,6 +375,19 @@ private void setupTableViewHeader()
...
+
+        private void layoutTableHeader()
+        {
+            TimeEntriesLogTableView.TableHeaderView = tableHeader;
+            tableHeader.SetNeedsLayout();
+            tableHeader.LayoutIfNeeded();
+            var frame = tableHeader.Frame;
+            frame.Size = tableHeader.SystemLayoutSizeFittingSize(UIView.UILayoutFittingCompressedSize);
+            tableHeader.Frame = frame;
+            TimeEntriesLogTableView.TableHeaderView = tableHeader;
```
simonrozsival: This method is super confusing 😆 at least to the same extend as it was before. Could you maybe add a comment with a link to some source or at least describe what this method does and why we added it here so we don't mess it up with future rewrites of the app?

Toggl.iOS/ViewControllers/MainViewController.cs
```diff
@@ -371,6 +375,19 @@ private void setupTableViewHeader()
 
             suggestionsContaier.AddSubview(suggestionsView);
             suggestionsView.ConstrainInView(suggestionsContaier);
+
+            layoutTableHeader();
+        }
+
+        private void layoutTableHeader()
```
simonrozsival: I think this method should be called also whenever we show or hide the rating view.

PR #6259, opened 2019-10-21
Dark mode: outdated app view

### LINKED ISSUE #6258: Dark mode: `OutdatedAppViewController`
Turn the view dark.

simonrozsival:
### Relationships
Closes #6258 

## Review hints
To force the app to show the view:
- go to settings
- long press the "Settings" title at the top
- select either "Outdated client error" or "Outdated API error"
- the `OutdatedAppViewController` will open

This is how it'll look like:
image
image

## :squid: Permissions
🦑

### DIFF
--- a/Toggl.iOS/Presentation/RootPresenter.cs
+++ b/Toggl.iOS/Presentation/RootPresenter.cs
@@ -22,14 +22,19 @@ namespace Toggl.iOS.Presentation
             typeof(OutdatedAppViewModel),
         };
 
+        private HashSet<Type> viewModelsNotWrappedInNavigationController { get; } = new HashSet<Type>
+        {
+            typeof(MainTabBarViewModel),
+            typeof(OutdatedAppViewModel),
+        };
+
         public RootPresenter(UIWindow window, AppDelegate appDelegate) : base(window, appDelegate)
         {
         }
 
         protected override void PresentOnMainThread<TInput, TOutput>(ViewModel<TInput, TOutput> viewModel, IView view)
         {
-            var shouldWrapInNavigationController = !(viewModel is MainTabBarViewModel);
-            var rootViewController = shouldWrapInNavigationController
+            var rootViewController = !viewModelsNotWrappedInNavigationController.Contains(viewModel.GetType())
                 ? ViewControllerLocator.GetNavigationViewController(viewModel)
                 : ViewControllerLocator.GetViewController(viewModel);
 
--- a/Toggl.iOS/ViewControllers/OutdatedAppViewController.cs
+++ b/Toggl.iOS/ViewControllers/OutdatedAppViewController.cs
@@ -18,6 +18,8 @@ namespace Toggl.iOS.ViewControllers
         {
             base.ViewDidLoad();
 
+            NavigationController.NavigationBar.Hidden = true;
+
             HeadingLabel.Text = Resources.Oops;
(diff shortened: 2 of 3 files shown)


Toggl.iOS/ViewControllers/OutdatedAppViewController.cs
```diff
@@ -18,6 +18,8 @@ public override void ViewDidLoad()
         {
             base.ViewDidLoad();
 
+            NavigationController.NavigationBar.Hidden = true;
```
Odrakir: 🤔 I see why you are doing this, but I don't know why this is presenting with the `RootPresenter` instead of the `ModalDialogPresenter`, like most of the other errors do. I think we should change that for `TokenResetViewModel` and `OutdatedAppViewModel`.

Toggl.iOS/Presentation/RootPresenter.cs
```diff
@@ -22,14 +22,19 @@ public sealed class RootPresenter : IosPresenter
             typeof(OutdatedAppViewModel),
         };
 
+        private HashSet<Type> viewModelsNotWrappedInNavigationController { get; } = new HashSet<Type>
```
Odrakir: I don't get why you do this... 🤔 Why don't you just remove the `typeof(OutdatedAppViewModel)` from `RootPresenter` and put it in `ModalDialogPresenter`?

simonrozsival: It's not a modal dialog, you can't dismiss it. Once you get in this state, your app is stuck like that until you update it, so it should be a root VC (it just looks like a modal dialog).

Odrakir: But modal dialogs can't be dismissed unless you add a way to do it, like a button. And you avoid having to do special cases.

Odrakir: Also, it won't look like a modal dialog if it completely replaces what's on the background.

simonrozsival: OK. I'll try it.

simonrozsival: I actually think we're trying to solve a different problem here. Only the `TokenResetViewModel` should actually be wrapped in the navigation VC, because login and signup hide the navbar as well.

When the app starts and the app is in the "blocked" state, it navigates directly to the `OutdatedAppViewModel/Controller` - can our model dialog be the root VC like this? I think this could throw and crash:
image

simonrozsival: (the code is from `ModalDialogPresenter.cs`)

simonrozsival: And I actually think that the `TokenResetViewController` shouldn't be wrapped in the navigation VC, which would fix this glitch: 
image

So in the end, none of the root VCs should be wrapped in the navigation VC. Maybe we should do it in a separate PR though.

Odrakir: Yeah, that probably should go into another release.

I don't understand what you are saying about the code that could crash... 🤔 I think the outdated app should appear over any view the app shows right before we do that check.

simonrozsival: OK. Don't listen to me. They have to be navigation VCs (to _navigate_ to other subviews).

PR #6271, opened 2019-10-21
Merge back develop into project/widgets

dshr:
## What's this?
A mergeback, so we can use all the colour goodness from Dark Mode.

### Relationships
Nothing

## Why do we want this?
🕶 

## :squid: Permissions
🎛 folx, and make sure it's a merge and not a squash


dshr:
Should've branched first 🤦‍♂

Issue #6229, opened 2019-10-18
Intents sent from widgets not working when app is not running


juanlaube:
When the app is killed manually by the user, the pending intents sent from the widgets are not being handled, a google search shows that in this case the app is not started again, and the intents will be handled only after the app launches again, either by the user or in background.

We need to find a solution or propose an alternative. RIght now I can only think of using a foreground service, but that's not ideal because a persistent notification will be visible in the notification area.

Note: this doesn't seem to be a problem when the app is killed by the OS, in this case it is restarted and the intents are handled correctly.

juanlaube:
Something that doesn't solve the problem entirely, but will help, is to reduce the `updatePeriodMillis` in the widget providers to 30 mins (this is the minimum time allowed), this will help us to relaunch the app in background and be able to handle those intents.

PR #6279, opened 2019-10-21
Use foreground service to handle widget actions

juanlaube:
## What's this?
Replaces the broadcast receiver with a foreground service to handle all the widget actions so the widget can be used even after the user kills the app.

### Relationships

Refs #6229 

## Why do we want this?
We want the widget to be useful, all the time.

## How is it done?
Using a foreground service.

### Why not another way?
This is the only way

### Side effects
When the app is not running in background, the actions are a bit slow because the app has to start. In a second branch, I tried to show an indeterminate progress bar when the user taps on the start/stop button, but even that is a slow because it needs the app to be running to update the widget, I will push that if I can make it work better somehow.

## :squid: Permissions
🦑


juanlaube:
@semanticer thanks for the feedback, I've found the problem that caused the flickering, it's solved now in this branch.
As for that search result, I can't find the link now (classic), but it was a stackoverflow answer that said the app is not started when sending a pending intent from the widgets if the user kills the app manually, it does only if the app is killed by the system.

PR #6280, opened 2019-10-21
🚧 Fix layout when changing size class

### LINKED ISSUE #6154: Fix calendar view for iPad
How the iPad layout should work:
1. In portrait mode the calendar view fills the whole screen width (minus any padding)
2. In landscape mode the calendar should take up the same amount of space as it would in portrait mode.
3. In split screen mode (when size class is compact) the week view headers should have only the initial letter

sofokli1:
## What's this?
Fixes the insets of the calendar day view for the different size classes and the weekly date picker

### Relationships
mostly deals with #6154

## Why do we want this?
we want it to look good but with simple code

## How is it done?
some simple inset calculation code and reload of the weekly date picker when the layout changes

## Review hints
test it and try to break it

## :squid: Permissions
📆 pals


arturs105:
There seems to be a problem with the week view:
1. Start in portrait mode
2. Rotate to landscape
3. Rotate back to portrait
4. Now the weekly picker takes up all the screen width - in step one there was some padding on the sides.

Also: seems like like the padding on the sides has increased on iPhones.

Also also: i didn't manage to test the app in split screen view, since the gestures aren't really working on the simulator.

sofokli1:
@arturs105 For split view do `CMD + SHIFT + H` twice to open the UI to that shows previous open apps. from the bottom bar select safari and drag it on the toggl app. wait over the app with the safari icon until the toggl app opens and now you can drop the safari app on the side.

for the other issues you mention can you make a video and screen shot for the second one?

arturs105:
Everything seems to be working fine apart from the problem that i described in the previous comment. Here's a recording (pay attention to the space between the screen edge and "Monday"):
Oct-25-2019 11-54-39

sofokli1:
@arturs105 i changed the description at the top not to close the issue so we can merge this as is. On my simulator i can't not reproduce the issue you mention.

Toggl.Core.UI/ViewModels/Calendar/CalendarViewModel.cs
```diff
@@ -86,7 +90,11 @@ public sealed class CalendarViewModel : ViewModel
                 .Select(days => days.ToImmutableList())
                 .AsDriver(schedulerProvider);
 
-            WeekViewHeaders = beginningOfWeekObservable
+            WeekViewHeaders = Observable
+                .CombineLatest(
+                    beginningOfWeekObservable,
+                    realoadWeekView.AsObservable(),
```
heytherewill: There's a `ReemitWhen` operator that would make this more readable

PR #6290, opened 2019-10-22
Fix undeletable time entries problem

### LINKED ISSUE #6289: A time entry in our app can become "undeletable"
When our user has a TE in our app, which does not exist on the server, it becomes "undeletable". An attempt to delete the TE ends with the TE reappearing and the error message in the edit view says `Oops. Time entry not found.`.

This message is confusing for the users and the ux is not good. If the TE is not on the server and the user wants to delete it, then we should just delete it in our app and ignore the server error. I'm not sure how we would check that this specific "error" can be ignored and treated as a "success". We might want to check if this error has some specific numeric code (errors messages now have error codes, right? is it already in production?) which we could hardcode in our app.

The ideal way of solving this issue is to figure out how the app got into the state where there is a TE in the app's database which has a server-assigned (positive) ID, but does not exist on the server. The particular user who reported this problem has a single workspace and is the only user in this workspace. This shouldn't be related to "losing access to a workspace". We must figure out how to reproduce this without messing with the data in Realm Studio.

Intercom link: https://app.intercom.io/a/apps/ayixs927/inbox/inbox/conversation/24064700153

simonrozsival:
## What's this?
This is a fix for an issue which was reported by at least one user. The core of the problem was that if a time entry is deleted using API v9 and we attempt to delete it again, we get a 404 error that the time entry was not found and we treat that as marking the time entry as unsyncable and showing it to the user.

### Relationships
Closes #6289 

## Why do we want this?
We want the delete operation to be idempotent. When we send the DELETE reqeust, but we don't receive a response because we lose connection with the server or the app is killed on the device, we want to be able to remove the TE from the local database later.

## How is it done?
I simply catch the `NotFound` API error exception for the `Delete` operation of the Time Entries API client class and handle it as if the request was successful. This makes sense, because if the time entry does not exist on the server and the user wants to delete it in the app, we should just delete it from the app. If the database is consistent, then there should never ever be a time entry with that ID which would be possible to delete.

### Why not another way?
It could be handled in the interactor or in the syncing algorithm, but that doesn't seem right to me. The way I implemented it hides an ugly behavior of our API, the business logic does not have to know about how we interpret status codes of HTTP responses.

### Side effects
We cannot distinguish deleting already deleted TE from deleting a non-existing random TE in the user's workspace. I don't think that's a problem.

## Review hints
Pay attention to the integration tests.

## Tests
Integration tests: https://app.bitrise.io/build/dc5fa0f20fb23778 - do not merge this branch unless these tests are passing.
Sync tests: https://app.bitrise.io/build/8fac857cb5210382

## :squid: Permissions
Leave it up to me.


simonrozsival:
The regular case is covered by integration tests, so I won't bother for now :D I hope we'll be able to get rid of the whole endpoint soonish.

PR #6308, opened 2019-10-23
Fix siri reports shortcut

### LINKED ISSUE #6254: Reports Siri Shortcuts don't work
They are supposed to open the app and show the reports tab with the selected interval, right now they just open the app and do nothing.

simonrozsival:
## What's this?
There was a bug, I think I fixed it.

### Relationships
Closes #6254

## How is it done?
I was debugging the app and I noticed that the report intents get thrown away at `AppDelegate.Siri.cs:28` because `interaction.IntentHandlingStatus` was `Unspecified`. It used to be `DeferredToApplication` but something changed (iOS 13? Xcode 11? Xamarin?) and now that doesn't hold anymore.

### Why not another way?
🤷‍♂ 

### Side effects
I hope none.

## Review hints
Trust me or try this on your phone (if you can).

## :squid: Permissions
🦑


Toggl.iOS/Startup/AppDelegate.Siri.cs
```diff
@@ -23,7 +23,7 @@ public partial class AppDelegate
             var navigationService = IosDependencyContainer.Instance.NavigationService;
 
             var interaction = userActivity.GetInteraction();
-            if (interaction == null || interaction.IntentHandlingStatus != INIntentHandlingStatus.DeferredToApplication)
+            if (interaction == null)
```
Odrakir: If I remember correctly this runs when you tap the Siri UI. I think the problem here might be that when tapping an intent that's already finished, and should just open the app, we might execute it again if you don't do that check. How about, instead of this, adding the case for `INIntentHandlingStatus.Unspecified` just in case?

simonrozsival: The docs say to check if the `IntentResponse` is null or not and in case of it being null, skip processing it for the second time. Basically what I did with the first commit (before I decided rather to not do it). If you think adding `Unspecified` as an extra case, I can do it.

PR #6310, opened 2019-10-23
🐛 Fix select default workspace fragment being empty

### LINKED ISSUE #6268: Select default workspace view is empty
## 🐛 Bug Report
The view we present to the user in case they have no default workspace selected is rendering empty.

## Context

Build/Branch: 2.5
Phone OS: Android
Phone model: XT1092

## Expected behavior
User being presented with workspaces, from which they can choose a default one.

## To reproduce
Use a debug build, long press on 'About' in settings. Trigger the 'no default workspace' view.

pdrj:
## What's this?
This makes the select default workspace view work and display data.

### Relationships

Closes #6268 

## Why do we want this?
Because for some users the bug makes the app unusable.

## How is it done?
- Setting correct string resources
- Fixing initialization

### Why not another way?
This one is pretty simple.

## Review hints
Do a debug build, long press on About, trigger the 'no default workspace error'. Alternatively, make it appear by modifying code in `MainViewModel`.

## :squid: Permissions
Me.

## 🤔 
What's interesting is that this has been broken for more than two months and hasn't been noticed.


semanticer:
Maybe nobody understood the test case correctly? You can submit PR to the testing template. @pdrj

pdrj:
Either that, or due to using a faster device the view had a workspace presented (because the viewmodel finished loading before presenting the fragment). The text still wouldn't have been visible though, but the test case doesn't say what to expect 😄 .

PR #6312, opened 2019-10-23
Implement analytics event for the weekly date picker

### LINKED ISSUE #5996: Calendar: Analytics event for the weekly date picker
### New `CalendarWeeklyDatePickerSelectionChanged` event with 2 parameter
    - [ ] `DaysSinceToday` int. The difference in days between the new date shown and today. If negative it is in the past, if positive it is in the future (with our  implementation all the values should be negative since we do not allow swiping to future dates, for now)
    - [ ] `DayOdTheWeek` string. The name of the day of the week

arturs105:
### Relationships

Closes #5996
## :squid: Permissions
🦑


arturs105:
It won't, because `SelectDayFromWeekView` action is not invoked when swiping left/right.

PR #6313, opened 2019-10-23
Implement analytics event when swiping the calendar

### LINKED ISSUE #5995: Calendar: Analytics event when swiping the calendar
### New `CalendarSingleSwipe` event with 3 parameters
    - [ ] `SwipeDirection` as a string, possible values:
        - `Left` when view scrolls to the left
        - `Right` when view scrolls to the right
    - [ ] `DaysSinceToday` int. Difference in days between the new date shown and today. If negative it is in the past, if positive it is in the future (with our  implementation all the values should be negative since we do not allow swiping to future dates, for now)
    - [ ] `DayOdTheWeek` string. The name of the day of the week

arturs105:
### Relationships

Closes #5995

## :squid: Permissions
🦑


sofokli1:
@arturs105 if i change the date from the weekly date picker won't the event be triggered?

arturs105:
I don't think it will be triggered, since it is called from a page view controller delegate method. 

However, i can't test this, since i can't build until XCode updates. And it may be a while, because i have to wait for it to download, then it fails to update without any useful error and i have to download it again (tfw no caching). I've tried to update it 3 times already.

PR #6314, opened 2019-10-23
Fix the "Tap to edit" onboarding bubble positioning

### LINKED ISSUE #6244: Firs entry onboarding hint is out of place
It's not pointing to the first entry, but to the top of the table.

IMG_B27D3D85D78A-1

simonrozsival:
## Relationships
Closes #6244 

## Why do we want this?
We want nicer UI.

## How is it done?
There are two components, which make this work better than it used to:
1. the bubble is not shown not only as the data is in the VM, but also when the data source receives the data and the first cell is created (line 742)
2. we adjust the position of the bubble when the data in the table changes

### Why not another way?
I'm not happy with the way we display and manage the onboarding tooltips, but I think it's good enough for now.

### Side effects
I'm not aware of any.

## Review hints
Remove the app from the simulator before you upload it from VS, otherwise you won't see the tooltips after signing up.

## :squid: Permissions
🦑


Toggl.iOS/ViewControllers/MainViewController.cs
```diff
@@ -128,7 +128,9 @@ public override void ViewDidLoad()
             TimeEntriesLogTableView.BackgroundColor = ColorAssets.TableBackground;
 
             ViewModel.TimeEntries
-                .Subscribe(TimeEntriesLogTableView.Rx().AnimateSections<MainLogSection, DaySummaryViewModel, LogItemViewModel, IMainLogKey>(tableViewSource))
+                .Do(TimeEntriesLogTableView.Rx().AnimateSections<MainLogSection, DaySummaryViewModel, LogItemViewModel, IMainLogKey>(tableViewSource))
```
heytherewill: I dislike this double `Do`. I'd rather you create a proper subscription that calls those two methods in a row

Odrakir: I agree, actually I think the subscription should be the table view animation, and the tooltip update should be a side effect.

simonrozsival: I agree. The "problem" is that the  original line `131` should stay, but the `Do(update...)` should come _after_ it. I'll try to do something about it.

simonrozsival:
@heytherewill I created a proper subscription, please approve now.

Issue #6264, opened 2019-10-21
Calendar Bottom sheet: Dark mode


sofokli1:
Add the appropriate colors and changes for dark mode. Talk with people that worked in PR #6053 for help.

daividssilverio:
🤔 I'm not sure how to improve this, I just know that the contextual menu doesn't look great in dark mode.

Any idea on how to improve this @AlicjaSuska

![image](https://user-images.githubusercontent.com/2173493/67421034-b3663000-f59d-11e9-8978-f544e2d69b6f.png)

daividssilverio:
Maybe @Odrakir and @simonrozsival could give their thoughts as well?

semanticer:
Since it should be very easy to implement, I'm adding this to Android 2.8 - @AlicjaSuska is now aware of this and should provide designs soon. If not, I will try to fix this myself - it shouldn't be too hard to improve this.

AlicjaSuska:
About the icons: I'm not sure what happened but those that are used for light designs look good on dark background. Maybe opacity is not 100%? Please, see the picture:

![Android - Calendar - Dark mode - Icons](https://user-images.githubusercontent.com/20326070/68851302-fcf5f680-06d5-11ea-8ee4-b7341352c880.png)

semanticer:
Ok, it looks like we did something wrong with the opacity - thanks I will look at it soon.

PR #6316, opened 2019-10-23
Improve dark mode on calendar view texts

daividssilverio:
## What's this?
This PR, at least for now only does one thing:
- It changes the text color of items in the calendar to use the primary text color (which adapts based on dark mode) if the contrast against the calendar item background is too low.

### Relationships
Closes Nothing, but part of #6264 

## Why do we want this?
So that the calendar looks nice in Dark mode.

## :squid: Permissions
🦑 📆 pals


Toggl.Droid/Views/Calendar/CalendarDayView.Events.cs
```diff
@@ -41,20 +41,22 @@ public partial class CalendarDayView
...
         private CalendarItemEditInfo itemEditInEditMode = CalendarItemEditInfo.None;
         private readonly CalendarItemStartTimeComparer calendarItemComparer = new CalendarItemStartTimeComparer();
         private int? runningTimeEntryIndex = null;
         private int editingHandlesHorizontalMargins;
         private int editingHandlesRadius;
 
+        private Bitmap calendarIconBitmap;
         private ImmutableList<CalendarItem> originalCalendarItems = ImmutableList<CalendarItem>.Empty;
         private readonly Paint calendarIconPaint = new Paint(PaintFlags.AntiAlias);
+        private readonly double minimumTextContrast = 1.6;
```
heytherewill: Why are you mixing the readonly members with the mutable ones 🙈

daividssilverio: This whole class and its parts need a nice clean up. Specially the drawing stuff, they could have a separate delegate to help clean up some of the code.

PR #6327, opened 2019-10-24
Remove unused resources

heytherewill:
## What's this?
This removes a bunch of resources that were not being used

## Why do we want this?
To reduce the burden of translating our apps!

## How is it done?
🔥 

## Review hints
Go over the removed strings and ask yourself: Are these supposed to be relevant somewhere?
For example, is the `DeleteNTimeEntries` string really never meant to be used? That's odd. ping @maritoggl 

## :squid: Permissions
Only myself.


maritoggl:
I can see `DeleteNTimeEntries` being used when deleting grouped time entries. In case there's a different string that's used for that purpose, we should be good.

PR #6335, opened 2019-10-25
Run full sync in background

simonrozsival:
## What's this?
This is a result of internal discussion. Since we resolved the "mysterious" failing bg sync (#6318), we can revert #6047 and enjoy full sync again.

## Why do we want this?
Some users report that their TEs are not being pushed to the server. This is caused by the fact that they have "many" unsynced TEs (like 20) and they close the app before it manages to sync them. With full background syncing, these TEs will be eventually pushed to the server and users will be happy.

### Side effects
If background syncing is killed during creating a new entity before we receive and process the response, we will create a duplicate of this entity eventually (the API endpoint isn't idempotent). This can happen in theory and we can't rule out that it will actually happen to some of our users. During the last 6 months, when background syncing was enabled in iOS, it preformed full sync the whole time, and we didn't have to deal with (m)any problem reports from our users.

We should definitely keep an eye on it and if it proves to cause more harm than we expected, we should revisit this again.

## :squid: Permissions
Let me do it once this is reviewed sufficiently.


daividssilverio:
But, isn't "light sync" better to have instead of "full sync" in background anyway? Won't it sync all time entries, adding placeholders where needed?

"Light sync" is faster, has less chances to fail, might provide a better user experience overall.

heytherewill:
I'm with Daivid here. I'd rather resolve the biggest issue (TEs) in a fast and efficient way rather than trying to fetch everything and increasing chances of timing out before we fetch TEs

simonrozsival:
One of our problems that TEs are not pushed to the server because user does something in the app, then closes the app and the changes are not propagated to the server. "Full sync" would attempt to push these changes to the server, while "Light sync" does not push anything.

Another possibility would be to chain "light sync" and then immediately perform "full sync" and sync as much as possible in the remaining time the OS gives us to run in background.

daividssilverio:
> Another possibility would be to chain "light sync" and then immediately perform "full sync" and sync as much as possible in the remaining time the OS gives us to run in background.

That seems very reasonable.

simonrozsival:
I changed the behavior to pull just TEs first and then run the full sync. Should we move forward with this? what do you think?

heytherewill:
I think this is sensible enough, but I'm putting @amulware in the loop just in case

amulware:
I don't think this is the right way forward. I think changing full sync to pull TEs first would be the better solution (see my proposal on Notion).

However, what bothers me more is that the main concern raised above is not addressed: What about taking too much time to do too much syncing in the background? Do we have any data on this? Can we be sure we're not just going to increase occurrences of duplicated time entries? Or maybe we have so little time it would never get to push data anyways?

simonrozsival:
> Can we be sure we're not just going to increase occurrences of duplicated time entries? 
I'd like to stress that it's similar to how syncing in foreground works. If we start syncing and then the user puts the app into background, it will suspend and kill the app after some time. I don't think we have significantly more time in foreground than in background 🤔

> Or maybe we have so little time it would never get to push data anyways?
We don't have "performance" metrics for background syncing. Maybe we should measure how long it takes. We know that syncing on iOS finishes in 94.4% of background syncs in 2.5 which performs only "sync light". (194.6k/206.1k). For full sync, which we used in iOS 2.2, the ran-to-finish rate was 79.5%. We don't have any usable analytics data for 2.3 and 2.4.

All the versions, which used background syncing until 2.5, used full sync, and I don't remember people complaining about duplicates.

The chaining of sync light & full sync is a "hack" which emulates the changes for full syncing you describe in the Notion document, but with very little coding. I'm not sure hwo

PR #6342, opened 2019-10-25
Fix enabled calendars not being saved

arturs105:
## What's this?
This fixes a bug, where enabled calendars would not save.

### Relationships
Doesn't have an issue that i know of.

## How is it done?
By making sure that `UserPreferences.SetEnabledCalendars` is not ever called with the default (initial) values and that the calendars are saved after selecting or unselecting a calendar.

## :squid: Permissions
🦑


daividssilverio:
This breaks android 😢 
Can't we simply add a "done" button like android does? That would solve the problem and keep things uniform.

Screen Shot 2019-10-25 at 1 30 51 PM

arturs105:
I'm not sure about that done button - to me it seems like the "Done" button closes the whole settings popup, not just the calendar settings, though i didn't find anything on that in the apple design docs.

I'll see if i can make this work for both OS'es.

daividssilverio:
The latest commit works almost alright on droid, except by the initial flow when permissions were not yet given/granted.

The switch displays the wrong state because an observable isn't awaited and another is modified with `.DistinctUntilChanged()`.

It's fixable on droid with: 
```cs
public CalendarSettingsViewModel(...)
        {
...
>>> CalendarListVisible = calendarListVisibleSubject.AsObservable(); // .DistinctUntilChanged(); removed
        }
```

and

```cs
private async Task togglCalendarIntegration()
        {
            var calendarListVisible = calendarListVisibleSubject.Value;
            if (calendarListVisible) ...
            else
            {
                await requestCalendarPermissionsIfNeeded();
>>>                var authorized = await permissionsChecker.CalendarPermissionGranted; // await added
                if (authorized) ...
            }
            calendarListVisibleSubject.OnNext(calendarListVisible);
        }
```

and it doesn't seem to affect iOS. So win-win. I've created an issue to handle further droid changes #6361

PR #6225, opened 2019-10-18
project/widgets

juanlaube:
## What's this?
The PR to merge `project/widgets` into `develop`

### Relationships

Closes the widgets project

## Why do we want this?
🎛 Because widgets!

## :squid: Permissions
No.

PR #6343, opened 2019-10-25
Address comments in the widgets

juanlaube:
## What's this?
Addresses some comments in the Widgets PR (#6225)

## Why do we want this?
To make William happy 🙂 

## Review hints
Most of the request were trivial, but please check the last two commits, those are fixes for the dependencies and the mess in `Toggl.sln`

## :squid: Permissions
Only @heytherewill


Toggl.iOS.TimerWidgetExtension/NetworkingHandler.cs
```diff
@@ -27,58 +26,53 @@ public NetworkingHandler(ITogglApi togglApi)
...
+
+                SharedStorage.Instance.SetNeedsSync(true);
+                SharedStorage.Instance.AddWidgetTrackingEvent(WidgetTrackingEvent.StartTimer());
+
+                return createdTimeEntry;
+            }
+            catch (Exception e)
+            {
+                SharedStorage.Instance.AddWidgetTrackingEvent(WidgetTrackingEvent.Error(e.Message));
+                throw e;
```
heytherewill: Don't `throw e`, as that will make you lose the stack. Simply `throw;` to re-throw

Toggl.iOS.TimerWidgetExtension/NetworkingHandler.cs
```diff
@@ -27,58 +26,53 @@ public NetworkingHandler(ITogglApi togglApi)
...
+                    .Select(stopTimeEntry)
+                    .FirstAsync();
+
+                SharedStorage.Instance.SetNeedsSync(true);
+                SharedStorage.Instance.AddWidgetTrackingEvent(WidgetTrackingEvent.StopTimer());
+            }
+            catch (Exception e)
+            {
+                SharedStorage.Instance.AddWidgetTrackingEvent(WidgetTrackingEvent.Error(e.Message));
+                throw e;
```
heytherewill: Same thing about the re-throw

PR #6344, opened 2019-10-25
Use regular services for pre-oreo devices

juanlaube:
## What's this?
Uses a regular service instead of a foreground service for pre-oreo devices.

## Why do we want this?
Because the app crashes otherwise.

## How is it done?
Using our `OreoApis` helper.

## :squid: Permissions
🦑


Toggl.Droid/Helper/AppWidgetProviderUtils.cs
```diff
@@ -27,7 +27,10 @@ public static class AppWidgetProviderUtils
         {
             var intent = new Intent(context, typeof(TService));
             intent.SetAction(action);
-            return PendingIntent.GetForegroundService(context, 0, intent, 0);
+
+            return OreoApis.AreAvailable
```
heytherewill: Since the check is being done at two places, please consider what we are doing for colors and adding a `SafeGetService` extension to make things a little more ©️🍝 proof

PR #6401, opened 2019-10-30
Update Xamarin.Android csproj to improve build times

jonathanpeppers:
## What's this?

I heard from @JonDouglas that the developer loop on this app was slow. Like 1-2 minutes or something...

I noticed a few settings that will drastically improve your Debug build times:

* The linker should be set to `None` for Debug builds.
* Only use ProGuard in Release builds.
* I moved the settings for Fast Deployment around, so things were explicit for Debug vs Release.

For details, see: https://devblogs.microsoft.com/xamarin/optimize-xamarin-android-builds/

However, I did find some bugs in Xamarin.Android's build system, when profiling builds of this app.

### Missing AndroidResource files

This app lists two files that do not exist:

    <AndroidResource Include="Resources\layout\SelectDateFormatFragment.axml" />
    <AndroidResource Include="Resources\layout\SelectDateFormatFragmentCell.axml" />

Due to the way MSBuild works, this causes:

    Building target "_CompileResources" completely.
    Input file "Resources\layout\SelectDateFormatFragment.xml" does not exist.
    Building target "_UpdateAndroidResgen" completely.
    Input file "Resources\layout\SelectDateFormatFragment.xml" does not exist.

Basically expensive `aapt2` calls happen on every build... We can just
remove these lines for now, until we get a Xamarin.Android update.

Fixed in: https://github.com/xamarin/xamarin-android/pull/3837

### Issue with GPS and aapt2

I noticed usage of Google Play Services and aapt2 causes a few MSBuild
targets to always run.

I don't have a workaround for this, but we have fixes in:

https://github.com/xamarin/GooglePlayServicesComponents/pull/271
https://github.com/xamarin/xamarin-android/pull/3856

### Relationships

https://github.com/xamarin/xamarin-android/pull/3837
https://github.com/xamarin/GooglePlayServicesComponents/pull/271
https://github.com/xamarin/xamarin-android/pull/3856

### DIFF
--- a/Toggl.Droid/Toggl.Droid.csproj
+++ b/Toggl.Droid/Toggl.Droid.csproj
@@ -15,11 +15,7 @@
     <AndroidManifest>Properties\AndroidManifest.xml</AndroidManifest>
     <MonoAndroidResourcePrefix>Resources</MonoAndroidResourcePrefix>
     <MonoAndroidAssetsPrefix>Assets</MonoAndroidAssetsPrefix>
-    <AndroidLinkMode>Full</AndroidLinkMode>
-    <AndroidLinkSkip>Toggl.Giskard;Toggl.Storage.Realm</AndroidLinkSkip>
-    <EnableProguard>true</EnableProguard>
     <JavaMaximumHeapSize>2G</JavaMaximumHeapSize>
-    <AndroidUseSharedRuntime>false</AndroidUseSharedRuntime>
     <LangVersion>7.1</LangVersion>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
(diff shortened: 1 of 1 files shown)


Toggl.Droid/Toggl.Droid.csproj
```diff
@@ -33,10 +29,16 @@
     <AndroidSupportedAbis>arm64-v8a;armeabi-v7a;x86;x86_64</AndroidSupportedAbis>
     <AndroidHttpClientHandlerType>Xamarin.Android.Net.AndroidClientHandler</AndroidHttpClientHandlerType>
     <AndroidUseSharedRuntime>true</AndroidUseSharedRuntime>
+    <EmbedAssembliesIntoApk>false</EmbedAssembliesIntoApk>
+    <AndroidLinkMode>None</AndroidLinkMode>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
     <AndroidHttpClientHandlerType>Xamarin.Android.Net.AndroidClientHandler</AndroidHttpClientHandlerType>
+    <AndroidUseSharedRuntime>false</AndroidUseSharedRuntime>
```
heytherewill: This change ( and all other included in `Release.AnyCPU`) should also be included in `Release.PlayStore`

Toggl.Droid/Toggl.Droid.csproj
```diff
@@ -33,10 +29,16 @@
...
     <EmbedAssembliesIntoApk>true</EmbedAssembliesIntoApk>
+    <EnableProguard>true</EnableProguard>
+    <AndroidLinkMode>Full</AndroidLinkMode>
```
heytherewill: I'm not entirely sure about this one. The Proguard is probably fine, since we don't add that much code that could be problematic to the Proguard anymore, but turning the linker off sounds like inviting hotfixes into releases

I guess we could try and revert it if we get problems in the next release? 
Pinging @amulware for ➿

Toggl.Droid/Toggl.Droid.csproj
jonathanpeppers: This PR is a suggestion, feel free to take what you need/don't need. Using Progard & Linker in debug mode is not ideal, we have added [build warnings](https://github.com/xamarin/xamarin-android/pull/3630) for these in the 16.4 preview. 

Removing the missing files also will help. 😄

Toggl.Droid/Toggl.Droid.csproj
jonathanpeppers: I would not recommend using the linker in Debug builds -- as the cost to your development loop is not worth it. Even if there was a linker bug in the Release build it should go through some kind of QA testing, as there are many other things that could go wrong besides just the linker.

heytherewill:
I've incorporated the suggestions from this PR into [this commit](https://github.com/toggl/mobileapp/commit/7ba58913ecca4733a6e9325b9d7ee9fd4861161a), so we can safely close this one.

Thanks again for helping out, I hope Xamarin and Toggl can collaborate like this more times in the future :)

PR #6403, opened 2019-10-30
Fix UI thread bug in the reports view model

### LINKED ISSUE #6402: [Regression] Reports are not working anymore after changing some observables into tasks
#5693 has introduced some issues with reports.

The reports screen does not show the pie chart, the projects and the summaries correctly.

juanlaube:
## What's this?
Fixes a bug in the reports view model when updating the UI from a background thread.

### Relationships

Closes #6402 

## Why do we want this?
We don't like bugs.

## How is it done?
`.AsDriver` all the things!

## :squid: Permissions
🦑


Toggl.Core.UI/ViewModels/Reports/ReportsViewModel.cs
```diff
@@ -175,9 +176,17 @@ public IObservable<bool> TotalTimeIsZeroObservable
                 .Select(prefs => prefs.DurationFormat)
                 .AsDriver(schedulerProvider);
 
-            SegmentsObservable = segmentsSubject.CombineLatest(DurationFormatObservable, applyDurationFormat);
-            GroupedSegmentsObservable = SegmentsObservable.CombineLatest(DurationFormatObservable, groupSegments);
-            ShowEmptyStateObservable = SegmentsObservable.CombineLatest(IsLoadingObservable, shouldShowEmptyState);
+            SegmentsObservable = segmentsSubject
+                .CombineLatest(DurationFormatObservable, applyDurationFormat)
```
heytherewill: Make this a variable so you don't chain Drivers:

```csharp
var segmentsObservable = segmentsSubject.CombineLatest(DurationFormatObservable, applyDurationFormat);

SegmentsObservable  = segmentsObservable.AsDriver(schedulerProvider);

GroupedSegmentsObservable = segmentsObservable
    .CombineLatest(DurationFormatObservable, groupSegments)
    .AsDriver(schedulerProvider);

ShowEmptyStateObservable = segmentsObservable
    .CombineLatest(IsLoadingObservable, shouldShowEmptyState)
    .AsDriver(schedulerProvider);
```

PR #6405, opened 2019-10-30
Track password complexity

### LINKED ISSUE #6152: Track password complexity
To allow us to make good decisions on security VS convenience, we would like to track the password complexity of our users.

Every time a user signs up with a new account, I'd like us to track the following event and properties:

`SignupPasswordComplexity`
- `Length` - total length of the password
- `LowerCaseCount` - count of lower case Latin letters (characters between and including `a` and `z`)
- `UpperCaseCount` - count of upper case Latin letters (characters between and including `A` and `Z`)
- `DigitCount` - count of numerical digits (characters between and including `0` and `9`)
- `OtherASCIICount` - count of all printable ASCII characters that are not letters or digits (characters between ` `(space) and `~` but not already included above)
- `NonASCIICount` - any non(-printable) ASCII/ characters (any character not already counted above)

Feel free to consult an ASCII table to see what order I'm referring to. Note that C# allows smaller than/larger than comparisons between `Char`s, which is probably your best bet for implementing this.

heytherewill:
## What's this?
This tracks the complexity of the passwords used to sign up

### Relationships
Closes #6152

## Why do we want this?
D A T A

## :squid: Permissions
Anyone, anytime


Toggl.Core/Analytics/SignupPasswordComplexityEvent.cs
```diff
@@ -0,0 +1,68 @@
...
+                {
+                    digitCount++;
+                    continue;
+                }
+
+                if (character >= ' ' && character <= '~')
+                {
+                    otherAsciiCount++;
+                    continue;
+                }
```
pdrj: Rather than having all these `continue`s, how about using `else if` and `else` instead? I think, since we want the conditions to be exclusive, that would be more readable.

Toggl.Core.Tests/AnalyticsService/SignupPasswordComplexityEvent.cs
```diff
@@ -0,0 +1,118 @@
...
+
+            public static IEnumerable<object[]> PasswordsAndCounts
+                => new[]
+                {
+                    new object[] { "123456", 6, 0, 0, 0, 0 },
+                    new object[] { "🔥🔥🔥asd123", 3, 6, 3, 0, 0 },
+                    new object[] { "$J?=f6&([)+7LQAx", 2, 0, 2, 4, 8 },
+                    new object[] { "JQR2$PK?=PGzqGyAMxRLB=QC_LZNYLVgZ6_8Ej", 3, 0, 6, 23, 6 },
+                    new object[] { "_DK*xF_ZjnZXLAqQPUrkLVn4AJZJM5tG^YxvuE", 2, 0, 11, 21, 4 },
+                    new object[] { "&=-DwdzDAh848v&uKpm&XM%MXA!ZF-5fmT9jP5", 6, 0, 11, 13, 8 },
```
pdrj: For completeness, can we add a test that uses a more exotic alphabet? For instance Yiddish or Arabic.

Toggl.Core/Analytics/SignupPasswordComplexityEvent.cs
```diff
@@ -0,0 +1,68 @@
...
+
+                nonAsciiCount++;
+            }
+        }
+
+        public string EventName { get; } = eventName;
+
+        public Dictionary<string, string> ToDictionary()
+            => new Dictionary<string, string>
```
pdrj: I know it's not introduced in this PR, but why don't we allow `int` values in events? That should make manipulating the data easier, if it's just numbers. Right now if we wanted to let's say count the number of events with password length `> 4`, it'll probably be the case that every password with length `>= 10` wouldn't be counted, if the filtering just uses lexicographical order on strings.

Toggl.Core.Tests/AnalyticsService/SignupPasswordComplexityEvent.cs
```diff
@@ -0,0 +1,118 @@
...
+
+                tryingToCreateAnEventFromAnInvalidPassword.Should().Throw<ArgumentException>();
+            }
+
+        }
+
+        public sealed class TheToDictionaryMethod
+        {
+            [Property]
+            public void AlwaysHasHasSixParameters(string passwordString)
```
pdrj: Typo: `HasHas`.

Toggl.Core/Analytics/SignupPasswordComplexityEvent.cs
heytherewill: Good point. I don't know how AppCenter deals with it, though. I could simply pad the strings with zero. ping @amulware

PR #6409, opened 2019-10-30
Track when the user has no workspaces

### LINKED ISSUE #6287: Track how often users don't have ANY workspace after sync
We track not having any default workspace, but we don't track when users have no workspaces AT ALL. Please add an event:

`NoWorkspace`

that gets sent whenever we detect the user not having any workspace during sync.

## Suggestion

Implement this in https://github.com/toggl/mobileapp/blob/3faa63ac478fcc66898183ed0e36e10a353104ab/Toggl.Core/Sync/States/Pull/DetectNotHavingAccessToAnyWorkspaceState.cs

Similar to

https://github.com/toggl/mobileapp/blob/3faa63ac478fcc66898183ed0e36e10a353104ab/Toggl.Core/Sync/States/Pull/DetectUserHavingNoDefaultWorkspaceSetState.cs#L28-L40

heytherewill:
## What's this?
This adds an event to track how often users are left with no workspaces after sync

### Relationships
Closes #6287 

## Why do we want this?
D A T A 

## :squid: Permissions
Anyone, anytime


Toggl.Core/Sync/States/Pull/DetectNotHavingAccessToAnyWorkspaceState.cs
```diff
@@ -11,16 +12,29 @@ internal sealed class DetectNotHavingAccessToAnyWorkspaceState : ISyncState<IFet
...
 
         public IObservable<ITransition> Start(IFetchObservables fetch)
             => dataSource.Workspaces.GetAll()
-                .Select(workspaces => workspaces.Any()
+                .Select(ws => ws.Any())
+                .Do(trackNoWorkspacesNeeded)
+                .Select(userHasAccessToAnyWorkspace => userHasAccessToAnyWorkspace
                     ? Done.Transition(fetch)
                     : throw new NoWorkspaceException());
+        private void trackNoWorkspacesNeeded(bool userHasAccessToAnyWorkspace)
```
sofokli1: shouldn't there be an empty line here?

heytherewill: Fixed

PR #6410, opened 2019-10-30
Fix calendar day view offset handling

### LINKED ISSUE #6387: Closing the contextual menu while at bottom of the calendar view leaves an empty space
image

How to replicate:
- Select any calendar item
- Scroll all the way to the bottom of the calendar page
- Close the calendar contextual menu
- It should be enough to replicate the screenshot above

daividssilverio:
## What's this?
This fixes #6387

### Relationships
Closes #6387

## Why do we want this?
So that calendar page doesn't look buggy when the contextual menu is closed.

## :squid: Permissions
🦑 📆 pals


Toggl.Droid/Views/Calendar/CalendarDayView.Background.cs
```diff
@@ -4,6 +4,7 @@
 using Toggl.Shared;
 using System.Globalization;
 using System.Linq;
+using Android.Util;
```
pdrj: 🔥 ?

Toggl.Droid/Views/Calendar/CalendarDayView.Background.cs
```diff
@@ -47,6 +48,9 @@ public partial class CalendarDayView
 
         partial void processBackgroundOnLayout(bool changed, int left, int top, int right, int bottom)
         {
+            if (!changed) 
+                return;
+
```
pdrj: Is this related to the issue, or just an optimization? The fix seems to work without it 😃.

Toggl.Droid/Views/Calendar/CalendarDayView.Background.cs
daividssilverio: 🔥

Toggl.Droid/Views/Calendar/CalendarDayView.Background.cs
daividssilverio: Yeah, it's just an optimization I forgot to add. Without that, the lines bellow are executed on every invalidation.

Issue #6368, opened 2019-10-29
Calendar: Fix total duration of TEs not working


sofokli1:
Fix the total duration of the day's TEs.

PR #6414, opened 2019-10-31
Add missing total duration indicator on calendar pages

daividssilverio:
## What's this?
This adds that little piece of information that was not being show in the calendar pages.
I'm talking about this:
Screen Shot 2019-10-30 at 8 54 08 PM

### Relationships
Droid bits of #6368 

## How is it done?
It simply observes the time entries for a specific day view model and specifically for "Today" it uses the `ObserveTimeTrackedToday` instead of the `CalendarItems` observable list.

### Why not another way?
Please, feel free to suggest something different. It would be nice to use an `Rx` operator that would select the right stream for us instead of this:
```cs
            if (timeService.CurrentDateTime.LocalDateTime.Date == Date)
            {
                timeTrackedOnDaySubject = timeTrackedOnDay
                    .CombineLatest(timeTrackedToday.ReemitWhen(timeService.MidnightObservable.SelectUnit()), 
                        selectTrackedTimeSource);
            }
            else
            {
                timeTrackedOnDaySubject = timeTrackedOnDay;
            }
```

## :squid: Permissions
🦑 📆 pals, after I add some tests for added observables.


Toggl.Core.UI/ViewModels/Calendar/CalendarDayViewModel.cs
```diff
@@ -108,8 +113,43 @@ public sealed class CalendarDayViewModel : ViewModel
...
+                .Select(items => items
+                        .Where(item => item.Source == CalendarItemSource.TimeEntry)
+                        .Sum(item => item.Duration(timeService.CurrentDateTime)))
+                .StartWith(TimeSpan.Zero);
+
+            if (timeService.CurrentDateTime.LocalDateTime.Date == Date)
+            {
+                timeTrackedOnDaySubject = timeTrackedOnDay
+                    .CombineLatest(timeTrackedToday.ReemitWhen(timeService.MidnightObservable.SelectUnit()), 
+                        selectTrackedTimeSource);
```
arturs105: Is this necessary, can't we just do `timeTrackedOnDaySubject  = timeTrackedToday `, since `selectTrackedTimeSource ` is gonna return it anyways.

daividssilverio: I added that because the user might be using the app around the clock. So If the user is around midnight on day X with a time entry running, when midnight arrives, the now day X is not today anymore, and the X+1 day should use the `timeTrackedToday` instead.

Even though those kind of time entries don't appear in the calendar (entries across days).

daividssilverio:
Tests are under progress and can be done in another branch if this can help getting the iOS side done faster.

PR #6421, opened 2019-10-31
Track number of workspaces

### LINKED ISSUE #4344: Track no-default-workspace case
I'd like to track how often it occurs that users have workspaces, but do not have a default workspace set.

Please check that this event won't accidentally be triggered in the no-workspace case (which should set the default, but I'm not sure if it does or relies on the no-default-handling to do so).

Also please include a property that tells us how many workspaces the user has. If it is 1, we know that the default is assigned automatically, otherwise, they see a dialogue to choose it.

heytherewill:
## What's this?
This tracks the amount of workspaces the user has when it reaches a no default workspace situation

### Relationships
Closes #4344

## Why do we want this?
D A T A

## Tests
[Syncing tests](https://app.bitrise.io/build/bafb1c1ee61f321a#?tab=log)

## :squid: Permissions
Anyone, anytime *but only if sync tests are passing*


sofokli1:
@heytherewill not sure if ti my issue but the test link seems broken

heytherewill:
@sofokli1 Bitrise is painfully slow today, but the link opens and the tests are passing :)

PR #6423, opened 2019-10-31
Fix tab bar items on iOS 13.0 onwards

### LINKED ISSUE #6046: Bottom navigation icons not properly aligned
## 🐛 Bug Report
Icons in the bottom navigation bar are not properly aligned as seen on the screenshot attached. 
IMG_3679

## Context
Build/Branch: Daneel 2.4
Phone OS: iOS 13.1.2
Phone model: iPhone 7

## Expected behavior
Icons are vertically centered.

dshr:
## What's this?
A small fix to tab bar item rendering

### Relationships
Closes #6046

## Why do we want this?
Pixel perfect designs

## How is it done?
By applying the tab bar item shimming only to devices running iOS 12 or earlier.

### Why not another way?
Seems like the simplest way?

## :squid: Permissions
All yours!


Toggl.iOS/ViewControllers/MainTabBarController.cs
```diff
@@ -96,8 +96,9 @@ private void recalculateTabBarInsets()
             ViewControllers.ToList()
                            .ForEach(vc =>
             {
-                if (TraitCollection.HorizontalSizeClass == UIUserInterfaceSizeClass.Compact)
+                if (TraitCollection.HorizontalSizeClass == UIUserInterfaceSizeClass.Compact && !UIDevice.CurrentDevice.CheckSystemVersion(13, 0))
```
sofokli1: This will be an issue in iOS 14. Isn't there a ways to say iOS 13 and lower?

arturs105: Why would this be a problem on iOS 14? It does one thing on everything that is less than 13, and another for 13 and higher, doesn't it?

sofokli1: `UIDevice.CurrentDevice.CheckSystemVersion` sounds more like it is checking for a specific version and not a range like` >= the given value`.

PR #6430, opened 2019-11-01
Fix calendar week stripe over scroll on beginning of week changes

### LINKED ISSUE #6386: It's possible to try to scroll to the future in the week stripe view in the calendar
How to replicate?
- Set the current day of the week to the next week day compared to the current day of the week, e.g. If today is a Tuesday, set the beginning of the week to Wednesday in the app settings.
- Try to scroll the page stripe to the next week.

daividssilverio:
### Relationships
This fixes #6386

## How is it done?
The pager `GetItemPosition` is used in the process of telling if a page is part of the adapter/pager.
Making it return `PositionNone` when for some reason the pager contains a page it shouldn't, triggers the page removal. That way we don't get the weird behaviour described on #6386. 

And yes, I couldn't tell why it was happening. The adapter looks fine to me.

## :squid: Permissions
🦑 📆 pals


daividssilverio:
Thanks for trying to find a better solution @semanticer . I couldn't find a better way.

PR #6431, opened 2019-11-01
🐛 Fix reports and crash on Android 2.6 🍧

kornelijepetak:
## What's this?
Incorporates #6403 and #6425 in order to fix the failing release.
These are
* Reports failing to show because of threading issues (no drivers)
* Off-by-one error in reports bar chart

### Relationships
Refs #6403
Refs #6425

## Why do we want this?
So that 📊 look nice and that there are no crashes.

## How is it done?
By 🍒-picking the two bug fixes from `develop`.

### Why not another way?
Because this is the sanest way.

## :squid: Permissions
Me, asap.


Toggl.Core.Tests/UI/ViewModels/ReportsViewModelTests.cs
```diff
@@ -555,7 +555,7 @@ public async Task SetsOtherProjectWithOneSegmentToThatSegmentButWithOnePercentIf
                     .ForEach(percentage => percentage.Should().BeGreaterOrEqualTo(5));
             }
         }
-
```
dshr: Can we remove the strange whitespace artifacts?

PR #6436, opened 2019-11-01
Fix sync failure caused by aggregate exception

### LINKED ISSUE #6434: Client end server error exceptions are wrapped in AggregateException
There's a weird behavior in the app, where sync fails and data is not pushed to the server, when some entity fails to sync and the server returns a 4xx or 5xx error, showing "Sync Failed" and terminating sync.

I'm not sure if it affects only the current 2.6 releases, which are being tested, or if it is in 2.5. If it's already shipped to the users, it's a major problem, which we must fix ASAP. If it's just in 2.6, we must fix it before we ship this to the users.

How to reproduce:
- start a time entry (running)
- change the duration to 999:00 hours
- let it tick for a while, then stop it and wait for the app to sync
- on iOS, you'll see the gray bar saying "Sync Failed", on Android you'll see a toast msg saying "Sync failed".

What should happen:
- syncing succeeds
- the TE is marked with the red `(!)` icon and it shows an error message in the edit view when you tap on it

I think there is a problem with rethrowing somewhere in the networking package, but I haven't looked into it too much just yet.

simonrozsival:
## What's this?
The syncing code catches exceptions thrown by the `BaseApi` class for requests, which result in a client or server error. Syncing has different behaviors for different responses and there is also a default error for unexpected/unknown errors, which aren't API exceptions. `System.AggregateException` is an "unexpected" error. When it is thrown, syncing is terminated. That's a problem.

### Relationships
Closes #6434, introduced in #5693 (most likely)

## How is it done?

`task.ToObservable()` apparently translates any exceptions thrown in the task into an `AggregateException`. ~Therefore, we must catch the aggregate exception for every `ToObservable` and re-throw the internal exception so our code can handle it correctly. There are many places, where we need to fix this - basically for every API call which needs to be changed into an observable, so I created an extension method `AggregateExceptionUnwrappingToObservable`. It's quite long and I'm not 100% happy with it, but it sounds quite scary, which is a good thing in this case IMO :D suggestions for a better name are welcome.~

_Edit:_ After a call with @heytherewill, I changed the strategy and simplified this PR. It now doesn't touch interactors etc. and it is used only where we need the type-matching of the exception in the sync state.

_Edit 2:_ I changed the code once more and I hopefully made it more explicit this time. I removed the extension method, because it's not gonna be reused in the codebase anyway, and I updated the unit tests to include aggregate exceptions. I also added a comment explaining why `Single` is used intentionally and that I am aware that it can throw.

### Side effects
The user has to know that he shouldn't use the default `ToObservable`. I'm not sure what to do about it.

## Review hints
Run the app in a simulator/emulator or on a device.

Test this:
- Create a time entry
- In the creation view, set the duration to 999 hours and let it tick for a few seconds
- Confirm the new TE
- Let the app sync, it should show "Sync succeeded"
- When you tap on the running TE, it should show an "Oops" error

Then test this:
- Create a new TE
- Confirm it
- Tap on it and set the duration to 999 hours (don't stop it though)
- Let it tick for a few seconds, then confirm the edit
- Let the app sync, it should show "Sync succeeded"
- When you tap on the running TE, it should show an "Oops" error

## :squid: Permissions
I'll do it. We'll then have to figure out how to cherry-pick the squash commit into iOS and Android 2.6 release branches.


Toggl.Core/Extensions/ExceptionExtensions.cs
```diff
@@ -7,5 +8,10 @@ public static class ExceptionExtensions
     {
         public static bool IsAnonymized(this Exception exception)
             => Attribute.IsDefined(exception.GetType(), typeof(IsAnonymizedAttribute));
+
+        public static Exception UnwrapSingle(this Exception exception)
+            => exception is AggregateException aggregate
+                ? aggregate.Flatten().InnerExceptions.Single()
```
heytherewill: Can you ensure that InnerExceptions *always* has only one member and that said member will never be another AggregateException after Flatten is called?

simonrozsival: `Flatten` should return a list of exceptions which are aggregate exceptions - it does it recursively. It might contain multiple exceptions and it will throw in that case. That's why I used `Single`. If used resposibly, it's not that dangerous :)

PR #6446, opened 2019-11-04
Use new bundle version format for the widget extension

juanlaube:
## What's this?
Replaces the commit count with the new bundle version format for the widget extension.

### Relationships
I just found this problem because I got an email from Apple.

## Why do we want this?
To upload and release new versions.

## :squid: Permissions
🦑


heytherewill:
Why is this also including other changes? Are those cherry picked? Will this branch get squashed or merged?

PR #6477, opened 2019-11-06
💄 Fix current time supplementary view stilling touches from underling cells

sofokli1:
## What's this?
Fix current time supplementary view stilling touches from underling cells

### Relationships
none

## Why do we want this?
we never want to touch the current time supplementary view 

## :squid: Permissions
:squid:


Toggl.iOS/Cells/Calendar/CurrentTimeSupplementaryView.cs
```diff
@@ -16,6 +16,7 @@ static CurrentTimeSupplementaryView()
 
         protected CurrentTimeSupplementaryView(IntPtr handle) : base(handle)
         {
+            UserInteractionEnabled = false;
```
heytherewill: Please don't run any initialization in `IntPtr` constructors

PR #6479, opened 2019-11-06
Bumped up targetSdkVersion to 29

### LINKED ISSUE #6478: Bump up the targetSdkVersion from 28 to 29
## 🚧 Refactor Proposal
We should always target the current sdk version. 

## Technical aspects
I think we just need to change the number in manifest `android:targetSdkVersion="28"` 
I will submit PR in a sec and we should test if anything breaks.

semanticer:
## What's this?
Bumping up targetSdkVersion from 28 (Android 9) to 29 (Android 10)

### Relationships
Closes #6478

## Why do we want this?
We would like to take advantage of those 29 APIs at some point in the future. Also, we want google to approve our future app updates. 

## How is it done?
Bumped up targetSdkVersion number in the manifest (I hope this is the correct way in xamarin??) 

### Why not another way?
Tell me if there is another way. This is typically changed in Gradle in normal Android apps.

### Side effects
Non that I know of but we should test this carefully and look at the guidelines from google mentioned in review hints.

## Review hints
https://developer.android.com/about/versions/10/migration
https://developer.android.com/about/versions/10/behavior-changes-10

## :squid: Permissions
Me or @heytherewill


semanticer:
@amulware @daividssilverio should we add this to the current release or wait for 2.8?

heytherewill:
2.8 for sure

heytherewill:
Do not merge this year; it's on hold until 2.7 is branched out

PR #6480, opened 2019-11-06
🌟 Implement a SearchEngine for quick in-memory searching of Time Entries

### LINKED ISSUE #5473: Improve start view autocomplete performance
## 🐛 Bug Report

When I type very fast in the start view, it takes a long time after I stop typing for suggestions to pop up.

## Context

Has been happening on iOS for months now.

## Expected behavior

Whenever I stop typing in the start view, it should not take very long to show suggestions. More importantly, my typing speed shouldn't affect it. Typing 'this is a test string' fast should have the same delay before suggestions show afterwards as typing 'this is a test strin', waiting for suggestions, then hitting 'g'.

## To reproduce
Type a recent description of the name of a longer project very fast. Stop typing, notice that it takes a long time.
(You need to have a good amount of data in your account - in my case it's ~30 projects with clients, and ~500 time entries, all with projects and descriptions)

## Suggestions
I highly suspect we are not cancelling the background searches triggered by each keystroke so they stack up and overload even modern multi-core phones.

I see several solutions:
1. A simple throttle (not a big fan, because we don't know how long search will take for different users)
2. Always cancel previous unfinished searches when a new one is started (can we do this easily?) - this will lead to the fastest result when the user stops typing, but it also means that while typing fast the screen stays empty
3. Don't start a search while a previous is running (but do check if a new search is needed when one finishes, so we catch the final search) - this will lead to slightly slower final results, but for fast typing there will be intermediate results, which might get the user what they are looking for earlier (the final result might be the same as the previous one anyways)

Additionally, but separately to this, we may want to do the search entirely in memory, instead of accessing realm every time. Assuming all data always fits into memory (and right now we read it anyways, on every search), searching in memory will be significantly faster. It might not eliminate the need for one of the above, but it might mean we get really really nice and responsive results, especially with approach 3.

We can make it even faster if we are smart and search only in the already filtered set when the user appends characters to the search string (which is the most common use case).

Let's be careful to make sure we address all different searches that can happen in this view (TE, project, tag), though perhaps project/tag search can be handled simpler, since the data searched is simpler, and usually considerably less.

kornelijepetak:
## What's this?
Implements caching and in-memory search of the time entries to speed up the typing in the Start TE view.

⚠ Please read the whole PR description before looking at the code.

### Relationships
Closes #5473

## Why do we want this?
For speed.

## How is it done?
By 
* fetching data in memory
* caching filtered time entries for various used query texts
* optimizing changes (appending and prepending)

I've used the `Task<IEnumerable<T>>` as a return type, but I'm thinking whether it maybe makes sense to use `IObservable<IEnumerable<T>>` instead. What do you think?

### Why not another way?
That's where you come in.

### Side effects
There's a thing. I have a couple thousand time entries on the phone. Now, it seems that there's still a small delay and it seems to be related to the refresh/redraw of the UI after the change.

However, it feels like the opening of the Start TE now happens with a small delay. I'm not sure if I'm maybe initializing the inital data on the UI thread?

## :squid: Permissions
Me, @heytherewill or @amulware


Toggl.Core/SearchBox/SearchEngine.cs
```diff
@@ -0,0 +1,77 @@
...
+                return (dataSource, Prepend);
+
+            // No optimization found, use all 
+            return (completeData, Other);
+        }
+
+        public async Task<IEnumerable<T>> Get(string query)
+        {
+            await initialDataAvailable.Task;
+            return await Task.Run(() => get(query).AsEnumerable());
```
amulware: There's a lot of accesses to the cache in get(), but nothing to ensure thread safety. What will happen if a user types faster than we can search, or other threading stuff happens?

Toggl.Core/SearchBox/SearchEngine.cs
```diff
@@ -0,0 +1,77 @@
...
+        }
+
+        private ImmutableList<T> get(string query)
+        {
+            var (dataSource, changeType) = determineBaseDataSource(query);
+
+            if (changeType == Append || changeType == Prepend)
+                cache[query] = dataSource;
```
amulware: Why do we store the base data source for the new query? That doesn't seem right.

Toggl.Core/SearchBox/SearchEngine.cs
```diff
@@ -0,0 +1,77 @@
...
+
+            return dataSource.Filter(query).Items;
```
amulware: We should be storing the filtered one here instead, right? Instead of forgetting it?

Toggl.Core/SearchBox/SearchEngineCacheItem.cs
```diff
@@ -0,0 +1,39 @@
...
+using System.Collections.Immutable;
+using System.Linq;
+using System.Reactive.Linq;
+using Toggl.Core.Extensions;
+
+namespace Toggl.Core.Search
+{
+    internal class SearchEngineCacheItem<T>
+    {
+        private readonly Func<string, T, bool> filter;
```
amulware: It seems very strange to me that we store this filter here, it's not really needed for anything, is it? We could just inject it in Filter()?

Toggl.Core/SearchBox/SearchEngine.cs
```diff
@@ -0,0 +1,77 @@
...
+            return (completeData, Other);
+        }
+
+        public async Task<IEnumerable<T>> Get(string query)
+        {
+            await initialDataAvailable.Task;
+            return await Task.Run(() => get(query).AsEnumerable());
+        }
+
+        private ImmutableList<T> get(string query)
```
amulware: Why is there no shortcut in here to return the cached list if it's a known query?

Toggl.Core/SearchBox/SearchEngine.cs
```diff
@@ -0,0 +1,77 @@
...
+                .OrderByDescending(cachedItem => cachedItem.Key.Length)
+                .Where(cachedItem => cachedItem.Key.Length < query.Length)
+                .FirstOrDefault(cachedItem => query.EndsWith(cachedItem.Key))
+                .Value;
+
+            if (dataSource != null)
+                return (dataSource, Prepend);
+
+            // No optimization found, use all 
+            return (completeData, Other);
```
amulware: We can remove this case and the `completeData` field if we add the `completeData` cache item to the dictionary. Maybe worth simplifying? Or is there a benefit in telling apart the 'other' case for something?

Toggl.Core/SearchBox/SearchEngine.cs
```diff
@@ -0,0 +1,77 @@
...
+                .Where(cachedItem => cachedItem.Key.Length < query.Length)
+                .FirstOrDefault(cachedItem => query.StartsWith(cachedItem.Key))
+                .Value;
+
+            if (dataSource != null)
+                return (dataSource, Append);
+
+            // Character preppended to a previously known query
+            dataSource = cache
+                .OrderByDescending(cachedItem => cachedItem.Key.Length)
```
amulware: This is the same enumeration as just above, could we maybe do them both in the same loop instead?

amulware: Also, since we're always enumerating the cache sorted by size, should we maybe just keep it ordered and use a different collection for that?

Toggl.Core/SearchBox/SearchEngine.cs
kornelijepetak: Thread safety concerns are a valid point. I will take it into account.

Toggl.Core/SearchBox/SearchEngine.cs
kornelijepetak: Yes, in addition to your next comment, this is a bug. I've been reworking this method a lot and it ended up wrong. Fixing it.

Toggl.Core/SearchBox/SearchEngine.cs
kornelijepetak: See previous comment.

Toggl.Core/SearchBox/SearchEngineCacheItem.cs
kornelijepetak: Yes, that could work. Less memory.

Toggl.Core/SearchBox/SearchEngine.cs
kornelijepetak: How is there not?
The first thing that `determineBaseDataSource()` does is check for existance of the cached entry.

Toggl.Core/SearchBox/SearchEngine.cs
kornelijepetak: `completeData` should be returned for all queries that do not exist in the dictionary (and for whose there's no append/prepend combination in the dictionary as well). The number of those is infinite. I don't really understand how you mean to add such case to a dictionary. What kind of key would that be?

Toggl.Core/SearchBox/SearchEngine.cs
kornelijepetak: I changed this to a single loop.

However, using any other collection seems like an overkill (if there is such a collection at all). AFAIK, there's no collection that can handle key-value pairs and be sorted by something else than the same key. Because that's what we would need. Keys need to be queries, but the ordering is done on the query lengths. 
The only 2 ways I see it working is `Dictionary<int, Dictionary<string, DataSource>>` (which then has a problem of unwrapping for direct access) or a combination of two different collection (a dictionary to enable O(n) access, and a `SortedList` to work with fast enumeration). Both seem like even a more complex idea.

Especially since this part of the code is not even performance critical. The dictionary search (even with two loops as before) lasts less than a few `ms`. I don't think it's worthwile introducing more complex structures, let alone writing our own.

Toggl.Core/SearchBox/SearchEngine.cs
amulware: But there is no early return in get() it will still do a Filter() call, which is the most expensive thing in the whole algorithm.

Toggl.Core/SearchBox/SearchEngineCacheItem.cs
amulware: Also semantically better I think.

Toggl.Core/SearchBox/SearchEngine.cs
amulware: There is SortedSet<T>, but I don't know right now how you'd best implement the lookup for the exact match (binary + linear search I guess, not great..), though you could use both a sorted set and a dictionary. But I agree that it might not be worth the complication (though I generally don't like OrderBy calls when we're only retrieving a single item in the end).

Toggl.Core/SearchBox/SearchEngine.cs
amulware: If you add it to the dictionary for they "" key, then any other query will be an extension of it in the Append/Prepend case.

kornelijepetak: Oh, yes, you are right. 
But it does introduce a layer of questioning when reading the code, because these are implicit, but I've tried to signify that in the exception message. Not sure if it's any clearer though.

Toggl.Core/SearchBox/SearchEngine.cs
kornelijepetak: For now I won't introduce any more complexity.
Also, the performance of this should always be fine, since we should probably limit the number of cached items. Otherwise, it could lead to some memory problems. Not that fast, but still.
How do you suggest we approach the limit?

Toggl.Core/SearchBox/SearchEngine.cs
```diff
@@ -0,0 +1,80 @@
...
+                    .Where(cachedItem => cachedItem.Key.Length < query.Length);
+
+                foreach (var cachedItem in sortedCacheItems)
+                {
+                    if (query.StartsWith(cachedItem.Key))
+                        return (cachedItem.Value, Append);
+
+                    if (query.EndsWith(cachedItem.Key))
+                        return (cachedItem.Value, Prepend);
+                }
```
amulware: A lot of work in this lock. Any way to simplify this? Since you're ordering anyways (which enumerates the items into a list internally, maybe enumerate them into a list manually, and then sort (maybe with a list.sort or whatever there is) and search outside the lock?

Toggl.Core/SearchBox/SearchEngine.cs
```diff
@@ -0,0 +1,80 @@
...
+        {
+            lock (cache)
+            {
+                // If the term is already in the cache, return its data source.
+                if (cache.TryGetValue(query, out var dataSource))
+                    return (dataSource, NoChange);
+
+                var sortedCacheItems = cache
+                    .OrderByDescending(cachedItem => cachedItem.Key.Length)
+                    .Where(cachedItem => cachedItem.Key.Length < query.Length);
```
amulware: Maybe switch these two lines to order fewer items, since the Where is checking all of them anyways?

Toggl.Core/SearchBox/SearchEngine.cs
amulware: Also, do we care about the difference between startswith/endswith? We could detect the key as substring instead?

Toggl.Core/SearchBox/SearchEngine.cs
amulware: This will be recreated everytime we open the startview, right? I.e. the cache is pretty shortlived.. I don't know if we really care about limiting it?

kornelijepetak: Yes. It is short lived, unless somebody really likes our Start TE view and wants to type into the filter for hours on end.

Toggl.Core/SearchBox/SearchEngine.cs
kornelijepetak: I replaced this with the `Contains` and ditched the enum.

Toggl.Core/SearchBox/SearchEngine.cs
kornelijepetak: Switched.

kornelijepetak:
Tests coming soon.

kornelijepetak:
In several commits, I've done several changes: 

* I've had to eradicate the old interactor and its tests and fix the new tests.
* Additionally, I've moved the data initialization to the background thread because of some tests deadlocks.
* I've abstracted an implementation of the search engine for time entries. 
* I've added interface to search box for testability reasons.
* I've added tests.

@amulware @heytherewill @daividssilverio please recheck

Toggl.Core.Tests/SearchEngine/SearchEngineTests.cs
```diff
@@ -0,0 +1,171 @@
...
+            .Select(i => new MockDataItem(i.ToString()))
+            .ToImmutableList();
+
+        [Fact]
+        public void DoesNotCompleteIfNoInitialDataIsSet()
+        {
+            var waitPeriod = TimeSpan.FromSeconds(2);
+            var searchEngine = new SearchEngine<MockDataItem>(pass);
+
+            var hasCompleted = searchEngine.Get("").Wait(waitPeriod);
```
daividssilverio: Hey @kornelijepetak , these `Wait(waitPeriod)` are all blocking, right? Why not make these tests async?

Toggl.Core.Tests/SearchEngine/SearchEngineTests.cs
```diff
@@ -0,0 +1,171 @@
...
+        }
+
+        [Fact]
+        public void CompletesIfInitialDataIsSet()
+        {
+            var waitPeriod = TimeSpan.FromSeconds(2);
+            var searchEngine = new SearchEngine<MockDataItem>(pass);
+            searchEngine.SetInitialData(ImmutableList<MockDataItem>.Empty);
+
+            var hasCompleted = searchEngine.Get("").Wait(waitPeriod);
```
daividssilverio: Here as well.

Toggl.Core.Tests/SearchEngine/SearchEngineTests.cs
heytherewill: Why not:

`var hasCompleted =await Task.WhenAny(searchEngine.Get("").ContinueWith(_ => throw new XUnitException()), Task.Delay(2)`?

heytherewill: Moreover, why is this even tested? 🤔

Toggl.Core.Tests/SearchEngine/SearchEngineTests.cs
```diff
@@ -0,0 +1,171 @@
...
+
+            hasCompleted.Should().BeFalse();
+        }
+
+        [Fact]
+        public void CompletesIfInitialDataIsSet()
+        {
+            var waitPeriod = TimeSpan.FromSeconds(2);
```
heytherewill: Why? Also why is this test not async?

kornelijepetak: 🤦‍♂

Toggl.Core.Tests/SearchEngine/SearchEngineTests.cs
kornelijepetak: True.

Toggl.Core.Tests/SearchEngine/SearchEngineTests.cs
kornelijepetak: True.

kornelijepetak:
For logistical reasons, I'm yielding the authorship of this PR to whoever wants to complete it. I will simply not have enough time to see what's wrong with the tests.

PR #6490, opened 2019-11-07
🚧 Fix delete glitch

### LINKED ISSUE #6486: Calendar: the day view freezes when a TE is deleted
#5650  🐛 Bug Report
When a TE is deleted, the UI showing the TEs for the given day stops updating. If you scroll up and down, you don't see the horizontal hourly lines and the labels on the left side, older and newer TEs are not shown either. It still registers touch events, so you can edit some TE and you'll be getting haptic feedback, if you try to edit a duration of a TE. You can also create a new TE if you long press at some free slot.

## Context
Build/Branch: ad-hoc from `project/calendar-improvements`
Phone OS: iOS 13.1.2
Phone model: iPhone 7

## Expected behavior
A time entry is deleted, you can continue editing and viewing other TEs.

## To reproduce
- Go to the calendar
- Delete a time entry

sofokli1:
## What's this?
Fixes the delete glitch in the calendar. Also some flickering views seem fixed 

### Relationships
Closes #6486

## Why do we want this?
No freezes

## How is it done?
collectionView reloadItems was causing internal collectionView crashes mostly because we were trying to reload items that are already deleted.

### Why not another way?
this mostly removes code, so it's the best way

### Side effects
Only good stuff

## Review hints
Test before and after

## :squid: Permissions
:squid:


sofokli1:
@arturs105 did you get a change to test this?

PR #6492, opened 2019-11-07
💄 Minor tweaks to the cal project code 

sofokli1:
## What's this?
Addresses some of the requests for changes in the cal project

### Relationships
none

## :squid: Permissions
:squid:


Toggl.Core.UI/Calendar/CalendarLayoutCalculator.cs
```diff
@@ -54,8 +54,15 @@ public List<CalendarItemLayoutAttributes> CalculateTwoHoursOrLessGapsLayoutAttri
             var nextGapStart = dayStart;
             var gaps = new List<(DateTimeOffset, TimeSpan)>();
 
+            DateTimeOffset lastTeEndTime = calendarItems[0].EndTime ?? now;
```
heytherewill: Why first instead of initializting it to null and then checking for nulls at the end? 

The way it's right now, if the first TE doesn't have an endTime it will compare everything against now, something that didn't happen in the previous implementation

sofokli1: int he previous implementation the  `.EndTime ?? now;` was useless since at the top of the chain we check `.Where(te => te.EndTime.HasValue)`. Will update as you suggested but will drop the `now` past altogether

PR #6510, opened 2019-11-08
❔ Add the Option type

### LINKED ISSUE #6451: Add an Option type
Can we even call ourselves software developers in the 21st century without using superior these functional constructs???????

heytherewill:
## What's this?
This is just a monoid in the cathegory of the endofunctors, so what?

### Relationships
Closes #6451

## Why do we want this?
For better describing our types

## How is it done?
This was implemented by @amulware. I simply changed the naming to better match .NET standards (I dream of the day where we will use F# in this codebase) and removed the option to get the value or default, which will force us to always match against the type (as god intended).

### Why not another way?
No reason to import a library when permission was given to use well tested and performant code

## :squid: Permissions
Anyone, anytime


Toggl.Shared/Option.cs
```diff
@@ -0,0 +1,87 @@
...
+            => hasValue ? EqualityComparer<T>.Default.GetHashCode(value) : 0;
+
+        public override string ToString()
+            => hasValue ? $"Some {value}" : "None";
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static implicit operator Option<T>(NothingMaybe _) => None;
+    }
+
+    public static class Maybe
```
amulware: Rename?

Toggl.Shared/Option.cs
```diff
@@ -0,0 +1,87 @@
...
+
+        public static Option<T> FromNullable<T>(T? value) where T : struct =>
+            value.HasValue ? Option<T>.Some(value.Value) : None;
+
+        public static Option<T> Some<T>(T value) => Option<T>.Some(value);
+
+        public static NothingMaybe None => default;
+    }
+
+    public struct NothingMaybe
```
amulware: Tut tut ;)

Toggl.Shared/Option.cs
heytherewill: 🙈

Issue #6239, opened 2019-10-21
Password suggestion in dark mode


Odrakir:
When signing up, if the system fills the password with a suggested one, the field looks weird.

![IMG_BA1634457DE1-1](https://user-images.githubusercontent.com/724743/67189107-ebc70c00-f3ed-11e9-83b9-aca68fde1904.jpeg)

This is true also for light mode, but it's not that bad.

![IMG_2DC06417919A-1](https://user-images.githubusercontent.com/724743/67189119-ee296600-f3ed-11e9-863b-291d9c7db7c4.jpeg)

dshr:
Can't even get this option to show up on iPhone 7, plus nothing comes up in the docs regarding customising this.

heytherewill:
This seems to be ugly on purpose (to denote that you filled your password using a service) and it's ugly on every other app as well, so I guess there's nothing we can do here 🤷‍♂

arturs105:
It seems like we can't do anything about the fact that it's ugly jellow, however we could do something about the size of it - make it smaller and/or add rounded corners. I suspect it is as ugly because we are using a custom text field.

simonrozsival:
If we can detect that this is happening, I think it would be a good solution to:
- set text color to black (the background color is always yellow in both light and dark theme,
- hide the 👁 button.

It would be nice if we could move the "Password" label to be above the yellow background (and with black text color).

simonrozsival:
Maybe we don't have to set the text color to black, but rather to "default text input color". The OS will handle the change of color for us.

One thing that makes it look weird is that the base line "Strong password" text isn't aligned with the base line of the actual password. Maybe it could be possible to fix it somehow.

simonrozsival:
I think this might have been a bug in iOS which they now fixed:
![image](https://user-images.githubusercontent.com/374616/68870912-efebfe00-06fb-11ea-991f-65b8dec2717d.png)
![image](https://user-images.githubusercontent.com/374616/68870957-009c7400-06fc-11ea-8bbd-c215c4ccdf33.png)

I suggest closing this issue for now.

simonrozsival:
OK. Don't listen to me. It doesn't work properly in the simulator for some reason, it looks exactly how Ricardo described it on my device.

simonrozsival:
My proposal is to simplify the login/signup fields. We can keep our custom look, but we won't have the over-complicated animating placeholders/labels, which don't even feel very "native" on iOS IMO.

[...]

PR #6511, opened 2019-11-08
Remove yellow overlay when autocompleting

heytherewill:
## What's this?
This removes the autocomplete overlay that appears when you use a password manager

### Relationships
While looking for a solution to #6239 I stumbled upon [this doc](https://developer.android.com/guide/topics/text/autofill-optimize#highlighted) and figured out: Why not fix it on droid as well?

## Why do we want this?
It currently looks ugly as hell 🤷‍♂ 

## How is it done?
By using this nice little attribute added in Oreo

### Why not another way?
Right now, we are making the overlay transparent. We could, alternatively, use two transparent colors (one for light theme and one for dark theme) so that the overlay is still visible but better matches our color scheme. I'll put @maritoggl in the loop so she can make a decision or involve designers

## :squid: Permissions
Anyone, anytime, but only after @maritoggl gives her thumbs up!


maritoggl:
Go wild. If you need additional input, let me know, but if you can decide on the colors without the designers, I won't be blocking anything.

heytherewill:
Well, screw colors then, transparent is best!

PR #6513, opened 2019-11-08
Draw behind navigation

### LINKED ISSUE #6113: Draw behind navigation bar on Android 10
## Design Improvement Proposal
 it is now strongly recommended for apps to draw behind the navigation bar when running on Android Q+ for a more compelling and modern UX. https://medium.com/androiddevelopers/gesture-navigation-going-edge-to-edge-812f62e4e83e

Basically we want to achieve this: https://miro.medium.com/max/1200/0*5Eqz13ClzwZlKLMV 

## Dependencies
Hopefully - we don't need AndroidX migration for this one

semanticer:
## What's this?
We are now handling navigation bar ourselves in order to go fullscreen on Android 10. This brings nice subtle changes everywhere but most prominently in our bottom menu:

**Before:**
edgetoedgebottomlighold

**Now:**
edgetoedgebottomlight

### Relationships

Closes #6113

## Why do we want this?
We want to show our Android 10 users that we care.

## How is it done?
I did a small refactor around our styles in order to make it less redundant so it was much easier to add new themes for sdk29+. Then I needed to add a couple of inset handling helper extension methods to make sure we layout bottom navigation and toolbar correctly. As a side effect, we now have a nice toolkit for any future inset layout issues. 

### Why not another way?
🤷‍♂️

### Side effects
Tell me if you find anything.

## Review hints
- Make sure you look at both light & dark themes
- Check if everything stayed the same on pre-10 versions
- Check this out on your weird non-Pixel phone

## :squid: Permissions
Me


Toggl.Droid/Extensions/ActivityExtensions.cs
```diff
@@ -155,5 +155,19 @@ public static IObservable<bool> ShowDestructiveActionConfirmationDialog(this Act
 
             throw new ArgumentOutOfRangeException(nameof(type));
         }
+
+        public static void SetQFlags(this Activity activity)
```
heytherewill: Does this work alright when toggling dark theme and the app is open?

Toggl.Droid/Extensions/ViewExtensions.cs
```diff
@@ -88,5 +89,109 @@ public void OnScrollChange(View v, int scrollX, int scrollY, int oldScrollX, int
                 }
             }
         }
+
+
+        public static void UpdatePadding(this View view, int left = -1, int top = -1, int right = -1, int bottom = -1)
```
heytherewill: Why not `int?` instead of using a magic constant?

Toggl.Droid/Extensions/ViewExtensions.cs
```diff
@@ -88,5 +89,109 @@ public void OnScrollChange(View v, int scrollX, int scrollY, int oldScrollX, int
...
+                this.left = left;
+                this.top = top;
+                this.right = right;
+                this.bottom = bottom;
+            }
+        }
+
+        public static void RequestApplyInsetsWhenAttached(this View view)
+        {
+            if (view.IsAttachedToWindow) {
```
heytherewill: No egyptian brackets please

Toggl.Droid/Extensions/ViewExtensions.cs
semanticer: because I didn't expect primitives to have optionals :-) cool

Toggl.Droid/Extensions/ActivityExtensions.cs
semanticer: Yes, it works as expected. At least for me.

Toggl.Droid/Extensions/ViewExtensions.cs
semanticer: Such a waste of lines!

heytherewill:
Please do not merge this until @daividssilverio says that it can be included in 2.7 or has branched out 2.7 :)

Toggl.Droid/Toggl.Droid.csproj
```diff
@@ -745,6 +745,24 @@
     <AndroidResource Include="Resources\menu\GenericSaveMenu.xml" />
     <AndroidResource Include="Resources\menu\DoneOrClearMenu.xml" />
     <AndroidResource Include="Resources\menu\OneButtonMenu.xml" />
+    <AndroidResource Include="Resources\values-night\Environment.xml">
```
daividssilverio: Hey @heytherewill these tags `SubType` and `Generator` inside the `AndroidResource` tags aren't needed, are they?

semanticer: Glad you are not asking me 😅

daividssilverio: I'm never sure about them, and @heytherewill sometimes leaves a comment about them.

daividssilverio:
By the way, this seems like a nice improvement and I don't see why we wouldn't add it to 2.7.

PR #6514, opened 2019-11-08
Support english copy

### LINKED ISSUE #6466: Include information about support being available only in English
We are slowly starting to localize our app.

However, that fact alone can cause the users to attempt to request help from support in non-english languages. While our support indeed can talk various languages, the official support language is English and that is the ONLY language that is available for communication when users are contacting the support.

**Task:** So, we need to include such a note in the Feedback form (or some other place @maritoggl decides) so that we avoid polyglottic problems.

@maritoggl, please provide information about:
* Where to put such note
* What is the exact wording we should put there.

heytherewill:
## What's this?
This adds a piece of copy that properly explains to our users that support is only available in English

### Relationships
Closes #6466 

## Why do we want this?
To avoid possible confusions once the app is out in multiple languages

## How is it done?
Change some copy here and there. As per @maritoggl's instructions, the title of the view has been unified on both platforms

## :squid: Permissions
Anyone, anytime


Toggl.Shared/Resources.resx
```diff
@@ -856,17 +856,16 @@ Use them to add keywords to your time entries.</value>
...
         <value>Something went wrong. Please try again</value>
     </data>
     <data name="ContactUsSend" xml:space="preserve">
         <value>Send</value>
     </data>
     <data name="FeedbackFieldPlaceholder" xml:space="preserve">
-        <value>Let us know what you think. Send us your feedback and questions.</value>
+        <value>Let us know what you think. Send us your feedback and questions.
+
+PS. Our customer support team currently operates in English only. Sorry for any inconvenience this may bring about!</value>
```
dshr: Can we go with slightly less copy? The `may bring about` bit feels a bit awkward
```suggestion
PS. Our customer support team currently operates in English only. Sorry for any inconvenience!</value>
```

PR #6516, opened 2019-11-08
Update widgets when running BG sync

### LINKED ISSUE #6489: PN+Widgets: The widget is not reloaded when the app receives a PN
We must tell the widget that the data reloaded. The `IWidgetService.Start` method should be called before syncing starts in `SyncJobService.runSync` so it can observe the running time entry changes. The widget service will subscribe to the data source changes and notify the widget when the running time entry changes.

heytherewill:
## What's this?
This makes sure that the widgets service are ready to receive updates when sync runs becuase of Push Notifications of the Background service

### Relationships
Closes #6489 

## Why do we want this?
So that widgets are always up-to-date

## How is it done?
Running WidgetsService.Start

### Why not another way?
You tell me. I'd like @daividssilverio to make sure that this is correct

### Side effects
None, I think

## Review hints
Ping @simonrozsival, he knows how to test this

## :squid: Permissions
Anyone, but only before 2.7 is branched out or @daividssilverio agrees to include it in 2.7


daividssilverio:
Feel free to 🦑 this for 2.7.

Issue #6500, opened 2019-11-08
Create a new ReportsViewModel


kornelijepetak:
### **🎯 What needs to be done?**
Create a **new** view model for the reports page.

> During the project it may be useful to retain the old VM just for the sake of potentially salvaging reusable parts. But let's rename the old one and use the `ReportsViewModel` for the new one.

> ⚠ Considering the fact that many things need to be changed in parallel, try to introduce this view model with a mocked/empty parts that will be introduced in later PRs. It makes most sense to implement the infrastructure first and only then to plug in the new parts. For example, even though the donut will be implemented later, this PR can introduce the necessary classes (which would be empty and provide no behavior) to satisfy the API requirements of this view model for the purposes of data propagation. You will be able to prove the correctness by the tests. 

### **📝 Detailed explanation**
The reports should consist of three parts:

* The summary header (#6502)
* The barchart for clocked hours (#6503)
* The donut chart with data grouped by project including the list of projects that are used in the donut (#6506).

For each of those parts, a different view model is created.
The job of this view model is to:
* Handle workspace selection
* Handle time range selection
* Fetch data for selected parameters (workspace + time range)
* Propagate that data to the sub-view models
* Handle and report errors to consumers*
* Handle and report no-data state to consumers
* Handle and report loading state to consumers

And **nothing more**.

> * `Consumers` in this case means activities, fragments, view controllers and similar.

### **🛠 Technical remarks**
In order to simplify the usage, each of the parts of the report should be an item of the RecyclerView/TableView. For that reason, a marker interface `IReportElement` should be used and the view model should return the collection of those to the consumers. 

> Note that it's probably best if the `no-data state` and the `error state` are also items that implement such interface so that in that case, this view model can emit a list of report elements consisting only of such state (no reason to show an empty barchart if there's no data or if there's an eror).

### **🏆 Outcome**
The issue is complete if there is:
* a **new view model** fulfilling all of the above and
* a collection of tests that prove the correctness of the view model

### **👕 Estimated complexity**
`LARGE`

### **🖼 Express your sentiment towards the issue**

![image](https://user-images.githubusercontent.com/7033216/68472679-469b9880-0221-11ea-988b-a9173b5579c9.png)

kornelijepetak:
Completed by #6552 and #6590

PR #6552, opened 2019-11-13
🦴 ReportsViewModel barebone

kornelijepetak:
## What's this?

* Renames the `ReportsViewModel` into `ReportsViewModelOld` so that currently, there's no change in behavior.
* Adds a new `ReportsViewModel` with some barebone to make development of further steps easier.

### Relationships
Touches certain parts of #6500 

## Why do we want this?
So that it's easier to parallelize from here.

### Side effects
None. The new code is not even used anymore.

## Review hints
If you do it commit-by-commit, it'd be trivial.

## :squid: Permissions
Any 📊 developer.


Toggl.Core.UI/ViewModels/Reports/ReportsViewModel.cs
```diff
@@ -1,533 +1,112 @@
...
-                   startDate == default(DateTimeOffset) || endDate == default(DateTimeOffset)
-                       ? Observable.Empty<ProjectSummaryReport>()
-                       : interactorFactory.GetProjectSummary(workspaceId, startDate, endDate).Execute())
-                .Catch(Observable.Return<ProjectSummaryReport>(null))
-                .Subscribe(onReport)
-                .DisposedBy(disposeBag);
-
-            dataSource.Preferences.Current
-                .Subscribe(onPreferencesChanged)
+            workspacesObservable
```
heytherewill: Let's take this rewrite opportunity as a way to get rid of such side-effecty pieces of code

kornelijepetak: Please elaborate a bit more.
How would you go about it?

kornelijepetak:
@heytherewill I have addressed your requests, please check it again. Especially https://github.com/toggl/mobileapp/pull/6552/commits/e993416d7650d81f226290032f0c87a536a88f36

Toggl.Core.UI/ViewModels/Reports/ReportsViewModel.cs
```diff
@@ -48,52 +45,51 @@ public sealed class ReportsViewModel : ViewModel
...
                 .AsDriver(schedulerProvider);
 
-            workspacesObservable
-                .Subscribe(ws => workspaces = ws)
-                .DisposedBy(disposeBag);
-
-            Elements = elementsSubject
+            Elements = Observable.CombineLatest(
+                SelectWorkspace.Elements.WhereNotNull(),
+                SelectTimeRange.Elements,
```
heytherewill: I think tests are failing because this is not considering the emission of the first element

kornelijepetak: The tests can't fail because of that cause no tests are currently testing the new vm. At least they shouldn't be.

PR #6557, opened 2019-11-13
Fix error messages when logging in and signing up

### LINKED ISSUE #6538: The wrong message is shown when logging in with invalid credentials
## 🐛 Bug Report
When logging in with an invalid e-mail and password combination, the generic error message is shown instead of "Password and/or e-mail is wrong".

## Context

Build/Branch: iOS 2.7
Phone OS: iOS 13
Phone model: iPhone X

## Expected behavior
Incorrect e-mail & password combination should show an error message along the lines of "Password and/or e-mail is wrong".

## To reproduce
Just try logging in with gibberish e-mail and password.

simonrozsival:
## Relationships
Closes #6538 

## Why do we want this?
We want the app to work as it used to.

## How is it done?
The problem is that `Upcast` turns any exceptions which are thrown into `AggregateException`s and we can't use type matching to determine the error then. By `await`ing, we can avoid upcasts.

### Why not another way?
I don't know any other way 🤷‍♂ 

## Review hints
Use the simulator to test logging in with incorrect email and password - check the issue for what you should see if this works well.

## Tests

Integration tests: https://app.bitrise.io/build/0015665bb0f98d49

## :squid: Permissions
🦑


heytherewill:
Code looks good, but integration tests are not passing, so don't merge this yet :)

simonrozsival:
I'm not sure what you mean:
image

PR #6560, opened 2019-11-13
Remove the Upcast extension method

simonrozsival:
## What's this?
This removes an extension method and replaces it with `async`/`await`.

### Relationships
Follow-up of #6557

## Why do we want this?
We don't want `AggregateExceptions`, which wrapped any exception thrown in a `Task` continued by an `Upcast` call.

## How is it done?
`async`/`await` solves the problem with the return types.

### Side effects
We can now mostly revert #6436.

## Review hints
Definitely try this in simulator/emulator to confirm, that there are no runtime errors caused by the casts.

## Tests
Integration tests: ~https://app.bitrise.io/build/fc4c70456325c0b7~ https://app.bitrise.io/build/82feba2c0ae13a2a - do not merge this branch unless these tests are passing.
Syncing tests: https://app.bitrise.io/build/bc9b00efdbf1121d - do not merge this branch unless these tests are passing.


heytherewill:
I'd keep the aggregate catching code there; just in case

PR #6577, opened 2019-11-14
Keep the cursor position when changing the duration in the edit view

### LINKED ISSUE #5839: Cursor jumps to the end when inserting emoji
## 🐛 Bug Report
When inserting an emoji in the description field in the edit time entry view, the cursor jumps to the end.
For some reason it only happens with emojis, works fine for regular characters.
Deleting the emoji also works fine, i.e: no cursor jumps.

## Context

Build/Branch: iOS 2.2.1, AppStore build
Phone OS: iPhone
Phone model: any

## Expected behavior
The cursor shouldn't jump.

## To reproduce
Insert an emoji at the beginning or in the middle of the description.

simonrozsival:
## What's this?
This is a bugfix.

### Relationships
Closes #5839, introduced by #5669

## Why do we want this?
We want the UI to behave predictably.

## How is it done?
I explained it in a comment so nobody removes it if they refactor the class.

### Why not another way?
This is a simple solution, I didn't want to refactor this class yet again. I would break something else.

### Side effects
I hope there are none. I haven't tested it with a RTL language just yet, but I assume it must work as well as for LTR.

## Review hints
Try this in a simulator. Try this with an RTL language if you can.

## :squid: Permissions
🦑


Toggl.iOS/Views/TextViewWithPlaceholder.cs
```diff
@@ -40,7 +40,14 @@ public override string Text
             get => base.Text;
             set
             {
+                // Overriding the Text property will place the cursor at the end
+                // of the new text. We don't want to control the position of the
+                // cursor ourselves, so we'll simply remember the cursor position
+                // before we overwrite the text and then we restore it.
+                var originalSelectedRange = SelectedRange;
                 base.Text = value;
+                SelectedRange = originalSelectedRange;
```
sofokli1: what if `value` is shorter, isn't there a chance that the `originalSelectedRange` would end up out of range?

simonrozsival: That's not a problem. The `SelectedRange` is already updated after you press a key or change the text of the text input - deleting things works just fine. I actually think that `base.Text == value` while you're editing the input field. This actually gives me an idea how to simplify this even further.

simonrozsival: Actually it doesn't make it simpler. I think this is as good as we can get without making significant changes to the current code.

PR #6579, opened 2019-11-14
✨ Add SummaryReportElement

### LINKED ISSUE #6502: Create a ReportSummaryViewModel
Create a `ReportSummaryViewModel` that governs the total tracked time and the ratio of tracked billable time.

> This one view model is relatively easy and effectively controls only a few things on top of the reports. For the sake of potentially expanding the amount of summary data in the future, let's keep it as a separate view model.

`SMALL`

pdrj:
## What's this?
This adds `SummaryReportElement`.

### Relationships

Closes #6502 

## Why do we want this?
To later use this element in the view layer.

## How is it done?
By writing some code.

### Why not another way?
Please suggest improvements.

## :squid: Permissions
📊 mates.


Toggl.Core.UI/ViewModels/Reports/ReportSummaryViewModel.cs
```diff
@@ -0,0 +1,39 @@
...
+            : base(navigationService)
+        {
+            Ensure.Argument.IsNotNull(navigationService, nameof(navigationService));
+            Ensure.Argument.IsNotNull(schedulerProvider, nameof(schedulerProvider));
+
+            BillablePercentage = billablePercentageSubject.AsDriver(0, schedulerProvider);
+            TotalTime = totalTimeSubject.AsDriver(TimeSpan.Zero, schedulerProvider);
+        }
+
+        public void UpdateReportData(ProjectSummaryReport report)
```
heytherewill: Why is this whole VM not a `IReportElement`? Why is it not immutable and a new one is created and added to the list of reports elements whenever a new project summary appears?

pdrj: I thought it was `ReportsViewModel`'s responsibility to create an `IReportElement` from data stored in `ReportSummaryViewModel`. 

What is the benefit of having an immutable viewmodel? In my view, view models are mostly mutable by nature (they hold data that might change). What if this VM becomes more complex in the future and has some setup penalty? Recreating it each time then would not be very performant.

If we want it to be immutable, maybe we shouldn't even call it a 'view model', assume it'll be this simple forever and just make it a wrapper that holds the data and is an `IReportElement`?

Toggl.Core.UI/ViewModels/Reports/SummaryReportElement.cs
```diff
@@ -0,0 +1,29 @@
...
+        
+        public static SummaryReportElement FromProjectSummaryReport(ProjectSummaryReport report)
+        {
+            var summaryReportElement = new SummaryReportElement();
+            summaryReportElement.setupWithReportSummary(report);
+            
+            return summaryReportElement;
+        }
+
+        private void setupWithReportSummary(ProjectSummaryReport report)
```
heytherewill: This method should literally be part of the constructor

Toggl.Core.UI/ViewModels/Reports/SummaryReportElement.cs
```diff
@@ -0,0 +1,29 @@
+using System;
+using Toggl.Core.Reports;
+
+namespace Toggl.Core.UI.ViewModels.Reports
+{
+    public sealed class SummaryReportElement : IReportElement
+    {
+        public float BillablePercentage { get; private set; }
```
heytherewill: No need for the private setters if these are set in the constructor
```suggestion
        public float BillablePercentage { get;  }
```

PR #6580, opened 2019-11-14
Fix the problem with holding backspace in the start view

### LINKED ISSUE #5785: Holding backspace doesn't delete the first word
## 🐛 Bug Report
This is a follow up issue to #5474 and #5775. The behavior when holding the backspace button on the description field in the start time entry view has been fixed except for one case, it doesn't delete the first word.

## Context

Build/Branch: develop
Phone OS: iOS
Phone model:

## Expected behavior
Holding the backspace button should delete the first word too, it's not special or anything.

## To reproduce
Have a description long enough so the text field starts deleting by words when holding the backspace button.
Hold the backspace button and delete everything.

## Suggestions
I tried to fix this and couldn't find the problem. I found that changing the text in the text field triggers an update in the view model, which triggers an update in the view, couldn't link this to the problem as everything seems to work as expected, but don't discard that when looking into it again.

simonrozsival:
## What's this?
This PR fixes the issue while it's trying not to break anything that already worked.

### Relationships
Closes #5785 

## Why do we want this?
We want the users to enjoy good UX.

## How is it done?
It's only a few lines of code, but it's not very obvious. It took me a while to find a way to workaround the issue. This view is unfortunately very fragile. If you remove one line, it 💥s. I described what's going on in a comment in the source code so we don't break it next time we touch it.

### Why not another way?
🤷‍♂ I suppose you don't like what I did, I'm not proud of it either 😄 suggestions for improvements are welcome.

### Side effects
No side effects.

## Review hints
Please try breaking this in the simulator. Play with it for a few minutes. Make sure it doesn't crash or freeze, or behave weirdly.

## :squid: Permissions
🦑


simonrozsival:
Damn. I'll revisit this than. Thanks for catching this!

simonrozsival:
@sofokli1 That's one of the limitations (caused by the condition on line 224) and I would consider it an edge case. Maybe somebody will update this in the future if we think it's worth the effort.

simonrozsival:
It's weird 🤔 the button in the bottom bar for tags works well. It should be something minor, I'll look into it after lunch.

simonrozsival:
Nov-25-2019 14-23-44

@arturs105 I rebuilt the app and I can't replicate it. Can you try this in a new simulator to check that you're not testing some older cached build?

Toggl.iOS/ViewControllers/StartTimeEntryViewController.cs
```diff
@@ -200,25 +200,66 @@ public override void ViewDidLoad()
...
-            var attributedText = textFieldInfo.AsAttributedTextAndCursorPosition();
-            if (DescriptionTextView.AttributedText.GetHashCode() == attributedText.GetHashCode())
-                return;
+            // When the user adds a token, then the cursor will be in an empty Text span. This is also
+            // true when the description is totally empty, so we have to take care of this special case.            
+            var likelyJustAddedToken =
+                textFieldInfo.Spans.Count > 1
+                    && textFieldInfo.GetSpanWithCurrentTextCursor()?.Text.Length == 0;
+
+            // When the user tapps a button with the `@` or `#` symbol, we programatically adjust the description
```
arturs105:
```suggestion
            // When the user taps a button with the `@` or `#` symbol, we programatically adjust the description
```

Toggl.iOS/ViewControllers/StartTimeEntryViewController.cs
```diff
@@ -200,25 +200,66 @@ public override void ViewDidLoad()
...
+                    return false;
+
+                if (text.Length == 1)
+                    return text.EndsWith(QuerySymbols.Projects) || text.EndsWith(QuerySymbols.Tags);
+
+                return text.EndsWith($" {QuerySymbols.Projects}", StringComparison.Ordinal)
+                    || text.EndsWith($" {QuerySymbols.Tags}", StringComparison.Ordinal);
+            }
+
+            // Unfortunatelly, this will cause a minor glitch when the user has a ` #` or ` @` somewhere in the middle
```
arturs105:
```suggestion
            // Unfortunately, this will cause a minor glitch when the user has a ` #` or ` @` somewhere in the middle
```

simonrozsival:
The "jumping bar" is a different issue and it is already in the current production version. Please create another issue for that (unless there is already an issue for that).

I'll try to fix the problem with the inserted symbols. At this point I think this whole thing should be rewritten from scratch, it has become a mess.

PR #6588, opened 2019-11-15
Move the eye icon into the password text field

### LINKED ISSUE #6239: Password suggestion in dark mode

simonrozsival:
## What's this?
This PR makes the "Strong Password" overlay of Autofill Password less glitchy.

### Relationships
Closes #6239 

## Why do we want this?
We want the UI to look nice.

## How is it done?
So far the button with the eye icon was placed on top of the password input view, now I changed it to be inside of the `rightView` subview (see https://developer.apple.com/documentation/uikit/uitextfield/1619596-rightview). The yellow overlay will therefore hide it, when it's shown.

### Side effects
This is not a side effect, but I'm still unable to fix the color of the password text in the dark mode. It still appears white on yellow, so it's practically "invisible".

## Review hints
Test this on your iPhone.

## :squid: Permissions
🦑


simonrozsival:
I know that this PR doesn't make it beautiful and so it doesn't solve it 100%, on the other hand I don't think this issue has very high priority and that we shouldn't spend any more time on it.

arturs105:
This breaks the reveal password button on iOS 11 - it is invisible, but tapping it still works though.

simonrozsival:
Damn. It should work across all iOS supported versions (https://developer.apple.com/documentation/uikit/uitextfield/1619596-rightview). I'll look into it.

simonrozsival:
@arturs105 I fixed the problem, I had to set the `Frame` correctly.

BTW this is what the yellow thing looks like now:
image
image

PR #6590, opened 2019-11-18
Implement the rest of the ReportsViewModel backbone

kornelijepetak:
## What's this?
Implements the rest of the `ReportsViewModel` backbone.

For efficiency, I have created this PR without tests so that other elements of the report can utilize the API provided by the ViewModel. Tests are coming in a new PR.

This concludes this view model, the rest of it should come in form of other report parts, as they are being developed.

### Relationships
Partially solves #6500 (no tests)

## Why do we want this?
So that it's easy to add other elements to the reports.

### Side effects
Not that I know of.

## Review hints
Do it commit-by-commit. And pay attention.

## :squid: Permissions
Me.


Toggl.Core.UI/ViewModels/Reports/CompositeReportElement.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System.Collections.Immutable;
+
+namespace Toggl.Core.UI.ViewModels.Reports
+{
+    public abstract class CompositeReportElement : ReportElementBase
```
heytherewill: Where will we want that? I see no reason why the list wouldn't be completely flat

Toggl.Core.UI/ViewModels/Reports/IReportElement.cs
```diff
@@ -1,6 +1,32 @@
...
+
+namespace Toggl.Core.UI.ViewModels.Reports
 {
     public interface IReportElement
     {
     }
+
+    public static class IReportElementExtensions
+    {
+        public static ImmutableList<IReportElement> Flatten(this IEnumerable<IReportElement> elements)
```
heytherewill: If the above is not needed, this also won't be necessary

Toggl.Core.UI/ViewModels/Reports/ReportDonutChartElement.cs
```diff
@@ -0,0 +1,36 @@
+﻿using System.Linq;
+using System.Collections.Immutable;
+using Toggl.Core.Reports;
+using Toggl.Shared.Models.Reports;
+
+namespace Toggl.Core.UI.ViewModels.Reports
+{
+    public sealed class ReportDonutChartElement : CompositeReportElement
```
heytherewill: I see what you are doing here, but I'm not sure I like it. This will require additional logic for calculating/adding/removing items from the UI. I see reason why, so please enlighten me on the reasoning for doing it nested rather than flat

heytherewill: My idea is to make it flat and enforce that you can't create each project segment without adding the header (easily done by marking the constructor as private and forcing the class to return a list of everything).

kornelijepetak: You are confusing me. How will this require any additional adding/removing items from the UI?
The `Elements` property returns an immutable list. 
When report filter changes (workspace or time range), all items will be recalculated and flattened. The recycler will receive simply a list of elements to redraw (which can also properly implement Equals to avoid unnecessary redraws).
Flatten is `O(n)` and even if it weren't, it's still in most cases `<20` elements.

The main reason for this composite was the separation of concerns.
I don't think that `ReportsViewModel` should itself calculate anything.
Alternatively, the `ReportDonutChartElement` doesn't have to be `CompositeReportElement` and it can simply return the list of items that are utilized in the `ReportsViewModel`.

Otherwise, I didn't understand your last comment:
> My idea is to make it flat and enforce that you can't create each project segment without adding the header (easily done by marking the constructor as private and forcing the class to return a list of everything).

kornelijepetak: But the changes I've described ☝️  (removal of `CompositeReportElement` inheritance) require the main view model to be less readable because then I can't treat all report elements the same way. I mean, that is the whole point of the Composite pattern.

kornelijepetak: We've discussed this internally and agreed upon the solution. It was all a misunderstanding.

PR #6592, opened 2019-11-18
Fix null description problems in the UI

### LINKED ISSUE #6591: Support `null` description for time entries
It is possible to create a time entry with `"description": null` and our apps crash or show a blank log (I'm not sure why sometimes it's the former and sometimes the latter) when this happens.

We must check every place we work with TE description (e.g., the edit view, grouping TEs) and add the default value for `null`s.

simonrozsival:
## What's this?
This PR fixes a crash for users, who recently started using the new Toggl app for FitBit.

### Relationships
Closes #6591 

## Why do we want this?
We want to be defensive against the data which we get form other clients.

## How is it done?
I decided to provide default value (empty string) for time entries' descriptions for the `Toggl.Core.Model.TimeEntry` class (which implements `IThreadSafeTimeEntry`).

### Why not another way?
The other possibility is to check every place where we use `Description` and provide a default value there (e.g., when we're grouping TEs).

There's one more solution and it's probably the correct solution:
- add the default value directly during deserialization
- add migration to the database to get rid of any `null`s which arleady made it into the database.

As I'm writing this PR, I'm leaning more to the "proper" solution and I might close this PR and replace it with this implementation. On the other hand, migrations are always a bit scary. I'd like to hear your ideas before I submerge myself into that 💩.

### Side effects
This solution doesn't have any side effects.

## Review hints
You can use the API directly, the Fitbit app, or Realm Studio to create TEs with `null` descriptions.

## :squid: Permissions
🚫  not yet. Let's discuss the alternative solution first.


amulware:
If we agree internally that handling null-descriptions on client-side is necessary, I agree with the current approach here for its simplicity.

simonrozsival:
I think that the internal agreement was that we won't change anything in BE and change users' data, but we will handle `null == ""` on clients.

I think this solution is the simplest and we should go with it. The other solution would be to add the default value whenever we accept some data from the API and already store everything in the DB with `""`. We don't have to add the migration in this case, because only a handful of users use the TogglBit app.

Toggl.Core/Models/FoundationConstructors.cs
```diff
@@ -186,7 +186,7 @@ private TimeEntry(ITimeEntry entity, SyncStatus syncStatus, string lastSyncError
             Billable = entity.Billable;
             Start = entity.Start;
             Duration = entity.Duration;
-            Description = entity.Description;
+            Description = entity.Description ?? string.Empty;
```
kornelijepetak: Why not just `"" `

simonrozsival: Because I like `string.Empty` better.

kornelijepetak: Fun stat of the day:
In Toggl's [Mobile App](https://github.com/toggl/mobileapp/) there are
* 90 occurences of `string.Empty`
* 437 occurences of `""`

simonrozsival: 😢

Toggl.Core.Tests/Models/TimeEntriesTests.cs
```diff
@@ -0,0 +1,166 @@
...
+
+        public class TheDirtyFactoryMethod
+        {
+            [Fact]
+            public void SetsDefaultValueForDescriptionIfDatabaseStoresNullValue()
+            {
+                var databaseEntry = Substitute.For<IDatabaseTimeEntry>();
+                databaseEntry.Description.Returns(null as string);
+
+                var timeEntry = TimeEntry.Clean(databaseEntry);
```
daividssilverio:
```suggestion
                var timeEntry = TimeEntry.Dirty(databaseEntry);
```

Toggl.Core.Tests/Models/TimeEntriesTests.cs
```diff
@@ -0,0 +1,166 @@
...
+
+        public class TheUnsyncableFactoryMethod
+        {
+            [Fact]
+            public void SetsDefaultValueForDescriptionIfDatabaseStoresNullValue()
+            {
+                var databaseEntry = Substitute.For<IDatabaseTimeEntry>();
+                databaseEntry.Description.Returns(null as string);
+
+                var timeEntry = TimeEntry.Clean(databaseEntry);
```
daividssilverio: I guess you meant `Unsyncable`

Toggl.Core.Tests/Models/TimeEntriesTests.cs
```diff
@@ -0,0 +1,166 @@
...
+
+        public class TheCleanDeletedFactoryMethod
+        {
+            [Fact]
+            public void SetsDefaultValueForDescriptionIfDatabaseStoresNullValue()
+            {
+                var databaseEntry = Substitute.For<IDatabaseTimeEntry>();
+                databaseEntry.Description.Returns(null as string);
+
+                var timeEntry = TimeEntry.Clean(databaseEntry);
```
daividssilverio: `CleanDeleted`

Toggl.Core.Tests/Models/TimeEntriesTests.cs
```diff
@@ -0,0 +1,166 @@
...
+
+        public class TheDirtyDeletedFactoryMethod
+        {
+            [Fact]
+            public void SetsDefaultValueForDescriptionIfDatabaseStoresNullValue()
+            {
+                var databaseEntry = Substitute.For<IDatabaseTimeEntry>();
+                databaseEntry.Description.Returns(null as string);
+
+                var timeEntry = TimeEntry.Clean(databaseEntry);
```
daividssilverio: `DirtyDeleted`

Toggl.Core.Tests/Models/TimeEntriesTests.cs
```diff
@@ -0,0 +1,166 @@
...
+
+        public class TheUnsyncableDeletedFactoryMethod
+        {
+            [Fact]
+            public void SetsDefaultValueForDescriptionIfDatabaseStoresNullValue()
+            {
+                var databaseEntry = Substitute.For<IDatabaseTimeEntry>();
+                databaseEntry.Description.Returns(null as string);
+
+                var timeEntry = TimeEntry.Clean(databaseEntry);
```
daividssilverio: `UnsyncableDeleted`

PR #6595, opened 2019-11-18
Start button animation while being pressed

### LINKED ISSUE #5796: Add long press hint on iOS
The start time entry button on the main view has two modes: normal mode as defined by the user preferences when tapping and alternate mode when long pressing.

The problem is this feature is not really easy to find unless you know it's there. We we'll probably need to add some kind of animation to make it clear to the user.

There's a current discussion here: https://toggl.slack.com/archives/CMB9FDE5D/p1565778022020100

simonrozsival:
## What's this?
This PR adds an animation for the "|>" Start button similar to the one on the 

### Relationships
Closes #5796 

## Why do we want this?
It's described in the issue.

## How is it done?
I added an animation. Which is bound to a long press gesture recognizer - a different one from the one which triggers the action.

### Why not another way?
I think there must be a simpler solution 😄 but I think it kinda looks good.

### Side effects
No side effects.

## Review hints
Nov-18-2019 17-56-37

## :squid: Permissions
🦑


Toggl.iOS/ViewControllers/MainViewController.cs
```diff
@@ -596,6 +596,9 @@ private void prepareViews()
             //Prevent bounces in UIScrollView
             AutomaticallyAdjustsScrollViewInsets = false;
 
+            // The "+" button
```
arturs105: This comment is a bit confusing - it is the "+" button only in manual mode. I think the method name is descriptive enough.

Toggl.iOS/ViewControllers/MainViewController.cs
```diff
@@ -639,6 +642,57 @@ private void prepareViews()
...
+                                    cancellationToken: cts.Token,
+                                    completion: () =>
+                                    {
+                                        AnimationExtensions.Animate(bounceAnimationDuration / 2, noDelay, Curves.Bounce,
+                                            () => StartTimeEntryButton.Transform = normalScale,
+                                            cancellationToken: cts.Token);
+                                    });
+                            },
+                            cancellationToken: cts.Token);
+                        break;
```
arturs105: 🙈  So much nesting. Can we do something abut this? At least name the animations with nested functions.

Toggl.iOS/ViewControllers/MainViewController.cs
```diff
@@ -639,6 +642,76 @@ private void prepareViews()
...
+        {
+            const double longPressMinimumPressDuration = 0.5; // default OS long press duration is 0.5s
+            const double startAnimatingAfter = 0.1;
+            const double bounceAnimationDuration = 0.1f;
+            const double shrinkingAnimationDuration = longPressMinimumPressDuration - startAnimatingAfter - bounceAnimationDuration;
+            nfloat noDelay = 0.0f;
+
+            var shrunk = CGAffineTransform.MakeScale(0.9f, 0.9f);
+            var bigger = CGAffineTransform.MakeScale(1.05f, 1.05f);
+            var normalScale = CGAffineTransform.MakeScale(1f, 1f);
```
arturs105: These can be made into constants

Toggl.iOS/ViewControllers/MainViewController.cs
```diff
@@ -639,6 +642,76 @@ private void prepareViews()
...
+                    Curves.Bounce,
+                    () => StartTimeEntryButton.Transform = normalScale,
+                    cancellationToken: cts.Token);
+            }
+
+            var press = new UILongPressGestureRecognizer(startButtonAnimation);
+            press.MinimumPressDuration = startAnimatingAfter;
+            press.ShouldRecognizeSimultaneously = (_, __) => true;
+
+            StartTimeEntryButton.AddGestureRecognizer(press);
```
arturs105: Not a big fan of putting executed code after nested functions

Toggl.iOS/ViewControllers/MainViewController.cs
```diff
@@ -596,6 +596,8 @@ private void prepareViews()
             //Prevent bounces in UIScrollView
             AutomaticallyAdjustsScrollViewInsets = false;
 
+            prepareStartButtonLongPressAnimation();
```
arturs105: There are similar methods executed a bit later in `prepareViews`. I think `prepareStartButtonLongPressAnimation ` would fee better if he was together with his friends: `prepareWelcomeBackViews` and `prepareEmptyStateView` on line 628.

Toggl.iOS/ViewControllers/MainViewController.cs
simonrozsival: These are "local constants" of the animation. I decided to leave them here because they won't be reused and this VC is so big that it would make the code even more confusing and harder to understand. This way, they don't pollute the scope of the VC and they aren't suggested by Intellisense.

Issue #5954, opened 2019-09-24
Measure what's wrong with our startup


heytherewill:
Our startup times aren't the best. We've been working on improving things, but we need to do more.

This issue is iOS specific, and in it we should list all things that can be improved (be they .NET specific or iOS specific) to make our startup as close to instantaneous as possible

To consider this issue closed we should have a write up on what can be improved and a list of follow-up issues to fix those things

simonrozsival:
When measured in simulator, these parts of the initialization code have non-trivial durations:
```
EnsureInitialized - 80 ms
AppStart Initialization - 60 ms
GetAccessLevel - 180 ms
navigateAccordingToAccessLevel - 265 ms
LoginViewController.ViewDidLoad - 107 ms
```

The whole app start time when the user is not logged in is not very bad. On my iPhone 7 it is less than 2 seconds.

When logged in, there are these two additional non-trivial methods:
```
AppDelegate.loginWithCredentialsIfNecessary - 400 ms
AppDelegate.navigateAccordingToAccessLevel - 1250 ms
```

The start up time when logged in takes roughly 3.5 seconds on my iPhone 7.

heytherewill:
Did you measure using a release build? Debug ones are really misleading in this sense. Also, the simulator data is really worthless for performance debugging purposes, so let's stick to actual device data.

If you drill down `navigateAccordingToAccessLevel`, what are the most problematic methods?
Same thing for `loginWithCredentialsIfNecessary`, 400ms is not ideal

simonrozsival:
It was all debug builds. I'll try measuring it again tomorrow.

simonrozsival:
I ran this on my iPhone 7 and this is what I got similar results:
```
// when logged in
EnsureInitialized - 77 ms
AppStart - 76 ms
GetAccessLevel - 150 ms
oginWithCredentialsIfNecessary - 140 ms
navigateAccordingToAccessLevel - 720 ms
```
Some of the times are better, but it's not very different from the simulator. I'll look into `navigateAccordingToAccessLevel` and find out what exactly is taking so long.

simonrozsival:
I broke down the `navigateAccordingToAccessLevel`:
```
SyncManager.ForceFullSync - 186 ms -- this could happen in a different thread and we can shave off most of these 180 ms
Load MainTabBarViewModel - 231 ms
MainTabBarViewModel.Initialize - 184 ms
ViewControllerLocator.GetViewController - 127 ms
```

[...]

PR #6618, opened 2019-11-19
Improve startup time: Remove unnecessary recreation of WidgetsService on startup

simonrozsival:
## What's this?
This PR tries to avoid totally unnecessary work during startup.

### Relationships
Ref #5954 

## Why do we want this?
We want 200ms faster startup on iOS.

## How is it done?
Don't recreate what hasn't been created yet. Disposing required to initialize the service in the first place. Go figure.

### Side effects
~I'm not sure what will be the consequences of the first commit. Keeping just the second commit will solve the startup times, so we might do just that and we can create a follow-up issue to revisit this further.~ _I reverted the first commit, now it should be safe. I'll create a follow-up issue._

## :squid: Permissions
🚫 Not until we are sure this is what we want


kornelijepetak:
I am really confused about how this helps with anything.

If the value is not yet created, then `?.Dispose()` should be a no-op and the creation of Lazy should not take 200 ms.

@heytherewill any comments?

PR #6627, opened 2019-11-19
Fix fragment not added crash

### LINKED ISSUE #6044: java.lang.IllegalStateException: Fragment already added: MainFragment
## 💥 Crash Report
This type of crash generally happens when we try to add the same fragment twice. The most common issue is with the DialogFragment being added twice before dismissing.

It looks like this event plays some major role: 
```
ReportsSuccess
Source:“ShortcutToday”
LoadingTime:“1000”
ProjectsNotSynced:“0”
TotalDays:“7”
```
### 👥 Impact
This impacts around 0.08% of android 2.4.1 users

### 🔗 Context
??

### 🛰 Link
https://appcenter.ms/orgs/Toggl/apps/Toggl-Android/crashes/errors/1324550673u/overview

`java.lang.IllegalStateException: Fragment already added: MainFragment`
```cs
java.lang.IllegalStateException: Fragment already added: MainFragment{123f7b5 #1 id=0x7f0d017e}
  at android.support.v4.app.FragmentManagerImpl.addFragment(FragmentManager.java:1916)
  at android.support.v4.app.BackStackRecord.executeOps(BackStackRecord.java:765)
  at android.support.v4.app.FragmentManagerImpl.executeOps(FragmentManager.java:2625)
  at android.support.v4.app.FragmentManagerImpl.executeOpsTogether(FragmentManager.java:2411)
  at android.support.v4.app.FragmentManagerImpl.removeRedundantOperationsAndExecute(FragmentManager.java:2366)
  at android.support.v4.app.FragmentManagerImpl.execPendingActions(FragmentManager.java:2273)
  at android.support.v4.app.FragmentManagerImpl$1.run(FragmentManager.java:733)
  at android.os.Handler.handleCallback(Handler.java:873)
  at android.os.Handler.dispatchMessage(Handler.java:99)
  at android.os.Looper.loop(Looper.java:214)
  at android.app.ActivityThread.main(ActivityThread.java:7037)
  at java.lang.reflect.Method.invoke(Method.java:-2)
  at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:494)
  at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:965)
```

heytherewill:
## What's this?
This is a best effort attempt at solving a bug we can't really reproduce reliably

### Relationships
Closes #6044 

## Why do we want this?
This is the most common crash, so Kornelije will be happy!

## How is it done?
This basically tries two different approaches at solving the problem. I was really at a loss here, since the lack of reproducibility makes it hard to figure out what's going on.

For context, we *are* checking for the `IsAdded` property before adding our fragments, but that is not preventing the app from crashing once it remains on the background for a while. I could not reproduce it (not even using the `Don't Keep Activities` setting), but that's my assumption after checking the crash reports.

After reading [this answer](https://stackoverflow.com/a/22486246/3465182) I thought it is worth giving it a shot. It's side effect free and we don't really lose much by ensuring transactions are run before we add fragments. That's the one I'm the most hopeful about and you can see it in the first commit.

The second commit (and second possible solution) comes from [here](https://stackoverflow.com/a/24629280/3465182). I'm using tags rather than IDs, but the gist is the same: We check whether the fragment manager is holding an instance of the fragment we're trying to add or not. If it is, we don't do anything. I made it an extension method.

*UPDATE*
I followed what @pdrj said and I will not release the second commit. If the first one does not cut it, we can always return to this PR and ship the other thing

### Why not another way?
You tell me, I'd love to hear other possible ideas.

### Side effects
The second commit might be problematic, that's why I want other droid devs to chime in. I have no idea what might happen if a fragment instance is kept but the fragment is reporting not being added! Right now we crash (and since we've been on the background for a while, the user probably notices it as a restart), but I have no idea what kind of weird state the user might see, so if anyone has experience on that I'd love to hear.

## :squid: Permissions
Only myself after a couple of androids have chimed in


Toggl.Droid/Activities/MainTabBarActivity.cs
```diff
@@ -233,15 +233,25 @@ private void onTabSelected(IMenuItem item)
 
         private void showFragment(Fragment fragment)
         {
+            SupportFragmentManager.ExecutePendingTransactions();
```
pdrj: Not a `C#` rx expert, so just want to make sure - this will always execute on main thread, right? Subscription happens in `InitializeBindings` which gets called in `OnCreate`, so I guess all is good?

pdrj:
I forgot to mention that I tested it on a device and it works OK (with `Don't keep activities` as well).

heytherewill:
@pdrj I think I will do as you say and rollback the change. If anything, we can just pick that commit from this PR again and see if that works :)

Toggl.Droid/Activities/MainTabBarActivity.cs
heytherewill: Yes, this is guaranteed to run on the Main Thread (after all, we are doing transactions right below, which are also meant to only run on the UI thread)

PR #6628, opened 2019-11-19
Main log refactor

### LINKED ISSUE #6083: Merge suggestions, time entries and rating in a single list
As of right now, the main log on Android is composed by:

- The recycler view to display the time entries
- A header that is also a recycler view, creating nesting
  - This made sense when suggestions were horizontal
- Some logic to inject a new item when we want to show the rating view

This leads to all sorts of rendering weirdness, because:

- Nested RecyclerViews simply don't play well
- The main recycler view gets weird update everything messages when it wants to show the rating view
- It's tricky to reason about whether the logic is right or not
- Everything looks very flashy during startup

We should refactor this whole infra to be a single list where all components inherit from a base type, this way we can simply treat everything as a single list (which it is) and make items easier to update.

@Odrakir says that iOS doesn't need it that much but that he is not opposed to implementing it, so given how much benefit android will get this should be done properly, in the shared layer

semanticer:
## What's this?
Merge suggestions, time entries and user feedback in a single list and refactoring out `MainRecyclerAdapter` to handle all of this.

### Relationships
Closes #6083

## Why do we want this?
We want to simplify and flatten our most important `RecyclerView` 

## How is it done?
`MainViewModel` has a new property, `MainLogItems` that combines both time entries and suggestions into a single collection (still probably needs tests!). Then we got rid of the `ReactiveSectionedRecyclerAdapter` and introduced the new `MainRecyclerAdapter` that supports all view types we need and it inherits from already battle-tested `BaseRecyclerAdapter`. The idea was to get rid of any special headers, sections etc - everything is just an item now and we use view types to differentiate. 

### Why not another way?
This seemed like the most straight forward way. 

### Side effects
There shouldn't be any but who knows...  🤷

## Review hints
I'm still not super sure the feedback view implementation works in all cases.

## :squid: Permissions
Me Me


semanticer:
TODO:
- [ ]  Fix TapSubjects
- [ ]  Cleanup some suggestions mess
- [ ]  Maybe some new tests

Toggl.Droid/Adapters/MainRecyclerAdapter.cs
```diff
@@ -1,161 +1,167 @@
...
-        public const int UserFeedbackViewType = 3;
+        public const int TimeEntryLogItemViewType = 1;
+        public const int SuggestionLogItemViewType = 2;
+        public const int DaySummaryViewType = 3;
+        public const int SuggestionsHeaderViewType = 4;
+        public const int UserFeedbackViewType = 5;
 
-        private readonly Context context;
-        private readonly ITimeService timeService;
+        private bool isRatingViewVisible;
```
heytherewill: Why not simply make the rating view another item that is included in the main log list?

semanticer: Rating view IS actually another item in main log list but this flag decides if it is included in SetItems(). So it looks the same way from the outside as the previous API but it is implemented as any other log item.

heytherewill: I don't like this kind of logic in the adapter. We are specifically moving away from that. Simply including or not including the rating view would make things much simpler and more testable (we had problems in the past where the rating view simply did not appear and that affected our ratings)

Toggl.Droid/Adapters/MainRecyclerAdapter.cs
```diff
@@ -1,161 +1,167 @@
...
+        protected override void SetItems(IImmutableList<MainLogItemViewModel> newItems)
         {
-            if (viewType == SuggestionViewType)
+            if (isRatingViewVisible && Items.Count > 0)
             {
-                var suggestionsView = LayoutInflater.FromContext(parent.Context).Inflate(Resource.Layout.MainSuggestions, parent, false);
-                var mainLogSuggestionsListViewHolder = new MainLogSuggestionsListViewHolder(suggestionsView, SuggestionsViewModel);
-                return mainLogSuggestionsListViewHolder;
+                var ratingIndex = newItems.Select((value, index) => new {index, value})
+                    .Single(p => p.value is DaySummaryViewModel)?.index ?? 0;
```
daividssilverio: This crashes with 
```
System.InvalidOperationException: Sequence contains more than one matching element
  at System.Linq.Enumerable.Single[TSource] (System.Collections.Generic.IEnumerable`1[T] source, System.Func`2[T,TResult] predicate) [0x0004a] in <ea96a889a49d4f9bbdce079153f71d9d>:0
```

it looks like we probably want to use `FirstOrDefault` instead.

Toggl.Droid/ViewHolders/MainLog/MainLogUserFeedbackViewHolder.cs
```diff
@@ -34,98 +34,97 @@ public class MainLogUserFeedbackViewHolder : RecyclerView.ViewHolder
...
+            laterButton = ItemView.FindViewById<TextView>(Resource.Id.UserFeedbackLaterButton);
+
+            questionGroup = ItemView.FindViewById<Group>(Resource.Id.QuestionView);
+            impressionGroup = ItemView.FindViewById<Group>(Resource.Id.ImpressionView);
+
+            userFeedbackTitle.Text = Shared.Resources.RatingTitle;
+            yesText.Text = Shared.Resources.RatingYes;
+            noText.Text = Shared.Resources.RatingNotReally;
+            laterButton.Text = Shared.Resources.Later;
+
```
daividssilverio: 🔥 the extra line

Toggl.Droid/Fragments/MainFragment.cs
```diff
@@ -136,10 +136,9 @@ public override void OnViewCreated(View view, Bundle savedInstanceState)
                 .Subscribe(onSyncChanged)
                 .DisposedBy(DisposeBag);
 
-            mainRecyclerAdapter = new MainRecyclerAdapter(Context, ViewModel.TimeService)
+            mainRecyclerAdapter = new MainRecyclerAdapter()
             {
-                SuggestionsViewModel = ViewModel.SuggestionsViewModel,
-                RatingViewModel = ViewModel.RatingViewModel,
+                UserFeedbackViewModel = ViewModel.RatingViewModel
```
daividssilverio: Attaching the `MainFragment` as the view of the `ViewModel.RatingViewModel` before setting it to adapter could prevent a crash when selecting the `tell us what's wrong` in the feedback view holder.
That could be fixed in the shared layer as well, but I'm not sure how that would affect iOS.

Toggl.Droid/Adapters/MainRecyclerAdapter.cs
semanticer: Understood, I will change this.

Toggl.Droid/ViewHolders/MainLog/MainLogSuggestionsListViewHolder.cs
```diff
@@ -0,0 +1,103 @@
...
+using Toggl.Droid.Widgets;
+using Android.Appwidget;
+using Android.App;
+using Toggl.Core.UI.ViewModels.TimeEntriesLog;
+
+namespace Toggl.Droid.ViewHolders.MainLog
+{
+    public class MainLogSuggestionsListViewHolder// : BaseRecyclerViewHolder<MainLogItemViewModel>
+    {
+//        private SuggestionsViewModel suggestionsViewModel;
```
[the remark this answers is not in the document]
semanticer: 🔥 ✅

Toggl.Droid/ViewHolders/BaseRecyclerViewHolder.cs
```diff
@@ -10,7 +10,7 @@ public abstract class BaseRecyclerViewHolder<T> : RecyclerView.ViewHolder
     {
         private bool viewsAreInitialized = false;
 
-        public ISubject<T> TappedSubject { get; set; }
+        public ISubject<T> ContinueSuggestionSubject { get; set; }
```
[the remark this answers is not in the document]
semanticer: Wow, who did this?  :-D

semanticer:
@daividssilverio 
> I don't like all of the ViewHolders being MainLogItemViewModels directly and all the casts around, but I guess there's no way around it.

that makes two of us :-( I don't see any solution without some bigger changes to BaseRecyclerAdapter but I didn't want to go that far at this point

Toggl.Droid/ViewHolders/MainLog/MainLogCellViewHolder.cs
```diff
@@ -113,19 +120,19 @@ public void HideSwipeBackgrounds()
 
         private void onContinueClick(object sender, EventArgs e)
         {
-            var continueMode = Item.ViewModel.IsTimeEntryGroupHeader
+            var continueMode = ((TimeEntryLogItemViewModel) Item).IsTimeEntryGroupHeader
```
daividssilverio: Maybe create a variable for this cast.

Toggl.Droid/ViewHolders/MainLog/MainLogCellViewHolder.cs
```diff
@@ -113,19 +120,19 @@ public void HideSwipeBackgrounds()
...
+            ContinueButtonTappedSubject?.OnNext(new ContinueTimeEntryInfo(((TimeEntryLogItemViewModel) Item), continueMode));
         }
 
         private ConstraintLayout.LayoutParams getDurationPaddingWidthDependentOnIcons()
         {
             var whitePaddingWidth =
                 72
-                + (Item.ViewModel.IsBillable ? 22 : 0)
-                + (Item.ViewModel.HasTags ? 22 : 0);
+                + (((TimeEntryLogItemViewModel) Item).IsBillable ? 22 : 0)
```
daividssilverio: And this ones.

Toggl.Droid/Fragments/MainFragment.cs
daividssilverio: @semanticer Did you take a look at this? If you rather not touch it in this PR, do you mind creating an issue for it?

Toggl.Droid/Fragments/MainFragment.cs
```diff
@@ -212,10 +215,6 @@ public override void OnViewCreated(View view, Bundle savedInstanceState)
                 .Subscribe(onEmptyStateVisibilityChanged)
                 .DisposedBy(DisposeBag);
 
-            ViewModel.ShouldShowRatingView
-                .Subscribe(setupRatingViewVisibility)
```
daividssilverio: Please remove the leftover field as well.

semanticer:
@daividssilverio I'm still not sure about the rating view - rate button, for example, doesn't do anything. Is this only debug feature or.. ? I guess I'm not 100% overall about rating view's behavior so it's hard for me to test it.

daividssilverio:
@semanticer the rate button just have the wrong package set.
The `Toggl.Droid/Services/RatingServiceAndroid.cs` is sending an intent to the PlayStore with the debug package, which doesn't exist, that's why it won't do anything in debug.

You can change the service to the production package and it should work.

Toggl.Droid/Fragments/MainFragment.cs
semanticer: I'm not even sure I understand the issue here so maybe you should create this :-D

Toggl.Core.UI/ViewModels/TimeEntriesLog/Identity/SuggestionsSectionKey.cs
```diff
@@ -0,0 +1,21 @@
...
+
+        public override bool Equals(object obj)
+        {
+            if (ReferenceEquals(null, obj)) return false;
+            if (ReferenceEquals(this, obj)) return true;
+            return obj is SuggestionsSectionKey other && Equals(other);
+        }
+
+        public override int GetHashCode()
+            => 13853211;
```
daividssilverio: Interesting detail: this fixed hash code can collide with other hash codes from the collection being compared during the `DiffUtil.Callback`. In fact, during my tests I actually got to accidentally achieve that, and the animation breaks of course.

Toggl.Core.UI/ViewModels/TimeEntriesLog/Identity/UserFeedbackKey.cs
```diff
@@ -0,0 +1,23 @@
...
+
+        public override bool Equals(object obj)
+        {
+            if (ReferenceEquals(null, obj)) return false;
+            if (ReferenceEquals(this, obj)) return true;
+            return obj is SuggestionsSectionKey other && Equals(other);
+        }
+
+        public override int GetHashCode()
+            => 11235813;
```
daividssilverio: Another hash code that can accidentally collide during the `AreItemsTheSame` callbacks of the diff util.

Toggl.Core.UI/ViewModels/TimeEntriesLog/Identity/SuggestionsSectionKey.cs
daividssilverio: Maybe pick a big prime number, that might help avoiding clashes.

semanticer:
@kornelijepetak Thank you for your very detailed complaint. I completely agree, know when I look at it I don't even understand why I did it :-) Probably seen this in our codebase.  
Solution 1: I'm currently thinking about some proper mapping. 
Solution 2: I'm not sure if I understand completely, you suggest using equality for both `AreContentsTheSame` AND `AreItemsTheSame` ?

kornelijepetak:
> @kornelijepetak Thank you for your very detailed complaint. I completely agree, know when I look at it I don't even understand why I did it :-) Probably seen this in our codebase.
> Solution 1: I'm currently thinking about some proper mapping.
> Solution 2: I'm not sure if I understand completely, you suggest using equality for both `AreContentsTheSame` AND `AreItemsTheSame` ?

1. You CAN'T use proper mapping in Main Log, because Time Entries already cover ALL of the `long` range: DB items are positive values, local unsynced identifiers are negative values. You will not be able to map it uniquely and thus should not use it in such a way.

2. I mean, if you just implemented `Equals` as I've described above, even using `EquatableDiffingStrategy<T>` would be correct albeit not performance-optimal. Optimal solution would be to implement your own `IDiffingStrategy<T>` and implement both  `AreContentsTheSame` AND `AreItemsTheSame` properly. `AreContentsTheSame` would of course need to rely on complete `.Equals` comparison, but `AreItemsTheSame` could be smarter in cases where compared items are of a type that actually does have a unique identifier (i.e. Time Entries).

So, the `AreItemsTheSame` would simply do something like this:

```cs
public bool AreItemsTheSame(T item, T other)
{
    if(item.GetType() != other.GetType())
		return false;

	// Shortcut check this instead of Equals because 
	// Equals does a lot of comparisons for the time entries
	if(item is IThreadSafeTimeEntry timeEntryA && other is IThreadSafeTimeEntry timeEntryB) {
		return timeEntryA.Id == timeEntryB;
	}

	// For all other items, Equals is rather fast, so just do this
    return item.Equals(other);
}
```

_I assumed that the underlying TE class is IThreadSafeTimeEntry, replace this with whatever you actually get._

Don't worry about the `is` pattern matching. It will not create performance issues.

semanticer:
@kornelijepetak I don't think we should be using equals in AreItemsTheSame unless there is no other way and not because of the performance reasons but because we actually want to return **true** when contents are **different** but identity is the same.

kornelijepetak:
> @kornelijepetak I don't think we should be using equals in AreItemsTheSame unless there is no other way and not because of the performance reasons but because we actually want to return **true** when contents are **different** but identity is the same.

That is true. My example was not supposed to be complete, but to show difference between checking the exact ID of a identifiable entity and the rest. Of course, wherever the two methods (items/contents) would not return the same result in all cases, custom logic should be used, but it has to be matched by exact type and then drill down into specific way of comparing elements for that exact type. 

As the matter of fact, for (almost) all non-unique-identifiable elements, I believe we're going to have only one of each, so just comparing their types in `AreItemsTheSame` should be enough. I think only TE suggestion rows would be different.

daividssilverio:
So, after the latest changes, the diffing was still broken.
Using the new idea of the `TypeIdentifierEqualityDiffingStrategy`, we can modify it a little bit so that we compare the their inner `IMainLogKey`s instead of the view model type directly.

That's mostly because of how our single time entries and grouped time entries share the same `Type`.
That is causing some confusing when comparing in the `AreItemsTheSame`.

We have a number of options here, like:
1. Making the adapter diffing strategy specific (which seems fine for now) since we `TypeIdentifierEqualityDiffingStrategy`'s only client is the main log adapter; 
2. Making the `TypeIdentifierEqualityDiffingStrategy` take another generic type in its template. It would look like something like this: 
```
public sealed class TypeIdentifierEqualityDiffingStrategy<T, KT> : IDiffingStrategy<T>
        where T : IEquatable<T>
        where KT : IEquatable<KT>
```

and 
```
public bool AreItemsTheSame(T item, T other)
        {
            var itemDiffable = (IDiffable<KT>) item;
            var otherDiffable = (IDiffable<KT>) other;

            if (itemDiffable.Identity.GetType() != otherDiffable.Identity.GetType())
            {
                return false;
            }
            
            var itemDiffableById = (IDiffableByIdentifier<T>)item;
            var otherDiffableById = (IDiffableByIdentifier<T>)other;

            return itemDiffableById.Identifier == otherDiffableById.Identifier;
        }
```
; or
3. Something completely different

@kornelijepetak @heytherewill could give better options considering how generics work best in C#.
Notice how my suggestion isn't type safe, it would crash if the items didn't implement `IDiffable<KT>`

daividssilverio:
Oh, my suggestion above can be made type-safe by doing: 

```
public sealed class TypeIdentifierEqualityDiffingStrategy<T, KT> : IDiffingStrategy<T>
        where KT : IEquatable<KT>
        where T : IEquatable<T>, IDiffable<KT>
```

Toggl.Droid/Adapters/DiffingStrategies/TypeIdentifierEqualityDiffingStrategy.cs
```diff
@@ -0,0 +1,45 @@
+﻿using System;
+using Android.Support.V7.Widget;
+using Toggl.Core.UI.Collections;
+using Toggl.Core.UI.Interfaces;
+using Toggl.Shared;
+
+namespace Toggl.Droid.Adapters.DiffingStrategies
+{
+    public sealed class TypeIdentifierEqualityDiffingStrategy<T, KT> : IDiffingStrategy<T>
+        where T : IEquatable<T>
```
daividssilverio:
```suggestion
        where T : IEquatable<T>, IDiffable<KT>, IDiffableByIdentifier<T>
```

Toggl.Droid/Adapters/DiffingStrategies/TypeIdentifierEqualityDiffingStrategy.cs
```diff
@@ -0,0 +1,45 @@
...
+        }
+
+        public bool AreContentsTheSame(T item, T other)
+        {
+            return item.Equals(other);
+        }
+
+        public bool AreItemsTheSame(T item, T other)
+        {
+            var itemDiffable = (IDiffable<KT>) item;
```
daividssilverio: These lines can be simplified as:
```
            if (item.Identity.GetType() != other.Identity.GetType())
            {
                return false;
            }
            
            return item.Identifier == other.Identifier;
```

considering my previous suggestion

Toggl.Core.UI/ViewModels/TimeEntriesLog/Identity/DayHeaderKey.cs
```diff
@@ -12,7 +12,7 @@ public DayHeaderKey(DateTime date)
         }
 
         public long Identifier()
-            => GetHashCode();
+            => date.GetHashCode();
```
kornelijepetak: How is `date.GetHashCode` unique?

Toggl.Droid/Adapters/DiffingStrategies/TypeIdentifierEqualityDiffingStrategy.cs
```diff
@@ -0,0 +1,40 @@
...
+
+        public bool AreItemsTheSame(T item, T other)
+        {
+            if (item.GetType() != other.GetType())
+            {
+                return false;
+            }
+
+            var itemDiffable = (IDiffableByIdentifier<T>)item;
+            var otherDiffable = (IDiffableByIdentifier<T>)item;
```
kornelijepetak: This is a copy/paste error.
The `otherDiffable` should cast the `other` value, not the `item`.

Toggl.Droid/Adapters/DiffingStrategies/TypeIdentifierEqualityDiffingStrategy.cs
```diff
@@ -22,15 +24,18 @@ public bool AreContentsTheSame(T item, T other)
 
         public bool AreItemsTheSame(T item, T other)
         {
-            if (item.GetType() != other.GetType())
+            var itemDiffable = (IDiffable<KT>) item;
+            var otherDiffable = (IDiffable<KT>) other;
+
+            if (itemDiffable.Identity.GetType() != otherDiffable.Identity.GetType())
```
kornelijepetak: I don't think there's a need to compare the Identities' type. Simply compare the `item` and the `other`.

Toggl.Core.UI/ViewModels/TimeEntriesLog/Identity/DayHeaderKey.cs
kornelijepetak: https://referencesource.microsoft.com/#mscorlib/system/datetime.cs,979
This is in no way unique.

You can use `date.ToBinary()` which will give you a long. That one uniquely represent a certain DateTime.

Toggl.Droid/Adapters/DiffingStrategies/TypeIdentifierEqualityDiffingStrategy.cs
```diff
@@ -0,0 +1,45 @@
...
+            {
+                return false;
+            }
+
+            var itemDiffableById = (IDiffableByIdentifier<T>)item;
+            var otherDiffableById = (IDiffableByIdentifier<T>)item;
+
+            return itemDiffableById.Identifier == otherDiffableById.Identifier;
```
kornelijepetak: This is still not correct because of the following.

In the `TimeEntriesGroupKey` you have this:

```cs
public long Identifier()
    => groupId?.GetHashCode() ?? 0;
```

This is in no way unique and you can confuse the diffing algo into treating two groups that are not the same as they are.

Since there's no theoretical way to conflate the identity of the TE Group into a `long` (it depends on multiple overlapping longs and some time etc), you need to compare it with the `Equals`.

```cs
return itemDiffableById.Identifier.Equals(otherDiffableById.Identifier);
```

would of course be less performant for keys that are actually able to conflate to `long`, so let's not do this.

But, since for groups, you actually have to check `Equals` so you have to do something like this:

```cs
if(item is TimeEntryLogItemViewModel te1 && other is TimeEntryLogItemViewModel te2)
{
    return te1.Identity.Equals(te2.Identity);
}

// Do the similar check as above for all items that do not intrinsically posses a unique long key.

return itemDiffableById.Identifier == otherDiffableById.Identifier;
```

What you provide as the result of the `long Identifier()` function does not even matter, because it's impossible to create a meaningful value. That's why I was always screaming that this concept of `Identifier` is problematic, but nobody would listen at the time. 🤷‍♂

Toggl.Droid/Adapters/DiffingStrategies/TypeIdentifierEqualityDiffingStrategy.cs
semanticer: I have experimented with this `TimeEntryBaseLogItemView` idea and it would probably need more thought and more extensive changes. One of the issues is the `LogItemVisualizationIntent` enum and it's connected logic. I suggest to explore this in the future but not now in this PR.

Toggl.Droid/Adapters/DiffingStrategies/TypeIdentifierEqualityDiffingStrategy.cs
semanticer: @kornelijepetak Your last suggestion would take us from "universal" `TypeIdentifierEqualityDiffingStrategy` to the concrete implementation of DiffStrategy. This might be inevitable or just more practical since we only have this one case but let me suggest two different options where we can keep this universal approach. 

1) We can add one more method to `IDiffableByIdentifier` like:
```csharp
public interface IDiffableByIdentifier<T> : IEquatable<T>
    {
        long Identifier { get; }
        boolean IsIdentifiable { get; }
    }
```
and then in our `TypeIdentifierEqualityDiffingStrategy` we can have:
```csharp
public bool AreItemsTheSame(T item, T other)
        {
            var itemDiffable = (IDiffable<KT>) item;
            var otherDiffable = (IDiffable<KT>) other;

            if (itemDiffable.Identity.GetType() != otherDiffable.Identity.GetType())
            {
                return false;
            }

            if (!item.isIdentifiable)
            {
                item.Identifier.Equals(other.Identifier)
            }

            return item.Identifier == other.Identifier;
        }
```

2) Add simple nullable parameter to `TypeIdentifierEqualityDiffingStrategy` constructor where client can provide `function Func<T, boolean> isIdentifiable` and then simply: 
```csharp
public bool AreItemsTheSame(T item, T other)
        {
            var itemDiffable = (IDiffable<KT>) item;
            var otherDiffable = (IDiffable<KT>) other;

            if (itemDiffable.Identity.GetType() != otherDiffable.Identity.GetType())
            {
                return false;
            }

            if (isIdentifiable(item))
            {
                item.Identifier.Equals(other.Identifier)
            }

            return item.Identifier == other.Identifier;
        }
```

Does any of this make sense to you?

kornelijepetak: Both make sense, but here's my take on them 😄 

1. This would be ok if the #1 goal is to make this thing universal. But the problem is the semantics: then you can have a `IDiffableByIdentifier` that is... well... NOT diffable by identifier. If you get my point. In ideal scenario, those keys that are NOT diffable by the identifier should not even implement that interface in the first place. (this is the main pain point I have with current implementation). The easy solution would be the change of the name, but as you already know, naming is hard. 😄 

2. Works as well, but is clunky because then you need a particular method that will return `true/false` depending on the type. Performance-wise this does not differ much from checking the exact type in the `AreItemsTheSame`, you are checking the types in both cases (which unfortunatelly you will have to do anyway). The only reason why I don't like it is that then the information about whether the type is diffable by identifier is far from the type itself, breaking encapsulation a bit.

Honestly, I see no particular problem with having a custom diffing strategy for this particular use-case. I mean, that's why it's a strategy, so you can implement a different strategy when you need a different behavior. And I doubt we'll ever have that many lists of so many different kinds of elements to make the code-reuse that useful. And I'd rather abstract it then, when such a need arises, than overengineer it now.

semanticer: Thanks, you made me feel good about the non-abstract solution.

Toggl.Core.UI/ViewModels/TimeEntriesLog/Identity/TimeEntriesGroupKey.cs
```diff
@@ -10,7 +12,7 @@ public TimeEntriesGroupKey(GroupId groupId)
         }
 
         public long Identifier()
-            => groupId?.GetHashCode() ?? 0;
+            => throw new InvalidOperationException("TimeEntriesGroupKey should be always identified only by using Equals");
```
kornelijepetak: Brave lad 😄

Toggl.Droid/Adapters/MainLogRecyclerAdapter.cs
```diff
@@ -140,5 +143,42 @@ public void DeleteTimeEntry(int position)
...
+            {
+                return item.Equals(other);
+            }
+
+            public bool AreItemsTheSame(MainLogItemViewModel item, MainLogItemViewModel other)
+            {
+                var itemDiffable = (IDiffable<IMainLogKey>) item;
+                var otherDiffable = (IDiffable<IMainLogKey>) other;
+
+                if (itemDiffable.Identity.GetType() != otherDiffable.Identity.GetType())
```
kornelijepetak: In lines `171-174` you have covered the only case where a single class covers multiple cell types (TEs and TE Groups). This is then simplifiable to:

```cs
if (itemDiffable.GetType() != otherDiffable.GetType())
```

kornelijepetak: Despite the approval, consider changing this as per comment, @semanticer

semanticer: Theoretically, if we get rid of the TEs vs TE Groups dilemma faster by keeping `itemDiffable.Identity.GetType() != otherDiffable.Identity.GetType()` we could save us a little time by not running the Equals for these cases right? Do you thinks it's worth it?

Toggl.Droid/Fragments/MainFragment.cs
```diff
@@ -136,34 +135,37 @@ public override void OnViewCreated(View view, Bundle savedInstanceState)
                 .Subscribe(onSyncChanged)
                 .DisposedBy(DisposeBag);
 
-            mainRecyclerAdapter = new MainRecyclerAdapter(Context, ViewModel.TimeService)
+            mainLogRecyclerAdapter = new MainLogRecyclerAdapter()
             {
-                SuggestionsViewModel = ViewModel.SuggestionsViewModel,
-                RatingViewModel = ViewModel.RatingViewModel,
+                UserFeedbackViewModel = ViewModel.RatingViewModel
```
heytherewill: 🔥 this property

Toggl.Droid/Adapters/MainLogRecyclerAdapter.cs
kornelijepetak: We would, yes. And it is worth it. But bear in mind that this is not obvious by reading the code. So, add a comment explaining why types of the Identity are checked at all.

semanticer:
Extracted some parts to a static extension method in order to make it testable as suggested by @simonrozsival

PR #6632, opened 2019-11-20
✨ Implement the list to show the report segments

### LINKED ISSUE #6518: Implement the visual infrastructure for reports segments
Create the ability to easily add report segments in the future.

The reports screen should basically be only a single recycler view where each segment is a different row type. The goal of this issue is to provide the infrastructure so that later on, new report segments can just simply be integrated into the system.

### **🛠 Technical remarks**
Utilize the fact that all report segments should be able to be represented by the marker interface, as described at `Technical remarks` section of #6500.

### **🏆 Outcome**
If adding new segments is trivial (and intuitive), this issue is complete.

`SMALL`

kornelijepetak:
## What's this?
Implements the adapter and the recycler view so that the segments can be shown on the reports screen.

Note that many of the things are just placeholders for the future time when specific parts of reports are being implemented.

### Relationships
Closes #6518

## Why do we want this?
Better reports.

## How is it done?
Basically, `IEquatable<IReportElement>` and the facilities of the `BaseRecyclerView<T>`. 'Nuff said.

### Side effects
I've introduced a dummy view holder that will just show a name of the underlying type. That one thing will be removed when all segments are implemented.

## Review hints
Definitely commit by commit. It's going to be very easy that way.

## :squid: Permissions
📊 guys and @heytherewill


Toggl.Core.UI/ViewModels/Reports/ReportBarChartElement.cs
```diff
@@ -18,5 +18,10 @@ private ReportBarChartElement(bool isLoading)
 
         public static ReportBarChartElement LoadingState
             => new ReportBarChartElement(true);
+
+        // TODO: Do not forget to update this method and write tests for it when the element is implemented
+        public override bool Equals(IReportElement other)
+            => base.Equals(other)
+            && other is ReportBarChartElement barChartElement;
```
heytherewill: Why not check if the types match before calling base? Why call base at all?

kornelijepetak: because of `IsLoading`.

kornelijepetak: but you are right about the order.

heytherewill: I see. Still, wouldn't it be better to check types prior to calling base?

PR #6633, opened 2019-11-20
Update widget when PN triggered sync updates the database

### LINKED ISSUE #6491: PN+Widgets: Update the widget content every 5 seconds
When push notifications are enabled, the running time entry will be updated in background, if the widget is visible when that happens, it won't update the UI until it disappears and appears again.

Unfortunately it's not possible to notify the widget from the host app that it needs to update, but there are two things we can try:
- Use key-value observing on the user defaults and observe the changes on the running time entry key and then update the content.
- If that doesn't work, use a timer and every X seconds (I'd say 5 seconds is enough), if the time entry id is different, then we update the UI. This timer needs to be started and stopped on `viewDidAppear` and `viewDidDisappear` because of the memory and execution time limits on the widget process.

simonrozsival:
## What's this?
This makes iOS widgets magical.

### Relationships
Closes #6491 

## Why do we want this?
We want more magic.

## How is it done?
KVO on `NSUserDefaults`.

### Why not another way?
This worked, so the other option proposed by @juanlaube (periodically poll data) wasn't necessary.

### Side effects
I don't think there are any side effects. The only weird thing was when I discovered this line:
```cs
public static SharedStorage Instance => new SharedStorage();
```
This doesn't seem right. Every time we use `Instance`, we create a new instance of `SharedStorage`. I think it's mostly OK, because up until now the object didn't hold any state and it was just a bunch of methods around the `NSUserDefaults`. I needed to add a reference to a disposable object, so I need to get the same instance of `SharedStorage` every time. I actually don't like the idea of a global instance, but I don't want to rewrite the iOS extensions (widgets, siri) and test if I didn't break anything.

## Review hints
Try this ad-hoc build: https://toggl.slack.com/archives/C02APMYTM/p1574256947445800 - it has PNs enabled (on staging).

## :squid: Permissions
🦑


Toggl.iOS.Shared/SharedStorage.cs
```diff
@@ -33,7 +33,7 @@ private SharedStorage()
             userDefaults = new NSUserDefaults($"group.{bundleId}.extensions", NSUserDefaultsType.SuiteName);
         }
 
-        public static SharedStorage Instance => new SharedStorage();
+        public static SharedStorage Instance { get; } = new SharedStorage();
```
juanlaube: Oh nice catch!

PR #6636, opened 2019-11-20
Fix copy of the signup terms

### LINKED ISSUE #6571: The terms of service/privacy policy dialog texts are broken
Well, this image says everything.
broken terms of service/privacy policy

I thought we had an issue for this already, but I couldn't find it.

heytherewill:
## What's this?
This fix the copy formatting so it doesn't look weird (on Android) and updates the copy so that it's i18n ready  (iOS)

### Relationships
Closes #6571 

## Why do we want this?
i18n is just around the corner! We need this to merge #6389 

## How is it done?
I'm basically making Android behave properly like iOS does. I'm making things more i18n friendly by making the links dynamic

## :squid: Permissions
Only myself, after approvals from both teams


Toggl.Core.UI/ViewModels/TermsOfServiceViewModel.cs
```diff
@@ -23,6 +26,13 @@ public TermsOfServiceViewModel(IRxActionFactory rxActionFactory, INavigationServ
 
             ViewPrivacyPolicy = rxActionFactory.FromAsync(openPrivacyPolicy);
             ViewTermsOfService = rxActionFactory.FromAsync(openTermsOfService);
+            FormattedDialogText = string.Format(
```
simonrozsival: I think we should have a runtime check that `Resources.TermsOfServiceDialogMessage` contains both `{0}` and `{1}` and that either is in the text just once and crash the app if it doesn't (so that we can fix it for the particular language asap). This is something we might overlook if we have many localizations and it might be hard to figure out what's the problem from crash logs (e.g., which language contains the problem?).

heytherewill: I think this is better as an unit test, don't you think? 
I could do one that fetches the resources for every language we support officially

Toggl.Core.Tests/ResourcesTests.cs
```diff
@@ -0,0 +1,27 @@
...
+                "Toggl.Shared.Resources",
+                typeof(Shared.Resources).Assembly);
+
+            foreach (var code in Helper.Constants.SupportedLanguageCodes)
+            {
+                var culture = new CultureInfo(code);
+
+                var dialogMessage = resourceManager.GetString("TermsOfServiceDialogMessage", culture);
+                dialogMessage.Should().Contain("{0}");
+                dialogMessage.Should().Contain("{1}");
```
arturs105: This test is missing the assertion that `{0}` and `{1}` appears only once.

Toggl.Droid/Extensions/SpannableExtensions.cs
```diff
@@ -0,0 +1,36 @@
...
+            return spannable;
+        }
+
+        private sealed class ActionClickableSpan : ClickableSpan
+        {
+            private readonly ViewAction viewAction;
+
+            public ActionClickableSpan(ViewAction viewAction)
+            {
+                this.viewAction = viewAction;
```
arturs105: A null check wouldn't hurt

daividssilverio:
With some touches in the new Japanese copy, both copies are looking great!

image

image

Issue #6616, opened 2019-11-19
January 2020 Campaign Popup Plumbing and VM


amulware:
## 🎯 What needs to be done?
We want to add a relatively simple popup for our upcoming winter discount campaign.

This issue deals with the shared remote config, view model and metrics.

## 📝 Detailed explanation

Showing the popup:
- The popup will be disabled by default (we will enable it using a firebase remote config parameter).
- The popup will show when the main view is shown (be that on a fresh start, activation from background, or navigation from another screen in-app), **if the user has installed the app more than 48 hours ago** and **if the user has at least two time entries in the local database**.
- As soon as the popup is shown, we set a flag in shared storage that will prevent it from ever being shown again.
- There is no need to close the popup when the user leaves the app with it open (i.e. it should still be open when the user comes back quickly). However, if the app gets cleaned up in the background, we also don't have to restore the popup if it was open. I.e. we're keeping it as simple as possible.

The popup itself:
- There are two slightly different versions (Option A and Option B), with slightly different copy and a slightly different link. Which option to use will be part of the remote config parameter.
- The popup will have various images and/or text specified in platform-specific follow-up issues.
- The popup will have a call-to-action button which, when tapped, will close the popup and open a URL (specified in platform-specific issues) in the browser (NOT in the in-app browser!) 
- The popup will have a dismiss button that closes it.

Events:
- A. One when the popup is shown (with a parameter indicating what campaign this is)
- B. One when the call-to-action is hit (with a parameter indicating what campaign this is)
- C. One when the dismiss button is hit (with a parameter indicating what campaign this is)

Note that if correctly implemented, B. and C. are mutually exclusive, and B+C<=A (< because some users will close the app and lose the popup that way).

## 🎨 Graphical resources
Provided in follow up platform-specific issues.

## 🛠 Technical remarks
The firebase remote config flag will be called `January2020CampaignOption`. It defaults to `none` and can be `A` or `B` which will cause the correct option of the popup to be shown.

The shared-storage parameter will be called "ShowedJanuary2020Campaign" (I'm not sure about the canonical formatting here, feel free to change that if camel or snake case are most customary). It defaults to `false` and is set to `true` when the popup is shown. 

The new events are called:
- A. `MarketingMessageShown` with property `Campaign="january2020"`
- B. `MarketingMessageCallToActionHit` with property `Campaign="january2020"`
- C. `MarketingMessageDismissed` with property `Campaign="january2020"`

[...]

PR #6638, opened 2019-11-20
January campaign: Add storage methods

simonrozsival:
## What's this?
This PR adds shared storage for the campaign popup - Has the popup already been shown?

### Relationships
Ref #6616 

## :squid: Permissions
🦑


heytherewill:
One tiny Q though: Can't we push something to firebase to say "Hey, we're done here" and from there onwards we start receiving `AlreadyShown` instead of `A`, `B` or `None`? 🤔

amulware:
We can just switch back to `None`. What's the benefit of making it more complicated?

simonrozsival:
You know what, I'll create a feature branch, so it's one commit in the end and so it's easy to revert it and get rid of it. I'll update the base branch of this PR and I'll merge it afterwards.

PR #6640, opened 2019-11-20
January campaign: Shared code + iOS UI + Android UI

simonrozsival:
## What's this?
This is a feature branch. It was originally meant to contain just the shared code (VM), but since the UI is almost trivial, I decided that we'll simply merge everything into this branch and then we'll SQUASH this branch into `develop`. This will allow us to roll back these changes in the following versions, once this is no longer needed.

### Relationships
Closes #6616, Closes #6634, Closes #6635 

## Review hints
Wait until all of the subparts are merged:
- [x] Shared code is finished and merged into the feature branch
- [x] iOS UI is finished and merged into the feature branch
- [x] Android UI is finished and merged into the feature branch

## :squid: Permissions
@daividssilverio - once you're done testing and you approve this, please 🦑 it!

PR #6641, opened 2019-11-20
January campaign: Analytic events

simonrozsival:
### Relationships
Ref #6616 

## Review hints
Please check if I copy-pasted everything from Notion correctly - we've had bugs caused by this in the recent months - https://github.com/toggl/mobileapp/issues/6616

## :squid: Permissions
🦑


simonrozsival:
I decided this is too simple and it doesn't have to wait to be reviewed at this point 😄 It will be reviewed later anyway in #6640

PR #6643, opened 2019-11-20
January campaign: View Model

simonrozsival:
## What's this?
This PR adds a VM which should be used for the January 2020 Campaign popup with a dismiss and CTA button.

### Relationships
Ref #6616, requires #6637 to be merged first (to correctly set the URL).

## How is it done?
VM + tests. The app does not navigate to this VM just yet, because it doesn't have any UI on either platform. This logic will be added later in a separate PR.

### Side effects
No side effects.

## :squid: Permissions
~🚫 this is still WIP~

The PR is now ready to be reviewed and 🦑 ASAP - do it!


Toggl.Core.UI/ViewModels/January2020CampaignViewModel.cs
```diff
@@ -0,0 +1,92 @@
...
+        }
+
+        public override async Task Initialize()
+        {
+            await base.Initialize();
+
+            if (remoteConfigService.GetJanuary2020CampaignConfiguration().Option == None
+                || onboardingStorage.WasJanuary2020CampaignShown())
+            {
+                // todo: test if this works OK - can we safely close the view from within Initialize?
```
daividssilverio: It's a bit weird... I'm assuming this will cause the dialogs/modals that will display the campaign to still be displayed even though the `ViewModel` was already closed before they had a chance to appear. It's a weird state, shouldn't this be controlled outside of the viewModel? Shouldn't the consumer check for the `WasJanuary2020CampaignShown` flag before navigating to this VM?

daividssilverio: Maybe simple prefer to do this check in one of these two callbacks: `ViewAppeared()` or `ViewAppearing`

simonrozsival: The caller will definitely check that this is enabled or not. I wanted to be sure 😄 Maybe it's an overkill though, I'll get rid of it.

daividssilverio: Ok, with the last commit, the caller will also check for campaign remote config being set to `None` as well, right? Just making sure, because `None` is the default value.

Toggl.Core.UI/ViewModels/January2020CampaignViewModel.cs
```diff
@@ -0,0 +1,90 @@
...
+            var os = platform == Platform.Giskard ? "android" : "ios";
+
+            return $"https://toggl.com/winter-is-here/?utm_source=mobile-{os}&utm_medium=popup&utm_campaign=winter-is-here&utm_content=copy-{group}";
+        }
+
+        public override async Task Initialize()
+        {
+            await base.Initialize();
+
+            onboardingStorage.SetJanuary2020CampaignWasShown();
```
daividssilverio: Maybe it makes more sense to add this to `ViewAppeared`, because it might happen that the VM is initialized but not shown.

daividssilverio: Just gotta make sure how you want to track `analyticsService.MarketingMessageShown.Track(campaignName);`, since `ViewAppeared` will be called again if the app goes into background and back to foreground without closing the modals.

Issue #6519, opened 2019-11-11
Implement the visual infrastructure for reports segments


kornelijepetak:
### **🎯 What needs to be done?**
Create the ability to easily add report segments in the future.

### **📝 Detailed explanation**
The reports screen should basically be only a single table view where each row is a different cell type. The goal of this issue is to provide the infrastructure so that later on, new report segments can just simply be integrated into the system.

### **🛠 Technical remarks**
Utilize the fact that all report segments should be able to be represented by the marker interface, as described at `Technical remarks` section of #6500.

### **🏆 Outcome**
If adding new segments is trivial (and intuitive), this issue is complete.

### **👕 Estimated complexity**
`SMALL`

PR #6644, opened 2019-11-20
Reports: A new reports view controller

### LINKED ISSUE #6507: Rewrite the reports ViewController to work with the new VM
After the view model is ready, utilize it in the view controller.

Of course, it's obvious that it's not just exchanging classes because there are obvious discrepancies between how the things are done in this improvement. Do not just mess with the old one, create a new empty view controller and start from scratch. 

Then, as each particular issues will be solved that each deals with a particular report segment, it is the duty of that PRs to also integrate themselves into this (currently empty) view controller.

### **🏆 Outcome**
* The reports screen can be navigated to, and be closed. 
* Also, the workspace can be changed. 
* **No other behavior** or reports segments are expected in this issue.

`MEDIUM`

dshr:
## What's this?
A brand-new reports view controller.

### Relationships
Closes #6507 and #6519 

## Why do we want this?
For a better, more sensible and cohesive reports view.

## How is it done?
Everything is fed into one giant UICollection view with custom layout, which them automagically lays out our items for us.

### Why not another way?
Just look at how nice this animates:
reports-rotation-animation

## Review hints
Commit by commit pls

## :squid: Permissions
Reports folks, it's all yours


Toggl.iOS/ViewControllers/ReportsViewController.cs
```diff
@@ -1,25 +1,96 @@
...
+        }
+
+        public override void TraitCollectionDidChange(UITraitCollection previousTraitCollection)
         {
+            if (TraitCollection.HorizontalSizeClass == UIUserInterfaceSizeClass.Regular)
+                CollectionView.SetCollectionViewLayout(regularLayout, false);
+            else
+                CollectionView.SetCollectionViewLayout(compactLayout, false);
+            CollectionView.CollectionViewLayout.InvalidateLayout();
+            base.TraitCollectionDidChange(previousTraitCollection);
```
pdrj: Shouldn't this be called first? [Docs](https://developer.apple.com/documentation/uikit/uitraitenvironment/1623516-traitcollectiondidchange).

Toggl.Core.UI/ViewModels/Reports/ReportsViewModel.cs
```diff
@@ -55,6 +56,9 @@ public sealed class ReportsViewModel : ViewModel
                 .DistinctUntilChanged()
                 .AsDriver(schedulerProvider);
 
+            // TODO: Get the current WS name instead of this placeholder
```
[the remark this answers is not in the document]
kornelijepetak: Yes, `ReportsViewModel` should not emit specific elements per se, but always thru the `Elements` property. Please implement the workspace name as a `IReportElement` cell, just as all other parts of the report view will happen. Also, remove that from the design in xib, so that basically the only thing in the view is a collection view.

Toggl.iOS/ViewSources/ReportsCollectionViewSource.cs
```diff
@@ -0,0 +1,77 @@
...
+
+namespace Toggl.iOS.ViewSources
+{
+    public class ReportsCollectionViewSource : UICollectionViewSource
+    {
+        private readonly UICollectionView collectionView;
+
+        private ImmutableList<IReportElement> elements;
+
+        private const string summaryCellIdentifier = "summaryCellIdentifier";
```
kornelijepetak: please use `nameof(summaryCellIdentifier)` instead of hard-coded strings so that it's easier to rename if need be.

Toggl.iOS/ViewSources/ReportsCollectionViewSource.cs
```diff
@@ -0,0 +1,77 @@
...
+
+        public override nint NumberOfSections(UICollectionView collectionView)
+        {
+            return 1;
+        }
+
+        public override nint GetItemsCount(UICollectionView collectionView, nint section)
+        {
+            // TODO: replace with actual elements count
+            return 33;
```
kornelijepetak: Shouldn't this get the count from `elements.Count`?

Toggl.iOS/ViewControllers/ReportsViewController.cs
```diff
@@ -1,25 +1,96 @@
...
         }
 
         public override void ViewDidLoad()
         {
             base.ViewDidLoad();
-            // Perform any additional setup after loading the view, typically from a nib.
+            prepareViews();
+
+            ViewModel.CurrentWorkspaceName
```
kornelijepetak: Should not be used. Use a separate cell for workspace name as it is emitted by `Elements`. (check the latest `project/reports-rewrite`.

Toggl.iOS/ViewSources/ReportsCollectionViewSource.cs
```diff
@@ -0,0 +1,77 @@
...
+        {
+            this.collectionView = collectionView;
+
+            collectionView.RegisterNibForCell(ReportsSummaryCollectionViewCell.Nib, summaryCellIdentifier);
+            collectionView.RegisterNibForCell(ReportsBarChartCollectionViewCell.Nib, barChartCellIdentifier);
+            collectionView.RegisterNibForCell(ReportsDonutChartCollectionViewCell.Nib, donutChartCellIdentifier);
+            collectionView.RegisterNibForCell(ReportsProjectCollectionViewCell.Nib, projectCellIdentifier);
+        }
+
+        public void SetNewElements(IEnumerable<IReportElement> elements)
```
kornelijepetak: The `Elements` emits `ImmutableList<T>`, so to avoid turning it again into another with `ToImmutableList`, just accept `ImmutableList<IReportElement>` as the parameter.

Toggl.iOS/Views/Reports/ReportsCollectionViewRegularLayout.cs
```diff
@@ -12,6 +12,7 @@ public class ReportsCollectionViewRegularLayout : UICollectionViewLayout
         private const int maxWidth = 834;
         private const int horizontalCellInset = 8;
         private const int verticalCellInset = 12;
+        private const int numberOfNonProjectItems = 3;
```
kornelijepetak: I highly dislike the hardcoded version of this. I think you should be able to calculate this from the `Elements` you get and update accordingly.

kornelijepetak: The reasoning: adding new report element should not require changes to any UI code except the new cell for that element.

kornelijepetak:
Additionally, the `switch-case` by the items order won't work.

`Elements` property will emit various number of elements at different times. In case of an error, there will be only a single element (error-state) emitted via `Elements`, the same holds for the no-data state.

You have to rely on the underlying item type to determine which cell you need to create/render.

Please adjust accordingly.

Toggl.iOS/ViewSources/ReportsCollectionViewSource.cs
```diff
@@ -0,0 +1,77 @@
...
+            this.elements = elements.ToImmutableList();
+            collectionView.ReloadData();
+        }
+
+        public override UICollectionViewCell GetCell(UICollectionView collectionView, NSIndexPath indexPath)
+        {
+            UICollectionViewCell cell;
+            switch (indexPath.Item)
+            {
+                case 0:
```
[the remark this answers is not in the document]
kornelijepetak: The order does indeed change in regular use. No-data and error states will happen as `Elements` emission with a single item. So 0 will in other cases be something else. 
Actual underlying item type needs to be consulted to determine which cell to create.

dshr:
@kornelijepetak can I have that documented somewhere? I def feel like I'm in the dark here implementation-wise.

Toggl.iOS/ViewControllers/ReportsViewController.cs
dshr: No, since that way we'd invalidate the old layout, and re-draw that, instead of the new one.

Toggl.iOS/Views/Reports/ReportsCollectionViewCompactLayout.cs
```diff
@@ -0,0 +1,98 @@
+﻿using System;
+using System.Collections.Generic;
+using CoreGraphics;
+using Foundation;
+using Toggl.iOS.Cells.Reports;
+using UIKit;
+
+namespace Toggl.iOS.Views.Reports
+{
+    public class ReportsCollectionViewCompactLayout : UICollectionViewLayout
```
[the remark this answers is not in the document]
dshr: `UICollectionViewFlowLayout` is a completely different thing altogether, not quite fitting with the designs that we aim for.

kornelijepetak:
We have internally discussed some changes.

`WorkspaceName` will be emitted by the VM directly (Android will adapt to the change), as is already implemented by this PR (disregard my comments re: this). Workspace name will no longer be emitted by the `Elements` property as an `IReportElement`.

Toggl.iOS/ViewSources/ReportsCollectionViewSource.cs
```diff
@@ -1,13 +1,24 @@
...
 using UIKit;
 
 namespace Toggl.iOS.ViewSources
 {
+    public enum ReportsCollectionViewCell
+    {
+        Summary,
+        BarChart,
+        DonutChart,
+        DonutChartLegend,
```
kornelijepetak: Since, the cell represents a single legend item, it's better if it's `DonutChartLegendItem`.

Toggl.iOS/ViewSources/ReportsCollectionViewSource.cs
```diff
@@ -27,51 +41,88 @@ public ReportsCollectionViewSource(UICollectionView collectionView)
...
+            this.elements = elements.Where(e => e.GetType().Name != nameof(ReportWorkspaceNameElement)).ToImmutableList();
             collectionView.ReloadData();
+            collectionView.CollectionViewLayout.InvalidateLayout();
         }
 
         public override UICollectionViewCell GetCell(UICollectionView collectionView, NSIndexPath indexPath)
         {
             UICollectionViewCell cell;
-            switch (indexPath.Item)
+            switch (elements[(int)indexPath.Item].GetType().Name)
```
kornelijepetak: You can use pattern matching here, see [here](https://github.com/toggl/mobileapp/blob/project/reports-rewrite/Toggl.Droid/Adapters/ReportsAdapter.cs#L33) and [here](https://github.com/toggl/mobileapp/blob/project/reports-rewrite/Toggl.Droid/Adapters/ReportsAdapter.cs#L65).

Toggl.iOS/ViewSources/ReportsCollectionViewSource.cs
```diff
@@ -27,51 +41,88 @@ public ReportsCollectionViewSource(UICollectionView collectionView)
...
         public override nint NumberOfSections(UICollectionView collectionView)
+            => 1;
+
+        public override nint GetItemsCount(UICollectionView collectionView, nint section)
+            => elements == null ? 0 : elements.Count();
+
+        public ReportsCollectionViewCell CellTypeAt(NSIndexPath indexPath)
         {
-            return 1;
+            switch (elements[(int)indexPath.Item].GetType().Name)
```
kornelijepetak: Pattern matching here as well.

Toggl.Core.UI/ViewModels/Reports/ReportsViewModel.cs
```diff
@@ -25,6 +26,8 @@ public sealed class ReportsViewModel : ViewModel
 
         private IInteractorFactory interactorFactory;
 
+        private readonly ISubject<string> currentWorkspaceNameSubject = new Subject<string>();
```
kornelijepetak: There is no need for a subject. See other comments for explanation.

Toggl.Core.UI/ViewModels/Reports/ReportsViewModel.cs
```diff
@@ -148,6 +154,8 @@ private async Task<ImmutableList<IReportElement>> reportElementsProcess(ReportFi
         {
             try
             {
+                currentWorkspaceNameSubject.OnNext(filter.Workspace.Name);
```
kornelijepetak: There is no need to fire this change every time the report is processed, only when there's a workspace change. Hence, an earlier point should be intercepted for that.

Toggl.Core.UI/ViewModels/Reports/ReportsViewModel.cs
```diff
@@ -56,8 +59,11 @@ public sealed class ReportsViewModel : ViewModel
                 .DistinctUntilChanged()
                 .AsDriver(schedulerProvider);
 
-            // TODO: Get the current WS name instead of this placeholder
-            CurrentWorkspaceName = Observable.Return("Workspace");
+            CurrentWorkspaceName = currentWorkspaceNameSubject
```
kornelijepetak: If you look a bit down from this code, you will see that there's a `workspaceSelector` observable, on line `:71`.

This is all you need, as that observable will emit whenever the workspace changes, and also on the initial workspace retrieval.

Therefore, move this code of yours below `:71` and do something like:

```cs
CurrentWorkspaceName = workspaceSelector
    .Select(ws => ws.Name)
    .StartWith("")
    .DistinctUntilChanged()
    .AsDriver("", schedulerProvider);
```
See that the `workspaceSelector` concats `SelectWorkspace`'s action's `Elements` property which emits whenever the handler for that action returns a workspace. That way, the workspace name will be set whenever there's a change in selected workspace.

(quick note, `WhereNotNull` is there because in the case the user cancels the dialog, the handler returns `null` signifying that there should be no change).

Toggl.iOS/ViewControllers/ReportsViewController.cs
```diff
@@ -48,6 +48,8 @@ public override void ViewDidLoad()
                 .DisposedBy(DisposeBag);
 
             ViewModel.HasMultipleWorkspaces
+                .CombineLatest(ViewModel.CurrentWorkspaceName,
```
kornelijepetak: I did not get this relation between the current workspace's name and whether the user has multiple namespaces.

What your condition effectively says is:
`if` the user HAS multiple workspaces `and` current workspace's name is not empty, `then` emit TRUE.

But if the user has multiple workspaces, their names are definitely not empty. Now, if your idea was to cover the situation when the retrieval of the workspaces has yet not completed and you can't fill the button with the proper name, I'd rather not, for three reasons
1. the availability of multiple workspaces and the current workspace perceived name are semantically unrelated
2. the time needed for the first emission of the name is small and you will (in 99.99% cases) not notice it
3. even if you do notice it, there's no problem in a button that shows just an icon, and then after 300 ms gets the name of the workspace.

I simply think this change is not needed.

dshr:
@kornelijepetak I requested a review from a Random iOS Dev, so we're proper sure.

Toggl.iOS/ViewSources/ReportsCollectionViewSource.cs
```diff
@@ -0,0 +1,127 @@
...
+    {
+        Summary,
+        BarChart,
+        DonutChart,
+        DonutChartLegendItem,
+        Error,
+        NoData
+    }
+
+    public class ReportsCollectionViewSource : UICollectionViewSource
```
Odrakir: This is just a suggestion, but you could inherit from `ReloadCollectionViewSource`

dshr: I'll do that in a separate cleanup PR, we need this as asap as possible 😅

PR #6645, opened 2019-11-20
January campaign: Navigate to the popup

### LINKED ISSUE #6616: January 2020 Campaign Popup Plumbing and VM

simonrozsival:
## What's this?
This PR implements navigation to the Campaign VM if some conditions are met.

### Relationships
Ref #6616, Requires #6643 

## How is it done?
Most of the code changes is just caused by injecting dependencies into the `MainViewModel`.

The `shouldShowJanuary2020Campaign` mirrors the business logic described in the issue. I had some follow-up issues and so this logic might not be final.

## Review hints
Check the specs in the issue + my comments in the issue thread.

## :squid: Permissions
🦑


Toggl.Core.UI/ViewModels/MainViewModel.cs
```diff
@@ -391,6 +402,22 @@ private async Task handleNoDefaultWorkspaceState()
             }
         }
 
+        private async Task<bool> shouldShowJanuary2020Campaign()
+        {
+            var isDisabled = remoteConfigService.GetJanuary2020CampaignConfiguration().Option == January2020CampaignConfiguration.AvailableOption.None;
+            if (isDisabled) return false;
```
daividssilverio: No early returns on single lines if the condition has a return value.

Toggl.Core.UI/ViewModels/MainViewModel.cs
```diff
@@ -391,6 +402,22 @@ private async Task handleNoDefaultWorkspaceState()
...
+
+            var wasAlreadyShown = OnboardingStorage.WasJanuary2020CampaignShown();
+            if (wasAlreadyShown) return false;
```
daividssilverio: same here.

Toggl.Core.UI/ViewModels/MainViewModel.cs
```diff
@@ -391,6 +402,22 @@ private async Task handleNoDefaultWorkspaceState()
...
+
+            var isNewUser =
```
daividssilverio: This chain of conditions does not define a new user, right? Only an user that has logged in more than 48 hours ago and has at last one time entry.

simonrozsival: I'll change it to some better name.

daividssilverio:
Dismissing my request for changes as they were met.

simonrozsival:
I added unit tests. They are ugly, but they test exactly what they should. I don't want to clean up the code, because this code will be removed in January anyway. Let's get this over with.

PR #6646, opened 2019-11-21
Prevent a remote notifications related startup crash

arturs105:
## What's this?
I tried debugging the app on my iPhone and got a `NullReferenceException` here. No idea why or how the thing is null, but this PR fixes the crash.

### Relationships
`null`

## Why do we want this?
No startup crashes.

## :squid: Permissions
🦑

### DIFF
--- a/Toggl.iOS/Startup/AppDelegate.PushNotifications.cs
+++ b/Toggl.iOS/Startup/AppDelegate.PushNotifications.cs
@@ -13,6 +13,7 @@ namespace Toggl.iOS
     {
         public override void RegisteredForRemoteNotifications(UIApplication application, NSData deviceToken)
         {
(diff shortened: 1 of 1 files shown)


heytherewill:
I'd like @simonrozsival to evaluate what are the possible implication of dropping that `deviceToken`

simonrozsival:
I think we should rather use `#if DEBUG` here, that would be safer. There are a few places in `develop` which are problematic and we should address all of them.

arturs105:
Do you mean something like this?
```
#if DEBUG
    return;
#else
    Messaging.SharedInstance.ApnsToken = deviceToken;
#endif
```

arturs105:
A better solution from Simon incoming.

Issue #4732, opened 2019-03-19
Enable Nullable Reference Types


heytherewill:
This works in Visual Studio 2019 preview, so we can already start working on it.

We need to add the following to our `.csproj`s:

```xml
<LangVersion>8.0</LangVersion>
<NullableContextOptions>enable</NullableContextOptions>
<WarningsAsErrors>CS8600;CS8602;CS8603</WarningsAsErrors>
```

After that, we just have to whack-a-mole our way into making the project build again. 

How hard can it be?®

kornelijepetak:
> How hard can it be?®
Very hard. 😛

PR #6652, opened 2019-11-21
Migrate to C# 8

### LINKED ISSUE #4893: Replace our implementation of HashCode with Microsoft's
So apparently Microsoft had [the same idea as us](https://github.com/dotnet/corefx/blob/master/src/Common/src/CoreLib/System/HashCode.cs) of naming a HashCode combining class `HashCode`, so there is no reason to keep ours if there is a version of it included in the base library

heytherewill:
## What's this?
This migrates all of our projects to use C# 8!

### Relationships
Collaterally closes #4893 in order to make the app build
Unblocks #4732

## Why do we want this?
We wanna be able to use C# 8 features, the best of which is nullable reference types!

## How is it done?
We need to [migrate projects to netstandard2.1/netcore3.0](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version) for all of our libraries. For the xamarin projects, we just change the language target to 8.

### Side effects
Probably none, since C# doesn't have breaking changes, but it won't hurt to play with the app a bit

## Tests
[Integration tests](https://app.bitrise.io/build/5b1d29361b8be359#?tab=log)
[Syncing test](https://app.bitrise.io/build/1ef5ed220aa5b600#?tab=log)

## :squid: Permissions
Only myself


amulware:
What is my review needed for?

Issue #6635, opened 2019-11-20
January 2020 Campaign Popup Android


amulware:
## 🎯 What needs to be done?
Implement the Android UI for the January 2020 Campaign Popup.

All logic is given in the shared issue https://github.com/toggl/mobileapp/issues/6616 

## 📝 Detailed explanation
The popup is a relatively simple model with an image, some text, and two buttons: One to open a URL in the browser (NOT the in-app browser) and one to dismiss the popup.

There are two options for the popup for A/B testing. Which one is shown is controlled by remote config. Each option has a slightly different copy, and the call-to-action button opens a different link.

All details and designs can be found at https://www.notion.so/toggl/Mobile-in-January-Campaign-89776753de184c6cbafc3f4da90a2380

## 🎨 Graphical resources
Option A: https://app.zeplin.io/project/5b9f4b52a7b6df68f4c212dc/screen/5dd53b5869edd745b015f519
Option B: https://app.zeplin.io/project/5b9f4b52a7b6df68f4c212dc/screen/5dd53b5912ea124028b7b0e5

## 🛠 Technical remarks
Be very careful to get the copy and call-to-action URL for both options correct.

Copy Option A
> Upgrade to any annual plan and pocket some extra cash just in time for the holiday season!

Copy Option B
> Psst... we've got an early gift for you! Sign up for any annual plan between now and January 17th and get 20% off.

Link Option A
`https://toggl.com/winter-is-here/?utm_source=mobile-android&utm_medium=popup&utm_campaign=winter-is-here&utm_content=copy-a`

Link Option B
`https://toggl.com/winter-is-here/?utm_source=mobile-android&utm_medium=popup&utm_campaign=winter-is-here&utm_content=copy-b`

## 🔗 Dependencies
Shared code issue: https://github.com/toggl/mobileapp/issues/6616 

## 🏆 Outcome

- [x] Both options carefully implemented
- [x] Test that the popup is shown on the right conditions only
- [x] Test that the popup is never shown more than once
- [x] Double check that the call to action opens in the actual browser, not the in-app browser
- [x] Test option A using actual remote config in an AdHoc build, and make sure it opens the correct link
- [x] Test option B using actual remote config in an AdHoc build, and make sure it opens the correct link

PR #6654, opened 2019-11-21
January campaign: A/B campaign message

simonrozsival:
## What's this?
This PR exposes a property with the correct text message for both Android and iOS based on A or B group.

### Relationships
Ref #6634, ref #6635 

## Why do we want this?
We want to make shared parts of the logic in the shared layer.

## How is it done?
Ternary operator. I don't check if the option is `None`, it doesn't seem necessary. It would default to the `B` message, but it won't happen anyway.

### Why not another way?
Let's not overengineer this (what would I unit test?).

### Side effects
🦑 

## Tests

Integration tests:  - do not merge this branch unless these tests are passing.

Syncing tests:  - do not merge this branch unless these tests are passing.


Toggl.Core.UI/ViewModels/January2020CampaignViewModel.cs
```diff
@@ -25,6 +25,11 @@ public sealed class January2020CampaignViewModel : ViewModel
         public ViewAction Dismiss { get; }
         public ViewAction OpenBrowser { get; }
 
+        public string CampaignMessage
+            => remoteConfigService.GetJanuary2020CampaignConfiguration().Option == A
+                ? "Upgrade to any annual plan and pocket some extra cash just in time for the holiday season!"
```
daividssilverio: 🤔 I'm thinking here... Should this be translated to Japanese? Maybe a question for @maritoggl or @amulware Aleks, not sure.

maritoggl:
By default, we should add all new strings to resources and create issues for localizing them to languages we support. Whether we actually do translate them or not can be decided separately (sometimes we might not need to), but I don’t think we should have any strings in app that couldn’t be found from the resources file.

PR #6669, opened 2019-11-22
✨ Report Donut Chart Element

### LINKED ISSUE #6506: Create a DonutChartViewModel
Create a view model that governs the donut chart.

Basically it goes hand in hand with the DonutChart UI control.

### **🛠 Technical remarks**
Make sure it is as generic as possible and work with any type of data that has values.

### **🏆 Outcome**
There is a view model and the tests that prove that the view model behaves correctly.

`MEDIUM`

kornelijepetak:
## What's this?
Describes the proposed solution to ensure that

* the UI controls (such as the actual Donut Chart View) do not depend on any domain entity
* we are able to utilize the donut chart for other entity types (such as tags, clients, etc.)
* we can unit test the behavior of the calculations (which would be unlikely were this code inside the particular UI control.
* we can share the (testable) code and make only primitives available to the UI controls (which is the only thing they need anyway).

### Relationships
Almost closes #6506 (still needs tests)

## Why do we want this?
I have explained this under `What's this.`

## How is it done?
The `Report [Donut Chart] Element` is the composite element that has
* `Report [Donut Chart] [Donut] Element` and 
* a collection of `Report [Donut Chart] [Legend Item] Element`s

_I have used the brackets to emphasize the naming scheme._

These classes are domain agnostic. They operate in terms of:
* `ReportDonutChartElement.Segment` class (that is the basic information for each chart segment), which in turn works only with primitives
* `IReportElement` interface for the donut part and the list of legend items parts.

Now for each domain entity, we inherit from these three classes to get the following:
* `Report [Entity] [Donut Chart] Element`
* `Report [Entity] [Donut Chart] [Donut] Element`
* `Report [Entity] [Donut Chart] [Legend Item] Element`

where `[Entity]` is a `Project`, `Tag` or something similar.

> Considering that the legend items for multiple domain entities could be basically the same, it makes sense to maybe merge those into a functionally-named (as opposed to entity-named) classes, such as `Report [Percentage] [Donut Chart] [Legend Item] Element`.

### Why not another way?
Because of the reasons described under `What's this`.

## Review hints
Please think about it and give comments.

## :squid: Permissions
Me.


kornelijepetak:
Since the Project entity does not require any particulars in comparison to the default donut visuals, the donut component default is used.

Toggl.Core.UI/ViewModels/Reports/ReportDonutChartDonutElement.cs
```diff
@@ -1,25 +1,82 @@
...
+                }
+            }
+
+            return segmentsWithOther.ToImmutableList();
+        }
+
         public static ReportDonutChartDonutElement LoadingState
             => new ReportDonutChartDonutElement(true);
 
-        // TODO: Do not forget to update this method and write tests for it when the element is implemented
```
pdrj: Since you're removing this comment, should we compare `Segments`?

Toggl.Core.UI/ViewModels/Reports/ReportDonutChartElement.cs
```diff
@@ -36,11 +58,72 @@ private ReportDonutChartElement(bool isLoading)
...
+        }
+
         // TODO: Do not forget to update this method and write tests for it when the element is implemented
         public override bool Equals(IReportElement other)
             => other is ReportDonutChartElement donutChartElement
             && donutChartElement.IsLoading == IsLoading
             && donutChartElement.donutElement.Equals(donutElement)
             && donutChartElement.legend.SequenceEqual(legend);
+
+        public struct Segment
```
pdrj: Shouldn't this `struct` live outside of this file and even out of `Reports` directory? Isn't it the domain-agnostic element that a donut can just display?

Toggl.Core.UI/ViewModels/Reports/ReportProjectsDonutChartElement.cs
```diff
@@ -0,0 +1,27 @@
...
+namespace Toggl.Core.UI.ViewModels.Reports
+{
+    public class ReportProjectsDonutChartElement : ReportDonutChartElement
+    {
+        public ReportProjectsDonutChartElement(ProjectSummaryReport summary, DurationFormat durationFormat)
+            : base(convertToSegments(summary), null, segments => createLegendItems(segments, durationFormat))
+        {
+        }
+
+        private static IEnumerable<Segment> convertToSegments(ProjectSummaryReport summary)
```
pdrj: How about making an abstract property in `base` that would provide the `Segments` as well as `Legend items`?

Toggl.Core.UI/ViewModels/Reports/ReportDonutChartElement.cs
kornelijepetak: It is, but the fact is that it's only basically needed in Report's donut, I'd rather keep it there. Especially with the fact that `NormalizedValue` is used in donut exclusively. I believe it is part of the donut.

Toggl.Core.UI/ViewModels/Reports/ReportProjectsDonutChartElement.cs
kornelijepetak: That is not safe for abstract or virtual members

The order of execution is
* Base-class fields
* Inherited-class fields
* Base-class constructor
* Inherited-class constructor

Now, if I access abstract members inside of the base-class constructor, I can't guarantee that I'm getting the correct thing, because the inherited-class' constructor has not yet been called and initialized things properly.

PR #6670, opened 2019-11-22
Create and inject the services needed for the queries project

### LINKED ISSUE #6449: Create and inject the services needed for the queries project
We need to introduce the following types:

- IQuery<T>
- IQueryFactory
- RealmQueryFactory

and add them to our dependency container.

The exact way in which the types are to be implemented can be found in the queries project documentation, which you can find [here](https://www.notion.so/toggl/Realmoval-part-one-Queries-2870434ba3eb4c8b80e18f22f95828a6)

arturs105:
## What's this?
The groundwork for upcoming db queries refactors

### Relationships

Closes #6449

## How is it done?
More or less the way that the linked notion page from the issue describes things.

Decided not use `public delegate Realms.Realm RealmProvider();`, since named delegates always cause confusion for me. `Func<Realms.Ream>` is clearer - you don't have to look up the declaration to see, what it does.

Also had to refactor out the realm configuration, since now it's needed in several places.

## :squid: Permissions
🦑


Toggl.Storage.Realm/RealmConfigurator.cs
```diff
@@ -0,0 +1,50 @@
+using Realms;
+using System.Linq;
+
+namespace Toggl.Storage.Realm
+{
+    public sealed class RealmConfigurator
```
heytherewill: Why do we need this abstraction class? Why not simply inject the `RealmConfiguration` itself? 🤔

PR #6678, opened 2019-11-25
Add base class for query tests

### LINKED ISSUE #6455: Create Realm testing infrastructure
Depends on #6449 

We need to create the infrastructure needed to test the Realm code which we will introduce for the queries project.

I already have a working proof of concept for this that I put up together while writing the specs for the queries project, so I will just submit that as a PR and there's no need for no one else to worry about it

heytherewill:
## What's this?
This adds the boilerplate we'll need when adding query tests

### Relationships
Closes #6455 

## Why do we want this?
Our queries will need to be tested

## How is it done?
This just adds the base class that will contain everything we need

### Why not another way?
We could alternatively set these up as integration tests rather than unit tests. I, however, prefer keeping things like this and, if speed becomes an issue, we move this whole thing to its own test project.

## :squid: Permissions
Anyone, anytime


Toggl.Storage.Tests/Realm/Queries/QueryTests.cs
```diff
@@ -0,0 +1,25 @@
...
+        {
+            var databaseId = $"{Guid.NewGuid().ToString()}.realm";
+            var configuration = new InMemoryConfiguration(databaseId);
+            RealmProvider = () => Realms.Realm.GetInstance(configuration);
+            QueryFactory = new RealmQueryFactory(RealmProvider);
+        }
+
+        public virtual void Dispose()
+        {
+        }
```
arturs105: The code sample on Notion has some cleanup here 👀

heytherewill:
The code on notion had some deletion because I used disk realms in the PoC. Simon pointed out the existence of in memory realms, which are preferred for this usecase

PR #6680, opened 2019-11-25
Add bar chart elements

### LINKED ISSUE #6503: Rewrite the ReportsBarChartViewModel to adhere to the new ReportsViewModel
Make sure that the `ReportsBarChartViewModel` can be easily integrated into the new `ReportsViewModel` and then integrate it.

Currently, the `BarChartViewModel` knows a lot about our domain, since it's used for billalbe chart on reports. We should have a UI control that's independent of the specific domain data and it should just simply show the given values (with optional percentages).

That means that this VM should be split into:
* Classes that represent the data displayable on the barchart in its most generic form possible, without having any references to particular business entities. Those classes should be in shared code so that both platform's UI controls can reuse them without too much code duplication.
* `ReportsBarChartViewModel` that would get the data from `ReportsViewModel` and convert the data for display in `BarChartView/ReportsBarChartCardView ` by utilizing the above mentioned classes.

This way, this chart will be able to be reused in the future for other uses.

### **🛠 Technical remarks**
* Make sure to check whether the code in this viewmodel can be simplified any further. 
* Make sure that it is properly written in terms of the way we do things with RX
* Make sure that the tests for the view model are updated to cover the use cases.

### **🏆 Outcome**
The issue is complete when 
* the view model correctly handles the data from the `ReportsViewModel` 
* there are tests to confirm this,
* there exist some data types that abstract the data required for the UI control to display everything

`MEDIUM`

pdrj:
## What's this?
This adds elements ('view models') needed to display the barchart.

### Relationships

Closes #6503 

## Why do we want this?
To be able to display the barchart.

## How is it done?
Similar to #6669. Added elements for every part of the barchart view.

### Why not another way?
Please suggest something.

## :squid: Permissions
Reports team.


Toggl.Core.UI/ViewModels/Reports/ReportBarChartElement.cs
```diff
@@ -1,14 +1,30 @@
...
-        public ReportBarChartElement(ITimeEntriesTotals reportsTotal, ProjectSummaryReport summary)
+        private const int maximumLabeledNumberOfDays = 7;
+        private const int roundToMultiplesOf = 2;
+
+        public ImmutableList<Bar> Bars { get; } = ImmutableList<Bar>.Empty;
+        public ImmutableList<DateTimeOffset> HorizontalLegendItems { get; } = ImmutableList<DateTimeOffset>.Empty;
+
+        public ReportBarChartElement(
+            IEnumerable<Bar> bars,
+            IEnumerable<DateTimeOffset> horizontalLegendItems
```
kornelijepetak: I would actually advise moving `DateTimeOffset` into the `Bar` struct, as each bar indeed belongs to a certain moment in time. That will simplify a code a bit.

Toggl.Core.UI/ViewModels/Reports/ReportBarChartElement.cs
```diff
@@ -19,9 +35,70 @@ private ReportBarChartElement(bool isLoading)
...
             => other is ReportBarChartElement barChartElement
-            && barChartElement.IsLoading == IsLoading;
+               && barChartElement.IsLoading == IsLoading
+               && barChartElement.Bars.SequenceEqual(Bars)
+               && barChartElement.HorizontalLegendItems.SequenceEqual(HorizontalLegendItems);
+
+        public struct Bar
+        {
+            public double FilledValue { get; }
+            public double TransparentValue { get; }
```
kornelijepetak: `FilledValue` and `TransparentValue` seem to related to UI to me.
I'd maybe change the `TransparentValue` into `TotalValue`.
Couldn't think of a better name for `Filled` so let's leave it be.

PR #6629, opened 2019-11-19
Contextual menu improvements

heytherewill:
## What's this?
This makes changes to make the contextual menu better

### Relationships
Closes #6574
Closes #6576

## Why do we want this?
Better calendar experience

## How is it done?
I've increase the padding area for the close button and also piggybacked on @semanticer's code to react to inset changes. I tested changing from gesture navigation to 2/3 button navigation on Android 10 and it works great 👌 

### Why not another way?
I don't think we need to check for QApis anywhere, but I might be wrong. I'd also test that on Samsung devices, just in case 🤷‍♂ 

## Review hints
Please do a functional test on as many devices as possible!

## :squid: Permissions
Anyone, but not before @semanticer  and @daividssilverio approve

PR #6681, opened 2019-11-25
Fix settings insets

### LINKED ISSUE #6674: Drawing behind the navigation bar is obtrusive when using 2-button navigation
## ⛑️ User Support
This user had the system navigation set to use the `2-button navigation` setting.

## About the user
App version: 2.7
Phone OS: Android 10
Phone model: Pixel 2

## Expected behaviour
The device's navigation bar shouldn't be able to obstruct permanently the app's content.
When using whichever system navigation settings, the user should be able to scroll and see the content without any obstructions.

## To Reproduce
I was able to replicate this issue using an emulator accessing the Settings screen.
1. Go to the settings screen
2. Scroll all the way to the bottom of the screen
You should see this:
image

## Suggestions
This problem has been partially fixed in develop after 2.7 was released, so I guess we just have to do whatever we did in the `CalendarView` for the settings view as well.

Here's a comparison of what we have right now:

| 2.7 | develop |
| --- | --- |
| image | image |
| image | image |

heytherewill:
## What's this?
This ensures that even users who are not using the correct type of navigation still have a seamless experience

### Relationships
Closes #6674

## Why do we want this?
_I don't agree with the way you navigate but I will defend to the death your right to navigate like that_
Some french dude, probably

## How is it done?
Similar to #6629

## :squid: Permissions
Anyone, anytime, but please functional test!


Toggl.Droid/Toggl.Droid.csproj
```diff
@@ -21,7 +21,7 @@
     <EnableProguard>true</EnableProguard>
     <JavaMaximumHeapSize>2G</JavaMaximumHeapSize>
     <AndroidUseSharedRuntime>false</AndroidUseSharedRuntime>
-    <LangVersion>8.0</LangVersion>
+    <LangVersion>7.1</LangVersion>
```
kornelijepetak: 😱

heytherewill: Nice catch!!!!!

PR #6689, opened 2019-11-25
Make campaign dialog not dismissible when tapping outside

daividssilverio:
## What's this?
This makes campaign dialog not dismissible when tapping outside.
Ref #6616

## Why do we want this?
To make sure the users see the campaign, avoiding mistaps when the campaign dialog shows up.
Ref https://github.com/toggl/mobileapp/issues/6616#issuecomment-558209945
Internal conversation: /archives/C02APMYTM/p1574697864231100

## How is it done?
The change is very simple, take a look at the single line change.

## :squid: Permissions
🦑 it


daividssilverio:
> You did this a bit differently than we do this for some other views, where this property is set by the DialogFragmentPresenter.createReactiveDialog, but let's not overengineer this, we'll get rid of it soon anyway 😄

I didn't see that. Well, too late.

PR #6690, opened 2019-11-25
Add Japanese localization for the Jan/2020 campaign

daividssilverio:
## What's this?
This adds the Japanese localization for the Jan 2020 campaign

### Relationships
Ref #6616 

## Why do we want this?
To avoid having an English dialog appear in an app that's localized to Japanese.

## Localizations
I've wrote the localization myself, feel free to suggest something different.

## :squid: Permissions
🦑 it after the translations are greenlit internally


heytherewill:
> feel free to suggest something different

WEIRD FLEX BUT OKAY

simonrozsival:
We won't be translating this view into Japanese, we will instead skip the dialog completely for users with 🇯🇵 locale.

PR #6691, opened 2019-11-25
Add nullable ref types to the shared lib

heytherewill:
## What's this?
This adds nullable reference types to the Shared lib

### Relationships
Ref #4732 

## Why do we want this?
null bad

## How is it done?
The first commit enables null reference checking and then I'm making all the nullable related warnings be treated like errors.
The other commits are fixing the errors that showed up

### Why not another way?
This adds null checking to a single library instead of the whole app because I wanna verify that we are doing things The Right Way:tm:

### Side effects
None! Since this is only a compile time check, these changes simply make the app more stable <3

## :squid: Permissions
Only myself


Toggl.Shared/Either.cs
```diff
@@ -3,6 +3,8 @@
 namespace Toggl.Shared
 {
     public sealed class Either<TLeft, TRight>
+        where TLeft : notnull
```
kornelijepetak: I don't agree with this change.

`Either` should just as well be able to be used for `Either<int?, string?>`. The type `Either` does not deal with values but with types. It's `either` **this type** or **the other**. But these types should be able to be whatever.

kornelijepetak: Completely unrelated to this PR, the error messages for `Left` and `Right` properties should be switched. 😄

Toggl.Shared/Extensions/RandomExtensions.cs
```diff
@@ -11,7 +12,8 @@ public static class RandomExtensions
         private static readonly ThreadLocal<Random> random =
             new ThreadLocal<Random>(() => new Random(Interlocked.Increment(ref seed)));
 
-        public static T RandomElement<T>(this IList<T> collection)
+        [return: MaybeNull]
+        public static T RandomElement<T>(this IList<T>? collection)
```
kornelijepetak: If `T` is not marked as `where notnull`, how does `[return: MaybeNull]` change anything?
`T` can then be any type (including nullable types), so by default it can return `null`. What value does the attribute add here? I may not be seeing something.

Toggl.Shared/Extensions/RxAction/InputAction.cs
```diff
@@ -10,7 +10,7 @@ namespace Toggl.Shared.Extensions
 {
     public class InputAction<TInput> : RxAction<TInput, Unit>
     {
-        private InputAction(Func<TInput, IObservable<Unit>> workFactory, IScheduler mainScheduler, IObservable<bool> enabledIf = null)
+        private InputAction(Func<TInput, IObservable<Unit>> workFactory, IScheduler mainScheduler, IObservable<bool>? enabledIf = null)
```
kornelijepetak: I don't like this. I mean, not your code, the C#. 😄 
That means that basically for every optional parameter with a null default value we now need a `?`.
I mean, there's no other way, but I still don't have to like it 😄

kornelijepetak:
@heytherewill I should also state that it may be wise for you to do a short talk about what we plan to do with non-nullable reference types and what kind of constraints and attributes we can now use. I am sure this will be appreciated in the team before the codebase is full of small shady annotations people are not 100% sure why they are there. Even I had to go look a few times during the review to check the exact implication of some of the things. 

And also, we are now in a transitional period, slowly switching to using them, so it'd be good if you also provided some insight into when and how aggresively should we use them in the new code.

2 cents.

amulware:
Should we be performing this refactoring while we're also working on several others at the same time?

Toggl.Shared/Extensions/RandomExtensions.cs
heytherewill: The return type is not the same as T. The return type is `T?`, but one can't mark `T` with `T?`, so I need to specify that the return type is nullable and this is gross but that's what we have 🤷‍♂

Toggl.Shared/Extensions/RxAction/InputAction.cs
heytherewill: What do you not like about it? If the type is default and the default is null, then the type is nullable. If the type is nullable, then it needs to be marked with a `?`. What is there not to like about it? Same thing happens for value types. This makes nullability explicity, I really don't get what's not to like

Toggl.Shared/Either.cs
heytherewill: Agreed, this was done purely to fix things. I'll look into a different solution :)

heytherewill:
> Should we be performing this refactoring while we're also working on several others at the same time?

I think so, since it can be done per library and it literally just adds safety rather than make things more breakable

Toggl.Shared/Extensions/RandomExtensions.cs
kornelijepetak: I see. It's because the `T` is unconstrained so `T` and `T?` would mean different things. I wish MSFT introduced breaking changes instead of this bullshit syntax 🤷‍♂

heytherewill: b-but m-muh big cop

Toggl.Shared/Extensions/RandomExtensions.cs
```diff
@@ -13,15 +13,14 @@ public static class RandomExtensions
             new ThreadLocal<Random>(() => new Random(Interlocked.Increment(ref seed)));
 
         [return: MaybeNull]
-        public static T RandomElement<T>(this IList<T>? collection)
+        public static T Sample<T>(this IList<T>? list)
```
kornelijepetak: This method is not used at all.
It was supposed to be used to select a random starting color in `EditProjectViewModel`, but for some reason it was never used there.

Honestly, I'd rather leave this method's parameter as non-nullable and remove the `throw`. It doesn't make sense to take a sample out of `null` list.

kornelijepetak: It may not be part of `Toggl.Shared`, but I'd still use this method there (in the `EditProjectViewModel`.

Issue #4777, opened 2019-03-25
Reduce the number of build warnings


simonrozsival:
When we build our app using 🍰 `sh build.sh`, it gives us *284* warnings. I don't think we should ignore the warnings. Let's *get rid of all of them* 🔥 🔥 🔥

simonrozsival:
It's mostly either warnings because of obsolete properties being used (which will be resolved after Mvexit is completed), or the lack of `await` which should be fairly easy to fix.

PR #6700, opened 2019-11-26
Firebase Remote Config: Avoid using ActivateFetched which is obsolete

### LINKED ISSUE #6698: Avoid calling obsolete method of Firebase Remote Config iOS library
When investigating what's wrong with #6539, I found this interesting warning:

```cs
 Services/FetchRemoteConfigServiceIos.cs(29,21): warning CS0618: 'RemoteConfig.ActivateFetched()' is obsolete: 'Use the Activate method instead.' [/Users/vagrant/git/Toggl.iOS/Toggl.iOS.csproj]
```

We had problems with Firebase Remote Config blocking syncing for over a minute (on Android mostly) and that is why we implemented a custom wrapper around it to prevent these problems. I want to be cautious with these methods and I would like to use the library the way the library is meant to be used.

Docs: https://firebase.google.com/docs/remote-config/use-config-ios#fetch-and-activate-values

simonrozsival:
## What's this?
This is a low priority issue/PR, but it's also quite simple.

### Relationships
Closes #6698, ref #4777

## Why do we want this?
We want to use the library code the way it is meant to be used. Also we have get rid of 1 warning.

### Why not another way?
We could combine and nest `Fetch` and then `Activate` but we're not doing anything in between. Why not reduce the number of callbacks and make this simpler? We could also make the method `async void` and use async methods to make it nicer, but that would be frowned upon 😄

### Side effects

## Review hints
Try this ad hoc build: https://app.bitrise.io/build/5a19d3d747d33cb5

## :squid: Permissions
🦑


daividssilverio:
And yes. Latest `Xamarin.Firebase.Config` doesn't have the updated bindings.

PR #6701, opened 2019-11-26
January campaign: don't show it when the UI is in Japanese

simonrozsival:
## What's this?
This PR will make sure that we don't show the UI to users who will see the UI in Japanese.

### Relationships
Ref #6616 

## Why do we want this?
We agreed on this in an internal discussion.

## How is it done?
I need to check the `Thread.CurrentThread.CurrentUICulture` global variable to see if the UI language was set to 🇯🇵 or not (see how `AppStart.setLocale` works).

### Side effects
I hope this doesn't break the unit tests or make them flaky 👀 

## Localizations
No localization needed 😆 

## :squid: Permissions
🦑


Toggl.Core.Tests/UI/ViewModels/MainViewModelTests.cs
```diff
@@ -288,6 +291,42 @@ public async ThreadingTask NavigatesToJanuary2020CampaignPopup(string group)
...
+                {
+                    Start = DateTimeOffset.Now,
+                    Duration = 1,
+                    IsDeleted = false,
+                    ServerDeletedAt = null,
+                    TagIds = new long[0],
+                    Workspace = new MockWorkspace { IsInaccessible = false }
+                };
+                var twoTEs = new BehaviorSubject<IEnumerable<IThreadSafeTimeEntry>>(new[] { mockTimeEntry, mockTimeEntry });
+                Thread.CurrentThread.CurrentUICulture = new CultureInfo("ja-JP");
```
heytherewill: Are we sure that each test runs in an isolated thread? Can't this mess up some other unrelated test?

Toggl.Core.UI/ViewModels/MainViewModel.cs
```diff
@@ -404,6 +406,11 @@ private async Task handleNoDefaultWorkspaceState()
 
         private async Task<bool> shouldShowJanuary2020Campaign()
         {
+            var japaneseCultureInfo = new CultureInfo("ja-JP");
```
heytherewill: Let's maybe. instead, check for the list of all supported languages? Just in case

simonrozsival: Here is the relevant thread: https://toggl.slack.com/archives/C02APMYTM/p1574697864231100

Toggl.Core.Tests/UI/ViewModels/MainViewModelTests.cs
simonrozsival: I wasn't sure. I agree that we should figure this out before we proceed with this PR. I tried addressing this in: https://github.com/toggl/mobileapp/pull/6701/commits/5d9c9d2d22ba7289bba000e0ce5da666df4664eb

In the commit I define a collection `"global"` and tests in this group won't run in parallel. If we agree that this is a good solution for this problem, I'd add a paragraph about tests which access the global scope (static classes) and that they should be included in this group and that they should clean-up after themselves.

Toggl.Core.UI/ViewModels/MainViewModel.cs
daividssilverio: I'd rather check for the two letters language code instead of the full `CultureInfo` object.
This way we can prevent funky iOS language codes, like `ja-US`.

simonrozsival: The culture info is set correctly in `AppStart`  and for Japanese it'll always be `ja-JP`. But we can check the two letter ISO language name.

simonrozsival: I changed it to check the ISO two letter name.

daividssilverio: > The culture info is set correctly in AppStart and for Japanese it'll always be ja-JP. But we can check the two letter ISO language name.

Not when I run it on the simulator at least. When the language code is `ja-US` nothing works.
I'm actually fixing that and another PR will come soon.

simonrozsival: Ok.

PR #6702, opened 2019-11-26
💚 Fix donut grouping and add tests

### LINKED ISSUE #6506: Create a DonutChartViewModel

kornelijepetak:
## What's this?
This PR does two things:

* Fixes previously implemented algorithm to adhere to #2439
* Adds tests for the implemented report elements

### Relationships
Closes #6506 

## Why do we want this?
For some nice 🍩.

## How is it done?
Math, man! Math.

## :squid: Permissions
📊 guys only.


Toggl.Core.UI/ViewModels/Reports/ReportDonutChartElement.cs
```diff
@@ -12,19 +13,32 @@ public class ReportDonutChartElement : CompositeReportElement
         private ImmutableList<IReportElement> legend;
 
         /// <summary>
-        /// The maximum percentage of elements under the Other category.
+        /// Elements with larger than the fraction specified by this constant
+        /// are included in the chart as individual slices.
         /// </summary>
-        private readonly double groupUnderOtherThreshold = 0.1;
+        public static readonly double GuaranteeSliceThreshold = 0.05;
```
pdrj: Since you used 'Guaranteed' for the other constant, please use it here as well.

Toggl.Core.UI/ViewModels/Reports/ReportDonutChartElement.cs
```diff
@@ -12,19 +13,32 @@ public class ReportDonutChartElement : CompositeReportElement
...
+
+        /// <summary>
+        /// Group into 'Other' all elements that would fill less of
+        /// a chart fraction than this constant specifies
+        /// </summary>
+        public static readonly double GuaranteedForOtherThreshold = 0.01;
+
+        /// <summary>
+        /// The maximum base size of the 'Other' item, up to which to add items
+        /// that are between 1% and 5% of the total chart.
```
pdrj: It's probably because I'm already spent today, but it took me about 10 minutes to get the meaning of this 😄.

Toggl.Core.UI/ViewModels/Reports/ReportDonutChartElement.cs
```diff
@@ -66,33 +83,48 @@ public static ReportDonutChartElement LoadingState
...
+            }
+
+            if (other.Count == 1)
+            {
+                slices.Add(other.Single());
+                other.Clear();
+            }
+
+            // Order by descending
+            slices.Sort((a, b) => b.Value.CompareTo(a.Value));
```
pdrj: Microoptimization, but wouldn't it be sufficient to just reverse the slices?

Toggl.Core.UI/ViewModels/Reports/ReportDonutChartElement.cs
```diff
@@ -66,33 +83,48 @@ public static ReportDonutChartElement LoadingState
...
-            var total = segments.Sum(s => s.Value);
+            var slices = new List<Segment>();
+            var other = new List<Segment>();
+            var otherPercentage = 0.0;
 
-            var normalizedSegments = segments
-                .Select(s => s.Normalized(s.Value / total))
-                .OrderByDescending(s => s.NormalizedValue)
-                .ToList();
+            var total = segments.Sum(s => s.Value);
```
pdrj: Just want to make sure - there will never be a case where that sum is 0? In our case segments will be summed time entries, right? Is it possible to create ones with 0 duration?

kornelijepetak: Probably not, but I still created this one case that will be handled in the UI control.

Toggl.Core.UI/ViewModels/Reports/ReportDonutChartElement.cs
kornelijepetak: True. Nice catch.

Toggl.Core.UI/ViewModels/Reports/ReportDonutChartElement.cs
kornelijepetak: I have updated the comment to better explain the purpose.

PR #6704, opened 2019-11-26
Possible fix for a crash in the calendar

### LINKED ISSUE #5869: CalendarCollectionViewSource.updateCalendarItem index out of range
## 💥 Crash Report
🤷‍♂ Check App Center for all the info on the crash.

### 👥 Impact
This crash has occurred 136 times to 54 users in the last 30 days, which translates to 0.1% affected users

### 🔗 Context
We've had index out of range crashes in `CalendarCollectionViewSource.updateCalendarItem` (#4886) and seems like it was fixed, however there are new crashes in this method in 2.2.1, so we need to do something about this.

### 🛰 Link
https://appcenter.ms/orgs/Toggl/apps/Toggl-iOS/crashes/errors/775237137u/overview

`SIGABRT: Index was out of range. Must be non-negative and less than the size of the collection. Parameter name: index`
```cs
List`1[T].get_Item (System.Int32 index) in <939d99b14d934342858948926287beba#01b7521b705f92edab0b11232d0eaaea>:0
CalendarCollectionViewSource.updateCalendarItem (Foundation.NSIndexPath indexPath, System.DateTimeOffset startTime, System.Nullable`1[T] duration)
CalendarCollectionViewSource.UpdateItemView (System.DateTimeOffset startTime, System.TimeSpan duration)
CalendarCollectionViewCreateFromSpanHelper.longPressChanged (CoreGraphics.CGPoint point)
CalendarCollectionViewCreateFromSpanHelper.onLongPress (UIKit.UILongPressGestureRecognizer gesture)
UILongPressGestureRecognizer+Callback.Activated (UIKit.UILongPressGestureRecognizer sender)
```

juanlaube:
## What's this?
A possible fix for a crash in the calendar view. While I couldn't reproduce the crash in any way, I think this happens in certain situations where the user is editing a time entry while syncing happens at the same time.

### Relationships

Closes #5869 

## Why do we want this?
Because we don't like crashes.

## How is it done?
Checking that the index path is in the range of the calendar items count.

### Why not another way?
🤷‍♂ 

### Side effects
None.

## Review hints
While I couldn't reproduce the crash, I added this piece of code in the calendar view controller to force a pull sync when pressing cmd+r, might be useful for testing.

```
        public override UIKeyCommand[] KeyCommands
        {
            get
            {
                return new[]
                {
                    UIKeyCommand.Create(new NSString("r"), UIKeyModifierFlags.Command, new Selector("pull")),
                };
            }
        }

        [Export("pull")]
        private void pull()
        {
            IosDependencyContainer.Instance.SyncManager.ForceFullSync().Subscribe();
        }
```

## :squid: Permissions
🦑


Toggl.iOS/ViewSources/CalendarCollectionViewSource.cs
```diff
@@ -383,6 +380,9 @@ private NSIndexPath insertCalendarItem(DateTimeOffset startTime, TimeSpan durati
 
         private NSIndexPath updateCalendarItem(NSIndexPath indexPath, DateTimeOffset startTime, TimeSpan? duration)
         {
+            if (indexPath == null || indexPath.Item >= calendarItems.Count)
+                return indexPath;
```
simonrozsival: 🤔 I don't understand this. At this point we know that `indexPath` is invalid. Why are we returning it?

Toggl.iOS/ViewSources/CalendarCollectionViewSource.cs
```diff
@@ -309,11 +309,8 @@ private void timeOfDayFormatChanged(TimeFormat timeFormat)
 
         private void onCollectionChanges()
         {
-            if (IsEditing)
```
simonrozsival: I think we might just `if (editingItemIndexPath.Item < 0 || editingItemIndexPath.Item >= calendarItems.Count) return;` and avoid the crash. What's worse?

Toggl.iOS/ViewSources/CalendarCollectionViewSource.cs
juanlaube: Oops 🙈

Toggl.iOS/ViewSources/CalendarCollectionViewSource.cs
juanlaube: I prefer to avoid the crash 👍

PR #6705, opened 2019-11-26
Make Japanese localization selection less strict

daividssilverio:
## What's this?
This is a "fix" for a problem found when testing the Jan. 2020 campaign (#6701).

### Relationships
Closes #Nothing
Ref #6701 
Ref #6389 

## Why do we want this?
Because it seems like iOS can have funky language codes that shouldn't even be possible, stuff like `ja-US`. 

On android, everything seems to work fine and when the system language is set to Japanese, the language code returned by
```cs
PlatformInfoAndroid:CurrentNativeLanguageCode => Locale.Default?.ToString()
```
is `ja-JP` alright.

But on iOS
```cs
PlatformInfoIos:CurrentNativeLanguageCode:getPreferredLanguageCode => NSLocale.PreferredLanguages.FirstOrDefault()
```
it can be something other than `ja-JP`.

## How is it done?
By ignoring what comes after the `jp`, and counting only on the two letter language code.

### Why not another way?

### Side effects
Nothing really.

## Review hints
Test develop, then test this PR's branch while setting the device language do Japanese.

## Localizations
Everything should still work.

## :squid: Permissions
🦑 it.


heytherewill:
Oh, BTW, you need to update `Toggl.Shared.csproj` to change the name of the file there to `Resources.ja.resx` :)

daividssilverio:
Hey @simonrozsival 
The `AppStartTests` already cover this. The biggest problem here I guess was the `resx` file being suffixed with `ja-JP`. Our localization framework would not select `ja-JP` (more specific localization) when we set the `CurrentCultures` to `ja`.

PR #6710, opened 2019-11-27
✅ Add tests for bar chart elements

### LINKED ISSUE #6503: Rewrite the ReportsBarChartViewModel to adhere to the new ReportsViewModel

pdrj:
## What's this?
This adds tests for recently added bar chart elements.

### Relationships

Closes #6503 

## Why do we want this?
Safety first.

## :squid: Permissions
Reports crew.


Toggl.Core.Tests/UI/ViewModels/Reports/ReportProjectsBarChartElementTests.cs
```diff
@@ -0,0 +1,139 @@
...
+            [Fact, LogIfTooSlow]
+            public void ConvertsWeeklyReportToBars()
+            {
+                var startDate = DateTimeOffset.ParseExact("200101011453 +00:00", "yyyyMMddHHmm zzz", CultureInfo.InvariantCulture);
+
+                var timeEntriesTotals = new TimeEntriesTotals();
+                timeEntriesTotals.Resolution = Resolution.Week;
+                timeEntriesTotals.StartDate = startDate;
+
+                var group1 = new TimeEntriesTotalsGroup();
```
kornelijepetak: Since you are using the same code, please reuse it and give it some meaningful name.

PR #6712, opened 2019-11-27
Remove strings that are only used in Analytics

heytherewill:
## What's this?
This prevents us from getting localized telemetry

### Relationships
Followup of #6389 

### Why not another way?
It makes no sense to localize such strings. It also makes no sense to mark them as untranslatable, since someone might mistakenly use them. We should simply use an inline string.

## Review hints
@daividssilverio is marked as a reviewer since he mentioned that calendar had a similar translated string, but I couldn't find it. If there are any other cases like this one, please speak up so we can fix it.

## :squid: Permissions
Only me


Toggl.iOS.SiriExtension.UI/Toggl.iOS.SiriExtension.UI.csproj
```diff
@@ -137,6 +136,8 @@
     <ProjectReference Include="..\Toggl.iOS.Shared\Toggl.iOS.Shared.csproj">
       <Project>{aca2821e-0ff9-4036-bff6-e775bb4e6732}</Project>
       <Name>Toggl.iOS.Shared</Name>
+      <IsAppExtension>false</IsAppExtension>
```
simonrozsival: 🔥

Toggl.iOS.SiriExtension/Toggl.iOS.SiriExtension.csproj
```diff
@@ -124,6 +124,8 @@
     <ProjectReference Include="..\Toggl.iOS.Shared\Toggl.iOS.Shared.csproj">
       <Project>{aca2821e-0ff9-4036-bff6-e775bb4e6732}</Project>
       <Name>Toggl.iOS.Shared</Name>
+      <IsAppExtension>false</IsAppExtension>
```
simonrozsival: 🔥

Toggl.iOS.TimerWidgetExtension/Toggl.iOS.TimerWidgetExtension.csproj
```diff
@@ -148,6 +143,8 @@
     <ProjectReference Include="..\Toggl.iOS.Shared\Toggl.iOS.Shared.csproj">
       <Project>{aca2821e-0ff9-4036-bff6-e775bb4e6732}</Project>
       <Name>Toggl.iOS.Shared</Name>
+      <IsAppExtension>false</IsAppExtension>
```
simonrozsival: 🔥

PR #6715, opened 2019-11-27
🍩 Add DonutChart Android UI control

### LINKED ISSUE #6504: Create a DonutChart control for Android
Create a generic Donut chart.

We are replacing the PieChart with a donut chart.

Make sure that the chart is made in a generic way, with as much things parameterized as possible because we will probably have to tweak some things during testing. Let's write it good.

### **🛠 Technical remarks**
There is probably a lot of our current PieChart that can be salvaged, or at least to speed up the implementation. Make use of it.

### **🏆 Outcome**
The control that can be used anywhere in the UI that responds to any data provided from outside.

`EXTRA LARGE`

image

kornelijepetak:
### What's this?
Implements the donut chart.
Also, fixes the elements required for projects-based donut.

### Relationships
Closes #6504 
Opens #6716 

## Why do we want this?
Because donuts are yummy.

## How is it done?
📐 Math.

## Review hints
Commit by commit, of course.

## :squid: Permissions
📊 guys only.


Toggl.Droid/Views/DonutChartView.cs
```diff
@@ -0,0 +1,99 @@
...
+            angleOffset += sliceAngle;
+        }
+
+        private void drawInnerCircle(Canvas canvas)
+        {
+            const float donutInnerCircleFactor = 0.6f;
+
+            paint = new Paint();
+            paint.AntiAlias = true;
+            paint.Color = Color.ParseColor("#FFFFFF");
```
heytherewill: Color constants are a big no no

kornelijepetak:
I did reuse the layouts. However, this is why it may be misleading:
1. The legend is a direct copy of the old one (with different name). I have not yet fired the old one as it is referenced in the old view model, which we are saving until it's time to :fire: all old things.
2. The piechart used to be inside a bunch of other elements within weird hierachy. I have literally pulled it out from there (copy/paste) and moved it into a separate donut layout (also changing the type from `PieView` into `DonutView`). But it's obvious that some of the margins/constraints are not the same in the new environment so I gotta take a look at it in the future.

Toggl.Core.UI/Helper/Colors.cs
```diff
@@ -328,6 +328,11 @@ public static class BarChart
 
             public static readonly Color OtherProjectsSegmentBackground = pinkishGrey;
 
+            public static class Donut
```
heytherewill: Use the native color assets instead, since this doesn't support dark mode

Toggl.Droid/Views/DonutChartView.cs
```diff
@@ -0,0 +1,104 @@
...
+        }
+
+        protected override void OnMeasure(int widthMeasureSpec, int heightMeasureSpec)
+        {
+            base.OnMeasure(widthMeasureSpec, widthMeasureSpec);
+        }
+
+        protected override void OnDraw(Canvas canvas)
+        {
+            const float angleRightToTopCorrection = -90;
```
daividssilverio: why define a constant inside the `OnDraw` loop?

Toggl.Droid/Views/DonutChartView.cs
```diff
@@ -0,0 +1,104 @@
...
+
+            if (data == null)
+                return;
+
+            var totalValue = data.Segments.Sum(s => s.Value);
+
+            var angleOffset = angleRightToTopCorrection;
+
+            using (var circle = new RectF(0, 0, Width, Height))
```
daividssilverio: Please refactor this to have a single `RectF` reference, create a field in the `DonutChartView`.
You can use the `rectF.Set(left, top, right, bottom)` to set the props.

Toggl.Droid/Views/DonutChartView.cs
kornelijepetak: Because it's not needed anywhere else? It's rather local and only important to the process of drawing. It's not even a constant we can change, it's just a number to transform the coordinate system to match what we want. Why would you need it anywhere else?

daividssilverio: Do you know if the compiler will move this somewhere else and this won't be instantiated a lot during the draw loops?

PR #6717, opened 2019-11-27
Let's get rid of warnings

### LINKED ISSUE #4777: Reduce the number of build warnings

simonrozsival:
## What's this?
In this PR I'm trying to get rid of as many build warnings as possible. This PR is not meant to be merged at the moment, it's here for us to discuss about which compiler warnings we should turn into errors and which we shouldn't.

### Relationships
Closes #4777

## Why do we want this?
Our apps might have some potential problems which are hidden in the sea of unnecessary warnings in our apps.

Some of these warnings are easily fixed, others can be ignored (e.g., in test projects). We should end up with build logs, which we will actually read and check, if it might actually be a problem or not.

## How is it done?
_I'll add more to this section once I'm finished with this draft._

### Side effects
This PR touches a lot of files. We might be seeing more and more red squiggles during development and we'll have to write code more consistently.

## Review hints
There's no other way to review this than commit by commit.

## :squid: Permissions
🚫 Don't even think about it.


simonrozsival:
Android build:
image

iOS build:
image

I currently disabled two warnings for `Toggl.Core.Tests`, because there are TOO MANY of them. I'll revisit this in the future.

Toggl.Shared/Math.cs
```diff
@@ -22,7 +22,9 @@ public static double ToAngle(this TimeSpan time)
 
         public static double ToPositiveAngle(this double angle)
         {
-            while (angle < 0) angle += FullCircle;
+            while (angle < 0)
```
heytherewill: This is evil, please add curly braces

Toggl.ruleset
```diff
@@ -6,9 +6,11 @@
     <Rule Id="CS0168" Action="Error" />
     <Rule Id="CS0169" Action="Error" />
     <Rule Id="CS0414" Action="Error" />
+    <Rule Id="CS0618" Action="None" />
```
heytherewill: Why were these disabled again?

simonrozsival: That's a temporary thing because `Toggl.Core.Tests` has like a million of these two warnings (the two that are set to `None`).

simonrozsival:
@heytherewill I think we can now consider going forward with this or not (it might break the reports rewrite project though, maybe we should wait until it's merged before we actually do something this radical).

Toggl.Droid/Adapters/BaseSectionedRecyclerAdapter.cs
```diff
@@ -34,10 +34,8 @@ public abstract class BaseSectionedRecyclerAdapter<TSection, TItem> : RecyclerVi
         private readonly Subject<Unit> itemsUpdateCompletedSubject = new Subject<Unit>();
         public IObservable<Unit> ItemsUpdateCompleted { get; }
 
-        private IImmutableList<SectionModel<TSection, TItem>> items;
```
kornelijepetak: I do not want you to delete this getter. While it may not be used at the moment, it is a valid and semantically sound element of the `Items` API. The ability to only set the items without having a way of getting them back makes no sense. The consumer of the adapter SHOULD be able to access the items should they want to.

Toggl.Droid/Services/SyncJobService.cs
```diff
@@ -22,7 +22,7 @@ public override bool OnStartJob(JobParameters parameters)
                 {
                     runSync(parameters);
                 }
-                catch (Exception exception)
+                catch (Exception)
```
kornelijepetak: this is the same as the catch without exception type specified.
```cs
catch
{

}
```

Toggl.ruleset
```diff
@@ -6,9 +6,11 @@
...
     <Rule Id="CS0628" Action="Error" />
     <Rule Id="CS0649" Action="Error" />
     <Rule Id="CS1998" Action="Error" />
+    <Rule Id="CS4014" Action="None" />
```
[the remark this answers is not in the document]
simonrozsival: This is temporarily disabled, because our tests are full of it and it's not something I want to fix right now. I'm mostly interested in making the app code warning-free and then I can look into the problems in the test projects. I'll definitely turn that into a warning later :)

Toggl.iOS/Resources/Images.xcassets/confirmation.imageset/Contents.json
```diff
@@ -1,85 +1,121 @@
 {
-  "images": [
+  "images" : [
```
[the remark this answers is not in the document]
simonrozsival: This file is generated by Xcode. Maybe they changed how they serialize JSONs in the new Xcode? 🤷‍♂

Toggl.Droid/Adapters/BaseSectionedRecyclerAdapter.cs
simonrozsival: We're not writing a library though. Why doesn't it make sense, if it's the way we're using it? 🤔

kornelijepetak: While language allows you to, I think set-only properties are an antipattern. If you intended to set something only, I think you are better of with a function.

Toggl.iOS/Resources/Images.xcassets/confirmation.imageset/Contents.json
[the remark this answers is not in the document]
kornelijepetak: oh, I see.

Toggl.ruleset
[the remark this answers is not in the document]
kornelijepetak: That is... well... a nice explanation. :)

Toggl.Core.Tests.Sync/Toggl.Core.Tests.Sync.csproj
```diff
@@ -4,7 +4,7 @@
     <IsPackable>false</IsPackable>
   </PropertyGroup>
   <PropertyGroup>
-    <CodeAnalysisRuleSet>$(SolutionDir)Toggl.ruleset</CodeAnalysisRuleSet>
+    <CodeAnalysisRuleSet>..\Toggl.ruleset</CodeAnalysisRuleSet>
```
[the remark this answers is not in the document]
simonrozsival: But `Cake` and `dotnet` don't seem to know what it is 🤔 What can we do about it?

heytherewill:
Not worth merging so much changes, which might bring instability

PR #6719, opened 2019-11-27
📊 Reports: new iOS report summary cell

### LINKED ISSUE #6520: Create a cell type for Report Summary
Create a cell type for Reports Summary.

This requires no user interaction. It simply shows:
* Total tracked time
* Ratio of billable time

### 🎨 Graphical resources
You can find the [designs at Zeplin](https://app.zeplin.io/project/5728ce853dd1b182475f25e9/screen/5b42fcda770d8fc76a8f1f53).

### **🛠 Technical remarks**
Utilize the view model created in #6502 

### **🏆 Outcome**
The reports now show the correct total tracked time and the appropriate billable ratio.

`SMALL`

dshr:
## What's this?
This adds the reports summary cell to our reports view.

### Relationships
Closes #6520 

## Why do we want this?
Better reports?

## How is it done?
I copied the old cell internals and populated it. No Rx stuff this time, to simplify retain cycles and all.

### Why not another way?
Seems as simple as it gets.

## Review hints
Commit by commit works best!

## :squid: Permissions
Reports folx only


Toggl.iOS/Cells/Reports/ReportsSummaryCollectionViewCell.cs
```diff
@@ -1,17 +1,47 @@
...
+            ForegroundColor = disabledColor
+        };
+
         public static readonly NSString Key = new NSString("ReportsSummaryCollectionViewCell");
         public static readonly UINib Nib;
-
         public static readonly int Height = 124;
 
+        public ReportSummaryElement element {
+            set
```
kornelijepetak: This setter does nothing. Why?

Toggl.iOS/Cells/Reports/ReportsSummaryCollectionViewCell.cs
```diff
@@ -1,17 +1,47 @@
...
+            {
+                var element = value;
+            }
+            private get { return element; }
```
kornelijepetak:
```cs
private get => element;
```

But you should also be aware that this getter is recursive. It returns itself. What's going on here? 👀

Toggl.iOS/Cells/Reports/ReportsSummaryCollectionViewCell.cs
```diff
@@ -21,6 +51,62 @@ protected ReportsSummaryCollectionViewCell(IntPtr handle) : base(handle)
...
+                    : totalTimeEnabledColor;
+            }
+        }
+
+        private NSAttributedString billableFormattedString(float? value)
+        {
+            var isDisabled = value == null;
+            var actualValue = isDisabled ? 0 : value.Value;
+
+            var percentage = $"{actualValue.ToString("0.00")}%";
```
kornelijepetak: There's a shortcut for this
`$"{actualValue:0.00}%";`

Toggl.iOS/Cells/Reports/ReportsSummaryCollectionViewCell.cs
```diff
@@ -21,6 +51,62 @@ protected ReportsSummaryCollectionViewCell(IntPtr handle) : base(handle)
...
+            TotalTimeTitleLabel.SetKerning(-0.2);
+            TotalTimeLabel.SetKerning(-0.2);
+            BillableTitleLabel.SetKerning(-0.2);
+            BillablePercentageLabel.SetKerning(-0.2);
+        }
+
+        public void SetElement(ReportSummaryElement element)
+        {
+            LoadingView.Hidden = !element.IsLoading;
+            if (element.BillablePercentage != null)
```
kornelijepetak: Since there's no `else` I wonder what happens when an RSE element comes in with the `BillablePercentage` set to some value, and then when another one comes after that with it set to `null`, there will be no change, the ui will still show the old value, won't it?

I am not quite sure in what case something like this would happen, but do you think it'd be wise to clear the text from the text views?

But I am not sure about all of that, so correct me if I got something wrong.

dshr: Will he ever have a case where we'd have `IsLoading` unset and no `BillablePercentage` value?

kornelijepetak: I assume not, but then I'd rather have the flow split on whether the `IsLoading` is set.
Then these two `if`s you have would not be needed. I find them a bit misleading because they somewhat look like there's a chance for them to be null.

But I don't know. this is ok as well. 🤷‍♂

Toggl.iOS/Cells/Reports/ReportsSummaryCollectionViewCell.cs
```diff
@@ -1,17 +1,47 @@
...
+using Toggl.Shared.Extensions;
 using UIKit;
 
 namespace Toggl.iOS.Cells.Reports
 {
     public partial class ReportsSummaryCollectionViewCell : UICollectionViewCell
     {
+        private const int fontSize = 24;
+
+        private static readonly UIColor percentageEnabledColor = Colors.Reports.PercentageActivated.ToNativeColor();
```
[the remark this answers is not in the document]
kornelijepetak: Some insight, please. 😄 How does iOS handle darkness?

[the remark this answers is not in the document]
dshr: So, there's two ways: 

1. Via `ColorAssets`: this is the way we use for colors that change when going to dark mode.
2. Using colors that work well with both light and dark bgs, like in our case:

Screenshot 2019-11-28 at 16 01 19 1
Screenshot 2019-11-28 at 16 03 09

PR #6732, opened 2019-11-28
Load calendar views lazily

### LINKED ISSUE #6541: Calendar view: optimize initial load time
Apart from the main view, which is slow due to (amongst other things 😄) loading other dependencies, calendar view is the slowest one to load.

We need to figure out how to make it at least four times as fast. 'Four times' is not arbitrary - more info in the [OKR page](https://www.notion.so/toggl/2019-Q4-OKRs-1671c3057a5f49f7b54d090c560f4192?p=7d2e61a160ec4d7c95b10faeebe14913).

To be clear, we're talking about the initial load time here. Use peter/measure branch to help with measurements.

daividssilverio:
## What's this?
This PR improves the loading time for the calendar view

### Relationships
Closes #6541

## Why do we want this?
Because nobody likes janking apps.

## How is it done?
By simply (not that simple though) using stubs for both calendar contextual menu and day views.
They were one of the main causes of the slow load times for the `CalendarFragment`.
By using stubs and only inflating the contextual menu when it's actually needed, the loading time of the calendar view improved a lot.
Using a stub for the day view didn't seem as effective to be honest... But I couldn't really tell for sure in my tests.

### Why not another way?
Feel free to suggest something else.
I'm not sure if the making the loading of the bindings themselves would improve things further, but I'll keep trying.
This already makes the loading times for the `CalendarFragment` twice as fast.

### Side effects
🤔 None that I could find.

## Review hints
Please review by commit.
Also, please check the branch out the `doivid/wip-calendar` and its two commits, two adhoc builds were provided for that branch, and they include @pdrj 's measuring utilities.
Test both builds (either using the bitrise builds or simply by building locally) 🙇 .

## :squid: Permissions
🦑 it after @pdrj review, another android review would also be nice.


Toggl.Droid/Fragments/CalendarFragment.cs
```diff
@@ -235,7 +235,8 @@ private class CalendarDayFragmentAdapter : FragmentStatePagerAdapter, ViewPager.
             private readonly ISubject<Unit> pageNeedsToBeInvalidated = new Subject<Unit>();
             private readonly ISubject<Unit> backPressSubject = new Subject<Unit>();
             public BehaviorRelay<int> OffsetRelay { get; } = new BehaviorRelay<int>(0);
-            public BehaviorRelay<int> CurrentPageRelay { get; } = new BehaviorRelay<int>(0);
+            public BehaviorRelay<int> CurrentPageRelay { get; } = new BehaviorRelay<int>(-1);
+            public Subject<int> ScrollingPageRelay { get; } = new Subject<int>();
```
semanticer: Why `ScrollingPageRelay` when it is a `Subject` type? Also, I guess there is some reason why this isn't  `ISubject`?

Toggl.Droid/Fragments/Calendar/CalendarDayViewPageFragment.cs
```diff
@@ -27,6 +28,7 @@ public partial class CalendarDayViewPageFragment : Fragment, IView
         private CalendarContextualMenuActionsAdapter menuActionsAdapter;
 
         public CalendarDayViewModel ViewModel { get; set; }
+        public Subject<int> ScrollingPageRelay { get; set; }
```
semanticer: Why `ScrollingPageRelay` when it is a `Subject` type? Also, I guess there is some reason why this isn't `ISubject`?

Toggl.Droid/Fragments/CalendarFragment.cs
daividssilverio: It's a subject because we need to fire events from it. It's not a behaviour or relay because we don't want to receive the historic values when the fragment day page subscribes to it, it would cause the day view stubs to load before they are needed.

I forgot to use the `ISubject` type, there's no reason not to use it.

Toggl.Droid/Fragments/Calendar/CalendarDayViewPageFragment.cs
daividssilverio: Mistype :) Also, no reason for not being an `ISubject`.

PR #6737, opened 2019-11-29
Fix ios11 modals position

simonrozsival:
## What's this?
This fixes weird start view positioning on iOS 11 (and maybe 12).

### Relationships
I was too lazy to create an issue, sorry.

## Why do we want this?
We want nice UI on all supported devices.

## How is it done?
The vertical offset has to account for the status bar height and the height of the modal has to account for the vertical offset of the view.

Before:
image

After:
image

### Side effects
I haven't noticed any. The iOS 13 devices (with or without a notch) still look the same.

## :squid: Permissions
🦑


simonrozsival:
That definitely looks off. Thanks for pointing it out!

simonrozsival:
@Odrakir I created a follow-up issue https://github.com/toggl/mobileapp/issues/6772 for the problem you pointed out. I will now merge this PR so we can continue with the release.

PR #6740, opened 2019-11-29
🤖 Create code owners file for automatic review assignment

amulware:
## What's this?
This creates a code owners file which uses github teams and the auto pull assigner to automatically assign reviewers based on what code was changed.

## Why do we want this?
To further automate reviewer assignment.

## How is it done?
Using GitHub's code owner feature. Docs here: https://help.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners#codeowners-file-location

### Why not another way?
Pretty sure this is the only way of doing it. And I'm not 100% sure this will even work.

### Side effects
If this works, it could lead to automatic assignments we don't want, causing spam. Or even worse, we forget to change the automatic assignment, and things are confusing until someone notices.

## Review hints
Create a pull request against this branch changing various parts of the code and see if the assignment works.

Oh look, I did it for you and it works: https://github.com/toggl/mobileapp/pull/6741

I'm not sure what happens if we edit both iOS and Android files though 🤔 

## :squid: Permissions
~~Only me. I want this tested first, and I also would like comments from multiple people on whether you think this is a good thing.~~
It works. Anyone can 🦑 this.


amulware:
Implemented the first change asked for. We currently don't have a sync guardian setup, and I have no time to create it. That's something for down the line.

PR #6741, opened 2019-11-29
Code owners test

amulware:
## What's this?
This is a test to see if the code owners file created in #6740 works.

If it works, I'd expect the Random iOS Dev team to be assigned, and for that to be replaced by a random iOS developer within moments.

## :squid: Permissions
Do NOT merge this under any circumstances.


amulware:
The test worked. The above assignment, though it looks like I did it, was automatic. :)

PR #6742, opened 2019-11-29
Fix Japanese copy about support being offered in English only

daividssilverio:
## What's this?
This adds the note about the customer support being only offered in English in the feedback screen.

### Relationships
Ref #6514 

## Why do we want this?
Because we only offer support in English at the moment.

## How is it done?
I've translated it using my knowledge about Japanese and checked using some online dictionaries.
Please note that the space after `追伸` is intentional.

### Why not another way?
Feel free to suggest something different. (You don't have to be from the Toggl team to make a suggestion).

## Localizations
This is changing the `FeedbackFieldPlaceholder` copy in Japanese.

## :squid: Permissions
🦑 it.


heytherewill:
Should this not target the release branch?

simonrozsival:
I suppose this should be cherry-picked into both release branches (both have already branched out).

PR #6745, opened 2019-11-29
📊 Reports: new Bar Chart cell

### LINKED ISSUE #6522: Rewrite the ReportsBarChartCardView to be independent of the business model
Rewrite the `ReportsBarChartCardView` to be independent of our business model. 

More details about the reasoning about the need for separation is given in #6503. That issue also describes the purpose of the generic data structure needed to show the barchart data without the need for the chart to know anything about _time entries_ or _billables_.

### **🛠 Technical remarks**
Utilize the classes mentioned in #6503 and implemented in a corresponding PR.

### 🔗 Dependencies
Requires completion of #6503.

### **🏆 Outcome**
The UI control can display any kind of data in a barchart where each bar can display also a differently-styled portion, such as seen in [this design](https://app.zeplin.io/project/5a0d3cd27a75add1a8924f6e/screen/5a8613eca212ab0471a3def0)).

`MEDIUM`

image

dshr:
## What's this?
This is the new and sparkly bar chart cell, with simpler data binding and all.

### Relationships
Closes #6522

## Why do we want this?
For a reports view that is proper magical ✨ 

## How is it done?
Pop everything in a UICollectionViewCell and bind stuff from the IReportElement

### Why not another way?
Simple enough tbh

## :squid: Permissions
Reports folks only!


Toggl.Core.UI/ViewModels/Reports/ReportProjectsBarChartElement.cs
```diff
@@ -11,40 +13,60 @@ public class ReportProjectsBarChartElement : ReportBarChartElement
...
+        public ReportProjectsBarChartElement(ITimeEntriesTotals report, DateFormat dateFormat)
+            : base(convertReportTimeEntriesToBars(report), convertReportTimeEntriesToXAxisLabels(report, dateFormat), convertReportTimeEntriesToYAxisLabels(report))
         {
         }
 
         private static IEnumerable<Bar> convertReportTimeEntriesToBars(ITimeEntriesTotals report)
+            => report.Groups
+                .Select(group => new Bar(group.Billable.TotalHours, group.Total.TotalHours));
+
+        private static IEnumerable<string> convertReportTimeEntriesToXAxisLabels(ITimeEntriesTotals report, DateFormat dateFormat)
```
kornelijepetak: Since you are sending this collection into the constructor of `ReportBarChartElement` which does the `ToList()`, there is no need to create a list here.

This function returns the `IEnumerable` so you can simply yield values out instead of putting them into the list and returning that list.

Toggl.Core.UI/ViewModels/Reports/ReportProjectsBarChartElement.cs
```diff
@@ -11,40 +13,60 @@ public class ReportProjectsBarChartElement : ReportBarChartElement
...
             }
-
-            return new DateTimeOffsetRange(start.AddMonths(howMany), start.AddMonths(howMany + 1));
+            else
+            {
+                maxValue = maxTime.TotalSeconds;
+                halfValue = maxValue / 2;
+                timeUnitString = Resources.UnitSecond;
+            }
+            return new List<string> {$"0 {Resources.UnitSecond}", $"{halfValue} {Resources.UnitSecond}", $"{maxValue} {Resources.UnitSecond}"};
```
kornelijepetak: Just `yield return` this.

Toggl.iOS/Cells/Reports/ReportsBarChartCollectionViewCell.cs
```diff
@@ -20,6 +31,99 @@ protected ReportsBarChartCollectionViewCell(IntPtr handle) : base(handle)
...
+        public void SetElement(ReportBarChartElement element)
+        {
+            if (element.Bars.IsEmpty)
+            {
+                BarChartContainerView.Hidden = true;
+                return;
+            }
+            BarChartContainerView.Hidden = false;
+
+            MaximumHoursLabel.Text = element.YLabels[2];
```
kornelijepetak: If this will always be just three values and not N values distributed equally, maybe it makes sense to create a struct with 3 props for each label, to avoid this index access which holds no obvious semantics.

Toggl.iOS/Cells/Reports/ReportsBarChartCollectionViewCell.cs
```diff
@@ -20,6 +31,99 @@ protected ReportsBarChartCollectionViewCell(IntPtr handle) : base(handle)
...
+            }
+
+            var barViews = createBarViews(element.Bars);
+            foreach (var barView in barViews)
+            {
+                BarsStackView.AddArrangedSubview(barView);
+            }
+            BarsStackView.Spacing = (nfloat)getSpacingForNumberOfItems(element.Bars.Count());
+
+            if (element.XLabels.Count() <= 2)
```
kornelijepetak:
```suggestion
            if (element.XLabels.Count <= 2)
```

kornelijepetak: However, this is sketchy condition. The problem is in the scenario where it is actually `< 2`, this condition will be `true`, but the body will fail on index `1`. I'd set this to `== 2`, and instead of `else` do a `else if (count > 2)` and in `else` simply draw nothing. Or crash the app so that it's obvious that it should have at least 2 labels.

Toggl.iOS/Cells/Reports/ReportsBarChartCollectionViewCell.cs
```diff
@@ -20,6 +31,99 @@ protected ReportsBarChartCollectionViewCell(IntPtr handle) : base(handle)
...
+            }
+        }
+
+        private IEnumerable<UIView> createBarViews(IEnumerable<Bar> bars)
+            => bars.Select<Bar, UIView>(bar =>
+            {
+                if (bar.TotalValue == 0 && bar.FilledValue == 0)
+                    return new EmptyBarView();
+
+                return new BarView(bar.FilledValue, bar.TotalValue);
```
kornelijepetak:
```cs
return bar.TotalValue == 0 && bar.FilledValue == 0
	? new EmptyBarView()
	: new BarView(bar.FilledValue, bar.TotalValue);
```

Toggl.iOS/Views/Reports/BarView.cs
```diff
@@ -20,14 +21,15 @@ internal sealed class BarView : UIView
...
         {
-            this.bar = bar;
+            this.filledValue = filledValue;
+            this.totalValue = totalValue;
 
             nonBillableView = new BarSegmentView(Colors.Reports.BarChart.NonBillable.ToNativeColor());
             billableView = new BarSegmentView(Colors.Reports.BarChart.Billable.ToNativeColor());
             minimumLevelView = new BarSegmentView(
-                bar.BillablePercent > bar.NonBillablePercent
+                filledValue > totalValue
```
kornelijepetak: how can this happen? I mean, previously `billable + nonbillable = total`, but now, you don't have two additive values, but `filled` is only a fraction of `total`. Check this math to make sure it's fine.

Issue #6498, opened 2019-11-08
Create DateRangePickerViewModel


kornelijepetak:
### **🎯 What needs to be done?**
Create a `DateRangePickerViewModel` to govern the behavior of the `DateRangePicker` control (#6497 & #6499). 

♻ As part of this issue, also rename the `ReportsSource` into `DateRangePickerShortcut`.

### **📝 Detailed explanation**
This view model serves threefold purpose:
1. Provide month data for the `DateRangePicker` items (months), also making sure that the user's **beginning of week** settings is taken into account.
2. Coordinate the relationship between current selection and the currently selected shortcut. This should work in both ways.
3. Manage the state of selection during the two-tap range selection.

**Other constraints.**
* The calendar should not show months past the current month.
* The calendar should show only months back until the start of previous year (so, for Nov 8 2019, the first month to include is: January 2018.

### **🛠 Technical remarks**
🧪 In normal circumstances, to make the `DateRangePicker` a black box as a UI control, the behavior for it should be contained inside it. But since it's hard to test it that way, it's much easier to extract the behavior into a view model that can then be tested. Additional factor here is that the behavior should be the same across platforms so it makes sense to make it shared. 

Note, however, that this is a view model that can basically be instantiated by the control itself as it defines the UI control's behavior and nothing else.

**Code consideration.** To make things simpler from the perspective of interaction between the control and its view model, the suggested (use your best judgement) interface should be similar to:

```cs
InputAction<DateTimeOffset> DaySelected { get; }
InputAction<DateRangePickerShortcut> ApplyShortcutRange { get; }
IObservable<DateRangePickerShortcut?> SelectedShortcut { get; }
IObservable<ImmutableList<DateRangePickerMonthInfo>> Months { get; }
```

In the above proposal, `DateRangePickerMonthInfo` would contain information about a month (such as the day with which it starts), but also, what days in it are currently selected.

### **🏆 Outcome**
The output artifacts are:
* the view model and 
* the tests that prove that the view model is behaving correctly in all use cases.

### **👕 Estimated complexity**
`EXTRA LARGE`

### **🖼 Express your sentiment towards the issue**

![image](https://media.giphy.com/media/3o7btPCcdNniyf0ArS/giphy.gif)

kornelijepetak:
[...]

PR #6752, opened 2019-12-02
✨ Add DateRange picker View Model

kornelijepetak:
## What's this?
Implements the view model for the DateRange picker.

* ✅ Still needs tests (but for the sake of finishing this sooner for the Android UI and subsequently for the iOS version, it will be added in another PR)

### Relationships
Closes the #6498 sans the tests.

## Why do we want this?
To simplify the Date Range selection.

## How is it done?
The `DateRangePickerMonthInfo` represents a single month and everything the view needs to dumbly draw the month item. The DateRange picker UI control should only inform the viewmodel about what date was tapped, everything else is inside the viewmodel.

## Review hints
Commit by commit.

## :squid: Permissions
Me.


Toggl.Core.UI/ViewModels/DateRangePicker/DateRangePickerViewModel.Shortcuts.cs
```diff
@@ -0,0 +1,151 @@
...
+
+        private class LastYearShortcut : DatePickerShortcut
+        {
+            public override ReportPeriod Period => ReportPeriod.LastYear;
+
+            public override DateRange DateRange
+            {
+                get
+                {
+                    var year = DateTime.Today.Year - 1;
```
heytherewill: Please use the date service to calculate the current year

Toggl.Core.UI/ViewModels/DateRangePicker/DateRangePickerViewModel.cs
```diff
@@ -46,6 +59,10 @@ public sealed partial class DateRangePickerViewModel : ViewModel<DateRange, Date
             this.interactorFactory = interactorFactory;
             this.schedulerProvider = schedulerProvider;
 
+            WeekDaysLabels = Enumerable.Range(0, daysInWeek)
+                .Select(_ => "")
```
heytherewill: wat

kornelijepetak: I was trying to prevent uninitialized access, but after checking the code, it seems that `OnCreate` happens after the call to the ViewModel's `Initialize`, so this should not be necessary.

PR #6754, opened 2019-12-02
Android Calendar zoom

### LINKED ISSUE #5861: Calendar: Add zoom in and out in the calendar view
Add zoom in and out in the calendar view same as iOS

heytherewill:
## What's this?
This lets our users pinch to zoom in the calendar

### Relationships
Closes #5861 

## Why do we want this?
It's a bloody cool feature

## How is it done?
Some math 

### Why not another way?
Right now, zoom only applies to the current page. Should it apply for all pages at once? I don't think it should, but I'll ping @maritoggl for answers.

## Review hints
Fiddle around with the app and make sure the zoom is TIGHT

## :squid: Permissions
Only myself after one droid dev and @daividssilverio have reviewed


daividssilverio:
I really think the zoom level should apply to all pages to make things simpler and consistent. The pages already share offsets when switching pages after all.

If we are not doing that, we've got to handle it, maybe by converting offsets to use the base scale level and let other pages receive the offset change signal using that. Or maybe by removing the shared pages offset.

As it is right now, there's a bug when you max zoom in one page, scroll all the way to the bottom and change pages. You'll get a white page momentarily.

daividssilverio:
:sogood:

PR #6765, opened 2019-12-02
Make the calendar menu actions feel faster

### LINKED ISSUE #6384: Some contextual actions take too long to finish and there's no progress indication
slow-close 2019-10-29 14_41_13

The app looks unresponsive when you press some actions, namely all actions that don't involve navigation. It's possible to execute other actions while the first action doesn't complete.

What should we do? Simply early closing the menu helps with the look and feel of the view, but the effect of the menu action would take some time to complete e be rendered.

Displaying some sort of progress indicator is also an option.

daividssilverio:
## What's this?
This changes how the calendar contextual menu actions behave.
Making the contextual menu close when certain actions are executed.
It makes the actions feel a tad faster.

### Relationships
Maybe closes #6384

## Why do we want this?
To avoid having slow-looking actions in our apps.

## How is it done?
By closing the menu before the contextual menu main actions run (the ones with IO).

### Why not another way?

### Side effects
Right now, it simply doesn't work right.
If make changes by dragging to start/end times, closing the menu also has the side effect of discarding the changes in the day view (at least on android).

## Review hints
Check the section below.

## :squid: Permissions
🦑 after you check the following:
- [ ] This draft has become a PR
- [ ] Every action in the contextual menu closes the menu immediately
- [ ] For the `Save` actions, please check that the action is "optimistic" and the calendar item won't blink with value before the changes then back into the saved changes after IO changes finish executing
- [ ] For every action, please check that if you execute some action and then quickly open another calendar item contextual menu, the latter will work alright (that is, the calendar now being edited is rendered in "edit mode"
- [ ] Please check if iOS works alright as well. Please help fixing it if you know how.


daividssilverio:
Supersed by #6775

PR #6778, opened 2019-12-04
🐛Fix auth card bottom padding

### LINKED ISSUE #6760: Login navbar overlaps bottom Sign up button
## 🐛 Bug Report
Users on Android 10 but with classical navbar are unable to sign in because of this: 
device-2019-12-02-113702

## Context

Build/Branch: android-2.8
Phone OS: Android 10
Phone model: OG Pixel

## Expected behavior
Bottom card view should be displayed above the navbar and should be clickable with ease

## To reproduce
Take your android 10 phone
Go back to classical navigation
Logout from Toggl app 
Try to sign up

semanticer:
## What's this?
Fix for our bottom login and sign up cards being overlapped by navbar.

### Relationships

Closes #6760

## Why do we want this?
Users should be able to sign up right?

## How is it done?
Using our magical `FitBottomInset()` method on the ViewGroup we use inside of the Card. Also adding extra `6dp` to make a little less room for accidents. Also, the card view has wrap_content height now and all styling attributes are extracted to style. This style is then used in login & sign up layouts.

## Review hints
Functional tests on multiple different devices are welcome

## :squid: Permissions
Anyone, anytime


semanticer:
Yeah, I realized that the second I push the button :-D Sorry for that, it's correct now.

PR #6781, opened 2019-12-04
Reports: Remove old stackview contents

dshr:
## What's this?
A small fix to a small bug where we'd hang onto old UIStackView contents in CollectionViewCells when the cells get reused.

### Relationships
Closes nothing.

## Why do we want this?
No UI bugs!

## How is it done?
By calling `RemoveFromSuperview()`.

## :squid: Permissions
Reports folks only!


simonrozsival:
For future reference: https://developer.apple.com/documentation/uikit/uistackview/1616235-removearrangedsubview

PR #6784, opened 2019-12-05
Android widget improvements

### LINKED ISSUE #6768: Widgets: client name in dark theme
## 🐛 Bug Report
Client name remains black even in dark theme, therefore they are almost invisible
blackclientsinwidgets

## Context

Build/Branch: android-2.8
Phone OS: Android 10
Phone model: OG Pixel with dark theme

## Expected behavior
Client names should probably be white in dark theme

### LINKED ISSUE #6769: Widgets: cut off client name
## 🐛 Bug Report
The client name is cut off in our running TE widget
widget fail

## Context

Build/Branch: android-2.8
Phone OS: Android 10
Phone model: OG Pixel

## Expected behavior
We should either move the client name after the project name or change the layout completely to get more vertical space.

## To reproduce
This might be specific to the phone and launcher used. I'm using default google launcher on OG Pixel (5 inches)

heytherewill:
## What's this?
This fixes some things about our widgets that are off

### Relationships
Closes #6768 
Closes #6769

## Why do we want this?
The toggl widget is pretty useful and we need to polish it!

## How is it done?
- I'm using the client color in the client label
- I'm moving the client label next to the project name (Like it's in the main app and in the suggestions widget)
- I'm adding an empty state to the timer widget

### Why not another way?
We need to figure out the empty state's copy and visuals
Also, I'd like to center the start/stop button.

## Review hints
Play around with the widgets

## Localizations
I need to extract the string resource and later translate it to japanese. I'm waiting for the final copy from @maritoggl or @amulware 

## :squid: Permissions
Only myself


juanlaube:
The text could use a padding on the left, other than that it looks good
Screenshot_1575658691

heytherewill:
Japanese translation is missing, but you can review as though it's there because @daividssilverio will provide one as soon as he starts working

Toggl.Shared/Resources.ja.resx
```diff
@@ -1331,4 +1331,8 @@
     <data name="LoginToShowSuggestions" xml:space="preserve">
         <value>提案を表示するためにログイン</value>
     </data>
+    <data name="NoRunningTimeEntry" xml:space="preserve">
+        <value>No running time entry</value>
```
daividssilverio: 計測中記録はありません
```suggestion
        <value>No running time entry</value>
```

Toggl.Shared/Resources.ja.resx
```diff
@@ -1331,4 +1331,8 @@
...
+        <comment>Used in the Android widget to indicate that there are no time entries running</comment>
```
daividssilverio: Should we leave the comments in the translations as well?

heytherewill: No, I'll remove this

PR #6786, opened 2019-12-06
Reduce iOS app size

simonrozsival:
## What's this?
This PR updates the project configuration files in order to decrease the size of the app and also to simplify these project files so that they're not so scary (at least the iOS ones).

### Relationships
There was no issue, but I think it is actually useful, so I'm presenting this PR.

## Why do we want this?
Smaller size of the file our users download from the AppStore.

## How is it done?
- The widget project included native debugging symbols - I ticked the checkbox to strip these symbols.
- I also unified the settings of all `Release|*` configurations of all iOS projects.
- I then tried to remove as many references to assemblies from the csproj files as possible so that it still builds.

Release build before this change: https://app.bitrise.io/build/20320711ac777734#?tab=artifacts (the iOS 2.8 release)
- compressed size (.ipa): 45.5 MB
- extracted size: 151.2 MB
Release build after this change: https://app.bitrise.io/build/d2a66d1f07287c8c#?tab=artifacts
- compressed size (.ipa): 41.3 MB
- extracted size: 116.6 MB

That's a saving of 34.6 MB (23%). That's quite significant. I hoped we could reduce the size even further (For example the app file contains the `System.Xml.dll` file 4 or 5 times and this file is 0.5 MB large. I don't know which of our libraries requires `System.Xml` or maybe it is always required by Mono 🤷‍♂ it would be interesting to get rid of it completely).

The sizes are taken from Finder, Bitrise measures file sizes differently than Apple (1024 vs. 1000).

### Why not another way?
🤷‍♂ 

### Side effects
Something might break, this requires careful testing.

## Review hints
- Try the ad-hoc build on your device.
- Test Widget
- Test Siri (UI is important)

## :squid: Permissions
Me. When there are at least 2 reviews.


juanlaube:
The changes make sense to me, however I tested an ad hoc build and the app install size is 173 Mb, vs 153 Mb from the latest AppStore build 🤔

PR #6787, opened 2019-12-06
📆 Add DateRangePicker for android

### LINKED ISSUE #6497: Create a DateRangePicker for Android
Create a `DateRangePicker` UI widget that user can use to
* manually select a time range
* quickly select a specific time range by using shortcuts (such as today, yesterday, this week, or similar).

As already explained, the control should contain both the calendar and the shortcuts. Shortcuts are the constituent part of the control.

**Behavior considerations.** 
The selection is done by two taps.
* First tap determines the start of the period
* First tap still does not output a range
* Second tap determines the end of the period
* Second tap outputs a selected time range
* Tapping the same day twice will set the range to include only that one day
* If the second tap happens on the day that is earlier than the first tap, invert the **start** and **end** so that the those boundaries are always chronologically consistent.

### **🛠 Technical remarks**
This control should utilize the `DateRangePickerViewModel` (#6498) to coordinate the selection of the range and also the interaction between the shortcuts and the range.

**Public interface.** To conform to the remainder of native controls interface, the relevant public interface of the control should be only
```cs
void Select(DateTimeOffsetRange range);
void SetBeginningDayOfWeek(DayOfWeek beginning);
DateTimeOffsetRange SelectedRange { get; }
event EventHandler<SelectedRangeChangedChangedEventArgs> SelectedRangeChanged { get; }
event EventHandler<ShortcutClickedEventArgs> ShortcutClicked { get; }
```

No other abstractions should be leaked from the control.

**Day/Month architecture.** Previously, the UI for the calendar was made out of `ViewPager` with a `RecyclerView` as each page, with each day as a `RecyclerView` item. This is needlessly too many items.

For this issue, we will make a `ViewPager` that can show months as pages, where each page is a `DateRangePickerMonth` `(☑️ insert issue here)` view that uses canvas to draw all that's needed. Considering that we do not have animations or any complex UI, canvas solution will work really fast.

### **🏆 Outcome**
A developer can use the `DateRangePicker` by placing it somewhere in the UI. It should correctly output the date ranges as the user is selecting them and fire the event when a shortcut has been used.>

`LARGE`

image

kornelijepetak:
## What's this?
Adds a DateRangePicker for android.
There is a lot here, so follow the commits to understand what's going on.

This PR should be helpful for iOS version, especially the drawing code.

### Relationships
Closes #6497 

## Why do we want this?
Because we want to be able to select the date range.

### Side effects
I have changed the behavior from the original. The dialog should not do any loading of data until the user has accepted the time range. Otherwise, there will be multiple needless calls to our API.

## Review hints
Commit by commit, from https://github.com/toggl/mobileapp/pull/6787/commits/c7ec376c4ed1c49e72dfb4324535d8a309b430ee. (For some reason, I can see some older commits that were already squashed into the project branch. 🤷‍♂ 

image


kornelijepetak:
One thing I know I still have to fix is:
* Use time service instead of relying on `DateTime.Today`.

Toggl.Droid/Resources/layout/DatePickerShortcutCell.axml
```diff
@@ -0,0 +1,15 @@
...
+    xmlns:android="http://schemas.android.com/apk/res/android"
+    tools:text="This week"
+    android:textSize="14sp"
+    android:gravity="center"
+    android:letterSpacing="0.01"
+    android:fontFamily="sans-serif-medium"
+    android:textColor="@android:color/white"
+    android:background="@drawable/ReportsCalendarShortcutRoundedRectangle"
+    android:layout_width="100dp"
+    android:layout_height="36dp"
```
daividssilverio: Text with fixed height can end up getting cut if the font size is too big. `minHeight` can be more prudent.

Toggl.Droid/Extensions/ViewExtensions.cs
```diff
@@ -74,6 +75,12 @@ public void OnScrollChange(View v, int scrollX, int scrollY, int oldScrollX, int
             }
         }
 
+        public static Point GetLocationOnScreen(this View view)
+        {
+            var coordinates = new int[2];
+            view.GetLocationOnScreen(coordinates);
+            return new Point(coordinates[0], coordinates[1]);
```
daividssilverio: Could you maybe use the `Toggl.Shared.Point` instead? You are not using this android related APIs, I see no reason to use `Android.Graphics.Point`. Besides, you will be creating a bunch of hard-to-guarantee-that-they-will-be-disposed objects if you use this.

Toggl.Droid/Views/DateRangePickerMonthView.cs
```diff
@@ -0,0 +1,248 @@
...
+        }
+
+        private void drawDay(Canvas canvas, DateTime date)
+        {
+            textPaint.Color = monthInfo.Month == date.Month
+                ? monthTextColor
+                : extraMonthTextColor;
+
+            if (monthInfo.Selection?.Contains(date) ?? false)
+            {
```
daividssilverio: Maybe add another two methods for each branch of this if.

Toggl.Droid/Fragments/DateRangePickerFragment.cs
```diff
@@ -0,0 +1,140 @@
...
+
+        public override void OnResume()
+        {
+            base.OnResume();
+
+            var (widthPixels, _, _) = Activity.GetMetrics(Context);
+            var width = widthPixels - 24.DpToPixels(Context);
+            Dialog.Window.SetLayout(width, ViewGroup.LayoutParams.WrapContent);
+
+            Dialog.Window.DecorView.SetBackgroundColor(backgroundColor);
```
daividssilverio: It feels odd to see this... Why isn't the the dialog theme enough?

Toggl.Droid/Adapters/DateRangePickerMonthsPagerAdapter.cs
```diff
@@ -0,0 +1,94 @@
...
+
+        public void UpdateMonths(ImmutableList<DateRangePickerMonthInfo> months)
+        {
+            this.months = months;
+            NotifyDataSetChanged();
+
+            if (isFirstUpdate)
+            {
+                isFirstUpdate = false;
+                pager.CurrentItem = Count - 1;
```
daividssilverio: This does not look good. You've added a reference to the pager just to do this? This can cause leaks and it looks like a bad idea. Just do this outside of the adapter using the pager `pager.SetCurrentItem`.

Toggl.Droid/Toggl.Droid.csproj
```diff
@@ -1476,6 +1480,24 @@
       <SubType>Designer</SubType>
     </AndroidResource>
   </ItemGroup>
+  <ItemGroup>
+    <AndroidResource Include="Resources\layout\DatePickerShortcutCell.axml">
+      <Generator>MSBuild:UpdateGeneratedFiles</Generator>
```
daividssilverio: Wouldn't hurt to remove those `Generator`, `Subtype` tags, right? AFAIK, they are not doing anything here.

heytherewill: They indeed don't do a thing, yes

Toggl.Droid/ViewHolders/DateRangePicker/DateRangePickerShortcutViewHolder.cs
```diff
@@ -0,0 +1,49 @@
...
+        public DateRangePickerShortcutViewHolder(View itemView, Context context) : base(itemView)
+        {
+            normalShortcutColor = context.SafeGetColor(Resource.Color.shortcutBackground);
+            selectedShortcutColor = context.SafeGetColor(Resource.Color.selectionBackground);
+        }
+
+        protected override void InitializeViews()
+        {
+            shortcutText = ItemView as TextView;
+            backgroundDrawable = shortcutText?.Background as GradientDrawable;
```
heytherewill: You're null coalescing and casting here and latter you try to access this member, which can be null, without any null checks

Toggl.Droid/Views/DateRangePickerMonthView.cs
```diff
@@ -0,0 +1,248 @@
...
+                else if (date == monthInfo.Selection?.End && monthInfo.IsSelectionEndBoundary)
+                {
+                    drawBoundarySelection(canvas, false);
+                }
+                else
+                {
+                    drawFullSelection(canvas);
+                }
+            }
+            else
```
heytherewill:
```suggestion
            else if (monthInfo.Today.HasValue && monthInfo.Today == date && !monthInfo.IsTodaySelected)
```

Toggl.Droid/Views/DateRangePickerMonthView.cs
```diff
@@ -0,0 +1,248 @@
...
+                if (monthInfo.IsSelectionPartial)
+                {
+                    textPaint.Color = monthTextColor;
+                    drawPartialSelection(canvas);
+                }
+                else if (monthInfo.Selection.Value.IsSingleDay)
+                {
+                    drawSingleDaySelection(canvas);
+                }
+                else if (date == monthInfo.Selection?.Beginning && monthInfo.IsSelectionBeginningBoundary)
```
heytherewill: I wouldn't hate local extension methods on date that would made these assertions read like english, but that's optional

Toggl.Droid/Views/DateRangePickerMonthView.cs
```diff
@@ -0,0 +1,248 @@
...
+
+            canvas.DrawRect(xStart, verticalPadding, xEnd, rowHeight - verticalPadding, backgroundPaint);
+        }
+
+        private void drawCircle(Canvas canvas, int xOffsetFromCenter = 0)
+        {
+            canvas.DrawCircle(dayWidth / 2 + xOffsetFromCenter, rowHeight / 2, radius, backgroundPaint);
+        }
+
+        private void drawText(Canvas canvas, object obj)
```
heytherewill: `object`? Really? Why? Just pass a string here.

Toggl.Droid/Fragments/DateRangePickerFragment.cs
kornelijepetak: This seems to be the only solution for this for now. 🤷‍♂ 
If you manage to do this automatically by using theme values, please tell me how.

Toggl.Core.UI/ViewModels/DateRangePicker/DateRangePickerViewModel.cs
```diff
@@ -34,17 +34,23 @@ public sealed partial class DateRangePickerViewModel : ViewModel<Either<ReportPe
         private readonly ITimeService timeService;
 
         private BeginningOfWeek beginningOfWeek;
-        private BehaviorSubject<SelectionState> selectionState;
+        private SelectionState selectionState;
 
-        private readonly List<DatePickerShortcut> shortcuts = new List<DatePickerShortcut>();
+        private DateRange? result;
+
+        private readonly List<DateRangePickerShortcut> shortcuts = new List<DateRangePickerShortcut>();
```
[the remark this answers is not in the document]
kornelijepetak: I am not sure what kind of consistency you are talking about, but I've changed it to `ImmutableList` class. I see no added benefit of using the interface: it's a private variable anyway.

Toggl.Droid/ViewHolders/DateRangePicker/DateRangePickerShortcutViewHolder.cs
kornelijepetak: Indeed, but that is on purpose. If those items are actually not what I expect them to be, I'd rather have it crash. 🤔

Toggl.Droid/Views/DateRangePickerMonthView.cs
kornelijepetak: This is a derp. I was trying something before, forgot to change. 🤷‍♂

Toggl.Droid/ViewHolders/DateRangePicker/DateRangePickerShortcutViewHolder.cs
heytherewill: Why not crash here instead, then?

Toggl.Droid/Fragments/DateRangePickerFragment.cs
daividssilverio: If it doesn't work out of the box, I rather take a look later. Let's keep it as it is for now since it's working.

Toggl.Shared/Point.cs
```diff
@@ -2,10 +2,20 @@
...
     public struct Point
     {
-        public double X;
+        public double X { get; private set; }
 
-        public double Y;
+        public double Y { get; private set; }
 
-        public static Point Zero { get; } = new Point { X = 0, Y = 0 };
+        public Point(double x, double y)
```
daividssilverio: It looks like you broke iOS with the changes made on this file.

kornelijepetak: Yes, indeed. For some reason it was temporarily unloaded so it did not detect any errors. 🤷‍♂

PR #6789, opened 2019-12-06
Move day specific styles and colors to base res folder

### LINKED ISSUE #6788: No default resource values provided
## 🚧 Refactor Proposal
Some of our resource values are not present in the base res folder.
So far, I've found these colors which we have defined in `values-notnight` and 'values-night' but not in the base `values` folder
Screenshot 2019-12-06 at 15 19 10

This might be connected to this weird new crash https://appcenter.ms/orgs/Toggl/apps/Toggl-Android/crashes/errors/3767013764u/reports

We should probably just move our `values-notnight` values to `values` - it should work exactly the same.

semanticer:
## What's this?
We move some day-specific resources to the base resource folder because every resource should be represented in the base folder.

### Relationships

Closes #6788

## Why do we want this?
We don't want warnings and we need to make sure our app works on all devices. Also, we don't want our app potentially crashing on some devices https://appcenter.ms/orgs/Toggl/apps/Toggl-Android/crashes/errors/3767013764u/overview

## How is it done?
notnight folder was removed and everything was moved to main resource folder. Day values will be still taken from there and night values still have their folder

### Side effects
None I hope...

## Review hints
Check if colors are still selected correctly on your device when switching between day and night

## :squid: Permissions
🍣 it!


heytherewill:
@semanticer Do you want this in 2.8.1?

semanticer:
@heytherewill yes, I will cherry-pick it probably

PR #6790, opened 2019-12-06
🍩 Reports: iOS donut chart

### LINKED ISSUE #6505: Create a DonutChart control for iOS
Create a generic Donut chart.

We are replacing the PieChart with a donut chart.

Make sure that the chart is made in a generic way, with as much things parameterized as possible because we will probably have to tweak some things during testing. Let's write it good.

### **🛠 Technical remarks**
There is probably a lot of our current PieChart that can be salvaged, or at least to speed up the implementation. Make use of it.

### **🏆 Outcome**
The control that can be used anywhere in the UI that responds to any data provided from outside.

`EXTRA LARGE`

image

dshr:
# What's this?
Some spicy iOS 🍩 s. 

### Relationships
Closes #6505, partially. This does not do the "1% Other split", just plainly prints the labels.

## Why do we want this?
✨ 

## How is it done?
By salvaging as much of the old pie chart as possible and then cutting a hole in the middle.

### Why not another way?
This is simple. What else does one need?

### Side effects
Craving doughnuts like crazy.

## Review hints
Commit by commit should work best.

## :squid: Permissions
Reports peeps only


Toggl.iOS/Cells/Reports/ReportsDonutChartCollectionViewCell.cs
```diff
@@ -20,6 +23,42 @@ protected ReportsDonutChartCollectionViewCell(IntPtr handle) : base(handle)
...
+        public void SetElement(ReportDonutChartDonutElement element)
+        {
+            DonutChartView.Segments = element.Segments;
+        }
+
+        public void SetIsLast(bool last)
+        {
+            isLast = last;
+            SetNeedsLayout();
+
```
kornelijepetak: 🔥

Toggl.iOS/ViewSources/ReportsCollectionViewSource.cs
```diff
@@ -64,15 +64,15 @@ public override UICollectionViewCell GetCell(UICollectionView collectionView, NS
                     var barChartCell = collectionView.DequeueReusableCell(barChartCellIdentifier, indexPath) as ReportsBarChartCollectionViewCell;
                     barChartCell.SetElement(element);
                     return barChartCell;
-                case ReportDonutChartDonutElement _:
+                case ReportDonutChartDonutElement element:
                     var donutCell = collectionView.DequeueReusableCell(donutChartCellIdentifier, indexPath) as ReportsDonutChartCollectionViewCell;
-                    // TODO: populate cell
-                    donutCell.BackgroundColor = UIColor.SystemRedColor;
+                    donutCell.SetElement(element);
```
kornelijepetak: Since you are always calling `.SetIsLast` after `.SetElement`, I'd rather merge those two functions into one. You need layout refresh after both anyway. So there's no reason not to put the is-last information as a parameter of the SetElement method.

kornelijepetak: Same for 🍩 legend item cell.

Toggl.iOS/Views/Reports/DonutChartView.cs
```diff
@@ -0,0 +1,90 @@
...
+            {
+                ctx.SetFillColor(new Color(segment.Color).ToNativeColor().CGColor);
+
+                var percent = (float) segment.Value / totalValue;
+                var endAngle = startAngle + (float) FullCircle * percent;
+                var textAngle = (startAngle + endAngle) / 2;
+
+                // Draw arc
+                ctx.MoveTo(centerX, centerY);
+                Console.WriteLine(ctx.GetPathCurrentPoint());
```
kornelijepetak: I've checked, but I see no fancy donut graphics in console. Too bad. 😄

Toggl.iOS/Views/Reports/DonutChartView.cs
```diff
@@ -0,0 +1,90 @@
...
+        private const float donutInnerCircleFactor = 0.6f;
+
+        private static readonly UIStringAttributes attributes = new UIStringAttributes
+        {
+            Font = UIFont.SystemFontOfSize(10, UIFontWeight.Semibold),
+            ForegroundColor = UIColor.White
+        };
+
+        private IEnumerable<Segment> segments = new Segment[0];
+        public IEnumerable<Segment> Segments
```
kornelijepetak: I like this better than my solution. I'm gonna rewrite the droid code to use Segments property instead of a method. 

Nice one! 👍

Toggl.iOS/Views/Reports/DonutChartView.cs
```diff
@@ -0,0 +1,90 @@
...
+            var radius = centerX;
+            var innerRadius = centerX * donutInnerCircleFactor;
+            var textRadius = (radius + innerRadius) / 2;
+            var totalValue = (float) Segments.Sum(s => s.Value);
+
+            var startAngle = (float) Math.PI * -0.5f;
+
+            foreach (var segment in Segments)
+            {
+                ctx.SetFillColor(new Color(segment.Color).ToNativeColor().CGColor);
```
kornelijepetak: Can you please extract into named methods the drawing of different parts of the donut chart?

PR #6791, opened 2019-12-06
Track report events

### LINKED ISSUE #6524: Make sure we track all that is worth tracking
We need to track some events so that we know how users use the reports.

Currently, we are tracking:
* Reports loading success 
* Reports loading failures

Events track these parameters
* how the time range was selected
* how long the fetch took
* the total time range duration in days
* [Only `ReportsSuccess`]: number of unsynced projects

For now, we will only continue tracking what we are already tracking.

### **🛠 Technical remarks**
We have noticed that the numbers at AppCenter are somehow invalid, so let's really take care that these are implemented and tested well.

### **🏆 Outcome**
The events are appearing in AppCenter in correct events. 
Try to use adhoc builds (and make sure that the events are actually getting there).

`SMALL`

image

pdrj:
## What's this?
[WIP] This adds event tracking for reports.

### Relationships

Closes #6524 

## Why do we want this?
More data, which we might or might not use.

## How is it done?
Emitting events when generating `elements`.

### Why not another way?
Open to suggestions.

## Review hints
I tested that the events actually reach firebase from my debug build. I've not been able to validate the `Failure` one, because, I suppose, we have not had it in staging firebase events yet and I can't select it (even when manually crafting an URL).

Screenshot 2019-12-06 at 14 14 52

## :squid: Permissions
Reports crew.


Toggl.Core.UI/ViewModels/Reports/ReportsViewModel.cs
```diff
@@ -169,38 +182,66 @@ private async Task<ImmutableList<IReportElement>> reportElementsProcess(ReportFi
...
 
                 if (summaryData.Segments.None())
-                    return elements(new ReportNoDataElement());
-
-                return elements(
-                    new ReportWorkspaceNameElement(filter.Workspace.Name),
-                    new ReportSummaryElement(summaryData, durationFormat),
-                    new ReportProjectsBarChartElement(reportsTotal, dateFormat),
-                    new ReportProjectsDonutChartElement(summaryData, durationFormat));
+                    elementsToReturn = elements(new ReportNoDataElement());
```
kornelijepetak: Revert this to how it was before, and call the tracking method in the `finally` block.

Toggl.Core.UI/ViewModels/Reports/ReportsViewModel.cs
```diff
@@ -169,38 +182,66 @@ private async Task<ImmutableList<IReportElement>> reportElementsProcess(ReportFi
...
         }
 
         private ImmutableList<IReportElement> elements(params IReportElement[] elements)
             => elements.Flatten();
 
+        private void trackReportElementProcessCompletion(ReportsAnalyticsEventData eventData)
+        {
+            var loadingTime = timeService.CurrentDateTime.UtcDateTime - eventData.StartTimestamp;
+
+            if (eventData.IsSuccessful)
```
kornelijepetak: Since both `if` and `else` bodies are non-trivial, wrap them with braces.

kornelijepetak:
Can't say I have anything better to suggest.
Ping me on slack when you write tests for this.

kornelijepetak:
I have pushed to your branch what we talked about.

PR #6803, opened 2019-12-12
🍩 Make very small slices still take some space

### LINKED ISSUE #6798: Do not group legend items. Make Other category take at least 1%.
## 🎯 What needs to be done?
The donut has to reserve at least 1% for the `Other` category.
Additionally, legend items should not be grouped at all.

## 📝 Detailed explanation
Items smaller than 1% are grouped into the `Other` category. However, that can be even as small as 0.0001%. Such slices would be invisible on the donut chart, so this Other category needs to take at least 1% in the donut.

## 🔗 Dependencies
Related to #6504, #6505 and #6716 

## 🏆 Outcome
The `Other` element is at least 1%, if present.
Legend items are not grouped at all.

## 👕 Estimated complexity
`MEDIUM`

image

kornelijepetak:
## What's this?
Moves the grouping out of ReportElement and moves it to the UI layer (in a reusable way: it's defined in shared code because it's used by both platforms, but it's used only from UI)

### Relationships
Closes #6798

## Why do we want this?
So that we can also reasonably show very small slices.

## Review hints
Commit by commit.

## :squid: Permissions
📊


Toggl.Core.UI/ViewModels/Reports/ReportDonutChartElement.Segments.cs
```diff
@@ -0,0 +1,48 @@
...
+                => HashCode.Combine(Color, Label, Value);
+
+            public static bool operator ==(Segment left, Segment right)
+                => left.Equals(right);
+
+            public static bool operator !=(Segment left, Segment right)
+                => !(left == right);
+        }
+
+        public class PercentageDecoratedSegment
```
pdrj: Should we add `Equals` for this type of segment in case we need to compare them in the future?

Toggl.Core.Tests/UI/ViewModels/Reports/ReportDonutChartElementTests.cs
```diff
@@ -92,24 +92,6 @@ public void DefaultOverloadCreatesCorrectNumberOfSubElements()
                 donutChart.SubElements.Skip(1).Should().AllBeOfType<ReportDonutChartLegendItemElement>();
             }
 
-            [Fact, LogIfTooSlow]
-            public void DefaultOverloadCreatesCorrectNumberOfSubElementsWithOtherCategory()
```
pdrj: We should now create `DonutChartGroupingTests` for the new logic, right?

Toggl.Core.UI/ViewModels/Reports/ReportDonutChartElement.Segments.cs
kornelijepetak: Not really. This decorator is used only to provide wrapping for display. There's no really logic happening at that point, or after. If at some point, we do need to introduce equality, it's easy to do.

Toggl.Core.Tests/UI/ViewModels/Reports/ReportDonutChartElementTests.cs
kornelijepetak: Does https://github.com/toggl/mobileapp/issues/6805 cover what you needed?

pdrj: Yes.

Issue #6805, opened 2019-12-13
Validate the reports behavior by tests


kornelijepetak:
## 🎯 What needs to be done?
Write tests for newly introduced components.

## 📝 Detailed explanation
During development, for the sake of speed and parallizing, we have omitted tests here and there. They still need to be done. Here's a list of things that needs more tests:
_(I will complete this list in this list during the day)_

* [x] `ReportsViewModel`
    * [x] Analytics
* [x]  Report elements (construction, equality)
	* [x] `ReportElementBase`
	* [x] `ReportErrorElement`
	* [x] `ReportNoDataElement`
	* [x] `ReportSummaryElement`
	* [x] `ReportDonutChartElement`
		* [x] `ReportProjectsDonutChartElement`
	* [x] `ReportDonutChartDonutElement`
	* [x] `ReportDonutChartLegendItemElement`
		* [x] `ReportProjectsDonutChartLegendItemElement`
	* [x] `ReportBarChartElement`
		* [x] `ReportProjectsBarChartElement`
	* [x] `ReportWorkspaceNameElement`
* [x] `DonutChartGrouping`

## 🏆 Outcome
All relevant items have tests and they all pass.

## 👕 Estimated complexity
`MEDIUM`

PR #6806, opened 2019-12-15
🐛 Makes the label show the shortcut name if the range is appropri…

### LINKED ISSUE #6801: The date label does not show a shortcut name
## 🎯 What needs to be done?
The button for opening the DateRange picker does not show the shortcut names, such as `This week`, `Yesterday`, etc. It shows dates for all cases. Make it so that the label shows the shortcut name if selected.

## 👕 Estimated complexity
`SMALL`

kornelijepetak:
## What's this?
This fixes a small bug with a big change.

I've introduced another service: `CalendarShortcutsService`. This service is used by multiple pieces of code and serve as some kind of generic store for dealing with things like what `This week`, or `Last year` means. It is based on the `TimeService` to get the correct current local time and is not tied directly to the reports which means it can be used by other parts of system as well, wherever those concepts are needed.

Then, I simply refactored the `ReportsViewModel` and the `DateRangePickerViewModel` to use this service instead of the time service (which was basically injected for the same reason).

### Relationships
Closes #6801

## Why do we want this?
Easier calendar shortcut manipulations.

## Review hints
As usual. Commit by commit.

## Tests
There will be much more tests coming as part of #6805.

## :squid: Permissions
Me.


Toggl.Core.UI/UIDependencyContainer.cs
```diff
@@ -41,29 +44,43 @@ private IDeeplinkParser createDeeplinkParser()
...
         protected override void RecreateLazyDependenciesForLogout()
         {
             base.RecreateLazyDependenciesForLogout();
-            WidgetsService?.Dispose();
-            widgetsService = new Lazy<IWidgetsService>(CreateWidgetsService);
+
+            widgetsService = reinitializeService(widgetsService, CreateWidgetsService, forced: true);
+        }
+
+        private Lazy<TService> reinitializeService<TService>(Lazy<TService> service, Func<TService> serviceCreator, bool forced = false)
```
simonrozsival: YAGNI

Toggl.Core.UI/ViewModels/Reports/ReportsViewModel.cs
```diff
@@ -151,10 +155,14 @@ private IObservable<ImmutableList<IReportElement>> reportElements(ReportFilter f
             .ToObservable()
             .StartWith(createLoadingStateReportElements());
 
-        private string formattedTimeRange(DateTimeOffsetRange range, DateFormat dateFormat)
+        private string formattedTimeRange(DateRange range, DateFormat dateFormat)
         {
-            var startDateText = range.Minimum.ToString(dateFormat.Short, DateFormatCultureInfo.CurrentCulture);
-            var endDateText = range.Maximum.ToString(dateFormat.Short, DateFormatCultureInfo.CurrentCulture);
+            var knownShortcut = calendarShortcutsService[range];
+            if (knownShortcut != null)
```
simonrozsival: I actually don't like that you return `null` if you don't find anything. What if we used a `TryGet(range, out var knownShortcut)`? I would prefer that over the indexer. What do you think about that?

Toggl.Core.UI/Services/CalendarShortcuts/ICalendarShortcutsService.cs
```diff
@@ -0,0 +1,16 @@
...
+using Toggl.Core.UI.Models;
+using Toggl.Shared;
+
+namespace Toggl.Core.UI.Services
+{
+    public interface ICalendarShortcutsService : IDisposable
+    {
+        ImmutableList<CalendarShortcut> Shortcuts { get; }
+
+        CalendarShortcut this[ReportPeriod period] { get; }
```
simonrozsival: Let's not use indexers please :/

Toggl.Core.UI/UIDependencyContainer.cs
kornelijepetak: IAGN what?
It is used. Even multiple times. 🤔

Toggl.Core.UI/Services/CalendarShortcuts/ICalendarShortcutsService.cs
kornelijepetak: You are right, I'm actually not using an `index`, makes more sense as a named function.

Toggl.Core.UI/ViewModels/Reports/ReportsViewModel.cs
kornelijepetak: You may not like it but it's precisely what `null` is for: the lack of value.
I don't want to do `out` vars for that reason. 
I will, however change the indexer.

Toggl.Core.UI/UIDependencyContainer.cs
simonrozsival: I meant the `forced` boolean and I overlooked that it is used on like 71. My comment was irrelevant.

Toggl.Core.UI/ViewModels/Reports/ReportsViewModel.cs
simonrozsival: Now that it's not an indexer, I don't mind the `null` return value (I believe the indexer should throw "out of bounds" instead of returning `null`).

PR #6820, opened 2019-12-17
📅 Reports: iOS DateRangePicker

### LINKED ISSUE #6499: Create a DateRangePicker for iOS
Create a `DateRangePicker` UI widget that user can use to
* manually select a time range
* quickly select a specific time range by using shortcuts (such as today, yesterday, this week, or similar).

As already explained, the control should contain both the calendar and the shortcuts. Shortcuts are the constituent part of the control.

**Behavior considerations.** 
The selection is done by two taps.
* First tap determines the start of the period
* First tap still does not output a range
* Second tap determines the end of the period
* Second tap outputs a selected time range
* Tapping the same day twice will set the range to include only that one day
* If the second tap happens on the day that is earlier than the first tap, invert the **start** and **end** so that the those boundaries are always chronologically consistent.

### **🛠 Technical remarks**
This control should utilize the `DateRangePickerViewModel` (#6498) to coordinate the selection of the range and also the interaction between the shortcuts and the range.

**Public interface.** To conform to the remainder of native controls interface, the relevant public interface of the control should be only
```cs
void Select(DateTimeOffsetRange range);
void SetBeginningDayOfWeek(DayOfWeek beginning);
DateTimeOffsetRange SelectedRange { get; }
event EventHandler<SelectedRangeChangedChangedEventArgs> SelectedRangeChanged { get; }
event EventHandler<ShortcutClickedEventArgs> ShortcutClicked { get; }
```

No other abstractions should be leaked from the control.

**Day/Month architecture.** Previously, the UI for the calendar was made out of small items for each day. In the new way of doing things, the view model will give only information about months. It is the duty of this control to find a way to display such info on the control. (For reference, #6497 describes that Android is going to make it by utilizing custom views for each month and canvas-draw the days). Make sure to use a solution that is a good compromise between simple and performant (performance wins in this case).

### **🏆 Outcome**
A developer can use the `DateRangePicker` by placing it somewhere in the UI. It should correctly output the date ranges as the user is selecting them and fire the event when a shortcut has been used.>

`LARGE`

image

dshr:
#  What's this?
The long-awaited new datepicker for iOS

### Relationships
Closes #6499 

## Why do we want this?
Better date picking and not breaking on iPad.

## How is it done?
Rewriting the datepicker from scratch, with a more of a unidirectional data flow.

### Why not another way?
Seems like the better option

### Side effects
None.

## Review hints
Commit by commit

## :squid: Permissions
Reports folx


Toggl.iOS/ViewControllers/DateRangePickerViewController.cs
```diff
@@ -0,0 +1,119 @@
...
+
+            PreferredContentSize = popoverPreferedSize;
+
+            AcceptButton.SetTitle(Resources.Done, UIControlState.Normal);
+            CloseButton.SetTitle(Resources.Cancel, UIControlState.Normal);
+
+            setupDayHeaders(ViewModel.WeekDaysLabels);
+
+            calendarCollectionViewSource = new DateRangePickerViewSource(CalendarCollectionView);
+            calendarCollectionViewLayout = new ReportsCalendarCollectionViewLayout();
```
kornelijepetak: Just out of curiosity, how come you haven't gone the route of drawing each month with CGContext and having at most 25 elements in the collection? I am wondering whether having only 25 cells (of which many would be recycled) would be better and simpler than having several months with 30+ cells each.

I mean, now that I see that you haven't moved away from the old implementation, I understand the concerns you stated about passing data around.

Toggl.iOS/ViewSources/DateRangePickerViewSource.cs
```diff
@@ -0,0 +1,95 @@
...
+{
+    public sealed class DateRangePickerViewSource : UICollectionViewSource
+    {
+        private const string cellIdentifier = nameof(ReportsCalendarViewCell);
+
+        private readonly UICollectionView collectionView;
+        private readonly ISubject<DateRangePickerMonthInfo> currentPageNotScrollingSubject = new Subject<DateRangePickerMonthInfo>();
+        private readonly ISubject<DateTime> dayTaps = new Subject<DateTime>();
+
+        private IImmutableList<DateRangePickerMonthInfo> months = ImmutableList<DateRangePickerMonthInfo>.Empty;
```
kornelijepetak:
```suggestion
        private ImmutableList<DateRangePickerMonthInfo> months = ImmutableList<DateRangePickerMonthInfo>.Empty;
```
The list is immutable and the concrete implementation is as safe to use as the interface. Even worse, the interface hides some members that the concrete implementation has (see another comment). That being said, the variable is also private, so you are not even hiding the impl details with this.

Toggl.iOS/ViewSources/DateRangePickerViewSource.cs
```diff
@@ -0,0 +1,95 @@
...
+        {
+            months = newMonths;
+            collectionView.ReloadData();
+        }
+
+        public void ScrollToCurrentPage()
+        {
+            if (page == null)
+            {
+                for (var i = 0; i < months.Count; i++)
```
kornelijepetak: If I get correctly what you wanted, you are missing a return in that `if`, because this will continue looking thru the months until it goes thru them all, effectively giving you the LAST that satisfies the condition. I assume that is not what you want as I'd expect you to reverse the iteration in that case. So, I'll assume you wanted the first one.

In that case, if the `months` is `ImmutableList` instead of the counterpart interface, you can replace the `for` loop with:

`page = months.FindIndex(m => m.IsSelectionBeginningBoundary);`

In the case that my assumption is incorrect, and you actually did want the last one, you can use

`page = months.FindLastIndex(m => m.IsSelectionBeginningBoundary);` instead.

Toggl.iOS/ViewSources/DateRangePickerViewSource.cs
```diff
@@ -0,0 +1,95 @@
...
+                {
+                    if (months[i].IsSelectionBeginningBoundary)
+                        page = i;
+                }
+            }
+
+            if (page == null)
```
kornelijepetak: With the above change, you'd switch this into

```suggestion
            if (page < 0)
```

as the `FindIndex` returns `-1` for not-found.

Toggl.iOS/Views/Reports/ReportsCalendarViewCell.cs
```diff
@@ -8,15 +8,21 @@
 
 namespace Toggl.iOS.Views
 {
-    public sealed partial class ReportsCalendarViewCell : ReactiveCollectionViewCell<ReportsCalendarDayViewModel>
+    public sealed partial class ReportsCalendarViewCell :
+        ReactiveCollectionViewCell<(
```
kornelijepetak: Please change this into a proper named struct.

Tuples are not really a nice syntax for moving too much data around.

Also, if you are naming tuple elements, it's best to follow the usual conventions.

If the elements of the tuples are public (which they are and can't be private), PascalCase them.
Only when destructuring tuples inside of functions use camelCase, i.e.

```cs
var (_, isToday, isSelected, _, _, _) = tuple;
```

because those are then local variables.

dshr:
Why do you think a `CGContext`-based implementation would be more performant? Generally, going with a `UICollectionView` gives us much better flexibility in terms of animations/interactions further down the line, along with it just being "the way".
As for the other points:
1. Good catch — "gentler" swipes seems to glitch out, it's probably caused by a logic error in the "scroll to initial page" flow.
2. Yup, haven't merged that yet.
3. That's how it is in the current implementation.
4. Is that documented somewhere? I really just matched the implementation we currently have, so this is news to me.

Toggl.iOS/ViewControllers/DateRangePickerViewController.cs
```diff
@@ -0,0 +1,119 @@
...
+
+            ViewModel.Months
+                .Subscribe(calendarCollectionViewSource.UpdateMonths)
+                .DisposedBy(DisposeBag);
+
+            ViewModel.Shortcuts
+                .Subscribe(quickSelectCollectionViewSource.UpdateShortcuts)
+                .DisposedBy(DisposeBag);
+
+            calendarCollectionViewSource.CurrentPageNotScrollingObservable
```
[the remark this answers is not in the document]
dshr: Yeah, this could use a rename — this is fired when the current page has finished scrolling onto a new month.

Toggl.iOS/ViewControllers/DateRangePickerViewController.cs
```diff
@@ -0,0 +1,119 @@
...
+using UIKit;
+
+namespace Toggl.iOS.ViewControllers
+{
+    public partial class DateRangePickerViewController : ReactiveViewController<DateRangePickerViewModel>
+    {
+        private CGSize popoverPreferedSize = new CGSize(319, 397);
+
+        private DateRangePickerViewSource calendarCollectionViewSource;
+        private ReportsCalendarCollectionViewLayout calendarCollectionViewLayout;
```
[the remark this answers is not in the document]
dshr: This file is reused from the old impl, I thought I'd do the big rename in #6816. Can do that in this PR just as well.

kornelijepetak:
1. Because drawing on canvas is superfast (and on most platforms cacheable) and the total amount of instantiated elements would be 25. With each date being a cell (which amounts to potentially 750 controls, not sure how aggresively iOS recycles them), the processing time and memory goes up (multiple objects instantiated for each cell). As I said, it may not be as problematic on iOS as there are much fewer shitty phones. I am not forcing anything here. I am definitely not an expert for iOS in any stretch of imagination, so I leave this decision to you 😄 If you say it's "the way", fine by me.

2. Ok.

3. Ok, leave it like that then. Although, how hard would it be to make it so that single date markers (both selection and current date) are circular, not oval? With canvas it'd be trivial, not sure what you have to do here to achieve the same effect.

4. After some internal discussion, make it so that the page with the Beginning of the period is selected.

5. (new ✨) One more thing, which came to mind while I was doing the droid version, and is not strictly required by the issue itself, but would be a nice UX: To make it so that the partial selection is differently displayed to denote that something is still missing.
On droid this looks like this: 

image

Is this something that can be done in a reasonable amount of time?

Toggl.iOS/ViewControllers/DateRangePickerViewController.cs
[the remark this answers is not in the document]
kornelijepetak: Whatever is easier for you.

Toggl.iOS/Views/Reports/ReportsCalendarViewCell.cs
```diff
@@ -8,17 +8,40 @@
 
 namespace Toggl.iOS.Views
 {
-    public sealed partial class ReportsCalendarViewCell :
-        ReactiveCollectionViewCell<(
+    public struct DatePickerCellData
+    {
+        public DateTime Date;
```
kornelijepetak: Turn these into `{ get; }` properties for immutability reasons.

PR #6824, opened 2019-12-18
Go Loco: A better debug commands view

dshr:
## What's this?
A slight reiteration of the debug commands view we have on iOS. Initially planned to be x-plat, but didn't have enough time for it.

## Why do we want this?
Pretty things

## How is it done?
By reusing @Odrakir's composable settings table view.

### Why not another way?
But it's nice!

## Review hints
Commit by commit

## :squid: Permissions
No done yet, but all yours when done

### DIFF
--- a/Toggl.Core.Tests/Interactors/Settings/GetFeedbackInfoInteractorTests.cs
+++ b/Toggl.Core.Tests/Interactors/Settings/GetFeedbackInfoInteractorTests.cs
@@ -1,0 +1,222 @@
+using FluentAssertions;
+using FsCheck;
+using FsCheck.Xunit;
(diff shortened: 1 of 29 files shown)


simonrozsival:
I have a feature request: Show the name of the current sync state. I'm willing to implement it myself. Can I push to this branch? 😄

dshr:
@simonrozsival go for it! Just noticed I have a DateTime formatting issue in tests as well 🤦‍♂

simonrozsival:
I currently don't have enough time to implement my idea, so let's merge it as is. I pushed a fix for the failing test and I think we can merge it and use it in the next release!

PR #6829, opened 2019-12-18
🎨 Add dropdown next to report period description

pdrj:
## What's this?
This adds dropdown character next to report period description.

### Relationships

Closes #issue in my head.

## Why do we want this?
So it's clear users can select something.

## How is it done?
Adding a character.

### Why not another way?
I don't know.

### Side effects
There's a dropdown character next to description.

## Review hints
Screenshot 2019-12-18 at 13 04 20

## :squid: Permissions
Rprtz krew.


Toggl.Core.UI/ViewModels/Reports/ReportsViewModel.cs
```diff
@@ -115,6 +116,7 @@ public override async Task Initialize()
             FormattedTimeRange = Observable.Merge(Observable.Return(initialSelection), SelectTimeRange.Elements.WhereNotNull())
                 .CombineLatest(dateFormatObservable, resultSelector: formattedTimeRange)
                 .DistinctUntilChanged()
+                .Select(description => $"{description} {dropdownCharacter}")
```
kornelijepetak:
```suggestion
                .Select(dateRange => $"{dateRange} {dropdownCharacter}")
```

Toggl.Core.UI/ViewModels/Reports/ReportsViewModel.cs
```diff
@@ -26,6 +26,7 @@ namespace Toggl.Core.UI.ViewModels.Reports
 {
     public sealed class ReportsViewModel : ViewModel
     {
+        private static readonly string dropdownCharacter = "▾";
```
kornelijepetak:
```suggestion
        private const string dropdownCharacter = "▾";
```

use `static readonly` only when you have to actually calculate (or get) the value from the constructor.

PR #6832, opened 2019-12-18
🍩 Android DonutChart interaction

kornelijepetak:
## What's this?
Implements interaction with the donut on the reports.

### Relationships
🎉 GO LOCO PROJECT

## Why do we want this?
So that one can see what's what among slices.

## How is it done?
By some animations.

### Why not another way?
@amulware said the other one didn't cause butterflies in the stomach.

## Review hints
Slowly and carefully.

## :squid: Permissions
Me.


Toggl.Droid/Views/DonutChart/DonutChartView.SlicesCollection.cs
```diff
@@ -0,0 +1,55 @@
...
+                foreach (var percentageSegment in percentageSegments)
+                {
+                    var percentage = (float)percentageSegment.NormalizedPercentage;
+                    var sweepAngle = fullCircle * percentage;
+
+                    slices.Add(new Slice(percentageSegment, angleOffset, sweepAngle));
+                    angleOffset += sweepAngle;
+                }
+
+                slices[^1].CorrectEndAngle();
```
daividssilverio: what does the `[^1]` do? Gets the last element?

kornelijepetak: Yes.

Toggl.Droid/Views/DonutChart/DonutChartView.cs
```diff
@@ -0,0 +1,290 @@
...
+
+        private void drawSelectedSliceLabels(Canvas canvas)
+        {
+            textPaint.TextSize = selectedSliceTextSize;
+
+            textPaint.Color = selectionAnimator.Color;
+            var labelPosition = center + new Point(0, selectedSliceLabelVerticalOffset);
+            drawLabel(canvas, selectedSlice.PercentageSegment.Segment.Label, labelPosition);
+
+            textPaint.Color = Color.ParseColor("#888888");
```
daividssilverio: This color shouldn't be here, right? Please add it to the `colors.xml` file (or use one from there).

kornelijepetak:
1. Fixed.
2. Since the reports are gotten from the API, it always requires an additional request.
3. Fixed.
4. Unfortunatelly this is not that easy to solve atm. If you think it's really annoying, let's make an issue and discuss how we can solve it in future.
5. Similar to 4.

daividssilverio:
Hey @kornelijepetak , I'm pretty sure 2. will come back as a bug report from users.
If it's not fixed here, do you mind creating an issue for it? Maybe we can fix it in regular backlog.

PR #6833, opened 2019-12-18
Fix a bug in a "hack" to reject autocompletion when using Kanji

### LINKED ISSUE #6812: It's not possible to input kanji when searching for tags in the edit time entry view
## 🐛 Bug Report
Users using the Japanese soft input are not able to add/search for tags with Kanji in the Add/Search Tags view in iOS.

## Context
It happens in the Edit Time Entry view when trying to add/search for tags

Build/Branch: 2.8 (latest release, current develop)
Phone OS: iOS
Phone model: Tested on the iPhone 8 simulator

Reported by a user through app reviews.

## Expected behavior
The users should be able to add and search Tags that include kanji in their names.
It should work when using the qwerty keyboard and the 12 keys keyboard both. Kanjis should be able to be suggested by the system for the user to pick.

## To reproduce
1. Use a Japanese keyboard.
2. Click on an existing time entry.
3. Try to add a tag with Kanjis.

## Suggestions
It's possible to use Kanjis alright when starting a time entry and when adding/editing Projects on existing time entries. It should work the same way.

juanlaube:
## What's this?
Fixes a bug in the start time entry view when selecting a tag after using a filter with Kanji symbols that caused the tag to be ignored.

### Relationships

Closes #6812 

## Why do we want this?
Because 🇯🇵 

## How is it done?
A "hack" was in place to reject autocompletion from the keyboard, but this hack used `UITextView.Text` instead of `UITextView.AttributedText`.

### Why not another way?
🤷‍♂ 

### Side effects
I can't think of any.

## :squid: Permissions
🦑


juanlaube:
@simonrozsival don't worry, this textfield is a mess and I hate it, we should do something to have a nice UX without the code being too complicated, perhaps while merging the start and edit views 🤔

PR #6836, opened 2019-12-19
Fix the notifications permission request

### LINKED ISSUE #6252: User cannot turn on notfications
## 🐛 Bug Report
A user says they cannot turn on notifications, the dialog doesn't show in the calendar tab, in the smart alerts setting it looks like the notifications are not authorized and in the settings app the notification settings is not showing.

image
image (1)

## Context

Build/Branch: iOS 2.4
Phone OS: iOS 13.1.2
Phone model: iPhone11,6 (Xs Max)

## Expected behavior
User should be able to turn on notifications.

### LINKED ISSUE #6797: Notification permission flow broken on iOS
It appears that there's currently no way for users to grant us permission to send them notifications on iOS. This is relevant when users want to receive smart alerts about their calendar events. 

Currently, after installing the app and navigating to the smart alerts settings, users are displayed 
84A77AA1-3611-496F-8A6F-B5F5F35F3150_1_105_c

In the app settings, however, there's no notification setting
29A55826-D63B-4EAF-B725-64C794E8AAD7_1_105_c

Let's fix the flow. My proposal is following:
- when a user first navigates to the smart alerts menu, we ask them permission for sending notifications
- - if a user decides to grant permission, we take them to app settings where they can confirm their decision. Then they can go back to in-app smart alerts settings where they are displayed the list of possible alerts settings (the list that we already should have - disabled, when the event starts, 5 minutes before, 10 minutes before etc)
- - if user denies the permission the first time, we show them the "No worries, you can enable notifications later from the Settings" popup (similar to what we show when calendar access permission is denied). 
- when a user wants to disable smart alerts in the future, they can do that from the in-app smart alerts settings menu. At this point, permission to notify stays intact. 
- when user wants to revoke the permission to send notifications, they can do that via the app settings where they initially granted the permission. 
- we introduce a new first row in the smart alerts menu that says "Allow sending notifications" that's followed by a short explanation "Toggl needs permission to send you notifications in order to remind you of upcoming events in your calendar." That's the menu item that takes the user to app settings where the permission can either be granted or revoked - similar to what we have for the calendar access permission. 

Share your objections or proposals for a better flow. Also, seeking feedback from @alicjasuska if such UX sounds alright.

juanlaube:
## What's this?
Fixes the notifications permission flow.

### Relationships

Closes #6252 
Closes #6797 

## Why do we want this?
Because without this users can't enable calendar notifications.

## How is it done?
By requesting the notifications permission from the smart alert view as proposed.

### Side effects
None.

## Review hints
Please test it, here's a chart I made as a guide.

permissions

## :squid: Permissions
🚫 NO U


amulware:
Does this close #6252?

juanlaube:
Correct, I edited the description so it will close automatically

Toggl.Core.Tests/UI/ViewModels/NotificationSettingsViewModelTests.cs
```diff
@@ -69,12 +70,12 @@ public async Task GetsInitialisedToTheProperValue(bool permissionGranted)
         public sealed class TheRequestAccessAction : NotificationSettingsViewModelTest
         {
             [Fact, LogIfTooSlow]
-            public async Task OpensAppSettings()
+            public async Task RequestThePermission()
```
dshr: Can we rephrase it to this?
```suggestion
            public async Task RequestsPermission()
```

PR #6850, opened 2019-12-23
Project: Reports Rewrite

kornelijepetak:
## What's this?
Well, these are the reports. The new ones.

### Relationships
📊 Reports Rewrite project

## Why do we want this?
Faster and nicer reports.

## How is it done?
Well, by lots of efforts.

## Review hints
Make sure to functionally test this.

## :squid: Permissions
Me.


pdrj:
> I noticed one weird thing... are we always flooring the numbers? Is that a bug or is that by design:
> image

@simonrozsival I've checked the code and indeed, we cast to `int` on both platforms, which results in rounding down. The fact that the numbers don't sum to `100` looks weird to me as well. What do you think should be done here? I don't think we can just round one of them up - this would be very misleading if we had 9 entries of `11.1%` or other entry set with equal fractional parts. We could maybe display one fractional digit (it still might not sum to `100` though)? 

As far as I'm concerned, I'd leave it as it is, if we can't find an elegant solution.

Any thoughts @dshr or @kornelijepetak?

simonrozsival:
My suggestion: https://toggl.slack.com/archives/C02APMYTM/p1577800409288000

kornelijepetak:
> 2. Looks like the only way to refresh the reports page is to change the date range or kill and re-launch the app. I'm not sure if we are ok with this behavior or if there are some technical limitations but it's a bit confusing to stop some TE, switch to reports to see how it impacted my stats and there is nothing updated

I've opened #6879 and it waits for @maritoggl and/or @amulware to chime in.

PR #6851, opened 2019-12-23
Use native HTTP client handlers

### LINKED ISSUE #3588: Proxy Support

### LINKED ISSUE #6865: 🔍 Consider enabling proxy for Android
Given https://github.com/toggl/mobileapp/issues/3588 and https://github.com/toggl/mobileapp/pull/6851 we might be able to monitor our requests with https://www.charlesproxy.com/ 

The only issue is that Android is surprisingly more secure than iOS and you need to specifically enable trust in the app: 

> As of Android N, you need to add configuration to your app in order to have it trust the SSL certificates generated by Charles SSL Proxying. This means that you can only use SSL Proxying with apps that you control.
> 
> In order to configure your app to trust Charles, you need to add a Network Security Configuration File to your app. This file can override the system default, enabling your app to trust user installed CA certificates (e.g. the Charles Root Certificate). You can specify that this only applies in debug builds of your application, so that production builds use the default trust profile.

- https://www.charlesproxy.com/documentation/using-charles/ssl-certificates/

We need to decide if we want to enable this in debug builds only or everywhere.

simonrozsival:
## What's this?
This PR _finally_ adds support for native HTTP client handlers. We've been using the managed `HttpClient` up until now, which caused problems with VPNs and proxies and it made it impossible to intercept networking traffic using tools like Charles (https://www.charlesproxy.com/documentation/ios/).

### Relationships
Closes #3588, closes #6865

## How is it done?
Dependency container -> inject the correct HTTP client based on app platform ~and production/staging server~.

### Side effects
~We (must?) use the managed clients for the staging server (because of the self-signed certificate).~ _I figured this part out, we can use the native handlers even for staging and it's easy to integrate it with Proxyman (or Charles) for easy networking debugging._

## Review hints
We need to test this:
- [ ] against staging on iOS (including Siri and Widget)
- [ ] against staging on Android (including Widget)
- [ ] against production on iOS (including Siri and Widget)
- [ ] against production on Android (including Widget)

To test the app against the production API (toggl.com):
- in VS4Mac, right click on `Toggl.iOS` or `Toggl.Droid` project, go to `Options`
- in the `Compiler` tab, add `;USE_PRODUCTION_API` to the `Define Symbols` input
- save with OK
- rebuild the app

Make sure the [staging certificate is installed AND trusted](https://github.com/toggl/mobileapp/blob/develop/docs/certificate.md) on the device or simulator/emulator you use for testing. You can try intercepting the HTTP requests with Proxyman - I added the docs in this PR as well so check the MD file.

On iOS you can test the TestFlight build 2080101 (2.8.1) to test against production.

## Tests
- ✅ Sync tests: https://app.bitrise.io/build/b3355a87512e8f0f
- ✅ Integration tests: https://app.bitrise.io/build/25ec0a1224009145

## :squid: Permissions
:squid:


Toggl.Droid/Properties/AndroidManifest.xml
```diff
@@ -1,18 +1,13 @@
-<?xml version="1.0" encoding="utf-8"?>
-<manifest
-    xmlns:android="http://schemas.android.com/apk/res/android"
-    android:versionCode="987654321"
-    android:versionName="2.8.1"
-    package="com.toggl.giskard.debug"
-    android:installLocation="auto">
+﻿<?xml version="1.0" encoding="utf-8"?>
+<manifest xmlns:android="http://schemas.android.com/apk/res/android" android:versionCode="987654321" android:versionName="2.8.1" package="com.toggl.giskard.debug" android:installLocation="auto">
```
semanticer: Do we want to lose the formating here?

Toggl.Droid/Properties/AndroidManifest.xml
```diff
@@ -1,18 +1,13 @@
...
 	<uses-sdk android:minSdkVersion="21" android:targetSdkVersion="29" />
 	<uses-permission android:name="android.permission.INTERNET" />
 	<uses-permission android:name="android.permission.VIBRATE" />
 	<uses-permission android:name="android.permission.RECEIVE_BOOT_COMPLETED" />
 	<uses-permission android:name="android.permission.READ_CALENDAR" />
 	<uses-permission android:name="android.permission.WAKE_LOCK" />
 	<uses-permission android:name="android.permission.FOREGROUND_SERVICE" />
-	<application android:allowBackup="false" android:icon="@mipmap/ic_launcher" android:label="Toggl for Devs">
+	<application android:allowBackup="false" android:icon="@mipmap/ic_launcher" android:label="Toggl for Devs" android:networkSecurityConfig="@xml/network_security_config">
```
semanticer: `network_security_config` - I think we used PascalCase for other XML resources because C# or something? But I'm not sure about the decision process here.

Toggl.Droid/Properties/AndroidManifest.xml
simonrozsival: I'm not sure why this happened 🤔 I'll fix it.

Toggl.Droid/Properties/AndroidManifest.xml
simonrozsival: For some reason it doesn't work when I rename i to PascalCase. I think we can leave it like this 🤷‍♂

PR #6852, opened 2019-12-23
Fix selecting projects with the same name but different clients

### LINKED ISSUE #6568: Selecting projects with the same name but different clients doesn't work
If you are adding an existing project (same description), it still shows the "add new project" hint. (it's not possible to create it though)

If the two projects have the same name, but different clients, clicking on the project during the autocomplete search does nothing.

How to replicate:
- Go to the start time entry view
- Create a project "X" without a client
- Create another project "X" but with a client
- Go back to the start time entry view
- Click on the project icon in the bottom bar
- Search for "X"
- Select the first project "X" without a client
- Click on the project icon in the bottom bar again
- Search for "X" again
- Try clicking on the project "X" that has a client
- There you go.

semanticer:
## What's this?
This is fixing the corner case where the user was not able to select a different project with the same name.

### Relationships

Closes #6568

## Why do we want this?
We want to give users the opportunity to mess with their projects however they want. (well almost)

## How is it done?
We had a condition that prevents us from setting the same `Spannable` again to avoid event loops. The issue was that this condition was only comparing the results of `toString()` and those are the same if you have a project with the same name. I solved this by providing proper `Equals()` to all implementations of `ISpan` and then creating a method that compares all ISpans in a spannable. This approach takes into account different ProjectSpan properties, thus it allows the update to proceed.

### Why not another way?
Please let me know if you find a better way. Keep in mind that the span classes in Android lack equals and hashCode methods and there is also a bug in the SpannableStringBuilder.equals() 🤦‍♂️

### Side effects
Shouldn't be any

## Review hints
Check if the equals implementations are correct and some functional test can't hurt.

## :squid: Permissions
Anyone, anytime


Toggl.Droid/Autocomplete/AutocompleteExtensions.cs
```diff
@@ -19,6 +19,17 @@ public static class AutocompleteExtensions
         public static IImmutableList<ISpan> AsImmutableSpans(this ICharSequence text, int cursorPosition)
             => text.AsSpans(cursorPosition).ToImmutableList();
 
+        public static bool isEqualTo(this ISpannable first, ISpannable second)
+        {
+            if (first.ToString() != second.ToString()) return false;
```
kornelijepetak: control-flow blocks should go to a new line
```suggestion
            if (first.ToString() != second.ToString()) 
                return false;
```

Toggl.Droid/Autocomplete/AutocompleteExtensions.cs
```diff
@@ -19,6 +19,17 @@ public static class AutocompleteExtensions
...
+
+            var firstSpans = first.AsImmutableSpans(0);
+            var secondSpans = second.AsImmutableSpans(0);
+
+            if (firstSpans.Count != secondSpans.Count) return false;
```
kornelijepetak: control-flow blocks should go to a new line

Toggl.Droid/Autocomplete/AutocompleteExtensions.cs
```diff
@@ -19,6 +19,17 @@ public static class AutocompleteExtensions
...
+            return firstSpans.Zip(secondSpans, (f, s) => f.Equals(s)).All((isEqual) => isEqual);
```
kornelijepetak:
```suggestion
            return firstSpans.Zip(secondSpans, (f, s) => f.Equals(s)).All(isEqual => isEqual);
```

The parentheses are not needed. Even better, maybe use:

```cs
            return firstSpans.Zip(secondSpans, (f, s) => f.Equals(s)).All(CommonFunctions.Identity);
```

Also, please format linq operators each in a new line (except maybe the first one).

Toggl.Core/Autocomplete/Span/TextSpan.cs
```diff
@@ -8,5 +8,22 @@ public TextSpan(string text)
...
+        }
+
+        public override bool Equals(object obj)
+        {
+            if (ReferenceEquals(null, obj)) return false;
+            if (ReferenceEquals(this, obj)) return true;
+            return obj.GetType() == GetType() && Equals((TextSpan) obj);
+        }
+
+        public override int GetHashCode()
```
kornelijepetak: There is no need to do these things manually, just return
```cs
public override int GetHashCode() => HashCode.Combine(Text);
```

Toggl.Core/Autocomplete/Span/TextSpan.cs
```diff
@@ -8,5 +8,22 @@ public TextSpan(string text)
         {
             Text = text;
         }
+
+        public bool Equals(ISpan other)
```
kornelijepetak: The equality of this class can be simplified;

```cs
public bool Equals(ISpan other)
    => other?.GetType() == GetType()
    && other is TextSpan otherSpan
    && otherSpan.Text == Text;

public override bool Equals(object obj)
    => Equals(obj as ISpan);
```

If `obj` in `Equals(object)` is `null`, it will be passed to `Equals(ISpan` as null, which will return false on `other?.GetType() == GetType()`. If it's not `null` but is not a `ISpan`, `as` operator will again pass `null` to `Equals(ISpan)` causing the same behavior.
Hence, only `TextSpan`s will pass the conditions (and only those with the same text).

Consider also whether the exact type check is needed. This depends on whether you want the inherited classes to also satisfy the condition or not.

For example, `QueryTextSpan` inherits from `TextSpan`. If you always want to consider them different, regardless of their `Text` representation, then `other?.GetType() == GetType()` is not strictly needed because `is` operator will pass thru even instances of inherited classes. But if they should always be different, then leave the exact type comparison.

Toggl.Core/Autocomplete/Span/TagSpan.cs
```diff
@@ -18,5 +19,20 @@ public TagSpan(IThreadSafeTag tag)
             : this(tag.Id, tag.Name)
         {
         }
+
+        public bool Equals(ISpan other)
```
kornelijepetak: Similarly, as described for `TextSpan`, this can be simplified to:

```cs
public bool Equals(ISpan other)
    => other is TagSpan otherSpan
    && otherSpan.TagId == TagId
    && otherSpan.TagName == TagName;

public override bool Equals(object obj)
    => Equals(obj as ISpan);
```

Note that the `is` operator will always return `false` for `null`, regardless of the containing variable's type.

Toggl.Core/Autocomplete/Span/TagSpan.cs
```diff
@@ -18,5 +19,20 @@ public TagSpan(IThreadSafeTag tag)
...
+        {
+            return other is TagSpan otherTagSpan && TagId == otherTagSpan.TagId && TagName == otherTagSpan.TagName;
+        }
+
+        public override bool Equals(object obj)
+        {
+            return ReferenceEquals(this, obj) || obj is TagSpan other && Equals(other);
+        }
+
+        public override int GetHashCode()
```
kornelijepetak: Turn this into an expression-bodied member.

Toggl.Core/Autocomplete/Span/ProjectSpan.cs
```diff
@@ -44,5 +45,25 @@ private ProjectSpan(IThreadSafeProject project, IThreadSafeTask task)
             : this(task.ProjectId, project.Name, project.Color, task.Id, task.Name)
         {
         }
+
+        public bool Equals(ISpan other)
```
kornelijepetak: Similarly to the `TagSpan` and `TextSpan`, make `Equals(object)` just a proxy for proper typed equality. I leave this as an excercise to you 😄

Toggl.Droid/Autocomplete/AutocompleteExtensions.cs
```diff
@@ -19,6 +19,17 @@ public static class AutocompleteExtensions
         public static IImmutableList<ISpan> AsImmutableSpans(this ICharSequence text, int cursorPosition)
             => text.AsSpans(cursorPosition).ToImmutableList();
 
+        public static bool isEqualTo(this ISpannable first, ISpannable second)
```
kornelijepetak: Don't you `camelCase` me! 😋 

```suggestion
        public static bool IsEqualTo(this ISpannable first, ISpannable second)
```

Toggl.Droid/Autocomplete/AutocompleteExtensions.cs
semanticer: `CommonFunctions.Identity` looks cool but seems less readable to non-C# guy like me so I would still prefer the original solution if there is no big argument against it.

Toggl.Core/Autocomplete/Span/TextSpan.cs
semanticer: 👍

Toggl.Droid/Autocomplete/AutocompleteExtensions.cs
semanticer: 👍

Toggl.Core/Autocomplete/Span/TextSpan.cs
semanticer: I guess it makes more sense to distinguish between `TextSpan` and `QueryTextSpan` for now so I will use the explicit check 👍

Toggl.Core/Autocomplete/Span/TagSpan.cs
semanticer: 👍

Toggl.Core/Autocomplete/Span/TagSpan.cs
semanticer: 👍

Toggl.Droid/Autocomplete/AutocompleteExtensions.cs
semanticer: Just found out `CommonFunctions.Identity` is not some C# `deep knowledge` but actually part of our codebase that we use normally.  I guess I will use it after all.

PR #6853, opened 2019-12-24
⌨ Reports: re-enable Handoff and Keyboard shortcuts

### LINKED ISSUE #6848: Integrate with Handoff and the new KB shortcuts
Handoff and the KB shortcuts relied on the old reports structure. Change their code to work with the new reports view.

dshr:
# What's this?
A set of changes that makes Handoff and the new KB shortcuts work with the new reports

### Relationships
Closes #6848 

## Why do we want this?
✨ 

## How is it done?
By reconnecting the appropriate observables

## :squid: Permissions
📊 folks only


Toggl.iOS/ViewControllers/ReportsViewController.KeyCommands.cs
```diff
@@ -12,6 +14,14 @@ public partial class ReportsViewController
         protected override void ConfigureKeyCommands()
         {
             base.ConfigureKeyCommands();
+            AddKeyCommand(shortcutCommand("1", Resources.Today));
```
kornelijepetak: I'd rather you send in the `DateRangePeriod` value instead of the title. See other relevant comments.

Toggl.iOS/ViewControllers/ReportsViewController.KeyCommands.cs
```diff
@@ -27,6 +37,15 @@ protected override void ConfigureKeyCommands()
             modifierFlags: UIKeyModifierFlags.Command,
             propertyList: null);
 
+        private UIKeyCommand shortcutCommand(string input, string title)
+            => UIKeyCommand.Create(
+            title: title,
```
kornelijepetak: If `DateRangePeriod` is sent in as an argument instead of a string `title`, just use
```suggestion
            title: period.ToHumanReadableString(),
```

Toggl.iOS/ViewControllers/ReportsViewController.KeyCommands.cs
```diff
@@ -27,6 +37,15 @@ protected override void ConfigureKeyCommands()
...
+            image: null,
+            action: new Selector("selectShortcut:"),
+            input: input,
+            modifierFlags: UIKeyModifierFlags.Command | UIKeyModifierFlags.Alternate,
+            propertyList: null);
```
kornelijepetak: If `DateRangePeriod` is sent in as an argument instead of a string `title`, add the enum value as a property here.

Toggl.iOS/ViewControllers/ReportsViewController.KeyCommands.cs
```diff
@@ -46,5 +65,22 @@ private void selectDateRange()
...
         }
+
+        [Export("selectShortcut:")]
+        private void selectShortcut(UIKeyCommand command)
+        {
+            var title = command.Title;
+            if (title == null)
+                return;
+
+            var shortcut = IosDependencyContainer.Instance.DateRangeShortcutsService.Shortcuts.FirstOrDefault(s => s.Text == title);
```
kornelijepetak: Instead of relying on text, I'd rely on the value of `DateRangePeriod` value, taken from `command.PropertyList`, which has previously been set to an appropriate value on `:47`

PR #6862, opened 2019-12-27
Cache calendar layout attributes

### LINKED ISSUE #2914: Optimizations in calendar collection view layout

dshr:
## What's this?
A set of changes to introduce caching to out layout attributes.

### Relationships
Closes #2914

## Why do we want this?
✨ performance ✨

## How is it done?
By creating two dicts and storing data in them, and purging that data when needed.

## :squid: Permissions
Go for it!


Toggl.iOS/Views/Calendar/CalendarCollectionViewLayout.cs
```diff
@@ -196,29 +204,40 @@ public override UICollectionViewLayoutAttributes[] LayoutAttributesForElementsIn
...
+            if (supplementaryViewLayoutAttributes.ContainsKey(kind)
+                && supplementaryViewLayoutAttributes[kind].ContainsKey(indexPath))
+                return supplementaryViewLayoutAttributes[kind][indexPath];
+
             if (kind == HourSupplementaryViewKind)
             {
                 var attributes = UICollectionViewLayoutAttributes.CreateForSupplementaryView(kind, indexPath);
                 attributes.Frame = frameForHour((int)indexPath.Item);
                 attributes.ZIndex = 0;
+                supplementaryViewLayoutAttributes[kind] = new Dictionary<NSIndexPath, UICollectionViewLayoutAttributes> { [indexPath] = attributes };
```
juanlaube: This invalidates all the cached attributes for this kind. I would extract this to a method that checks if the cache for this kind exists and either save the attributes in that cache or create a new one.

Toggl.iOS/Views/Calendar/CalendarCollectionViewLayout.cs
```diff
@@ -196,29 +204,40 @@ public override UICollectionViewLayoutAttributes[] LayoutAttributesForElementsIn
...
                 return attributes;
             }
             else if (kind == EditingHourSupplementaryViewKind)
             {
                 var attributes = UICollectionViewLayoutAttributes.CreateForSupplementaryView(kind, indexPath);
                 attributes.Frame = frameForEditingHour(indexPath);
                 attributes.ZIndex = 200;
+                supplementaryViewLayoutAttributes[kind] = new Dictionary<NSIndexPath, UICollectionViewLayoutAttributes> { [indexPath] = attributes };
```
juanlaube: Same thing here.

PR #6875, opened 2020-01-03
✨ Add animation around play fab

### LINKED ISSUE #5795: Add long press hint on Android
The start time entry button on the main view has two modes: normal mode as defined by the user preferences when tapping and alternate mode when long pressing.

The problem is this feature is not really easy to find unless you know it's there. We we'll probably need to add some kind of animation to make it clear to the user.

There's a current discussion here: https://toggl.slack.com/archives/CMB9FDE5D/p1565778022020100

pdrj:
## What's this?
This adds animated circle around the play fab.

### Relationships

Closes #5795

## Why do we want this?
To better indicate that long pressing the fab has a side effect.

## How is it done?
Extending `FloatingActionButton` and some custom drawing.

### Why not another way?
I wanted to implement it using animated vector drawables, but couldn't get the right animation out of these. To not waste more time, I chose this approach.

## Review hints
Give the fab a long press, use a thumb that is not too big.

This is how completed animation looks like:
Screenshot 2020-01-03 at 15 18 02

## :squid: Permissions
Anyone.


Toggl.Droid/Views/FloatingActionButtonAnimation.cs
```diff
@@ -0,0 +1,27 @@
...
+
+        public FloatingActionButtonAnimation(AnimatedFloatingActionButton animatedFloatingActionButton)
+        {
+            targetAngle = fullCircle + animatedFloatingActionButton.AngleStartPoint;
+            this.animatedFloatingActionButton = animatedFloatingActionButton;
+        }
+
+        protected override void ApplyTransformation(float interpolatedTime, Transformation t)
+        {
+            if (interpolatedTime < animationStartDelayThreshold) return;
```
kornelijepetak: Do not put blocks in the same line as flow-control statements.
```suggestion
            if (interpolatedTime < animationStartDelayThreshold) 
                return;
```

Toggl.Droid/Views/FloatingActionButtonAnimation.cs
```diff
@@ -0,0 +1,27 @@
+using Android.Views.Animations;
+
+namespace Toggl.Droid.Views
+{
+    public class FloatingActionButtonAnimation : Animation
```
kornelijepetak: I think it'd be easier to use `ValueAnimator.ofFloat(0f, 100f)`
and use `setStartDelay` when setting the animator up.

Then this whole class can be dropped by using:

```cs
animator = ValueAnimator.OfFloat(0, 360);
animator.SetDuration(Math.Max(ViewConfiguration.LongPressTimeout, minAnimationDurationMs));
animator.StartDelay = 500; // or whatever the correct initial delay is
animator.Update += (_, args) => fab.SweepAngle = (float)args.Animation.AnimatedValue;

// when needed:
animator.Start();
```

This code also assumes that the extended FAB itself should take care of properly applying the StartingAngle. Using `animatedFloatingActionButton.AngleStartPoint` is not the job of the animator. Handle it inside of the FAB.

Toggl.Droid/Views/AnimatedFloatingActionButton.cs
```diff
@@ -0,0 +1,94 @@
...
+
+        public AnimatedFloatingActionButton(Context context, IAttributeSet attrs)
+            : this(context, attrs, 0)
+        {
+        }
+
+        public AnimatedFloatingActionButton(Context context, IAttributeSet attrs, int defStyle)
+            : base(context, attrs, defStyle)
+        {
+            paint = new Paint { Flags = PaintFlags.AntiAlias };
```
kornelijepetak: Are you sure this is the only constructor that's ever (and always) be called? Maybe this setup belongs to an isolated method called from all constructors? I'm not exactly sure how mono-java interop handles the constructors in this regard. Please check this with @heytherewill, unless, of course, you are completely sure yourself.

Toggl.Droid/Views/AnimatedFloatingActionButton.cs
```diff
@@ -0,0 +1,94 @@
...
+            paint.SetStyle(Paint.Style.Stroke);
```
kornelijepetak: I think this one is not needed (not 100% sure), because the `Stroke` is a default Paint style. 🤔

Toggl.Droid/Views/AnimatedFloatingActionButton.cs
```diff
@@ -0,0 +1,94 @@
...
+            drawingRect.Top = -animationPaddingPx;
+            drawingRect.Left = -animationPaddingPx;
+
+            drawingRect.Right = w + animationPaddingPx;
+            drawingRect.Bottom = h + animationPaddingPx;
+        }
+
+        public override void Draw(Canvas canvas)
+        {
+            base.Draw(canvas);
```
kornelijepetak: Since the call to `base.Draw(canvas)` is before your custom code, does that not overdraw the fab with your `canvas.DrawArc`? Am I missing what you're doing here?

Toggl.Droid/Views/AnimatedFloatingActionButton.cs
pdrj: I believe `Fill` is the default.

Toggl.Droid/Views/AnimatedFloatingActionButton.cs
pdrj: I'm not sure. I based it on what we do in other custom views. 

That being said, I think that a constructor that takes a `context` must be called at some point... But yeah, I don't have a source for that.

Toggl.Droid/Views/AnimatedFloatingActionButton.cs
pdrj: Nope, you're not missing anything. That is intentional - I decided that it looks best like that, with a slight overdraw. Do you think the animation should have a bigger inner radius?

Toggl.Droid/Views/AnimatedFloatingActionButton.cs
kornelijepetak: True, but there's also another constructor with the `Context` parameter, just without `defStyle`, so I am not sure if maybe that one is called in some other situations... 🤔

pdrj: Yeah, but it is
```
        public AnimatedFloatingActionButton(Context context, IAttributeSet attrs)
            : this(context, attrs, 0)
```
(it calls the one with `defStyle`).

Toggl.Droid/Views/FloatingActionButtonAnimation.cs
pdrj: Sure, `less code == better`, I'll change it.

Toggl.Droid/Views/AnimatedFloatingActionButton.cs
kornelijepetak: True. Somehow missed that. 
All well 😄

Toggl.Droid/Views/AnimatedFloatingActionButton.cs
kornelijepetak: I see. I've tried it now on a device.

I thought the FAB icon is also drawn before the call to this Draw. I find it weird that it does not.

All well.

PR #6883, opened 2020-01-08
🐛 Display onboarding tip only for fully visible TE log items

### LINKED ISSUE #6567: The "tap to edit" onboarding tooltip misbehaves when the initial sync fails during login
After logging out and logging in again with my account the sync initially failed causing the spider to appear.

After pulling to refresh, the time entries loaded alright, but the "Tap to edit" onboarding tooltip appeared in a very weird position, covering the running time entry card in the main log.

How to replicate:
- Log in into your account that has a bunch of time entries (it should cover a least one screen)
- Log out
- Log in again
- Pray for the initial sync to fail (or set it up in debug mode so that it always happens)
- Checkout the spidey bro
- Pull to refresh
- Wait a bit and the popup will appear in the bottom of the screen

SS for reference:
broken-onboarding

semanticer:
## What's this?
This is an attempt on fixing our weird glitchy behavior of "tap to edit" tooltip. 

This is a pretty straightforward solution for tooltips that appear for invisible TEs and tooltips that are squashing at the bottom of the bottom bar like this: 

Screenshot 2020-01-08 at 14 28 48

**It also improves the situation with the running TE layout where we had situations like:**

Screenshot 2020-01-08 at 14 43 11

**Now the worst case scenario is this:**

Screenshot 2020-01-08 at 14 26 19

### Relationships

Original idea was to close 6567 but as @daividssilverio pointed out below 👇 this is not the ultimate solution. 
 
## Why do we want this?
We want our tooltip to appear below TEs that our users can actually see and we want the tooltip itself to be displayed readable and not squashed or anything. 

## How is it done?
Previously, we were looking for the oldest **partially** visible TE log layout. I changed the logic to skip the partially visible layouts and proceed in finding some fully visible ones. 

### Why not another way?
This seems to be the most straight forward solution. We could probably improve the situation with the running TE even more by applying special cases based on the visibility of our running TE layout but I'm not sure if it's worth it to go that far since the worst-case scenario for tooltip over the running TE layout is now pretty fine I guess because you are still more or less able to see the TE layout that the tooltip is pointing at.

### Side effects
None that I know of

## Review hints
You can use https://github.com/toggl/mobileapp/tree/debug-onboarding branch if you want to see the tooltip all the time and 🍒this commit to test. 

## :squid: Permissions
Anyone


daividssilverio:
As described [here](https://github.com/toggl/mobileapp/issues/6567#issuecomment-572810774), the issue this is trying to close isn't being solved. If you are not making further changes to this PR, please do not close #6567 yet.

semanticer:
Ok, I will merge this one without closing #6567 and look further into it.

PR #6885, opened 2020-01-08
🐛 Fix the tests

### LINKED ISSUE #6594: Fix failing tests caused by hard-coded implicit culture strings
The tests are failing with one or two of the following culture-related errors:

```
Expected result to be 
"Sun, 26 Feb", but 
"ned, 26 vlj" differs near "ned" (index 0).
```
Obviously a problem with culture settings (ned is 🇭🇷 for Sunday)...

And also something like
```
The UTC time represented when the offset is applied must be between year 0 and 10,000.
```

which may not even be a problem with the culture per se, but with using `default(DateTimeOffset)` where it should not be used, but it fails after moving the default an hour or two in the past (now being an unrepresentable time).

[TestResultsReport.xml.txt](https://github.com/toggl/mobileapp/files/3859035/TestResultsReport.xml.txt)

kornelijepetak:
## What's this?
Fixes some of the culture-depending tests.

### Relationships
Closes #6594

## Why do we want this?
So that we can pass tests even on non-EN cultures.

## How is it done?
Several things:
* Using `InvariantCulture`
* Using `DiscardCultureChanges` 

### Why not another way?
Not even this is foolproof.
The problem with `DiscardCultureChanges` is that it's executed before and after specific test, changing the culture values. However, because of the asynchronous execution of tests, you can imagine the racing conditions here.

In my opinion no static elements should be used to determine the culture (if it is at all needed). Instead, we should always use `InvariantCulture` in all callsites where we are actually NOT interested in the exact user's culture and `CultureInfo.CurrentCulture` in all other places. That would make redundant the following artefacts:

* `DiscardCultureChanges`
* `DateTimeToFormattedString` (this one is particularly messy)
* Culture-based shenanigans in AppStart/AppDelegate

Also, these fixes are now solving only the tests, although I believe the tested code is actually at fault here. But since that requires systemic changes, I opted not to touch that.

### Side effects
Tests will work even on non-EN cultures.

## Review hints
Run the tests in Tunguska Culture. Or Azerbaijan. Or find another random point on the Globe which is not english-based.

## Tests
Indeed.

## :squid: Permissions
Mobile people


kornelijepetak:
@heytherewill no, I haven't created any issues for that explicitely. 
I'd have to look into the technicalities and see all the usages of such constructs across the app. Also, I'd need to find out what are the cases in which we take the offsets that we get from the API and what are the cases that we use the local culture...
And I can't promise I'll have time to do that while being a syncer.
If it can wait some time, I can do it when I return 😄

PR #6889, opened 2020-01-09
Make the calendar time entry dragging smooth

### LINKED ISSUE #6002: Improve drag-to-edit auto scrolls
Currently, when you drag to edit start/end times or offset calendar items near the edges of the screen, it's not smooth.

Let's make it smooth.

daividssilverio:
## What's this?
This smooths the calendar time entries dragging!

### Relationships
Closes #6002 

## Why do we want this?
Because it was clanky before.

## How is it done?
I've replaced the **old strategy** of using the `scroller` to trigger a scroll event when the auto scroll would be triggered with a **new strategy** that handles the auto scrolling with an animator that offsets/changes time entries and moves the scroll offset at the same time.

The dragging of time entries has also changed. Now it takes a snapshot of the event being dragged when the dragging starts and the new position of the time entry being edited is updated based on the difference between the snapshot and current user touched position on screen.

### Why not another way?
By all means, suggest something different if have something in mind.
I had to change my mindset to make this work. The old way didn't work really well.
If you think you can simplify this our the whole calendar, please speak up.

Suggestions are always welcome.

### Side effects
It's smooth.

## Review hints
Unfortunately, none.

## :squid: Permissions
🦑 it.


Toggl.Droid/Views/Calendar/CalendarDayView.cs
```diff
@@ -108,7 +114,8 @@ private void init()
             baseHourHeight = 56.DpToPixels(Context);
             hourHeightSubject = new BehaviorSubject<int>(baseHourHeight);
             maxHourHeight = 112.DpToPixels(Context);
-
+            distanceFromEdgesToTriggerAutoScroll = 8.DpToPixels(Context);
```
semanticer: Consider extracting some of the dimension constants in this file to android resources. They can be better adjusted for different devices if needed.

Toggl.Droid/Views/Calendar/CalendarDayView.EditItem.cs
```diff
@@ -90,61 +173,80 @@ private void calculateBottomDragRect(CalendarItemEditInfo activeItemEditInfo)
...
+            => currentTouchY + scrollOffset < topAreaTriggerLine
+               && scrollOffset > 0
+               && itemInEditModeRect.Top > 0;
+
+        private bool shouldAutoScrollDown()
+            => currentTouchY + scrollOffset > bottomAreaTriggerLine
+               && scrollOffset < maxHeight - Height
+               && itemInEditModeRect.Bottom < maxHeight;
+
+        private void changeStartTime()
```
semanticer: After changing this to a parameter-less method, it would maybe make more sense to change the name to `updateStartTime` and  `updateEndTime`?  When I first read call code for changeStartTime() I was like "change to what?"

semanticer: Maybe there is no clean way to do this but I was just thinking if you could share some code between changeStartTime and changeEndTime but there are probably too many little details different...

daividssilverio: 🤔 I'm not sure it's worth the effort. Please let me know if you have an interesting idea on how to do it. The internals of the method look similar, but they are different in what they are meddling with.

I rather not refactor that here. But I'm all up to a follow up issue or even any comments.

daividssilverio: Makes sense! 👍

Toggl.Droid/Views/Calendar/CalendarDayView.cs
daividssilverio: Alright!

Toggl.Droid/Views/Calendar/CalendarDayView.EditItem.cs
semanticer: Sure, I'm also not sure about this. Just ignore this now and if I feel some OCD urgency I will attempt to try something later  (not in this PR)

daividssilverio:
@semanticer I've addressed some of your change requests, how does it look for you?

semanticer:
@daividssilverio Looks great! And the new extension method `GetDimen` is a useful addition.

PR #6891, opened 2020-01-10
:bug: Fix empty main log and rating view

### LINKED ISSUE #6873: Main Log is empty until a TE is stopped within the app
## 🐛 Bug Report
Right after login or after starting the app when the user is already logged in, the main log is shown completely empty. Syncing doesn't fix it (even if it fetches new TEs) and the only thing so far is stopping the running TE (either on the main screen or in the calendar screen).

## Context

Build/Branch: build 2.8.1 - `2080103` from TestFlight
Phone OS: iOS
Phone model: doesn't seem to matter

For some reason it doesn't affect debug builds.

## Expected behavior
The main log is shown as usual.

## To reproduce
Try installing the build from TestFlight.

## Suggestions
No idea.

pdrj:
## What's this?
This makes items appear in the main log in release builds. It also fixes several issues with rating view on Android.

### Relationships

Closes #6873 

## Why do we want this?
Seems to improve the experience.

## How is it done?
The main log was empty because we use `CombineLatest` to produce the items and in release mode that waited for the `ShouldShowRatingView` observable to emit something. That doesn't happen straight away. To fix that, I added `StartWith(false)` to the `ShouldShowRatingView` observable.

The button to go to Play Store and rate was not working, because a `NewTask` flag was needed.

The send feedback button was crashing (on Android) due to lack of the flag as well. The flag is now needed because there is no `View` associated with `RatingViewModel` and we start the activity using application context.

### Why not another way?
🤷‍♂ 

## Review hints
Force the rate view to appear and test all the buttons.

## :squid: Permissions
Anyone.


pdrj:
Do you remember what these problems were @Odrakir, so that I can test?

Toggl.Core/Services/UpdateRemoteConfigCacheService.cs
```diff
@@ -85,4 +85,4 @@ private void onFetchFailed(Exception exception)
             }
         }
     }
-}
\ No newline at end of file
+}
```
juanlaube: 👀

Toggl.Droid/Adapters/MainLogRecyclerAdapter.cs
```diff
@@ -114,13 +115,18 @@ public override int GetItemViewType(int position)
...
                 default:
-                    throw new Exception("Invalid view type");
+                    throw new Exception($"Invalid view type: {viewType}");
             }
         }
 
         public void UpdateCollection(IImmutableList<MainLogSection> items)
         {
-            var flattenItems = items.Aggregate(ImmutableList<MainLogItemViewModel>.Empty, (acc, nextSection) => acc.AddRange(nextSection.Items.Prepend(nextSection.Header)));
+            var flattenItems = items.Aggregate(ImmutableList<MainLogItemViewModel>.Empty,
```
juanlaube: I would rather have a view with 0dp height for this section header instead of checking the type, in the future we could add more sections without a header, or decide that this one needs a header now, and it won't be easy to find this.

daividssilverio: What @juanlaube makes sense, even though it's not nice. 
If not the view with `0dp`, leave a comment here to help people understand why we were doing this.

Toggl.Core/Services/UpdateRemoteConfigCacheService.cs
pdrj: I don't know what's up with that. I know Rider adds newline on save automatically, I deleted it in vim but a change is still there 🤷‍♂.

Toggl.Droid/Adapters/MainLogRecyclerAdapter.cs
pdrj: I'm not sure about the `0dp` view approach. It seems more hacky than this one tbh - having items in adapter that won't ever be rendered.

How about having a set of classes that won't be rendered (this will now have just one element) and comparing each header's type against that set? Not sure how to achieve that in `c#` though ;).

daividssilverio:
Hint for reviewers: to replicate the issue, remove the changes made and make the change `RatingViewExperiment` with

```
            => Observable.Never<bool>();
```

You'll be able to replicate the issue. Nice catch @pdrj .

Toggl.Droid/Adapters/MainLogRecyclerAdapter.cs
pdrj: I mean something like:
```            
var nonRenderableHeaderTypes = new List<Type>()
            {
                typeof(UserFeedbackSectionViewModel)
            };
```

Toggl.Core/Services/UpdateRemoteConfigCacheService.cs
pdrj: Did a checkout from develop. It's no longer there.

dshr:
+1 on @Odrakir's concern about feeding drivers into other things feels off. Perhaps we can have an private `shouldShowRatingViewObservable`, that then goes `ShouldShowRatingView = shouldShowRatingViewObservable.AsDriver(schedulerProvider)`? Everything else is on point though 👌

Toggl.Core.UI/ViewModels/MainViewModel.cs
```diff
@@ -84,7 +85,7 @@ public sealed class MainViewModel : ViewModel
         public IObservable<bool> ShouldShowRunningTimeEntryNotification { get; private set; }
         public IObservable<bool> ShouldShowStoppedTimeEntryNotification { get; private set; }
         public IObservable<IThreadSafeTimeEntry> CurrentRunningTimeEntry { get; private set; }
-        public IObservable<bool> ShouldShowRatingView { get; private set; }
+        public IObservable<bool> ShouldShowRatingViewDriver { get; private set; }
```
dshr: All good, but we don't suffix name public observables with `Driver`, just the `ShouldShowRatingView` is more than enough.

pdrj: It was to make a clear distinction that there is now `*Observable` and `*Driver`. I'll rename it :).

dshr:
Done, let's 🚢 🇮🇹 !

Odrakir:
@pdrj I don't remember the specifics. I think it was related to running tests in parallel, but we never found out the exact problem. We just learned not to use drivers for anything else.

PR #6892, opened 2020-01-10
Clean up onboarding storage on logout

### LINKED ISSUE #6598: Onboarding steps are not appearing after signups after logout
How to replicate:
- Log in once, make sure you don't see the onboarding steps
- Log out
- Sign up with a new account
- You shouldn't see the main onboarding steps

What should be working:
- [x] Check that steps are being displayed correctly for first entry
- [x] Check that steps to edit first entry are being displayed correctly

daividssilverio:
## What's this?
This fixes the misbehaving onboarding steps during logout/new signups.

### Relationships
Closes #6598 

## Why do we want this?
Because you never know when some user is going to show toggl to a friend, maybe he/she logs out and helps his/her friend to create an account using the mobile apps. This new user should still see the onboarding steps o/

## How is it done?
By resetting the onboarding storage during the logout process. I'm not completely sure why we were not doing that already. Maybe @maritoggl or @amulware or even @simonrozsival knowns something that I don't.

### Why not another way?
I'm not sure why we are not clearing the whole private key/value storage during logout.
We could also do that. Maybe there's a historic reason for that.

### Side effects
There shouldn't be any.

## Review hints
Try to replicate the problem first in another branch, then check if it's fixed here.

## :squid: Permissions
🦑 it


maritoggl:
Yes, the new user should still see the onboarding steps. With that said, if the old user logs in again, they should not see the onboarding steps. Please confirm that's the approach you've taken as well?

daividssilverio:
@maritoggl , yeah, that part still works as it worked before. There's only two onboarding steps that still appear for old users. The `AddProjectOrTagOnboardingStep` and `CategorizeTimeUsingProjectsOnboardingStep`.

PR #6893, opened 2020-01-10
Recreate tap to edit onboarding tooltip when null

### LINKED ISSUE #6599: Tap to edit onboarding step isn't appearing at all
How to replicate:
- Create a new account
- Go through the two first onboarding steps:
  - "Tap to start time entry"
  - "Tap to stop time entry"
- The "Tap to edit time entry" should appear then

PS: If you tap the time entry, you should see the "Categorize your time by adding a project" step appears alright.

daividssilverio:
## What's this?
This fixes the onboarding step to teach users how to tap to edit time entries.

### Relationships
Closes #6599 

## Why do we want this?
Because we do want to show the onboarding step to teach users how to edit time entries.

## How is it done?
It wasn't appearing because the popup tooltip was being set `onStop` to null and never recreated when it was time to show it.
Now the popup is recreated if it's null when it's time to show it.

### Why not another way?

### Side effects
None really.

## Review hints
Follow the replication instructions in the original issue.
It's better if you test this after #6892 is merged.

## :squid: Permissions
:squid: it


Toggl.Droid/Fragments/MainFragment.Onboardings.cs
```diff
@@ -143,6 +141,10 @@ private void setupTapToEditOnboardingStep()
         private void updateTapToEditOnboardingStep(MainLogCellViewHolder oldestVisibleTimeEntryViewHolder)
         {
             tapToEditPopup?.Dismiss();
+            if (tapToEditPopup == null)
```
kornelijepetak: A shorter, more C#-ian version of this would be: 

```cs
tapToEditPopup ??= createTapToEditPopup();
```

Toggl.Droid/Fragments/MainFragment.Onboardings.cs
```diff
@@ -116,11 +115,10 @@ private void setupStopTimeEntryOnboardingStep()
 
         private void setupTapToEditOnboardingStep()
         {
-            tapToEditPopup = PopupWindowFactory.PopupWindowWithText(
-                Context,
-                Resource.Layout.TooltipWithLeftTopArrow,
-                Resource.Id.TooltipText,
-                Shared.Resources.TapToEditIt);
+            if (tapToEditPopup == null)
```
kornelijepetak: A shorter, more C#-ian version of this would be: 

```cs
tapToEditPopup ??= createTapToEditPopup();
```

daividssilverio:
@kornelijepetak I did think about your suggestion. But I did not want to touch the others onboarding steps as well.

PR #6938, opened 2020-01-17
Fix the DateRangePickerViewSource.ScrollToCurrentPage crash

### LINKED ISSUE #6937: DateRangePickerViewSource.ScrollToCurrentPage: index out of range
## 💥 Crash Report
A crash that randomly happens when opening the date range selector.

### 👥 Impact
~4% of users on 2.9

### 🔗 Context
Introduced in the Reports rewrite

`SIGABRT: Specified argument was out of the range of valid values. Parameter name: index`
```cs
    Requires.FailRange (System.String parameterName, System.String message)
    ImmutableList`1+Node[T].ItemRef (System.Int32 index)
    ImmutableList`1[T].get_Item (System.Int32 index)
    DateRangePickerViewSource.ScrollToCurrentPage (System.Boolean animated)
    DateRangePickerViewController.TraitCollectionDidChange (UIKit.UITraitCollection previousTraitCollection)
```

dshr:
## What's this?
A small bit of defensive programming to help with the crash

### Relationships
Closes #6937 

## Why do we want this?
No :boom:

## How is it done?
🛡 

## :squid: Permissions
Me only


Odrakir:
[User report](https://app.intercom.io/a/apps/ayixs927/inbox/inbox/unassigned/conversations/25430632714)

dshr:
Yeah, the closing thing is a separate crash. That one is related to Handoff and som `CombineLatest` weirdness. Check Slack for more info, I was asking about it just now.

Toggl.iOS/ViewSources/DateRangePickerViewSource.cs
```diff
@@ -60,6 +60,8 @@ public void UpdateMonths(ImmutableList<DateRangePickerMonthInfo> newMonths)
 
         public void ScrollToCurrentPage(bool animated = false)
         {
+            if (months.IsEmpty) return;
```
kornelijepetak:
```suggestion
            if (months.IsEmpty) 
                return;
```

dshr: We do `return`s on the same line quite often, I even remember something about it in the code guidelines too, so I'm a bit confused about your reasoning behind this. Could you elaborate?

Odrakir: Yep, it is in the code guidelines, des is right: https://github.com/toggl/mobileapp/blob/develop/docs/style-guide.md

kornelijepetak: The original guidelines are very old. We have since introduced stylecop rules and one of them says that this should not be done. We have not updated the guidelines tho to reflect all the stylecop changes.

The reasoning is that the flow-control usually consists of a flow-control **statement** (such as `if`, `while`, `foreach`) and its **block**. Usually in the braces.
There is a visual demarcation of such a block with a single level of indentation in the _block_ line. We don't ever do this:

```cs
if(something) { someCallHere(); thisThere(); return; }
```
The first statement of the block is always in its own line and indented:

```cs
if(something)
{
	someCallHere();
	thisThere();
	return;
}
```

Why would any other block be different? You can say that it's because it's short, but that breaks the flow/format mental paradigm of flow-control syntax elements.

The braces are of course optional for a single statement, but it is still a flow-control `block`. That being said, the natural expectation for C-based languages is that if you see a flow-control statement **without braces** that you expect its block to be found in the following line and indented.

If this is the code:
```cs
if(something) return;
doSomethingElse();
doSomethingOtherThanElse();
```

speedy reading will feel that `doSomethingElse()` is the block of the if, because with no braces that's what is expected. But then of course, you get confused that it's not indented and only then you get to look at the return that's in a _somewhere unexpected_.

Well, that's at least my reasoning. I don't require anyone to agree with me. And despite the guidelines, I'll always put a return in a newline/indented. 

That being said, if you don't want to change this, I don't want my comment blocking you.

kornelijepetak: For reference:
https://github.com/toggl/mobileapp/blob/develop/Toggl.ruleset#L104
https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1107.md

Odrakir: In that case I think this went under the radar, we should make style cop follow our code style guides and no the other way around, in my opinion. No that this one specifically is a big deal.

dshr:
The fix from #6946 is now applied to this branch via a mergeback.

Toggl.iOS/ViewSources/DateRangePickerViewSource.cs
kornelijepetak: Our stylecop rules ARE the guidelines. We just had noone update the guidelines page and in my opinion it's not really necessary. `.ruleset` file in your root solution folder should be picked up by your IDE and acted upon accordingly. Then you don't need a markdown file somewhere telling you how to do it, because the IDE will warn you or prevent doing things. That's my take on these...

In any case, this didn't go under the radar because in two separate occurences, for two separate PRs (one mine, one Williams), all mobile devs were pinged to chime in on the potential changes. I just think nobody cared as much at the time.

PR #6939, opened 2020-01-17
🚧 Use sync error to trigger noWS UI

### LINKED ISSUE #6615: Revisit how we navigate to NoWorkspaceViewModel
We recently found out that we're not handling data consistency errors correctly: just after we detect that the user has no workspaces, we should navigate him/her to `NoWorkspaceViewModel`. We're not doing that properly, the user first has to go back to the main log to be redirected. The same problem would happen also for the case when the user doesn't have a default workspace.

The PR which probably caused this: https://github.com/toggl/mobileapp/pull/3226
Internal discussion in Slack: https://toggl.slack.com/archives/CQ5R2RWTS/p1574148830007800

sofokli1:
## What's this?
Instead of basing the UI presentation based on the `viewAppeard` now it is also triggered by sync errors.

### Relationships
Closes #6615

## Why do we want this?
We want to show the noWS UI without needing `viewAppeard` to be triggered (at least for iOS).

## How is it done?
very few changes

### Why not another way?
This is the simplest.

### Side effects
None that I know of

### Hint
To trigger noWS error, in `DetectNotHavingAccessToAnyWorkspaceState.cs` replace `public IObservable<ITransition> Start(IFetchObservables fetch)` implementation with `=> Observable.Throw<ITransition>(new NoWorkspaceException());`

To trigger noDefaultWS error, in `DetectUserHavingNoDefaultWorkspaceSetState.cs` replace `public IObservable<ITransition> Start()` implementation with `=> Observable.Return(NoDefaultWorkspaceDetected.Transition());`

## :squid: Permissions
:squid:


Toggl.Core.UI/ViewModels/MainViewModel.cs
```diff
@@ -195,6 +196,9 @@ public override async Task Initialize()
             SyncProgressState = syncManager.ProgressObservable
                 .AsDriver(schedulerProvider);
 
+            SyncErrors = syncManager.Errors
```
daividssilverio: Why not unify this `SyncErrors` initialization with what is being done on L296?
Why is it public? Just for tests?

Toggl.Core.UI/ViewModels/MainViewModel.cs
```diff
@@ -537,5 +547,13 @@ private void postAccessibilityAnnouncementAboutSync(SyncProgress syncProgress)
 
             accessibilityService.PostAnnouncement(message);
         }
+
+        private async Task<Unit> handleSyncError(Exception exception)
+        {
+            await handleNoWorkspaceState();
+            handleNoDefaultWorkspaceState();
```
daividssilverio: why is `handleNoWorkspaceState` awaited and `handleNoDefaultWorkspaceState` isn't?

sofokli1: we do not want to choose a default WS if we are not done with the noWS case first. does this make sense?

Toggl.Core.UI/ViewModels/MainViewModel.cs
sofokli1: public is for the tests. will unify the other part

sofokli1: we might not even need `SyncErrors` for the tests

Toggl.Core.UI/ViewModels/MainViewModel.cs
daividssilverio: What you said does make sense.
My question is more about the second method not being awaited, it's a fire and forget thing?

Toggl.Core.UI/ViewModels/MainViewModel.cs
```diff
@@ -360,7 +366,15 @@ public override void ViewDisappeared()
         public override void ViewAppeared()
         {
             base.ViewAppeared();
-            SuggestionsViewModel.ViewAppeared();
```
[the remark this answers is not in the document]
sofokli1: not sure what happened but now it is fixed

Toggl.Core.UI/ViewModels/MainViewModel.cs
sofokli1: I just copied it from the `viewAppeardAsync` method. Not sure why we should await it. Do you think we need to await it?

daividssilverio: Not sure either. It doesn't seem harmful at least.

PR #6941, opened 2020-01-17
Track calendar swipe events on droid

### LINKED ISSUE #6907: Track calendar pages swipes on Android
We just need to do the Android side of #5995 .
We need this data to better understand the calendar usage patterns on Android as well.

daividssilverio:
## What's this?
This adds the missing events tracking for the calendar swipes on android.

### Relationships
Closes #6907 

## Why do we want this?
Because we forgot to add it :/

## How is it done?
Very similarly to how iOS is done.

### Why not another way?

### Side effects
🤔 The clicks on the top week-stripe day picker also track events. Is that alright @amulware ?

## Review hints
Check the logic added to track the events, double check if everything is alright.

## :squid: Permissions
🦑 it after the draft is turned into a PR. I still have to double-check the events.


amulware:
> 🤔 The clicks on the top week-stripe day picker also track events. Is that alright @amulware ?

What kind of events? Swipe events? That doesn't make sense to me. Other events?  Perhaps that's how it should be. There should be documentation in the related project on how the events should work.

daividssilverio:
@amulware I didn't see anything in the project docs explaining if the top date picker should trigger or not the swipe events.

> Swipe events? That doesn't make sense to me

Yeah... I thought so... but it's gonna get cryptic... 
Just to confirm, how is iOS handling this?  @arturs105 @sofokli1

daividssilverio:
Well, the changes were made in the latest commit.
Android won't trigger analytics events when the user clicks or swipes on the top week date picker.

arturs105:
iOS does not track any events on top week picker. It only tracks the `AnalyticsService.CalendarSingleSwipe` after the swipe is finished (that's done inside `IUIPageViewControllerDelegate.DidFinishAnimating`).

sofokli1:
@daividssilverio 
>> I didn't see anything in the project docs explaining if the top date picker should trigger or not the swipe events.

it should not trigger, I remember talking about this with product at some point. only user manual swipe.

daividssilverio:
@sofokli1 thanks for the clarification. I think we are safe to merge this PR now.

PR #6942, opened 2020-01-20
Refactor calendar settings

### LINKED ISSUE #6130: Calendar: Refactor the calendar settings into 1 VM
merge `CalendarSettingsViewModel` and `SelectUserCalendarsViewModelBase` into one VM

### LINKED ISSUE #6361: Make the calendar settings permission's flow more similar in both platforms
Please refer to https://toggl.slack.com/archives/C02APMYTM/p1572254416278900 for more details.

The idea is to replace the "Link calendars" switch in the calendar settings on droid to behave similar to iOS.

- Change the text to something similar to iOS.
- The initial permissions' flow should be the same, the user can give permissions directly from the app until the permissions are granted or permanently denied.
- After the permissions were given/permanently denied, a message stating the state of the permissions should appear as a helper text. And clicking (somewhere) should lead to the system settings so that the user can grant/revoke permissions manually.

arturs105:
## What's this?
Started working on #6361 and it became clear that the view model was quite a mess due to it being separated in 2 classes, so ended up refactoring it into a single class and closing 2 issues with 1 ~stone~ PR.

### Relationships

Closes #6361
Closes #6130

## Why do we want this?
Better permission requesting flow & cleaner code.

## How is it done?
The calendar permission flow works like this:
* The permission is asked when user opens calendar settings or calendar view for the very first time
* The "Link Calendars" setting is no longer tied to the calendar permission, instead it is just a stored setting. Though, it is a little tied to the calendar permission since if it is not granted, the setting can't be on.
* If the permission is granted then turning the setting on/off only changes the corresponding value in `UserPreferences`
* If the permission is not granted then turning on the setting:
  * 🍏 Takes the user to the Settings.app where user can enable the permission, since it either was denied or revoked through Settings.app
  * 🤖 If possible, shows the permission request dialog. If not (user has checked "Do not ask again"), takes the user to system settings where the permission can be turned on.

### Why not another way?
This is what was decided. Check #6361 for the full discussion.

## :squid: Permissions
⛔️ Still WIP


sofokli1:
@arturs105 if I do a fresh install and go to cal, asks for permissions and I approve. It opens the cal settings and the `allow calendar access` switch is on.

but if I do a fresh install and open settings, go to cal settings. Asks for permissions and I approve. It does not enable the `allow calendar access` switch.

is this in purpose?

arturs105:
@sofokli1 This is not on purpose. Thanks for finding this out

Toggl.Core.UI/ViewModels/Calendar/CalendarViewModel.cs
```diff
@@ -104,6 +118,43 @@ public sealed class CalendarViewModel : ViewModel
...
+                onboardingStorage.SetCalendarPermissionWasAskedBefore();
+            }
+        }
+
+        private void onCalendarPermission(bool granted)
+        {
+            if (granted)
+            {
+                userPreferences.SetCalendarIntegrationEnabled(true);
+                Navigate<IndependentCalendarSettingsViewModel>();
```
daividssilverio: The view responsible for this VM on android isn't done, right? Was that intended?

Toggl.Core.UI/ViewModels/Settings/CalendarSettingsViewModel.cs
```diff
@@ -42,89 +58,160 @@ public sealed class CalendarSettingsViewModel : SelectUserCalendarsViewModelBase
...
+        private async Task toggleCalendarIntegration()
+        {
+            //Disabling the calendar integration
+            if (userPreferences.CalendarIntegrationEnabled())
             {
-                calendarListVisible = false;
+                userPreferences.SetCalendarIntegrationEnabled(false);
+                userPreferences.SetEnabledCalendars(null);
+                selectedCalendarIds.Clear();
+                reloadCalendars();
```
daividssilverio: why isn't this awaited?

Toggl.Core.UI/ViewModels/Settings/CalendarSettingsViewModel.cs
```diff
@@ -42,89 +58,160 @@ public sealed class CalendarSettingsViewModel : SelectUserCalendarsViewModelBase
...
-                {
-                    calendarListVisible = true;
-                }
+                var permissionGranted = await View.RequestCalendarAuthorization(true);
+                if (!permissionGranted)
+                    return;
             }
-            calendarListVisibleSubject.OnNext(calendarListVisible);
+            userPreferences.SetCalendarIntegrationEnabled(true);
+            reloadCalendars();
```
daividssilverio: Same question as above.

Toggl.Storage/Settings/IUserPreferences.cs
```diff
@@ -37,6 +39,10 @@ public interface IUserPreferences
 
         void Reset();
 
+        bool CalendarIntegrationEnabled();
+
+        void SetCalendarIntegrationEnabled(bool enabled);
```
daividssilverio: Please clear this new key-value on `Reset`.

arturs105:
Android parts are still WIP, so all of those things don't work because i only fixed build errors on droid, no further changes have been made.

Toggl.Core.UI/ViewModels/Calendar/CalendarViewModel.cs
arturs105: It's because Droid is still WIP

PR #6948, opened 2020-01-20
Use resources colours for auth screen dividers

### LINKED ISSUE #6858: 🎨 Darker divider lines on Login and SignUp with darkTheme
Related slack thread: https://toggl.slack.com/archives/CMB9FDE5D/p1576679351017400

## 🎯 What needs to be done?
Divider lines used in ------- or ------- should be darker in dark theme. We should try `#616161` color. We can also try to apply some opacity on text color to get divider color that works in both light and dark themes.

daividssilverio:
## What's this?
This handles #6858

### Relationships
Closes #6858

## Why do we want this?
So that the login/signup divider looks better on dark mode.

## How is it done?
By moving static colors to the colors.xml file.

### Why not another way?
This is the way.

### Side effects
It also fixes the broken sign up divider, which was simply back in light mode.

## Review hints
Have a phone/emulator that supports dark mode.

## :squid: Permissions
🦑 it


Toggl.Droid/Resources/values/Colors.xml
```diff
@@ -100,5 +100,8 @@
   <color name="darkBackground">#303030</color>
   <color name="primaryTextOnDarkBackground">#DDFFFFFF</color>
   <color name="weekStripeDisabledDayColor">#BDBDBD</color>
+  
+  <!-- Login/Sign Up -->
+  <color name="loginSignUpOptionsDivider">#f3f3f3</color>
```
pdrj: Total nitpick, but the other colors use capital letters for the color definition 😄 Feel free to ignore this comment.

daividssilverio: Let's maintain the pattern.

PR #6951, opened 2020-01-21
💥 Yet another null check for reports

### LINKED ISSUE #6950: Reports WS selector crash
## 💥 Crash Report
A crash that happens when closing the WS selector when no WS was selected.

### 👥 Impact
~4% of users on 2.9

### 🔗 Context
Introduced in the Reports rewrite

dshr:
## What's this?
Yet another `WhereNotNull` to prevent a crash in reports, this time in the WS selector.

### Relationships
Closes #6950 

## Why do we want this?
No :boom:

## :squid: Permissions
Only me.


dshr:
@Odrakir there was this long discussion during development, and the WS Id is only needed for iOS handoff, while Droid only needs the name, so the rest of the team felt like it's unnecessary to add the WS Id just for iOS.

dshr:
I'd rather discuss and fix this in a separate PR in order not to delay the release any more.

PR #6954, opened 2020-01-22
Revamp droid settings

### LINKED ISSUE #6908: Slight settings redesign
## 🎯 What needs to be done?
Currently, the settings view looks a bit odd when you look at these gaps. Every gap functions as a separator of some logical sections but it look very odd when you have sections without a title:

Screenshot 2020-01-09 at 21 55 29 (1)

Instead, we should try a cleaner layout inspired by newer versions of the Android settings app:

Screenshot 2020-01-09 at 21 55 29 (1)

## 📝 Detailed explanation
The key is to drop the dividers between every single item and use them only to separate sections instead of the vertical spaces we use now. For the sections that currently have some title, we can change the way the title is laid out to get even closer to the native setting app.

Check out this example of sections with and without a titles:

Screenshot 2020-01-09 at 21 55 29 (1)

## 🎨 Graphical resources
I don't think we need any

## 🛠 Technical remarks
This should be pretty straightforward

## 👕 Estimated complexity
This should be a matter of hours not days. I have no idea what 👕 is the right one in this case, **M**?

❤️

daividssilverio:
## What's this?
This PR does tow things:
- Makes the layout building strategy similar to its iOS counterpart
- Handles #6908 

### Relationships
Closes #6908

## Why do we want this?
Because it's annoying to touch the settings layout. There was a bunch of copy-pasta and so much boilerplate.
The changes make it so that it's easier to add/remove new rows and their bindings.
Also, it's more or less easy to replace the new strategy with a recycler view if we ever decide to do it, but I don't it's worth it.

## How is it done?
By creating view holders for each type of settings row, inflating and adding them to the view hierarchy manually.

### Why not another way?
Feel free to suggest something new.

### Side effects
Should be alright.

## Review hints
Please help making sure everything still works as it should.
Feel free to make suggestions to change paddings, margins directly from github suggestions.

## :squid: Permissions
🦑 it after the draft is done


daividssilverio:
Here's how it looks after the layout changes:

| Dark mode | Light mode | Inspect |
| --- | --- | --- |
| image | image  | image |
| image | image | image |

*The user name, e-mail and workspace views are not broken, I've changed the text views for those screenshots.

daividssilverio:
> This is god's intended use case for a recycler view.

@heytherewill @juanlaube maybe... but I'm not sure in this case.

While I made the views already `ViewHolders` in case someone wanted to do that. I have to disagree that this is an "intended use" for the `RecyclerView`, simply because how the collection would have to be created. (Plus I have my feelings telling me this will look bad on earlier droid versions, gotta check)

Please check how the collection would be created here: 

https://github.com/toggl/mobileapp/blob/efb9dde6cdf0ccbd1ff1cf9a1947c2d18ac5e850/Toggl.iOS/ViewControllers/Settings/SettingsViewController.cs#L82-L167

It looks really wasteful to me. Maybe someone could clarify to me why it would be a good idea?
I'll change if you insist, but it feels weird.

heytherewill:
@daividssilverio iOS looks odd because it's fighting against the tides! A full blown refactor would move the code that creates the grouping to *within* the ViewModel, fully embracing the collection nature of the view.

daividssilverio:
@heytherewill but... won't that mean wrapping the collections as sections, just to flatten them for android (as it is done in our codebase)? Plus we have android specific and iOS specific settings... we would have to work around that as well.

Are you against adding this as is, then adding the full blown refactor proposal in the refactor drawing board? This way this doesn't have to be done in a haste and can be carefully designed to not look odd in either platform.

The changes made here are a step towards a full refactor anyway.

daividssilverio:
Hey @AlicjaSuska @semanticer here's how it looks with 64 dp of left margins. (I meant to post it here)

| dark mode | light mode |
| --- | --- |
|  |  |

semanticer:
Oh god, this would be so simple, short and much more readable with Epoxy & MvRx 😞which would be using recyclerView underneath but keep this simple declarative nature.
Anyway...
In this case, I'm still slightly with @daividssilverio (I'm also the one who seduced him to fallow this path)
I think the solution with RecyclerView and sections would be unnecessarily complex, for no gain. 
Recycler view is great for long, changing lists that need to be updated and where views need to be recycled. It is not ideal for static predictable short lists with lost of different interactions and item types .. 
That being said, the current solution is quite repetitive and verbose - maybe there are ways to improve it?

Toggl.Droid/Activities/SettingsActivity.cs
```diff
@@ -35,68 +38,88 @@ protected override void InitializeBindings()
...
-                .Subscribe(dateFormatTextView.Rx().TextObserver())
-                .DisposedBy(DisposeBag);
-
-            ViewModel.BeginningOfWeek
-                .Subscribe(beginningOfWeekTextView.Rx().TextObserver())
-                .DisposedBy(DisposeBag);
-
-            ViewModel.DurationFormat
-                .Subscribe(durationFormatTextView.Rx().TextObserver())
+                .Select(stoppedTimerNotificationsEnabled => new ToggleRow(NotificationsRunningTimer, stoppedTimerNotificationsEnabled, ViewModel.ToggleRunningTimerNotifications))
```
juanlaube: Found this bug while testing it...
```suggestion
                .Select(stoppedTimerNotificationsEnabled => new ToggleRow(NotificationsRunningTimer, stoppedTimerNotificationsEnabled, ViewModel.ToggleStoppedTimerNotifications))
```

daividssilverio:
Hey @semanticer that can be changed easily whenever we want now thanks to your suggestion.
Let's merge this as is and create another issue for the keylines, maybe even fixing other areas of the app.

PR #6966, opened 2020-01-22
Consider latest tag when building for release

heytherewill:
## What's this?
This fixes the build system so the generated bundle uses the correct tags

## Why do we want this?
So we can upload 2.10

## How is it done?
By adding the date to the git output that lists the tags and then ordering by the date

### Why not another way?
This is straightforward enough, I guess

### Side effects
Working builds

## :squid: Permissions
Only me or @daividssilverio


build.cake
```diff
@@ -134,12 +134,15 @@ private string GetVersionNumberFromTagOrTimestamp()
...
     StartProcess("git", new ProcessSettings
     {
-        Arguments = "tag --list '" + platform + "-*'",
+        Arguments = "tag --list '" + platform + "-*' --format='%(creatordate:short)%09%(refname:strip=2)",
         RedirectStandardOutput = true
     }, out var redirectedOutput);
-
     
-    var tagName = redirectedOutput.DefaultIfEmpty(formattedTimestamp).Last();
+    var tagName = redirectedOutput
```
daividssilverio: There's no need for the old `DefaultIfEmpty` now? Why was it there?

build.cake
```diff
@@ -134,12 +134,15 @@ private string GetVersionNumberFromTagOrTimestamp()
     
     StartProcess("git", new ProcessSettings
     {
-        Arguments = "tag --list '" + platform + "-*'",
+        Arguments = "tag --list '" + platform + "-*' --format='%(creatordate:short)%09%(refname:strip=2)",
```
daividssilverio: It's not obvious what this does. Do you mind adding a comment?

build.cake
```diff
@@ -132,14 +132,21 @@ private string GetVersionNumberFromTagOrTimestamp()
...
         RedirectStandardOutput = true
     }, out var redirectedOutput);
-
     
-    var tagName = redirectedOutput.DefaultIfEmpty(formattedTimestamp).Last();
+    var tagName = redirectedOutput
+        .DefaultIfEmpty(formattedTimestamp)
+        .Split(Environment.NewLine)
```
daividssilverio: The build still fails because it can't find this `Split` on a `IEnumerable<string>`.
Would it work if we just add a `Last()` before the `Split`?

```suggestion
        .Last()
        .Split(Environment.NewLine)
```

PR #6971, opened 2020-01-23
🐛 Fix google login issue

### LINKED ISSUE #6572: it's not possible to sign up with google after logging in with google once
How to replicate:

1. Log in with google
2. Log out from the settings screen
3. Try to sign up with google from the Sign up screen
4. The account picker dialog won't appear and an error message will appear (tried to sign up with an existing account).

The account picker should appear after a logout.

semanticer:
## What's this?
Fix for a weird issue where google login would just blink n fail if you had some successful/half-successful login before.

### Relationships

Closes #6572

## Why do we want this?
Google login should work every time

## How is it done?
The issue was that when you requested the google sign dialog for the second time, it would automatically choose and return the last logged user; thus the dialog would only blink and disappear.
When I finally understood where the problem is, I found out some related StackOverflow thread https://stackoverflow.com/questions/36659753/google-login-uses-same-account-everytime-users-login

Then I just carefully choose the right place for `googleApiClient.ClearDefaultAccountAndReconnect();  ` 

### Why not another way?
This seems to be the simplest way

### Side effects
Hopefully none

## Review hints
Here is the bundle build https://app.bitrise.io/build/033959b8f4386ce6#?tab=artifacts
I successfully installed `com.toggl.giskard-Signed.aab.universal.apk` and google login doesn't work there at all but it doesn't work even for the current develop(or 2.10) so it's probably not my fault 🤷‍♂ 
I've even tried the new simple https://play.google.com/apps/publish/internalappsharing/ method but also without success.
How do you test it then you ask? I have no idea - the only way I can think of now is to release it to the internal channel.

## :squid: Permissions
Anyone, anytime

### DIFF
--- a/Toggl.Droid/Activities/ReactiveActivity.GoogleTokenProvider.cs
+++ b/Toggl.Droid/Activities/ReactiveActivity.GoogleTokenProvider.cs
@@ -95,6 +95,10 @@ namespace Toggl.Droid.Activities
                     finally
                     {
                         isLoggingIn = false;
+                        if (googleApiClient.IsConnected)
+                        {
+                            googleApiClient.ClearDefaultAccountAndReconnect();
+                        }
                     }
                 });
             }


Toggl.Droid/Activities/ReactiveActivity.GoogleTokenProvider.cs
```diff
@@ -95,6 +95,10 @@ private void onGoogleSignInResult(Intent data)
                     finally
                     {
                         isLoggingIn = false;
+                        if (googleApiClient.IsConnected)
+                        {
+                            googleApiClient.ClearDefaultAccountAndReconnect();
```
heytherewill: Is calling clear and reconnect right after the user signs in correct? Also, can this never throw? Shouldn't it be around its own try/catch block?

daividssilverio: Based on the docs, it looks like this will cause the google account picker to appear again.

https://developers.google.com/android/reference/com/google/android/gms/common/api/GoogleApiClient.html#public-abstract-pendingresultstatus-cleardefaultaccountandreconnect-

Also, we are using a deprecated api. It seems like we should've been using the `GoogleApi` by now.

semanticer: Oh... I didn't notice the deprecation. I guess those deprecation warnings doesn't propagate into C# :-( 
I've read the documentation but for some reason, it behaves differently - I would love to see the source code, maybe I will be able to find something.
Anyway - I need to look at the way to bring this whole think up to date.

semanticer:
This is superseded by #7061

Issue #6953, opened 2020-01-21
Refactor CalendarCollectionViewSource


juanlaube:
## 🚧 Refactor Proposal
`CalendarCollectionViewSource` has become a major pain to maintain and it's the biggest source of bugs and crashes in our app, it has become is the core of the calendar view and how users interact with it, so for every new thing we added to the calendar, we had to touch this class.

## Technical aspects
1. We need to stop using index paths to track the item being edited, index paths are not stable because it could change when inserting, reordering or deleting items, instead we should rely on calendar item ids.
1. Stop reloading the collection view on every datasource change, we should start using our diffing algorithm and apply individial insert/delete/reoder operations collection, the implementation should be almost identical to what we already do for table views.
1. At the same time, we should stop using `ObservableGroupedOrderedCollection` (we can delete that class after this refactor) and start using `AnimatableSectionModel`, the code will be cleaner and is needed to use the diffing algorithm.

juanlaube:
Related issues: #6739 and #6748

PR #6985, opened 2020-01-24
Refactor CalendarCollectionViewSource

### LINKED ISSUE #6739: Rewrite CalendarCollectionViewSource so that it doesn't remember indexPath
Ref: #5869, #4886, #5432

The problem with all the crashes ☝️ is that the indexes of the items in the calendar view change over time and the `indexPath` state variable now might not point to the same item, or it can even point outside of the range of items in the collection.

The latest report is from 2.8: https://appcenter.ms/orgs/Toggl/apps/Toggl-iOS/crashes/errors/3327916630u/reports/2518272647600009999-608a2aa2-054e-437d-90db-5cf016fde339/threads

We tried preventing these crashes, but it seems pointless. We should find a better way to represent the state of the collection while editing.

juanlaube:
## What's this?
This is a refactor on the calendar collection view source to make things independent of the selected item index path, instead of that it uses the item ids now because those are stable.

### Relationships

Closes #6739 
Refs #6953 

## Why do we want this?
Depending on the index paths was error prone, we had a lot of crashes and fixing them made our code smell. The refactor removes this dependency on index paths and makes our code cleaner.

## How is it done?
Did I say we don't use index paths anymore? 😄 

### Why not another way?
🤷‍♂ I'm not sure there's a better way.

### Side effects
None?

## Review hints
Please do functional testing, I tested heavily but I want to make sure everything is still working.

## :squid: Permissions
🚫 No, I want at least two reviews from people who worked on the calendar recently.


heytherewill:
Why does it ref and not close #6953?

Toggl.iOS/ViewSources/CalendarCollectionViewSource.cs
```diff
@@ -40,15 +40,18 @@ public sealed class CalendarCollectionViewSource : UICollectionViewSource, ICale
         private IList<CalendarItemLayoutAttributes> layoutAttributes;
         private TimeFormat timeOfDayFormat = TimeFormat.TwelveHoursFormat;
         private DateTime date;
-        private NSIndexPath editingItemIndexPath;
-        private NSIndexPath runningTimeEntryIndexPath;
-        private UICollectionView collectionView;
+        private string selectedItemId;
+        private string runningTimeEntryId;
+        private string newItemId = "";
```
heytherewill: Why only this field is being initialized?

arturs105:
After stopping a time entry from the calendar it sometimes doesn't update to appear stopped. Steps to reproduce:
1. Start or continue a time entry
2. Wait for he running entry to sync
3. Stop the running time entry from the calendar

Toggl.iOS/ViewSources/CalendarCollectionViewSource.cs
juanlaube: I should have used a const 👀

juanlaube:
@arturs105 I fixed the bug when stopping a time entry, however I cannot reproduce the null ref, do you have more details?

juanlaube:
@heytherewill #6953 is bigger than this, we need to replace the observable collection with what we use now in tableview and implement the diffing for collectionviews, that's why it's only refs and not closes.

PR #6995, opened 2020-01-25
Fix widgets authenticated tap to open app

### LINKED ISSUE #6991: Start time entry from widget creates another activity instance
Why is that a problem?

Our app wasn't designed to work with multiple instances of the activities.

How to replicate:
1. Start the application
2. Go to the main log
3. Go to the start time entry widget
4. Click on the running time entry
5. Another instance of the main log activity will be created, you can verify that by pressing back on it, which returns to the existing main log activity

daividssilverio:
## What's this?
This fixes #6991

## Why do we want this?
Because our apps don't were not designed to support multiple activity instances.

## How is it done?
By changing the default tap intent in the widgets that should lead to the app.
The new way works similar to how we currently handle the android persistent notifications.

### Why not another way?

### Side effects
There shouldn't be any. But please help us out testing it.

## Review hints
None.

## :squid: Permissions
🦑 it


Toggl.Droid/Widgets/WidgetIntentHelper.cs
```diff
@@ -0,0 +1,16 @@
+using Android.App;
+using Android.Content;
+
+namespace Toggl.Droid.Widgets
+{
+    public static class WidgetIntentHelper
```
heytherewill: Aren't we triggering app opens from other places? Shouldn't all these places use the same consistent way of opening the app? IOW, shouldn't this method be an extension method on context that gets used all throughout the app rather than a helper that only applies to widgets? 🤔

daividssilverio: Sure, that can be done.

Issue #6977, opened 2020-01-23
Review MainVM observables to prevent expensive operations on the main thread


heytherewill:
## 🎯 What needs to be done?
We need to go through all the observables of the `MainViewModel` (and its sub-ViewModels as well) to ensure that none of them are executing expensive operations on the main thread.

## 📝 Detailed explanation
Since this is the first VM the user sees, if there are any duplication of call and the code that produces the initial values runs on the Main thread, this may slow down the startup of our app. We need to go through each of the observables and verify if the code that runs to generate the values is running on the main thread or not. If it is, we should marshall it. We also need to ensure that multiple subscriptions are not triggering multiple disk accesses. Most (if not all) observables in ViewModels should have shared subscriptions

## 🛠 Technical remarks
A couple of `.SubscribeOn(schedulerProvider.BackgroundScheduler);` might do the trick, but we need to verify where this is needed and where it isn't.

## 🔗 Dependencies
This will be particularly useful once #6679 is merged, but it doesn't depend on it

## 🏆 Outcome
Observables in the MainViewModel only produce values once and expensive operations needed to produce said values should run on a thread that does not block  the UI

## 🖼 Express your sentiment towards the issue
![image](https://user-images.githubusercontent.com/7688727/73033838-fd231880-3e21-11ea-82b1-ea10819f0866.png)

heytherewill:
No perceivable improvements in this area

PR #7021, opened 2020-01-27
Lazy Tab ViewModels

### LINKED ISSUE #6679: Load and initialize the VMs for all tabs lazily
Currently when the user opens the app (cold start), the `MainTabBarViewModel` is loaded. This VM creates its child VMs in its constructor (main VM, reports VM, calendar VM) and the initialization method of the VM initializes all of the tab VMs.

I think we should change this:
- we should only create and initialize the main VM during startup
- the rest of the VMs should be either created and initialized in a background thread without blocking the UI
- or we should simply load the VMs and initialize them at the first occasion when the user navigates to the tabs (the user might not go to calendar or reports every time he/she opens the app)

Since this is all `Toggl.Core` territory, we must make sure we don't break either the iOS or the Android app while we're doing this. Both of these app's performance would benefit from this change.

heytherewill:
## What's this?
This makes the ViewModels used in tabs lazy

### Relationships
Closes #6679 
Brings the startup speed time from 1700ms to 1360ms (-340ms)

## Why do we want this?
No real reason to load everything up-front when you can wait

## How is it done?
Make things lazy on Droid. I'm not really making anything lazy on iOS, mostly because performance there is not heavily affected anyways.

### Side effects
The PR is as small as possible, but let's make sure we are not introducing bugs here, testing specially what happens when the app is collected on the BG

## :squid: Permissions
Me


daividssilverio:
Btw, I made a quick experiment replacing the tabs in the main tab bar to start with the reports view and the cold startup itself wasn't affected... (In debug though).

heytherewill:
@daividssilverio The problem is that the dependencies themselves aren't lazy enough. See https://github.com/toggl/mobileapp/issues/6977 and https://github.com/toggl/mobileapp/issues/7024

Toggl.iOS/ViewControllers/MainTabBarController.cs
```diff
@@ -35,10 +37,16 @@ public sealed partial class MainTabBarController : UITabBarController
...
+                .GetAwaiter()
+                .GetResult();
 
-            UIViewController createTabFor(ViewModel childViewModel)
+            async Task<UIViewController> createTabFor(Lazy<ViewModel> lazViewModel)
             {
+                var childViewModel = lazViewModel.Value;
+                await childViewModel.Initialize();
                 var viewController = ViewControllerLocator.GetNavigationViewController(childViewModel);
                 var childViewModelType = childViewModel.GetType();
```
arturs105: Don't think the `child` prefix of `childViewModel` is necessary anymore.

heytherewill:
@semanticer Problem number 1 is fixed by @daividssilverio's proposal for #4423, so I'll let that be solved in the other issue :)

semanticer:
but #2 is still not solved for me though @heytherewill

daividssilverio:
Quick question, why did the `Toggl.sln` file changed?

heytherewill:
@daividssilverio I rolled that back, thanks for catching it!

PR #7029, opened 2020-01-28
Fix date format

### LINKED ISSUE #6975: Contextual menu start/end dates are always in AM/PM mode.
Title says it all.
To see what I'm talking about, try the following:

- Check your `Use 24-hour format` user preferences (true)
- Go to the calendar view
- Select any entry from the calendar
- Check the string at the top of the contextual menu, it will be using the (AM-PM) format.

heytherewill:
## What's this?
This makes the contextual menu respect the user's preferences for date 12/24 hours

### Relationships
Closes #6975 

## Why do we want this?
We should respect the user

## How is it done?
The first commit renames an interactor so that it reflects its observing nature
The second one changes the way the formatting is calculated to be more reactive and to consider format changes

### Side effects
None, really

## Review hints
Do a functional test, just in case!

## :squid: Permissions
Anyone, anytime


Toggl.Core.UI/ViewModels/Calendar/ContextualMenu/CalendarContextualMenuViewModel.cs
```diff
@@ -196,20 +205,17 @@ private void handleCalendarItemUpdate(CalendarItem calendarItem)
                 timeEntryInfoSubject.OnNext(currentTimeEntryDisplayInfo);
             }
 
-            if (calendarItem.StartTime != currentStartTimeOffset || calendarItem.Duration != currentDuration)
```
daividssilverio: it seems like the `DistinctUntilChanged`s  used in place of this if are making the tests fail.

heytherewill:
I don't get what you mean, @dshr, tests were fixed in https://github.com/toggl/mobileapp/pull/7029/commits/1c7f9ce618d2655007de572fbad631d91cb528b7

dshr:
> I don't get what you mean, @dshr, tests were fixed in [1c7f9ce](https://github.com/toggl/mobileapp/commit/1c7f9ce618d2655007de572fbad631d91cb528b7)

I tested a stale ver, it seems. Gonna have another look!

PR #7045, opened 2020-01-29
Fix main log section header reloading

### LINKED ISSUE #7013: Main log section header total duration is not updating after changes in the section
## 🐛 Bug Report
Main log section header total duration does not update after changes to the data in the section.

## Context
Build/Branch: ios 2.10
Phone OS: 13.3
Phone model: iphone 11 pro

## Expected behavior
To show the correct duration of the section

## To reproduce
- Start a TE
- Stop it
- Continue that TE
- Stop it
- Section duration and sum of TE duration does not match
If we scroll the section header out of bounds and back in, it will refresh

## Suggestions
We need to reload sections headers when datasource changes.

dshr:
## What's this?
A fix to the missed header updates that started happening in the main log after the flattening effort.

### Relationships
Closes #7013

## Why do we want this?
✨ 

## How is it done?
By removing one layer of inheritance in the generic param of `BaseTableHeaderFooterView` subclasses that we use for the headers, instead checking/casting it during runtime.

### Why not another way?
No covariance for generic params :(

## Review hints
Try to reproduce.

## :squid: Permissions
@sofokli1 only.


Toggl.iOS/Cells/MainLog/SuggestionsHeaderView.cs
```diff
@@ -9,6 +9,8 @@
 
 namespace Toggl.iOS.Cells.MainLog
 {
+    // this is supposed to work with SuggestionsHeaderViewModel, but due to variance constraints on
+    // generic types in C# we have to use the base type
```
sofokli1: is this comment needed here?

Toggl.iOS/Cells/MainLog/TimeEntriesLogHeaderView.cs
```diff
@@ -8,7 +8,9 @@
 
 namespace Toggl.iOS.Views
 {
-    public partial class TimeEntriesLogHeaderView : BaseTableHeaderFooterView<DaySummaryViewModel>
+    // this is supposed to work with DaySummaryViewModel, but due to variance constraints on
+    // generic types in C# we have to use the base type
+    public partial class TimeEntriesLogHeaderView : BaseTableHeaderFooterView<MainLogSectionViewModel>
```
arturs105: It's not easy to understand what `this` refers to. Maybe change to
* `...to work with DaySummaryViewModel as T, but due...`
* `...to work with DaySummaryViewModel as the generic type parameer, but due...`

PR #7046, opened 2020-01-29
Add entity models

### LINKED ISSUE #6448: Introduce entity classes in Toggl.Shared
In our ever going effort to make the Spanish contingent of Toggl happier by removing interfaces from the app, our entities are the next target!

We need to create one class for each entity type. e.g.: we shall introduce a `TimeEntry` type and the type should have the same fields as an `IDatabaseTimeEntry`, but it should not implement that interface! It will have a `Workspace` rather than an `IDatabaseWorkspace`, for example.

These types won't be used anywhere for now, but will in the future

The documentation for the queries project can be found [here](https://www.notion.so/toggl/Realmoval-part-one-Queries-2870434ba3eb4c8b80e18f22f95828a6)

heytherewill:
## What's this?
This adds versions of our models that are completely independent from interfaces

### Relationships
Closes #6448

## Why do we want this?
For the upcoming database refactor

## Review hints
Please make sure that the models are correct and that I didn't forget any property or misnamed anything.

## :squid: Permissions
Only Myself


Toggl.Shared/Models/Client.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System;
+
+namespace Toggl.Shared.Models.DoNotUse
+{
+    public sealed class Client
```
daividssilverio: Why doesn't it have the `WorkspaceId`,  `SyncStatus`,  `LastSyncErrorMessage`,  `IsDeleted` props from the current `Client : IThreadSafe` ? We won't be needing them anymore?

Toggl.Shared/Models/User.cs
```diff
@@ -0,0 +1,42 @@
+﻿using System;
+
+namespace Toggl.Shared.Models.DoNotUse
+{
+    public sealed class User
+    {
+        public long Id { get; }
+        public string ApiToken { get; }
+        public long? DefaultWorkspaceId { get; }
```
sofokli1: based on the other models, why not `public Workspace? DefaultWorkspace { get; }`?

Toggl.Shared/Models/WorkspaceFeatureCollection.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System.Collections.Generic;
+
+namespace Toggl.Shared.Models.DoNotUse
+{
+    public sealed class WorkspaceFeatureCollection
+    {
+        public long WorkspaceId { get; }
```
sofokli1: same as the other comment. Why just the id, instead of the whole model?

Toggl.Shared/Models/Client.cs
amulware: We definitely need @kornelijepetak and @simonrozsival to weigh in on this.

amulware:
I agree with Eugene that we should probably expose a proper model tree, not ids, which require extra fetching - if the intention is to change this, I haven't seen it in any of the docs, please refresh my memory. :)

Toggl.Shared/Models/Client.cs
kornelijepetak: I need to properly understand the underlying purpose of these models, so I'll have a talk with @heytherewill next week and see what to do with those properties as well.

Toggl.Shared/Models/WorkspaceFeature.cs
```diff
@@ -0,0 +1,14 @@
+﻿namespace Toggl.Shared.Models.DoNotUse
+{
+    public sealed class WorkspaceFeature
```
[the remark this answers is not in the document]
heytherewill: Great point

heytherewill:
it ded

PR #7050, opened 2020-01-30
Fix bar chart view crash

### LINKED ISSUE #7025: Crash when updating constraints in the BarView
## 💥 Crash Report
Due to some unknown course of events the constraint multiplier ends up being a NaN, which causes the app to crash.

### 👥 Impact
17 cases in the last 4 days since the rollout of 2.9.1

### 🔗 Context
`SIGABRT: Objective-C exception thrown. Name: NSInternalInconsistencyException Reason: Multiplier is not finite! That's illegal.`

dshr:
## What's this?
A fix to the occasional crashes in the Bar Chart view — the VM would return Bars with all values set to 0, which would trigger a divide by 0, returning an infinity. Adding the check anywhere else doesn't quite make sense, since at those points we don't care about the the whole bar group.

### Relationships
Closes #7025

## Why do we want this?
No :boom:

## How is it done?
`float.IsFinite(value)`

## :squid: Permissions
Go for it!


sofokli1:
is there an easy way to reproduce for testing?

dshr:
> is there an easy way to reproduce for testing?
I didn't manage to reproduce it in debug at all, but the crash logs say it all

PR #7051, opened 2020-01-30
Add start time entry DTO

### LINKED ISSUE #6452: Introduce a DTO for the Start Time Entry Query
This type must be added to the `Toggl.Storage` project.

With the queries project, we wanna make the types in our database more reasonable. With that in mind, we wanna introduce DTOs that are unable to express invalid states.

For this DTO in particular, no invalid states means that the DTO should not have a `Duration` or `StopTime` fields, since a running time entry does not possess either.

For more info, please check the documentation for the queries project which can be found [here](https://www.notion.so/toggl/Realmoval-part-one-Queries-2870434ba3eb4c8b80e18f22f95828a6)

heytherewill:
## What's this?
This adds a DTO that will be later used in the queries project

### Relationships
Closes #6452

## Why do we want this?
For the queries project

## Review hints
Make sure this is everything that's needed for a time entry to be created

## :squid: Permissions
Anyone, anytime


heytherewill:
The idea was to make issues atomic so I made them atomic 🤷‍♂

Issue #7053, opened 2020-01-30
Add a temporary animated "splash screen"/transition continuation for the MainTabBar


daividssilverio:
The idea is to have something following the splash screen.

We could even create a transition between the `SplashScreen` and the `MainTabBar`. We could maybe animate the logo up and display the `Toggl` name bellow it.

This transition can ease up how the users feel about the load times during cold starts on Android.

Since we are currently working on improving the overall performance of the app, this could be a temporary measure or it could stay there if people enjoy it.

daividssilverio:
Hey @maritoggl here's the details for the idea discussed on https://toggl.slack.com/archives/C02APMYTM/p1580342393146800

Let's do this?

maritoggl:
I'm ok with what you have proposed. Let's make things smooth and improve the perceived speed of the app, but be super careful of adding any transitions/animations that seem like they're there just to be there. I also want us to avoid any personalized screens for the sake of simplicity - we don't have good enough data to do anything with usernames, various cool quotes and stuff would also be tricky since we need to consider localization etc. So let's just play with the Toggl logo and brand name for the time being.

heytherewill:
Superseded by #7055

PR #7055, opened 2020-01-30
Add placeholders directly in the MainTabBarActivity

### LINKED ISSUE #4423: Android Bottom Bar: Create placeholders for pages

daividssilverio:
## What's this?
This adds placeholders directly in the `MainTabBarActivity`.

### Relationships
Closes #4423 

## Why do we want this?
Because our load times aren't fast enough to not require any placeholders.

Why directly on the `MainTabBarActivity`? 
- It takes some time to load even the first fragments.

Why the other fragments don't have placeholders yet?
- They can be done in other PRs or not done at all. The Reports for example doesn't look like it needs one.
- ~Still validating the idea~ As shown internally, the calendar also doesn't seem to need the placeholder, it appears and disappears too fast for it to look good.

## How is it done?
By adding the placeholders directly in the main tab bar activity and handling their visibility based on the "readiness" of each fragment.

### Why not another way?
Feel free to suggest something different or completely refuse this idea (don't do it if you can't suggest something better though).

### Side effects
None.

## Review hints
- Review by commit.
- Do test functionally. I did test, but I'm asking for more tests.

## :squid: Permissions
🦑 it into develop ~after #7021 is merged or directly into #7021~
~This PR is currently targeting #7021 to make the reviews easier.~


daividssilverio:
Also, we might or might not need to do #7053 depending on the results of this PR.

daividssilverio:
The last two commits also fix a bug in develop that makes android crash when changing configurations (enabling/disabling dark mode).

Toggl.Droid/Activities/MainTabBarActivity.ViewBinds.cs
```diff
@@ -1,12 +1,20 @@
...
 using Toggl.Droid.Extensions;
 using Google.Android.Material.BottomNavigation;
+using Toggl.Droid.Fragments;
 
 namespace Toggl.Droid.Activities
 {
     public sealed partial class MainTabBarActivity
     {
         private BottomNavigationView navigationView;
+        private Dictionary<Type, bool> readyLayouts = new Dictionary<Type, bool>();
```
pdrj: I don't know if that is a c# thing, but I'd normally use a set if the values are not used.

Toggl.Droid/Activities/MainTabBarActivity.cs
```diff
@@ -238,16 +235,16 @@ private async Task showFragment(int fragmentId)
...
             SupportFragmentManager.ExecutePendingTransactions();
 
             var initialFragment = await getCachedFragment(initialTabItemId);
+            setPlaceholderVisibility(typeof(MainFragment),  initialTabItemId == Resource.Id.MainTabTimerItem && !initialFragment.IsAdded);
             if (!initialFragment.IsAdded)
             {
                 SupportFragmentManager
                     .BeginTransaction()
-                    .Add(Resource.Id.CurrentTabFragmmentContainer, initialFragment, initialTabItemId.ToString())
+                    .Replace(Resource.Id.CurrentTabFragmmentContainer, initialFragment, initialTabItemId.ToString())
```
pdrj: Could you explain a bit why is there a replace now?

Toggl.Droid/Activities/MainTabBarActivity.ViewBinds.cs
heytherewill: Agreed

Toggl.Droid/Activities/MainTabBarActivity.cs
daividssilverio: Basically, a crash was going on when the view was being restored after being killed by the system for some reason.
After some trials, that fixed it. This makes sure if any existing fragments with same Id get replaced with the new fragment.

Toggl.Droid/Activities/MainTabBarActivity.ViewBinds.cs
daividssilverio: Sure. Done.

PR #7057, opened 2020-01-30
Fix syncing of user properties

### LINKED ISSUE #6585: Workspace preferences are not propagated when you go online after switching workspaces offline
To replicate:
1. Have at least two workspaces A, B
2. While online, be on workspace A
3. Check that the webapp also has the workspace set to A
4. Go offline
5. In the mobile app, switch to workspace B
6. Go online & Sync
7. Go to the web app and refresh
9. Sync again
8. Check that the web app is still on workspace A and now the mobile app is also on workspace A.

heytherewill:
## What's this?
This makes the user model sync properly

### Relationships
Closes #6585

## Why do we want this?
Proper syncing

## How is it done?
See issue discussion

## Tests
[Integration tests](https://app.bitrise.io/build/0fae6021d620d8a1)
[Syncing tests](https://app.bitrise.io/build/39c6c9e67c3c3cd1)

## :squid: Permissions
Anyone, anytime


amulware:
To catch more flies and hit more birds: Consider comparing the entire model for equality, since all other fields should be the same as well :)

PR #7059, opened 2020-01-31
🐛 Prevent RatingViewShouldBeVisible from emitting too often

### LINKED ISSUE #7035: Fix rating view
## 🐛 Bug Report
As discussed at length here: https://github.com/toggl/mobile-discussions/issues/51 it looks like something is very wrong with our rating view. 

Our event data says that the view is shown to a great number of users, however less than half of the expected number interact with them (this is compared to numbers from before the view was broken for a while).

Note that on iOS, the view behaves fine and we see the data we expect. Something broke on Android specifically, between September and November.

## Context
Rating view was entirely broken in October until: https://github.com/toggl/mobileapp/pull/6087
Data and discussions on what we observed recently is here: https://github.com/toggl/mobile-discussions/issues/51

## Expected behavior
The rating view should work properly for all users, as configured with remote config. The conversion rates of each step should be the same as they were throughout most of last year.

## To reproduce
N/A

## Suggestions
I'm not sure if we can easily reproduce it, but it's worth a try. Otherwise, look at what code changes happened between September and November that could have broken it on Android specifically.

pdrj:
## What's this?
This makes the rating view not disappear as soon as it appears.

### Relationships

Closes #7035 

## Why do we want this?
To get ratings.

## How is it done?
We have the following observable:
```        
            => updateRemoteConfigCacheService.RemoteConfigChanged
                .Select(_ => remoteConfigService.GetRatingViewConfiguration())
                .SelectMany(criterionMatched) // 1
                .Select(tuple => tuple.criterionMatched && dayCountPassed(tuple.configuration))
                .DistinctUntilChanged();
```

The place marked with `1` ran twice in a very quick succession, but I unfortunately don't know why. The config has not beed updated in the meantime. Maybe it's something about `SelectMany`? I'd like some rx experts to weigh in. 

The first emission of `RatingViewShouldBeVisible` caused, in the main view model, a `RatingViewOutcome` value of no interaction with the rating view to be saved in onboarding storage. Then, when the observable emitted `true` again, the main log list got constructed again, but this time the rating view would not be shown, as the storage said user has not interacted with the view when it was previously shown.

In any case, my quasi-fix is to not let the whole thing emit if the emitted value would be the same as the last one. 
 
### Why not another way?
Please suggest something better.

## Review hints
I'd like at least one @toggl/androids and one @toggl/ios-senseis to test this. You can get the staging plist/json from firebase and make the debug app initialize with that config. Launch the app once, change the device date to one day ahead, go back to the app and start->stop a TE. The rating view should appear and stay there. 

I couldn't provide a release apk to test with, as the build is failing on bitrise with a cryptic message and I wanted to get this out to testing as soon as possible.

## :squid: Permissions
Anyone.


heytherewill:
While I totally see this solution working, why was it broken only on Android if the fix happens in shared code?

pdrj:
I don't know 🤔 I've just run some additional tests on an iOS emulator and couldn't find an answer to your question. I'll do a bit more analysis, but I've a feeling that it might be something really time consuming to figure out.

pdrj:
I wasn't able to let this one go unexplained and did some more testing. Surprise, surprise, I've been able to reproduce it in iOS as well 😄

Repro steps (in simulator, debug build with firebase initialized):
- Do a fresh install (so uninstall any previous versions to clean up storage)
- Login, minimize the app, enter the app again (this should fetch RC)
- Minimize the app
- Change your system's date to one day ahead, the simulator should update its date too
- Enter the app again and stop a TE
- You should get a rating prompt blink.

It's important to not do all the steps when the app is in foreground because in that case I couldn't reproduce.

I've yet to find out why does the observable emits twice. I added some logging around the stop interactor to see if it maybe gets called twice, but it doesn't.

Toggl.Core/Experiments/RatingViewExperiment.cs
```diff
@@ -23,7 +23,8 @@ public IObservable<bool> RatingViewShouldBeVisible
             => updateRemoteConfigCacheService.RemoteConfigChanged
```
sofokli1: it could be like, it fire once when getting the local config and once when it gets the remote config

PR #7061, opened 2020-01-31
🚧 Update google signIn APIs

### LINKED ISSUE #6572: it's not possible to sign up with google after logging in with google once

### LINKED ISSUE #7060: Transition to new (non-deprecated) Google SignIn APIs
## 🚧 Refactor Proposal
Move away from managing connection manually with deprecated GoogleApiClient https://developers.google.com/android/reference/com/google/android/gms/common/api/GoogleApiClient.html and use the https://developers.google.com/android/reference/com/google/android/gms/common/api/GoogleApi.html instead.

## Technical aspects
Here is a blog post about the difference between the old and new APIs
https://android-developers.googleblog.com/2017/11/moving-past-googleapiclient_21.html

semanticer:
## What's this?
Update of our Google SignIn APIs.  I was able to simplify the logic and maybe even fixed the small bug we had (#6572)

### Relationships

Closes #7060
Maybe also #6572

## Why do we want this?
We don't want to use deprecated APIs and we want to simplify Google SignIn logic. We also want to fix #6572 bug along the way.

## How is it done?
I've tried to maintain the same API structure and just replaced the internals with new GoogleApi

### Why not another way?
Feel free to suggest

### Side effects
I hope none

## Review hints
I have no idea how to fully test this without making a release build. 

## :squid: Permissions
Me

### DIFF
--- a/Toggl.Droid/Activities/ReactiveActivity.GoogleTokenProvider.cs
+++ b/Toggl.Droid/Activities/ReactiveActivity.GoogleTokenProvider.cs
@@ -1,154 +1,106 @@
-using Android.App;
 using Android.Content;
-using Android.Gms.Auth;
-using Android.Gms.Auth.Api;
 using Android.Gms.Auth.Api.SignIn;
-using Android.Gms.Common;
 using Android.Gms.Common.Apis;
 using System;
 using System.Reactive;
 using System.Reactive.Linq;
 using System.Reactive.Subjects;
-using System.Threading.Tasks;
-using Toggl.Core.Exceptions;
+using Android.Gms.Tasks;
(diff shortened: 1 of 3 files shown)


heytherewill:
Does this supersede #6971?

semanticer:
Yes, I'm gonna close the #6971 to make it explicit.

semanticer:
It doesn't work and I don't want to do anything about it ever again (ok maybe until aurora)

PR #7063, opened 2020-01-31
🖼 Experiment with the selection list item background

### LINKED ISSUE #6847: Try to change background in SelectProjectActivity and SelectClientActivity
Origin story in https://toggl.slack.com/archives/CMB9FDE5D/p1576592001002000 and https://toggl.slack.com/archives/CMB9FDE5D/p1576592796007700   
## 🎯 What needs to be done?
We should try to use `background` instead of `card` bg for list items in SelectProjectActivity and SelectClientActivity. The result should be consulted with a designer to make sure contrasts work in both themes and also make sure it's not uglier than before.

## 🎨 Graphical resources
Current unsatisfying state
dark
awesomecolors

semanticer:
## What's this?
Experimentation with the background colors of selection items. I did this for project, client and tag selections. I accidentally didn't change the color of `AbcCreateEntityCell` but now I feel like it might be a good idea to let this one pop out a bit - let me know what you think.

**Tags:**

**Projects:**

**Clients:**

### Relationships

Closes #6847

## Why do we want this?
Because it might look better

## How is it done?
Just some background resource changes

### Why not another way?
We are just trying this. Feel free to suggest alternatives 

## :squid: Permissions
Me


Toggl.Droid/Resources/layout/SelectClientActivityCell.axml
```diff
@@ -2,7 +2,7 @@
 <FrameLayout
     xmlns:tools="http://schemas.android.com/tools"
     xmlns:android="http://schemas.android.com/apk/res/android"
-    android:background="@color/cardBackground"
+    android:background="@color/background"
```
daividssilverio: Quick question, if you are using `@color/background`, isn't it easier to remove the background as a whole?

Does it have something beneath it or is it needed for the ripple effect?

PR #7074, opened 2020-02-09
Add dummy endpoints for new sign in with Apple API

### LINKED ISSUE #6910: Implement dummy `SignUpWithApple` endpoint
## 🎯 What needs to be done?
Implement the interface for a new endpoint in `UserApi` to sign up with Apple.
We don't know the interface of this endpoint yet, so until we have more details, just create a method that does nothing.

## 📝 Detailed explanation
The method should look like this:

`public async Task<IUser> SignUpWithApple(string authToken, bool termsAccepted, int countryId, string timezone)`

### LINKED ISSUE #6911: Implement dummy `GetWithApple` endpoint
## 🎯 What needs to be done?
Implement the interface for a new endpoint in `UserApi` to login with Apple.
We don't know the interface of this endpoint yet, so until we have more details, just create a method that does nothing.

## 📝 Detailed explanation
The method should look like this:

`public async Task<IUser> GetWithApple()`

juanlaube:
## What's this?
Adds dummy methods that `throw` to implement the api contract for sign in and sign up with Apple

### Relationships

Closes #6910 
Closes #6911 

## Why do we want this?
🍎 

## How is it done?
It's just a couple of methods that throw, we will replace them later with an actual implementation.

### Why not another way?
This is the way.

## :squid: Permissions
🦑


Toggl.Networking.Tests/ApiClients/UserApiTests.cs
```diff
@@ -84,6 +84,12 @@ protected override Task<IUser> CallEndpoint(IUserApi api)
                 => api.GetWithGoogle();
         }
 
+        public class TheGetWithAppleMethod : FailOnPurpose
+        {
+            protected override Task<IUser> CallEndpoint(IUserApi api)
+                => api.GetWithGoogle();
```
daividssilverio: Shouldn't it be `GetWithAppleId` or something like that?

PR #7088, opened 2020-02-12
Implement AppleID token provider

### LINKED ISSUE #6912: Add `IAppleSignInProvider` interface
## 🎯 What needs to be done?
Create a new interface, `IAppleSignInProvider`, with one method, `IObservable<string> GetAppleSignInToken`, similar to `IGoogleTokenProvider`.

## 📝 Detailed explanation
`IView` should implement this new interface, and because of that also `ReactiveViewController`, `ReactiveTableViewController`, `ReactiveActivity` and `ReactiveFragment`, in all those cases throw `InvalidOperationException` for now.

This observable will be used by `LoginViewModel` and `SignUpViewModel` to get the token and login or signup with Apple.

### LINKED ISSUE #6916: Request Authorization with Apple ID
## 🎯 What needs to be done?
Implement `IAppleSignInProvider` on `ReactiveViewController` to make the request to sign in with Apple using `ASAuthorizationAppleIDProvider` and the email and full name scopes.
On the success and error handlers, pass the sign in data using the `GetAppleSignInToken` observable.

juanlaube:
## What's this?
Implements the AppleID token provider, following the same architecture as the Google token provider.

### Relationships

Closes #6912 
Closes #6916 

## Why do we want this?
This is the way to get the JWT we will send to backend.

## How is it done?
Using the `AuthenticationServices` framework to sign in with Apple.

### Why not another way?
_Mandalorian intensifies_

### Side effects
None.

## Review hints
If you want to test it, replace the line 197 in `LoginViewModel.cs` with `GetAppleToken` and use the button to login with Google, you need to sign in with your AppleID on the device/simulator first.

## :squid: Permissions
🦑


Toggl.Core/Exceptions/ThirdPartyLoginException.cs
```diff
@@ -1,11 +1,11 @@
 ﻿using System;
 namespace Toggl.Core.Exceptions
 {
-    public sealed class GoogleLoginException : Exception
+    public sealed class ThirdPartyLoginException : Exception
     {
         public bool LoginWasCanceled { get; }
 
-        public GoogleLoginException(bool loginWasCanceled)
+        public ThirdPartyLoginException(bool loginWasCanceled)
```
sofokli1: how can we distinguish between apple exception and google exception?

Toggl.iOS/ViewControllers/Reactive/ReactiveViewController.ThirdPartyTokenProvider.cs
```diff
@@ -0,0 +1,132 @@
...
+    {
+        private const int cancelErrorCode = -5;
+
+        private bool loggingIn;
+        private Subject<string> tokenSubject = new Subject<string>();
+        private ASAuthorizationAppleIdProvider appleIdProvider;
+
+        public IObservable<string> GetToken(ThirdPartyLoginProvider provider)
+        {
+            switch (provider)
```
heytherewill: You can hoist the return using pattern matching and EBM, but I won't block you further

PR #7090, opened 2020-02-13
Fix requesting calendar permission from the settings app

arturs105:
## What's this?
This fixes the problem described in this [comment](https://github.com/toggl/mobileapp/pull/6942#pullrequestreview-357011225).

Basically the problem was:
1. Deny the calendar permission and check "do not ask again"
2. Go to the calendar settings
3. Try to enable the calendar integration by tapping on the switch (not the whole cell)
4. System settings are opened
5. Go back to the Toggl app without changing any permissions
6. The switch is no in the on state even though the calendar integration is not enabled

This PR fixes it.

### Relationships

Ref #6942

## :squid: Permissions
🦑


heytherewill:
Approving the technical bit; this still needs someone to functional test it

Issue #7079, opened 2020-02-10
Object reference not set to an instance of an object CalendarCollectionViewSource.LayoutAttributesForItemAtIndexPath


sofokli1:
## 💥 Crash Report
Seems to be related to layout in the calendar 

### 👥 Impact
717 in last 30 days

### 🔗 Context
Ref https://github.com/toggl/mobileapp/issues/5432 https://github.com/toggl/mobileapp/issues/4304 https://github.com/toggl/mobileapp/issues/6722

### 🛰 Link
https://appcenter.ms/orgs/Toggl/apps/Toggl-iOS/crashes/errors/3672084728u/overview

### 📝  Exception & Stack Trace
`SIGABRT: Object reference not set to an instance of an object`
```cs
CalendarCollectionViewSource.LayoutAttributesForItemAtIndexPath (Foundation.NSIndexPath indexPath)
CalendarCollectionViewLayout.frameForEditingHour (Foundation.NSIndexPath indexPath)
CalendarCollectionViewLayout.LayoutAttributesForSupplementaryView (Foundation.NSString kind, Foundation.NSIndexPath indexPath)
(wrapper managed-to-native) ObjCRuntime.Messaging.objc_msgSend(intptr,intptr)
UIView.LayoutIfNeeded ()
CalendarDayViewController.showContextualMenu ()
CalendarDayViewController.<ViewDidLoad>b__20_6 (System.Boolean _)
(wrapper other) System.Object.gsharedvt_out()
AnonymousSafeObserver`1[T].OnNext (T value)
(wrapper other) System.Object.gsharedvt_in()
Sink`1[TTarget].ForwardOnNext (TTarget value)
(wrapper other) System.Object.gsharedvt_out()
Where`1+Predicate+_[TSource].OnNext (TSource value)
(wrapper other) System.Object.gsharedvt_in()
Sink`1[TTarget].ForwardOnNext (TTarget value)
(wrapper other) System.Object.gsharedvt_out()
ObserveOnObserverNew`1[T].DrainStep (System.Collections.Concurrent.ConcurrentQueue`1[T] q)
(wrapper other) System.Object.gsharedvt_in()
(wrapper other) System.Object.gsharedvt_out()
ObserveOnObserverNew`1[T].DrainShortRunning (System.Reactive.Concurrency.IScheduler recursiveScheduler)
(wrapper other) System.Object.gsharedvt_in()
(wrapper other) System.Object.gsharedvt_out()
[...]

PR #7093, opened 2020-02-13
🐛 Fix null selected indexPath

### LINKED ISSUE #7078: Value cannot be null. Parameter name: indexPath CalendarCollectionViewEditItemHelper.panBegan
## 💥 Crash Report
Some indexPath var is null while using `CalendarCollectionViewEditItemHelper`

This might be fixed in the 2.11 release, if so close the issue.

### 👥 Impact
48 in last 30 days

### 🛰 Link
https://appcenter.ms/orgs/Toggl/apps/Toggl-iOS/crashes/errors/2298738814u/overview

`SIGABRT: Value cannot be null. Parameter name: indexPath`
```cs
UICollectionView.CellForItem (Foundation.NSIndexPath indexPath)
CalendarCollectionViewEditItemHelper.panBegan (CoreGraphics.CGPoint point)
CalendarCollectionViewEditItemHelper.onPan (UIKit.UIPanGestureRecognizer gesture)
UIPanGestureRecognizer+Callback.Activated (UIKit.UIPanGestureRecognizer sender)
```

sofokli1:
## What's this?
This `public NSIndexPath IndexPathForSelectedItem => indexPathFor(selectedItemId);` in `CalendarCollectionViewSource` can return `null` as indexPath and that `null` is used to look for a specific cell in CalendarCollectionViewEditItemHelper, which ends up in crashes. I make sure in all situations where `IndexPathForSelectedItem` is used to check it it is null or not.

### Relationships
Closes #7078 #7079

## Why do we want this?
Less crashes

## How is it done?
Checking for null and early returns.

### Side effects
Not that I know of.

## Review hints
You can force a crash by making `public NSIndexPath IndexPathForSelectedItem => indexPathFor(selectedItemId);` in `CalendarCollectionViewSource` always return null, in the 2.11 branch and this branch to see the difference. In real life situation it is not that easy to reproduce. crating, deleting and editing TEs very fast during sync may help with recreating the crash. (use `Network link Conditioner` to simulate slow internet)

## :squid: Permissions
Only @arturs105


Toggl.iOS/ViewSources/CalendarCollectionViewSource.cs
```diff
@@ -132,7 +132,14 @@ public override UICollectionReusableView GetViewForSupplementaryElement(UICollec
             else if (elementKind == CalendarCollectionViewLayout.EditingHourSupplementaryViewKind)
             {
                 var reusableView = collectionView.DequeueReusableSupplementaryView(elementKind, editingHourReuseIdentifier, indexPath) as EditingHourSupplementaryView;
-                var attrs = LayoutAttributesForItemAtIndexPath(IndexPathForSelectedItem);
+
+                var selectedindexPath = IndexPathForSelectedItem;
```
arturs105: `camelCase` instead of `camelcase` pls.

PR #7094, opened 2020-02-13
Add sign in and sign up buttons

### LINKED ISSUE #6914: Add the Sign in with Apple button in the Login view
## 🎯 What needs to be done?
In order to add the button, we need to:
1. Add the required app entitlement
1. Link AccountServices.framework
1. Finally add the Sign in with Apple button

Pressing on the button should call `AppleLogin` on `LoginViewModel`, similar to what we already do with Google.

Documentation about the required entitlement can be found here: https://developer.apple.com/documentation/bundleresources/entitlements/com_apple_developer_applesignin

### LINKED ISSUE #6915: Add Sign in with Apple button in the Sign up view
## 🎯 What needs to be done?
Add the sign in with Apple button in the Sign up view (unfortunately we cannot change the title).
Pressing on the button should execute the `AppleSignUp` action on `SignUpViewModel`, similar to what we already do with Google.

juanlaube:
## What's this?
Adds the sign in and sign up buttons for sign in with Apple, with the corresponding logic to call the API.

### Relationships

Closes #6914 
Closes #6915 

## Why do we want this?
🍎 🔑 

## How is it done?
Adding the buttons and a bit of refactor in user access manager to support both providers.

### Why not another way?
This is the way.

### Side effects
None.

## :squid: Permissions
🦑


Toggl.iOS/ViewControllers/SignupViewController.cs
```diff
@@ -56,6 +55,8 @@ public override void ViewDidLoad()
 
             prepareViews();
 
+            setupSignInWithAppleButton();
```
sofokli1: no `TraitCollectionDidChange` in this file?

PR #7105, opened 2020-02-17
Avoid reloading all items when closing contextual menu

### LINKED ISSUE #6748: Calendar contextual menu dismissal triggers strange cell animation
## 🐛 Bug Report
When I close the contextual menu (bottom bar), either cancel or save, all calendar cells animate strangely, as if they start too large and have to be scaled down.

## Context
Build/Branch: 2.7
Phone OS: ios
Phone model: 11 pro

## Expected behavior
Just dismiss and maybe perform needed changes to the selected cell where the changes are performed on.

## To reproduce
Check description

## Suggestions
May be related to some reload data calls in the viewSource

juanlaube:
## What's this?
Fixes a glitch in the calendar view when closing the contextual menu, it caused all the cells to reload.

### Relationships

Closes #6748 

## Why do we want this?
It doesn't look great otherwise.

## How is it done?
By reloading only the cell for the selected item and some supplementary items.

### Why not another way?
The proper solution would be to use do diffing and changes in the collection view, but it's a bigger refactor and we decided not to do that because of Aurora.

### Side effects
Didn't found any, but please do functional testing in case I missed something, the calendar is complex :)

## :squid: Permissions
🦑


Toggl.iOS/ViewSources/CalendarCollectionViewSource.cs
```diff
@@ -210,10 +210,15 @@ public void StartEditing(CalendarItem calendarItem)
 
         public void StopEditing()
         {
+            if (!IsEditing)
+                return;
+
             IsEditing = false;
             layout.IsEditing = false;
+            layoutAttributes = calculateLayoutAttributes();
+            collectionView.ReloadItems(new [] { IndexPathForSelectedItem });
```
sofokli1: `IndexPathForSelectedItem` can be `null`, this was one of the main crashes in 2.10. Please check for `null`

juanlaube:
@sofokli1 take a look again please, I found some crashes on the way that I also fixed, mostly null refs and indices out of bounds.

Toggl.iOS/ViewSources/CalendarCollectionViewSource.cs
```diff
@@ -179,7 +179,7 @@ public List<DateTimeOffset> AllItemsStartAndEndTime()
 
         public CGRect? FrameOfEditingItem()
         {
-            if (!IsEditing)
+            if (!IsEditing || IndexPathForSelectedItem == null)
```
sofokli1: wouldn't it be better to have a local var for `IndexPathForSelectedItem` so that when used in the `return` we are sure that the value has not changed in the mean time in some kind of race condition?

Toggl.iOS/ViewSources/CalendarCollectionViewSource.cs
```diff
@@ -216,7 +216,10 @@ public void StopEditing()
             IsEditing = false;
             layout.IsEditing = false;
             layoutAttributes = calculateLayoutAttributes();
-            collectionView.ReloadItems(new [] { IndexPathForSelectedItem });
+            if (IndexPathForSelectedItem != null)
```
sofokli1: same as the other one comment

Toggl.iOS/ViewSources/CalendarCollectionViewSource.cs
sofokli1: I know it sounds like a stretch but it would make it easier to think about future bugs/crashes if we know for sure that this cases are not to blame. If you think about it, we shouldn't even have to check for null if it was working correctly. Yet we need to.

juanlaube: I extracted to a local var in the scope of the function, but having a local var in the scope of class is exactly what we removed in a previous refactor due to lot of crashes, so it's not exactly a good idea.

sofokli1: by local var I meant scope of the func. bad wording on my part

PR #7106, opened 2020-02-17
Update androidx libs

daividssilverio:
## What's this?
This is an update to the androidX libraries. They went stable.

### Relationships
Closes Nothing, but since I'm planning to use `ViewPager2` in #7000 

## Why do we want this?
Because we were using the release candidates. Let's update it.

## How is it done?
I've pressed the update buttons and fixed some references in the droid `csproj` file.

### Why not another way?
This is the way.

### Side effects
There shouldn't be any.

## Review hints
Make sure you can still build.

## :squid: Permissions
🦑 it already.


Toggl.Droid/Toggl.Droid.csproj
```diff
@@ -159,6 +159,182 @@
     <Reference Include="Newtonsoft.Json">
       <HintPath>..\packages\Newtonsoft.Json.12.0.3\lib\netstandard2.0\Newtonsoft.Json.dll</HintPath>
     </Reference>
+    <Reference Include="Xamarin.AndroidX.Activity, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null">
+      <HintPath>..\packages\Xamarin.AndroidX.Activity.1.0.0\lib\monoandroid90\Xamarin.AndroidX.Activity.dll</HintPath>
+      <Private>True</Private>
```
pdrj: What's the deal with all the `<Private>` stuff? Why was it added now and what are the implications? I did some googling, but not being a xamarin master, I can't say if this should stay or not.

daividssilverio: I don't know, I just pressed the buttons to update. Well, it wasn't there before, so I think it's safe to remove. @heytherewill could know better, but I'll remove it if there's no feedback from him.

daividssilverio: I've removed the `<Private>True</Private>` as @heytherewill said it's ok to do so :) @pdrj

PR #7115, opened 2020-02-19
Create new Onboarding VM

### LINKED ISSUE #6998: Create Onboarding ViewModel
This is the crossplatform part of the onboarding. It has to include actions for social sign in (just Google for now) and a way to navigate to the email sign up / login flow. No actual onboarding for now, this will be added later. Implement or reuse current interactors where possible.

In the case of login, this will take the user directly to the timeline. In the case of sign in, it'll take him to a country selector, TOS check screen that's being done in another issue. This screen will probably be a modal dialog and won't requiere its own ViewModel.

In the case of email sign in, we'll direct the user to sign up or login, depending on factors. This is done in another issue as well.

Most signup/login logic should be taken from the current implementation, but bear in mind that this time signup/login can be done from two different screens with different ViewModels, so ideally we would move shared logic to interactors.

dshr:
## What's this?
A brand new onboarding VM, complete with a new Terms and Country VM.

### Relationships
Closes #6998

## Why do we want this?
We need this for the refactor

## How is it done?
✨

## Review hints
Commit by commit works best

## :squid: Permissions
All yours!


Toggl.Core.Tests/TermsAndCountryViewModelTests.cs
```diff
@@ -0,0 +1,215 @@
...
+
+                tryingToConstructWithEmptyParameters
+                    .Should().Throw<ArgumentNullException>();
+            }
+        }
+
+        public sealed class TheInitializeMethod : TermsAndCountryViewModelTest
+        {
+            [Fact, LogIfTooSlow]
+            public async Task GetstheCurrentLocation()
```
arturs105:
```suggestion
            public async Task GetsTheCurrentLocation()
```

Toggl.Core.UI/ViewModels/TermsAndCountryViewModel.cs
```diff
@@ -0,0 +1,163 @@
...
+        public IObservable<bool> IsCountryErrorVisible { get; }
+
+        public int IndexOfPrivacyPolicy { get; }
+        public int IndexOfTermsOfService { get; }
+        public string FormattedDialogText { get; }
+        public ViewAction PickCountry { get; }
+        public ViewAction ViewTermsOfService { get; }
+        public ViewAction ViewPrivacyPolicy { get; }
+        public ViewAction Accept { get; }
+        public ViewAction Cancel { get; }
```
arturs105: These methods are lacking tests

Toggl.Core.UI/ViewModels/OnboardingViewModel.cs
```diff
@@ -0,0 +1,189 @@
...
+using Toggl.Shared;
+using Toggl.Shared.Extensions;
+using Toggl.Shared.Models;
+using Toggl.Storage.Settings;
+
+namespace Toggl.Core.UI.ViewModels
+{
+    public class OnboardingViewModel : ViewModel
+    {
+        private readonly ISchedulerProvider schedulerProvider;
```
arturs105: The field is not necessary

Toggl.Core.Tests/UI/ViewModels/OnboardingViewModelTests.cs
```diff
@@ -0,0 +1,374 @@
...
+                LastTimeUsageStorage
+                    .Received()
+                    .SetLogin(Arg.Is(now));
+            }
+        }
+
+        public sealed class ContinueWithEmail : OnboardingViewModelTest
+        {
+            [Fact, LogIfTooSlow]
+            public void GoesToLogInWhenNoPreviousLoginHasBeenDetected()
```
daividssilverio:
```suggestion
            public void GoesToSignupWhenNoPreviousLoginHasBeenDetected()
```

Toggl.Core.UI/ViewModels/OnboardingViewModel.cs
```diff
@@ -0,0 +1,189 @@
...
+                    .Subscribe(_ => onAuthenticated(), onGoogleLoginFailure, onCompleted);
+        }
+
+        private async void onAuthenticated()
+        {
+            lastTimeUsageStorage.SetLogin(timeService.CurrentDateTime);
+
+            interactorFactory.GetCurrentUser().Execute()
+                .Select(u => u.Id)
+                .Subscribe(analyticsService.SetAppCenterUserId);
```
daividssilverio: This subscription isn't being handled (not explicitly disposed).
Is there a reason why not?

daividssilverio: This subscription isn't being handled (not explicitly disposed).
Is there a reason why not?

Toggl.Core.UI/ViewModels/TermsAndCountryViewModel.cs
dshr: All of these are rather flat, and either explicitly call `Browser.OpenAsync` or just close the VM.

Toggl.Core.UI/ViewModels/OnboardingViewModel.cs
```diff
@@ -0,0 +1,187 @@
...
+            else
+            {
+                return Navigate<LoginViewModel, CredentialsParameter>(CredentialsParameter.Empty);
+            }
+        }
+
+        private void tryLoggingInWithGoogle()
+        {
+            isLoadingSubject.OnNext(true);
+            disposeBag = View?
```
daividssilverio: Why not simply go with `.DisposedBy(disposeBag)` and dispose it when the view is destroyed or something like that?
I see you are overriding a single `disposeBag` on `tryLoggingInWithGoogle` and `onGoogleLoginFailure`. That looks error prone. 

I'm not sure what would happen if, for example, the user manages to click quickly in the `continue with google` button.

Toggl.Core.UI/ViewModels/OnboardingViewModel.cs
```diff
@@ -0,0 +1,187 @@
...
+        {
+            onCompleted();
+
+            var e = exception as GoogleLoginException;
+            if (e == null)
+            {
+                onLoginError(exception);
+                return;
+            }
+            else if (e.LoginWasCanceled == true) return;
```
daividssilverio: IMHO the `else` and the `== true` aren't adding much here. Feel free to leave it as is if you like it this way.

Toggl.Core.UI/ViewModels/OnboardingViewModel.cs
```diff
@@ -0,0 +1,192 @@
...
+            {
+                analyticsService.UnknownSignUpFailure.Track(exception.GetType().FullName, exception.Message);
+                analyticsService.TrackAnonymized(exception);
+            }
+            else if (e.LoginWasCanceled) return;
+
+            View.Alert(Resources.Oops, Resources.GenericSignUpError, Resources.Ok);
+        }
+
+        private void onCompleted()
```
daividssilverio: I'm assuming this will cause the app to crash after the first time it's called. Isn't that the case?
I think the `ViewDestroyed` callback is good place to dispose the `disposeBag`.

dshr: Oh, ofc 🤦‍♂

Toggl.Core.UI/ViewModels/OnboardingViewModel.cs
```diff
@@ -28,7 +29,7 @@ public class OnboardingViewModel : ViewModel
         private readonly ILastTimeUsageStorage lastTimeUsageStorage;
         private readonly IInteractorFactory interactorFactory;
 
-        private IDisposable disposeBag;
+        private CompositeDisposable disposeBag;
```
arturs105:
```suggestion
        private CompositeDisposable disposeBag = new CompositeDisposable();
```

PR #7117, opened 2020-02-19
Prevent phantom calendar items which create artificial columns

### LINKED ISSUE #6972: Time entry creation in the calendar visually divides the time entry being created when you go and come back from the edit view
How to replicate:
- Long press to create a time entry
- Click `Edit` in the contextual menu
- Press back in the edit view that appears
- You should go back to the calendar view and the time entry you were working with should be divided now.

It should look like this:
image

pdrj:
## What's this?
This makes sure that in calendar view there are no cached items that won't get rendered, but will still be taken into consideration when computing columns.

### Relationships

Closes #6972 

## Why do we want this?
UX.

## How is it done?
Checking for such items in `onDraw`.

### Why not another way?
This is not a great solution, but other things I had in mind would take much more effort. Since we're doing the rewrite...

### Side effects
I hope not.

## Review hints
Follow the repro steps from the issue and validate it no longer happens.
Also please try to interact with calendar a bit to make sure all other stuff works.

## :squid: Permissions
Anyone.


Toggl.Droid/Views/Calendar/CalendarDayView.cs
```diff
@@ -247,6 +247,14 @@ public override bool CanScrollVertically(int direction)
 
         protected override void OnDraw(Canvas canvas)
         {
+            var invalidItemsCount = calendarItems.Count(item => item.Id == "");
+            if (!itemEditInEditMode.IsValid && invalidItemsCount > 0)
+            {
+                var validItems = calendarItems.Where(item => item.Id != "").ToImmutableList();
```
daividssilverio: 🤔 Maybe instead of doing this `OnDraw`, it would be better to have a field for these `validItems` that are populated every time the `calendarItems` are changed.

pdrj: Yeah, would be better. Thanks.

PR #7120, opened 2020-02-19
Implement endpoints used for Sign in with Apple

### LINKED ISSUE #6918: Implement final `SignUpWithApple` endpoint
## 🎯 What needs to be done?

This depends on backend, we don't know the interface of this endpoint yet.
The issue will be updated when we have more information.

### LINKED ISSUE #6919: Implement final `GetWithApple` endpoint
## 🎯 What needs to be done?

This depends on backend, we don't know the interface of this endpoint yet.
The issue will be updated when we have more information.

juanlaube:
## What's this?
This implements the actual sign in and sign up endpoints for Sign in with Apple.

### Relationships

Closes #6918 
Closes #6919 

## Why do we want this?
🍎 🔑 

## How is it done?
It's two new requests, to `/me` and `/signup`. Because of how sign in with Apple works, we need to send the client id to the API, and backend decided to do this in the `Referer` header because they do the same for webapp.

### Why not another way?
This is the way.

### Side effects
None.

## :squid: Permissions
🦑


juanlaube:
@sofokli1 it should work now, can you give me more info about the errors? Perhaps adding a few  breakpoints `BaseApi

sofokli1:
- in signup request
refere: `daneel.debug`

- in response
raw data:`invalid token:  signature_verification_failed`
status code: `ServiceUnavailable`

does this help or do you need more?

PR #7122, opened 2020-02-20
🚧Add email login VM

### LINKED ISSUE #7006: Create email login ViewModel
This will take some logic from the current login implementation, but share some other with the new social signup screen, that we'll be put in interactors.

It'll take the email and password from the user, validate the data and then try to log him in. If it succeeds it'll take the user to the timeline, otherwise it should show the appropriate error.

Designs and full specs: https://www.notion.so/toggl/Designs-wip-d767ebc157c542c790e1bc93eeaf006f#835299ea40344044b2572f356b4f47c1

sofokli1:
## What's this?
As suggested in the issues by @daividssilverio, this PR adds the email login VM as it was in the LoginVM file without the google sign in logic. Also renames the tests and removes the google login related tested to work for the new VM. I kept the old VM so that the project doesn't break. When we migrate the UI we can just delete the old VM.

### Relationships
Closes #7006

## :squid: Permissions
🦑


Toggl.Core.UI/ViewModels/EmailLoginViewModel.cs
```diff
@@ -0,0 +1,250 @@
...
+
+            loginDisposable =
+                userAccessManager
+                    .Login(emailSubject.Value, passwordSubject.Value)
+                    .Track(analyticsService.Login, AuthenticationMethod.EmailAndPassword)
+                    .Subscribe(_ => onAuthenticated(), onError, onCompleted);
+        }
+
+        public void TogglePasswordVisibility()
+            => isPasswordMaskedSubject.OnNext(!isPasswordMaskedSubject.Value);
```
heytherewill: Is this still needed? I think both platforms have an UI built-in way of handling password visibility now

sofokli1: I do not know if they have, but shouldn't that logic be in the VM?

Toggl.Core.UI/ViewModels/EmailLoginViewModel.cs
```diff
@@ -0,0 +1,241 @@
...
+using Toggl.Storage.Settings;
+
+namespace Toggl.Core.UI.ViewModels
+{
+    [Preserve(AllMembers = true)]
+    public sealed class EmailLoginViewModel : ViewModelWithInput<CredentialsParameter>
+    {
+        [Flags]
+        public enum ShakeTargets
+        {
```
daividssilverio: Are we still shaking input fields in the new screens?

sofokli1: I do not know. @Odrakir Are we shaking things here?

arturs105: I discussed the same thing for the sign up view with Janika. We agreed to shake only the e-mail field if any errors occur.

arturs105: So i 🔥ed all the `ShakeTargets` code from sign up.

Toggl.Core.UI/ViewModels/EmailLoginViewModel.cs
```diff
@@ -0,0 +1,241 @@
...
+        private readonly BehaviorSubject<Email> emailSubject = new BehaviorSubject<Email>(Shared.Email.Empty);
+        private readonly BehaviorSubject<Password> passwordSubject = new BehaviorSubject<Password>(Shared.Password.Empty);
+
+        public IObservable<string> Email { get; }
+        public IObservable<string> Password { get; }
+        public IObservable<bool> HasError { get; }
+        public IObservable<bool> IsLoading { get; }
+        public IObservable<bool> LoginEnabled { get; }
+        public IObservable<ShakeTargets> Shake { get; }
+        public IObservable<string> ErrorMessage { get; }
```
arturs105: Don't think one error observable is enough though. The designs show error messages in 3 different places:
* E-mail input error below the e-mail field
* Password input error below the password field
* Error returned from API a bit lower than the Password error

Granted, the email and password errors are shown only in the sign up view designs and not in the login view designs, but it's safe to assume that such similar views would behave similarly as well.

Toggl.Core.UI/ViewModels/EmailLoginViewModel.cs
```diff
@@ -0,0 +1,241 @@
...
+        private readonly Subject<ShakeTargets> shakeSubject = new Subject<ShakeTargets>();
+        private readonly Subject<bool> isShowPasswordButtonVisibleSubject = new Subject<bool>();
+        private readonly BehaviorSubject<bool> isLoadingSubject = new BehaviorSubject<bool>(false);
+        private readonly BehaviorSubject<string> errorMessageSubject = new BehaviorSubject<string>("");
+        private readonly BehaviorSubject<Email> emailSubject = new BehaviorSubject<Email>(Shared.Email.Empty);
+        private readonly BehaviorSubject<Password> passwordSubject = new BehaviorSubject<Password>(Shared.Password.Empty);
+
+        public IObservable<string> Email { get; }
+        public IObservable<string> Password { get; }
+        public IObservable<bool> HasError { get; }
```
arturs105: Based on the comment below, the `HasError` property doesn't make much sense.

Toggl.Core.UI/ViewModels/EmailLoginViewModel.cs
sofokli1: so do you think wee need 3 IObservable<string>?

Toggl.Core.UI/ViewModels/EmailLoginViewModel.cs
sofokli1: do we need 3 of these flags now?

Toggl.Core.UI/ViewModels/EmailLoginViewModel.cs
arturs105: Yes, i think it makes the most sense.

Toggl.Core.UI/ViewModels/EmailLoginViewModel.cs
arturs105: Don't think the flags are necessary at all, since the same for each error observable can be accomplished in the view controller with `ViewModel.EmailError.Where(string.IsNutNullOrEmpty)`.

Toggl.Core.UI/ViewModels/EmailLoginViewModel.cs
sofokli1: what about something like shakeTarget but for errors?

Toggl.Core.UI/ViewModels/EmailLoginViewModel.cs
```diff
@@ -155,20 +155,25 @@ public void SetIsShowPasswordButtonVisible(bool visible)
...
+                shakeEmailSubject.OnNext(Unit.Default);
+                return;
+            }
+            else if (!emailSubject.Value.IsValid)
             {
-                shakeSubject.OnNext(shakeTargets);
+                emailErrorMessageSubject.OnNext(Resources.InvalidEmailError);
+                shakeEmailSubject.OnNext(Unit.Default);
                 return;
             }
```
arturs105: Password should have similar errors as well: `Resources.NoPasswordError` and `Resources.InvalidPasswordError`.

Toggl.Core.UI/ViewModels/EmailLoginViewModel.cs
```diff
@@ -222,12 +227,12 @@ private void onError(Exception exception)
             switch (exception)
             {
                 case UnauthorizedException forbidden:
-                    errorMessageSubject.OnNext(Resources.IncorrectEmailOrPassword);
+                    passwordErrorMessageSubject.OnNext(Resources.IncorrectEmailOrPassword);
```
arturs105: I think any error that comes from the API should go through `loginErrorMessageSubject`: https://www.notion.so/toggl/Designs-d767ebc157c542c790e1bc93eeaf006f#4a9ef1322e5242ccbb72e0eeb6316e21

Toggl.Core.UI/ViewModels/EmailLoginViewModel.cs
```diff
@@ -96,7 +93,7 @@ public enum ShakeTargets
             Signup = rxActionFactory.FromAsync(signup);
             ForgotPassword = rxActionFactory.FromAsync(forgotPassword);
 
-            Shake = shakeSubject.AsDriver(this.schedulerProvider);
+            ShakeEmail = shakeEmailSubject.AsDriver(this.schedulerProvider);
```
arturs105: 🤔  I think we could manage without `shakeEmailSubject` by emitting from `ShakeEmail` whenever there's any error on any of the 3 error message observables, but this is fine as well.

Toggl.Core.UI/ViewModels/EmailLoginViewModel.cs
sofokli1: is `InvalidPasswordError` returned from the API?

arturs105: Nope. `InvalidPasswordError` can be detected locally - it's when the password is shorter than 6 characters. 🤔 Maybe the resource title is a bit misleading.

Toggl.Core.UI/ViewModels/EmailLoginViewModel.cs
```diff
@@ -0,0 +1,252 @@
...
+            {
+                emailErrorMessageSubject.OnNext(Resources.InvalidEmailError);
+                shakeEmailSubject.OnNext(Unit.Default);
+                return;
+            }
+
+            if (passwordSubject.Value.IsEmpty)
+            {
+                passwordErrorMessageSubject.OnNext(Resources.NoPasswordError);
+                return;
```
arturs105: You need to remove the `return` statements if the email is invalid - currently the password error shows only if a valid e-mail is entered. We want to show both errors at the same time. This is how i did in the `SignUpViewModel`:
```
if (Email.Value.IsEmpty)
    emailErrorSubject.OnNext(Resources.NoEmailError);
else if (!Email.Value.IsValid)
    emailErrorSubject.OnNext(Resources.InvalidEmailError);
else
    emailErrorSubject.OnNext(String.Empty);

if (Password.Value.IsEmpty)
    passwordErrorSubject.OnNext(Resources.NoPasswordError);
else if (!Password.Value.IsValid)
    passwordErrorSubject.OnNext(Resources.InvalidPasswordError);
else
    passwordErrorSubject.OnNext(string.Empty);

//credentialsAreValid is a get property that returns Email.Value.IsValid && Password.Value.IsValid;
if (!credentialsAreValid)  return;
```

sofokli1: done

Toggl.Core.UI/ViewModels/EmailLoginViewModel.cs
```diff
@@ -43,6 +43,9 @@ public sealed class EmailLoginViewModel : ViewModelWithInput<CredentialsParamete
         private readonly BehaviorSubject<Email> emailSubject = new BehaviorSubject<Email>(Shared.Email.Empty);
         private readonly BehaviorSubject<Password> passwordSubject = new BehaviorSubject<Password>(Shared.Password.Empty);
 
+        private bool credentialsAreValid
+             => emailSubject.Value.IsValid && !passwordSubject.Value.IsEmpty;
```
arturs105: Why `!passwordSubject.Value.IsEmpty` instead of `passwordSubject.Value.IsValid`?

sofokli1: Because invalid pass error should come from the api and not blocked by us before the request. In the near future we are going to change the rules for valid pass, that would block existing users from accessing their account. Furthermore, I can not guaranty how the pass validity is checked on other platforms we provide, this can lead to us not allowing someone to login on a pass that was set on an other client and considered valid there.

sofokli1: All we care is that we have a pass to send to BE in the login, it does not matter if it is valid or not because it won't login anyway.

arturs105: `password.IsValid` is literally: `password?.Length >= minimumLength`. I don't see a reason to make a web request with credentials that we _know_ are invalid and with which login will definitely fail. Checking for this without making the user wait is just good UX. **Furthermore, in the current implementation we do check whether the password is too short**. We don't show any error messages (we shake the password field), but we do not allow the user to proceed with logging in unless the password is valid.

Sure the rules for a valid password will change and _maybe_ we'll go with password validation just in the backend (we shouldn't though), but that's not how login currently works.

PR #7123, opened 2020-02-20
:arrow_up: Update WorkManager deps

pdrj:
## What's this?
This updated WorkManager deps from the RC ones to stable.

### Relationships

Closes #NaN

## Why do we want this?
Other androidx dependencies are already the stable ones.

## How is it done?
Some clicking in rider.

### Why not another way?
I don't like VS4Mac.

## Review hints
I tested that the firebase events relying on WorkManager get to firebase.
You can test the same stuff on your device.

## :squid: Permissions
Anyone.


Toggl.Droid/Toggl.Droid.csproj
```diff
@@ -160,19 +160,20 @@
       <HintPath>..\packages\Newtonsoft.Json.12.0.3\lib\netstandard2.0\Newtonsoft.Json.dll</HintPath>
     </Reference>
     <Reference Include="Xamarin.AndroidX.Room.Common, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null">
-      <HintPath>..\packages\Xamarin.AndroidX.Room.Common.2.2.2-rc1\lib\monoandroid90\Xamarin.AndroidX.Room.Common.dll</HintPath>
+      <HintPath>..\packages\Xamarin.AndroidX.Room.Common.2.2.2\lib\monoandroid90\Xamarin.AndroidX.Room.Common.dll</HintPath>
+      <Private>True</Private>
```
daividssilverio: Well, since we removed those `<Private>True</Private>`  in the other package updates, let's not add them back here.

pdrj: :D and I wonder who was the one to point it to you the last time.

PR #7131, opened 2020-02-21
Track Google Login attempts

### LINKED ISSUE #6997: Review Google Sign In
I personally haven't ever been able to Sign In with Google, not with my Toggl account and not with my personal Gmail account. Apparently there are some users with this problem too.

Analytics show this is not a big deal,  but still it'll be interesting to investigate. Maybe even adding more analytics events, mostly to user actions like tapping the Google Sign In button, which we are currently not tracking.

Someday said we should update the Google Sign In package, but I believe it's on the latest version, so I'm not sure if that was referring to using some other means of authenticating with Google...

dshr:
## What's this?
Adding tracking to the Continue With Google button.

### Relationships
Closes #6997

## Why do we want this?
See issue.

## :squid: Permissions
All yours!


Toggl.Core.UI/ViewModels/OnboardingViewModel.cs
```diff
@@ -93,6 +93,7 @@ private void continueWithGoogle()
         private Task continueWithEmail()
         {
             isLoadingSubject.OnNext(true);
+            analyticsService.ContinueWithGoogle.Track();
```
arturs105: Why is this event tracked in `continueWithEmail()`?

dshr: Me not paying enough attention 🤦‍♂

Odrakir:
Hey @dshr this is not fixing the thing about the timezone you mentioned in the issue.

dshr:
@Odrakir yeah, it was fixed in #7115, my message was concerning only the old onboarding.

PR #7150, opened 2020-03-03
Use AsDriver and DistinctUntilChanged in SignUpViewModel

arturs105:
## What's this?
Does what the title says and also adds shake behaviour based on a subject because `DistinctUntilChanged` calls make it no possible to rely on error messages to shake the email field.

## :squid: Permissions
🦑


Toggl.Core.UI/ViewModels/SignUpViewModel.cs
```diff
@@ -196,6 +199,8 @@ private void onError(Exception exception)
             if (errorHandlingService.TryHandleDeprecationError(exception))
                 return;
 
+            shakeEmailFieldSubject.OnNext(Unit.Default);
```
sofokli1: shouldn't this be in the `EmailIsAlreadyUsedException` case?

Toggl.Core.Tests/UI/ViewModels/SignUpViewModelTests.cs
```diff
@@ -162,6 +162,55 @@ public void EmitsTheOppositeValueWhenTogglePasswordVisibilityActionIsExecuted(in
...
+                    .Returns(
+                        Observable.Throw<Unit>(exception)
+                    );
+                var observer = TestScheduler.CreateObserver<Unit>();
+                ViewModel.ShakeEmailField.Subscribe(observer);
+
+                ViewModel.SignUp.Execute();
+                TestScheduler.Start();
+
+                observer.Values().Should().HaveCount(1);
```
sofokli1: can you add a test where you check that you do not get a shake. Like correct signup or other kind of exception that should not trigger a shake.

Toggl.Core.UI/ViewModels/SignUpViewModel.cs
```diff
@@ -143,7 +142,11 @@ private async Task signUp()
             else
                 passwordErrorSubject.OnNext(string.Empty);
 
-            if (!credentialsAreValid)  return;
+            if (!credentialsAreValid)
+            {
+                shakeEmailFieldSubject.OnNext(Unit.Default);
```
sofokli1: Won't this shake the email if the password is the wrong size? Isn't that a bit confusing?

arturs105: Me and Janika discussed this a while ago (when i was writing the `SignUpViewModel`) and decided to shake just the email field for any error message, be it email error, password error, or sign in error.

Toggl.Core.UI/ViewModels/SignUpViewModel.cs
arturs105: Same explanation as below - the email field is shaken for any displayed error message.

Toggl.Core.Tests/UI/ViewModels/SignUpViewModelTests.cs
```diff
@@ -209,6 +209,71 @@ public void EmitsWhenTryingToSignUpWithAlreadySignedUpEmail()
...
+                ErrorHandlingService.TryHandleDeprecationError(exception).Returns(true);
+                var observer = TestScheduler.CreateObserver<Unit>();
+                ViewModel.ShakeEmailField.Subscribe(observer);
+
+                ViewModel.SignUp.Execute();
+                TestScheduler.Start();
+
+                observer.Values().Should().BeEmpty();
+
+
```
sofokli1: 🔥, up to you

PR #7159, opened 2020-03-04
Add new analytics events

### LINKED ISSUE #7015: Add/Review analytics
Some of this will already be done, since we are taking some logic from the existing view models, but I want to have a separate issue for this because I think analytics for this specific part of the app are very important. 

I want a list of everything we are tracking in these screens (if possible confirming they are working) and maybe a list of more things we are not tracking, like UI stuff, for example.

sofokli1:
## What's this?
Just adds new analytics events

### Relationships
Closes #7015

## Why do we want this?
Analytics are good, more analytics are better

## :squid: Permissions
🦑


Toggl.Core/Analytics/IAnalyticsService.cs
```diff
@@ -155,6 +155,32 @@ public interface IAnalyticsService
...
 
+        IAnalyticsEvent EmailIsAlreadyInUsedSignUpFailure { get; }
+
+        IAnalyticsEvent<bool> LocalEmailValidationSignUpCheck { get; }
+
+        IAnalyticsEvent<bool> LocalPasswordValidationSignUpCheck { get; }
+
+        IAnalyticsEvent<bool> LocalCountryValidationSignUpCheck { get; }
+
+        IAnalyticsEvent TaCCountryNotSelected { get; }
```
juanlaube: I'm not a fan of these `TaC` prefixes, it's hard to understand what it means, I would use `TermsAndConditions` or perhaps `Onboarding` 🤔

sofokli1: Onboarding sound better

PR #7163, opened 2020-03-04
Clear rating view related settings storage

### LINKED ISSUE #7158: Rating view card appeared right after creating a new account
This is a corner case, but it can happen when an existing user signs off and a new account is created in that device (actually... it might even happen if you login back right away).

Basically, we are not cleaning the `firstAccessDateKey` and `ratingViewNumberOfTimesShownKey` keys on `IOnboardingStorage.Reset().

Because of that, if you logout after 5 days and tries to create a new account (maybe even simply signing back in), you should see the rating when even though you shouldn't.

daividssilverio:
## What's this?
This fixes #7158 

## Why do we want this?
Because it's not cool to show the rating view as soon as the user signs up/logins.

## How is it done?
By simply cleaning the settings storage keys related to the rating view.

### Why not another way?

## Review hints
To replicate the original issue:
1. Set your device date to 5+ days in the past.
2. Login in some account.
3. Logout
4. Set your device date to today.
5. Create a new account
6. Create a time entry
7. You should see the new rating view card appear

Do the same for this branch.

## :squid: Permissions
🦑 it


amulware:
I don't think this storage should ever be cleared. Where are these specs from?

daividssilverio:
Hey @amulware , there are no specs for the changes made here, just the original issue #1972 (and I didn't find anything there about this).

But, as I reported in #7158 , the rating view appeared right after signing up for a new account in the app (after stopping a time entry ofc), ignoring the expected 5 days before using the app to show the rating view.

Did I understand it wrong? The currently configured value to **wait for 5 days** to show the rating view refers to the f**irst time the user opened the app** and **not each log in/sign in event**? That makes sense now that I've looked at when the `firstAccess` key is set, which is during the app start.

If that's the case, we can close this PR and #7158 .

amulware:
Let's close it then. I agree this is maybe not 100% ideal, but it also doesn't do  much harm (users don't logout that much, and probably almost never to let another personn use the app). This change would mean that we might ask the same person twice within a couple of days, which is not what we want.

daividssilverio:
Closed it is.

PR #7165, opened 2020-03-05
Don't show fragments until the activity is started

### LINKED ISSUE #7132: MainTabBarActivity.showFragment (System.Int32 fragmentId)
## 💥 Crash Report
The most common crash in 2.11.1. Might be fixable in the usual `commitAllowingStateLoss` way. 

### 👥 Impact
Happens to about 0.05% of users. 25 times in the last 5 days.

### 🔗 Context
We've made some changes in how we handle fragments when we added placeholders. That being said, we had a similar version of the crash before, so that is unlikely to be the root cause.

### 🛰 Link
[Here](https://appcenter.ms/orgs/Toggl/apps/Toggl-Android/crashes/errors/116989627u/overview)

```cs
MainTabBarActivity.showFragment (System.Int32 fragmentId)
Java.Lang.RuntimeException: Can not perform this action after onSaveInstanceState
JniEnvironment+InstanceMethods.CallIntMethod (Java.Interop.JniObjectReference instance, Java.Interop.JniMethodInfo method, Java.Interop.JniArgumentValue* args)
JniPeerMembers+JniInstanceMethods.InvokeAbstractInt32Method (System.String encodedMember, Java.Interop.IJavaPeerable self, Java.Interop.JniArgumentValue* parameters)
FragmentTransactionInvoker.Commit ()
MainTabBarActivity.showFragment (System.Int32 fragmentId)
MainTabBarActivity.onTabSelected (Android.Views.IMenuItem item)
AsyncMethodBuilderCore+<>c.<ThrowAsync>b__7_0 (System.Object state)
SyncContext+<>c__DisplayClass2_0.<Post>b__0 ()
Thread+RunnableImplementor.Run ()
IRunnableInvoker.n_Run (System.IntPtr jnienv, System.IntPtr native__this)
(wrapper dynamic-method) Android.Runtime.DynamicMethodNameCounter.20(intptr,intptr)
java.lang.IllegalStateException: Can not perform this action after onSaveInstanceState
androidx.fragment.app.FragmentManagerImpl.checkStateLoss(Unknown Source:11)
androidx.fragment.app.FragmentManagerImpl.enqueueAction(Unknown Source:2)
androidx.fragment.app.BackStackRecord.commitInternal(Unknown Source:67)
androidx.fragment.app.BackStackRecord.commit(Unknown Source:1)
mono.java.lang.RunnableImplementor.n_run(Native Method)
mono.java.lang.RunnableImplementor.run(Unknown Source:0)
android.os.Handler.handleCallback Handler.java:873
android.os.Handler.dispatchMessage Handler.java:99
android.os.Looper.loop Looper.java:193
android.app.ActivityThread.main ActivityThread.java:6894
java.lang.reflect.Method.invoke(Native Method)
com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run RuntimeInit.java:493
com.android.internal.os.ZygoteInit.main ZygoteInit.java:883
```

daividssilverio:
## What's this?
This is yet another attempt to fix the crash when committing fragment transactions in the main tab bar.

### Relationships
Closes #7132 

## Why do we want this?
To avoid crashes when the user opens the app after leaving it in the background for a long time.

## How is it done?
By checking if it's safe to commit a fragment transaction.

### Why not another way?

### Side effects
There shouldn't be any, but please help testing it out.

## Review hints
One way to test this is to open the app, do something in one of the main tab bar fragments, then go to your home screen, forget about the app and come back in a couple of hours.

If you open the app from the launcher and it crashes, the crash wasn't fixed. If it doesn't crash, it could be fixed.

## :squid: Permissions
🦑 it


heytherewill:
Won't this do more harm than good and get us in a state where no fragments are shown, which is worse than a simple crash that the users barely perceive?

daividssilverio:
@heytherewill This crash is actually happening quite a lot to me 🤔 
I didn't manage to get into a state where no fragment is shown at all, but I could throw in a guard there that triple checks if a fragment is in place and adds one if not.

daividssilverio:
> Please correct me if I'm wrong, but the crash is saying the operation can't be completed after onSaveInstanceState. Doesn't it mean that we're trying to commit the transaction after the activity is no longer alive, hence the check you added is in a wrong place?

You are not wrong. I'm not wrong either. The crash actually happens when the fragment transaction is committed after `onSaveInstanceState`.

`Lifecycle.CurrentState.IsAtLeast(AndroidX.Lifecycle.Lifecycle.State.Started)` will only return true when:
```
 /**
         * Started state for a LifecycleOwner. For an {@link android.app.Activity}, this state
         * is reached in two cases:
         * <ul>
         *     <li>after {@link android.app.Activity#onStart() onStart} call;
         *     <li><b>right before</b> {@link android.app.Activity#onPause() onPause} call.
         * </ul>
         */
```

> I thought what was happening was await getCachedFragment(fragmentId); took too long, activity got destroyed in the meantime and then we're trying to commit the transaction.

That's what's happening I guess.

> Since you were able to reproduce this and fix it with the proposed change, I must have got it wrong - just wondering what's happening here ;).

I guess the docs above serve as an explanation, let me know if that didn't help.

pdrj:
Hm, yeah I get what you're saying. But, you first check for the state, and after that you call the async method. What if, when `showFragment` is called, the lifecycle check passes, we then block on the `await`, something happens and activity gets killed, `await` returns and then we try to commit the transaction? How does the added check handle that?

pdrj:
[Related slack thread](https://toggl.slack.com/archives/CD40QCV25/p1583417344057100)

Issue #7125, opened 2020-02-20
Onboarding view: new copy translation request


juanlaube:
**Target Languages:** 
🇯🇵 Japanese

**New Copies:**
```
<data name="OnboardingMessageVariantA" xml:space="preserve">
    <value>Simple time tracking.
Powerful reporting.</value>
    <comment>Message shown in the onboarding view</comment>
</data>
<data name="ContinueWithGoogle" xml:space="preserve">
    <value>Continue with Google</value>
    <comment>Text of the button that allows to enter your account using Google from the onboarding view</comment>
</data>
<data name="ContinueWithEmail" xml:space="preserve">
    <value>Continue with Email</value>
    <comment>Text of the button that allows to enter your account using email and password from the onboarding view</comment>
</data>
```

PR #7171, opened 2020-03-11
Add the onboarding pages

### LINKED ISSUE #7011: Add actual onboarding for iOS
This is about adding the actual onboarding steps to the onboarding screens.

Bear in mind that this will be the first screen the user will see when first opening the app, so take extra care here, everything has to be perfect, beautiful, animate when possible, load fast and generally be a pleasure to use.

We should also add analytics on what steps does the user view and more.

#6998 should be done first.

Designs and full specs: https://www.notion.so/toggl/Design-specs-for-onboarding-carousel-a55437219e074f6a869ab4f9ab8e5ff4

juanlaube:
## What's this?
Adds the onboarding pages from the design.

### Relationships

Closes #7011 

## Why do we want this?
🚪 We want more users!

## How is it done?
The new onboarding pages are subviews of the onboarding view controller, swiping happens with two gestures recognizers to go left or right.
First page contains the video we had already.
Second page has a custom view that animates the eye in the periscope.
Third page contains just an image for now.

### Why not another way?
This is the way.

### Side effects
None.

## Review hints
Try it!

## Localizations
Added new copies to #7125 

## :squid: Permissions
🚫 Not yet, a designer should review the animations first.


Toggl.iOS/Views/Onboarding/OnboardingPageView.cs
```diff
@@ -0,0 +1,80 @@
...
+        {
+        }
+
+        public override void LayoutSubviews()
+        {
+            base.LayoutSubviews();
+            videoView?.RestartVideo();
+        }
+
+        public void AddVideo(NSUrl url)
```
arturs105: What about `SetVideo` and making the `videoView` readonly?

Toggl.iOS/Views/Onboarding/PeriscopeView.cs
```diff
@@ -0,0 +1,55 @@
...
+using CoreGraphics;
+using UIKit;
+
+namespace Toggl.iOS
+{
+    public class PeriscopeView : UIView
+    {
+        private CALayer periscopeLayer;
+        private CALayer eyeballLayer;
+        private CALayer pupilLayer;
```
arturs105:
```suggestion
        private readonly CALayer periscopeLayer;
        private readonly CALayer eyeballLayer;
        private readonly CALayer pupilLayer;
```

Toggl.iOS/ViewControllers/OnboardingViewController.cs
```diff
@@ -15,8 +13,17 @@ namespace Toggl.iOS.ViewControllers
...
     public sealed partial class OnboardingViewController : ReactiveViewController<OnboardingViewModel>
     {
-        private AVPlayer player;
-        private AVPlayerLayer playerLayer;
+        private UIView containerView;
+        private OnboardingPageView page1;
+        private OnboardingPageView page2;
+        private OnboardingPageView page3;
+
+        private double duration = 0.3;
```
arturs105:
```suggestion
        private const double animationDuration = 0.3;
```

Toggl.iOS/ViewControllers/OnboardingViewController.cs
```diff
@@ -15,8 +13,17 @@ namespace Toggl.iOS.ViewControllers
...
+        private UISwipeGestureRecognizer swipeLeftGesture;
+        private UISwipeGestureRecognizer swipeRightGesture;
```
arturs105: Not a big fan of using `UISwipeGestureRecognizer`s for this, but will do for mvp.

Toggl.iOS/ViewControllers/OnboardingViewController.cs
```diff
@@ -15,8 +13,17 @@ namespace Toggl.iOS.ViewControllers
...
+
+        private int currentPage = 0;
+        private int totalPages = 3;
```
arturs105:
```suggestion
        private const int totalPages = 3;
```

Toggl.iOS/Views/Onboarding/OnboardingPageView.cs
juanlaube: `videoView` cannot be readonly because it's not set in the constructor.

Toggl.iOS/ViewControllers/OnboardingViewController.cs
juanlaube: Me neither, in another situation I would use `UIPageViewController`, but since the final designs have that wobbly effect when you swipe between pages, this is the only way to do it.

PR #7175, opened 2020-03-13
Add loading view

### LINKED ISSUE #7036: Create loading screen for iOS
After the user finishes the onboarding (signup/login, but mostly login) we need to show a loading screen while the app loads whatever data is available. This will take between 2 and 10 seconds.

Light mode: https://toggl.invisionapp.com/share/J3VWZDDQVFD#/screens/404298503
Dark mode: https://toggl.invisionapp.com/share/J3VWZDDQVFD#/screens/404298502

juanlaube:
## What's this?
Adds the loading view

### Relationships

Closes #7036 

## Why do we want this?
To indicate something is happening.

## How is it done?
A view that is added/removed depending on the loading observable, the animation is a gif.

### Why not another way?
We could also use a video for the animation, but this is the asset we had.

### Side effects
None.

## :squid: Permissions
🦑


Toggl.Core.UI/ViewModels/OnboardingViewModel.cs
```diff
@@ -95,7 +95,6 @@ private void continueWithGoogle()
 
         private Task continueWithEmail()
         {
-            isLoadingSubject.OnNext(true);
```
sofokli1: why change this logic?

juanlaube: Because we don't want to do this when using email and password, it's only for social login according to the designs.

PR #7181, opened 2020-03-16
Update onboarding copy

juanlaube:
## What's this?
Updates the copy on the first slide in the onboarding

### Relationships

Closes nothing

## Why do we want this?
We want to use the correct strings

## How is it done?
Updating Resources.resx

## :squid: Permissions
🦑


Toggl.Shared/Resources.resx
```diff
@@ -138,8 +138,8 @@ Let’s get you set up to start tracking.</value>
         <comment>Displayed in the login view right above the email text field</comment>
     </data>
     <data name="OnboardingMessagePage1" xml:space="preserve">
-        <value>Simple time tracking.
-Powerful reporting.</value>
+        <value>Time tracking so simple
+you’ll actually use it </value>
```
pdrj: Is there an extra space at the end? 🤔 Also, the animations in notion have a '.' at the end, but in the text further down the page there aren't any, so I'm not sure which is right. We can go without punctuation for now I guess.

juanlaube: Fixed! Yeah, the strings in the notion page don't have a `.` so I would go with that unless someone requests a change.

PR #7212, opened 2020-03-18
Fix ToS & Country welcome message copy

### LINKED ISSUE #7211: In ToS screen on Android, the copy is wrong. Should be as per design.
That's pretty much it.

daividssilverio:
This fixes #7211 

## Why do we want this?
Because the copy was wrong.

## :squid: Permissions
🦑 it.


pdrj:
I am looking at the desings and the copy is different in the Google Signup flow and the email flow: https://toggl.invisionapp.com/share/ANW6KYJQM6H#/screens/406975649_Toggl_-_Android_-_Continue_With_Google_-sign_Up-

Specifically:
https://toggl.invisionapp.com/share/ANW6KYJQM6H#/screens/406975636
https://toggl.invisionapp.com/share/ANW6KYJQM6H#/screens/406975649

We should agree on one way before I accept this PR. I'll ping Janika.

JanikaMandel:
@pdrj That is done on purpose. Because in Google and Apple sign up when user gets there they might not know that they are now signing up (so we welcome them), but in email flow, they already know that and it is the second step in their sign up flow (why we have "Just one more thing.")

daividssilverio:
Oh, I see, thanks for clarifying.

daividssilverio:
Well, now I think iOS has the same problem @sofokli1 @juanlaube

daividssilverio:
I'll be creating the translations request issues for the new copies.

PR #7223, opened 2020-03-20
Fix TypeFace crash on lower API versions

pdrj:
## What's this?
This fixes TypeFace crash on lower API versions (no method exception)

### Relationships

Closes #nothing

## How is it done?
Using an older constructor.

## Review hints
Try to login on api 22 and see no crash.

## :squid: Permissions
Anyone.


daividssilverio:
I'm still getting a crash, but unrelated to fonts.
```
Android.Util.AndroidRuntimeException: Only the original thread that created a view hierarchy can touch its views.
```

And... I know where it's from... It was introduced by #7219  (The `AnimatedLoadingMessage` is not running in the main thread)

PR #7224, opened 2020-03-20
Fix TextObserver crash on old and slow APIs 

pdrj:
## What's this?
This prevents making UI changes from a wrong thread.

### Relationships

Closes #

## Why do we want this?
Otherwise we get crashes on old, slow devices.

## How is it done?
Anything that modifies UI state is observed on the main scheduler.

### Why not another way?
I couldn't pinpoint exactly which one is causing the crash and each build takes about 10 minutes to install on the old device. I just handled them all to be safe.

## Review hints
Login with this on API 22 device which is slow - you should get no crashes. Then try without these changes.

## :squid: Permissions
Anyone.


daividssilverio:
I'm pretty sure the problem is caused by the  `TextHelpers.AnimatedLoadingMessage()` not being a driver.  I've introduced it in #7219 .

A simple `.AsDriver(AndroidDependencyContainer.Instance.SchedulerProvider);` in that extension should fix the problem.

PR #7242, opened 2020-03-24
Track onboarding carousel analytics

### LINKED ISSUE #7197: Track onboarding carousel events
## 🎯 What needs to be done?
We want to track when a page is scrolled in the carousel, either because it was automatically scrolled, or because the user swiped.
We also want to track one single event before leaving the onboarding view, to know which pages have been viewed.

## 📝 Detailed explanation
Add the following two events:

`OnboardingPageScroll(trigger: Manual|Automatic, direction: Left|Right, nextPageNumber:int)`
`OnboardingPagesViewed(page1:bool, page2:bool, page3:bool)`

`OnboardingPageScroll` has to be tracked on each page scroll.
`OnboardingPagesViewed` has to be tracked when pressing any of the "Continue with..." buttons.

juanlaube:
## What's this?
Tracks the onboarding carousel events.

### Relationships

Closes #7197 

## Why do we want this?
📈 Analytics!

## How is it done?
It's two new events that are tracked in the `OnboardingViewModel`, in this PR I also added the code needed for iOS.

### Side effects
None.

## :squid: Permissions
🦑


Toggl.iOS/ViewControllers/OnboardingViewController.cs
```diff
@@ -108,6 +110,13 @@ private void configureOnboardingPages()
 
             View.AddGestureRecognizer(swipeLeftGesture);
             View.AddGestureRecognizer(swipeRightGesture);
+
+            ViewModel.OnOnboardingScroll.Execute(new OnboardingScrollParameters
```
daividssilverio: 🤔 With this, the first page will always count as automatic, right?

juanlaube: Yep

daividssilverio: And we don't have a automatic scroll based on how much time has passed as of yet as well, right? Just confirming.

juanlaube: Correct, not yet.

PR #7243, opened 2020-03-24
Easen up splashscreen transition

### LINKED ISSUE #7156: Splash screen is flashing weirdly before opening the main log
When clicking to open the app, it first closes very quickly, then the login view appears (after clearing data or closing all the apps).

PS: This happened consistently yesterday, but I could not replicate the issue today. This might need some investigation.

daividssilverio:
## What's this?
This is an attempt to easen up the transitions between the splash screen and other activities.

### Relationships
Closes #7156

## Why do we want this?
Because currently it can look broken at times.

## How is it done?
By finishing the splash screen after navigating to the target activities.

### Why not another way?
Please suggest something else if have something in mind.

### Side effects
The navigation could work weirdly now... But it seemed fine from my tests.

## Review hints
Please test the app, opening and closing from the launcher, tiles and shortcuts.
From my tests, it looked fine.

## :squid: Permissions
🦑 it


Toggl.Droid/Startup/SplashScreen.cs
```diff
@@ -102,22 +103,34 @@ private void registerTimezoneChangedBroadcastReceiver(ITimeService timeService)
...
+        private Intent createIntentForNotLoggedInAccessLevel(AccessLevel accessLevel)
         {
-            var navigationService = AndroidDependencyContainer.Instance.NavigationService;
-
             switch (accessLevel)
             {
                 case AccessLevel.AccessRestricted:
-                    navigationService.Navigate<OutdatedAppViewModel>(null);
-                    return;
+                    loadAndCacheViewModelWithParams<OutdatedAppViewModel, Unit>(Unit.Default);
```
pdrj: This calls `loadAndCacheViewModelWithParams` which in turn does `viewModel.Initialize()`. Initialization is `async` and is not awaited anywhere. Nav service does await it and we should too, IMO. Please let me know if I'm missing something.

pdrj:
Uploaded an mp4 with a blink to slack.

pdrj:
I also did some functional testing and it seems to be alright 👍

Toggl.Droid/Startup/SplashScreen.cs
daividssilverio: Those VMs in specific don't have anything in their `Initialize` methods, that's why it should be safe to do that, although I should at least call `Initialize` anyway.

pdrj: `TokenResetViewModel` has something in `Initialize` 🤔

daividssilverio: Just saw that. I'll put this PR on WIP for now. We'll need to figure out another way to handle this issue.

daividssilverio: So, I've moved the initialization needed in the `TokenResetViewModel` to its constructor. Now it's safe to not wait for it o/

PR #7248, opened 2020-03-25
Fix copy, buttons and dark mode in onboarding

juanlaube:
## What's this?
Three small fixes in the onboarding:
1. the copy in the terms and country selection
2. the back button in the terms and country selection when presented in the google flow
3. the text color in the forgot password's textfield

### Relationships

none

## Why do we want this?
🐛 

## How is it done?
👢 
🐛 

### Why not another way?
This is the way.

### Side effects
My patience with this release is running out.

## :squid: Permissions
@sofokli1 only


Toggl.Shared/Resources.resx
```diff
@@ -288,7 +288,7 @@ Please try again.</value>
         <comment>@Invariant</comment>
     </data>
     <data name="ReviewTheTerms" xml:space="preserve">
-        <value>Just one more thing.
+        <value>Welcome to Toggl!
```
heytherewill: Needs to change in other languages, right?

juanlaube: We don't have this string localized yet.

juanlaube:
Android changed it in their release branch.

PR #7249, opened 2020-03-25
Fix some onboarding visuals

pdrj:
## What's this?
This fixes (from [Janika's feedback notion page](https://www.notion.so/toggl/QA-feedback-1cd7b4d830d34809b7693836d3d74485#b1ddc437bfc748738dbb23d8ac8fe5b3)):

(new 25.03) The headline is visually slightly off. I was trying to understand why, but it seems it is because the line height is not as per design. Please check if the line-height is correct. 
(new 25.03) Onboarding image is too much in the bottom, but it should be vertically centered in the free space between headline and buttons.
(new 25.03) The toggl logo in the dark mode should also have dark text, not white. Please see designs.
(new) Please update the onboarding screen buttons to also have all caps like on the other pages (I updated the designs also to follow that)

### Relationships

Closes #

## How is it done?
In a way that took so much time it's not even funny (mainly centering the red dude caused the time loss).

### Why not another way?
Regarding the centering, the chain is not needed in the end. But, if we decide to add elements to the main activity, then it'll be easier to keep togglman centered.

## Review hints
Use the current version in internal testing, or checkout the main release branch. Validate the problems listed at the top of PR exist in that app version. Then validate they don't exist on this branch.

## :squid: Permissions
Anyone.


Toggl.Droid/Resources/values-night/Colors.xml
```diff
@@ -54,5 +54,5 @@
   <color name="onboardingLoadingIndicatorColor">#32A69E</color>
   <color name="onBoardingTabIndicatorSelected">#B3000000</color>
   <color name="onBoardingTabIndicatorDeselected">#33000000</color>
-  
+
```
juanlaube: No `onboardingTogglLogoText`?

pdrj: No, it is black in both modes.

Issue #7104, opened 2020-02-17
Dump unsynced/unsyncable changes to a local file


amulware:
## Overview

In order to make transitioning to the new apps a smooth process for all users, we need to make sure unsynced/unsyncable data gets transitioned correctly.

To do that, we will push that data (and only that data) into a local file, serialised as JSON.

The affected entities are: time entries, projects, clients, tags, since these are the ones the user can create/edit in our apps.

## Format

Everything should be serialised into a single file with the following layout:
```
{ "time_entries" : [], "projects" : [], "clients" : [], "tags" : []}
```
Each of these JSON arrays will contain the unsynced/unsyncable entities of the appropriate nature, exactly the way they would be serialised when sent to our API.

The file is overwritten every time.

## Analytics

I also want an analytics event added to track this, called: `UnsyncedDataDumped` with the following properties:
- `TimeEntries` - number of time entries serialised
- `Projects` - number of time entries serialised
- `Clients` - number of time entries serialised
- `Tags` - number of time entries serialised

## Open questions

A couple of questions remain:

1. How/where to put this file? What is it called?
2. When exactly do we create this file? Every time any of these entities are created/changed? At the beginning or end of sync? This is critical because we need to balance the possible performance impact of the realm queries with making sure the file is always in the correct state.
3. How can we test this?

Please propose answers to these three questions below (discuss if needed) and ask me to sign off on them before starting implementation.

juanlaube:
My thoughts regarding the open questions:

### How/where to put this file? What is it called?
We should use the same filename in both platforms, something like `unsynced-migration-dump.json` is a good name. As for the path, I'd put it directly in the documents folder, in iOS it's the first path in `NSDocumentDirectory` and in Android it's `DIRECTORY_DOCUMENTS`.

### When exactly do we create this file?
I think a new sync step that runs in it's own queue is the best option, this way we can schedule it after pulling, pushing or even cleanup.

### How can we test this?
We could add something that triggers sharing the json file with other apps so we can inspect the contents, maybe triggered by a not-so-obvious gesture like tapping 7 times on the version number or a keyboard shortcut?

sofokli1:
### How/where to put this file? What is it called?

Same as @juanlaube 

### When exactly do we create this file?

Agree with the separate queue, but should happen every time we enter background. I see no point in doing these at any other point since there will always be a chance of syncing the data while the app is still open.

### How can we test this?

[...]

PR #7262, opened 2020-03-27
Create interactor to generate syncable data dump

daividssilverio:
## What's this?
This is one of the interactors that will build the data dump of unsynced entities that will end up in a file to be consumed in the new apps.

More details and how everything else can/will be done can be found at #7104.

### Relationships
Part of #7104

## Why do we want this?
With this interactor, we have the first part needed to create the dump files: The data we'll need.
The `SyncableDataDump` can then be serialized into json and an actual file can be created in both platforms using whatever is decided at #7104.

## How is it done?
The `CreateUnsyncedDataDumpInteractor` takes in the toggl data sources and spits out a `SyncableDataDump`, which contains a list of `TimeEntries`, `Tags`, `Projects` and `Clients` which have their sync status in either `SyncNeeded`, `SyncFailed` or `RefetchingNeeded`.

### Why not another way?
By all means, suggest something different if think this won't work.

### Side effects
None yet, we are not running this yet.

## Review hints
Check the unit tests. 
You setup some offline testing and run this interactor artificially in some view model just to see it working as well.

## :squid: Permissions
🦑 it


Toggl.Core/Models/SyncableDataDump.cs
```diff
@@ -0,0 +1,25 @@
...
+using Toggl.Core.Models.Interfaces;
+
+namespace Toggl.Core.Models
+{
+    public class SyncableDataDump
+    {
+        public IEnumerable<IThreadSafeTimeEntry> TimeEntries { get; }
+        public IEnumerable<IThreadSafeTag> Tags { get; }
+        public IEnumerable<IThreadSafeProject> Projects { get; }
+        public IEnumerable<IThreadSafeClient> Clients { get; }
```
heytherewill: Make these `IImmutableList<IThreadSafeTimeEntry>` to ensure that the lists are enumerated and can't be further changed

Toggl.Core/Models/SyncableDataDump.cs
```diff
@@ -0,0 +1,25 @@
...
+
+        public SyncableDataDump(
+            IEnumerable<IThreadSafeTimeEntry> timeEntries,
+            IEnumerable<IThreadSafeTag> tags,
+            IEnumerable<IThreadSafeProject> projects,
+            IEnumerable<IThreadSafeClient> clients)
```
heytherewill: You can still receive `IEnumerable<T>` here, just make sure to force enumeration by calling `ToImmutableList`

Toggl.Core/Interactors/Sync/CreateUnsyncedDataDumpInteractor.cs
```diff
@@ -0,0 +1,42 @@
...
+            return new SyncableDataDump(timeEntries, tags, projects, clients);
+        }
+
+        private IObservable<IEnumerable<TThreadsafe>> getUnsyncedItems<TThreadsafe, TDatabase>(
+            IDataSource<TThreadsafe, TDatabase> source)
+            where TDatabase : IDatabaseSyncable
+            where TThreadsafe : TDatabase, IThreadSafeModel
+        {
+            return source.GetAll(p => p.SyncStatus != SyncStatus.InSync);
+        }
```
heytherewill: EBM wouldn't hurt here:
```suggestion
        => source.GetAll(p => p.SyncStatus != SyncStatus.InSync);
```

Toggl.Core/Interactors/Sync/CreateUnsyncedDataDumpInteractor.cs
```diff
@@ -0,0 +1,42 @@
...
+            Ensure.Argument.IsNotNull(dataSource, nameof(dataSource));
+            this.dataSource = dataSource;
+        }
+
+        public async Task<SyncableDataDump> Execute()
+        {
+            var clients = await getUnsyncedItems(dataSource.Clients);
+            var projects = await getUnsyncedItems(dataSource.Projects);
+            var tags = await getUnsyncedItems(dataSource.Tags);
+            var timeEntries = await getUnsyncedItems(dataSource.TimeEntries);
```
heytherewill:
```suggestion
            var clients = await dataSource.Clients.GetAll(notSynced);
            var projects = await dataSource.Projects.GetAll(notSynced);
            var tags = await dataSource.Tags.GetAll(notSynced);
            var timeEntries = await dataSource.TimeEntries.GetAll(notSynced);
```

heytherewill: This is actually shorter 😂

Toggl.Core/Models/SyncableDataDump.cs
```diff
@@ -0,0 +1,25 @@
+using System.Collections.Generic;
+using Toggl.Core.Models.Interfaces;
+
+namespace Toggl.Core.Models
+{
+    public class SyncableDataDump
```
juanlaube: Shouldn't this be `UnsyncableDataDump`? 😄

juanlaube: Wouldn't hurt to add a method in this class to serialize that dump into JSON so the next parts of this issue can do the writing only.

heytherewill: Unsyncable is not accurate either; `UnsyncedDataDump` is more adequate

daividssilverio: Not really, right? This is the the dump of the data that couldn't be synced, not the data that is unsyncable.

Maybe `UnsyncedDataDump` would make more sense?

juanlaube: You are right, `UnsyncedDataDump` is 💯

daividssilverio: @juanlaube I think the serialization can be done in another PR before or alongside the services for each platform.

PR #7288, opened 2020-04-07
Add UniqueId to entities and update DB with a migration

### LINKED ISSUE #7278: Create a realm migration for UUIDs
## 🎯 What needs to be done?
A migration is needed that will add a UUID field to all entities and it needs to generate a UUID for each entity.

## 🏆 Outcome
All entities have a UUID.

### LINKED ISSUE #7279: Assign UUID to all new entities
## 🎯 What needs to be done?
Whenever a new entity is created, it needs to be assigned a UUID.

## 📝 Detailed explanation
Go thru all the entities creation process and make sure they all have an UUID.

## 🔗 Dependencies
#7278 needs to be solved first.

## 🏆 Outcome
Every entity has a GUID after creation.

kornelijepetak:
## What's this?
This PR

* Adds UniqueId to the entities
* Adds a migration that will fill in the entities currently in the database with the UUIDs
* Makes sure that the all the new entities have this UUID set up on creation

### Relationships
Closes #7278 
Closes #7279 

## Why do we want this?
Because of the new syncing where UUIDs are sent with the push actions to prevent duplicates.

### Side effects
Considering the fact that I'm not extremely familiar with all of the nuisances of all the layers of entities, make sure to note any weirdness during review.

## Tests
Does this need any particular tests?

## :squid: Permissions
Me or @simonrozsival

### DIFF
--- a/Toggl.Core/Models/Preferences.cs
+++ b/Toggl.Core/Models/Preferences.cs
@@ -1,4 +1,5 @@
-﻿using Toggl.Core.Models.Interfaces;
+﻿using System;
+using Toggl.Core.Models.Interfaces;
 using Toggl.Shared;
 using Toggl.Shared.Models;
 using Toggl.Storage;
@@ -18,6 +19,9 @@ namespace Toggl.Core.Models
 
         public const long fakeId = 0;
         public long Id => fakeId;
+
+        public static readonly string fakeUniqueId = Guid.Empty.ToString();
+        public string UniqueId => fakeUniqueId;
 
         private Preferences(IPreferences entity, SyncStatus syncStatus, string lastSyncErrorMessage, bool isDeleted = false)
             : this(entity.TimeOfDayFormat, entity.DateFormat, entity.DurationFormat, entity.CollapseTimeEntries, syncStatus, lastSyncErrorMessage, isDeleted)
--- a/Toggl.Networking.Tests/Models/TimeEntryTests.cs
+++ b/Toggl.Networking.Tests/Models/TimeEntryTests.cs
@@ -11,10 +11,10 @@ namespace Toggl.Networking.Tests.Models
         public sealed class TheTimeEntryModel
         {
             private string validJson
-                => "{\"id\":525144694,\"workspace_id\":1414373,\"project_id\":3178352,\"task_id\":null,\"billable\":false,\"start\":\"2017-04-25T19:34:39+00:00\",\"stop\":null,\"duration\":-1493148879,\"description\":\"Some short description\",\"tag_ids\":[313040,3129041,319042],\"at\":\"2017-04-25T20:12:27+00:00\",\"server_deleted_at\":null,\"user_id\":0,\"created_with\":\"SomeApp\"}";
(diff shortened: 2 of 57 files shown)


Toggl.Core/Models/Preferences.cs
```diff
@@ -19,6 +20,9 @@ internal class Preferences : IThreadSafePreferences
         public const long fakeId = 0;
         public long Id => fakeId;
 
+        public static readonly string fakeUniqueId = Guid.Empty.ToString();
+        public string UniqueId => fakeUniqueId;
```
heytherewill, line 24: Why does the User model, which is also an singleton entity, need a real UniqueId and this model doesn't?

Toggl.Networking.Tests/Models/TimeEntryTests.cs
```diff
@@ -11,10 +11,10 @@ public sealed class TimeEntryTests
         public sealed class TheTimeEntryModel
         {
             private string validJson
-                => "{\"id\":525144694,\"workspace_id\":1414373,\"project_id\":3178352,\"task_id\":null,\"billable\":false,\"start\":\"2017-04-25T19:34:39+00:00\",\"stop\":null,\"duration\":-1493148879,\"description\":\"Some short description\",\"tag_ids\":[313040,3129041,319042],\"at\":\"2017-04-25T20:12:27+00:00\",\"server_deleted_at\":null,\"user_id\":0,\"created_with\":\"SomeApp\"}";
+                => "{\"id\":525144694,\"unique_id\":null,\"workspace_id\":1414373,\"project_id\":3178352,\"task_id\":null,\"billable\":false,\"start\":\"2017-04-25T19:34:39+00:00\",\"stop\":null,\"duration\":-1493148879,\"description\":\"Some short description\",\"tag_ids\":[313040,3129041,319042],\"at\":\"2017-04-25T20:12:27+00:00\",\"server_deleted_at\":null,\"user_id\":0,\"created_with\":\"SomeApp\"}";
```
heytherewill, line 14: Will the API actually return `unique_id:null`? 🤔

Toggl.Storage.Realm/Models/IModifiableId.cs
```diff
@@ -8,6 +8,8 @@ public interface IModifiableId : IIdentifiable
 
         long? OriginalId { get; set; }
 
+        new string UniqueId { get; set; }
```
heytherewill, line 11: Is this even needed?

Toggl.Core/Models/Preferences.cs
kornelijepetak: You are right. Somehow I missed that. `User` doesn't need a real one either.

Toggl.Networking.Tests/Models/TimeEntryTests.cs
kornelijepetak: No, the old API will not return anything. But this test is wrong. :)
We should never expect exact json string anyway. The test should've parsed the json and checked whether all the required properties with their respective values are in there.

The new API will not use this models in the same way anyway, so it will be covered by other stuff.

This test tho, has to assume that the server returns this as well.
Got any better solutions for this?

Toggl.Storage.Realm/Models/IModifiableId.cs
kornelijepetak: It is. In order to be able to perform DB migrations. If it's just a getter, realm migrations have no way of setting those to a meaningful value.

Toggl.Core/Models/Preferences.cs
kornelijepetak: Fixed it.

kornelijepetak:
> Mostly questions. Also, why use a String and not the GUID type?

@heytherewill the problem is that the Realm still does not support GUID types, or at least that's what I read.

kornelijepetak:
Considering some recent developings in Sync team, it is possible that this PR will not be needed after all.

Putting it on-hold for now.

kornelijepetak:
The Sync Team decided not to go this route.
Closing.

PR #7315, opened 2020-04-16
💾 Implement unsynced data file persistence service

### LINKED ISSUE #7306: Serialise and store the unsynced/unsyncable entities in a file
A followup of #7104

## Gathering all the entities
This has already been taken care of in #7262, so you should just use that interactor.

As agreed in #7104 this should happen every time the app enters background.

## Format

Everything should be serialised into a single file with the following layout:
```
{ "time_entries" : [], "projects" : [], "clients" : [], "tags" : []}
```
Each of these JSON arrays will contain the unsynced/unsyncable entities of the appropriate nature, exactly the way they would be serialised when sent to our API.

The file should be named `unsynced-migration-dump.json` and it should be stored in `NSDocumentDirectory `

The file is overwritten every time.

### LINKED ISSUE #7307: Serialise and store the unsynced/unsyncable entities in a file
A followup of #7104

## Gathering all the entities
This has already been taken care of in #7262, so you should just use that interactor.

As agreed in #7104 this should happen every time the app enters background.

## Format

Everything should be serialised into a single file with the following layout:
```
{ "time_entries" : [], "projects" : [], "clients" : [], "tags" : []}
```
Each of these JSON arrays will contain the unsynced/unsyncable entities of the appropriate nature, exactly the way they would be serialised when sent to our API.

The file should be named `unsynced-migration-dump.json`, but the location of it wasn't determined fully. @juanlaube suggested `DIRECTORY_DOCUMENTS `, but @pdrj proposed to go with `getFilesDir()`.

The file is overwritten every time.

semanticer:
## What's this?
This PR is making sure we persist unsynced data into a file every time the app goes into the background.

### Relationships

Closes #7307
Closes #7306

## Why do we want this?
We want to persist our unsynced data for future migration to Aurora.

## How is it done?
I created a new service interface `IUnsyncedDataPersistenceService` and an Android implementation of this service. The only method `persistUnsyncedData()` of this service is called from `TogglApplication` in the same place we call our platform-independent `backgroundService?.EnterBackground();`
Service is simply serializing `unsyncedDataDump` returned from the `CreateUnsyncedDataDumpInteractor` (added in #7262 ). I'm transforming everything to snake case since it seems to be required by #7307 .

### Why not another way?
I can't _see sharp_ so, please suggest any improvements.

### Side effects
None I hope

## Review hints
Take a look at my usage of **Task.Run** because I have no idea what I'm doing. 

## :squid: Permissions
Me


Toggl.Core/Services/IUnsyncedDataPersistenceService.cs
```diff
@@ -0,0 +1,8 @@
+namespace Toggl.Core.Services
+{
+    public interface IUnsyncedDataPersistenceService
+    {
+        public const string unsyncedDataFileName = "unsynced-migration-dump.json";
+        void persistUnsyncedData();
```
arturs105, line 6:
```suggestion
        void PersistUnsyncedData();
```

Toggl.Droid/Services/UnsyncedDataPersistenceServiceAndroid.cs
```diff
@@ -0,0 +1,38 @@
...
+using Newtonsoft.Json.Serialization;
+using Toggl.Core.Services;
+
+namespace Toggl.Droid.Services
+{
+    public class UnsyncedDataPersistenceServiceAndroid : IUnsyncedDataPersistenceService
+    {
+        public void persistUnsyncedData()
+        {
+            Task.Run(() =>
```
arturs105, line 15: Why dis? Why not make the whole method `async` and `await` the CreateUnsyncedDataDumpInteractor?

Toggl.Droid/Services/UnsyncedDataPersistenceServiceAndroid.cs
```diff
@@ -0,0 +1,38 @@
...
+                {
+                    NamingStrategy = new SnakeCaseNamingStrategy()
+                };
+                var serializedUnsyncedDataDump = JsonConvert.SerializeObject(unsyncedDataDump, new JsonSerializerSettings
+                {
+                    ContractResolver = contractResolver,
+                    Formatting = Formatting.Indented
+                });
+                var outputStream = Application.Context.OpenFileOutput(IUnsyncedDataPersistenceService.unsyncedDataFileName, FileCreationMode.Private);
+                byte[] bytes = Encoding.ASCII.GetBytes(serializedUnsyncedDataDump);
```
arturs105, line 30: What if some entities have unicode characters in them? Will `Encoding.ASCII` handle that properly?

semanticer: Yeah, I completely missed that. Looks like we are using UTF8 everywhere else so I switched to that.

Toggl.Droid/Services/UnsyncedDataPersistenceServiceAndroid.cs
```diff
@@ -0,0 +1,34 @@
...
+using Android.Content;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Serialization;
+using Toggl.Core.Services;
+
+namespace Toggl.Droid.Services
+{
+    public class UnsyncedDataPersistenceServiceAndroid : IUnsyncedDataPersistenceService
+    {
+        public async void PersistUnsyncedData()
```
arturs105, line 13: Doesn't this create some kind of warning that `async` methods should return `Task`?

semanticer: @arturs105 Looks like async void is valid in some cases I'm just not sure if this case is a callback case or Fire and Forget case where you should still return Task according to this article https://www.pluralsight.com/guides/returning-void-from-c-async-method

arturs105: I think event handlers are the _only_ acceptable case for `async void` and this does not qualify for that. MS even encourages avoiding `async void`: https://docs.microsoft.com/en-us/archive/msdn-magazine/2013/march/async-await-best-practices-in-asynchronous-programming

Toggl.Core/Services/IUnsyncedDataPersistenceService.cs
```diff
@@ -0,0 +1,10 @@
+using System.Threading.Tasks;
+
+namespace Toggl.Core.Services
+{
+    public interface IUnsyncedDataPersistenceService
+    {
+        public const string unsyncedDataFileName = "unsynced-migration-dump.json";
```
arturs105, line 7: 🤔 This can probably be made `protected`

Toggl.Droid/Services/UnsyncedDataPersistenceServiceAndroid.cs
```diff
@@ -0,0 +1,34 @@
...
+using Toggl.Core.Services;
+
+namespace Toggl.Droid.Services
+{
+    public class UnsyncedDataPersistenceServiceAndroid : IUnsyncedDataPersistenceService
+    {
+        public async Task PersistUnsyncedData()
+        {
+            var unsyncedDataDump = await AndroidDependencyContainer.Instance
+                .InteractorFactory.CreateUnsyncedDataDump().Execute();
```
arturs105, line 16: I think injecting the interactor factory would be more politically correct.

Toggl.Droid/Services/UnsyncedDataPersistenceServiceAndroid.cs
```diff
@@ -0,0 +1,34 @@
...
+using System.Threading.Tasks;
+using Android.App;
+using Android.Content;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Serialization;
+using Toggl.Core.Services;
+
+namespace Toggl.Droid.Services
+{
+    public class UnsyncedDataPersistenceServiceAndroid : IUnsyncedDataPersistenceService
```
arturs105, line 11: I think we should test whether the object is properly serialised before writing. Not sure how to do so though. 🤔 `FIleWriterService` would probably be overkill, so we might just inject `Func<string> writeToFile` and test whether the `func` gets the expected string.

Pinging @heytherewill for guidance.

Toggl.Core/Services/IUnsyncedDataPersistenceService.cs
```diff
@@ -4,7 +4,7 @@ namespace Toggl.Core.Services
 {
     public interface IUnsyncedDataPersistenceService
     {
-        public const string unsyncedDataFileName = "unsynced-migration-dump.json";
+        protected const string unsyncedDataFileName = "unsynced-migration-dump.json";
```
arturs105, line 7: `protected` members use `PascalCase`

Toggl.Droid/Services/UnsyncedDataPersistenceServiceAndroid.cs
semanticer: That could work yes, let's wait for @heytherewill to weigh in on this

heytherewill: I agree with Arturs, a function should be enough to test behavior and not insane enough that we need more services

semanticer:
@arturs105 Can you confirm if this new cross-platform implementation does the right thing on iOS?

Toggl.Droid/Services/UnsyncedDataPersistenceServiceAndroid.cs
semanticer: I ended up doing `Func<string, StreamWriter, Task> writeToFile;`. If you want to write the test feel free, otherwise I will do it tomorrow.

Toggl.Core/Services/UnsyncedDataPersistenceService.cs
```diff
@@ -0,0 +1,42 @@
...
+            {
+                NamingStrategy = new SnakeCaseNamingStrategy()
+            };
+            var serializedUnsyncedDataDump = JsonConvert.SerializeObject(unsyncedDataDump, new JsonSerializerSettings
+            {
+                ContractResolver = contractResolver,
+                Formatting = Formatting.Indented
+            });
+            var backingFile = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.Personal), IUnsyncedDataPersistenceService.UnsyncedDataFileName);
+            using var writer = File.CreateText(backingFile);
```
arturs105, line 38: 🤔 I'm not sure if this is correct `using` statement. They usually go like this:
```
using (var something = CreateTheThing())
{
    //Can use the something in this scope
}
```

However i haven't used `using` in ages, so maybe things have changed in a recent C# version.

Toggl.Core.Tests/Services/UnsyncedDataPersistenceServiceTests.cs
```diff
@@ -0,0 +1,241 @@
...
+                CreateUnsyncedDataDumpInteractor = Substitute.For<IInteractor<Task<UnsyncedDataDump>>>();
+                WriteToFile = Substitute.For<Func<string, StreamWriter, Task>>();
+            }
+        }
+
+        public sealed class TheConstructor
+        {
+            [Theory, LogIfTooSlow]
+            [ConstructorData]
+            public void ThrowsWhenAnyOfTheArgumensAreNull(bool useCreateUnsyncedDataDumpInteractor, bool useWriteToFile)
```
arturs105, line 37: Typo
```suggestion
            public void ThrowsWhenAnyOfTheArgumentsAreNull(bool useCreateUnsyncedDataDumpInteractor, bool useWriteToFile)
```

PR #7311, opened 2020-04-14
🤜 Add Sync Push actions classes

kornelijepetak:
## What's this?
Adds the Push Actions classes for Sync.

### Relationships
Closes #7280 

## Why do we want this?
To utilize the new sync API.

## How is it done?
Just some classes. I decided not to make things too generic as there's no particular appeal in doing that, since that would require introducing more interfaces which I'm not really eager to do.

## Review hints
Requires first look.

## Tests
Do we want/need tests for this?

## :squid: Permissions
Only Sync Team.

PR #7316, opened 2020-04-16
Move to push ns

simonrozsival:
## What's this?
This is just a bunch of renaming.

### Relationships
Follow-up of #7311 

## Why do we want this?
There's a lot of classes and this way they will be sorted in a cleaner way.

## How is it done?
Just moving files and changing `namespace` and `using`.

## :squid: Permissions
🦑


simonrozsival:
I actually realized I wanted to one more thing in this PR and I forgot 😱 Let me push one more commit.

sofokli1:
> I actually realized I wanted to one more thing in this PR and I forgot 😱 Let me push one more commit.

just 60 more file changes 😛

simonrozsival:
@sofokli1 Just 5 more!

Now it's ready 😆

Toggl.Networking/Sync/Push/Action.cs
```diff
@@ -1,9 +1,9 @@
 ﻿using Newtonsoft.Json;
 using System;
 
-namespace Toggl.Networking.Sync
+namespace Toggl.Networking.Sync.Push
 {
-    public class PushAction<TMeta> : IPushAction
+    public class Action<TMeta> : IAction
```
heytherewill, line 6: Is using a name that clashes with something inside the .NET framework a good idea?

Toggl.Networking/Sync/Push/IAction.cs
```diff
@@ -0,0 +1,6 @@
+﻿using System;
+
+namespace Toggl.Networking.Sync.Push
+{
+    public interface IAction { }
```
heytherewill, line 5: Why the marker interfaces?

simonrozsival: This interface was introduced in #7311

Toggl.Networking/Sync/Push/Action.cs
simonrozsival: It has the same name, but it's in a different namespace. That's one of the reasons we have namespaces 😄

heytherewill: It was called PushAction before

kornelijepetak: @simonrozsival Yeah, I'm not sure I like this either. Despite being in a `Push` namespace, you'll get into a file where you'll need same named stuff from both Push and Pull and then you'll have to use the fully qualifier anyway. I'd rather leave those as they were 🤷‍♂️

Toggl.Networking/Sync/Push/IAction.cs
kornelijepetak: Because it's primarily used for serialization and the actual classes don't have too much stuff in common to warrant non-marker interface.

kornelijepetak:
Decided to let that rename go.
If troubles with naming arise, I'll request a revert of some of these names.

My approval still stands.

PR #7324, opened 2020-04-17
🌟 Add fetch dirty entities interactor

### LINKED ISSUE #7281: Create an interactor for gathering all dirty entities
## 🎯 What needs to be done?
We need an interactor that will fetch all the dirty entities from the database.

## 🛠 Technical remarks
Well, in my opinion, instead of duplicating all the properties for 3-way merging, I'd rather have an additional copy of an entity in the database that represents a `base` entity (for 3-way merge).

When an entity is changed locally, a copy of it is made and from then on, the copy is edited and pushed to the server. During the pull/merge, both the `copy` (local) and the `base` (local) are used in 3-way merging with the `server` version. When all of the changes are merged, the `base` one is used as a merged version and the `copy` is deleted (until the next local change).

For this I'd propose the following extension on the data model:

* Additional boolean property `IsActiveEntity` (set by default to all entities) on an entity
* A new `enum EntityMode { Original, Copy }` type
* Additional `EntityMode` property `Mode` on an entity

Marking things as such, the local edits would simply copy the entity, mark the original with `Mode = Original, IsActiveEntity = false` and the copy with `Mode = Copy, IsActiveEntity = true`. The merging process would use all of the entities, while the rest of the system would simply work only on entities with the `IsActiveEntity == true`. 

That would make the interactor of this issue very trivial. The `dirtyness` would be implied by the `Mode`. So, just fetch all the copies.

### LINKED ISSUE #7282: Implement conversion from dirty entities to push actions
## 🎯 What needs to be done?
When dirty entities are to be pushed, they need to be converted into [actions](https://github.com/toggl/mobileapp/issues/7280).

It can probably be an interactor, but maybe it doesn't even have to be. It can be much simplified static converter.

## 🔗 Dependencies
#7280 needs to be solved before.

## 🏆 Outcome
The conversion from dirty entities to push actions is available.

kornelijepetak:
## What's this?
Adds an interactor that fetches all the dirty entries and turns them into a Push Request object.

I've also changed the `Request` API from _fluent_ to simple `void` methods.

### Relationships
Closes #7281 
Closes #7282 

## Why do we want this?
So that sync can simply get all things that need to be pushed.

## How is it done?
Well, by fetching stuff from database.

## Tests
Yet to be added. For now I'm setting this as wip beacuse of that.

## :squid: Permissions
Me or @simonrozsival


Toggl.Core/Interactors/Sync/GetDirtyEntitiesInteractors.cs
```diff
@@ -0,0 +1,84 @@
...
+
+        private IEnumerable<T> entitiesToCreate<T>(IEnumerable<T> entities) where T : IIdentifiable
+            => entities.Where(entity => entity.Id < 0);
+
+        private bool isDirty(IDatabaseSyncable entity)
+            => entity.SyncStatus == SyncNeeded;
+
+        private ActionType actionType<T>(T entity) where T : IIdentifiable, IDatabaseSyncable
+            => entity.IsDeleted
+            ? Delete
```
simonrozsival, line 81: There's one extra option: If `entity.IsDeleted == true && entity.Id < 0`, then you should just delete the entity locally and you shouldn't push it to the server. We could alternatively change the TE repository to hard-delete local-only TEs instead of just soft-deleting them.

Toggl.Networking/Extensions/TypeExtensions.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace Toggl.Networking.Extensions
+{
+    public static class TypeExtensions
+    {
+        public static IEnumerable<T> WhereMatchesGenericType<T>(this IEnumerable<T> collection, Type typeToMatch)
```
simonrozsival, line 10: Is this used just in the tests? It should be in the tests project.

Toggl.Core/DataSources/TimeEntriesDataSource.cs
```diff
@@ -89,6 +100,25 @@ public void OnTimeEntryStopped(IThreadSafeTimeEntry timeEntry)
             timeEntryStoppedSubject.OnNext(timeEntry);
         }
 
+        public IObservable<IEnumerable<IThreadSafeTimeEntry>> GetBackedUpTimeEntries()
+            => timeEntriesBackupRepository.GetAll().Select(entities => entities.Select(Convert));
+
+        public override IObservable<IThreadSafeTimeEntry> Update(IThreadSafeTimeEntry entity)
+            => update(entity).ToObservable(schedulerProvider.DefaultScheduler);
+
+        private async Task<IThreadSafeTimeEntry> update(IThreadSafeTimeEntry entity)
```
simonrozsival, line 109: Should we also call this for `BatchUpdate`?

Toggl.Core.Tests/Interactors/Sync/PreparePushRequestInteractorTests.cs
```diff
@@ -0,0 +1,216 @@
...
+            {
+                var request = await interactor.Execute();
+
+                request.IsEmpty.Should().BeTrue();
+            }
+
+            [Fact, LogIfTooSlow]
+            public async Task EmitsEmptyRequestWhenThereAreNoDirtyEntities()
+            {
+                returnsValueOrEmpty(DataSource.Tags, new[] { new MockTag(), new MockTag() });
```
simonrozsival, line 66: Please add "clean" entities to all of the data sources, not just the tags.

Toggl.Core.Tests/Interactors/Sync/PreparePushRequestInteractorTests.cs
```diff
@@ -0,0 +1,216 @@
...
+
+                request.IsEmpty.Should().BeFalse();
+                request.Clients.Count.Should().Be(createdClientsCount);
+            }
+
+            [Fact, LogIfTooSlow]
+            public async Task RequestContainsProjectsCreations()
+            {
+                var projects = new[] {
+                    new MockProject() { Id = -1 }.WithColor("#000000"),
```
simonrozsival, line 93: Couldn't you set the color directly in the initializer? 😄

Toggl.Core.Tests/Interactors/Sync/PreparePushRequestInteractorTests.cs
```diff
@@ -0,0 +1,216 @@
...
+                request.IsEmpty.Should().BeFalse();
+                request.TimeEntries.WhereMatchesGenericType(typeof(CreateAction<>)).Should().HaveCount(createdTimeEntriesCount);
+            }
+
+            [Fact, LogIfTooSlow]
+            public async Task RequestContainsTimeEntryUpdates()
+            {
+                var backupUpdates = new IThreadSafeTimeEntry[]
+                {
+                    new MockTimeEntry() { Id = 1, Description = "ABC-changed", SyncStatus = SyncNeeded },
```
simonrozsival, line 173: Shouldn't this be the other way round? This should be `ABC-original` and the other entity should have `ABC-changed`? I thought the _backup_ is the original entity from the server before the app changed it locally.

Toggl.Core.Tests/Interactors/Sync/PreparePushRequestInteractorTests.cs
```diff
@@ -0,0 +1,216 @@
...
+            }
+
+            private MockTimeEntry[] createTimeEntries()
+                => new[] {
+                    new MockTimeEntry() { Id = -100, SyncStatus = SyncNeeded },
+                    new MockTimeEntry() { Id = -200, SyncStatus = SyncNeeded },
+                    new MockTimeEntry() { Id = -300, SyncStatus = SyncNeeded },
+                    new MockTimeEntry() { Id = 1, Description = "ABC", SyncStatus = SyncNeeded },
+                    new MockTimeEntry() { Id = 2, Description = "DEF", SyncStatus = SyncNeeded },
+                    new MockTimeEntry() { Id = 99, SyncStatus = SyncNeeded, IsDeleted = true }
```
simonrozsival, line 152: I'm missing a test for a local-only deleted entity (negative ID and `IsDeleted == true`).

Toggl.Core.Tests/Interactors/Sync/PreparePushRequestInteractorTests.cs
```diff
@@ -0,0 +1,216 @@
...
+            public async Task RequestContainsTimeEntryCreations()
+            {
+                var timeEntries = createTimeEntries();
+                var createdTimeEntriesCount = timeEntries.Count(t => t.Id < 0);
+                returnsValueOrEmpty(DataSource.TimeEntries, timeEntries);
+
+                var request = await interactor.Execute();
+
+                request.IsEmpty.Should().BeFalse();
+                request.TimeEntries.WhereMatchesGenericType(typeof(CreateAction<>)).Should().HaveCount(createdTimeEntriesCount);
```
simonrozsival, line 165: Couldn't you just use `OfType<CreateAction<ITimeEntry>>()` to filter them (maybe it's not `ITimeEntry` in this case)?

Toggl.Networking/Sync/Push/Request.cs
```diff
@@ -17,7 +17,84 @@ public sealed partial class Request
...
+                .Select(timeEntry => new CreateAction<TimeEntry>(timeEntry))
+                .AddTo(TimeEntries);
+        }
+
+        public void UpdateTimeEntries(IEnumerable<ITimeEntry> timeEntries, IEnumerable<ITimeEntry> backedUpTimeEntries)
+        {
+            var timeEntriesMap = timeEntries.ToDictionary(te => te.Id); 
+
+            backedUpTimeEntries
+                .Select(bte => new UpdateAction<ITimeEntry>(timeEntriesMap[bte.Id], bte))
```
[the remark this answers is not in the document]
kornelijepetak: Well, this is called only for update time entries for which the following is true:
* `SyncStatus` == `SyncNeeded`
* TE is not a local delete.

So, in case the backup is not there, nothing will happen here. Nothing will throw.

However, as you mentioned, this would be a mess right after the migration as there would be some `SyncNeeded` updates, which wouldn't have a corresponding entry in the backup table. 

One solution is to create a migration that will transfer all non-synced elements into a completely meaningless entries with the same ID in the backup table, so that when the push gets to run, all props will differ and the item will be pushed.

The other is for me to add some code here to force the update of the whole item. Which seems the cleaner solution.

Toggl.Core.Tests/Interactors/Sync/PreparePushRequestInteractorTests.cs
kornelijepetak: DERP. Of course 🤦‍♂️

Toggl.Networking/Sync/Push/Serialization/PartiallySerializableTimeEntry.cs
```diff
@@ -0,0 +1,56 @@
...
+
+            if (!entityTags.SetEquals(backupEntityTags))
+            {
+                var array = new JArray();
+                entityTags.ForEach(tagId => array.Add(tagId));
+
+                obj.Add(new JProperty("tag_ids", array));
+            }
+
+            return obj;
```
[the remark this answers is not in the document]
kornelijepetak: Well, the issue with the suggestion you are proposing is that it treasts `null` as  a lack of `change`. But a change of a referenced project from `166324` to a `null` means the change from TE having a project to one that doesn't have a project, and this change would be missing this way.

If all properties were non-nullable, then it would maybe be a much more attractive option.

But that's just for TEs, so I'd rather not dabble in this too much longer. Let the aurora team think of an appropriate way to serialize these things in their respective platforms.

Toggl.Storage.Realm/Database.cs
```diff
@@ -27,6 +27,7 @@ public Database(RealmConfigurator realmConfigurator)
             Preferences = SingleObjectStorage<IDatabasePreferences>.For(getRealmInstance, (preferences, realm) => new RealmPreferences(preferences, realm));
             Projects = Repository<IDatabaseProject>.For(getRealmInstance, (project, realm) => new RealmProject(project, realm));
             TimeEntries = Repository<IDatabaseTimeEntry>.For(getRealmInstance, (timeEntry, realm) => new RealmTimeEntry(timeEntry, realm));
+            TimeEntriesBackup = Repository<IDatabaseTimeEntry>.For(getRealmInstance, (timeEntry, realm) => new RealmTimeEntry(timeEntry, realm));
```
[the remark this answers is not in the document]
kornelijepetak: Oh, no, this is definitely a mistake on my part.

Toggl.Networking/Extensions/TypeExtensions.cs
kornelijepetak: Will move.

Toggl.Core/DataSources/TimeEntriesDataSource.cs
kornelijepetak: Yes, I'll fix this as well.

Toggl.Core.Tests/Interactors/Sync/PreparePushRequestInteractorTests.cs
simonrozsival: I already forgot everything about how C# works ... 😆

Toggl.Networking/Sync/Push/Request.cs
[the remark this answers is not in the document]
kornelijepetak: This comment no longer applies.

Toggl.Core.Tests/Interactors/Sync/PreparePushRequestInteractorTests.cs
kornelijepetak: Indeed. Ignore this, it's changed anyway.

Toggl.Core/DataSources/TimeEntriesDataSource.cs
```diff
@@ -89,6 +96,34 @@ public void OnTimeEntryStopped(IThreadSafeTimeEntry timeEntry)
...
+        public override IObservable<IEnumerable<IConflictResolutionResult<IThreadSafeTimeEntry>>> BatchUpdate(IEnumerable<IThreadSafeTimeEntry> timeEntries)
+            => timeEntriesRepository.GetByIds(timeEntries.Select(te => te.Id).ToArray())
+            .Select(timeEntriesDb => timeEntriesDb.ToDictionary(te => te.Id))
+            .Select(dbMap => timeEntries.Select(te => new { TimeEntry = te, TimeEntryDb = dbMap[te.Id] }))
+            .Do(timeEntryPairs => timeEntryPairs.ForEach(pair => backupTimeEntry(pair.TimeEntryDb, pair.TimeEntry)))
+            .SelectMany(_ => base.BatchUpdate(timeEntries));
+
+        private void backupTimeEntry(IDatabaseTimeEntry timeEntryDb, IThreadSafeTimeEntry timeEntry)
+        {
+            if (!timeEntryDb.ContainsBackup)
```
simonrozsival, line 113: I don't understand this 🤔 are we gonna backup only the first edit? Or is this in place to avoid overwriting local changes with the data pulled from the server in `BatchUpdate`?

kornelijepetak: I'm confused about your comment.
Of course the original is backed up only in case of the first edit. After the first edit, there is no more original unless it was backed up.

So, 
* user has synced everything.
* Then the user changes entity A
_This A is the `base` in the 3-way merge_.
* The A is backed up and the edit is allowed
* Every other edit that happened since the last push is only local, so there's no need to backup anything else.
* At the moment of merge, there will be the latest changed entity (`local`), the `server` version (from pull) and the `backed-up` version.

When the 3-way merge completes, the backup is erased and the entity contains the result of the merge.

Does this clarify the problem, or am I missing something?

simonrozsival: > Of course the original is backed up only in case of the first edit. After the first edit, there is no more original unless it was backed up.

🤦‍♂️ obviously. For some reason I thought about it the other way round, which doesn't make any sense. Thanks for the clarification!

Issue #7309, opened 2020-04-14
Figure out testing for the unsyncable entities dump


arturs105:
Followup of #7104

This still wasn't 100% figured out in the issue above, so let's do it here.

Here's what was suggested in said issue:

>We could add something that triggers sharing the json file with other apps so we can inspect the contents, maybe triggered by a not-so-obvious gesture like tapping 7 times on the version number or a keyboard shortcut?

>Same way we test loss of workspace access UI. The popup in setting for testing these cases seems like a good place to manually trigger it.

>We can create an unsyncable TE (extra long description for instance) and then inspect the file's contents. That inspection tool can be accessed via settings, as suggested above.

semanticer:
I would be consistent with our previous hidden testing method and access this via Settings -> Help -> Long Press. But I'm not 100% sure how would that "inspection tool" look like.

arturs105:
Yeah, i agree that having something similar to previous hidden testing method makes sense. 🤔 Maybe we could even add another button to the same screen.

But as to what the inspection tool should look like, i think we could simply use the system's sharing functionality and then Airdrop, e-mail or do whatever with the file to inspect it's contents.

PR #7326, opened 2020-04-24
📨Debug unsynced data via mail

semanticer:
## What's this?
This PR adds one more option to our hidden debug menu that you can access via long press on **About**. This option triggers Intent that can be handled for example by Gmail or Google Drive app. It will attach a file with the currently usynced data dump so you can send it to yourself.

### Relationships

This should address the Android part of #7309

## Why do we want this?
We want to make unsynced data testing easier even for people with no knowledge of Android internals.

## How is it done?
The internal dump file is copied to a temporary file which is then sent via email or uploaded to google drive. I had to use this temporary file that is saved to external storage because Gmail's security policy prohibits attachments of files from internal storage.

### Why not another way?
We can probably make it a bit nicer, this was sort of hacked just to prove the concept. So let me know if you have any ideas. However, I don't think we should spend too much time on debug functionality.

### Side effects
none that am aware of

## Review hints
Just try to send yourself an email.

## :squid: Permissions
Anyone


Toggl.iOS.SiriExtension.UI/IntentViewController.cs
```diff
@@ -5,7 +5,6 @@
 using IntentsUI;
 using System;
 using System.Globalization;
-using Toggl.iOS.Intents;
```
heytherewill: Revert dis

dshr:
I might be a bit out of the loop — AFAIK the hidden debug menu is not available in production, is this for internal purposes only, or will this migrate to the main settings page in time?

Toggl.Droid/DebugHelpers/ErrorTriggeringFragment.cs
```diff
@@ -47,6 +52,7 @@ public override View OnCreateView(LayoutInflater inflater, ViewGroup container,
             var outdatedApi = createTextView("Outdated API error");
             var outdatedAppPermanently = createTextView("Permanent outdated client error");
             var outdatedApiPermanently = createTextView("Permanent outdated API error");
+            var unsyncedDataDump = createTextView("Unsynced data dump test");
```
arturs105, line 55: 🤔 Maybe "Share unsynced data dump" would be more accurate.

Toggl.Droid/DebugHelpers/ErrorTriggeringFragment.cs
```diff
@@ -98,18 +109,15 @@ private TextView createTextView(string text)
...
-            layoutParameters.BottomMargin = margin;
-
             var textView = new TextView(Context)
             {
                 Text = text,
                 LayoutParameters = layoutParameters
             };
 
+            var padding = 20.DpToPixels(Context);
+            textView.SetPadding(padding, padding, padding,padding);
```
arturs105, line 119:
```suggestion
            textView.SetPadding(padding, padding, padding, padding);
```

Toggl.Droid/DebugHelpers/ErrorTriggeringFragment.cs
```diff
@@ -176,6 +184,29 @@ private void outdatedApiPermanentlyErrorTriggered()
...
+        {
+            void copyFile(File src, File dst) {
+                FileInputStream inStream = new FileInputStream(src);
+                FileOutputStream outStream = new FileOutputStream(dst);
+                inStream.Channel.TransferTo(0, inStream.Channel.Size(), outStream.Channel);
+                inStream.Close();
+                outStream.Close();
+            }
+
+            var backingFile = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.Personal), IUnsyncedDataPersistenceService.UnsyncedDataFileName);
```
arturs105, line 197: 🤔 Maybe we can move this whole `Path.Combine` to `UnsyncedDataPersistenceService.UnyncedDataFilePath`, so callers of `UnsyncedDataPersistenceService.UnsyncedDataFileName` don't have to remember that the file is in `Environment.SpecialFolder.Personal`.

semanticer:
Good, idea. Even I did this mistake multiple times during debugging. I think it's totally fine to generate this file right before launching the sharing option. It's fast enough and straight forward.

PR #7328, opened 2020-04-27
Reload reports when the tab is opened if time entries have changed

### LINKED ISSUE #7170: Reports tab doesn't update unless restarted or date range changed
## ⛑️ User Support

User experiences issue where they add a time entry and switch to the reports tab. The time entries on the tab won't update unless the app is force closed and reopened OR the date range picker at the top left-hand corner has been changed to a different date and then changed back to the original date.

## About the user

App version: Giskard/2.11.1
Phone OS: Android 10
Phone model: Pixel 2

## Expected behaviour

User should be able to go to the report page and have it update straight away.

## To Reproduce

1. Add a time entry
2. Go to report tab - report will not include your latest entry
3. Change date range
4. Change date range back - report will now be updated

## Suggestions

[Case reference](https://app.intercom.com/a/apps/ayixs927/inbox/inbox/3540765/conversations/26053517491)

arturs105:
## What's this?
This PR makes the reports reload upon opening the reports tab if the time entries have changed while the reports tab was in the background.

### Relationships

Closes #7170

## How is it done?
By updating a flag on `dataSource.TimeEntries.ItemsChanged` and then checking that flag on `ViewAppeared`.

## :squid: Permissions
🦑


Toggl.Core.UI/ViewModels/Reports/ReportsViewModel.cs
```diff
@@ -27,10 +30,15 @@ namespace Toggl.Core.UI.ViewModels.Reports
 {
     public sealed class ReportsViewModel : ViewModel
     {
-        public const string DownArrowCharacter = "▾";
+        private const string DownArrowCharacter = "▾";
```
dshr, line 33:
```suggestion
        private const string downArrowCharacter = "▾";
```

PR #7340, opened 2020-05-04
🔒 Add IsStrong property to Password

### LINKED ISSUE #7329: Add password strength rules validation to passwords
## 🎯 What needs to be done?
Add the following checks to `Password`:
- at least 8 characters
- at least one uppercase letter
- at least one lowercase letter
- at least one number.

Expose this check as a property on `Password`.

## 🛠 Technical remarks
Make sure the lower/uppercase check works for different alphabets too.

## 🏆 Outcome
`Password` has a `IsStrong` [feel free to come up with a better name] property that returns `true` if all the rules above are matched, `false` otherwise.

There are tests to validate the property is computed properly.

## 👕 Estimated complexity
S

pdrj:
### What's this?
This adds `IsStrong` property to `Password`.

### Relationships

Closes #7329 

## Why do we want this?
To know if the password is strong 💪
 
## How is it done?
Just added combined some checks from `char` class.

### Why not another way?
🤷 

### Side effects
None

## Review hints
There's not much to test - you can run unit tests and read through the code.

## :squid: Permissions
Anyone


Toggl.Shared.Tests/PasswordTests.cs
```diff
@@ -39,6 +39,74 @@ public void ReturnsFalseForEmptyPassword()
...
+            [InlineData("")]
+            [InlineData("1")]
+            [InlineData("pAs1s")]
+            [InlineData("1aB4567")]
+            [InlineData("    aff")]
+            [InlineData("\t axxX")]
+            public void ReturnsFalseWhenPasswordIsShorterThan8Characters(
+                string passwordString)
+            {
+                Password.From(passwordString).IsStrong.Should().BeFalse();
```
heytherewill, line 61: This could be a property test

Toggl.Shared.Tests/PasswordTests.cs
```diff
@@ -39,6 +39,74 @@ public void ReturnsFalseForEmptyPassword()
...
+
+            [Theory, LogIfTooSlow]
+            [InlineData("12345678999aa")]
+            [InlineData("fdasfsf333qwerty1")]
+            [InlineData("من أي وقت مضى الكلب القط من أي وقت مضىx11")]
+            [InlineData("####@!@!@!@!@!@!123aaa")]
+            public void ReturnsFalseWhenNoUppercasePresent(
+                string passwordString)
+            {
+                Password.From(passwordString).IsStrong.Should().BeFalse();
```
heytherewill, line 72: This could also be a property test where you get the random string and call `.ToLower()`

pdrj:
> In general, I'd also like to see tests where the string matches more than one but not all constraints.

All the `ReturnsFalseWhenNo[Something]Present` test that @heytherewill. There's always a test case that matches just 2 or 3 out of 4 conditions.

pdrj:
> Also, can a password be weak but valid? Are we allowing new sign ups to have weak password or only logins?

We allow weak passes for logins only. We check the `IsValid` in `LoginEnabled` observable in the login view model, hence I introduced another property.

PR #7342, opened 2020-05-05
Deserialize pull response

### LINKED ISSUE #7341: Deserialize the `/pull` endpoint response
The docs are available in the `sync-server` project repo.

simonrozsival:
## What's this?
Just the deserialization logic, nothing more.

### Relationships
Closes #7341 

## :squid: Permissions
@kornelijepetak


Toggl.Networking/Sync/Pull/Response.cs
```diff
@@ -0,0 +1,56 @@
...
+using System.Collections.Generic;
+using Toggl.Shared;
+using Toggl.Shared.Models;
+using Toggl.Networking.Models;
+using Newtonsoft.Json;
+
+namespace Toggl.Networking.Sync.Pull
+{
+    [Preserve(AllMembers = true)]
+    internal class Response : IResponse
```
kornelijepetak, line 11: A much simpler solution is the following (the `int[]` is just an artificial example I've used to test this):

image

Json.NET will use the constructor and map the names of the constructor parameters to the properties in the json.

simonrozsival: I'll do that. Thanks for the suggestion!

Issue #7332, opened 2020-04-28
New Copy Translation Request - Password security criteria


pdrj:
**Target Languages:** 
PT
JP
FR

**New Copies:**
- Include 8 or more characters, with at least one uppercase letter, lowercase letter, and a number.
- One more try before account gets locked for 5 minutes.
- Account is locked for 5 minutes. Account owner has been notified.
- Incorrect email or password.

~~**EDIT - adding another string**
Set a secure password~~
It turns out 'Set a secure password' won't be needed after all

pdrj:
pt-br: Inclua 8 ou mais caracteres, com ao menos uma letra maiúscula, uma letra minúscula, e um número.

thanks to @heytherewill

pdrj:
Adding one more string:
Set a secure password

pdrj:
Already adding copy for the second phase:
- One more try before account gets locked for 5 minutes.
- Account is locked for 5 minutes. Account owner has been notified.
- Incorrect email or password.

heytherewill:
I'm by no means a french speaker, so take my french with many grains of salt and run the translations through a native speaker before adding them to the app :)

- Include 8 or more characters, with at least one uppercase letter, lowercase letter, and a number.
🇧🇷 Inclua 8 ou mais caracteres, com ao menos uma letra maiúscula, uma letra minúscula, e um número.
🇫🇷 Inclue 8 ou plus caractères, au moins une lettre majuscule, une lettre minuscule, et un numéro
- One more try before account gets locked for 5 minutes.
🇧🇷 Você tem apenas mais uma tentativa de login antes da sua conta ser bloqueada por 5 minutos.
🇫🇷 Vous pouvez essayer seulement un altre fois Puis, votre compt sera bloqué.
- Account is locked for 5 minutes. Account owner has been notified.
🇧🇷 Conta bloqueada por 5 minutos. O dono da conta foi notificado.
🇫🇷 Compte bloqué par 5 minutes. La propriétere de le compte a été notifié .
- Incorrect email or password.
🇧🇷 Email ou senha incorreta.
🇫🇷 Email ou mot de passe incorrect.

daividssilverio:
- Include 8 or more characters, with at least one uppercase letter, lowercase letter, and a number.
  🇯🇵 8文字以上でなければなりません。英字の大文字や小文字、数字、番号を一つずつー以上の文字を入れてください。
- One more try before account gets locked for 5 minutes.
[...]

PR #7346, opened 2020-05-05
🔒 Display password criteria in singnup screen

### LINKED ISSUE #7331: Include the password strength criteria in the Android UI
## 🎯 What needs to be done?
Inform the user of our password strength rules. If they try to signup with a password that doesn't match these rules, inform them even more [in red] 😃  

## 📝 Detailed explanation
Design links an UX: https://www.notion.so/toggl/UX-documentation-for-password-strenght-project-done-9779bca9938b46de8d1c6761b9f30bd3#89d2f5ae71a5444cbce72910a2e1579b

tldr:
- display the "Include 8 or more characters, with at least one uppercase letter, lowercase letter, and a number." below password field
- make sure the copy is always visible when the password field is active
- if user tries to signup with a password that doesn't match the criteria, make the copy red
- the copy turns white again, as soon as they type in a strong password (no need to actually press 'signup' again)

## 🔗 Dependencies
#7329 

## 🏆 Outcome
- The copy is displayed below the password field
- It is always visible when the password field is active
- If the user tries to signup with a password that doesn't match the rules, make the copy red
- The copy turns white again, as soon as they type in a strong password (no need to actually press 'signup' again)

## 👕 Estimated complexity
S

pdrj:
## What's this?
This displays password criteria in singnup screen. It also cleans up the IsValid property, which was used wrong in some places and turns out is not needed anymore.

### Relationships

Closes #7331 

## Why do we want this?
To let users know what's going on.

## How is it done?
To display the criteria, I'm reusing the 'error' label on textinput. It gives us some animations and styling will always be the same as of the error label on email field. I think it's better than adding a separate textview, but let me know what you think.

It turns out `IsValid` is not needed anymore, because the only places that still used it, used it incorrectly. For instance, in token reset screen, we wouldn't let users reset the token with a short password and that is wrong (some users may still be on short passwords). I've modified those usages to just check if password is non-empty and then removed the `IsValid` property.

## Review hints
- Try signing up with a weak password. The criteria should turn red. Then modify the password so that criteria are matched **the criteria should turn to regular text color as soon as they are matched, without actually pressing 'signup' again**. Delete some characters so that criteria are not matched again - try to signup and validate criteria turn red again. 
- Try to actually signup and validate it works.
 

## Localizations
#7332 

## :squid: Permissions
Anyone


Toggl.Droid/Activities/SignUpActivity.cs
```diff
@@ -115,5 +123,31 @@ string signupButtonTitle(bool isLoading, string currentLoadingMessage)
...
+        {
+            if (isStrong)
+            {
+                setPasswordErrorLabelTextColor("");
+            }
+        }
+
+        private void setPasswordErrorLabelTextColor(string errorText)
+        {
+            if (errorText == "")
```
heytherewill, line 137: This is really cryptic, why not just make this method receive and errorDrawable, a color and a string?

Toggl.Shared/Resources.resx
```diff
@@ -175,6 +175,10 @@ mobile, desktop or web</value>
         <value>Password should be at least 6 characters</value>
         <comment>Displayed when the user tries to login or sign up with an invalid password (less than 6 characters)</comment>
     </data>
+    <data name="StrongPasswordCriteria" xml:space="preserve">
+        <value>Include 8 or more characters, with at least one uppercase letter, lowercase letter, and a number.</value>
```
heytherewill, line 179: pt-br: Inclua 8 ou mais caracteres, com ao menos uma letra maiúscula, uma letra minúscula, e um número.

pdrj: Thanks, added to #7332

Toggl.Droid/Activities/SignUpActivity.cs
```diff
@@ -67,15 +69,15 @@ protected override void InitializeBindings()
...
                 .DisposedBy(DisposeBag);
 
             ViewModel.SignUpError
-                 .Subscribe(errorLabel.Rx().TextObserver())
-                 .DisposedBy(DisposeBag);
+                .Subscribe(errorLabel.Rx().TextObserver())
+                .DisposedBy(DisposeBag);
 
             var animatedLoadingMessage = TextHelpers.AnimatedLoadingMessage();
-
```
pdrj, line 70: I must have pressed reformat hotkey out of habit, sorry.

Toggl.Droid/Activities/SignUpActivity.cs
```diff
@@ -115,5 +125,32 @@ string signupButtonTitle(bool isLoading, string currentLoadingMessage)
...
+            {
+                setPasswordErrorLabelTextColor(error, errorTextColor, defaultErrorDrawable);
+            }
+        }
+
+        private void resetPasswordErrorLabelColorIfPasswordStrong(bool isStrong)
+        {
+            if (isStrong)
+            {
+                setPasswordErrorLabelTextColor("", secondaryTextColor, null);
```
heytherewill, line 145: Shouldn't this be `handlePasswordError("")`?

pdrj: Yes. Good catch - thanks.

PR #7350, opened 2020-05-06
Data dump analytics

### LINKED ISSUE #7308: Add analytics to the unsynced file dump.
Followup of #7104

Depends on #7307 and #7306 

Everytime the unsynced entries are dumped into a file, an analytics event should be tracked. The event should be called: `UnsyncedDataDumped` with the following properties:
- `TimeEntries` - number of time entries serialised
- `Projects` - number of time entries serialised
- `Clients` - number of time entries serialised
- `Tags` - number of time entries serialised

heytherewill:
## What's this?
This adds analytics events to our data dumps

### Relationships
Closes #7308 

## Why do we want this?
To better understand how the migration will work

### Why not another way?
O could've added the event to the interactor, but at that point we still haven't really committed to writing, so I think the service itself is a better place.

## :squid: Permissions
Anyone, anytime


Toggl.Core.Tests/Services/UnsyncedDataPersistenceServiceTests.cs
```diff
@@ -20,11 +22,15 @@ public sealed class UnsyncedDataPersistenceServiceTests
     {
         public abstract class UnsyncedDataPersistenceServiceTest
         {
-            protected IInteractor<Task<UnsyncedDataDump>> CreateUnsyncedDataDumpInteractor { get; }
+            protected IAnalyticsService AnalyticsService { get; }
             protected Func<string, StreamWriter, Task> WriteToFile { get; }
+            protected IInteractor<Task<UnsyncedDataDump>> CreateUnsyncedDataDumpInteractor { get; }
```
kornelijepetak, line 27: Oh, I still find it funny how you OCD all these lines according to their length 😄

PR #7359, opened 2020-05-07
👤 Add userAgent info to created time entries

### LINKED ISSUE #7353: Set `CreatedWith` for newly created time entries
When creating TEs, we must set the `CreateWith` property of `Toggl.Networking.Models.TimeEntry`.  Currently it's being done in `TimeEntriesApi.cs:66` just before the TE is POSTed to V9. We need to do something similar in `Toggl.Networking.Sync.Push.Request.CreateTimeEntries`.

kornelijepetak:
## What's this?
Adds a correct user agent to the created time entries.

### Relationships
Closes #7353

## Why do we want this?
So that the backend gets proper user agent data.

## How is it done?
Injection, ToString, pass by reference and other classical goodies.

### Why not another way?
No.

### Side effects
A fistbump 🤜🤛

## Review hints
This branches from #7324 (to avoid conflicts) and needs to be merged before #7324.

## :squid: Permissions
Sync-mobile boiz


Toggl.Networking/Sync/Push/Request.cs
```diff
@@ -20,6 +23,11 @@ public sealed partial class Request
         public SingletonUpdateAction<IPreferences> Preferences { get; set; }
         public SingletonUpdateAction<IUser> User { get; set; }
 
+        public Request(string userAgent = null)
+        {
+            this.userAgent = userAgent ?? "";
```
simonrozsival, line 28: 🤔 I think that empty user agent should actually throw if it's `null`. We shouldn't skip this, it's very important for analytics.

PR #7362, opened 2020-05-11
🐛 Fix initial sync failures

### LINKED ISSUE #7238: Sync fails on fresh login
## 🐛 Bug Report
When logging in to the app, the initial sync fails and the spider is displayed (the one with 'you haven't tracked time in a while' message). If you then pull to refresh, the TEs appear.

## Context

I've checked all Android releases back to v2.9. 2.9 is the last one that was working properly.

## Expected behavior
The initial sync to not fail and past time entries to appear.

## To reproduce
Logout of the app, login again.

pdrj:
### What's this?
This fixes initial sync failures (on most devices - more info below).

### Relationships

Closes #7238 

## How is it done?
By specifying a delay of the first background job.

### Why not another way?
The reason we got the bug in the first place is because on login, we do two syncs. One (pull TEs) is due to the background service. Another, full sync, has always been forced on login. For reasons unknown to me (and I've already spent a lot of time to try to get to the bottom of this), those two syncs one after the other, don't play along very well. What's more, they don't even make sense. If we run a full sync on login, why even bother with the pull one first.

I've looked into delaying the bg sync somehow, but:
- delaying it with a `Task.Delay` for instance, for several minutes, seems very wrong and we don't use the capabilities of `JobScheduler`
- making sure it runs after the first full sync requires some complicated changes.

That's why, in the end, I decided to go with what is provided by the framework. The caveat is that it's only api 24+, but since it covers 90%+ of users, I say it's OK. Not all users are even affected by this bug, since @semanticer was unable to reproduce it on his device, so we this PR might be fixing it for nearly everyone for all I know.

### Side effects
I hope none.

## Review hints
Try to repro as in the issue. If you get the buggy behavior, validate it doesn't happen on this branch. If you can't reproduce, just make sure the app works right after logging in.

## :squid: Permissions
Anyone


pdrj:
BTW: I did a run on iOS emulator and I didn't get the buggy behavior described in the issue.

PR #7369, opened 2020-05-12
🐎 Use higher default limits for leaky bucket

### LINKED ISSUE #7176: Mobile sync taking up to 10 seconds per sync
## ⛑️ User Support

Sync is taking up to 10 seconds for certain users. Android and iOS are both affected. This includes when you open the app, when you start a time entry, when you stop a time entry and when you pull down to manually sync.

## About the user

[Link to Google Drive sheet](https://docs.google.com/spreadsheets/d/1wkAWV_NYc_7IAGiWkR30y9Tg4GwRGvxvF2EaQTz6O_s/edit?usp=sharing)

## Expected behaviour

Sync would be a lot quicker 

## To Reproduce

[Please see linked video](https://drive.google.com/open?id=1pQVtm09PTiO5mcHmTID-ZzE3ppKI60md)
(This is my own video from my personal phone - no user data apart from my own)

pdrj:
### What's this?
This makes the burst size of the leaky bucket `10` instead of `3`. It also fixes a wrong condition when setting `burstsEnabled`.

### Relationships

Closes #7176 

## Why do we want this?
Sync is much faster. Way better UX.

## How is it done?
By sending more requests at once.

### Why not another way?
One more way discussed in slack (freeing slots when responses arrive) is very complicated and not guaranteed to be a solution.

### Related
https://toggl.slack.com/archives/C02APMYTM/p1588934950167400

## Review hints
Pull to refresh the main log on develop, observe that the spinner is visible for about `4s` at least. Then compare to this branch. Also, make sure the app works fine in a regular workflow and the changes make it to the web interface.

## :squid: Permissions
Anyone


daividssilverio:
So, the results of my tests on staging, purely empirical:

On develop:
- The first sync takes like 4~6 seconds
- Further pull to refresh syncs take from 7~10 seconds

On this branch:
- The first sync is definitely quicker, like 1~2 seconds
- Other pull to refresh syncs take from 2~6 seconds (depending on how far apart they are made)
- It feels slightly faster overall

On single pushes (simple changes to time entries, deletions), I didn't feel any difference.

PR #7379, opened 2020-05-12
🎨 Use a new default for custom project color

### LINKED ISSUE #7368: Have a pre-set default custom colour
## 🎯 What needs to be done?

Set `#3177BE` as the default colour when opening the custom colours menu.

pdrj:
###  What's this?
This uses a new default for custom project color

### Relationships

Closes #7368 

## Why do we want this?
It's required by the project.

## How is it done?
Hardcoding the default, as it was before :(

### Why not another way?
This is quite simple.

## Review hints
On a non-premium workspace, validate that you can pick a color from the list. On a premium workspace validate that the default custom color is the one from issue [and the pickers are setup on that color] and that you can select it and modify it.

## :squid: Permissions
Anyone


pdrj:
Yeah @dshr, that's what I got from the issue description 🤔. How should it work then?
-  Is it the case that the new custom color should always be preselected on the project creation page? Chances are that then users will not bother updating the preset color and end up with projects always having the same color (because there will be no more random choosing as we're doing now). 
- Should the custom color be selected just when they open the picker? I don't think that's a good idea - I think the picker should indicate what is currently selected in the project creation page. We display the 'tick' over the color and it'd be confusing to start with the 'tick' in a wrong place.

PR #7384, opened 2020-05-13
♻ Add new empty sync manager

### LINKED ISSUE #7383: Create a Sync Manager selector
## 🎯 What needs to be done?
A new sync manager selector needs to be created that will select an appropriate sync manager at the app start.

## 📝 Detailed explanation
The decision is based on the beta flag in the preferences payload. Additionally, to prevent being locked in a faulty sync-manager that can no longer fetch updates and thus revert to the old one, an additional API request needs to be done for the preferences and the local value should be updated so that on the next app start, the app will be able to chose the valid one again.

## 🏆 Outcome
The sync manager will correctly create a sync manager to be used in that session.

kornelijepetak:
## What's this?
Adds a new empty v2 sync manager and a sync manager selector that will choose the proper sync manager on app start.

Parallel to the selection, an API call to fetch up-to-date preferences is emitted. Check the in-code comment for explanation of its purpose.

### Relationships
Closes #7383

## Why do we want this?
So that we can slowly rollout the new sync manager.

## How is it done?
By the magic of choice.

### Why not another way?
I've spent toooooo much time doing this in different ways, in the end I just traded off some flexibility (being able to switch during runtime) for a lot of simplicity.

### Side effects
If sync manager really breaks down, sync for that user will work only from the next app start (after the beta flag is reset).

## Review hints
Commit by commit should really be straight forward.

## :squid: Permissions
Me and @simonrozsival.


Toggl.Core/Sync/v2/SyncManagerSelector.cs
```diff
@@ -0,0 +1,64 @@
...
+            IPreferencesApi preferencesApi,
+            Func<ISyncManager> oldSyncManagerCreator,
+            Func<ISyncManager> newSyncManagerCreator)
+        {
+            Ensure.Argument.IsNotNull(interactorFactory, nameof(interactorFactory));
+            Ensure.Argument.IsNotNull(preferencesApi, nameof(preferencesApi));
+            Ensure.Argument.IsNotNull(oldSyncManagerCreator, nameof(oldSyncManagerCreator));
+            Ensure.Argument.IsNotNull(newSyncManagerCreator, nameof(newSyncManagerCreator));
+
+            var preferences = interactorFactory.ObserveCurrentPreferences().Execute().GetAwaiter().Wait();
```
simonrozsival, line 24: This will block until the preferences are fetched from the DB? That could make launching the app noticeably slower. I'll try to think about a better way of doing this, but I guess it could be fine. Maybe some other dev can suggest some other way of doing it.

Toggl.Core/Sync/v2/SyncManagerSelector.cs
```diff
@@ -0,0 +1,64 @@
...
+            EditPreferencesDTO preferencesDto;
+
+            try
+            {
+                var preferences = await preferencesApi.Get();
+                preferencesDto = new EditPreferencesDTO { UseNewSync = preferences.UseNewSync };
+            }
+            catch
+            {
+                preferencesDto = new EditPreferencesDTO { UseNewSync = false };
```
simonrozsival, line 58: This will cause the app to revert to the old sync manager every time the app is offline. Is that the behavior you had in mind?

Toggl.Core/Sync/v2/SyncManagerSelector.cs
simonrozsival: Could we make this method `async` and push blocking one level above (where we might be able to await it somehow)?

Toggl.Core/Sync/v2/SyncManagerV2.cs
```diff
@@ -0,0 +1,31 @@
...
+using System.Reactive.Linq;
+
+namespace Toggl.Core.Sync.V2
+{
+    public class SyncManagerV2 : ISyncManager
+    {
+        public bool IsRunningSync => false;
+
+        public IObservable<SyncProgress> ProgressObservable
+            => Observable.Return(SyncProgress.Synced);
```
simonrozsival, line 11: Maybe we can always return `Failed` until it's implemented? 😄

Toggl.Networking.Tests/Models/PreferencesTests.cs
```diff
@@ -30,13 +37,35 @@ public void HasConstructorWhichCopiesValuesFromInterfaceToTheNewInstance()
         [Fact, LogIfTooSlow]
         public void CanBeDeserialized()
         {
-            SerializationHelper.CanBeDeserialized(validJson, validPreferences);
+            SerializationHelper.CanBeDeserialized(getValidJson(), validPreferences);
```
simonrozsival, line 40: 👀 `getValidJson(useSync: false)` --> but `validPreferences.UseNewSync == true`

Toggl.Networking/Models/Preferences.cs
```diff
@@ -20,5 +25,10 @@ internal sealed partial class Preferences : IPreferences
 
         [JsonProperty("CollapseTimeEntries")]
         public bool CollapseTimeEntries { get; set; }
+
+        [JsonProperty("alpha_features")]
+        [JsonConverter(typeof(AlphaFeaturesJsonConverter), newSyncClient, false)]
+        public bool UseNewSync { get; set; }
+        public bool ShouldSerializeUseNewSync() => false;
```
simonrozsival, line 32: The `[IgnoreWhenPosting]` attribute would be more elegant, check `Project.cs`.

Toggl.Networking/Serialization/AlphaFeaturesJsonConverter.cs
```diff
@@ -0,0 +1,39 @@
...
+                return value;
+            }
+            catch
+            {
+                return defaultValue;
+            }
+        }
+
+        public override void WriteJson(JsonWriter writer, bool value, Serializer serializer)
+        {
```
simonrozsival, line 36: I guess this should throw 😄. Plus maybe this could have `bool CanWrite => false`? (https://www.newtonsoft.com/json/help/html/P_Newtonsoft_Json_JsonConverter_CanWrite.htm)

Toggl.Core/Sync/v2/SyncManagerSelector.cs
kornelijepetak: No. Actually, this was my initial idea, but the way the dependency container works, it wouldn't be a simple thing to do.

I've talked to William about it, and considering the fact that this is a single call to a _table_ that has a single element, the thing would happen really fast.

So, let's not optimize prematurely, and let's see how it behaves during testing.

Toggl.Core/Sync/v2/SyncManagerSelector.cs
kornelijepetak: 🤔 hm... good point. I'll think about it.

Toggl.Core/Sync/v2/SyncManagerV2.cs
kornelijepetak: I see, you are a pessimist and I am an optimist 😄

Toggl.Networking.Tests/Models/PreferencesTests.cs
kornelijepetak: I'm just fixing this test. This way of testing json is not really well designed.

The proper way is to roundtrip serialization and then test for existence of properties, not by hardcoded strings. 

But I'll find a way to make this work.

Toggl.Core/Sync/v2/SyncManagerSelector.cs
simonrozsival: Alright!

PR #7394, opened 2020-05-14
📦 Add push request id to the storage

### LINKED ISSUE #7321: Store the ID of the current outstanding push request
We need to:
- generate a new random UUID for a push request,
- store the ID,
- check if there is an outstanding push request,
- ... and be able to remove it once we're finished.

We'll put it into Realm instead of the key-value storages similarly to since dates.

kornelijepetak:
## What's this?
Adds Push Request Identifier into the storage. 
Allows getting, setting and updating the ID.

### Relationships
Closes #7321

## Why do we want this?
So that the sync manager can utilize this information when sending out the request.

## How is it done?
Since there's only one item expected in that _table_, an `onlyId` of `1` is used.

### Why not another way?
I utilized the same way the `since parameter storage` works for ease of use. But ours will contain only a single item in the _table_.

### Side effects
None.

## Tests
Is there anything particularly worth testing? I don't see that database access is tested in any other class. Only certain _business logic_ is (such as `since parameter storage` type being checked etc.)

## :squid: Permissions
Only the ♻ crew.


Toggl.Storage.Realm/Models/RealmPushRequestIdentifier.cs
```diff
@@ -0,0 +1,32 @@
...
+        public RealmPushRequestIdentifier()
+        {
+        }
+
+        public RealmPushRequestIdentifier(IDatabasePushRequestIdentifier entity)
+        {
+            SetPropertiesFrom(entity, null);
+        }
+
+        public void SetPropertiesFrom(IDatabasePushRequestIdentifier entity, RealmDb realm)
```
attilio-work, line 26: `realm` is not used here, is this necessary? Or is this an overridden method?

Toggl.Storage.Realm/PushRequestIdentifierStorage.cs
```diff
@@ -0,0 +1,78 @@
...
+        {
+            lock (storageAccess)
+            {
+                try
+                {
+                    var record = realmAdapter.GetAll().FirstOrDefault();
+                    if (record == null)
+                        return;
+
+                    realmAdapter.Delete(onlyId);
```
attilio-work, line 34: Why does this have only the single ID?

Toggl.Storage.Realm/PushRequestIdentifierStorage.cs
```diff
@@ -0,0 +1,78 @@
...
+                }
+                catch
+                {
+                }
```
attilio-work, line 38: What is the reason for this empty catch? Shouldn't we at least log the error?

Toggl.Storage.Realm/PushRequestIdentifierStorage.cs
```diff
@@ -0,0 +1,78 @@
...
+        {
+            lock (storageAccess)
+            {
+                try
+                {
+                    return realmAdapter.Get(onlyId).PushRequestId;
+                }
+                catch
+                {
+                    return null;
```
attilio-work, line 52: We won't know if we are getting `null` due to some error, or because `onlyID` really does not exist...

Toggl.Storage.Realm/PushRequestIdentifierStorage.cs
```diff
@@ -0,0 +1,78 @@
...
+
+            lock (storageAccess)
+            {
+                try
+                {
+                    realmAdapter.Update(onlyId, record);
+                }
+                catch (InvalidOperationException)
+                {
+                    realmAdapter.Create(record);
```
attilio-work, line 73: I know this works, but... it seems a bit ugly to me... Since we are locking anyway, wouldn't it be possible to just check if `onlyId` exists already or not?

Toggl.Storage.Realm/Models/RealmPushRequestIdentifier.cs
kornelijepetak: Yes, it is an interface method so it has to be there. Some implementations may use it, but some do not need it. This one does not. :)

Toggl.Storage.Realm/PushRequestIdentifierStorage.cs
kornelijepetak: yes, precisely because of that.

Toggl.Storage.Realm/PushRequestIdentifierStorage.cs
kornelijepetak: We could log the error, but it's gonna be somewhere on somebody's device and we won't have it anyway... at least with current setup. So I don't think it's an issue for us. But considering that it can only fail if IO derps (as it only calls into realm db), maybe we should actually let it crash the app (as per mobile guidelines) so that we get the error and the stack trace on appcenter.

any comments on this @heytherewill or @simonrozsival?

Toggl.Storage.Realm/PushRequestIdentifierStorage.cs
simonrozsival: `Delete` should not throw anyway: https://realm.io/docs/dotnet/3.0.0/api/reference/Realms.Realm.html#Realms_Realm_Remove_Realms_RealmObject_

Toggl.Storage.Realm/PushRequestIdentifierStorage.cs
simonrozsival: Delete should not throw anyway: https://realm.io/docs/dotnet/3.0.0/api/reference/Realms.Realm.html#Realms_Realm_Remove_Realms_RealmObject

Toggl.Storage.Realm/PushRequestIdentifierStorage.cs
simonrozsival: The only reason why `Get` should throw is that there is no entity with the ID (internaly it uses Linq `Single`), so this is fine.

Toggl.Storage.Realm/PushRequestIdentifierStorage.cs
simonrozsival: This approach has worked quite well for us for storing `SinceParameter` (check the `SinceParameterStorage` class).

kornelijepetak: Yeah, if I had to implement this from my own mind, I'd go a bit different, but I was shamelessly just copying the behavior of a similar class so that we can be faster with it. Especially in the light of the soon deprecation of the app.

Issue #7280, opened 2020-04-06
Create models for push actions


kornelijepetak:
## 🎯 What needs to be done?
There are three actions that need JSON-serializable models:

* Create action
* Update action
* Delete action

Make those models according to sync API format.

## 🏆 Outcome
The models exist that allow serialization to JSON.

kornelijepetak:
Closed by #7311

PR #7413, opened 2020-05-22
Refactor and fix PUSH request serialization

simonrozsival:
## What's this?
While implementing the code for the `/push` endpoint integration, I found out that the serialization does not work well in the iOS simulator. I started debugging and changing the code and I arrived at this final state.

Although in the end the changes could have been smaller (we could still use some of the classes which I deleted), I think that this PR makes the code simpler, so I would keep this refactoring.

### Relationships
Ref #7280 

## Why do we want this?
We want a working and simple `sync-server` integration.

## How is it done?
The most important part of this PR are the `[Preserve(AllMembers = true)]` attributes. Without these attributes, the serialized JSON would be an empty object, because all the properties are stripped by the linker.

The second important change is the renaming of `DeleteMeta -> ModificationMeta` and its useage not only for delete actions, but also for updates.

The remaining changes might be unnecessary, but I think they are worth it. In the end, we don't need three classes for `create`, `udpate`, and `delete` actions, we need just one `PushAction` class with factory methods.

## :squid: Permissions
Me and @kornelijepetak


Toggl.Networking/Sync/Push/Request.cs
```diff
@@ -1,99 +1,183 @@
...
-            timeEntries
-                .Select(bte => new UpdateAction<ITimeEntry>(bte))
+            var list = timeEntries.ToList();
+            if (list.Count == 0)
+                return;
+
+            if (TimeEntries == null)
+            {
+                TimeEntries = new List<PushAction>();
+            }
```
kornelijepetak, line 82: This can be used on all of similar places in this file.

```suggestion
            TimeEntries ??= new List<PushAction>();
```

PR #7424, opened 2020-05-25
♻ Sync basics

simonrozsival:
This PR contains some changes I want to share with @kornelijepetak for the development of the new sync framework (notice that this PR targets the `sync-team/sync` branch).

🦑 just me or @kornelijepetak


Toggl.Core/Sync/v2/SyncManagerV2.cs
```diff
@@ -1,33 +1,154 @@
...
         public IObservable<SyncState> ForceFullSync()
-            => Observable.Return(SyncState.Sleep);
+            => sync()
+                .ToObservable(schedulerProvider.DefaultScheduler)
+                .SelectValue(SyncState.Sleep);
 
         public IObservable<SyncState> Freeze()
-            => Observable.Return(SyncState.Sleep);
+        {
+            lock (mutex)
```
kornelijepetak, line 58: I see your GO experience is making you write weird variable names 😄 

`mutex` sounds like a type name. So, I'd go with `lock(access)`, but considering that the sync manager is actually a singleton service, `lock(this)` would work as well.

Toggl.Core/Sync/v2/SyncManagerV2.cs
```diff
@@ -1,33 +1,154 @@
...
+                lock (mutex)
+                {
+                    isRunningSync = false;
+                }
+            }
+        }
+
+        private void processError(Exception error)
+        {
+            // shamelessly copy&pasted from SyncManager V1
```
kornelijepetak, line 122: There is no shame in copy pasting from another class if it does exactly the same thing.

Don't tell me you even considered the alternative of introducing a `SyncManagerBase` class 😄 .

simonrozsival: No, I didn't.. yuck :D

PR #7441, opened 2020-05-28
Add Sync Pulling

### LINKED ISSUE #7283: Implement a 3-way merge with conflict resolution
## 🎯 What needs to be done?
Implement the per-property 3-way merge with conflict resolution.

## 🛠 Technical remarks
This is an epic issue and this needs a bit more discussion.

The proper 3-way merge:

### LINKED ISSUE #7440: Add Sync Pull
## 🎯 What needs to be done?
A database query needs to be created and an interactor that will call it that will perform the sync pull.

## 📝 Detailed explanation
Sync pull needs to gets the `all` data from the sync server and perform an update of the local state by performing a conflict resolution on the time entries. It also needs to make sure to stop any locally running time entries if the payload contains a running time entry.

## 👕 Estimated complexity
`Large`

kornelijepetak:
## What's this?
Adds pulling support to sync.

### Relationships
Closes #7440 
Closes #7283 

## Why do we want this?
So that we can pull data from the server.

## How is it done?
Data is retrieved from the API. Then all of the entities in the payload are used to update local state, additionally performing 3way merge on the Time Entries.

## Review hints
As always, commit by commit is the easiest, especially so in this PR.

## :squid: Permissions
Me and @simonrozsival


Toggl.Storage.Realm/Models/RealmUser.cs
```diff
@@ -49,5 +50,18 @@ public void PushFailed(string errorMessage)
             SyncStatus = SyncStatus.SyncFailed;
             LastSyncErrorMessage = errorMessage;
         }
+
+        public void SaveSyncResult(IUser entity, Realms.Realm realm)
```
simonrozsival, line 54: In this method, you must also set `SyncStatus` to `InSync` and the last error message to `null`.

kornelijepetak: Done.

Toggl.Storage.Realm/Queries/ProcessPullRequestQuery.cs
```diff
@@ -0,0 +1,242 @@
...
+                if (dbTimeEntry == null)
+                {
+                    if (timeEntry.ServerDeletedAt.HasValue)
+                        continue;
+
+                    addTimeEntry(timeEntry, realm);
+                    continue;
+                }
+
+                if (incomingTimeEntryIsRunning && dbTimeEntry.IsRunning())
```
simonrozsival, line 102: This is not correct. The currently running time entry locally might not be among those pulled from the server (`timeEntries`) and you will end up with two running TEs after all. You need to query `realm.All<RealmTimeEntry>().Where(te => te.Duration == null)` and stop all those (except for the pulled running one).

Toggl.Storage.Realm/Queries/ProcessPullRequestQuery.cs
```diff
@@ -0,0 +1,242 @@
...
+        }
+
+        private void addTimeEntry(ITimeEntry timeEntry, RealmDb realm)
+        {
+            var dbTimeEntry = new RealmTimeEntry();
+            dbTimeEntry.SaveSyncResult(timeEntry, realm);
+            realm.Add(dbTimeEntry);
+        }
+
+        private void updateTimeEntry(ITimeEntry timeEntry, RealmTimeEntry dbTimeEntry, RealmDb realm)
```
simonrozsival, line 119: You must also set the `SyncState` and the last error message to `null`.

Toggl.Storage.Realm/Sync/ThreeWayMerge.cs
```diff
@@ -0,0 +1,29 @@
+﻿using System;
+
+namespace Toggl.Storage.Realm.Sync
+{
+    public static class ThreeWayMerge
+    {
+        public static T Merge<T>(T common, T local, T server) where T : IEquatable<T>
+        {
+            if (local.Equals(common))
```
simonrozsival, line 9: I think we could make this method a bit more readable to make sure that it serves as a good reference when it will be re-implemented by introducing a variable - something like `var localHasNotChanged = local.Equals(common);`.

Toggl.Storage.Realm/Sync/ThreeWayMerge.cs
```diff
@@ -0,0 +1,29 @@
...
+                return server;
+
+            if (server.Equals(common))
```
simonrozsival, line 12: Same thing here (`var serverHasNotChanged = server.Equals(common);`)

Toggl.Storage.Realm/Queries/ProcessPullRequestQuery.cs
```diff
@@ -0,0 +1,242 @@
...
+                dbEntity.SaveSyncResult(serverEntity, realm);
+                realm.Add(dbEntity);
+            }
+            else
+            {
+                dbEntity.SaveSyncResult(serverEntity, realm);
+            }
+        }
+
+        private bool identifierComparison(long? a, long? b) => a == b;
```
simonrozsival, line 238: Nullable aren't `IEquatable<>`? Interesting.

Toggl.Storage.Realm/Queries/ProcessPullRequestQuery.cs
```diff
@@ -0,0 +1,242 @@
...
+                    if (entity.ServerDeletedAt.HasValue)
+                        continue;
+
+                    dbEntity = new TRealmEntity();
+                    dbEntity.SaveSyncResult(entity, realm);
+                    realm.Add(dbEntity);
+                }
+                else
+                {
+                    dbEntity.SaveSyncResult(entity, realm);
```
simonrozsival, line 216: Here you might have to delete the entity, if `entity.ServerDeletedAt.HasValue` the same way you did for TEs.

Toggl.Storage.Realm/Models/RealmPreferences.cs
```diff
@@ -49,5 +50,14 @@ public void PushFailed(string errorMessage)
             SyncStatus = SyncStatus.SyncFailed;
             LastSyncErrorMessage = errorMessage;
         }
+
+        public void SaveSyncResult(IPreferences entity, Realms.Realm realm)
```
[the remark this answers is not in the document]
kornelijepetak: Fixed.

Toggl.Storage.Realm/Queries/ProcessPullRequestQuery.cs
kornelijepetak: 🤦‍♂️ of course. What was I thinking?!

Toggl.Storage.Realm/Queries/ProcessPullRequestQuery.cs
kornelijepetak: Yeah, for some reason, they did not implement it 🤷‍♂️

Toggl.Storage.Realm/Sync/ThreeWayMerge.cs
kornelijepetak: Is this really necessary?! I mean `local.Equals(original)` basically reads the same as `local has not changed`... I find your suggestion really superfluous. And if one knows anything about a 3way merge (which I assume that all git-using devs should), I don't think there's anything confusing or unreadable in this code as it is...

That being said, I did consider changing `common` into the `original` or `base` (although `base` is a keyword, so it'd have to be `@base`).

simonrozsival: Renaming would be a good solution as well 👍

simonrozsival:
(damn, sorry for messing with your branch resolving those conflicts! for some reason I completely forgot this is not my _push_ sync PR which was already merged...)

Toggl.Storage.Realm/Queries/ProcessPullRequestQuery.cs
```diff
@@ -99,16 +99,28 @@ private void processTimeEntries(RealmDb realm)
...
+            var time = currentTimeProvider();
+
+            var runningTimeEntries = realm.All<RealmTimeEntry>()
+                .Where(te => te.Duration == null)
+                .ToList();
+
+            foreach (var runningTimeEntry in runningTimeEntries)
+            {
+                var duration = (int)(time - runningTimeEntry.Start).TotalSeconds;
+                runningTimeEntry.DurationBackup = runningTimeEntry.Duration = duration;
```
simonrozsival: These changes should be pushed next time we sync - change the sync status to `SyncNeeded` as well please.

simonrozsival: Also should we really set the `DurationBackup` here? It's just the local change, it might be different on the server and we might want to change the duration in conflict resolution.

Toggl.Storage.Realm/Queries/ProcessPullRequestQuery.cs
```diff
@@ -83,8 +83,8 @@ public Unit Execute()
 
         private void processTimeEntries(RealmDb realm)
         {
-            var incomingTimeEntryIsRunning = timeEntries.Any(te => te.IsRunning());
-            var time = currentTimeProvider();
+            if (timeEntries.Any(te => te.IsRunning()))
+                stopLocallyRunningTimeEntries(realm);
```
simonrozsival: I have to think this through... can't this be messed up during conflict resolution? 🤔 In the old sync algorithm, we stop the second running TE after we persist all the pulled TEs.

Toggl.Storage.Realm/Queries/ProcessPullRequestQuery.cs
kornelijepetak: These are the local running time entries (even though there shouldn't be more than one, assume that somehow it happens). If they are running time entries, that means that they were created with the `Duration` of `null`, and that will also be the case for the backup. The only way for the `DurationBackup` and `Backup` to be different is if then the time entry is stopped so that the Duration of `null` would be copied into the `DurationBackup` and then the `Duration` would be set to the stop time. As long as the `Duration` is `null`, it is guaranteed that the `DurationBackup` would have the same value.

Now, if a time entry from the server comes as running, it is either
* the same time entry or
* another time entry.

If it's the same time entry, and it is still running, the 3way merge would not really change this at all, as all three values are `null`. If it was stopped, this code here will effectively stop it (albeit with a different time). Since the conflict resolution would happen afterwards (at least in my impl), what we'll have here is a 3way merge of (original: `now`, local: `now`, server: `some time`), which will cause the `now` to be overwritten with the server's value. That shows that this _has_ to happen after the resolution. 

Let's switch the story here and assume from now on that the stopping of potentially multiple running time entries will happen after the pulled TEs are persisted.

Let's now consider the posibilities.

If a local TE is running, its both properties are `null`.

**Scenario 1.** The payload running TE is also locally running TE.

The conflict resolution, would be (original: `null`, local: `null`, server: `null`), leaving both values at `null`, so this does not change the semantics and is thus correct. 

**Scenario 2.** Another TE is running locally, when the payload running TE comes

We need to stop local TE, but we have two subscenarios:

**Scenario 2.1** - Payload **does not contain** this TE at all.

That means that the 3way merge will not happen at all, and the only action we need to do is set the `Duration` to a correct value as `now - start`.

**Scenario 2.2** - Payload **does contain** this TE, but it wasn't running on server

The 3way merge will combine `(O: null, L: null, S: duration)` overwriting the values of both properties with server-based duration.

But that means that after the 3way merge is done, we can no longer fetch all of the running time entries as this one will no longer be running. So, that means, that we must `fetch` all locally running entries prior to the merge, and remember which those were. Then perform the merge with conflict resolution, and only then from the remembered list update the duration with a correct value depending on current time.

Toggl.Storage.Realm/Queries/ProcessPullRequestQuery.cs
kornelijepetak: Question answered in that big wall of text called a comment, a bit up above.

Toggl.Storage.Realm/Queries/ProcessPullRequestQuery.cs
simonrozsival: It makes sense and it will work.

I thought of this differently and it seemed straightforward to me this way:
1. fetch all TEs X and save them using 3WM
2. if there was a running TE A in X, select all the TEs Y which are running, but are not X
3. stop every TE B in Y by setting duration
4. set `SyncStatus` to `SyncNeeded` to make sure it propagates to the server in the next Push

To me, it's immediately obvious that this alg is correct and Pull won't break the invariant of at most one running TE. Your algorithm works as well, but it requires knowledge of how the conflict resolution works on the lowest level.

But ultimately it's up to you 😄 If you go for the solution you proposed, just please include a comment in the code to make it easier to understand. We're building a reference implementation for the mobile team after all.

kornelijepetak: I think we have described the same thing, it's just that your algorithm does not take into account in the step 1. that the 3WM will (depending on the state of the payload) smear the local information on what time entries are running and which are not. 

i.e. if TE(A) is a local running time entry, and a TE(A) comes from the server as a _stopped_ TE, 3WM will turn this time entry into a stopped time entry locally with duration corresponding to the server value (because of the server-wins nature of the 3WM). That's why the idea of `TEs Y` in step 2 would not work in cases when payload contained those TEs (but in a stopped state). During 3WM they would be stopped with a duration dictated by the server value, not by the current time.

So, the order of these operations is not really valid, but your whole idea is valid and is basically what I have described.

simonrozsival: > During 3WM they would be stopped with a duration dictated by the server value, not by the current time.

But that's actually what would be the preferred thing to do in most cases IMO:
- user stops a TE A in the webapp
- user starts a TE B in the webapp
- 5 minutes later user goes to the mobile app and TE A is still running here (let's assume a PN did not arrive or it was not processed for some reason, which happens from time to time)
- the TE A should be stopped and the duration from the server is the desired one
- TE B will be running, no problem there (it's a new TE)

I think that A would have the server time even in your version of the algorithm:
- we'll set the duration according to the current time
- the 3WM will overwrite the local data with the server change (`server != original`)

simonrozsival: Also if there aren't any two running TEs after the regular conflict resolution, then everything is great, it resolved itself. The only case when you have two running TEs and it doesn't resolve itself is when you create a TE in the mobile app/any client app, it isn't pushed to the server, and you create a new running TE on the server. In this case it won't resolve itself during the conflict resolution and it will require you to set the `Duration` manually.

kornelijepetak: But I understood the other day you that you wanted the `current time` to win over server value.

So, if a time entry is started on client A, then started (and pushed to server) on client B, then pulled into client A (from server), I understood you to want that time entry's duration to be calculated from current time, not overwritten by the server (which is basically what was done in client B).

So what are we finally doing? Server wins, or local time wins?

simonrozsival: In case of stopped TEs the server should always win. I think that's currently the weakest point of the mobile app's sync - if you stop the TE on the mobile after you already stopped it in the webapp, you lose the original stop time (which is probably the correct stop time from the user's perspective).

There was probably a misunderstanding between us. During our call the other day, I talked about setting the stop time to the start time of the TE which follows to fill the gap and to avoid TE overlap (a "smart" way of stopping the TE which we're using currently) and that we don't have to do it and that we can just set the current time as the stop time the same way other clients do it (at least the webapp).

Issue #7429, opened 2020-05-26
Add unit tests for Push sync


simonrozsival:
In order to move fast, I created a PR without unit tests for the newly added interactors and DB query. In order to not break things, we should have unit tests before we merge the code into `develop` and create builds for our users.

simonrozsival:
There are already unit tests for the `PushSyncInteractor` and `ResolveOutstandingPushRequestInteractor` classes but there's no test for the `ProcessPushResultQuery` which might be challenging (it includes an in-memory Realm instance). I will look into this after the first internal release is out.

PR #7445, opened 2020-05-28
Push sync interactor tests

simonrozsival:
## What's this?
Unit tests for the `PushSyncInteractor`.

### Relationships
Ref #7429  - it doesn't close it, because I'll add unit tests also for `ResolveOutstandingPushRequest` in a separate PR. I might also add a test for the `ProcessPushResultQuery` in a third PR, if testing in-memory Realm proves doable and reliable.

## Why do we want this?
We want tests 😄 

## :squid: Permissions
🦑


Toggl.Core.Tests/Interactors/Sync/PushSyncInteractorTests.cs
```diff
@@ -33,6 +39,177 @@ public sealed class TheConstructor : BaseInteractorTests
...
+            [Fact, LogIfTooSlow]
+            public async Task StoresResultsInTheDatabase()
+            {
+                var res = Substitute.For<IResponse>();
+                var throwingTask = Task.FromResult(res);
+                Api.SyncApi.Push(Arg.Any<Guid>(), Arg.Any<Request>()).Returns(throwingTask);
+
+                await interactor.Execute();
+
+                QueryFactory.Received().ProcessPushResult(Arg.Is(res));
```
kornelijepetak, line 90: I think this works normally like this as well:

```suggestion
                QueryFactory.Received().ProcessPushResult(res);
```

simonrozsival: In this case I prefer using `Arg.Is`, let's keep it 😄

